PR #25, opened 2014-01-14
Allow use of derived classes and add missing argument

vasole:
Unless one can recover the last event handled by the PlotCurveItem or the ScatterPlotItem, it is impossible to recover things like what mouse button was pressed and so on. The proposed modification allows to use derived classes of PlotDataItem, PlotCurveItem and ScatterPlotItem to handle events in a different manner.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -123,8 +123,8 @@ class PlotDataItem(GraphicsObject):
         #self.dataMask = None
         #self.curves = []
         #self.scatters = []
-        self.curve = PlotCurveItem()
-        self.scatter = ScatterPlotItem()
+        self.curve = self._getPlotCurveItem()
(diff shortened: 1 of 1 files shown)


2xB:
Is this necessary to achieve custom classes? You could also do:

ˋˋˋpython
yourplotdataitem.curve.__class__ = CustomPlotCurveItem
yourplotdataitem.scatter.__class__ = CustomScatterPlotItem
ˋˋˋ
, where ˋCustomPlotCurveItemˋ extends ˋPlotCurveItemˋ and ˋCustomScatterPlotItemˋ extends ˋScatterPlotItemˋ.

To the missing argument: PyQt allows you to receive events without their parameters.

2xB:
The last comment should hopefully work as seamless as the Pull Request aims at. If it does not or if you have any further comment, feel free to post here! I will of course also reopen this Pull Request if wished.
Thank you!

PR #45, opened 2014-03-05
Make the bars that are clickable visible

zarch:


ixjlyons:
As mentioned this change shouldn't be necessary (I would expect the default cosmetic arguments to result in visible items). Closing as part of an effort to address old PRs/issues.

PR #46, opened 2014-03-05
Add BarWidget

zarch:
Please check that the widget is correct.


j9ac9k:
Going to close this out due to the suggestion of just extending PlotWidget.  Thanks for the PR @zarch

PR #49, opened 2014-03-06
PyQt5 support for PyQtGraph

mfitzp:
Here is a pull-request for the PyQt5 support in pyqtgraph. Due to the API differences between Qt4 and Qt5 I've opted to use the existing PySide/PyQt4 wrapper `Qt.py` to implement a wrapper for PyQt5, re-implementing deprecated APIs in use in pyqtgraph and moving classes back to their previous places in the Qt namespace.

The approach to getting all the moved classes back into the correct namespace is as follows:

    # Import all QtWidgets objects into QtGui
    for o in dir(QtWidgets):
        if o.startswith('Q'):
            setattr(QtGui, o, getattr(QtWidgets,o) )

This is quite horrible, but I could not think of a nice way to do it. If we want to avoid shifting the lot a list of all QtGui objects used in the code that are now in QtWidgets would be needed (volunteers). This could be avoided entirely by importing the entire Qt namespaces within the Qt wrapper, e.g.

    from QtGui import *

...in this wrapper, then....

    import Qt
    Qt.<QtClass> 

...in the pyqtgraph source. But that requires a bit overhaul elsewhere so I avoided it in favour of a drop-in replacement. However, the import and shift doesn't appear to take a noticeable amount of time.

Deprecated APIs are re-implemented using Python versions of the suggested replacement code available in the Qt5 release notes. I've implemented a bare-minimum to get Qt5 working, but others are very easy to add.

    # Re-implement deprecated APIs
    def scale(self, sx, sy):
        self.setTransform(QtGui.QTransform.fromScale(sx, sy), True)
    QtWidgets.QGraphicsItem.scale = scale

As it stands, the implementation here works on all existing demos, except the OpenGL ones and a couple that directly import Qt (rather than using the wrapper) - these may be bugs in the demos rather than the wrapper, will need a look. I'm not familiar enough with how the demos are "supposed" to look under PyQt4 to be able to confirm that they are all working as expected though there are no errors.

Comments and suggestions welcome!

### DIFF
--- a/pyqtgraph/Qt.py
+++ b/pyqtgraph/Qt.py
@@ -121,10 +121,18 @@ elif USE_QT_PY == PYQT5:
         self.setRotation(self.rotation() + angle)
     QtWidgets.QGraphicsItem.rotate = rotate
 
+    def translate(self, dx, dy):
+        self.setTransform(QtGui.QTransform.fromTranslate(dx, dy), True)
+    QtWidgets.QGraphicsItem.translate = translate
 
     def setMargin(self, i):
-        self.setContentsMargins( i, i, i, i)
+        self.setContentsMargins(i, i, i, i)
     QtWidgets.QGridLayout.setMargin = setMargin
+
+    def setResizeMode(self, mode):
+        self.setSectionResizeMode(mode)
+    QtWidgets.QHeaderView.setResizeMode = setResizeMode
+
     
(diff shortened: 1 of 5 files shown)


campagnola:
Many thanks!

soupault:
Any dates on 0.9.11 release?

PR #64, opened 2014-05-10
Small Improvements for LegendItem

CPrescher:
Added 3 new parameters LegendItem:
horSpacing: Specifies the spacing between the line symbol and the label.
verSpacing: Specifies the spacing between individual entries of the
legend vertically. (Can also be negative to have them really close)
box: Specifies if the Legend should will be drawn with a rectangle
around it.

This made it possible to change the resize behavior by just using
layout.setVerticalSpacing and layout.setVerticalSpacing.

Changed the removeItem function to accept either name or the originally
added Item. I found the latter more useful, especially if several items
happen to have the same name.

Changed the rows in which addItem inserts a new item. Now the number
increases all the time. This was needed because there where some
overwriting issues (it did want to write in a row were already an item was existent, in the previous version) when using several times removeItem and addItem on
the same legend (in random order). The GraphicsGridLayout does not care
about empty rows. I found this approach to work best.

### DIFF
--- a/pyqtgraph/graphicsItems/LegendItem.py
+++ b/pyqtgraph/graphicsItems/LegendItem.py
@@ -19,7 +19,7 @@ class LegendItem(GraphicsWidget, GraphicsWidgetAnchor):
         legend.setParentItem(plotItem)
 
     """
-    def __init__(self, size=None, offset=None):
+    def __init__(self, size=None, offset=None, horSpacing = 25, verSpacing=0, box=True):
         """
         ==============  ===============================================================
         **Arguments:**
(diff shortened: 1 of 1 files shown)


rkoyama1623:
Hello, I'm a master student in Japan.
I want this PR to be merged too.
Could you rebase devel branch? @CPrescher 

In this PR, `LegendItem.items` is renamed to `LegendItem.legenditems`, 
but I don't think this renaming is good decision in terms of backward compatibility.

Also, I think the options used in `LegendItem.__init__` should be given in `PlotItem.addLegend`'s keyword arguments.

Regards,

SamSchott:
I just had a look at this and really like the new parameters `horSpacing ` and `verSpacing` and the simpler control of the legend size that comes with them. A few points:

- Why is it necessary to keep track of the number of rows in `numItems`? As far as I understand, adding new items to the row number returned by `layout.rowCount()` works just as well (see Qt docs [here](https://doc.qt.io/qt-5/qgraphicsgridlayout.html#rowCount)). Or is there any case where `not numItems == layout.rowCount()`?
- Why do we need to track `plotItems` separately? You can always access them through `sample.item`.
- The current implementation of `removeItem` already accepts both the actual plot item and its name. No need for changes (which only complicate things). Or am I missing something?

Finally, I would suggest new parameters `brush` and `pen` for the background and border of the legend, respectively, instead of the boolean parameter `box`. They can be set to `None` by default, resulting in no box being drawn. This seems to be more in line with the coding style of `pyqtgraph` but may be less intuitive. 

What do you think? I have incorporated in some of this PR [here](https://github.com/SamSchott/pyqtgraph/tree/patch-2). If there is sufficient interest, I can submit a new pull request from that branch.

j9ac9k:
@CPrescher thanks for this PR, sorry it's taken us so long to properly follow up.  I'm going to close this as #958 took components of this PR and added more on top of it, but still wanted to thank you for your contribution!

PR #65, opened 2014-05-14
A table parameter tree item

YiQinHuili:
Subclass of WidgetParameterItem, used to edit a numpy record array


outofculture:
It's been a while, and ParameterTree has changed a fair amount since this PR was submitted. We're having trouble justifying this feature now, in light of the new features that have been implemented. If this PR is still relevant to someone, can you describe a use-case and double-check that the current library doesn't already address your needs?

PR #67, opened 2014-05-14
Fix scale-rotate ROI handler to allow to use a random scale axis and not only x/y

vallsv:
This patch allow to use scale-rotate handler without orthogonal center.

```
    1-----2
    |     |
    |     |
    |     |
    4-----3
```

On my application the handler 1 use the center 3, the handler 2 use the center 4...
Without this patch it is not possible.

```
            self.addScaleRotateHandle(pos=(0, 1), center=(1, 0))
            self.addScaleRotateHandle(pos=(1, 0), center=(0, 1))
            self.addScaleRotateHandle(pos=(0, 0), center=(1, 1))
            self.addScaleRotateHandle(pos=(1, 1), center=(0, 0))
```

Unfortunatly i think this patch break the code relative to scaleSnap

Then are you interested by this patch? Else can we create another handler type too allow this kind of scale-rotation?

Thanks a lot

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -466,8 +466,6 @@ class ROI(GraphicsObject):
         """
         pos = Point(pos)
         center = Point(center)
-        if pos[0] != center[0] and pos[1] != center[1]:
-            raise Exception("Scale/rotate handles must have either the same x or y coordinate as their center point.")
         return self.addHandle({'name': name, 'type': 'sr', 'center': center, 'pos': pos, 'item': item}, index=index)
     
     def addRotateFreeHandle(self, pos, center, axes=None, item=None, name=None, index=None):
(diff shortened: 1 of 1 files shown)


outofculture:
I did an experimental merge of the conflicts, and this breaks in a couple ways. The scaling behavior in master does not impact sides parallel to the axis, so we should probably retain that. One of my choices of center, `.addScaleRotateHandle([1, 0.5], [1, 0.5])`, wouldn't work at all (obviously), but we should maybe throw a warning. I couldn't figure out what to do with one of the merge conflicts ( `if self.aspectLocked:...` ). It would be nice to add an example of this to `examples/ROIExample.py`.

vallsv:
You probably can close the PR. That's so 2010s :-D

j9ac9k:
> You probably can close the PR. That's so 2010s :-D

yeah.... sorry about that.  In the last year and a half or so we've gotten more aggressive about evaluating/merging/closing existing PRs in an attempt to make the library a bit more robust.  `ROI` s are a part of the library I don't interact with much so I was largely postponing those for aspects of the library that I am more familiar with.

Thanks again for the PR!

vallsv:
There was no problem to me. I just wanting to say that propably it's not a good idea to spend much time with it, and that it is fine for me if you prefer to close it.

outofculture:
Heh, well, it seemed too neat to just throw away, so I fixed it on [outofculture:arbitrary-scale-center](https://github.com/outofculture/pyqtgraph/tree/arbitrary-scale-center). We can just switch this PR.

j9ac9k:
#1572 super-seeded this PR, but includes the changes; so closing this out.  Thanks again @vallsv

PR #68, opened 2014-05-14
PyQt5 support via qt_backport package

rwarren:
This is a cleaned up version of the last pull request I closed.

Most of the `PyQt5` support comes directly from `qt_backport`'s emulation of the `PyQt4` (or `PySide`) API.

I modified your `Qt.py` mostly to provide a bit more insight to the rest of the code on what the underlying Qt implementation is so that intelligent switching could be done where required.  Pre-existing code used `USE_PYSIDE`, and the existing `pyqt5` branch had `QT_LIB`, but I expanded on that.  Qt.py is by far the biggest change I made.

All other file changes were pretty much minor, coming about in support of the above.  Specifically while working to make every single example work as well as I could (this process also helped flesh out corners of the `qt_backport` emulation).

I see that a `.gitignore` change slipped in there, too (sorry).  This was because I was using `pip install -e` on the package while working and constantly removing the egg info was tiresome (you'll see I missed a round of it in the commits).

---

**Additional notes:**

What were the "variety of bugs" fixed by `setGraphicsSystem('raster')`?  Qt5 does not have this and I am curious what bugs will surface.

One oddity I haven't been able to fix is your `FillBetweenItem.py` example.  The fills end up being very blocky, as in [this screenshot][1].  I'm not sure what is up here, especially since the fill between in the bottom left of the basic `Plotting.py` is perfect.

[1]: http://i.snag.gy/DRY86.jpg


campagnola:
I have just merged PyQt5 support into the develop branch (based on changes by @mfitzp and @compassym). Skimming through this PR, it appears there are some parts that would still be useful. I am going to close this PR now, but please feel free to reopen it (or a new one) if you'd like to rework it to match the current develop branch. Thanks!

rwarren:
This PR was only 20 days behind the current pyqtgraph dev branch.

I'll bring it up to current again, but was there something in particular that was problematic about the PR, aside from this?

rwarren:
Yikes... major merge conflicts due to the merge of the @mfitzp fork.  Not sure they can be unwound without huge effort. :(

OTOH, merging in the other commits worked perfectly.  Not sure where to go to next... this is going to be very painful to unravel.

campagnola:
Right; the problem was just that the two branches implemented overlapping PyQt5 support in different ways, so only one branch could be merged easily.

rwarren:
@campagnola: That makes sense, of course.  However, did you check the pyqt5 compatibility between the two?  The original @mfitzp fork was extremely basic and there were still a tonne of problems.  I'm not sure if you looked into the extent of fixes that were done in my PR, and especially in `qt_backport`, but a *lot* of work was done.  In fact, the primary driver for qt_backport was almost exclusively getting the full suite of pyqtgraph examples to run without a hitch.  The few issues that mfitzp addressed were included, but a LOT more as well.  It was removed to qt_backport rather than being specific to pyqtgraph, because it also makes a lot of other PyQt4 & PySide code work out of the box just by doing `pip install qt_backport` together with PyQt5.

For example, I just did quick check of your latest `pyqtgraph\examples` directory with the latest develop branch, and there are *tonnes* of errors still (and a lot of deja vu while seeing them!).  A reasonably quick way to check all of them for base functioanlity is this, run from the examples dir:

    find . -name "*.py" -exec echo "Running {}" \; -exec python {} \;

Doing that with your current develop branch, many of them immediately crash out exactly as they used to.  And that is without even checking internal functionality of the demos!!  There were many internal fixes required as well.  If you haven't tried this "run all the examples" checking, it would probably worth doing to let you see the extent of pyqt5 incompatibility that remains in `develop`.

For comparison, if you run that same blanket "run all examples in here" check on my PR with PyQt5, they all run perfectly, with these exceptions:

  1. the `FillBetweenItem.py` quirk mentioned above (which I never fixed).
  2. Remote proxy issues that are probably all from one thing (that I could not follow and debug):
    * `RemoteSpeedTest.py` crashes
    * `parallelize.py` has issues in mid-run, for reasons I'm sure are the same as in 
    * `multiprocess.py` throws exceptions

And that's pretty much it, out of > 100 test examples!!  As mentioned, the pyqtgraph examples dir was tested *extensively* for this PR!

Like I said before, I'm not sure where to go from here now.  Trying to "fix" my PR to rewind way back to an mfitzp-style approach looks like it will be brutal/intractable.

I could probably selectively merge in the handful of Feb 28 changes you made in advance of the full merge of the problematic 1fe5976 merge commit without too much hassle, but fear that I'll then be stuck maintaining a difficult fork.  To date the upstream merges haven't been bad at all, but that is no longer true.  Also, the mods that @compass made look pretty basic, and mostly seem to be a targeted subset of what I did already (like the addition of some missing pyqt5 templates), so matching up those wouldn't be bad.

Any chance you could change your mind on which pyqt5 path to choose? :)  One is definitely more complete/compatible than the other, and will also have ongoing support/help for it!

rwarren:
Just to highlight the larger-than-you-might-expect number of changes required for achieving PyQt5 compatibility, have a look at the [QWheelEvent patches][1] that were needed.  That was a particularly fun one due to Qt mouse handling fun and some stuff I can't even remember the origin of in pyqtgraph's `RemoteGraphicsView`.

The [QPainter patches][2] are also a good example of a case where there were a bunch of basic redirects, but a few trickier ones.

Again - most (all?) of the patches in `qt_backport` were required to get pyqtgraph code (mostly all examples) to work properly.  Blanketing pyqtgraph code with if statements to handle all of these differences was getting pretty nutty, so `qt_backport` was born to make pyqtgraph more clean, the work less painful, and also be useful to a wider variety of porting cases.

For the latter point, I recall being pretty satisfied when (after getting pyqtgraph working well) a Tetris clone that is out there and written in both [PyQt4][3] and [PySide][4]) works on PyQt5 without any code changes at all, just by doing `pip install qt_backport` before running the code. :)

[1]: https://github.com/rwarren/qt_backport/blob/master/qt_backport/qt5_backport.py#L467
[2]: https://github.com/rwarren/qt_backport/blob/master/qt_backport/qt5_backport.py#L359
[3]: http://zetcode.com/gui/pyqt4/thetetrisgame/
[4]: http://zetcode.com/gui/pysidetutorial/thetetrisgame/

campagnola:
Perhaps I misjudged the amount of effort that is really required here. I tested most of the examples on my system before pushing these changes (I used `python examples/ --test --pyqt5`; this is generally done for python 2 and 3 before every push although it only gives me a glance at each example). Everything looked ok, so I assumed the simpler solution made more sense. I missed the WheelEvent issue; can you point me to any others? 

I'll need some more time to review this. One issue I am grappling with is that I would rather avoid monkey-patching Qt as much as possible--it should be up to the user to decide whether to make use of qt-backports; the compatibility layer should be entirely internal to pyqtgraph.

rwarren:
> Everything looked ok, so I assumed the simpler solution made more sense

That command you gave is handy, but why is the execution list so short compared to running everything as with my `find .` version?  It seems to miss a lot of stuff.

I know the chosen PR obviously had less diffs, but I certainly don't think it was any simpler, especially in the long run.  I think that any perceived simplicity is mostly because it was an incomplete and broken attempt at pyqt5 compatibility.

The reliance on an external module (`qt_backport`) is a definite issue that I feared would affect the PR, and I really tried to avoid.  I have a long response with the rationality behind it.  However, I've stashed that for now due to length/coherence concerns... I'm a bit tired, and the response was definitely tinged by a lot of angst over potentially losing significant time and effort to-date on getting pyqtgraph working sweetly on pyqt5, as well as seeing a lot of future pain/time involved in getting it working again with your latest develop branch.

The TL;DR for the stashed response is (quite unfairly):

> The pyqt5 compatibility PR chosen was an incomplete monkey-patch, and its selection makes synchronizing the much more complete pyqt5 compatibility PR pretty difficult (it was really easy prior to the 1fe5976 commit). The longer it stays in mainline develop, the harder it will be to unwind.  A more in-between approach would be much easier, if this PR can't be taken as-is. 

... which misses all the carefully reasoned rationale and examples, but so it goes.

I also wanted to point out that, although the changed file list was pretty long for my PR, the changes boil down to:

  1. Required pyqt5 templates
  1. Minor Qt.py re-org to clean up globals that indicate what API and LIB are in use
    * Definitely the most significant looking PR diff, although I did expend a bunch of effort trying to make the diff look less scary (although looking again now, that effort seems to have failed... it really is not bad, though).
  1. associated constant changes in source files
    * eg: `if Qt.QT_LIB == Qt.LIB_PYSIDE:` instead of `if USE_PYSIDE:`
  1. Low level special stuff like required sip-layer changes
  1. Some small fixes to make your demos work under XFCE
    * addressing a [Qt bug][xfce_bug]... probably should have separated these
  1. Requiring `qt_backport` to work with pyqt5
    * kaboom

> I missed the WheelEvent issue; can you point me to any others?

I can't remember any offhand.  I've been working cleanly with pyqt5 for so long (last significant pyqt5 compatibility change back in May 2014) that it is hard to recall the functional issues that drove the patches. Sorry!

As with the WheelEvent case, the easiest thing to do is peruse the clean list of `_patch_foo()` patch functions called out by `patch_api()` in the [main patching file][main_patching], since they were dominantly driven by making pyqtgraph work on pyqt5.  That is where the meat of the work is done.  The rest of `qt_backport` is just usability glue (and some tests).

That said, I do recall having some annoying mouse interactivity issues with GradientWidget that needed fixing, but I can't remember what they were exactly.  A quick check of your latest develop trunk didn't have the problem jump at me, but maybe it will for you?

> I'll need some more time to review this. One issue I am grappling with is that I would rather avoid monkey-patching Qt as much as possible--it should be up to the user to decide whether to make use of qt-backports; the compatibility layer should be entirely internal to pyqtgraph.

That would be nice for sure, but there really is a lot required to do this.  Reminiscent of Python 2->3 fun, but probably worse due to current (and future) diversity.

Also - You *are* [already monkey patching][partial_patching] with the PR you chose... just nowhere near how much is required!

Trying to maintain Qt API differences internally to the pyqtgraph code while the differences settle out, and while not interfering with code outside of pyqtgraph internals (either directly, or indirectly by how people expect your Qt subclasses to behave) is tricky.  It seems cleaner to me for pyqtgraph to be as adherent as possible to a single API and let the users manage the differences... at least for now.  Offloading this job to qt_backport (for now?) works *extremely* well (for me, anyway :( ).  Kind of like using [six][sixpy], but with less code changes required since you just use the API you have already been using.

I have more details/examples/points if you want more food for thought.

Bah - ended up long again. :(  If you can't tell, pyqt5 compatibility of pyqtgraph is a big deal for us... don' t want to lose it!! :)

[xfce_bug]: https://bugreports.qt-project.org/browse/QTBUG-39019
[partial_patching]: https://github.com/pyqtgraph/pyqtgraph/blob/develop/pyqtgraph/Qt.py#L144
[sixpy]: http://pythonhosted.org/six/
[main_patching]: https://github.com/rwarren/qt_backport/blob/master/qt_backport/qt5_backport.py

campagnola:
> That command you gave is handy, but why is the execution list so short compared to running everything as with my find . version? It seems to miss a lot of stuff.

That command only runs the scripts that are listed in the example loader menu. If we run everything in *.py, there will be errors because many of those files are not meant to be executed directly.

> The reliance on an external module (qt_backport) is a definite issue that I feared would affect the PR, and I really tried to avoid.

If I do end up going this way, it would be by embedding qt_backport as a subpackage so there would be no extra dependency.

>  if `Qt.QT_LIB == Qt.LIB_PYSIDE:` instead of `if USE_PYSIDE:`

I had actually made this same change because I thought it would be cleaner and then backtracked because (1) it wasn't necessary or helpful and (2) it is slightly more effort for the other modules to import one more symbol and do the comparison.

> Some small fixes to make your demos work under XFCE

I saw those; that's actually what I had in mind when I suggested that some of the changes could still be merged in. I could probably cherrypick those pretty easily, but it's likely to appear again when I add new examples and forget where to put `show()`. Perhaps there is a better solution for this?

> Also - You are already monkey patching with the PR you chose... just nowhere near how much is required!

Right, this is because I was lazy  :)
For example, we now have a monkey-patch for `QGridLayout.setMargin()`, which has been deprecated for a long time. It would be better for me to go through the library and convert these uses to `setContentsMargins()`, but then I need to worry about whether I am breaking backward compatibility, which takes a bit more research and testing. Over time I would like to remove these patches.

> It seems cleaner to me for pyqtgraph to be as adherent as possible to a single API and let the users manage the differences... at least for now. Offloading this job to qt_backport (for now?) works extremely well

It would certainly be _cleaner_ to use qt_backport and it is very tempting given the effort you have already put into making this work well. My concerns about the drawbacks of patching the global Qt API still stand, though. Years ago I took this approach to deal with python2-3 differences as well as numpy version differences and other pyqtgraph users suffered from these. If I remember correctly, `six.py` uses a better approach--each compatibility fix is implemented in a new function rather than patching the built-in namespace.

Here's an idea that might work: what if qt_backport has an option to prefix the methods that it patches so that the original namespace is left intact, and it will be obvious to the user if they are accessing a method that is not part of the base Qt API.

rwarren:
> That command only runs the scripts that are listed in the example loader menu. If we run everything in *.py, there will be errors because many of those files are not meant to be executed directly.

That is definitely the case, but the files not intended for execution are pretty easy to identify.  Especially when you run that find command 50-100 times :).  Done a boatload of testing that way,

> I had actually made this same change because I thought it would be cleaner and then backtracked because (1) it wasn't necessary or helpful and (2) it is slightly more effort for the other modules to import one more symbol and do the comparison.

Right... that distinction/usefulness is more directly tied to the `qt_backport` methodology of distinguishing the backing lib from the expected api (PySide, PyQt4, PyQt5).  Having application code (like pyqtgraph) have to deal with a wide variety of APIs is a heavy load, especially as it gets more diverse.  Right now it is (PySide, PyQt4, PyQt5) for available python libs/apis.  What happens when hell freezes over and PySide bumps up for Qt5 coverage, with inevitable subtle differences?  Then Qt6 for both? Ugh.  Python 2->3 is easy in comparison.  You/pyqtgraph may be trying to stay current with pyqt5 now, but not all apps/libs will be so current.

qt_backport was done to let old code expecting a certain API to just work as-is and not be distracted by api pain.  Leave that problem to qt_backport.

> If I remember correctly, six.py uses a better approach--each compatibility fix is implemented in a new function rather than patching the built-in namespace.

"better" is subjective, but you're right that there is no namespace patching with six. :)  The six approach is to provide a common calling method that will work on both py2 and py3 back ends.  This works well for packages you are writing fresh and aiming at both py2 and py3.  But then six is a hard dependency that you must have.  I did not want qt_backport to be a hard dependency today... only if you need PyQt5 support.

> Here's an idea that might work: what if qt_backport has an option to prefix the methods that it patches so that the original namespace is left intact, and it will be obvious to the user if they are accessing a method that is not part of the base Qt API.

I'm not sure what you are thinking here.  Do you mean something like six, with a slightly different name?  eg: to use the obsolete-in-Qt5 `QGraphicsItem.translate()` method you would, instead, call something like `QGraphicsItem.qb_translate()`?  The awareness of this would need to be far and wide, or code just breaks.

In the case of pyqtgraph there would need to be a hard dependency on qt_backport for *all* installations, even when using the originally targeted PyQt4 and PySide.  That doesn't seem palatable.

The six approach here would be to force the calling of something like:

    qt_backport.QGraphicsItem_translate(obj, dx, dy)

Which is also heavy for adoption and just trying to get old code to work, and requires a hard dependency on qt_backport.

I haven't audited the resulting patch set to classify them all into renames vs functionaility fixes, but most of them don't have namespace collisions, just namespace duplicates/pollutions.  eg: `QHeaderView::setClickable` to `QHeaderView::setSectionsClickable`.

There's already a major namespace hijacking requirement happening with the object re-organization, but I acknowledge that method hijacking is a different matter.

I definitely get why you don't like a monkey patch approach since there is magic under the covers (I hate hijacking the namespace)... being explicit is definitely a better.nobler goal.  I don't see how to really get around the magic, though... at least not without a hard dependency on a compatibility module.  You could definitely have that module internal to pyqtgraph.  eg: in your `Qt.py` compatibility module you have something like:

    Qt.QGraphicsItem_translate(obj, dx, dy)

but then you end up having to maintain that subset yourself and can drift away from other patch attempts, like qt_backport (also splitting development effort).  It also isn't clear to me what the impact is on the users of your subclasses.

One thing that makes me uncomfortable about qt_backport is the implicit hijacking of the PySide and PyQt4 imports.  I initially started by trying to make it an explicit hijacking to minimize magic.  ie: doing something like:

    import qt_backport
    qt_backport.hijack_PySide()

or

    from qt_backport import PySide

but was running into all kinds of problems trying to make this work.  I don't actually recall the details. At one point I was stuffing sys.modules to make it work across files once done, but the details are lost in history.

In the end I settled on the implicit takeover because the fact is that the installation of Qt and PySide/PyQt is such a major operation (or is buried in someone's Anaconda installation or something) that implicitly patching the differences with a single install wasn't so bad, after weighing everything.

If someone does have multiple major versions od Qt lib floating about with multiple wrapper APIs (eg: PySide) installed I think there might currently be an issue.  Sorting out the layers and choosing your API and lib is definitely possible with qt_backport, but I didn't get there yet since it wasn't needed.

Argh.  So messy.  Not sure any of the above resolved anything.

rwarren:
It might also be worth pointing out the extent to which I tried to isolate the backporting patches from code that should not be patched.

The typical qt_backport situation is when you have your own code that is using the latest lib+wrapper (eg: Qt5.4 + PyQt5), but you also have the need to import some old code that was written with expectations of a completely different lib+wrapper (eg: Qt4+PySide).

I built in [isolation code](https://github.com/rwarren/qt_backport/blob/master/qt_backport/qt5_backport.py#L175) to keep these completely separate.  When this isolation code is in use, they are niceley independent, per this console example:

https://bpaste.net/show/7288bdf72aae

Unfortunately, although this works for lots of code, it did not work for pyqtgraph for reasons I never resolved.  At the time I think I suspected some of the mouse handling code which I seem to recall you completely take over from Qt, but I never resolved it.  As a result, the current qt_backport release doesn't use the isolation code, and there is some resulting namespace pollution.  This has not affected me to date, since the fact is that there isn't much crossover in reality.

It comes down to the fact that if you use the API you expect to use in your app/lib, it just works fine when qt_backport is involved.  It really isn't doing that much to do it... the [patch file](https://github.com/rwarren/qt_backport/blob/master/qt_backport/qt5_backport.py) is currently < 600 lines with a tonne of comments.

campagnola:
> In the case of pyqtgraph there would need to be a hard dependency on qt_backport for all installations, even when using the originally targeted PyQt4 and PySide. That doesn't seem palatable.

That would be just fine--I can embed qt_backport as a subpackage and use subtree merges to synchronize code with the upstream repository. It's only slightly more effort and well worth avoiding dependency issues.

The fact that most of the patches do not override existing methods in the Qt5 namespace is encouraging..  I need time to look closely at the patch file.

PR #73, opened 2014-06-06
Cleanup ROI.py (mostly whitespace & line lengths).

anntzer:
Some cleanups to ROI.py: keep line lengths below 80 characters (most of the time), remove trailing whitespace, and replace `True if x else False` by `bool(x)`.


anntzer:
Now obsolete.

Issue #16, opened 2013-11-25
Image export gives incorrect results when specifying size.


campagnola:
When exporting with ImageExporter, specifying any width other than the default produces erroneous results. In some cases, axis labels are out of place, and in other cases lines do not match up to scatter plot points.

See this discussion: https://groups.google.com/d/msg/pyqtgraph/qkLUEiFt0Qk/pwiNToIKcMsJ

Example:

```
from PyQt4 import QtCore, QtGui
import pyqtgraph as pg
pg.mkQApp()

view = pg.PlotWidget(title='I0-Normalised reflectivity')
view.setLabel('left', "Reflectivity", units='arb. units')
view.setLabel('bottom', "Photon energy", units='eV')
view.showAxis('top', show=True)

exporter = pg.exporters.ImageExporter.ImageExporter(view.plotItem)
exporter.params['width'] = 1000
exporter.export('test.png')
```

mfitzp:
Also experiencing this bug with scatterplots + plots with markers on lines. In both cases the markers end up taking up 50% of the area they should, being placed in the top left of the plot.

aeroaks:
I tried changing the Widget for export and making it visible for processing the resizeevents and then again making it invisible. I get proper exported images using the following code, I do not change the Width parameter in the exporter class:

```
# plot export area
self.export_area = pg.GraphicsLayoutWidget(border=(50, 50, 50))
# Configure the colors:
self.export_area.setBackground('w')
# We resize the widget to our desired export dimensions.
# When the widget is hidden, the child widgets do not get resized.
# Only when the widget is made visible, the events are propagated to the child widgets.
# We immediately hide the widget again to prevent a pop-up or flashing screen.
self.export_area.resize(2000, 1000)
self.export_area.setVisible(True)
self.export_area.setVisible(False)
```

Then use self.export_area for drawing and exporting as an invisible widget

2xB:
Here is an updated version of the code given to reproduce this error with the latest pyqtgraph version from `develop`:

```python3
import pyqtgraph as pg
import pyqtgraph.exporters as ex
pg.mkQApp()

view = pg.PlotWidget(title='I0-Normalised reflectivity')
view.setLabel('left', "Reflectivity", units='arb. units')
view.setLabel('bottom', "Photon energy", units='eV')
view.showAxis('top', show=True)

exporter = ex.ImageExporter(view.plotItem)
exporter.params['width'] = 1000
exporter.export('test.png')
```

[...]

PR #88, opened 2014-08-05
Fixed image scatterplot export bug Issue #16

boylea:
Exporting an image with a ScatterPlotItem in it does not scale correctly. This seems to be due to the QImage that is being exported to not having the expected scaling in its transform attribute.

Fix for https://github.com/pyqtgraph/pyqtgraph/issues/16

### DIFF
--- a/pyqtgraph/graphicsItems/GraphicsItem.py
+++ b/pyqtgraph/graphicsItems/GraphicsItem.py
@@ -102,8 +102,9 @@ class GraphicsItem(object):
         Extends deviceTransform to automatically determine the viewportTransform.
         """
         if self._exportOpts is not False and 'painter' in self._exportOpts: ## currently exporting; device transform may be different.
-            return self._exportOpts['painter'].deviceTransform() * self.sceneTransform()
-            
+            scaler = self._exportOpts['resolutionScale']
+            return self.sceneTransform() * QtGui.QTransform(scaler, 0, 0, scaler, 1, 1)
+
         if viewportTransform is None:
             view = self.getViewWidget()
             if view is None:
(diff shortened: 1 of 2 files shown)


j9ac9k:
#412 should have patched this up; thanks for the PR

2xB:
#412 was not merged and issue #16 persists.

j9ac9k:
I checked out this branch, this is what it looks like in the app:

image

This is what the exported output looks like:

test-2

You can generate the results with the following code

```python
import pyqtgraph as pg
import pyqtgraph.exporters as ex

app = pg.mkQApp()

view = pg.PlotWidget(title='I0-Normalised reflectivity')
view.setLabel('left', "Reflectivity", units='arb. units')
view.setLabel('bottom', "Photon energy", units='eV')
view.showAxis('top', show=True)
view.show()

exporter = ex.ImageExporter(view.plotItem)
exporter.params['width'] = 1000
exporter.export('test-2.png')

if __name__ == '__main__':
    import sys
```

paulmueller:
Thanks, this fixes #16 / #105 / #484 for me.

j9ac9k:
@paulmueller thanks for following up, I'll take a second look at this PR vs. the issues you referenced, it's quite possible I did not understand the issue this PR was supposed to resolve.

paulmueller:
Here is an example plot with lines and a scatter plot (without code snippet, sorry):
before:
aa
after:
aaaaa

j9ac9k:
Thanks @boylea for this PR, I apologize it's taken so long for us to merge this.

Thanks @JohnLunzer @mmeiler and @paulmueller for reporting this issue and detailing it.

Merging shortly

PR #95, opened 2014-09-04
Python3 related fixes

anntzer:
Two fixes related to Python3:
- py.test for python3 is not necessarily called py.test3, and
- do not inject python2 builtins in the python3 builtins namespace.


anntzer:
Now obsolete.

PR #101, opened 2014-09-25
configurable GridItem tick spacing and pen color

jmakovicka:
this patch adds configuration options for GridItem grid and text color, and options for custom grid line spacing.

It also makes GridItem honor the 'foreground' config option.

this version uses setTextColor as advised. It also fixes setting of just one of the x/y tick spacings, and puts setting of the text color in the proper place.

### DIFF
--- a/pyqtgraph/graphicsItems/GridItem.py
+++ b/pyqtgraph/graphicsItems/GridItem.py
@@ -3,6 +3,7 @@ from .UIGraphicsItem import *
 import numpy as np
 from ..Point import Point
 from .. import functions as fn
+from .. import getConfigOption
 
 __all__ = ['GridItem']
 class GridItem(UIGraphicsItem):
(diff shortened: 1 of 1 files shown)


jmakovicka:
Rebased to current develop.

pyqtgraph/graphicsItems/GridItem.py
```diff
@@ -1,33 +1,99 @@
...
+    def __init__(self, pen='default', textColor='default'):
         UIGraphicsItem.__init__(self)
         #QtGui.QGraphicsItem.__init__(self, *args)
         #self.setFlag(QtGui.QGraphicsItem.ItemClipsToShape)
         #self.setCacheMode(QtGui.QGraphicsItem.DeviceCoordinateCache)
-        
+
+        self.opts = {}
+
+        if pen == 'default':
```
2xB: How about moving this logic to `setPen`, so both `__init__` and `setPen` accept the same input? Same goes for `setTextColor`.

pyqtgraph/graphicsItems/GridItem.py
```diff
@@ -1,33 +1,99 @@
...
+            self.setPen()
+        else:
+            self.setPen(pen)
+
+        if textColor == 'default':
+            self.setTextColor()
+        else:
+            self.setTextColor(textColor)
+
+        self.setTickSpacing(x=[None, None, None], y=[None, None, None])
```
2xB: This should probably also be specifiable in that method.

pyqtgraph/graphicsItems/GridItem.py
```diff
@@ -1,33 +1,99 @@
...
+            setTickSpacing(x=[1.0], y=[None, None])
+
+            Single scale with distance of 1.0 for X axis, Two scales
+            for Y axis, one with spacing of 1.0, other one automatic:
+            setTickSpacing(x=[1.0], y=[1.0, None])
+        """
+        self.opts['tickSpacing'] = (x or self.opts['tickSpacing'][0],
+                                    y or self.opts['tickSpacing'][1])
+
+        self.no_grids = max([len(s) for s in self.opts['tickSpacing']])
```
2xB: `no_grids` probably stands for 'number of grids'? That might be confusing. Perhaps `griddepth`?

pyqtgraph/graphicsItems/GridItem.py
```diff
@@ -1,33 +1,99 @@
...
+        else:
+            self.opts['pen'] = fn.mkPen(getConfigOption('foreground'))
+
+        self.picture = None
+        self.update()
+
+    def setTextColor(self, *args, **kwargs):
+        """Set the color used to draw the texts."""
+        if args or kwargs:
+            if len(args) == 1 and args[0] is None:
```
2xB: A codegolfy suggestion for readability: `if args == [None]:`

pyqtgraph/graphicsItems/GridItem.py
2xB: It is probably not that important to be able to specify the tick spacing when calling the constructor.

2xB:
@jmakovicka First, thank you very much for your fast response.
I lost some overview with the force-pushes - why is the commit **tweak alpha value to keep the default grid appearance** no longer included?

jmakovicka:
> @jmakovicka First, thank you very much for your fast response.
> I lost some overview with the force-pushes - why is the commit _tweak alpha value to keep the default grid appearance_ no longer included?

Sorry, I missed that change when updating the patch.

pyqtgraph/graphicsItems/GridItem.py
2xB: This is probably not that important, resolving.

j9ac9k:
Code looks good to me, only question I have is should we modify one of the examples to reflect the usage of this feature?

PR #114, opened 2014-10-23
Update GLViewWidget.py

rkedge:
Change mouseMoveEvent to use the right mouse button for panning. 
Event is currently set to the middle mouse button. This causes erratic bevhavior when trying to click and drag since the wheel usually gets activated to and the graphic starts zooming at the same time. The right button is not currently used for anything.
I have tested this and it works in all of the demos.


j9ac9k:
Closing due to time-single the original PR was made.

PR #117, opened 2014-11-02
Feature: Flow chart nodes resize based on inputs/outputs

onlyjus:
Changed the node sizing code so that nodes change size based on the number of inputs and outputs.


j9ac9k:
Tested this out, this works wonderfully, thanks for the PR @onlyjus sorry it's taken so long for us to follow up.  I'll merge pending a clean CI run.

PR #121, opened 2014-11-18
Fix axis range growing infinitely if we have no data or zero-range.

josselex:
Fix axis range growing infinitely if we have no data or zero-range.
Before, the padding was applied over and over again.


josselex:
nothing happend, so i closed this pull request

PR #123, opened 2014-11-25
Added arrow symbols for the ScatterPlotItem

Peque:
This commit adds arrows (up, right, down and left) to the ScatterPlotItem's symbol list.

When executing the ScatterPlot example, we can see they are significantly smaller than the rest of the symbols because, as I have seen, the symbols' path is to fit in the 1x1 square centered at (0, 0). Of course the arrows have been implemented to point to the actual point in the scatter plot, which means they only use a small part of that area.

Is there a better way to implement them? I will be glad to change anything if that leads to the acceptance of these symbols in PyQtGraph.


j9ac9k:
Hello @Peque I tried to update this PR resolving merge conflicts, and adding the symbols to (what didn't exist at the time you created this PR) `examples/Symbols.py`, unfortunately I didn't have permission to edit the PR.  I'm going to merge your suggested changes as part of #1244 

Thanks for the PR, sorry it's taken us so long to incorporate it.

Peque:
@j9ac9k Better late than never! :joy: 

Thanks a lot. :blush:

PR #126, opened 2014-12-02
UTF-8 support for CSVExporter

malazar:
Hello, I had a problem with exporting UTF-8 encoded headers with the CSVExporter, so I came up with this fix.


j9ac9k:
#1219 implemented the desired effect here while addressing #1081  Thanks for the PR @malazar sorry nobody followed up with you in a reasonable period of time.

PR #130, opened 2014-12-24
DockArea apoptose fix.

lidstrom83:
Fixes problem where docks can't be added to a non-temporary area once all of its docks have been closed.


campagnola:
Thanks very much!

PR #134, opened 2014-12-30
Fix bug in LayoutWidget.py

cjtk:
getWidget tries to get self.row which doesn't exist, get self.rows instead


j9ac9k:
Thanks for the PR, sorry it's taken us so long to get to this.

PR #136, opened 2015-01-12
Don't create extraneous AxisItems

welshjf:
Fixes a bug introduced in 0.9.10.

PR #139, opened 2015-01-15
bugfix: don't attach spurious AxisItem to plot

picketg:
Bug description: when PlotItem is given a custom axis in constructor:
PlotItem(axisItems={'bottom': myAxis})
the result is a plot with _2_ bottom axes visible instead of just user's custom axis.

Cause: axisItems.get(k, AxisItem(orientation=k, parent=self)) attaches an axis to the parent via "parent=self" directive, even if custom axis is present in 'axisItems'.

Fix: create and attach default AxisItem to parent _only_ after verifying that no custom AxisItem is available.

Fixes regression in f6ded808efc89cb65d51edd2257c5a204b856317
"Fixed a few exit crashes, added unit tests to cover them"
(where parent=self was added)

Signed-off-by: Picket G


welshjf:
Duplicate of #136

ixjlyons:
As mentioned, this looks like a duplicate which has been resolved. Thanks though!

PR #140, opened 2015-01-22
Pyqt5

compassym:
Fixed some bugs for running the examples


campagnola:
Many thanks! Shall I credit your real name or user name?

PR #142, opened 2015-01-26
Bugfix: PlotCurveItem.sigClicked emits MouseClickEvent

miili:
Hi all,

currently `PlotCurveItem.sigClicked` emits `self`, a `PlotCurveItem`:

```
def mouseClickEvent(self, ev):
    if not self.clickable or ev.button() != QtCore.Qt.LeftButton:
        return
    if self.mouseShape().contains(ev.pos()):
        ev.accept()
        self.sigClicked.emit(self)
```

Since it can be useful to get the MouseClickEvent in the GraphicsView I suggest to either emit
            `self.sigClicked.emit(ev)`
or
            `self.sigClicked.emit(self, ev)`

cheers!

### DIFF
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -567,7 +567,7 @@ class PlotCurveItem(GraphicsObject):
             return
         if self.mouseShape().contains(ev.pos()):
             ev.accept()
-            self.sigClicked.emit(self)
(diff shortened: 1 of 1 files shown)


lidstrom83:
In order to get access to the keyboard modifiers, I currently do the latter: `self.sigClicked.emit(self, ev)`.

I think this PR should similarly change ScatterPlotItem and PlotDataItem, as well as update the docstrings.

PR #143, opened 2015-01-30
Bugfix: a user provided handle for ROIs was not set to the correct posit...

CPrescher:
I just changed the position of the h.setPos so that also a user provided handle will get initially the correct position on the ROI

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -1200,14 +1200,16 @@ class Handle(UIGraphicsItem):
     sigClicked = QtCore.Signal(object, object)   # self, event
     sigRemoveRequested = QtCore.Signal(object)   # self
     
-    def __init__(self, radius, typ=None, pen=(200, 200, 220), parent=None, deletable=False):
+    def __init__(self, radius, typ=None, pen=fn.mkPen(200, 200, 220), parent=None, deletable=False,
+                 activePen = fn.mkPen(255, 255,0)):
         #print "   create item with parent", parent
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/ROI.py
```diff
@@ -1200,14 +1200,16 @@ class Handle(UIGraphicsItem):
     sigClicked = QtCore.Signal(object, object)   # self, event
     sigRemoveRequested = QtCore.Signal(object)   # self
     
-    def __init__(self, radius, typ=None, pen=(200, 200, 220), parent=None, deletable=False):
+    def __init__(self, radius, typ=None, pen=fn.mkPen(200, 200, 220), parent=None, deletable=False,
+                 activePen = fn.mkPen(255, 255,0)):
```
campagnola: Should generally avoid using mutable objects in a function signature (see for example http://docs.python-guide.org/en/latest/writing/gotchas/).

Luindil: Hei, I understand, but how would be the best way then to let the user choose different linewidths for the active and inactive state? Add a linewidth parameter for each?

I also have to apologize, those changes where not supposed to be in the pull request i did not know that all future commits also go into the pull request. The main change i wanted to propose where the position one in line 508.

campagnola: It would be better to leave the default arguments as tuples and put the calls to `mkPen()` inside the `__init__()` method.

Github pull requests always track the changes between two _branches_, so as soon as you update your develop branch, those changes are included in the PR (this allows you to continue modifying the PR until we decide it is ready to be merged). To avoid mixing unrelated changes, you should create a new topic branch for each PR.

PR #153, opened 2015-03-16
Dock to emit signal when closed

lidstrom83:


pyqtgraph/dockarea/DockArea.py
```diff
@@ -317,5 +321,12 @@ def dragLeaveEvent(self, *args):
     def dropEvent(self, *args):
         DockDrop.dropEvent(self, *args)
 
-        
-        
+
+class TempAreaWindow(QtGui.QMainWindow):
+    def __init__(self, area, **kwargs):
+        super().__init__(**kwargs)
```
campagnola: `super()` without arguments is not allowed in python 2.

lidstrom83:
Whoops, shouldn't have been using super() in the first place with PyQt4/PySide...  Fixed now.

campagnola:
Another down. I am digging all these compact PRs  :)

lidstrom83:
Oh good, I was hoping to make it easy on you since I knew I'd have a bunch.

campagnola:
It worked. If those three PRs had all been joined into one, I probably would have taken a lot longer to review them (and they probably would have been more difficult to review anyway).

PR #154, opened 2015-03-16
Fix DockArea save/restore when topContainer is None

lidstrom83:
I.e. when no docks are present, or when all are in temporary windows.


campagnola:
Thanks again!

PR #156, opened 2015-03-17
Make ViewBox update matrix for invertX and setState

lidstrom83:
I couldn't figure out why these updateMatrix calls were commented out, but they seem to be necessary (or equivalently as in invertY, _matrixNeedsUpdate can be set to True).


j9ac9k:
Thanks for the PR, I apologize it's taken so long for someone to follow up.  I'm going to close this guy out, as `self._updateMatrix` is currently called under both of those functions, and ViewBox.py looks a fair amount different from when the PR was submitted.

PR #165, opened 2015-04-01
Do not create an axis if a custom one is specified

jmakovicka:
Because get(key, default) creates the default object unconditionally,
and the default axis gets linked with the PlotItem, a bogus axis
ends up displayed in the top left corner when a custom axis is used.

Change-Id: I7bfd5e126cc7bd150122edd8436348f0c74ff947


jmakovicka:
should have read the previous pull requests

PR #168, opened 2015-05-14
New plotting items

lesauxvi:
Existing items that can be added to a PlotItem (such as InfiniteLine or LinearRegionItem) missed some functionnalities I needed. I've extended their functionalities with some extra features. I also added a CrossHair item and an example to illustrate these new functionnalities.

### DIFF
--- a/examples/Plotting_items.py
+++ b/examples/Plotting_items.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+"""
+This example demonstrates some of the new plotting items.
+"""
+
+import initExample ## Add path to library (just for examples; you do not need this)
+
+
+from pyqtgraph.Qt import QtGui, QtCore
+import numpy as np
+import pyqtgraph as pg
+
+#QtGui.QApplication.setGraphicsSystem('raster')
+app = QtGui.QApplication([])
+#mw = QtGui.QMainWindow()
+#mw.resize(800,800)
+
+win = pg.GraphicsWindow(title="Basic plotting with some added items")
+win.resize(1000,600)
+win.setWindowTitle('pyqtgraph example: Plotting with items')
+
+# Enable antialiasing for prettier plots
+pg.setConfigOptions(antialias=True)
+
+p1 = win.addPlot(title="Basic array plotting", y=np.random.normal(size=100))
+infiniteLine = pg.InfiniteLine(label=True)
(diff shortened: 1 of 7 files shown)


campagnola:
I really like these features. To start, I have some architectural questions:

* Why have separate InfiniteLineBase/InfiniteLine and LinearRegionBase/LinearRegion classes?
* Why reimplement text handling in LinearRegion rather than reusing it from InfiniteLine?

lesauxvi:
Initially, I didn't want to modify the pyqtgraph source code, which explains why I splitted the files into 2 differents files the original one, and the modified one). Now, when I think about it few months later, it seems odd to me. I should probably modifiy that and just extends the original files. Changes forthcoming.

For your second question, I do not remember why I did that. I'll check it right now.

lesauxvi:
Hi Luke,

I've started merging my modifications for the InfiniteLine object with the base class (the current PyQtGraph one) and I've a small problem. To be able to locate my label (which is a TextItem), I need the ViewBox range, that I can get from the UIGraphicsItem.boundingRect method. So basically, I have to change the inheritance of InfiniteLine (from GraphicsObject to UIGraphicsItem, but no big deal). However, depending on the angle value set when the object is instantiated, the boundingRect method can give surprising results. This is because there is a rotation in the setAngle method. So I need to completely modify the way the line is painted to avoid this rotation.

To sum-up, if I want to modify the InfiniteLine class in order to add the proposed modifications, I have to:
   - InfiniteLine is no longer a child of GraphicsObject, but a child of UIGraphicsItem
   - the way the line is painted can no longer take advantage of the rotation feature of QPainter, and I have to handle this differently by specifically define two points. Moreover, this will prevent the awkward situation when an angle different from 0 and 90 is set (set it to 45 for example, and look at the result ;-)).

Do you agree with these modifications, or do you see something I missed which can avoid all theses changes? I'm waiting for your feedback before starting the code modification.

Cheers

Vincent

lesauxvi:
I did not wait for your answer because I found some time yesterday to find out a way to properly merge InfinteLine class and InfiniteLineBase class. It took me some time because I had no particular experiences into GraphicsView and co'. Things are much more clear now and the InifiniteLine class is nearly finished. I think it would be easier for you to review it if I create a new PR dedicated to the InfinteLine class (LinearRegionItem and CrossHair classes will be also in dedicated PRs). I propose to close this one and create new ones. new PR incoming really soon.

PR #169, opened 2015-05-16
Properly remove select box when export dialog is closed

klickverbot:
Previously, only clicking the "Close" button would remove it,
but it would stay behind when directly closing the window.


campagnola:
Thanks very much!

PR #177, opened 2015-06-22
Fix clickRadius feature

vallsv:
Hi. The current version of PyQtGraph do not support anymore the setClickRadius feature. The code was broken somewhen.

To support again the clickRadius feature (which allow to select an item anyway the mouse is over it or not):
- `items` list must come from the `rgn` shape (computed from `_clickRadius`)
- `shape contains point` must be removed cause that's especially what must allow the clickRadius feature.

This patch does that.

Thanks a lot.

### DIFF
--- a/pyqtgraph/GraphicsScene/GraphicsScene.py
+++ b/pyqtgraph/GraphicsScene/GraphicsScene.py
@@ -426,7 +426,7 @@ class GraphicsScene(QtGui.QGraphicsScene):
         #self.searchRect.setRect(rgn)
 
 
-        items = self.items(point, selMode, sortOrder, tr)
+        items = self.items(rgn, selMode, sortOrder, tr)
         
         ## remove items whose shape does not contain point (scene.items() apparently sucks at this)
         items2 = []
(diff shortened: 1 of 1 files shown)


campagnola:
This one is tricky--long ago had to switch away from `QGraphicsScene.items(region, ...)` because it was failing to detect some items. The alternative has been to ensure that thin objects expand their `shape()` to be more permissive in catching clicks. This has the added benefit that individual items can customize the way they interact with the mouse. 

However I *would* like to have `clickRadius` working again; it will just take a lot of testing to be sure everything works as expected. It might be possible to make this change safer, though, by setting the default clickRadius to 0, and using  `QGraphicsScene.items(point, ...)` only if the user has not explicitly changed the clickRadius.

vallsv:
Hi. Thanks for the answer.

Then, will you accept a patch using ```0``` as default value for ```clickRadius``` and using the current ```QGraphicsScene.items(point, ...)``` code?

j9ac9k:
I'm going to close this PR out for house-keeping purposes, however @vallsv if you still want to discuss the implementation of `clickRadius` please open up an issue!

PR #178, opened 2015-06-23
ENH: Added a new shader

miha-pirnat:
Hi,
we are developing an engineering app and it would help a lot if a shader as proposed could be added to the pyqtgraph. The purpose of this shader is to ignore the order of the triangle nodes for shading purposes as discussed in pyqtgraph google group:
https://groups.google.com/forum/#!topic/pyqtgraph/F9tymaHmT8o

Maybe would be better to implement a way so that custom shader could be read from a file? I'm willing to try that.

### DIFF
--- a/pyqtgraph/opengl/shaders.py
+++ b/pyqtgraph/opengl/shaders.py
@@ -111,7 +111,97 @@ def initShaders():
                 }
             """)
         ]),
-        
+
+        ShaderProgram('customShader', [
+            VertexShader("""
+                varying vec4 diffuse,ambient,baseColor;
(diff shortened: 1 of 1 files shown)


miha-pirnat:
bump?

campagnola:
The `Shaders` variable you added to is global; can't you add your custom shader to this at runtime from your application? It should not be necessary to modify the original shaders file..

miha-pirnat:
I tried something like you said, but i had memory leak problems...now I tried again and it works. I guess a nudge in the form of your comment was all I needed :). Thanks.

PR #179, opened 2015-07-06
Log scale and fft transform fix

mrussell42:
If the plotted data is fourier transformed and an x log scale is chosen, the first bin causes an error because np.log10(0) doesn't make any sense.


campagnola:
Excellent, thank you!

PR #188, opened 2015-07-13
Clean up temp file from test suite

ericdill:


pyqtgraph/exporters/tests/utils.py
```diff
@@ -0,0 +1,47 @@
+import tempfile
+import uuid
+import os
+
+
+def gentempfilename(dir=None, suffix=None):
```
campagnola: why not just use the `tempfile` module for these?

ericdill: I have no good reason for not ultimately winding up with just using tempfile.

campagnola:
Awesome. Venti grazie.

Issue #144, opened 2015-02-02
HistogramLUTItem doesn't update when in LogMode


fedebarabas:
Hi all,

I stumpled upon this weird issue. I wanted to have an histogramLUT with its y-axis in log mode. Following the example:

<pre><code>from pyqtgraph.Qt import QtCore, QtGui
import numpy as np
import pyqtgraph as pg
import pyqtgraph.ptime as ptime

app = QtGui.QApplication([])

## Create window with GraphicsView widget
win = pg.GraphicsLayoutWidget()
win.show()  ## show widget alone in its own window
win.setWindowTitle('pyqtgraph example: ImageItem')
view = win.addViewBox()

## lock the aspect ratio so pixels are always square
view.setAspectLocked(True)

## Create image item
img = pg.ImageItem(border='w')
view.addItem(img)
hist = pg.HistogramLUTItem()
hist.gradient.loadPreset('yellowy')
hist.setImageItem(img)
hist.plot.setLogMode(False, True)
hist.vb.setLimits(yMin=0, yMax=20000)
win.addItem(hist)

## Set initial view bounds
view.setRange(QtCore.QRectF(0, 0, 600, 600))

## Create random image
data = np.abs(np.random.normal(size=(15, 600, 600), loc=1024, scale=64).astype(np.uint16))
i = 0

updateTime = ptime.time()
fps = 0

def updateData():
    global img, data, i, updateTime, fps

    ## Display the data
    img.setImage(data[i])
    i = (i+1) % data.shape[0]

    QtCore.QTimer.singleShot(1, updateData)
    now = ptime.time()
    fps2 = 1.0 / (now-updateTime)
    updateTime = now
    fps = fps * 0.9 + fps2 * 0.1

    #print "%0.1f fps" % fps

updateData()

## Start Qt event loop unless running in interactive mode.
if __name__ == '__main__':
    import sys
    if (sys.flags.interactive != 1) or not hasattr(QtCore, 'PYQT_VERSION'):
        QtGui.QApplication.instance().exec_()</code></pre>

If I just delete the line 
<code><pre>hist.plot.setLogMode(False, True)</code></pre>
the histogram updates just fine. I realize it must have something to do with applying np.log to zero counts, but PlotItem seems to take care of that problem.

Am I doing something wrong?
Thank you for such a wonderful library

2xB:
[...]

PR #190, opened 2015-07-14
trying to fix error in log scale Issue  #144

cweninger:
Trying to fix the error in log scaling in issue 144.
Running the example code in that issue i get:

 "RuntimeWarning: divide by zero encountered in log10"

going from np.log10(y) to np.log10(1.0+y) fixes the issue

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -530,7 +530,7 @@ class PlotDataItem(GraphicsObject):
             if self.opts['logMode'][0]:
                 x = np.log10(x)
             if self.opts['logMode'][1]:
-                y = np.log10(y)
+                y = np.log10(1.0+y)
             #if any(self.opts['logMode']):  ## re-check for NANs after log
                 #nanMask = np.isinf(x) | np.isinf(y) | np.isnan(x) | np.isnan(y)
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -530,7 +530,7 @@ def getData(self):
             if self.opts['logMode'][0]:
                 x = np.log10(x)
             if self.opts['logMode'][1]:
-                y = np.log10(y)
+                y = np.log10(1.0+y)
             #if any(self.opts['logMode']):  ## re-check for NANs after log
                 #nanMask = np.isinf(x) | np.isinf(y) | np.isnan(x) | np.isnan(y)
                 #if any(nanMask):
```
ericdill: it looks like this code block that is now commented out was meant to deal with this problem. @campagnola, what was the rationale for removing this code block?

campagnola: I think it was removed because it is computationally expensive and unnecessary--Qt should simply ignore any points that are NaN. If you just want to avoid the warning message, then the calls to `np.log10()` can be wrapped with `warnings.catch_warnings`. Issue #144 will require a more complicated solution because InfiniteLine and LinearRegionItem are not log-aware.

PR #191, opened 2015-07-17
Putting travis back to work!

ericdill:


ericdill:
https://travis-ci.org/ericdill/pyqtgraph/builds

ericdill:
@campagnola can you turn travis back on?

campagnola:
Travis is now on. Maybe send another commit to kick it?

ericdill:
Awesome, travis is running and passing.

Some notes:

1. That this is now using the containerized systems and is much faster (woo!)
1. I'm skipping 2 tests on python 2.7 that were failing and 4 extra tests on python 3.4 that were failing
1. I think we agreed that we didnt need to test EVERY combination of qt libs and python verions, right?  The test matrix is:
    - python 2.7 & pyside
    - python 2.7 & pyqt4
    - python 3.4 & pyqt4

(Let me squash the commits before this gets merged too)

ericdill:
Also, can you turn on codecov.io and/or coveralls?

ericdill:
note to self: `PySide.QtTest.QTest.qWait` does not exist and that test should be skipped in pyside

campagnola:
I think I enabled codecov.io, but it's not really clear from the website if this is the case..

ericdill:
@campagnola Travis is now passing. A number of tests are being skipped on python 3 + pyqt4 or python 3 + pyside.  All but three tests are being run on python 2 + pyqt4. These three tests are:
- https://github.com/pyqtgraph/pyqtgraph/pull/191/files#diff-37c86e4a401f35f7200decd43162ff55R17
- https://github.com/pyqtgraph/pyqtgraph/pull/191/files#diff-e8f342916888445248377e20df802e4eL61
- https://github.com/pyqtgraph/pyqtgraph/pull/191/files#diff-e8f342916888445248377e20df802e4eL70

Thoughts?

ericdill:
bump

ericdill:
@campagnola Can you have a look at this and (hopefully) merge it?  Tests are passing on travis!

campagnola:
Sorry Eric, I have been travelling and interviewing non-stop since SciPy ended. Review forthcoming!

README.md
```diff
@@ -1,3 +1,6 @@
+[![Build Status](https://travis-ci.org/pyqtgraph/pyqtgraph.svg?branch=develop)](https://travis-ci.org/pyqtgraph/pyqtgraph)
+[![codecov.io](http://codecov.io/github/Nikea/scikit-xray/coverage.svg?branch=develop)](http://codecov.io/github/Nikea/scikit-xray?branch=develop)
```
campagnola: Hopefully this will embarrass me into writing more tests.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -530,15 +530,18 @@ def getData(self):
             if self.opts['logMode'][0]:
                 x = np.log10(x)
             if self.opts['logMode'][1]:
+                if 0 in y:
```
campagnola: This does not belong. If it's necessary to fix a bug elsewhere, then we'll need a different solution..

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -530,15 +530,18 @@ def getData(self):
...
-            #if any(self.opts['logMode']):  ## re-check for NANs after log
-                #nanMask = np.isinf(x) | np.isinf(y) | np.isnan(x) | np.isnan(y)
-                #if any(nanMask):
-                    #self.dataMask = ~nanMask
-                    #x = x[self.dataMask]
-                    #y = y[self.dataMask]
-                #else:
-                    #self.dataMask = None
+                
+            if any(self.opts['logMode']):  ## re-check for NANs after log
```
campagnola: I think this will need a different solution as well. Maybe decouple these from this PR?

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
```diff
@@ -58,28 +65,27 @@ def test_ViewBox():
...
     
     # test tall resize
-    win.resize(400, 800)
-    app.processEvents()
-    w = vb.geometry().width()
-    h = vb.geometry().height()
-    view1 = QRectF(0, -5, 10, 20)
-    size1 = QRectF(0, h, w, -h)
-    assertMapping(vb, view1, size1)
+    # win.resize(400, 800)
```
campagnola: IIRC these are important tests (in that they remind me what is broken). They can be moved to a separate test function and skipped for now..

campagnola:
Travis changes look great! 
Apologies again for the hangup.

pyqtgraph/graphicsItems/PlotDataItem.py
ericdill: Oh, yeah, whoops. This was not intended to be part of this PR. Thanks for catching this.

pyqtgraph/graphicsItems/PlotDataItem.py
ericdill: Same as above... thanks for catching this

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
ericdill: Is there any significance to the order that the tests in this test function is run?  Can I move each `assertMapping` block to its own test function?

ericdill:
Also, welcome back!  I forgot that you were traveling, sorry for the noise coming from this thread :disappointed: .

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
campagnola: Ah.. yes, there are a few lines above that are required. Should start like this:

```
def test_ViewBox():
    qtest = pg.Qt.QtTest.QTest
    global app, win, vb
    QRectF = pg.QtCore.QRectF
    
    win = pg.GraphicsWindow()
    win.ci.layout.setContentsMargins(0,0,0,0)
    win.resize(200, 200)
    win.show()
    vb = win.addViewBox()
    
    # set range before viewbox is shown
    vb.setRange(xRange=[0, 10], yRange=[0, 10], padding=0)
    
    # required to make mapFromView work properly.
    qtest.qWaitForWindowShown(win)
    
    g = pg.GridItem()
    vb.addItem(g)
    win.resize(400, 400)
    vb.setAspectLocked()
    win.resize(800, 400)
    app.processEvents()

    # commented test code here
```

ericdill: Great, thanks

ericdill:
@campagnola should be good to go!

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
```diff
@@ -28,26 +26,41 @@ def test_ViewBox():
...
     app.processEvents()
+    
     w = vb.geometry().width()
     h = vb.geometry().height()
+    view1 = QRectF(0, 0, 10, 10)
     size1 = QRectF(0, h, w, -h)
-    assertMapping(vb, view1, size1)
-    
-    # now lock aspect
-    vb.setAspectLocked()
```
campagnola: What happened to this?

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
```diff
@@ -1,19 +1,17 @@
...
 
-def assertMapping(vb, r1, r2):
-    assert vb.mapFromView(r1.topLeft()) == r2.topLeft()
-    assert vb.mapFromView(r1.bottomLeft()) == r2.bottomLeft()
-    assert vb.mapFromView(r1.topRight()) == r2.topRight()
-    assert vb.mapFromView(r1.bottomRight()) == r2.bottomRight()
-
-def test_ViewBox():
-    global app, win, vb
+def setup_module():
```
campagnola: With this setup/teardown, the test results are order-dependent (because the view mapping may change depending on the state of the viewbox from the prior test). So I think either each test needs to create its own window+viewbox, or they need to stay as a single function to ensure their order is fixed.

campagnola: (for example, what happens if this is tested with multiprocessing enabled?)

ericdill: There did appear to be some bad state management going on.

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
ericdill: It got dropped by accident

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
```diff
@@ -76,10 +57,55 @@ def test_ViewBox():
...
+    win.resize(200, 200)
+    win.show()
+    vb = win.addViewBox()
     
+    # set range before viewbox is shown
+    vb.setRange(xRange=[0, 10], yRange=[0, 10], padding=0)
     
-if __name__ == '__main__':
-    import user,sys
-    test_ViewBox()
```
ericdill: What was the purpose of this `__main__` block?  Also `user` is a module that does not exist in python 3, so this is not python 2/3 kosher

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
ericdill: I switched to using the `setup_function` and `teardown_function` pytest functions to create an isolated environment for each test.

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
campagnola: It just lets me execute the script directly for debugging. Doesn't really need to be py3 compatible.

ericdill: Ah ok. Do you to be able to execute this file with `python test_ViewBox.py`? If so, I will need to rework this test file since this now uses the `setup_function` and `teardown_function`.

I think that `py.test test_viewBox.py` provides the same (or similar) functionality as executing the script without pytest.

campagnola: I wouldn't worry about it. I usually rewrite these blocks to set up a particular state every time I need to fix something, so it's fine if it does nothing for now. I usually start with `py.test --pdb`, but sometimes this is not convenient, particularly because the pdb event loop does not play nicely with the PyQt interactive event loop.

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
```diff
@@ -76,10 +57,62 @@ def test_ViewBox():
...
-        
\ No newline at end of file
+    # required to make mapFromView work properly.
+    qtest.qWaitForWindowShown(win)
+    
+    g = pg.GridItem()
+    vb.addItem(g)
+
+    g = pg.GridItem()
```
campagnola: duplicate

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
```diff
@@ -76,10 +57,62 @@ def test_ViewBox():
...
+
+
+function_set = set([test_resize, test_wide_resize, test_tall_resize,
+                    test_aspect_ratio_constraint])
+                    
+@pytest.mark.parametrize('function', function_set)
+def setup_function(function):
+    print('\nsetting up function %s' % function)
+    global app, win, vb, QRectF
+    app = pg.mkQApp()
```
campagnola: app should only be built once, and should never be exited in tests.

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
```diff
@@ -1,53 +1,27 @@
...
-    # required to make mapFromView work properly.
-    qtest.qWaitForWindowShown(win)
-    
-    g = pg.GridItem()
-    vb.addItem(g)
-    
+def test_resize():
+    # test resize
+    win.resize(400, 400)
     app.processEvents()
```
campagnola: There's an order issue here--previously this was tested before setting the aspect ratio constraint, but now it is after?

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
ericdill: Ah, I now see that the `pyqtgraph.mkQApp()` function is smart enough to look for an already running instance first.

campagnola: Even if it wasn't, the "official" rules for QApplication are that you may only create one per process, and it may only be exited once.

ericdill:
I over-engineered `test_ViewBox.py`.  How about I just do the minimum change that splits out the failing test into its own function

campagnola:
That sounds great to me :)

ericdill:
ok, so 2b07556 is the final overengineered solution that does produce the same initial error (though it is skipped) and  cb326c4 is the minimum change without just copy/pasting code. Assuming that it fails travis correctly, I will add a skipif decorator to `test_limits_and_resize`.

Thank you for the patience. Apparently I lose my way easily when diving into the test suite of a new library

ericdill:
@campagnola Can you have another look at `test_ViewBox.py`? Should pass muster this time

campagnola:
Awesome, thanks!

PR #194, opened 2015-07-22
Use glColor instead of mkColor to set GLViewWidget background color.

ixjlyons:


campagnola:
Looks good to me. The travis failures appear to be a problem on their end, not ours..

PR #196, opened 2015-07-24
Fixed ImageView custom axes.

derricw:
Custom axes, for example `{"t": 3, "c": 2, "x": 0, "y": 1}`, don't currently work with the `ImageView` widget.  It messes up the time slider, the key binds, and `play()` functionality because they all assume that the time axis is 0.

A simple way to fix this is to just adjust the `setImage` method to always transpose the image data after checking the `axes` kwarg, and to do that before the `tVals` attribute is set.

This minor adjustment fixed this issue for me.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,11 @@
+[run]
+source = 
+    pyqtgraph
+
+[report]
+omit =
+    */python?.?/*
+    */site-packages/nose/*
+    *test*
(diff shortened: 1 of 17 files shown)


campagnola:
This looks great, @derricw. Do you think you can add a unit test to cover these changes?

derricw:
Closing.  This is super old and has already been fixed by someone else.

PR #199, opened 2015-07-27
prevent divide-by-zero in PlotDataItem downsample

rabryan:
In the edge case of a PlotDataItem with only a single data point,
don't downsample since this leads to a divide by zero exception


campagnola:
This PR is mixed up with some unrelated changes..

PR #201, opened 2015-07-28
Smoother realtime scrolling via quantized downsampling option; support for individual min and max downsampling 

rabryan:
When downsampling, the indices chosen to bucket over change
whenever the x data is updated. This causes the plot line
to 'wiggle' when plotting a real-time data stream.

The "quanitizedDownsample" option snaps the bucket centers
to deterministic points such that the downsampled values
for the same underlying data don't change as new data is
added.

Individual options for min and max downsampling are now supported.

This also switches to using midpoint x-values
for downsampled data instead of the left edge of the bucket.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -111,6 +111,9 @@ class PlotDataItem(GraphicsObject):
                              multiple line segments per pixel. This can improve performance when
                              viewing very high-density data, but increases the initial overhead 
                              and memory usage.
+            quantizedDownsample  (bool) If True, downsampling will occur at deterministic
+                             x-intervals when updating data.  This allows for a smoother
+                             representation of scrolling data.
             clipToView       (bool) If True, only plot data that is visible within the X range of
                              the containing ViewBox. This can improve performance when plotting
                              very large data sets where only a fraction of the data is visible
@@ -170,6 +173,7 @@ class PlotDataItem(GraphicsObject):
             'autoDownsample': False,
             'downsampleMethod': 'peak',
             'autoDownsampleFactor': 5.,  # draw ~5 samples per pixel
+            'quantizedDownsample': True, #ensures downsample points align across scrolls
             'clipToView': False,
             
             'data': None,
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -111,6 +113,9 @@ def __init__(self, *args, **kargs):
                              multiple line segments per pixel. This can improve performance when
                              viewing very high-density data, but increases the initial overhead 
                              and memory usage.
+            quantizedDownsample  (bool) If True, downsampling will occur at deterministic
```
campagnola: Sphinx won't be able to render this.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -569,14 +578,41 @@ def getData(self):
                         x = x[x0:x1]
                         y = y[x0:x1]
                     
-            if ds > 1:
+        
+            if len(x) > 1 and ds > 1:
+                n = len(x) // ds
+                if self.opts['quantizedDownsample']:
+                    i0 = np.searchsorted( x, ds*(1 + x[0]//ds ))
```
campagnola: I think this will fail if x is floating point?

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -569,14 +578,41 @@ def getData(self):
...
+                    if i0 < x[-1]:
+                        indices = np.arange(i0, n*ds, ds)
+                    else:
+                        indices = [0]
+                    """use midpoint of x downsample points"""
+                    x = x[indices] + float(x[-1]-x[0]) / (len(x)-1)
```
campagnola: Why use the midpoint?

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -569,14 +578,41 @@ def getData(self):
...
+                    x = x[indices]
+
                 if self.opts['downsampleMethod'] == 'subsample':
-                    x = x[::ds]
-                    y = y[::ds]
-                elif self.opts['downsampleMethod'] == 'mean':
-                    n = len(x) // ds
-                    x = x[:n*ds:ds]
-                    y = y[:n*ds].reshape(n,ds).mean(axis=1)
+                    y = y[indices]
```
campagnola: This will probably cause a performance hit.. need some benchmarking to be sure.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -569,14 +578,41 @@ def getData(self):
...
+                        downsampler = np.mean
+                    elif self.opts['downsampleMethod'] == 'max':
+                        downsampler = np.max
+                    else:
+                        downsampler = np.min
+ 
+                    n = len(indices)
+                    y1 = np.zeros(n)
+                     
+                    for i in np.arange(1, n):
```
campagnola: Same here.

pyqtgraph/graphicsItems/PlotDataItem.py
rabryan: np.searchsorted isn't a search for an exact match -- its just a comparison based sort for inserting in-order

campagnola: Right, but if my x values are all in the range (0, 1) then the expression `ds*(1 + x[0]//ds` always evaluates to `ds`, and the bin edges are not quantized.

pyqtgraph/graphicsItems/PlotDataItem.py
rabryan: I guess it could be an option, but (especially with the mean) I think it makes more sense that a value calculated over a bucket interval should be plotted at the bucket's midpoint

PR #206, opened 2015-08-02
[py2/py3] dict.iteritems() --> six.iteritems(dict)

ericdill:


examples/test_examples.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import print_function, division, absolute_import
+import six
```
campagnola: I'd rather not add a dependency for this.. we can embed `six.py` under `pyqtgraph/util`, though.
Presently I use `pyqtgraph/python2_3.py` for this purpose (as it predates my awareness of `six`).

lidstrom83:
Two things that jumped out:
1. Needs an import of six in pyqtgraph/multiprocess/processes.py
2. Needs the six dependency removed from install_requires in setup.py

ericdill:
@lidstrom83 good catches, thanks!

@campagnola bump, this is probably ready to go, have another look?

lesauxvi:
ping @campagnola . Ready to go?

PR #207, opened 2015-08-02
WIP: Py26compat

ericdill:
@campagnola How important is python 2.6 compatibility?  I started tackling this problem before I asked if you are planning on keeping python 2.6 support alive. I'll wait to hear back from you before I do any more work on this PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,11 @@ install:
     - pip install pytest-xdist # multi-thread py.test
     - pip install pytest-cov # add coverage stats
     
+    # required for example testing on python 2.6
+    - if [ "${PYTHON}" == "2.6" ]; then
+        pip install importlib
+      fi;
+    
     # Debugging helpers
     - uname -a
     - cat /etc/issue
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ env:
     # Enable python 2 and python 3 builds
     # Note that the 2.6 build doesn't get flake8, and runs old versions of
     # Pyglet and GLFW to make sure we deal with those correctly
-    #- PYTHON=2.6 QT=pyqt TEST=standard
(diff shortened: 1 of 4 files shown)


campagnola:
I'd certainly like to keep it alive a while longer--usually the people stuck on 2.6 are also the ones with very large install bases. That said, feel free to just open a general "fix 2.6" issue and move on if this turns out to be a lot of work.

ericdill:
Great, thanks.  I'll keep hacking on this and see how far I get

ericdill:
Well that sure is an unhelpful test name... https://travis-ci.org/ericdill/pyqtgraph/jobs/73963240#L697

ericdill:
python 2.6 does not have `weakref.WeakSet`. In [test_stability.py] (https://github.com/pyqtgraph/pyqtgraph/blob/develop/pyqtgraph/tests/test_stability.py#L37), you are using this.  There is a WeakSet backport available on pypi, so it could be included in the requirements or shipped with the project. What would you like to do here?

https://pypi.python.org/pypi/weakrefset

campagnola:
In the past I have just used `WeakKeyDictionary` as a replacement for WeakSet. 
But on the other hand, I don't think test_stability needs to be 2.6 compatible--it doesn't contain any discoverable unit tests.

campagnola:
@ericdill, what's the status on this? It looks like tests will pass once the WeakSet is removed. If you're busy, I can take it the last step..

ericdill:
Ah thanks for circling back to this. Should I just get rid of `allWidgets` or turn it into a `WeakKeyDict` and modify the assignment on line 102?

campagnola:
A `WeakKeyDictionary` would be best. Thanks!

ericdill:
Alright, using a `WeakKeyDictionary`.  I'm using the widgets as the keys and using `1` as the value.  Seem sensible enough?

PR #208, opened 2015-08-03
Simpify ExampleLoader code and close #133

### LINKED ISSUE #133: Demo codeView font too small
Please increase the size of the text in the code view. It's so small that I have a very hard time reading it.

Linux Mint 17.1, pyqt, Python 3.4.2 (Anaconda)

Thanks

ericdill:
- Remove pre-compiled example loader for pyqt4, pyqt5 and pyside.
- Add a font-selector to the example loader
- Closes #133.


ericdill:
So I thought that this was going to be a simple fix.  Then I couldn't figure out how to compile the `exampleLoaderTemplate.ui` file into the `*_pyqt4.py`/`*_pyqt5.py`/`*_pyside.py` files.  How did you compile these originally?  Am I assuming correctly that I need all three qt versions installed in different environments so that I can convert the .ui to the three .py's?

Because I'm stubborn, I wanted to figure out how to load the .ui file at run-time so that the pre-compiled *.py files were not necessary.  Travis is failing because of path issues (ugh!). What are your thoughts on pre-compiling the ui files versus loading them at run-time?

campagnola:
Generally I prefer to pre-compile all the .ui files because I don't trust that all systems have uic available. Perhaps this is not a real problem. I have definitely found that the pyuic scripts are not always available, but this might not extend to the uic library itself..

In any case, I use `pyqtgraph/tools/rebuildUi.py` to generate all of the compiled .py files.

PR #213, opened 2015-08-09
do not close non-closable docks that are floating

sbourdeauducq:
Non-closable docks that are floating can still be closed by closing their window. This is a pretty serious bug as it may render the application completely unusable when the improperly closed docks contain important widgets.

This patch fixes the problem by moving non-closable docks to the home area when they are floating and their window is closed.


lidstrom83:
Seems like a nice way to handle this case.  PR looks good to me.

2xB:
Thank you for this Pull Request! This feature was later implemented again and merged with PR #657, so I close it for now.

PR #215, opened 2015-08-12
Fix to be able to use scatterplot in python 2.6

luddek:
Problem was that memoryview is used and does not exist in python 2.6


j9ac9k:
I know this PR was made a while ago, but as part of house-keeping I'm going to close this out as pyqtgraph no longer supports python 2.6.

PR #219, opened 2015-08-25
Ignore wheel events in GraphicsView if mouse disabled

rabryan:
This allows parent dialogs to receive these events
if needed.


campagnola:
Thanks again!
I think I'll fix the other three mouse events as well..

PR #224, opened 2015-08-31
Fix import of OrderedDict for old python API

vallsv:
Hi,

This file do not use your backport of OrderedDict (which is already part of pyqtgraph). This fix only use your default way to import OrderedDict.

The file ./examples/relativity/relativity.py should also be fixed in the same way.

We are using pyqtgraph tag 0.9.10 in production to have something stable. This bug is also part of this tag.

Thanks a lot.

### DIFF
--- a/pyqtgraph/parametertree/SystemSolver.py
+++ b/pyqtgraph/parametertree/SystemSolver.py
@@ -1,4 +1,4 @@
-from collections import OrderedDict
+from pyqtgraph.pgcollections import OrderedDict
 import numpy as np
 
 class SystemSolver(object):


lidstrom83:
Looks good to me.

lidstrom83:
Just noticed this is already included in https://github.com/pyqtgraph/pyqtgraph/pull/207

campagnola:
Thanks @vallsv, but I'm going to close this since it's a duplicate.
FYI: All imports inside `pyqtgraph` are required to be relative--I (and a few others, I think) use this to allow pyqtgraph to be embedded as a subpackage in larger projects.

PR #226, opened 2015-09-04
invalid slice fix

fedebarabas:
I don't understand why this line used to work without problems. But for the last few days it was crashing my HistogramLUTItem because np.ceil(self.image.shape[0] / targetImageSize) returned a float


lidstrom83:
Out of curiosity, how can I reproduce the crash?  As far as I can tell it has to be this line:

    stepData = self.image[::step[0], ::step[1]]

that causes the problem, but I don't have any problem doing array slicing with float64s...

campagnola:
Looks good to me. Might be another numpy version change..?

PR #227, opened 2015-09-06
Multiprocess cleanup

campagnola:
Minor updates to multiprocessing system:
* Debugging messages now avoid expensive string processing if debugging is disabled
* Unicode and bytes types added to the default no-proxy list (these types are serialized and passed by value instead of by proxy)
* Check for invalid arguments to `ObjectProxy._setProxyOptions`
* Fix error messages caused by trying to delete proxy objects after connection has closed


lidstrom83:
I couldn't find any issues with this one.

campagnola:
Thanks for the look-over, @lidstrom83!

PR #231, opened 2015-09-18
Python3 support for save and load of flowchart

duguxy:
- Fix errors encountered while trying to save and load a flowchart(.fc) file.
- Automatically add ".fc" to the end of the file name if omitted.

### DIFF
--- a/pyqtgraph/flowchart/Flowchart.py
+++ b/pyqtgraph/flowchart/Flowchart.py
@@ -514,7 +514,6 @@ class Flowchart(Node):
             return
             ## NOTE: was previously using a real widget for the file dialog's parent, but this caused weird mouse event bugs..
             #fileName = QtGui.QFileDialog.getOpenFileName(None, "Load Flowchart..", startDir, "Flowchart (*.fc)")
-        fileName = unicode(fileName)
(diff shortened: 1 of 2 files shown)


pyqtgraph/flowchart/Flowchart.py
```diff
@@ -535,7 +536,9 @@ def saveFile(self, fileName=None, startDir=None, suggestedFileName='flowchart.fc
             self.fileDialog.fileSelected.connect(self.saveFile)
             return
             #fileName = QtGui.QFileDialog.getSaveFileName(None, "Save Flowchart..", startDir, "Flowchart (*.fc)")
-        fileName = unicode(fileName)
+        if not fileName.endswith('.fc'):
```
campagnola: If I were to manually call `flowchart.saveFile('filename')`, I would not expect any `.fc` to be added automatically. Can you modify this such that the extension is only added to user-entered file names? You can connect `fileDialog.fileSelected` to a different method (maybe `saveFileSelected` or similar) that does the extension check and then calls `saveFile`.

campagnola:
Looks good except for my comment. Thanks!

j9ac9k:
Implemented with #886

PR #232, opened 2015-09-18
fix repetitive update() of nodes with multiple input

duguxy:
This commit fixes the issue below:
For a flowchart with connections between nodes like
node A -> node B -> node D
node A -> node C -> node D
The update() method of node D would be called twice if node A is changed and propagate the change through two different paths. This should not be the behavior expected. The update of D should only be called once after processing node B and node C.


campagnola:
Looks good! Thanks for digging in, that can't have been easy code to read ;)

PR #235, opened 2015-10-01
Axis fixes

JenSte:
These changes correct two problems the AxisItem has:
- It never shrinks when the space for the tick labels is not needed any more: example
- Depending on the zoom, sometimes a wrong value is used to calculate the distance between the axis label and the tick labels: example

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -910,6 +910,7 @@ class AxisItem(GraphicsWidget):
             #textHeight = self.style['tickTextHeight'] ## space allocated for horizontal text
             
         textSize2 = 0
+        last_textSize2 = 0
         textRects = []
         textSpecs = []  ## list of draw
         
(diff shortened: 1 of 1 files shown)


JenSte:
GH somehow f***ed up the links to the example images, here they are again:
https://www.dropbox.com/s/mdiodhlo4oap0tn/axis1.png?dl=0
https://www.dropbox.com/s/fnovmzngx55zv1f/axis2.png?dl=0

lidstrom83:
This looks good to me.

lesauxvi:
I agree with you, it seems ok!

campagnola:
I like these changes, and I think they should be merged. However, the reason I never fixed these issues in the first place is: if you have a plot that is rapidly auto-scaling, you can end up in a situation where the axis becomes completely unreadable because it changes size so quickly. So merging this code will introduce a new regression bug (and generally I prefer to keep existing bugs over regressions). 

A possible fix would be to allow the axis to shrink, but only after a certain amount of time has passed?

lesauxvi:
Ok Luke, I did not see this aspect! 

Could it be possible to circumvent this problem by adding an extra parameter in the constructor of the axis item (and thus in all the classes that use it) to allow it to shrink automatically when requested by the user (typically when the auto-scale is disabled)? Of course, the default value should be set to a value such as the default behavior is the actual one (no shrinkage)... Is it a good idea?

campagnola:
@lesauxvi, that's a good idea!

lesauxvi:
@JenSte , do you think you can take into account the remark mentionned above?

PR #239, opened 2015-10-23
Update functions.py

DavidLP:
BUG: fix scaling with numpy 1.10

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -826,7 +826,7 @@ def rescaleData(data, scale, offset, dtype=None):
         #p = np.poly1d([scale, -offset*scale])
         #data = p(data).astype(dtype)
         d2 = data-offset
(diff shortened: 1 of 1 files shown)


lidstrom83:
I've verified the bug (by running the ImageItem example with numpy 1.10 installed).  This fix works for me.

Not sure why Travis fails though.

ericdill:
Travis is failing due to an exception raised by the coverage library. While I am unclear why this exception is being raised, it is definitely safe to ignore

mivade:
Thanks for this. Just to add for cross-referencing, this addresses bug #243.

campagnola:
An in-place multiply should be faster and use less memory. There's a good discussion of the numpy changes that led to the bug and a suggested workaround here:
https://mail.scipy.org/pipermail/numpy-discussion/2012-September/063926.html

PR #241, opened 2015-10-25
TableWidget key press events

### LINKED ISSUE #240: arrow key navigation in table widget
Hi. Good job with pyqtgraph!. I like it!.
According to the manual of pyqt: "You can navigate the cells in the table by clicking on a cell with the mouse, or by using the arrow keys. Because QTableView enables tabKeyNavigation by default, you can also hit Tab and Backtab to move from cell to cell.".

However, when I create a table in pyqtgraph, I cannot navigate with arrow keys. This is true even for the TableWidget example in pyqtgraph.

Do you know how to make this work without the need to handle key events?

Hanan.

ixjlyons:
Key press events should be passed up to the parent class so arrow key and tab key navigation works. Fixes #240.

### DIFF
--- a/pyqtgraph/widgets/TableWidget.py
+++ b/pyqtgraph/widgets/TableWidget.py
@@ -353,11 +353,11 @@ class TableWidget(QtGui.QTableWidget):
         self.contextMenu.popup(ev.globalPos())
         
     def keyPressEvent(self, ev):
-        if ev.text() == 'c' and ev.modifiers() == QtCore.Qt.ControlModifier:
+        if ev.key() == QtCore.Qt.Key_C and ev.modifiers() == QtCore.Qt.ControlModifier:
             ev.accept()
-            self.copy()
+            self.copySel()
         else:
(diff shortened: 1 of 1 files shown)


lidstrom83:
Some issues I noticed:

1. Ctrl + c doesn't actually trigger the if clause
2. If it did, the copy method doesn't exist
3. Assuming copySel was what was intended, the value stored in the clipboard would ultimately be overwritten by a different value produced by the parent method

I think the following would deal with these issues:

    def keyPressEvent(self, ev):
        if ev.key() == QtCore.Qt.Key_C and ev.modifiers() == QtCore.Qt.ControlModifier:
            ev.accept()
            self.copySel()
        else:
            QtGui.QTableWidget.keyPressEvent(self, ev)

ixjlyons:
Good catch. I tested with only one cell selected, so it appeared to work.

campagnola:
Looks good to me. 
Thanks @ixjlyons and @lidstrom83!

PR #249, opened 2015-11-25
Fixing numpy bug in functions.py

Luindil:
Fixing an error which is caused by a new numpy version: 
d2 *= scale 
gave:
"TypeError: Cannot cast ufunc multiply output from dtype('float64') to dtype('int16') with casting rule 'same_kind'"
however, d2 = d2 * scale does seem to work without problems


campagnola:
Thanks for the tip! The proposed solution creates an extra copy, though. We should get better performance / memory usage with an in-place multiply. This can be done by manually specifying a different casting rule:
https://mail.scipy.org/pipermail/numpy-discussion/2012-September/063926.html

campagnola:
Duplicate of #239.

Issue #243, opened 2015-10-27
FutureWarning with numpy 1.10


Nodd:
I have this warning:

```
/usr/lib/python3.5/site-packages/pyqtgraph/functions.py:1317: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  if connect == 'all':
```

ixjlyons:
This has been fixed by #264 / #273

PR #264, opened 2016-01-12
Fix Numpy FutureWarning.

u55:
Don't accidentally compare an array to string. Fixes issue #243 when `connect=='finite'`.


u55:
Darn, I just realized that the fix is still incomplete. Creating another commit.

u55:
Okay, I think we're good now.

PR #266, opened 2016-01-15
new markers

lesauxvi:
Addition of several markers that can be used to plot data and creation of an example highlighting all the markers currently available in pyqtgraph.


campagnola:
1. I'm slightly conflicted about the use of "markers" because these are referred to as "symbols" in pyqtgraph. However matplotlib and vispy both use "markers", so it's possible people will be looking for that terminology.. 

2. Can you make the markers larger in the example? Some of them are difficult to tell apart.

3. Can you add a legend to the example that lists the symbol strings? You just need to add `plot.addLegend()` and  `name="xxx"` to each of the `plot()` calls.

lesauxvi:
1. You are the boss ;-). If you prefer the terminology 'symbol', I'll change it.
2. Of course!
3. OK. Good idea.

lesauxvi:
List of the modifications:
   - 'markers' has been renammed as 'symbols'
   - the marker size has been set to 14 (default value was used before, i.e. 10)
   - legend added!

Any other suggestions?

campagnola:
Looks good, thanks! I also added this to the example menu.

PR #267, opened 2016-01-15
InfiniteLine enhancement

lesauxvi:
Add some new features to InfiniteLine and an example illustrating some of them.


pyqtgraph/graphicsItems/InfiniteLine.py
```diff
@@ -37,79 +78,125 @@ def __init__(self, pos=None, angle=90, pen=None, movable=False, bounds=None):
...
-        Note that the use of value() and setValue() changes if the line is 
+
+        Note that the use of value() and setValue() changes if the line is
         not vertical or horizontal.
         """
         self.angle = ((angle+45) % 180) - 45   ##  -45 <= angle < 135
-        self.resetTransform()
-        self.rotate(self.angle)
+        # self.resetTransform()   # no longer needed since version 0.9.11
+        # self.rotate(self.angle) # no longer needed since version 0.9.11
```
campagnola: Any reason for this? This is an API change--previously, any child of an InfiniteLine would inherit its rotation. It also results in more complex mouse intersection / bounds calculation code.

pyqtgraph/graphicsItems/InfiniteLine.py
```diff
@@ -37,79 +78,125 @@ def __init__(self, pos=None, angle=90, pen=None, movable=False, bounds=None):
...
+        self.angle = ((angle+45) % 180) - 45
+        if textColor is None:
+            textColor = (200, 200, 200)
+        self.textColor = textColor
+        self.location = textLocation
+        self.shift = textShift
+        self.label = label
+        self.format = textFormat
+        self.unit = unit
+        self._name = name
```
campagnola: What is the purpose of `self._name` ?

pyqtgraph/graphicsItems/InfiniteLine.py
```diff
@@ -240,3 +362,122 @@ def setMouseHover(self, hover):
...
+        Set the location of the textItem with respect to a specific axis. If the
+        line is vertical, the location is based on the normalized range of the
+        yaxis. Otherwise, it is based on the normalized range of the xaxis.
+
+        ==============   ==============================================
+        **Arguments:**
+        loc              the normalized location of the textItem.
+        ==============   ==============================================
+        """
+        if loc > 1.:
```
campagnola: `loc = np.clip(loc, 0, 1)`

pyqtgraph/graphicsItems/InfiniteLine.py
```diff
@@ -240,3 +362,122 @@ def setMouseHover(self, hover):
...
+        state            If True, the label is shown. Otherwise, it is hidden.
+        ==============   ==============================================
+        """
+        if state:
+            self.text.show()
+        else:
+            self.text.hide()
+        self.update()
+
+    def setLocation(self, loc):
```
campagnola: `setLocation` and `setShift` are ambiguous names (especially compared to `setPos`). These should make it clear that the *label* will be affected, like `setLabelLocation` or `setTextLocation`.  For further simplicity, it would be nice for these two methods to be combined into one, like `setTextLocation([0.8, 1.0])`, where the first value is always along the line axis, and the second is always orthogonal (or however you think is most intuitive).

pyqtgraph/graphicsItems/InfiniteLine.py
lesauxvi: The purpose is to define a name, that can set by the user. I added this because I'm writing a list manager that will allow the user to interact with the content of a plotwidget (and the name of the item is based on this property). I can remove it if you prefer.

campagnola:
This is looking good. I should explain, though, that `InfiniteLine` is one of a few classes that was surprisingly tricky to get right; it went through a few different implementations before arriving at its current state. I am totally open to the possibility that the implementation I have chosen is *not* the best one, so do let me know if you think my suggestions don't make sense.

pyqtgraph/graphicsItems/InfiniteLine.py
lesauxvi: Yes. InfiniteLine no longer inherits GraphicsObject, but UIGraphicsItem, which means that the object needs to be defined in the viewbox coordinates, and no longer in its item coordinates. There is may be a clever way to do what i did, but I did not find it out.

pyqtgraph/graphicsItems/InfiniteLine.py
lesauxvi: Thank you. I missed that!

pyqtgraph/graphicsItems/InfiniteLine.py
lesauxvi: These modification seem very nice to me. I'll try to combine these methods into a single one, with an appropriate name!

pyqtgraph/graphicsItems/InfiniteLine.py
```diff
@@ -240,3 +362,122 @@ def setMouseHover(self, hover):
...
+        ==============   ==============================================
+        **Arguments:**
+        format           Any format compatible with the new python
+                         str.format() format style.
+        ==============   ==============================================
+        """
+        self.format = format
+        self.update()
+
+    def setUnit(self, unit):
```
campagnola: For consistency with the rest of the library, consider `suffix` instead of `unit`.

I am also wondering if this is necessary, given the availability of `setFormat` ? There's a general problem here that exists throughout the library that I have not yet found a good solution for:  how to let the user specify text formatting for numerical values (for example, in AxisItem, SpinBox, and ValueLabel). In most places, we use `siFormat` to generate the strings, but it would be even nicer to expose this via some string formatting syntax like `str.format()` uses. Imagine my line is at x=0.001, and I want the label to read "val: 1 mm":

    line.setFormat('val: {value}m')  # unscaled units
    line.setFormat('val: {siScaledValue}m')   # automatically scaled to 1 mm
    
Those two possibilities would be easy enough to support, but anything beyond that would be too difficult and perhaps just requires the user to provide a callable to generate the string. Perhaps `setFormat(callable)` should also be allowed? I don't know..

pyqtgraph/graphicsItems/InfiniteLine.py
lesauxvi: I should also emphasize that the rotation inheritance is something I did not want in this case because the label is a TextItem whose parent is self. If I apply a transformation, as it was done before, that transformation is also applied to the label. This behaviour results in something ugly (label rotated and scaled, not really readable actually). These API changes were a necessity considering the modification I wanted to add to this object (no such major changes will be necessary for the other modification I plan to propose).

I did not find another way (and I do not know if there is another way).

pyqtgraph/graphicsItems/InfiniteLine.py
lesauxvi: Your suggestions make perfect sense. It is always a pleasure to have your feedback. New commit incomming (except for your first remark, because I'm afraid I can not do anything.

allebacco:
@campagnola The current implementation of `InfiniteLine` is pretty slow. I didn'try the updates from @lesauxvi, but I think that the performance can't be improved with more things that must be managed.
So, my question is: do we need to add these features to the current `InfiniteLine`, or is it possible to define a new item name for example `AdvancedInfiniteLine` that can inherits from `InfiniteLine` (if possible) with these new features?

@lesauxvi : I'm sorry, I don't want to attack your work on this PR, that seems to be great, but, since during my work I'm using a lot of `LinearRegion` items (that contains 2 `InfiniteLine` each), I would like that my code will run with the same performance level.

lesauxvi:
No problem @allebacco. I've never encountered any performance issue with InfiniteLine. I never had to add many InfiniteLine items though. So, your question is  important and clearly must be discussed before going further (the PR is still pending, as I have to modify few things).

One way to check your concerns would be to compare their performance. The question is: how can we do that and what can we compare (I never had to check performance of my code - even if sometimes it is necessary)? Any suggestions?

allebacco:
In my graphs I could have up to 100 `InfiniteLine` instances.
Please, keep in mind that `InfiniteLine` is used by `LinearRegionItem`  for drawing its edges and for implementing resizing with mouse events.

In Pyqtgraph example, you can take a look at **Basic Plotting** demo. The **Region Selection** plot use a `LinearRegionItem` for highlighting an area of the graph. Is it still working with your modifications?

In the last hours i sent a PR #275 (that you have already seen). Please try the example code with your modifications for comparing the performance.

lesauxvi:
Yes, LinearRegionItem is still working with my modifications. I had to change few things but it is working.

I check the performance right now.

lesauxvi:
OK, just checked. On my laptop: here are the results (when hovering nothing)
   - with your modifications : 125fps (almost no drops when hovering objects)
   - with the current version of pyqtgraph InfiniteLine : 20 fps (huge drop in performance when hevering objects)
   - with my implementation : 27 fps (huge drop in performance when hovering objects)

So there is a small performance increase compared to the current PyQtGraph version ;-), but I'm still way behind you! It could be VERY interesting to try to merge your modifications to this PR in order to see if we can even do better that 125 fps.

campagnola:
> @campagnola The current implementation of InfiniteLine is pretty slow. I didn'try the updates from @lesauxvi, but I think that the performance can't be improved with more things that must be managed.

@allebacco, adding new features to InfiniteLine should not slow it down at all (if they are implemented correctly). If you are not adding text labels to your lines, then you should see the same performance. And as you have already discovered, there are many other areas we can work on to improve performance :)

pyqtgraph/graphicsItems/InfiniteLine.py
campagnola: I am not opposed to it, especially because it follows the same conventions as line plots--we use names there to allow automatic generation of plot legends.

lesauxvi:
Apparently, the implentation I propose seem to improve the performance (this may need to be checked on other computers, I'll check at work this morning). May be, it is related to the new inheritance (UIGraphicsItem over GraphicsObject). What do you think @campagnola ?

pyqtgraph/graphicsItems/InfiniteLine.py
lesauxvi: That was basically my idea. Add a name attribute to every items that can be added to a plotItem (with eventually, a default name given when nothing is specified by the user). I'm anticipating something I judge useful : a listwidget that contains all the items contained in a PlotItem, with the possibility to interact with the items (especially change swap their Zvalue). If you think it is a good idae, we can keep the self._name attribute, otherwise, I remove it (no problem about that).

pyqtgraph/graphicsItems/InfiniteLine.py
campagnola: > I should also emphasize that the rotation inheritance is something I did not want in this case because the label is a TextItem whose parent is self.

TextItem should always appear unscaled and unrotated; only its position is inherited from its parent (if that's not the case, then we have a bug). I suggest to leave the painting and bounds calculation unchanged--it keeps the code simpler, it'll make integrating #275 much easier, and I also have some feature additions to InfiniteLine that I'd like to merge after your PR. This is my suggestion:

* paint / boundingRect probably don't need to change at all
* Add a TextItem as child of the line (but only if text is explicitly enabled)
* Use `GraphicsItem.viewTransformChanged()` to detect when the text position needs to be recalculated.
* Use `setPos()` to detect when the text needs to be updated.

lesauxvi: OK, I forgot something when I wrote the previous message. I tried to add directly a text with the painter in InfiniteLine, and I got the rotation and scale problem in that condition (when I think about it, this was a stupid idea). I did not try to add a TextItem in the initial InfiniteLine class, because I thougth it was not a good idea, and the object are not in the same coordinates... But this could not be a problem thanks to the mapTo method. 

There is still something I missed. In the current implementation of InfinteLine, every child will inherit its rotation. If I had a TextItem, I will have to set its parent to the InfiniteLine (so that when I delete the InifiniteLine, it will also delete the TextItem). So the TextItem will be rotated. I miss something, but what?

campagnola: Have you tried it? TextItem should ignore the rotation of its parent.

lesauxvi: I'm on it. And TextItem is ignoring the rotation of its parent. You were definitely right!

Just a quick question about your suggestions. How can I use the GraphicsItem.viewTransformedChanged() method you mentionned two comments above for the update of the textItem position?
By just adding something:

        GraphicsObject.viewTransformChanged(self)

in the code? But what next? What is called next? A paint event?

campagnola: That method will be called for you whenever the ViewBox transform changes, so you want something like:

        self.updateTextPosition()

campagnola: (and simply updating the position of the textItem will be enough to trigger a repaint)

lesauxvi: Great! Last question (and after, it will be ready for the commit) : I need to get the viewRange of the PlotItem that contains the InfiniteLine. I tried few things, but nothing works. What is the recommended way to get the parent of the InfiniteLine?

campagnola: Well, the direct parent of the InfiniteLine is `infiniteline.parent()`. The ViewBox that it lives inside is `infiniteline.getViewBox()`, and the bounding rectangle of the view, in the local coordinate system of the infiniteline,  is `infiniteline.viewRect()`. (these methods are defined in GraphicsItem, which has many other useful methods as well)

campagnola: Sorry, that's `infiniteline.parentItem()`, not `.parent()`.  (but I don't think you'll need to use this either way)

allebacco:
@campagnola I have tested the @lesauxvi implementation and it is faster than the current pyqtgraph code.
For me, it is ok to merge this PR and ignore PR #275, if the last is not compatible with @lesauxvi code.

pyqtgraph/graphicsItems/InfiniteLine.py
lesauxvi: OK, is works! Thank you. 

However, there is a problem with my current implementation. I can get the viewBox only when the infiniteLine has been added to the PlotItem using the addItem method, right? So, I can't initialize the textItem when I create the InfiniteLine object, but only after the addItem method has been called. The way to use the item would therefore be;
infLine = InfiniteLine()
plotItem.addItem(infLine)
infLine.setLabel(True)

That a not very clean. Which do you think?

campagnola: You just need to check for the presence of a viewbox--if either `getViewBox()` or `viewRect()` returns None, then do not update the text position. When the line is added to the view, then `viewTransformChanged()` will be called automatically and the text position will be set.

lesauxvi:
Without your modifications @allebacco, my proposed PR is a bit faster than the current pyqtgraph implementation. However, with your modifications, they are almost identical (check it . So, I think, we should take @campagnola suggestions into account (avoid API changes, if possible). I'm curently working on it.

pyqtgraph/graphicsItems/InfiniteLine.py
lesauxvi: OK, I also came to that solution!

lesauxvi: Actually, it is not working. I need to check the nature of the object returned by `getViewBox()` because it return the first time a `GraphicsWindow` object which does not have a `viewRange()` method. I need to use: `isinstance(self.getViewBox(), ViewBox)`. And then it works.

lesauxvi:
I just commited modification of InfiniteLine. Basically, the capabilities are the same as before, but I tried to include Luke suggestions. There remains two things to improve:
 - the format handling. I did not try to take Luke remarks into account. This will be done soon
 - there is something I did not manage to do properly (exact location of the label, when requested by the user), but I kinda prefer the proposed solution (the user can "drag" the label with the mouse in order to locate it wherever he wants).

lesauxvi:
Addition of @allebacco proposition (#PR267) into the code.

pyqtgraph/graphicsItems/InfiniteLine.py
```diff
@@ -240,3 +336,50 @@ def setMouseHover(self, hover):
...
+
+        if self.getViewBox() is not None and isinstance(self.getViewBox(), ViewBox) and self.textItem is not None:
+            self.updateTextPosition()
+
+    def showLabel(self, state):
+        """
+        Display or not the label indicating the location of the line in data
+        coordinates.
+
+        ==============   ==============================================
```
campagnola: sphinx will complain--the table lines need to be at least as wide as the contents of the table.

lesauxvi: Corrected!

campagnola:
It used to be that the text remained stationary in the viewport along the axis orthogonal to the line, right? That seems to be broken since 51b8be2.

lesauxvi:
Right! I should be able to fix the problem, but I do not know what is better... What od you prefer? A perfect control of the location or the possibility to locate approximately the object during the dragging?

campagnola:
For now I think it is easiest to let the user specify the location along the line as 0.0-1.0 relative to the view bounds. If there is a good use case for it, we could let these be dragged (optional; probably disabled by default), but it should probably never be the case that panning/zooming the viewbox itself causes the text to change position along the line.

lesauxvi:
OK, thanks for your feedback. So, there is still a bit of work to manage to do that. In its current state, the text can be dragged and does not change position during panning and zooming. I still need to find a way to block the text position.

lesauxvi:
Problem solved! I add a method to authorize the relocation of the label during the dragging.

lesauxvi:
Addition of an option to make the label draggable (follow the motion of the mouse during the dragging).

There is currently a bug in the implementation: the line must be dragged a first time in order to being able to locate precisely the label (just execute the *plottingItems.py* example and drag the vertical line). I'll try to correct the bug today, but you already have a first idea of the current implementation (and I'm looking forward your feedback)

lesauxvi:
Bug solved. The current implementation does correctly what it is supposed to. It should be possible to improve the code, but I thnink it is a good starting point. Waiting for your feedback!

campagnola:
So now if you drag the line, then the label becomes fixed to its location in the viewbox. Is that really what you intend? As I said above, it should probably never be the case that panning/zooming the viewbox itself causes the text to change position along the line. But perhaps you have a use case for this?

lesauxvi:
You should try the 'plottingItems' example to see the new features in action, it should be much more clear ;-). You have the two behaviours (draggable label fo the horizontal line and fixed location in the orhogonal direction of the line for the vertical one).

campagnola:
Yes, my comment above is in response to the latest commit.

lesauxvi:
Oups. The bug before the lastest commit was related to the placement of the label when an exact position along the line is requested (default behaviour when thet label is set to visible). Actually, the position was not the right one, except if the line is dragged (problem when zooming/panning the viewBox). 

Is the implementation suitable for publication?

campagnola:
See my comment above, I am still confused about why the label should move when the viewport is panned?

lesauxvi:
Oups (again). The label does not move when the viewport is moved or I do not know what you mean? Did you tried the example. If yes, can you illustrate with a picture?

campagnola:
In examples/plottingItems, when I pan the viewbox to the right, the red label also moves to the right. That's not what you see?

lesauxvi:
OK, I understand! It is very early in the morning in France ;-)!

That is what I intended to. This way, the user wan locate the label wherever he wants by a simple dragging action. This is not the default behaviour when the label is set to visible. Actually, I first had this behaviour due to a bug in my implementation, but I found the result quite interesting so I kept it as an option. If you do not like it, I can remove it.

This behaviour can be interesting, to my option, when you do not want the label to superpose your data for sake of lisibility. So you darg the line to a location where there is more room available.

campagnola:
Right, I understand why you want the label to be draggable. I do not understand why the label moves when the viewbox is panned ;)

Let me have a crack at it; I'll open a sub-PR.

lesauxvi:
OK. Waiting for your PR!

PR #269, opened 2016-01-18
let cprint handle bytestrings

alexcjohnson:
Something in pyqtgraph tried to write a `bytes` using cprint (python 3.5 on a mac) - might as well let it, yes?

I chose `'latin1'` just to be failsafe - I presume the only things people care about here (this is only used for debugging, right?) are ascii anyway.

### DIFF
--- a/pyqtgraph/util/cprint.py
+++ b/pyqtgraph/util/cprint.py
@@ -69,27 +69,21 @@ def cprint(stream, *args, **kwds):
     else:
         err = kwds.get('stderr', False)
 
-    if hasattr(stream, 'isatty') and stream.isatty():
-        if _WIN:
-            # convert to win32 calls
-            for arg in args:
-                if isinstance(arg, basestring):
-                    stream.write(arg)
-                else:
-                    kwds = WIN[arg]
-                    winset(stderr=err, **kwds)
-        else:
-            # convert to ANSI
-            for arg in args:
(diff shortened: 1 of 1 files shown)


pyqtgraph/util/cprint.py
```diff
@@ -69,27 +69,21 @@ def cprint(stream, *args, **kwds):
...
-                else:
-                    stream.write(ANSI[arg])
-    else:
-        # ignore colors
-        for arg in args:
-            if isinstance(arg, basestring):
-                stream.write(arg)
+    for arg in args:
+        if isinstance(arg, bytes):
+            arg = arg.decode('latin1')
```
campagnola: Probably unicode is a better bet here, yeah?

alexcjohnson: @campagnola sorry, didn't notice that anyone had looked at this PR! It's been a long time, I'm going to close it since I'm no longer involved in the project this was useful for.

If someone feels like picking this back up, my rationale for using `latin1` instead of `utf8` is that, in as far as this is (solely?) used for debugging, I didn't want it to ever throw an error. `utf8` has invalid byte sequences, but single-byte encodings do not, and `latin1` is probably the most likely single-byte encoding. Anyway if it's just debugging info the input should just be ASCII, which is the same in both encodings. Could certainly try `utf8` then fall back on `latin1` if it failed...

PR #270, opened 2016-01-26
Fix issue where numpy is (planning on) overloading `==` to operate elementwise, while `is` operates object-wise.

fake-name:
So, the numpy people have decided that overloading `==` is a good idea, and are now throwing a warning message every time you use `==` to check if an array is anything. It's annoying, silence it.

Very, very minor change.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1305,17 +1305,17 @@ def arrayToQPath(x, y, connect='all'):
     arr[1:-1]['y'] = y
 
     # decide which points are connected by lines
-    if connect == 'pairs':
+    if connect is 'pairs':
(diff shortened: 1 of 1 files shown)


kalvdans:
Using 'is' on strings are unsafe. This is a transcript from a Python 2.7.6 session:
>>> ''.join(["a", "b"]) == "ab"
True
>>> ''.join(["a", "b"]) is "ab"
False

fake-name:
@kalvdans - Well, that's freaking annoying.

campagnola:
Could be dealt with just by reversing the order of the if-blocks: start by checking `if isinstance(connect, ndarray)`, and then do the string checks afterward.

fake-name:
Yeah, that's what I wound up doing. 

Overloading `==` is really, really annoying.

PR #271, opened 2016-01-27
Allow pens to be set for ROI handles and hover

nmearl:
Small tweaks to allow pens to be passed when instantiating an ROI. The `handlePen` is the pen for the handles on the ROI, and `hoverPen` is the pen used while the mouse is hovering over the ROI shape (or its handles; i.e. the shape and its handles share the same hover pen).


lesauxvi:
I really like these changes and think they are suitable for integration in the devel branch.

nmearl:
Yeah there. I apologize if I sound pushy, but our group is using this as part of an application that's being distributed, and currently we have to work around this not being in the main `pyqtgraph` repo. Is there anything I can do to push this along a bit? Thank you!

lesauxvi:
Could you be more precise because these changes will not affect the default behaviour of the ROI class?

Moreover, these changes, if judged acceptable for publication by @campagnola , will be incorporated into the develop version of pyqtgraph and I guess you were talking about the master branch...

nmearl:
Yeah, sorry. Ignore the last comment. My local fork of `pygraph` is set with the branch as `master`, even though the upstream is `develop`. When I saw that I was behind on `pyqtgraph` commits, I had made the snap assumption that `pyqtgraph`'s `master` branch was being updated, so I was hoping that this could be incorporated as well. But obviously that was incorrect! Sorry about the confusion.

lesauxvi:
This looks good to me and a simple PR. What do you thing @ericdill and @campagnola ?

pyqtgraph/graphicsItems/ROI.py
```diff
@@ -71,6 +71,10 @@ class ROI(GraphicsObject):
                      is generally not necessary to specify the parent.
     pen              (QPen or argument to pg.mkPen) The pen to use when drawing
                      the shape of the ROI.
+    handlePen        (QPen or argument to pg.mkpen) The pen to use when drawing
+                     the ROI handles.
+    hoverPen         (QPen or argument to pg.mkpen) The pen to use while the
+                     mouse is hovering over the ROI shape or its handles.
```
campagnola: Can we also have `handleHoverPen` for completeness? 

..or maybe it would be nicer to provide `Handle.setPen` and `Handle.setHoverPen` instead..  I'd be happy with either approach.

lesauxvi: What about adding the two? It could be interesting to be able to customize the handles after their creation. What do you think @campagnola?

campagnola: Sure, having both would be nice too.

lesauxvi:
@campagnola may be you can merge this PR. The missing features can be added later... What do you think?

lesauxvi:
ping @campagnola and @ericdill . The missing features can easily be added in another PR.

ibrewster:
Sorry to bump an old thread, but add me to the list of people wanting this feature to be added. At the moment, looking at the source code, the handlePen is hard coded to a color that does not work at all for my use. Apparently I am going to have to edit the PyQtGraph code just to be able to use the ROI feature and be able to see the handles. Meanwhile, a solution was developed three years ago, but never implemented?

j9ac9k:
hi @ibrewster, sorry for the long delay; we've started picking up on development again in an attempt to get the next stable release out.  We'll look over the PR as we get the test suite up and running.

PR #272, opened 2016-01-30
Image performance

campagnola:
Improves performance of ImageItem for video display by rescaling the LUT instead of the image whenever possible.


campagnola:
Also added some related unit tests, moved a function from flowcharts to pg.functions (that was needed by ImageItem), and did some dusting in flowchart code while I was there.

PR #281, opened 2016-02-10
Fixing order of positions in colormap

lionel-martin:
numpy.interp (called in ColorMap.map) forces pos to be in increasing order. Not following this requirement does not raise any Error but provide unexpected behavior.


ericdill:
Seems pretty straightforward. @campagnola any unexpected consequences of doing this?

lesauxvi:
Right. What do you think @campagnola ?

campagnola:
Looks good to me! Thanks @lionel-martin !

PR #284, opened 2016-02-15
Image view work

meganbkratz:
Added convenience functions to ImageView for setting colormaps and preset gradients.

### DIFF
--- a/pyqtgraph/graphicsItems/GradientEditorItem.py
+++ b/pyqtgraph/graphicsItems/GradientEditorItem.py
@@ -25,8 +25,7 @@ Gradients = OrderedDict([
 ])
 
 def addGradientListToDocstring():
-    ### create a decorator so that we can add construct a list of the gradients defined above in a docstring
-    ### Adds the list of gradients to the end of the functions docstring
+    """Decorator to add list of current pre-defined gradients to the end of a function docstring."""
     def dec(fn):
         fn.__doc__ = fn.__doc__ + str(Gradients.keys()).strip('[').strip(']')
         return fn
@@ -482,7 +481,6 @@ class GradientEditorItem(TickSliderItem):
     def loadPreset(self, name):
         """
         Load a predefined gradient. Currently defined gradients are: 
-    
         """## TODO: provide image with names of defined gradients
         
         #global Gradients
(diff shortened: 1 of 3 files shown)


pyqtgraph/graphicsItems/GradientEditorItem.py
```diff
@@ -25,6 +24,14 @@
     ('grey', {'ticks': [(0.0, (0, 0, 0, 255)), (1.0, (255, 255, 255, 255))], 'mode': 'rgb'}),
 ])
 
+def addGradientListToDocstring():
+    ### create a decorator so that we can add construct a list of the gradients defined above in a docstring
+    ### Adds the list of gradients to the end of the functions docstring
```
ericdill: Can you rewrite this in a more standard docstring format with 
```
"""
Decorator to add a list of valid gradients to the function docstring
"""
```

Feel free to write whatever prose you want between the triple quotes, the above is merely a suggestion.

I like the very verbose function name that clearly defines what it is that this function does :smile:

pyqtgraph/imageview/ImageView.py
```diff
@@ -717,4 +717,14 @@ def menuClicked(self):
         if self.menu is None:
             self.buildMenu()
         self.menu.popup(QtGui.QCursor.pos())
-        
+
+    def setColorMap(self, colormap):
+        """Set the color map. *colormap* is an instance of ColorMap()"""
+        self.ui.histogram.gradient.setColorMap(colormap)
+
+    def setPredefinedGradient(self, name):
```
ericdill: Why not use the gradient decorator here too?

pyqtgraph/imageview/ImageView.py
```diff
@@ -717,4 +717,14 @@ def menuClicked(self):
         if self.menu is None:
             self.buildMenu()
         self.menu.popup(QtGui.QCursor.pos())
-        
+
+    def setColorMap(self, colormap):
+        """Set the color map. *colormap* is an instance of ColorMap()"""
```
ericdill: Can you write this docstring following the convention outlined [here] (https://github.com/pyqtgraph/pyqtgraph/blob/develop/CONTRIBUTING.txt#L42)?

ericdill:
Thanks for the PR. The decorator is clever, :+1:   Just a couple of documentation related comments

campagnola:
Merged, with a different colormap in the example.
Thanks @meganbkratz and @ericdill !

PR #275, opened 2016-02-02
Improve InfiniteLine drawing performance by caching the line and bounding rect.

allebacco:
This PR is for improving `InfiniteLine` drawing performance by caching the line and the bounding rect.
The performance has been improved for situations in which items close to the `InfiniteLine` are changing.

Please, try the following code before and after the PR to have an idea of the improvements.
```
#!/usr/bin/python

from pyqtgraph.Qt import QtGui, QtCore
import numpy as np
import pyqtgraph as pg
from pyqtgraph.ptime import time
app = QtGui.QApplication([])

p = pg.plot()
p.setWindowTitle('pyqtgraph performance: InfiniteLine')
p.setRange(QtCore.QRectF(0, -10, 5000, 20))
p.setLabel('bottom', 'Index', units='B')
curve = p.plot()

# Add a large number of horizontal InfiniteLine to plot
for i in range(100):
    line = pg.InfiniteLine(pos=np.random.randint(5000), movable=True)
    p.addItem(line)

data = np.random.normal(size=(50, 5000))
ptr = 0
lastTime = time()
fps = None


def update():
    global curve, data, ptr, p, lastTime, fps
    curve.setData(data[ptr % 10])
    ptr += 1
    now = time()
    dt = now - lastTime
    lastTime = now
    if fps is None:
        fps = 1.0/dt
    else:
        s = np.clip(dt*3., 0, 1)
        fps = fps * (1-s) + (1.0/dt) * s
    p.setTitle('%0.2f fps' % fps)
    app.processEvents()  # force complete redraw for every plot


timer = QtCore.QTimer()
timer.timeout.connect(update)
timer.start(0)


# Start Qt event loop unless running in interactive mode.
if __name__ == '__main__':
    import sys
    if (sys.flags.interactive != 1) or not hasattr(QtCore, 'PYQT_VERSION'):
        QtGui.QApplication.instance().exec_()
```

PR #286, opened 2016-02-22
combined infiniteline updates

campagnola:
* Performance improvements from #275
* New unit tests covering InfiniteLine
* Labels provided by #267, with modifications:
    - Label handling code moved out of InfiniteLine, to its own subclass of TextItem
    - Orientation-independent code; labels can be attached to lines at any angle
    - Labels can be dragged along line (but position along the line is not affected by the position of the line itself)
* Supporting changes:
   - TextItem now supports mouse interaction and rotation
   - TextItem may be automatically aligned to an axis of its parent


lesauxvi:
Thank you for your code modifications. I've s problem when I tried to execute the plottingItems.py example:

lesauxvi:
Oups, the image is not visible. I copy paste the error:

```
  File "/home/lesaux/Documents/ENSTA/Python/Tests/056_pyqtgraph-infiniteline-tests/pyqtgraph/graphicsItems/GraphicsObject.py", line 23, in itemChange
    self.parentChanged()
  File "/home/lesaux/Documents/ENSTA/Python/Tests/056_pyqtgraph-infiniteline-tests/pyqtgraph/graphicsItems/GraphicsItem.py", line 444, in parentChanged
    self._updateView()
  File "/home/lesaux/Documents/ENSTA/Python/Tests/056_pyqtgraph-infiniteline-tests/pyqtgraph/graphicsItems/GraphicsItem.py", line 500, in _updateView
    self._replaceView(oldView)
  File "/home/lesaux/Documents/ENSTA/Python/Tests/056_pyqtgraph-infiniteline-tests/pyqtgraph/graphicsItems/GraphicsItem.py", line 515, in _replaceView
    child._updateView()
  File "/home/lesaux/Documents/ENSTA/Python/Tests/056_pyqtgraph-infiniteline-tests/pyqtgraph/graphicsItems/GraphicsItem.py", line 497, in _updateView
    self.viewTransformChanged()
  File "/home/lesaux/Documents/ENSTA/Python/Tests/056_pyqtgraph-infiniteline-tests/pyqtgraph/graphicsItems/InfiniteLine.py", line 420, in viewTransformChanged                                                                                            
    self.updatePosition()                                                                                                    
  File "/home/lesaux/Documents/ENSTA/Python/Tests/056_pyqtgraph-infiniteline-tests/pyqtgraph/graphicsItems/InfiniteLine.py", line 351, in updatePosition                                                                                                  
    paths = vr.intersected(p).toSubpathPolygons()                                                                            
TypeError: QPainterPath.toSubpathPolygons(QTransform): not enough arguments                                                  
Fatal Python error: Aborted                                                                                                  
                                                                                                                             
Current thread 0x00007f0218cb2700 (most recent call first):                                                                  
  File "/home/lesaux/Documents/ENSTA/Python/Tests/056_pyqtgraph-infiniteline-tests/pyqtgraph/graphicsItems/GraphicsItem.py", line 389 in setParentItem
  File "/home/lesaux/Documents/ENSTA/Python/Tests/056_pyqtgraph-infiniteline-tests/pyqtgraph/graphicsItems/ViewBox/ViewBox.py", line 404 in addItem
  File "/home/lesaux/Documents/ENSTA/Python/Tests/056_pyqtgraph-infiniteline-tests/pyqtgraph/graphicsItems/PlotItem/PlotItem.py", line 512 in addItem
  File "plottingItems.py", line 29 in <module>
Abandon (core dumped)
```

lesauxvi:
Just check with @sylm21, the problem occurs both with linux and Win7 (py3 and PyQt5), and does not occurs with py2 and PyQt4. The problem seems related to PyQt4. I'll try to understand why!

campagnola:
Thanks for the check! Latest commit should fix this.

campagnola:
What do you think about the API, @lesauxvi ? There's one feature missing that you had written, which is the ability to switch anchor point when the line crosses a threshold position. This could be added back in, but it didn't seem to be necessary anymore now that the labels default to vertical for vertical lines. Otherwise, I think I like this approach better where the InfiniteLine API and structure are almost completely unchanged, and the new functionality is provided by a new class.

lesauxvi:
Besides the pyqt5 problem, this is awesome. However, I've two modification proposals:
   - the shift argument has disappeared compared to the implementation I proposed. The question is: is this functionnality still relevant regarding the availability of the label dragging? The code modification to add this functionnality is simple (we just have to change the anchor attribute of the textItem should be enough), so it can be interesting to add it. What do you think?
   - I think it could be interesting to add the possibility to rotate the label, i.e. to set an angle of the label different from the infiniteLine one. I do not find the vertical line label very easy to read (the label is rotated), which can be circumventing by authorizing the label to be horizontal (as the blue on the picture below). But may be it is already possible?!?

pyqtgraph_label

As a complementary remark, this new implementation is more effective than mine. The infinite_performance example gives, on my laptop, around 170 fps (with my implementation, I was at 125fps). Congratulations!

lesauxvi:
Do not take into account my remark concerning the rotation of the label. It is already possible. May be you can add a convenience method for handling this easily directly from the InfiniteLine class. Something like  setLabelAngle(self, angle), with angle being the angle of the label relative to the line (in the example, a value of 270 make the vertical label looks like the blue one). Or a method called setHorizontalLabel() which does the job necessary for the label to be horizontal... I do not know what is better. What is your opinion?

campagnola:
> I think it could be interesting to add the possibility to rotate the label

It is already possible by setting `rotateAxis=None` and `angle=...`.

> the shift argument has disappeared compared to the implementation I proposed. The question is: is this functionnality still relevant regarding the availability of the label dragging?

For those like you who prefer to have unrotated labels, I think it does make sense to automatically shift the anchor (especially for vertical lines).

lesauxvi:
Oups, I did not see your remarks. Sorry.

With the last commit, things work perfectly fine with pyqt5 (under Archlinux, will check under Win7 when I'll at work within few minutes)

I really like your code and the API. I share your opinion. You manage to add these functionalities without changing the original InfiniteLine class, which is a very good point.

Just have to discuss the points I mentionned above, and ready for the merge!

Thank you very much for taking some time to improve my implemantation.

lesauxvi:
Ok for the rotation. I missed that point. 

But if the user set `rotateAxis=None and angle=...`, the angle is set relative to the infiniteline angle or relative to the viewBox?

campagnola:
The effects of `rotateAxis` and `angle` args are added independently. So for example:

* rotateAxis=None, angle=0  -> normal horizontal text
* rotateAxis=None, angle=90  -> normal vertical text
* rotateAxis=(1, 0), angle=0  -> text aligned with x axis of its parent
* rotateAxis=(0, 1), angle=0  -> text aligned with y axis of its parent
* rotateAxis=(1, 0), angle=90  -> text orthogonal to x axis of its parent

lesauxvi:
Perfectly clear for me. Thanks for the explanation (the docstrings make perfect sense now).

There remain the shift functionnality. May be you can add it and set the default value such as the behaviour is the current one (shift to 1.)? What do you think?

campagnola:
Done and done.
Thanks @lesauxvi and @allebacco !

lesauxvi:
Thank you @campagnola!

allebacco:
Thank you @campagnola and @lesauxvi  for the new features of InfiniteLine.
When I will have time, I will try to port these features to the InfiniteLine of the VisPy project.

PR #291, opened 2016-02-29
slight changes in TextItem

lesauxvi:
The default text color of any TextItem objects is reinitialized everytime the setText method is called. As a consequence, the color argument of labelOpts in the new InfiniteLine class can not be taken into account. This PR changes this behaviour.

### DIFF
--- a/pyqtgraph/graphicsItems/TextItem.py
+++ b/pyqtgraph/graphicsItems/TextItem.py
@@ -124,7 +124,6 @@ class TextItem(GraphicsObject):
         """
         self.color = fn.mkColor(color)
         self.textItem.setDefaultTextColor(self.color)
-        self.updateTextPos()
         
     def updateTextPos(self):
         # update text position to obey anchor
--- a/pyqtgraph/graphicsItems/TextItem.py
+++ b/pyqtgraph/graphicsItems/TextItem.py
@@ -50,8 +50,8 @@ class TextItem(GraphicsObject):
         self._lastTransform = None
         self._bounds = QtCore.QRectF()
         if html is None:
-            self.color = color
-            self.setText(text, color)
+            self.setColor(color)
+            self.setText(text)
         else:
             self.setHtml(html)
         self.fill = fn.mkBrush(fill)
@@ -64,10 +64,6 @@ class TextItem(GraphicsObject):
         
         This method sets the plain text of the item; see also setHtml().
         """
-        if color != self.color:
-            color = self.color
-        color = fn.mkColor(color)
-        self.textItem.setDefaultTextColor(color)
         self.textItem.setPlainText(text)
         self.updateTextPos()
         
@@ -116,6 +112,16 @@ class TextItem(GraphicsObject):
         
     def setAnchor(self, anchor):
         self.anchor = Point(anchor)
+        self.updateTextPos()
+
+    def setColor(self, color):
+        """
+        Set the color for this text.
+        
+        See QtGui.QGraphicsItem.setDefaultTextColor().
+        """
+        self.color = fn.mkColor(color)
+        self.textItem.setDefaultTextColor(self.color)
         self.updateTextPos()
         
     def updateTextPos(self):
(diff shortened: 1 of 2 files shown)


lesauxvi:
Small improvements of the TextItem class:
   - when a custom color is given, this color is kept even if the setText method is called (for example, every time an InfiniteLine is dragged, cf. #286 )
   - the text and its color are now defined into two separated methods and the color argument no longer need to be define each time the `setText` method is called

The `color=(200, 200, 200)` entry in the `setPos` is kept for API compatibility, but its use is deprecated.

pyqtgraph/graphicsItems/TextItem.py
```diff
@@ -59,12 +60,12 @@ def __init__(self, text='', color=(200,200,200), html=None, anchor=(0,0),
 
     def setText(self, text, color=(200,200,200)):
         """
-        Set the text and color of this item. 
+        Set the text of this item. 
+        
+        The color entry is deprecated and kept to avoid an API change.
```
ericdill: But this *is* an API change because now `setText` silently ignores the `color` kwarg.  Perhaps a better solution would be to make the default color `None` and then 

```python
if color is not None:
    self.setColor(color)
```

Doing it this way will not break existing code that *is* calling `setText` with a color kwarg, but will get the desired behavior of not having to send in the current text color every time you want to change the text.

pyqtgraph/graphicsItems/TextItem.py
```diff
@@ -114,6 +115,16 @@ def setAngle(self, angle):
...
+
+    def setColor(self, color):
+        """
+        Set the color for this text.
+        
+        See QtGui.QGraphicsItem.setDefaultTextColor().
+        """
+        self.color = fn.mkColor(color)
+        self.textItem.setDefaultTextColor(self.color)
+        self.updateTextPos()
```
ericdill: Do you need to call this function if you are only changing the color?

ericdill:
I am generally +1 on this. This change will unfortunately break code that is calling `setText` with the expectation that the font color will be reset to (200,200,200), but I think that using `setText` to reset the color to some default is not an intuitive API anyway, so I am +1 on removing that functionality.  I wonder what @campagnola thinks?

pyqtgraph/graphicsItems/TextItem.py
lesauxvi: Yes!

lesauxvi:
Just changed the PR taking into account @ericdill remarks (by the way, thank you!). Waiting for you @campagnola and hope we are good to go!

pyqtgraph/graphicsItems/TextItem.py
campagnola: Do you need `updateTextPos()` here?

campagnola:
I agree with the API change--it's so unintuitive that few people are likely to be affected by fixing it.
Can you also add a note to the changelog? It should go under the heading `API / behavior changes:`. (see the changes under 0.9.9 for example).

pyqtgraph/graphicsItems/TextItem.py
lesauxvi: No. It was just a mistake. I copy/paste setText method to create the setColor method and did not checkd carefully enough. Thanks!

lesauxvi:
Suggested changes done! Other remarks @ericdill & @campagnola ?

ericdill:
I have no idea why travis just started failing on py2.6... Could you humor me and undo this change? https://github.com/pyqtgraph/pyqtgraph/pull/291/commits/5cd9646fc82c36c27319748057a33dc7b8ed8e4d#diff-8dc53f300f50cb6caef9a3581449aa8fL127 

If that doesn't fix it, then I'll need to dig into travis and see what the heck is going on...

campagnola:
Looks like "interrupted system call" again; probably unrelated to this PR.
I restarted the tests; that should fix it.

ericdill:
Weird. I definitely restarted travis and got the same result. Glad that it was this simple!

lesauxvi:
@ericdill do you still want me to undo the change you mention, or is this PR mergeable?

campagnola:
Looks good!

campagnola:
Thanks @lesauxvi and @ericdill

lesauxvi:
Thanks @ericdill and @campagnola !

PR #293, opened 2016-03-03
Control si prefix

eisoku9618:
I would like to switch enableAutoSIPrefix to False by setLabel argument.

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -233,17 +233,18 @@ class AxisItem(GraphicsWidget):
     def setLabel(self, text=None, units=None, unitPrefix=None, enableAutoSIPrefix=None, **args):
         """Set the text displayed adjacent to the axis.
         
-        ==============  =============================================================
+        ==================  =============================================================
         **Arguments:**
-        text            The text (excluding units) to display on the label for this
-                        axis.
-        units           The units for this axis. Units should generally be given
-                        without any scaling prefix (eg, 'V' instead of 'mV'). The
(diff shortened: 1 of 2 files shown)


eisoku9618:
passed travis.
How about this PR?

ericdill:
Can you add a docstring entry for the new `enableAutoSIPrefix` kwarg in both `setLabel` functions?

ericdill:
other than that, LGTM

eisoku9618:
Thank your for reviewing, I've added the docstring.

eisoku9618:
passed Travis!

campagnola:
Can you give an example of where this will be useful? To disable si-prefixing, I just write the unit string directly into the label text, like:

    axis.setLabel('X axis (units)')

Since the `units` argument is not specified, no si-prefix will be added.

eisoku9618:
Indeed, ’’axis.setLabel('X axis (units)')’’ can disable auto si prefix and I use a similar code in other python graphics packages like matplotlib.

But, pyqtgraph has an interface to set unit **explicitly** while matplotlib does not, and I liked that.
I think spliting label text and unit improves readability, so I prefer to use units argument.

campagnola:
Hm.. I am not convinced. I prefer the API to be as simple as possible, and the `units` argument only exists to allow automatic SI prefixing, not to improve readability.

This issue follows in a long series of feature requests around text formatting, and I am actually considering going the opposite direction--instead of the `units` argument, let the user specify exactly how the label should be formatted like `setLabel("X Axis ({sipfx}Units)")`.

eisoku9618:
OK, I will use ``axis.setLabel('X axis (units)')`` style. Thank you for your reviews!

PR #302, opened 2016-03-27
fix a color name error

WayneTimer:
pen=(R,G,B), so actually the second curve is green and the third curve is blue.


ericdill:
Good catch. Thanks!

PR #304, opened 2016-03-29
OSX RemoteGrahpics.py: Add shared mem file open and close to get around no mremap.

cboulay:
[OS X does not have mremap](http://stackoverflow.com/questions/3521303/is-there-really-no-mremap-in-darwin) and therefore `mmap.mmap(...).resize(...)` does not work. Therefore, none of the RemoteGraphicsView examples worked ([offending line](https://github.com/pyqtgraph/pyqtgraph/blob/develop/pyqtgraph/widgets/RemoteGraphicsView.py#L197)).

I've hacked around the deficiency by simply closing the file and opening a new one of the appropriate size (similar to what is done for Windows). The RemoteGraphicsView examples now work in OS X.

Tested with python 2.7.10 on OS X 10.11.3


ericdill:
I see nothing that gives me pause.  Worth noting that I have little experience with osx dev though.  :+1: from me. @campagnola ?

campagnola:
Looks great, and tests ok on all my platforms. Thanks @cboulay !

PR #305, opened 2016-03-30
MetaArray: PrettyInfo for all axes, not skipping last axis.

cboulay:
It seems deliberate that the final axis in a MetaArray was excluded from the pretty info print treatment but I can't figure out why. This was frustrating to me because I often put 'time' on the last axis as that yields slightly shorter processing time in my signal processing pipeline but then my MetaArrays were messy when printed and I didn't get the best benefits (like sample interval estimation).

This simple change no longer omits the last axis from formatted printing.

### DIFF
--- a/pyqtgraph/metaarray/MetaArray.py
+++ b/pyqtgraph/metaarray/MetaArray.py
@@ -152,7 +152,7 @@ class MetaArray(object):
             if self._data is None:
                 return
             else:
-                self._info = [{} for i in range(self.ndim+1)]
+                self._info = [{} for i in range(self.ndim + 1)]
                 return
         else:
             try:
@@ -175,16 +175,16 @@ class MetaArray(object):
                     elif type(info[i]['values']) is not np.ndarray:
                         raise Exception("Axis values must be specified as list or ndarray")
                     if info[i]['values'].ndim != 1 or info[i]['values'].shape[0] != self.shape[i]:
-                        raise Exception("Values array for axis %d has incorrect shape. (given %s, but should be %s)" %\
+                        raise Exception("Values array for axis %d has incorrect shape. (given %s, but should be %s)" %
                                         (i, str(info[i]['values'].shape), str((self.shape[i],))))
                 if i < self.ndim and 'cols' in info[i]:
                     if not isinstance(info[i]['cols'], list):
                         info[i]['cols'] = list(info[i]['cols'])
                     if len(info[i]['cols']) != self.shape[i]:
-                        raise Exception('Length of column list for axis %d does not match data. (given %d, but should be %d)' %\
+                        raise Exception('Length of column list for axis %d does not match data. (given %d, but should be %d)' %
                                         (i, len(info[i]['cols']), self.shape[i]))
             self._info = info
-   
+
     def implements(self, name=None):
         ## Rather than isinstance(obj, MetaArray) use object.implements('MetaArray')
         if name is None:
@@ -647,7 +647,7 @@ class MetaArray(object):
             if len(axs) > maxl:
                 maxl = len(axs)
         
-        for i in range(min(self.ndim, len(self._info)-1)):
+        for i in range(min(self.ndim, len(self._info) - 1)):
             ax = self._info[i]
             axs = titles[i]
             axs += '%s[%d] :' % (' ' * (maxl - len(axs) + 5 - len(str(self.shape[i]))), self.shape[i])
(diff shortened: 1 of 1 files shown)


ericdill:
> excluded from the pretty info print treatment but I can't figure out why.

Me either!  The original commit that introduced this functionality only says "Added metaarray to support flowchart" 

This PR Looks fine to me.  @campagnola ?

pyqtgraph/metaarray/MetaArray.py
```diff
@@ -631,7 +631,7 @@ def prettyInfo(self):
         s = ''
         titles = []
         maxl = 0
-        for i in range(len(self._info)-1):
+        for i in range(len(self._info)):
```
campagnola: The last index in `_info` is supposed to be an extra dict containing general meta-info for the entire array. So maybe it should look like:

    for i in range(min(self.ndim, len(self._info))):

pyqtgraph/metaarray/MetaArray.py
```diff
@@ -663,7 +663,6 @@ def prettyInfo(self):
                     colstrs.append(cs)
                 axs += '[' + ', '.join(colstrs) + ']'
             s += axs + "\n"
-        s += str(self._info[-1])
```
campagnola: ..and this should be:

    if len(self._info) > self.ndim:
        s += str(self._info[-1])

cboulay:
Is there a need to limit `len(self._info)` to `self.ndim + 1`? I'm making your adjustments now and I realized it's easier to just implement a rule like, "if this is an axis descriptor then give it the treatment, else just print it". However, with this simple rule, this function won't generate errors on maybe undesirable cases. i.e., if i < self.ndim yet it's not properly formatted, or if i > ndim (i.e., more than 1 extra dict at the end). But maybe a print function shouldn't be the place to generate these errors if in fact these cases are undesirable.

What do you think? Simple rule that allows improperly formatted axes to go by OR strict format checking that limits the number of extra dicts to 1?

cboulay:
Also, is there any reason to explicitly refuse having an empty list of `axisdict['values'] = []`?

I sometimes create MetaArrays with shape[time_axis] = 0 as a place-holder before I know how many samples I'm about to receive, but I still need to prepare plots (e.g. frequency labels, data items for each channel). Then, when the plots are prepared and I'm receiving data, I simply use setData to save time.

If that is considered an acceptable use-case, then I'll fix the prettyInfo to check for non-empty axisDict['values'] before trying to print the values start, stop, and step size.

cboulay:
I squashed and rebased after making the changes described in my last two comments. I can go back if this is not desirable. In this PR's current form, a print may look something like:

```
array([], shape=(6, 0), dtype=float64)
-----------------------------------------------
"Channel"   [6] : columns: [RAW1 (V), SPK1 (V), RAW2 (V), SPK2 (V), RAW3 (V), SPK3 (V)]
"Time" (s)  [0] :
"2"             :{'nonimal_srate': 16384.0}
```

pyqtgraph/metaarray/MetaArray.py
```diff
@@ -631,39 +631,45 @@ def prettyInfo(self):
...
         maxl = 0
-        for i in range(len(self._info)-1):
+        for i in range(len(self._info)):
             ax = self._info[i]
             axs = ''
             if 'name' in ax:
                 axs += '"%s"' % str(ax['name'])
             else:
-                axs += "%d" % i
+                axs += '"%d"' % i
```
campagnola: Adding the quotes here implies that the axis has a string name rather than just the integer index.

campagnola:
> What do you think? Simple rule that allows improperly formatted axes to go by OR strict format checking that limits the number of extra dicts to 1?

I looked back through `checkInfo` and it looks like the `_info` list is forced to have exactly `ndim+1` elements, but there is a bug--we need to add `self._info = info` to the end of `checkInfo`. This is the "correct" fix for the prettyInfo bug; I recommend to revert the other changes.

> Also, is there any reason to explicitly refuse having an empty list of axisdict['values'] = []?

Not at all; if the axis has length 0, then an empty sequence should be the only allowed value. This is also enforced in `checkInfo`. So in this case, the values string should just be "[]"; can you fix this?

cboulay:
I made the suggested changes and also spotted a couple other things to change:

1. When None info is given, the empty ._info should be of len ndim + 1 to match what happens when real info is given.
2. Wrapped 2 very long lines.
3. Fixed some spacing issue when the axis length has different number of digits in different axes (i.e., [1000] has 3 more characters than [6] which offsets the output; works up to 5 digits).
4. Fixed a few other style warnings that I noticed Travis picked up on, but not all.

The Travis CI build failed but the errors don't seem to have anything to do with what I changed. I did a `git reset --hard upstream/develop` before making the latest round of changes so the errors must be in the current develop branch.

campagnola:
Looks great, and tests are passing (we have periodic failures with travis that just require resetting the tests). Many thanks, @cboulay!

PR #307, opened 2016-04-10
fix problems with high freequency game mouse

ombschervister:
The number of mouseMoveEvents was limited to ~60 per second.

Issue is analog of https://github.com/vispy/vispy/issues/1146
Pull request is analog of https://github.com/vispy/vispy/pull/1153

### DIFF
--- a/pyqtgraph/GraphicsScene/GraphicsScene.py
+++ b/pyqtgraph/GraphicsScene/GraphicsScene.py
@@ -6,6 +6,8 @@ from .. import functions as fn
 from .. import ptime as ptime
 from .mouseEvents import *
 from .. import debug as debug
+import time
+get_millis = lambda: int(round(time.time() * 1000))
 
 
 if hasattr(QtCore, 'PYQT_VERSION'):
@@ -104,7 +106,9 @@ class GraphicsScene(QtGui.QGraphicsScene):
         self.contextMenu[0].triggered.connect(self.showExportDialog)
         
         self.exportDialog = None
-        
+
+        self._last_move_event_time = 0
+
     def render(self, *args):
         self.prepareForPaint()
         return QtGui.QGraphicsScene.render(self, *args)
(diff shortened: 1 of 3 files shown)


pyqtgraph/GraphicsScene/GraphicsScene.py
```diff
@@ -154,35 +158,40 @@ def mousePressEvent(self, ev):
...
-        ## First allow QGraphicsScene to deliver hoverEnter/Move/ExitEvents
-        QtGui.QGraphicsScene.mouseMoveEvent(self, ev)
-        
-        ## Next deliver our own HoverEvents
-        self.sendHoverEvents(ev)
-        
-        if int(ev.buttons()) != 0:  ## button is pressed; send mouseMoveEvents and mouseDragEvents
+        cur_move_event_time = get_millis()
+        # ignore high frequency events
+        if cur_move_event_time - self._last_move_event_time > 15:
```
ericdill: can you make `15` an attribute of this object?  Seems like this should be a configurable value.  maybe `_milliseconds_between_updates` or something

ericdill:
I think you need to update the test_mouseInteraction test to either:

1. set the min time between updates to 0 in the GraphicsScene object (this is the `_milliseconds_between_updates` property that I suggested in-line) before testing the mouse move events 
2. import time and time.sleep(0.015) between calls to move.

I am not sure if this would fix it, but give it a try?

ericdill:
This is the test at `pyqtgraph/graphicsItems/tests/test_InfiniteLine.py::test_mouseInteraction`

ombschervister:
I update "test_mouseInteraction" test, move event delay move to pyqtgraph options.

ombschervister:
These changes lead to damage to the drag of legendItem.

pyqtgraph/GraphicsScene/GraphicsScene.py
```diff
@@ -6,6 +6,9 @@
 from .. import ptime as ptime
 from .mouseEvents import *
 from .. import debug as debug
+from pyqtgraph import getConfigOption
+import time
+get_millis = lambda: int(round(time.time() * 1000))
```
campagnola: 1. You can probably accomplish this more simply by using `pg.SignalProxy`
2. Pyqtgraph convention is to use `camelCase` (for consistency with Qt API). I often forget too..

pyqtgraph/__init__.py
```diff
@@ -59,6 +59,7 @@
     'exitCleanup': True,    ## Attempt to work around some exit crash bugs in PyQt and PySide
     'enableExperimental': False, ## Enable experimental features (the curious can search for this key in the code)
     'crashWarning': False,  # If True, print warnings about situations that may result in a crash
+    'millisecondsBetweenUpdates': 10,  # For ignoring frequent mouse event.
```
campagnola: I suggest `"mouseRateLimit": 30` or similar. the addition of "mouse" makes the parameter name more descriptive, and specifying the value in events per second should be more intuitive than milliseconds per event.

pyqtgraph/graphicsItems/tests/test_InfiniteLine.py
```diff
@@ -47,49 +47,59 @@ def test_InfiniteLine():
...
-    # test non-interactive line
-    pos = plt.plotItem.vb.mapViewToScene(pg.Point(5,-1)).toPoint()
-    pos2 = pos - QtCore.QPoint(50, 50)
-    mouseMove(plt, pos)
-    assert hline2.mouseHovering == False
-    mouseDrag(plt, pos, pos2, QtCore.Qt.LeftButton)
-    assert hline2.value() == -1
+    finally:
+        # restore delay of mouse move events
+        pg.setConfigOption('millisecondsBetweenUpdates', move_event_delay)
```
campagnola: I don't think resetting `millisecondsBetweenUpdates` is necessary. For unit tests, we expect all mouse moves to be processed immediately.

ombschervister:
Fine, this bug with the legend was not introduced by this pull request. It also happens in master branch as well.
legend_bug

PR #309, opened 2016-04-12
Encode QPropertyAnimation property name if not passed as bytes.

### LINKED ISSUE #257: unexpected 'str' during Arrow test
testing on Windows Python 3.4.4/Qt5.5.1/ PyQtgraph github of 20160102, I have thefollowing  error on "Arrow" test:

````
Using PyQt5 (default graphics system)
Using PyQt5 (default graphics system)
Using PyQt5 (default graphics system)
Using PyQt5 (default graphics system)
  File "D:\WinPython\basedir34\buildQt5\winpython-3.4.4.amd64\python-3.4.4.amd64
\lib\site-packages\pyqtgraph\examples\Arrow.py", line 50, in <module>
    anim = a.makeAnimation(loop=-1)
  File "D:\WinPython\basedir34\buildQt5\winpython-3.4.4.amd64\python-3.4.4.amd64
\lib\site-packages\pyqtgraph\graphicsItems\CurvePoint.py", line 94, in makeAnima
tion
    anim = QtCore.QPropertyAnimation(self, prop)
  QPropertyAnimation(QObject parent=None): too many arguments
  QPropertyAnimation(QObject, QByteArray, QObject parent=None): argument 2 has u
nexpected type 'str'

````

### PR DESCRIPTION
**Pull request by ixjlyons**
Fixes #257 

Note: someone should check compatibility with PySide

As mentioned in the comment I added, this is needed because PyQt stopped automatically converting as of v5.5 [(ref)](http://pyqt.sourceforge.net/Docs/PyQt5/incompatibilities.html). This was *probably* disabled because it's not a smart way to handle things, but I can't think of a better way to do it.

### DIFF
--- a/pyqtgraph/graphicsItems/CurvePoint.py
+++ b/pyqtgraph/graphicsItems/CurvePoint.py
@@ -91,7 +91,9 @@ class CurvePoint(GraphicsObject):
         pass
     
     def makeAnimation(self, prop='position', start=0.0, end=1.0, duration=10000, loop=1):
-        # automatic encoding when QByteString expected was removed in PyQt v5.5
+        # In Python 3, a bytes object needs to be used as a property name in
+        # QPropertyAnimation. PyQt stopped automatically encoding a str when a
+        # QByteArray was expected in v5.5 (see qbytearray.sip).
         if not isinstance(prop, bytes):
             prop = prop.encode('latin-1')
         anim = QtCore.QPropertyAnimation(self, prop)
--- a/pyqtgraph/graphicsItems/CurvePoint.py
+++ b/pyqtgraph/graphicsItems/CurvePoint.py
@@ -91,6 +91,9 @@ class CurvePoint(GraphicsObject):
         pass
     
     def makeAnimation(self, prop='position', start=0.0, end=1.0, duration=10000, loop=1):
+        # automatic encoding when QByteString expected was removed in PyQt v5.5
+        if not isinstance(prop, bytes):
+            prop = prop.encode('latin-1')
         anim = QtCore.QPropertyAnimation(self, prop)
         anim.setDuration(duration)
         anim.setStartValue(start)

### CODE REVIEW

File: pyqtgraph/graphicsItems/CurvePoint.py [hunk 1]
```diff
@@ -91,6 +91,9 @@ def paint(self, *args):
         pass
     
     def makeAnimation(self, prop='position', start=0.0, end=1.0, duration=10000, loop=1):
+        # automatic encoding when QByteString expected was removed in PyQt v5.5
+        if not isinstance(prop, bytes):
+            prop = prop.encode('latin-1')
```
ericdill: why latin-1 instead of utf-8?

ixjlyons: That's what PyQt was doing before they stopped automatically converting a `str` to `QByteArray`.

Looking more into this, I think this is probably not the right way to go, but expecting the user to explicitly pass a `bytes` type as a property name seems a little awkward. Here's a quote from the PyQt 5.4.2 doc strings (in `sip/QtCore/qbytearray.sip`):

```
// In Python v2.x we want a str object to be interpreted as the name of the
// file (as that is the historical behaviour).  This has the following
// implications.
//
// - The QString version of the ctor must appear before the QByteArray version
//   in the .sip file.  This rule should be applied wherever a similar
//   situation arises.
// - A QString must not automatically convert a QByteArray.
// - QByteArray must also exist in the v2 API.
//
// In Python v3.x we want a bytes object to be used wherever a QByteArray is
// expected.  This means that a QString must not automatically convert a bytes
// object.
//
// A QByteArray can be created from a Latin-1 encoded string.  This was a
// mistaken attempt to ease the porting of Python2 code to Python3.  It is
// deprecated in PyQt v5.4 and will be removed in v5.5.
```

ericdill: Might we need to make shim layer here for compatibility for pyside, pyqt4 and pyqt5?  cc @campagnola

campagnola: I don't think a shim is necessary--the pyqt folks probably had a good reason for the API change, and this is the only place in pyqtgraph that we use QAnimation.

campagnola: Can you add a comment here referencing qbytearray.sip and briefly explaining the pyqt API change from 5.4 to 5.5?

campagnola:
This looks good to me, and it tests ok on pyqt4, pyside, and pyqt5. Good to merge after adding the comment mentioned above. Thanks!

ixjlyons:
Something like that? Also, do you prefer squashed commits?

campagnola:
Generally I prefer commits unsquashed, but it's ok to squash for such small changes.
Thanks!

Issue #71, opened 2014-05-23
Crash on exit after exporting a plot


markmelvin:
I have a consistent crash on exit whenever I close a plot _after_ exporting.  My code is as follows:

``` python
def plot(data):
    x_data = [w[0] for w in data]
    y_data = [w[1] for w in data]
    win = pg.GraphicsWindow(title="Input versus Output")

    # Enable antialiasing for prettier plots
    pg.setConfigOptions(antialias=True)

    p1 = win.addPlot(title="Input versus Output")
    p1.setLogMode(False, False)
    p1.plot(x_data, y_data)

# --------------------------------------------------------------------------
if __name__ == "__main__":
    SHOULD_PLOT = True

    if not pyQA400.connect_to_analyzer():
        print "Can't find the analyzer. Exiting."
        sys.exit(1)

    initialize_analyzer()
    data = input_versus_output(-100, -10, 2,
                               generator=pyQA400.GEN2,
                               input_channel=pyQA400.RIGHTIN)
    if SHOULD_PLOT:
        import pyqtgraph as pg
        plot(data)
        pg.QtGui.QApplication.exec_()
    else:
        print data
```

This all runs fine but if I happen to right-click the plot and export it as a .png file, when I close the plot (and the program exits) I get a crash.

I am running Python 2.7, the latest pyqtgraph (from head) and PySide 1.0.9 on Windows 7, 64-bit.

2xB:
I can reproduce the crash with pyqtgraph 0.10.0 but not with the develop version of pyqtgraph, using the following standalone code (adapted from your extract):

```python3
def plot(data):
    x_data = [w[0] for w in data]
    y_data = [w[1] for w in data]
    print(x_data)
    print(y_data)
    win = pg.GraphicsWindow(title="Input versus Output")

    # Enable antialiasing for prettier plots
    pg.setConfigOptions(antialias=True)

    p1 = win.addPlot(title="Input versus Output")
    p1.setLogMode(False, False)
    p1.plot(x_data, y_data)

    return win

# --------------------------------------------------------------------------
if __name__ == "__main__":
    data = [[1,2],[1.1,2.1],[2,3],[2.4,3.4]]
    import pyqtgraph as pg
    win = plot(data)
    pg.QtGui.QApplication.exec_()
```

The error message I get on Xubuntu 16.04, python 3.6.3 (Anaconda), Qt 5.6.4, pyqtgraph 0.10.0:

```python3
[...]

PR #311, opened 2016-04-13
Update BarGraphItem.py

dindom999:
add setData()


2xB:
If I do not miss something, this PR just introduces a function `setData` that is a pure copy of an existing function `setOpts`. Is this a mistake?
Also, this PR does not use a seperate branch, so other commits to dindom999/pyqtgraph branch `develop` also occur in this PR. They fix issue #71, which is now fixed in branch `develop`.

Therefore, I close this Pull Request for now. If I missed something, especially in the new function `setData`, I would be happy to reopen this.

PR #312, opened 2016-04-13
Trying to add support for roiCurves with multi-color image stacks in ImageView

ptweir:
roiCurve is replaced by a list, roiCurves. When an roi is changed in a 4-dimensional image stack, an roiCurve is added for each color.

*I apologize in advance if these changes are undesirable. (This is my first time submitting a pull request on an open source project.)

### DIFF
--- a/examples/ImageViewColorZStack.py
+++ b/examples/ImageViewColorZStack.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+"""
+This example demonstrates the use of ImageView with 3-color image stacks
+ImageView is a high-level widget for displaying and analyzing 2D and 3D data.
+ImageView provides:
+
+  1. A zoomable region (ViewBox) for displaying the image
+  2. A combination histogram and gradient editor (HistogramLUTItem) for
+     controlling the visual appearance of the image
+  3. A timeline for selecting the currently displayed frame (for 3D data only).
+  4. Tools for very basic analysis of image data (see ROI and Norm buttons)
+
+"""
+## Add path to library (just for examples; you do not need this)
+import initExample
+
+import numpy as np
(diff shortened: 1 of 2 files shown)


ericdill:
Thanks for the PR @ptweir !  Could you add an example in https://github.com/pyqtgraph/pyqtgraph/tree/develop/examples that demonstrates this usage? Seeing it in action would help evaluate the PR

ptweir:
Hi @ericdill thanks for the suggestion. Let me know how to improve the example and/or changes.

ixjlyons:
It turns out `ImageView` has sort of independently evolved to support ROI average curves for each color channel, including stacked color images, so this example just works (once #1291 is merged). Making a note here that we should cherry-pick f5c4eee0c6d80f657dfc534fcdc0a635fedd9941 because that's a pretty cool example that really demonstrates what's going on.

j9ac9k:
@ixjlyons just saw your comment (why am I so bad at this); you suggesting either updating this PR (or creating a new PR) with just  https://github.com/pyqtgraph/pyqtgraph/commit/f5c4eee0c6d80f657dfc534fcdc0a635fedd9941  and things will "just work"?

@ptweir sorry this PR has sat idle for so long, we've had quite the PR back-log which I've made a priority to go through.

ixjlyons:
@j9ac9k yeah the example works with the current state of `ImageView` without further modification. My suggestion was to just grab this example script then add it to the example app. I started doing it and waffled on it because I'm not sure it makes sense to have both `examples/ImageView.py` and this one. Maybe they should just be merged.

outofculture:
I took @ixjlyons' thought and ran with it; this has a funner example than the original, so I say we just go with this one. I'm closing this so I can just merge that.

PR #313, opened 2016-04-14
add a CursorItem class

lesauxvi:
This pull request aims at providing a first implementation of a CursorItem, i.e. a cursor that can be dragged to a specific location. For the moment, the current implementation provides the following features : 
- convenience methods for customizing the look of the cursor
- a label attached to the object indicating its current location in the viewbox

A non neglectible part of the code comes from the Handle class (available in the ROI file) because some features where already available in this class. I just extended them. I think it should be possible to rewrite the Handle class taking this new class into account.

The idea of this class is to provide a basic item that can be use for more advanced widgets, such as  a crossHair item for example (already written, but waiting for the feedback on this item before proposing it as a pull request).

Some interesting new features can also be proposed in the future (if this class is judged useful), such as an equivalent to the ginput functionnality of matlab which allows the user to pick the exact value of a PlotDataItem...

Waiting for your feedback!

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -21,6 +21,7 @@ pyqtgraph-0.9.11  [unreleased]
 	- Dock titles can be changed after creation
         - Added Dock.sigClosed
     - Added TextItem.setColor()
+    - Added the CursorItem class
 
   Maintenance:
     - Add examples to unit tests
--- a/examples/InfiniteLine.py
+++ b/examples/InfiniteLine.py
@@ -41,6 +41,7 @@ label = pg.InfLineLabel(lr.lines[1], "region 1", position=0.95, rotateAxis=(1,0)
 cursor = pg.CursorItem(pos=[50, 20])
 p1.addItem(cursor)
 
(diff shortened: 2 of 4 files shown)


lesauxvi:
@campagnola and @ericdill , any chance one of you find the time to review this PR? I knwo it is quite long, but it is very interesting to my opinion.

lesauxvi:
ping @campagnola  and @ericdill

lesauxvi:

lesauxvi:
Besides the way this class is implemented, is this new feature of interest for the library @campagnola  and @ericdill ?

j9ac9k:
Hi @lesauxvi 

Sorry nobody followed up with you on this PR.  When this PR was created, @campagnola in parallel created TargetItem, which lacked in the features CursorItem had. Given TargetItem was already merged into the library, I decided to implement your suggested changes into TargetItem, which I did this with #1318   In addition, I added a number of other features as well.

Thanks for this PR, sorry we didn't follow up with you earlier.

PR #318, opened 2016-04-21
[WIP] Porting to C++ with Python wrapper using SIP

allebacco:
This PR is for introducing the (work in progress) porting of pyqtgraph to C++ and a SIP Python wrapper. Since Python is sometime limiting the performance of the visualization, my purpose is to translate the most part of the library to Qt C++. Actually only a limited set of classes have been ported:

- `Point`
- `GraphicsItem`
- `GraphicsObject`
- `GraphicsWidget`
- `ViewBox` (not 100%)
- `UIGraphicsItem`
- `ItemGroup`
- `ChildGroup`
- `HoverEvent`
- `MouseClickEvent`
- `MouseDragEvent`

There is a lot of work to be done before completing the porting only of the GraphicsView framework. The code is not 100% stable and not all examples are working.

Actually I'm working in my spare time to document the C++ code and then I will continue to porting the rest of the graphics items.

The actual performance are not better than the original Python code, but before investigating the reason, I need to port other computing intensive function and classes.

I hope you find my work interesting, but I think that this is a complicated PR that could break the most part of the library. Maybe it can be considered an initial working on the version 2.0 of the pyqtgraph library.

Thank you.


campagnola:
Woah! This is very interesting and looks like a lot of work. I have several conflicting thoughts:

* If you can significantly improve performance, that would be awesome, but:
* rather than port everything to C++, it makes more sense to me to profile the code and identify specific areas that could benefit from optimization. Even with that approach, though:
* turning pyqtgraph into a C++ library somewhat defeats one of the primary purposes of pyqtgraph, which is to have a library that is written in python. Being pure-python makes the library easy to develop and distribute, and in fact we already have other fast C/C++ plotting options available, which we do not use because of the difficulty of using these with python.

One option is to have an optional "acceleration" package that includes compiled classes/functions that optimize the slowest points of the library. That way most users get the benefit of extra optimization, but we can always fall back to the PP code if the acceleration package is not available. The downside to this approach is the need to maintain 2x code.

allebacco:
Thank you, Luke.

Your points are right, it is not possible to maintain 2 different code for the same library and the strength of the library is to be pure Python.

This porting project has been started to learning the inner part of this plot library and the SIP Python wrapper generator.

An interesting result of my work is that i have found some of what I think can be dead code within some functions. I think that, in the future, after some investigation, there can be some PR to remove that code.

About the "acceleration" package, I think that it can be a good solution, but it should not be integrated within pyqtgraph: my idea is the creation of a new library that contains only the accelerated functions that can be loaded (if available) by pyqtgraph. These functions, if general enough, could also be used by other "PyQt" projects.

campagnola:
An easy place to start would be `pyqtgraph/functions.py`. A lot of the code there uses numpy tricks for speed, but they could likely be optimized in C. In particular, `arrayToQPath`, `rescaleData`, `makeARGB`, `makeQImage`, `downsample`. Another obvious target is in `graphicsItems/ScatterPlotItem.py`: `SymbolAtlas` and `ScatterPlotItem.updateSpots`.

allebacco:
Actually I have not translated any of these function to C++, however I think that in the next future I will take a look at least at `arrayToQtPath` and `invertQtTransform`. But why do you need a custom matrix invert function? `QTransform.inverted` is not enough?

campagnola:
Yeah, from the docstring:

    bugs in that method. (specifically, Qt has floating-point precision issues
    when determining whether a matrix is invertible)

allebacco:
Are there any open issue or specification of this kind of bug? It is in all versions of Qt?

Matrix invert is everywere in the code. Porting the function to C++, without precision loss, could lead to performance improvements.

campagnola:
I usually report bugs as I find them. I'm not sure how easily it could be ported, though--it uses numpy, which probably uses lapack or blas..

allebacco:
This PR is closed because its purpose is out of the aims of `pyqtgraph` library. Interesting changes, if any, will be integrated with smaller PR.
Thank you Luke for the suggestion about the functions that need a speedup.

PR #320, opened 2016-04-25
Addition of the _name attribute to ROI

lesauxvi:
Small PR to add a _name attribute as well as a getter and a setter.

I'm developing a widget that list all the elements that are present in a plotItem and I'm taking advantage of the _name attribute to feed this widget with the items name. I'll propose it as PR when it will be ready... Currently, only PlotDataItem and InfiniteLine objects have this feature.

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -110,7 +110,9 @@ class ROI(GraphicsObject):
     sigClicked = QtCore.Signal(object, object)
     sigRemoveRequested = QtCore.Signal(object)
     
-    def __init__(self, pos, size=Point(1, 1), angle=0.0, invertible=False, maxBounds=None, snapSize=1.0, scaleSnap=False, translateSnap=False, rotateSnap=False, parent=None, pen=None, movable=True, removable=False):
+    def __init__(self, pos, size=Point(1, 1), angle=0.0, invertible=False, maxBounds=None,
+                 snapSize=1.0, scaleSnap=False, translateSnap=False, rotateSnap=False,
+                 parent=None, pen=None, movable=True, removable=False, name=None):
         #QObjectWorkaround.__init__(self)
         GraphicsObject.__init__(self, parent)
         self.setAcceptedMouseButtons(QtCore.Qt.NoButton)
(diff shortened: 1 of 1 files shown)


lesauxvi:
@campagnola and @ericdill , any chance one of you find the time to review this PR?

lesauxvi:
ping @campagnola  and @ericdill

lesauxvi:

lesauxvi:
ping @campagnola and  @ericdill

2xB:
You could keep all your ROIs together with their names in a dictionary. If this does not fit your demands, this sounds like a great use-case for inheritance:
```python3
class NamedROI(ROI):
        self._name = kwargs.pop("name", None)
        ROI.__init__(self, *args, **kwargs)

    def setName(self, name):
        self._name = name

    def getName(self):
        return name
```

I'd suggest to do it that way (as opposed to have a `name` attribute), as naming objects is nearly never done via a built-in attribute and still common enough that there are enough ways around it (as suggested above) that I would suggest not to make this an attribute of `ROI`.

On a more personal side: If a Pull Request is ignored, that mostly happens because the maintainer of a package has so much to do that the PR is left on hold. From a maintainers point of view, both Pull Requests and issues are better when closed or, for Pull Requests, of course merged. Therefore, you are not the only one wanting to see this discussed. So writing reminders to a Pull Request usually only brings the effect of the maintainer having even more mails in his inbox and therefore progressing even slower. After two months or half a year, one can of course gently ask to reconsider a request. But 4 reminders in under two months do not look that sophisticated and may be perceived tiring.

j9ac9k:
I'm not entirely against the PR, as @lesauxvi indicates, we use this naming feature in other places in the library, specifically in `Dock.py`, `Node.py`, `Terminal.py` and `InfiniteLine.py`.

We allow for naming plots via `addPlot` method of `GraphicsLayoutWidget`

We allow naming of plots for the purposes of legend items. 

I'm inclined to reject this PR as I'm not sure where else in the library we would use this parameter.  I'll wait for some other maintainers to chime in.

ksunden:
I also agree with @2xB that this is a fairly niche use case, served well by a small subclass.

Additionally, looking at the implementation, it is wholly unclear to me why this wouldn't just be an attribute `name` (or an `@property`, if we needed to do more than it currently does).

That said, we use `setName` in two places in the codebase now: Parameters (where the setter has a more complex implementation), and `InfiniteLine` (also implemented by @lesauxvi, essentially the same as here)

j9ac9k:
Given no other part of the library benefits from this, and if an end user wants to add this feature, it's trivial to do so, I'm going to close out this issue.

Issue #316, opened 2016-04-19
A GraphicsItem object overlaps the ViewBox border line


espdev:
An example:

``` python
from pyqtgraph.Qt import QtGui
import pyqtgraph as pg
from skimage import data

app = QtGui.QApplication([])

pg.setConfigOptions(antialias=True)

view = pg.GraphicsView()
view.resize(500, 500)

border = pg.mkPen(cosmetic=True, width=5, color='g')

glayout = pg.GraphicsLayout(border=border)
view.setCentralItem(glayout)

vb1 = pg.ViewBox(lockAspect=True, invertY=True)
vb2 = pg.ViewBox(lockAspect=True, invertY=True)

glayout.addItem(vb1, 0, 0)
glayout.addItem(vb2, 0, 1)

imi1 = pg.ImageItem()
imi1.setImage(data.camera().T)
vb1.addItem(imi1)

imi2 = pg.ImageItem()
imi2.setImage(data.camera().T)
vb2.addItem(imi2)

view.show()
app.exec()
```

Screenshot:
<img width="300" alt="pg_overlapping" src="https://cloud.githubusercontent.com/assets/1299189/14648472/24270754-066b-11e6-9ca7-e678853a18da.PNG">

sylm21:
No problem for me with your script with the last pyqtgraph develop version. I'm on windows 7 with pyqt5.5 and python 3.4

espdev:
@sylm21 Thank you for testing. My environment: windows 10, pyqtgraph 0.9.10-142-gfd76443, pyqt 5.5.1 and python 3.4. I use antialiasing and do not use opengl.

Currently I'm using workaround: double layout:
```python
view = pg.GraphicsView(self)

glayout = pg.GraphicsLayout(border=(150, 150, 150))
view.setCentralItem(glayout)

layout = glayout.addLayout(row=0, col=0)
layout.setContentsMargins(1, 1, 1, 1)

viewbox = pg.ViewBox()
layout.addItem(viewbox)
```

campagnola:
Confirmed on linux. This could be fixed easily by removing the border drawing from `ViewBox.paint()` and instead adding a child `QGraphicsRectItem` similar to `ViewBox.background`, with the Z value set high.  @espdev, would you like to make a PR?

espdev:
@campagnola GraphicsLayout has the same problem.

I tried to fix this problem for ViewBox using your method, but there is new problem. Depending on the size of the window some borders disappear.
[...]

PR #321, opened 2016-04-25
Fixing the border rect overlapping for ViewBox and GraphicsLayout

espdev:
This PR fixes issue #316 
Please review this PR.

### DIFF
--- a/pyqtgraph/graphicsItems/GraphicsLayout.py
+++ b/pyqtgraph/graphicsItems/GraphicsLayout.py
@@ -23,6 +23,7 @@ class GraphicsLayout(GraphicsWidget):
         self.setLayout(self.layout)
         self.items = {}  ## item: [(row, col), (row, col), ...]  lists all cells occupied by the item
         self.rows = {}   ## row: {col1: item1, col2: item2, ...}    maps cell location to item
+        self.itemBorders = {}  ## {item1: QtGui.QGraphicsRectItem, ...} border rects
         self.currentRow = 0
         self.currentCol = 0
         self.setSizePolicy(QtGui.QSizePolicy(QtGui.QSizePolicy.Expanding, QtGui.QSizePolicy.Expanding))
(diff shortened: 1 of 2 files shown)


half-adder:
Any news on this?

j9ac9k:
I'm not opposed to merging this (especially as @campagnola was involved with the referenced issue) . This is a tough thing to evaluate due to changes to ViewBox which we consider to be really fragile.  

Also our priorities are right now with getting our test suite to pass.  To move this up our queue some steps that would be good to add are

1. Resolve the conflicts with `ViewBox.py`
1. Re-run against our CI suite (something we can manage after the conflicts are handled).
2. Add a test that fails for the current implementation, but passes  with the fix.

espdev:
I'm sorry. I was working on it 3 years ago and did not get any feedback. Now I don’t remember the implementation details. If someone completes this work, I will be more than happy.

j9ac9k:
@espdev Thanks for the PR, nobody is expecting you to update this from 3 years ago.  This PR does address an issue with the library that we would like to have fixed.  I can see some of the other collaborators (or myself) picking this up and going with it; I was merely relaying to @half-adder where our current priorities were lying and laying out what it would take to get this merged right away.

espdev:
I have resolved conflicts but I cannot guarantee the code works correctly. :)
I will try to check it soon.

j9ac9k:
> I have resolved conflicts but I cannot guarantee the code works correctly. :)
> I will try to check it soon.

Thanks so much for both the original PR and the follow up/update.  If there is a specific use case you're concerned about, but can't get to it, let me know and I'll try and dive into it.

espdev:
I have checked this changes on my example from the issue. It seems to me it works.

`develop` branch (78df356c):


`fix-issue316` branch (1955ae0):


What additional automatic tests can we add?

j9ac9k:
I'm not sure what tests can be added honestly, but this is looking pretty good, I was hoping to merge this last night but we had a miniconda 4.7.x issue in our CI that caused all sorts of havoc.  I'll try and get to this today.

j9ac9k:
This looks good to me, I'm going to merge.  Thanks again for the PR, I'm sorry it took us so long to follow up with you.

PR #327, opened 2016-05-03
Made InfLineLabel.setFormat actually set the format string.

lidstrom83:


lesauxvi:
Nice catch! Thanks @lidstrom83?

Easy PR again @ericdill and @campagnola ;-)

ericdill:
Hah, yes. Silly to set the format string to the built in format function :grinning: Thanks @lidstrom83 and @lesauxvi

Issue #263, opened 2016-01-04
Can't hide/show SpinBox-based parameters from parameter tree


lidstrom83:
Here's a script to demonstrate:

```
import pyqtgraph as pg
from pyqtgraph.parametertree import ParameterTree, Parameter

if __name__ == "__main__":
    import sys
    app = pg.QtGui.QApplication(sys.argv)
    tree = ParameterTree()
    param = Parameter.create(name='param', type='float')
    tree.setParameters(param)
    tree.show()
    param.show(False)
    sys.exit(app.exec_())
```

Traceback:

```
Traceback (most recent call last):
  File "C:\Users\dlidstrom.RTC\AppData\Local\Continuum\Miniconda3\envs\dev\lib\site-packages\pyqtgraph\parametertree\parameterTypes.py", line 285, in optsChanged
    self.widget.setOpts(**opts)
  File "C:\Users\dlidstrom.RTC\AppData\Local\Continuum\Miniconda3\envs\dev\lib\site-packages\pyqtgraph\widgets\SpinBox.py", line 160, in setOpts
    raise TypeError("Invalid keyword argument '%s'." % k)
TypeError: Invalid keyword argument 'visible'.
```

PR #328, opened 2016-05-03
Fix Parameter.show for float, int, and group types

lidstrom83:
Fixes issue #263.  Here's a script for demonstrating the bugs/fixes:

```
from pyqtgraph.parametertree import ParameterTree, Parameter

app = QtGui.QApplication([])
tree = ParameterTree()
param = Parameter.create(name='param', type='group', children=[
    {'name': 'int', 'type': 'int'},
    {'name': 'float', 'type': 'float'},
    {'name': 'group', 'type': 'group'}
])
tree.setParameters(param)
tree.show()
for child in param.children():
    child.show(False)

app.exec_()
```

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes.py
+++ b/pyqtgraph/parametertree/parameterTypes.py
@@ -414,6 +414,8 @@ class GroupParameterItem(ParameterItem):
             ParameterItem.addChild(self, child)
             
     def optsChanged(self, param, changed):
+        ParameterItem.optsChanged(self, param, changed)
+
         if 'addList' in changed:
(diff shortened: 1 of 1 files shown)


lidstrom83:
Fixes issue #346.

2xB:
Thank you for the pull request! In `develop`, this is now fixed via commit https://github.com/pyqtgraph/pyqtgraph/commit/65e9052580fe93130738549dc374621efdc126fa#diff-66f0335c2c55dfc152aa64b292f07db6, which is very similar to the suggested changes here!

PR #330, opened 2016-05-03
Set parameter's default button autoDefault value to False

lidstrom83:
When a parameter tree is included in a dialog, currently when the user presses Enter, the first parameter's default button is clicked.  This seems unexpected.

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes.py
+++ b/pyqtgraph/parametertree/parameterTypes.py
@@ -48,6 +48,7 @@ class WidgetParameterItem(ParameterItem):
             w.setToolTip(opts['tip'])
         
         self.defaultBtn = QtGui.QPushButton()
+        self.defaultBtn.setAutoDefault(False)
         self.defaultBtn.setFixedWidth(20)
         self.defaultBtn.setFixedHeight(20)
         modDir = os.path.dirname(__file__)


lesauxvi:
I'm not very familiar with these functionalities of pyqtgraph, but it seems ok to me. What do you think @campagnola  and @ericdill?

ixjlyons:
Ah, don't think the close/reopen was necessary after the base branch change.

Anyway, looks good to me. Quick example:

```python
import pyqtgraph as pg
from pyqtgraph.parametertree import ParameterTree, Parameter

app = pg.mkQApp()

win = QtGui.QDialog()
layout = QtGui.QGridLayout()
win.setLayout(layout)

tree = ParameterTree()
param = Parameter.create(name='root', type='group', children=[
    {'name': 'a', 'type': 'int', 'value': 10, 'default': 0},
    {'name': 'b', 'type': 'float', 'value': 10.5, 'step': 0.1},
])
tree.setParameters(param)

layout.addWidget(tree)
win.show()

app.exec_()
```

I agree it seems unexpected even if there's an indication of keyboard focus on the button. It also doesn't match the behavior of a `ParameterTree` in a regular window or just showing the tree itself.

Thanks @lidstrom83, better late than never!

PR #331, opened 2016-05-04
Allow Parameter to have no default value.

lidstrom83:
Currently the default is always set to None, while it's clear from the rest of the code that a missing 'default' key in Parameter.opts is meant to represent the case of a missing default.

Edit: To be more clear, the default is set to None if neither 'value' nor 'default' are explicitly set.


lidstrom83:
When no default is set, the current behavior is:
- Parameter.defaultValue raises a KeyError
- Parameter.valueIsDefault returns None, instead of True or False
- The button is hidden

Does everyone else agree with these choices?

Issue #332, opened 2016-05-04
Bug with LegendItem drag


ombschervister:
In master branch In examples/legend.py after dragging the legend around and releasing the mouse, next dragging attempt drags the plot instead of the legend.

![legend_bug](https://cloud.githubusercontent.com/assets/3264643/15014859/c577e0e4-1221-11e6-996f-3c98a1be454c.gif)

lesauxvi:
I'm not able to reproduce the bug with the develop version of pyqtgraph (version of april 2016) both under linux and win7.

Can you also check on your side?

ombschervister:
develop branch, last commit "670d63c Eric Dill on 03.05.2016 at 22:11":
error is reproduced.

The bug is appear, when this code running:
GraphicsScene.mousePressEvent

       if self.lastHoverEvent is not None:
                # If the mouse has moved since the last hover event, send a new one.
                # This can happen if a context menu is open while the mouse is moving.
                if ev.scenePos() != self.lastHoverEvent.scenePos():
                    self.sendHoverEvents(ev)

For reproduce the bug you should not move the mouse after first drag and before next.
Maybe you need to add print before sendHoverEvent.

lesauxvi:
OK, with your guidelines, I'm able to reproduce it! My above comment should not be taken into account...

j9ac9k:
@ombschervister Thanks for the issue with the gif and the associated PR!

PR #333, opened 2016-05-04
fix legendItem drag

ombschervister:
Fix https://github.com/pyqtgraph/pyqtgraph/issues/332

### DIFF
--- a/pyqtgraph/graphicsItems/LegendItem.py
+++ b/pyqtgraph/graphicsItems/LegendItem.py
@@ -128,6 +128,7 @@ class LegendItem(GraphicsWidget, GraphicsWidgetAnchor):
         
     def mouseDragEvent(self, ev):
         if ev.button() == QtCore.Qt.LeftButton:
+            ev.accept()
             dpos = ev.pos() - ev.lastPos()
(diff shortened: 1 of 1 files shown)


ombschervister:
Can you explain what does it mean "Merging #333 into develop will decrease coverage by -<.01%"?

ombschervister:
1. Ev.accept() in mouseDragEvent fix problem, but i do not understand what behavior must be for ev.ignore. I think after ev.ignore mouseDrag must stop calling, but now it is not.

2. If I delete lines about hoverEvent from mousePressEvent, problem fix too.
       Deleted lines:
           if self.lastHoverEvent is not None:
               # If the mouse has moved since the last hover event, send a new one.
               # This can happen if a context menu is open while the mouse is moving.
               if ev.scenePos() != self.lastHoverEvent.scenePos():
                   self.sendHoverEvents(ev)

lesauxvi:
look at my remark at #332 ...

j9ac9k:
Thanks for the PR, sorry it took us so long to follow up.   It looks good to me, will be merging!.

PR #335, opened 2016-05-06
fix work of hasattr(PlotWidget)

ombschervister:
Change NameError to AttributeError in PlotWidgte.**getattr** to fix hasattr(PlotWidget) calling.


j9ac9k:
Looks like this got fixed in 52754d48594f713092a7e6b55a6fdcb046f76be6

Thanks for the PR regardless!

PR #340, opened 2016-05-17
Fix error in siEval for exponential notation with plus sign

josselex:
if number has  a plus sign in exponent, the conversion failed and e.g.
2.0e+03 returned 2.0 instead of 2000 in a SpinBox

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -115,7 +115,7 @@ def siEval(s):
     """
     
     s = asUnicode(s)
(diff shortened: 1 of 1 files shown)


josselex:
This happened on windows.
* If you input 100000 in a Spin box it got converted into 1e+06. 
* And when you then changed this notation it returned wrong numbers.

lesauxvi:
Seems ok for me even though I'm not very familiar with regular expressions.

What do you think @campagnola  and @ericdill ?

PR #343, opened 2016-05-20
PlotItem.addItem checks item's existence

xiaoqiangwang:
It should avoid adding an item twice.


lesauxvi:
That seems ok for me. What do you think @campagnola  and @ericdill ?

lesauxvi:

2xB:
Looks good to me. As in any case I can imagine adding a PlotItem twice is not intended, I'd suggest to also raise a RuntimeWarning stating that an item was added twice to help debugging.

j9ac9k:
I attempted to make the changes on this PR but couldn't (got a permission denied error).  

@xiaoqiangwang thanks for the PR, sorry it's taken us so long to get to this.  I'm implementing this change along w/ the suggested change from @2xB in #1239

PR #348, opened 2016-05-30
Update MatplotlibWidget.py

mhogg:
NavigationToolbar2QTAgg doesn't exist in matplotlib >= 1.4.0 anymore, and has been replaced with import NavigationToolbar2QT. This is detailed [here](http://matplotlib.org/api/api_changes.html). 
In this pull request, it is proposed that if there is an error importing NavigationToolbar2QTAgg, then NavigationToolbar2QT is imported instead. This is both backwards and forwards compatible.


lesauxvi:
Wouldn't it be cleaner to rather check the version of matplotlib and to import it differently depending on the version? Moreover, It would be easier to remember the necessity to this piece of code.

Agree @campagnola  and @ericdill with this change?

j9ac9k:
Thanks for the PR!

Looks like this was fixed with fcdc2a74adb5c52edfc045d39c4753b2072bcbda

PR #349, opened 2016-05-31
Roi tests

campagnola:
Adds unit tests and several bugfixes for ROI and subclasses.

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -215,15 +215,20 @@ class ROI(GraphicsObject):
         
     def setPos(self, pos, y=None, update=True, finish=True):
         """Set the position of the ROI (in the parent's coordinate system).
-        By default, this will cause both sigRegionChanged and sigRegionChangeFinished to be emitted.
         
-        If finish is False, then sigRegionChangeFinished will not be emitted. You can then use 
-        stateChangeFinished() to cause the signal to be emitted after a series of state changes.
+        Accepts either separate (x, y) arguments or a single :class:`Point` or
+        ``QPointF`` argument. 
         
-        If update is False, the state change will be remembered but not processed and no signals 
+        By default, this method causes both ``sigRegionChanged`` and
+        ``sigRegionChangeFinished`` to be emitted. If *finish* is False, then
+        ``sigRegionChangeFinished`` will not be emitted. You can then use 
+        stateChangeFinished() to cause the signal to be emitted after a series
+        of state changes.
+        
+        If *update* is False, the state change will be remembered but not processed and no signals 
         will be emitted. You can then use stateChanged() to complete the state change. This allows
         multiple change functions to be called sequentially while minimizing processing overhead
-        and repeated signals. Setting update=False also forces finish=False.
+        and repeated signals. Setting ``update=False`` also forces ``finish=False``.
         """
         if y is None:
             pos = Point(pos)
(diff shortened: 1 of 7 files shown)


lesauxvi:
Nice and quite long PR! Before looking at it in details, I've a question: do your modifications change the default behaviour of the getArrayRegion method (interpolation of the data at specific location rather than giving the real value displayed, i.e. the values of the pixels contained within the ROI)? For my personal purpose, I had to subclass ROI because of that and to my opinion, it could be interesting to let the user choose what outputs he wants (interpolated data or real pixels values). But my question/remark is may be out of the scope of this PR...

campagnola:
This does not change the default behavior, but I think you should find this is already possible without subclassing--you can pass `order=0` through to the interpolation functions to get nearest-neighbor, and you can force the ROI's position / size to snap to pixel boundaries.

pyqtgraph/graphicsItems/ROI.py
```diff
@@ -213,7 +213,7 @@ def angle(self):
         """Return the angle of the ROI in degrees."""
         return self.getState()['angle']
         
-    def setPos(self, pos, update=True, finish=True):
+    def setPos(self, pos, y=None, update=True, finish=True):
```
lesauxvi: What is the purpose of the y kwarg? Is there an interest to add some additional docstrings?

campagnola: `QGraphicsItem.setPos()` accepts separate x, y arguments. For some reason `ROI.setPos()` did not follow the same pattern and instead only accepted a single argument. Adding `y` here simply recovers the original behavior.

lesauxvi: OK, make perfect sense now. May be it could be interesting to add your comment in the docstrings of the method as it will help people understand if (if the need to look deeply at the source code)

campagnola:
For some reason, QGraphicsLayout is giving slightly different positions on travis than it is on my machine. This prevents the test images matching, so for now I might just disable the tests on travis. I'd like to get this merged so we can move toward the next release (two more PRs to follow).

lesauxvi:
Seems reasonable!

Do you think it is possible to integrate #320 and #271 into this PR? These are pretty straithforward PR (and I currently use them with a modified Roi class).

I really like these changes. Ok for me. Do you agree @ericdill ?

raphaelvalentin:
Dear pyqtgraph developers,
I experience some issues concerning ROIs by changing mode scales of axis (lin<->log<->fft).
First, ROIs do not scale in the plot correctly and can lead potentially to overflow issues if app developers do not overwrite ROI class.
Second, there is not general way to fix ROIs for different axis mode for all cases because handles coordinate can be negative or equal to zero. Depending of cases, app developers can choose to "delete", "hide" or "modifiy" the ROI. I believe that the best behaviour could be to hide ROIs when scale mode change (ROIs is defined in a axis mode scale and data environment, if one change one of this item, then ROIs can be mostly probably meaningless).
For general purposes, a very quick fix, without cost, could be to include a signal like "aboutToChangeAxisMode" (or whatever). This signal can possibly be added in 'PlotItem.py'. In that way, app developers have opportunities to modify data before plot update (or do whatever). It could be also a way for app developers to fix plot data when those are (partially) negative (it can avoid that numpy sends a Float Warning or Raise an error).
In addition, I notice that the functions "setLogMode" and "setFftMode" are missing in ROI.py. These functions could may do nothing. However in this case, they will be documented in 'help' and can be overwritten by app developers.
What are your opinions? 
Thank a lot for your feedback.

PR #357, opened 2016-07-09
Python 3 fixes

QuLogic:
This fixes various issues on Python 3:
- Broken multiprocessing/remote examples due to `iteritems` calls.
- Drop `sortList` compat function; key-based sorting is available on all supported Python versions, and is simpler.
- Use a context manager for file I/O in as many places as possible; could cause a `ResourceWarning` on Python 3, though it's not enabled by default.
- Fix some encoding issues with `MetaArray`. I'm not sure why the tests are hidden away in the file and not run on Travis, but they would have pointed out this problem.


ixjlyons:
Closing as #1073 was merged

PR #360, opened 2016-07-20
Allow MetaArray.__array__ to accept an optional dtype arg

### LINKED ISSUE #359: `examples/Flowchart.py` fails to produce output
The output plot is blank. The GaussianFilter node in the graph has the following traceback in it:

``` python
File "/usr/local/lib/python2.7/site-packages/pyqtgraph/flowchart/Node.py", line 304, in update
    out = self.process(**strDict(vals))
File "/usr/local/lib/python2.7/site-packages/pyqtgraph/flowchart/library/common.py", line 108, in process
    out = self.processData(In)
File "/usr/local/lib/python2.7/site-packages/pyqtgraph/flowchart/library/common.py", line 182, in newFn
    return fn(self, data, *args, **kargs)
File "/usr/local/lib/python2.7/site-packages/pyqtgraph/flowchart/library/Filters.py", line 167, in processData
    return pgfn.gaussianFilter(data, self.ctrls['sigma'].value())
File "/usr/local/lib/python2.7/site-packages/pyqtgraph/functions.py", line 1195, in gaussianFilter
    filtered = scale * np.fft.irfft(np.fft.rfft(filtered, shape, axis=ax) *
File "/usr/local/lib/python2.7/site-packages/numpy/fft/fftpack.py", line 366, in rfft
    a = array(a, copy=True, dtype=float)
```

Some digging reveals that `MetaArray.__array__` doesn't support numpy's `__array__` protocol fully: it doesn't support an optional `dtype`. I can patch around it by adding unused `*args` and `**kwargs` to it and just ignore them, but I'm not sure this is the correct way to go.

nfirvine:
Fixes #359

### DIFF
--- a/pyqtgraph/metaarray/MetaArray.py
+++ b/pyqtgraph/metaarray/MetaArray.py
@@ -358,7 +358,7 @@ class MetaArray(object):
         else:
             return np.array(self._data)
             
-    def __array__(self):
+    def __array__(self, dtype=None):
         ## supports np.array(metaarray_instance) 
         return self.asarray()
             


j9ac9k:
Merging this due to infact the `__array__` subclass does specify passing dtype along with it here:

https://docs.scipy.org/doc/numpy/reference/arrays.classes.html#numpy.class.__array__

Thanks for the PR @nfirvine sorry it's taken us so long to get to this!

pyqtgraph/metaarray/MetaArray.py
```diff
@@ -358,7 +358,7 @@ def asarray(self):
         else:
             return np.array(self._data)
             
-    def __array__(self):
+    def __array__(self, dtype=None):
```
campagnola: Shouldn't this do something like:
```
if dtype is not None:
    return self.asarray().astype(dtype)
else:
    return self.asarray()
```
 ?

j9ac9k: ....ahh you're right; I'll revert the merge and make that change

PR #361, opened 2016-07-25
Use pyqt5 canvas templates when needed.

sscherfke:


sscherfke:
Was by now fixed in f612d845fcf08601153d1e864240945f1c69a9e1

PR #370, opened 2016-08-28
Image alignment

### LINKED ISSUE #315: ImageItem shows an image as transposed
A simple example (uses image data from scikit-image):
```python
import pyqtgraph as pg
from skimage import data

app = QtGui.QApplication([])

win = QtGui.QMainWindow()
win.resize(800, 800)
imv = pg.ImageView()
win.setCentralWidget(imv)
win.show()

imv.setImage(data.camera())

app.exec_()
```

Result:
pg

campagnola:
Adds a global configuration option allowing image data to be interpreted in row-major format (by default, we assume col-major, which requires most users to transpose image data before displaying).

fixes #315


lesauxvi:
Huge amount of work for a feature that will please many users working with images (I'm among them!). I did some tests and everything works perfectly fine. Looking forward to seeing this feature as a built-in one. 

Luke, how far from the next stable version in terms of new features? Are there other things to implement?

campagnola:
I think this may be the last item on the list before release. I found a couple of minor issues with this PR that I'll try to fix tomorrow..

campagnola:
Tests pass; I'm calling this one good.

PR #376, opened 2016-09-24
modifications to support pyside2

maxpeng:
I followed this guide - [Compiling PySide2 from source · Fredrik Averpil](https://fredrikaverpil.github.io/2016/08/17/compiling-pyside2/) - to build pyside2 on Windows then tested the modifications, most of the examples can be executed.


asaezper:
Hi, I'm trying to use pyqtgraph with PySide2, so I got pyside2 from 5.6 branch and merged these changes in pyqtgraph. However, It doesn't work propertly when I try to plot data (using PlotCurveItem and ScatterPlot). It looks like some paint methods are not called, so curves aren't drawn. 
Do you know if current dev branch from PySide2 works? When do you freeze the PySide code?
Thanks.

klotzt-draeger:
how are chances to see this on master?

j9ac9k:
@klotzt-draeger `master` hasn't been updated in a while, but there is plan to rework how we do our `develop` and `master` relationships.  

Is there a particular reason you're interested in this change being on `master`?  `pip` can install from non-master branches without an issue (see the readme for the code to actually do so).

klotzt-draeger:
the reason why we want it is that we try to keep our requirements installation straight (everythign from pypi), it seems that only master versions are listed here: https://pypi.org/search/?q=pyqtgraph 
I would like to avoid to have extra ways of installing for every single packet in my list...

j9ac9k:
if you're using a `requirements.txt` you can still reference the installing pyqtgraph from a specific commit by including the hash of the commit as such:

```
...
numpy==1.16.4
git+https://github.com/pyqtgraph/pyqtgraph.git@ad7453c7818eddb3b18354c65370f416cb668b29#egg=pyqtgraph
scipy==1.3.0
...
```

FWIW I'm fighting with the same situation in my projects; but I'm not going to rush a release of pyqtgraph just to make my `setup.py` easier to handle; the state of the repository has to get there first, and we're getting close, but aren't there yet.

PR #381, opened 2016-10-16
Fix bounds when ROI snaps to pixel edge.

npisenti:
When ROIs snap to pixel boundaries, the bounds calculation in getArraySlice includes an extra row/column. If we take the ceiling of intBounds rather than truncate with int and then add 1, getArraySlice returns the correct bounds.

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -1047,8 +1047,8 @@ class ROI(GraphicsObject):
         
         ## Determine index values to use when referencing the array. 
         bounds = (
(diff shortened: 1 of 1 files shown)


npisenti:
A little more experimenting -- seems that sometimes `intBounds.top()` (/`left`/`right`/...) will return, for example, `2.000000001` instead of `2.0` which does not work with the above patch.

Is there a reason not to use `np.round(...)`?

j9ac9k:
Thanks for the PR @npisenti I'm going to close this out as you said, the proposed patch doesn't quite work.  We can discuss potential implementations in the issue tracker if this is still a problem.  Sorry it's taken so long to get back to you on this.

PR #383, opened 2016-10-28
Release 0.10.0

campagnola:
All updates in preparation for 0.10.0 release.

Major changes are to setup.py and support tools, especially the addition of a script for automating release package builds and publishing. 

There are also several fixes for broken examples and documentation thrown in.


.travis.yml
```diff
@@ -56,9 +56,12 @@ install:
     - echo ${TEST}
     - echo ${PYTHON}
 
-    - if [ "${QT}" == "pyqt" ]; then
+    - if [ "${QT}" == "pyqt5" ]; then
           conda install pyqt --yes;
       fi;
+    - if [ "${QT}" == "pyqt4" ]; then
+          apt-get install python-qt4 python-qt4-gl;
```
ericdill: You should be able to `conda install pyqt<5` which will pull ` pyqt 4.11.4-py35_4`

campagnola: OOOOOOO. I searched far and wide for "pyqt4" packages. Will try this..

ericdill: You can see the available installs here: https://anaconda.org/anaconda/pyqt/files.  You'll probably need to skip the py26 + pyqt5 combo as there does not appear to be a py26 package on the defaults channel for that

campagnola:
@ericdill  Are you seeing these empty logs on the travis build page? I assume that's not just on my end..
(also, HI!)

.travis.yml
ericdill: You can get up to pyqt-4.11.3 for support with py26, py27, py33, py34 and py35

ericdill:
@campagnola yup, empty on my end too. Weird.  @parente says he's seen that once and it was a screw-up on their end.  (Hi back!)

pyqtgraph/debug.py
```diff
@@ -126,7 +126,30 @@ def printTrace(msg='', indent=4, prefix='|'):
 
 def backtrace(skip=0):
     return ''.join(traceback.format_stack()[:-(skip+1)])    
+
+
+def formatException(exctype, value, tb, skip=0):
```
ericdill: Can you detail what the purpose of `skip` is? The first three arguments are the positionals to traceback.format_exception and traceback.TracebackException IIUC

pyqtgraph/tests/image_testing.py
```diff
@@ -212,7 +212,7 @@ def assertImageApproved(image, standardFile, message=None, **kwargs):
 
 
 def assertImageMatch(im1, im2, minCorr=None, pxThreshold=50.,
-                       pxCount=0, maxPxDiff=None, avgPxDiff=None,
+                       pxCount=-1, maxPxDiff=None, avgPxDiff=None,
```
ericdill: IIUC `-1` is being used here as a sentinel value.  Why not just let it be `None` like the other defaults?

pyqtgraph/widgets/RawImageWidget.py
```diff
@@ -3,7 +3,9 @@
     from ..Qt import QtOpenGL
     from OpenGL.GL import *
     HAVE_OPENGL = True
-except ImportError:
+except Exception:
```
ericdill: I'm 99% sure this is py26/27/34/35 compatible:

```python
except (ImportError, AttributeError):
    HAVE_OPENGL = False
```

pyqtgraph/tests/image_testing.py
campagnola: `None` would mean that pxCount is ignored.

pyqtgraph/debug.py
campagnola: `printException()` below uses `skip=1` to avoid having that function be the last entry in the printed stack trace.

pyqtgraph/widgets/RawImageWidget.py
campagnola: I cast a broad net here because I don't really care what the error is--if either of these packages cannot be imported, then the behavior should always be to handle it gracefully. Don't want to find that the next pyopengl update raises some other unexpected exception.

stonebig:
I had a python failure when leaving "DataTreeWidget" this morning with this patch

campagnola:
Here we go!

campagnola:
@stonebig, was that an exit crash? Let me know if you can produce a minimal working example. I have a deep disdain for exit crashes ;)

stonebig:
a python interpreter has stopped working (a microsoft windows message), while the PyQtGraph general test window remains operative. most probably a Qt bug. I just mentioned it in case you had an obvious fix

PR #385, opened 2016-11-06
Implement headWidth parameter for arrows

Griffon26:
Although the documentation used to say that specifying tipAngle would
override headWidth, headWidth was never used. The new behaviour is that
tipAngle will be used, with a default value of 25, unless headWidth is
specified.

I also adapted the arrow example to give the 3rd arrow a wider head to clearly 
show that the default tipAngle of 25 was being overridden.


Griffon26:
Am I expected to do something about the above checks before this pull request can be integrated? AFAIK there are no existing tests to extend.

Griffon26:
It would be nice to get some feedback on this.

j9ac9k:
Hi @Griffon26 sorry it has taken so long for us to get to this; the PR looks good, and as soon as the CI run finishes (assuming it's all green), I'll merge into the library. 

Thanks for the PR

PR #386, opened 2016-11-07
Suppress VisibleDeprecationWarning when using numpy 1.11.1

radjkarl:
np.ceil returns a float
however, slicing with a float step results in the following warning:

VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future


j9ac9k:
Thanks for the PR, sorry for the long time it's taken for us to follow up.  This issue was fixed with 0c1cda49730bdb7da4f9600dd69f55e8fd0cd4d1

PR #389, opened 2016-11-10
Revert "ignore wheel events in GraphicsView if mouse disabled"

ales-erjavec:
This reverts commit f49c179275e86786af70b38b8c5085e38d4e6cce.

On Qt 5.7 ignoring the initial `wheelEvent` when `phase() ==
Qt.ScrollBegin` suppresses all intermediate events (`Qt.ScrollUpdate`)
from being delivered to the view. This makes ViewBox zooming
unresponsive.

### DIFF
--- a/pyqtgraph/widgets/GraphicsView.py
+++ b/pyqtgraph/widgets/GraphicsView.py
@@ -325,7 +325,6 @@ class GraphicsView(QtGui.QGraphicsView):
     def wheelEvent(self, ev):
         QtGui.QGraphicsView.wheelEvent(self, ev)
         if not self.mouseEnabled:
-            ev.ignore()
             return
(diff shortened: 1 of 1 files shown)


ales-erjavec:
This behavior was introduced as a fix for [QTBUG-50199](https://bugreports.qt.io/browse/QTBUG-50199)

ozankececioglu:
This PR broke our `PlotItem` behavior :( 
Parent (`QScrollArea`) of the plots does not scroll anymore. We had to patch `GrapicsView.wheelEvent` back to the previous version where event is ignored. Just saying.

ixjlyons:
@ozankececioglu you might want to create a new issue linking to this PR so this doesn't get lost

PR #390, opened 2016-11-10
moved some functionality from method 'export' to new method

radjkarl:
'getSupportedFormats' making it acessible from outside


j9ac9k:
Thanks for the PR @radjkarl sorry it's taken us so long to get to this.

PR #391, opened 2016-11-10
changed structure to redefine axis via plotitem.setAxes

radjkarl:
this allows to redefine the plotItems axes after creation.


pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -289,7 +273,31 @@ def __init__(self, parent=None, name=None, labels=None, title=None, viewBox=None
         if len(kargs) > 0:
             self.plot(**kargs)
         
-        
+    def setAxes(self, axisItems):
```
lesauxvi: if `self.axes` is actually non empty when this method is called, should not they be removed from the layout before adding the new axes?

radjkarl: you are right. changed that. see last commit.

lesauxvi:
OK, good to me. Do you agree @campagnola and @ericdill ?

pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -96,6 +93,27 @@ def __init__(self, orientation, pen=None, linkView=None, parent=None, maxTickLen
...
+        """
+        if orientation != self.orientation:
+            if orientation not in ['left', 'right', 'top', 'bottom']:
+                raise Exception("Orientation argument must be one of 'left', 'right', 'top', or 'bottom'.")
+            #rotate absolute allows to change orientation multiple times:
+            if orientation in ['left', 'right']:
+                self.label.setRotation(-90)
+                if self.orientation:
+                    self._updateWidth()
+                    self.setMaximumHeight(16777215)
```
radjkarl: this restores origin value

radjkarl:
setAxes should work now as expected 

to try out this PR, please modify 'PlotWidget' example as follows:

ORIGINAL:
image

WITH ADDED CODE BEFORE LINE 'def rand(n):'
#swap axes
v = pw.plotItem
a1 = pw.getAxis('bottom')
a2 = pw.getAxis('left')
v.setAxes({'bottom': a2, 'left': a1})
image


WITH ADDED COSE BEFORE LINE 'def rand(n):'
#same with now axes at right:
v = pw.plotItem
a1 = pw.getAxis('bottom')
a2 = pw.getAxis('left')
v.setAxes({'bottom': a2, 'right': a1})
image

j9ac9k:
Thanks for the PR @radjkarl LTGM, merging, sorry it's taken so long for us to get to this.

j9ac9k:
Going to revert this PR due to some incompatibilities with PRs merged after this PR was created.

PR #394, opened 2016-11-10
NEW features for HistgramLUTItem

radjkarl:
* option to place gradient left or right
* link histograms to others
* speed up through only paint histogram if visible


outofculture:
First, sorry this has languished for so long. I looked at trying to merge it, and it should work with some slight modifications. The `isVisible()` speedup had conflicts, but was easy to recreate. The `gradientPosition` init arg can be used for the most part. The `linkHistogram` method needs to be tweaked to handle rgba-mode levels. We also need to include the new behaviors in an example or test. I'm going to close this PR and open a new one with these fixes. Thanks @radjkarl!

outofculture:
Ugh, nevermind. I'm going to cut out the `linkHistogram` method, actually. It needs to be reconsidered more completely with respect to rgba-mode levels.

PR #395, opened 2016-11-10
NEW options for LegendItem

radjkarl:
* align items in columns
* show/hide frame
* access labelItem from outside (e.g. to modify label text)


radjkarl:
normal:
image
no frame:
image
3 columns:
image

pyqtgraph/graphicsItems/LegendItem.py
```diff
@@ -74,13 +78,52 @@ def addItem(self, item, name):
...
+            self.updateSize()
     
+    def getLabel(self, plotItem):
+        """
+        return the labelItem inside the legend for a given plotItem
+        the label-text can be changed via labenItem.setText
+        """
+        for i in self.items:
+            if i[0].item == plotItem:
+                return i[1]
```
lesauxvi: can you change these three lines with something more compact/elegant. I see at least two possibilities:

1. use a dict. For exemple
`d = dict(self.items)`
`return d[plotItem]`
should work. You should also add some verification to make sure plotItem actually exists (just check the key exists).

2. use a comprehension list (I prefer this solution). Something like:
`out = [(it, lab) for it, lab in self.items if it==plotItem]`
`return out[0][1]`
should work. Here again, you should also add some verification to make sure plotItem actually exists (just check `out` is not empty).

Just a remark, should it be interesting to also handle the case where the same plotItem has been added several times ??? If yes, the second possibility is definitely the right solution since if the plotItem has been added several times, then the length of `out` is the number of times it has been added (this information is lost with the dict).

What do you think @campagnola and @ericdill ?

pyqtgraph/graphicsItems/LegendItem.py
```diff
@@ -19,7 +19,7 @@ class LegendItem(GraphicsWidget, GraphicsWidgetAnchor):
         legend.setParentItem(plotItem)
 
     """
-    def __init__(self, size=None, offset=None):
+    def __init__(self, size=None, offset=None, drawFrame=True):
```
lesauxvi: may be replace the `drawFrame` kwarg by `frame`?

lesauxvi:
Can you also change the `addLegend` method of `plotItem`?

pyqtgraph/graphicsItems/LegendItem.py
lesauxvi: Just a small complement. The possibility to add a PlotItem several times will no longer be possible when #343 will be merged.

radjkarl: done - see last commit

lesauxvi:
That is ok for me. Good job!
Let @campagnola  and @ericdill have a brief look at this PR to check I did not miss anything.

j9ac9k:
Thanks for the PR @radjkarl sorry it has taken us so long to get to it.  Merging.

PR #396, opened 2016-11-10
NEW features in imageView

radjkarl:
NEW  method setHistogramLabel
image

NEW method setHistogramPrintView
image

NEW option 'discreteTimeSteps'
to rather jump from 0->1->2 than to slide smoothly when changing time (if you have 3 images, people find it hard to understand why there should be time=2.5)

NEW opts{'autoLevels', 'autoRange'}
look in opts for values for 'autoLevels', 'autoRange' if not given in setImage(**kwargs)

DEBUGGED method play, keypressevent, timeIndex
allow to actually setup framerate

NEW property nframes


j9ac9k:
I don't often use the imageView, any suggestions as to how to go about testing the discrete time steps?

outofculture:
Moving this to #1828

PR #397, opened 2016-11-10
NEW method setColor

radjkarl:
color in glGridItem was fixed before. Now it can be changed


j9ac9k:
This should be fixed with #992

PR #400, opened 2016-11-12
ImageView.Timeline better visibility and fixed

radjkarl:
That's the state so far:
orig
Good enough - but why should it be possible to move the timeline area up:?
orig_timeline_moved
This can be nerving and doesn't help at all. This pull request fixes its position and changes its look slightly:
mod_1

Another issue is, that the timeline handle is barely visible when you chose a white background:
orig_white
I made it visible:
mod_2


j9ac9k:
CI came back all good, thanks for the PR @radjkarl merging.

PR #401, opened 2016-11-16
Join the fileforwarder threads in RemoteGraphicsView

jamessimard:
Deals with the occasional exception thrown when RemoteGraphicsView closes but these threads aren't stopped.


j9ac9k:
Thanks for the PR, sorry it's taken a while to follow up :|

Curious if this is related to #881 ... should test these merges early on.

PR #402, opened 2016-11-18
added method clear() to GLViewWidget

vladimir-kraus:
It can be very useful to be able to clear the scene in one call. It would also be consistent with clear() methods in other widgets. The base class does not have clear() method so there is not danger of changing existing behavior due to overriding.


j9ac9k:
Merging.  Thanks for the PR @vladimir-kraus sorry it's taken us so long to get to this.

PR #408, opened 2016-11-28
Fix `cleanup` when the running qt application is not a QApplication

ales-erjavec:
Fix a segmentation fault in atexit registered cleanup function (in QGraphicsScene constructor) when the running qt application is not a fully fledged QApplication but a QCoreApplication (or possibly QGuiApplication on PyQt5).


campagnola:
Looks good to me. Thanks!

PR #416, opened 2016-12-14
Add print wrapper to work around interrupted system calls on travis

campagnola:
Printing on travis occasionally fail with `IOError: Interrupted system call`. Usually just restarting the job on travis fixes the problem, but this is a pain.

Example: https://travis-ci.org/pyqtgraph/pyqtgraph/jobs/183496470#L1066

This PR is an attempt to work around the issue..

### DIFF
--- a/examples/test_examples.py
+++ b/examples/test_examples.py
@@ -3,13 +3,17 @@ from pyqtgraph import Qt
 from . import utils
 import itertools
 import pytest
-import os
-import __builtin__
+import os, sys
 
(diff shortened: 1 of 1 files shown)


campagnola:
Seems to be working..
@ericdill, what do you think?

campagnola:
I'm going to merge this because it should only affect testing and the IOErrors are blocking most other PRs. Please do comment and we can make changes later if needed.

PR #417, opened 2016-12-15
In PlotDataItem.clear() use corresponding curve.clear() and scatter.clear()

termim:
Otherwise when stepMode is True curve.setData([]) causes exception:
"len(X) must be len(Y)+1 ..."


termim:
closes #605

j9ac9k:
Looks good to me.

PR #419, opened 2016-12-15
Adding color support for GLGridItem

dbbh:
GLGridItem init has attribute color which is ignored. Currently the grid is hardcoded to draw in white. This minor modification takes into consideration a 4-element tuple (r,g,b,a) such as glColor passed by the user at the method's init to draw the grid at the desired color. The Standard grid color is still white.


2xB:
Thank you for this Pull Request! The functionality of the `color` parameter was later implemented in #648. As this is merged, the remaining feature here is the `setColor` method. For convenience, I cherry-picked the corresponding commit into the new Pull Request #992, and would suggest to continue discussion there. If I missed something, I will of course also re-open this PR.

PR #424, opened 2017-01-04
Fix crash when running pyqtgraph with python -OO

colinba:
Running pyqtgraph with python -OO gives a crash since version 0.10.0


campagnola:
Looks good, thanks!

PR #425, opened 2017-01-07
Removed unnecessary 'curve1' from 'global'

james1293:
It works without "curve1" being global, so I figured we should take it out to avoid confusing newcomers (like me).

(Note: This is the first time I've contributed to someone else's githib repo, so let me know if minor changes like this are silly.)


campagnola:
Thanks! I like the examples to be as clear as possible :)

PR #429, opened 2017-01-17
fix fft premature slicing away of 0 freq bin

pijyoi:
#179 already slices away the first element when doing X log scale.
the old implementation slices away the first frequency value but still computes the freq coord as starting from 0.

also fixes:
- use rfft for better efficiency
- use rfftfreq to compute coords correctly
  - works for both odd/even lengths
- python3: integer division needed for numpy indexing


campagnola:
Ah! That's embarrassing. 
Thanks for the fix. I'll add a unit test to cover this.

PR #423, opened 2017-01-03
python3 fixes for flowchart saving

Ylannl:
This fixes a crash on pressing the save button in the flowchart panel with python 3.

PR #435, opened 2017-02-05
fix flowcharts saveState in python3.5

cycomanic:
OrderedDict values can not be concatenated with '+' in python 3.5, which results in failure when trying to save the flowchart. Here is a trivial fix (sorry about the 3 commit messages, my editor automatically removed whitespace in my first commit).


cycomanic:
Sorry just saw that there's already a pull request (I only looked at the issues).

QuLogic:
The other PR is #423.

j9ac9k:
Implemented with #886

PR #443, opened 2017-02-15
Add bugfix for exportClicked() for PyQt5

mgrady3:
The Export functionality from the context menu was not working properly with PyQt5 if the User clicked the cancel button once the FileDialog was opened.

The is a result of an API change in PyQt5 where the return type of getSaveFileName (along with many other file dialog functions) is now a tuple. The first item in the tuple is the string of the file name which was the item returned in PyQt4. 

If the Cancel button is clicked, previously this situation was addressed by checking the fileName for an empty string.  Now an additional check is made to see if the return type is tuple. If so, then the first item of the tuple is checked for an empty string which indicates the Cancel scenario.

This will change nothing if PyQt4 is being used and will fix the Cancel scenario for PyQt5.

Tested on Mac OS X.11.6 and Ubuntu 16.0.4 with PyQt 5.6

### DIFF
--- a/pyqtgraph/imageview/ImageView.py
+++ b/pyqtgraph/imageview/ImageView.py
@@ -20,7 +20,7 @@ if USE_PYSIDE:
     from .ImageViewTemplate_pyside import *
 else:
     from .ImageViewTemplate_pyqt import *
-    
+
 from ..graphicsItems.ImageItem import *
 from ..graphicsItems.ROI import *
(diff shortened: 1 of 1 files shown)


ixjlyons:
Thanks for the pull request and sorry it took so long to address it. This looks like it was fixed by 3fd56eb195b10e209c3c00376b37408328fce58f

PR #444, opened 2017-02-15
Fix duplicate menus in GradientEditorItem

zachlowry:
Add call to ev.accept in Tivk.mouseClickEvent to prevent parent menu from opening on a right click of a Tick.


j9ac9k:
blerg, you would think i would have figured out how to do this without messing up people's branches... sorry about that @zachlowry I'll try and get this cleaned up

j9ac9k:
Thanks for the PR @zachlowry sorry it's taken us so long to get this merged.

Issue #442, opened 2017-02-11
error messages when working with pyqtgraph multiprocess


peendebak:
When I remotely imported a module, I cannot use the module properly in ipython. There is no issue in python. A minimal example is included below. The remote import of numpy results in an object that gives error messages. The `print` and `repr` function work fine, in ipython some other properties of the object are checked.
```
import pyqtgraph as pg
import pyqtgraph.multiprocess as pgmp

app=pg.mkQApp()
proc = pgmp.QtProcess()  # pyqtgraph multiprocessing
module = proc._import('numpy')
print(module) # fine
module # generates error messages
```

The error messages printed to screen are:
```
<module 'numpy' from '/home/eendebakpt/.local/lib/python3.5/site-packages/numpy/__init__.py'>
===== Remote process raised exception on request: =====
Traceback (most recent call last):
  File "/home/eendebakpt/develop/pyqtgraph-0.10.0/pyqtgraph/multiprocess/remoteproxy.py", line 218, in handleRequest
    result = getattr(opts['obj'], opts['attr'])
AttributeError: module 'numpy' has no attribute '_ipython_canary_method_should_not_exist_'

===== Local Traceback to request follows: =====
===== Remote process raised exception on request: =====
Traceback (most recent call last):
  File "/home/eendebakpt/develop/pyqtgraph-0.10.0/pyqtgraph/multiprocess/remoteproxy.py", line 218, in handleRequest
    result = getattr(opts['obj'], opts['attr'])
AttributeError: module 'numpy' has no attribute '_ipython_display_'

===== Local Traceback to request follows: =====
===== Remote process raised exception on request: =====
Traceback (most recent call last):
  File "/home/eendebakpt/develop/pyqtgraph-0.10.0/pyqtgraph/multiprocess/remoteproxy.py", line 218, in handleRequest
    result = getattr(opts['obj'], opts['attr'])
AttributeError: module 'numpy' has no attribute '_ipython_canary_method_should_not_exist_'

===== Local Traceback to request follows: =====
===== Remote process raised exception on request: =====
Traceback (most recent call last):
  File "/home/eendebakpt/develop/pyqtgraph-0.10.0/pyqtgraph/multiprocess/remoteproxy.py", line 218, in handleRequest
    result = getattr(opts['obj'], opts['attr'])
AttributeError: module 'numpy' has no attribute '_repr_jpeg_'

===== Local Traceback to request follows: =====
===== Remote process raised exception on request: =====
[...]

PR #445, opened 2017-02-16
add warnings for remote exceptions

eendebakpt:
For remote objects that generate an exception the exception is printed to screen. This is inconvenient of the software using the remote object wants to generate an exception (ipython does this to test some capabilities of objects). This PR solves this issue by replacing the `print` statement with a `warning.warn` This solves #442 by allowing the user to filter output on a `RemoteExceptionWarning`. 

An alternative to warnings would be to use the Python logging framework, or add an option to the `pyqtgraph` to not print exceptions at all.

### DIFF
--- a/pyqtgraph/multiprocess/remoteproxy.py
+++ b/pyqtgraph/multiprocess/remoteproxy.py
@@ -1,6 +1,7 @@
 import os, time, sys, traceback, weakref
 import numpy as np
 import threading
+import warnings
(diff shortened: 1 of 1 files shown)


peendebak:
@campagnola Can you have a look at this PR?

j9ac9k:
I'm good with this change.  Will ask someone else to look this over as well.

PR #446, opened 2017-02-17
Included matplotlib 2.0 colormaps

WFrsh:
Included the four Perceptually Uniform Sequential colormaps from the matplotlib 2.0 release.

- viridis
- inferno
- plasma
- magma

colormaps_reference_00


cjermain:
@campagnola, what do you think about including this PR in the next release? There is upstream interest for these colorschemes, and it appears to be an easy PR to review. Thanks!

PR #447, opened 2017-02-19
ScatterPlotWidget deprecated boolean index now int

zvodd:
Numpy doesn't like boolean indexes for ndarrays
```
IndexError: in the future, 0-d boolean arrays will be interpreted as a valid boolean index
```


2xB:
This should be fixed with https://github.com/pyqtgraph/pyqtgraph/commit/4d26b4bb20013eb5b90b5c5f5b058b26d0eb8906, as now no 0-d arrays are generated. It is also fixed by using a newer version of numpy that handles element-wise comparison as expected.
Thank you for the Pull Request!

PR #356, opened 2016-07-09
Fix some NumPy warnings

QuLogic:
This fixes several of the simpler NumPy warnings I found when running the examples. Either indexing with a non-integer or doing things that trigger future deprecation warnings.

PR #455, opened 2017-03-04
Use floor division

epifanio:
needed to avoid float as index in array slicing.


QuLogic:
The first block of changes is also in #356. I'm wondering if the remaining blocks are necessary? Does `.translate` require integers?

j9ac9k:
Thanks for the PR, looks like the first block of changes is already incorporated, and it does not appear that `translate` requires integers.

PR #458, opened 2017-03-07
fix image export bug

### LINKED ISSUE #454: export image bug related to type casting
Hi there,

I've got a Qt5-only Python distro installed (WinPython 3.5 Qt5) which includes pyqtgraph 0.10.0. Exporting images from PlotWidgets and ImageViews doesn't work anymore and gives this exception:

`Traceback (most recent call last):
  File "C:\WinPython35_Qt5\python-3.5.3.amd64\lib\site-packages\pyqtgraph\exporters\Exporter.py", line 77, in fileSaveFinished
  File "C:\WinPython35_Qt5\python-3.5.3.amd64\lib\site-packages\pyqtgraph\exporters\ImageExporter.py", line 70, in export
    bg = np.empty((self.params['width'], self.params['height'], 4), dtype=np.ubyte)
TypeError: 'float' object cannot be interpreted as an integer
QWaitCondition: Destroyed while threads are still waiting`

Didn't happen with WinPython 3.5 Qt4 (pyqtgraph 0.9.10 I think). Am I the only one experiencing this?

Update: simple fix: in ImageExporter.py, line 70:
`bg = np.empty((int(self.params['width']), int(self.params['height']), 4), dtype=np.ubyte)`

ZacDiggum:
fixes #454

### DIFF
--- a/pyqtgraph/exporters/ImageExporter.py
+++ b/pyqtgraph/exporters/ImageExporter.py
@@ -23,8 +23,8 @@ class ImageExporter(Exporter):
             bg.setAlpha(0)
             
         self.params = Parameter(name='params', type='group', children=[
-            {'name': 'width', 'type': 'int', 'value': tr.width(), 'limits': (0, None)},
-            {'name': 'height', 'type': 'int', 'value': tr.height(), 'limits': (0, None)},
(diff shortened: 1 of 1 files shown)


campagnola:
Looks good; thanks!
I'm also working on the int / parameter issue in #479.

PR #459, opened 2017-03-08
Fix color alpha limits in LinearRegionItem hover.

scamille:
Clip hover color alpha to limits (0, 255) for LinearRegionItem, to
remove QT warnings about calling the function QColor::setAlpha with a
invalid value.


2xB:
Thank you for this Pull Request! A similar fix was later implemented in https://github.com/pyqtgraph/pyqtgraph/commit/98cdc65049a8b61169982bb7add833f923979bf4.

PR #465, opened 2017-03-28
Fix np image exporter error

CPrescher:
Fixing the numpy error for non-integer dimensions when creating an empty array in the ImageExporter Class


QuLogic:
Also fixed in #356.

campagnola:
Looks like these changes are now in mainline.

Issue #466, opened 2017-03-29
Image (NumPy array) data not released when closing plot windows (memory leak?)


nanoant:
Long story short, pyqtgraph is not releasing NumPy array variables once they are used by some plots.

Run following commands line by line and observe your memory usage (eg. in task manager):
```python
import numpy
import pyqtgraph
a=numpy.ones((128,1024,1024)) #<- 1GB now used by python
del a                         #<- memory released python now is using just few MBs
a=numpy.ones((128,1024,1024)) #<- 1GB now used again by python
pyqtgraph.image(a) # opens a window
pyqtgraph.QtGui.QGuiApplication.exec() # just run Qt event loop, so we can close the window
# now close the plot window
del a #<- python still uses 1GB, even I have just released NumPy array
pyqtgraph.cleanup() #<- no effect, still 1GB used
import gc
gc.collect() #<- absolutely no efffect, still 1GB used
```
Is there any reason that this memory is not collected and pyqtgraph is keeping references to non-accessible plot objects?

nanoant:
Some update on this issue. pyqtgraph seems to track imageViews in `images` global. However clearing this global does not release underlying memory:
```python
pyqtgraph.images.clear() #<- no effect memory not released
```
But if we instead of the command above, assign some 1x1 image to existing images, then most (but not all) memory is reclaimed.
```python
for iw in pyqtgraph.images:
  iw.setImage(numpy.zeros((1,1))) #<- dummy 1x1 image
  iw.parent().close()
```
So it means altogether that pyqtgraph (or Qt?) keeps a reference to `pyqtgraph.graphicsWindows.ImageWindow` even it is no longer accessible by user of `pyqtgraph`, which means that's a memory leak.

Altogether I think that `pyqtgraph` should not keep any strong references, and release memories once the windows are closed and there's no explicit reference to the window or other `pyqtgraph` object in the user's code.

j9ac9k:
Since merging #468 we can close this issue out.

PR #468, opened 2017-04-10
Patch/window handling

nanoant:
This PR reduces number of references to PyQtGraph windows implicitly hold by the package. Also removes some memory leak in `ImageView.close()` not taking care of `self.imageItem`.

This PR is related to https://github.com/pyqtgraph/pyqtgraph/issues/466


pyqtgraph/__init__.py
```diff
@@ -378,8 +378,6 @@ def exit():
 
 ## Convenience functions for command-line use
 
-plots = []
```
campagnola: These references should be necessary to ensure the plots stay alive as long as they the user keeps them open. Perhaps we can have the references automatically removed when the window is closed?

nanoant: Alright, I'll try to keep the references, and remove them on window close.

nanoant:
@campagnola Hi I have updated my PR, but I have no idea why Travis fails with `IOError: [Errno 4] Interrupted system call`. I don't see any of updated code in the call-stack. Any clues?

j9ac9k:
Hi @nanoant sorry it's taken us so long to get to this.  I fixed a merge conflict, this looks good to me.  I'll be merging.

PR #470, opened 2017-04-11
Fixed PySide image memory leak

HashSplat:
PySide has a known memory leak issue when using QImage. It does not handle the reference counter correctly. I manually adjusted the reference counter to the data as suggested in a bug report by Neil Whelchel. This bug report can be found at https://bugreports.qt.io/browse/PYSIDE-140


HashSplat:
This is only an issue for Python 3. May want to put an additional check for Python 3, although I believe this fix should be compatible with Python 2. I have not checked this with Python 2.

campagnola:
Merged with minor modifications. Thanks @HashSplat !

PR #473, opened 2017-04-21
Correct wrong function call in LayoutWidget.addLabel and LayoutWidget…

### LINKED ISSUE #242: QLayoutWidget Add Label Error: No method AddItem
The code in QLayoutWidget looks something like:

```
    def addLabel(self, text=' ', row=None, col=None, rowspan=1, colspan=1, **kargs):
        """
        Create a QLabel with *text* and place it in the next available cell (or in the cell specified)
        All extra keyword arguments are passed to QLabel().
        Returns the created widget.
        """
        text = QtGui.QLabel(text, **kargs)
        self.addItem(text, row, col, rowspan, colspan)
        return text
```

I think it should have `self.addWidget` instead of `self.addItem` as addItem method is missing.

Error:

```
AttributeError: 'LayoutWidget' object has no attribute 'addItem'
```

tmtnslt:
….addLayout

Previously LayoutWidget.addLabel and LayoutWidget.addLayout called a function "addItem", that didn't exist. Corrected to call LayoutWidget.addWidget.

This fixes #242

### DIFF
--- a/pyqtgraph/widgets/LayoutWidget.py
+++ b/pyqtgraph/widgets/LayoutWidget.py
@@ -39,7 +39,7 @@ class LayoutWidget(QtGui.QWidget):
         Returns the created widget.
         """
         text = QtGui.QLabel(text, **kargs)
-        self.addItem(text, row, col, rowspan, colspan)
+        self.addWidget(text, row, col, rowspan, colspan)
         return text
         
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for the PR, sorry we're just now getting to this, this looks good to me; and I tested your fix with the examples.

PR #487, opened 2017-05-18
debugged: RGB images cause exception if GradientEditorItem is not 'grey'

radjkarl:
to reproduce error:
- load RGB image into imageView
- change gradient to anything, but 'grey'
--> you will see the error:

```
    argb, alpha = fn.makeARGB(image, lut=lut, levels=levels)
  File "/home/karl/git/pyqtgraph_karl/pyqtgraph_karl/functions.py", line 1110, in makeARGB
    imgData[..., i] = data[..., order[i]] 
ValueError: could not broadcast input array from shape (512,512,3) into shape (512,512)
```

no gradient can still be changed but no error is generated

### DIFF
--- a/pyqtgraph/graphicsItems/ImageItem.py
+++ b/pyqtgraph/graphicsItems/ImageItem.py
@@ -399,7 +399,11 @@ class ImageItem(GraphicsObject):
         
         if self.axisOrder == 'col-major':
             image = image.transpose((1, 0, 2)[:image.ndim])
-        
+
+        if image.ndim > 2:
+            #no lookup table form color images
+            lut = None
+
         argb, alpha = fn.makeARGB(image, lut=lut, levels=levels)
         self.qimage = fn.makeQImage(argb, alpha, transpose=False)
 


2xB:
I could reproduce this using the following (adapted example) code with pyqtgraph 0.10.0, but not with the version that is currently on `develop`. It seems to be fixed with commit https://github.com/pyqtgraph/pyqtgraph/commit/c3e52f15b0df0104455922512762887fdfc81e25 as part of #621.

```python3
import numpy as np
import pyqtgraph as pg

# Interpret image data as row-major instead of col-major

app = QtGui.QApplication([])

## Create window with ImageView widget
win = QtGui.QMainWindow()
win.resize(800,800)
imv = pg.ImageView()
win.setCentralWidget(imv)
win.show()
win.setWindowTitle('pyqtgraph example: ImageView')

## Create random 3D data set with noisy signals
img = np.random.normal(size=(200, 200, 3)) * 20 + 100
img = img[np.newaxis,:,:,:]
decay = np.exp(-np.linspace(0,0.3,100))[:,np.newaxis,np.newaxis,np.newaxis]
data = np.random.normal(size=(100, 200, 200, 3))
data += img * decay
data += 2

## Add time-varying signal
sig = np.zeros(data.shape[0])
sig[30:] += np.exp(-np.linspace(1,10, 70))
sig[40:] += np.exp(-np.linspace(1,10, 60))
sig[70:] += np.exp(-np.linspace(1,10, 30))

sig = sig[:,np.newaxis,np.newaxis,np.newaxis] * 3
data[:,50:60,30:40,:] += sig


## Display the data and assign each frame a time value from 1.0 to 3.0
imv.setImage(data, xvals=np.linspace(1., 3., data.shape[0]))

## Set a custom color map
colors = [
    (0, 0, 0),
    (45, 5, 61),
    (84, 42, 55),
    (150, 87, 60),
    (208, 171, 141),
    (255, 255, 255)
]
cmap = pg.ColorMap(pos=np.linspace(0.0, 1.0, 6), color=colors)
imv.setColorMap(cmap)

## Start Qt event loop unless running in interactive mode.
if __name__ == '__main__':
    import sys
```

Thank you for the report and the Pull Request!

PR #496, opened 2017-06-08
draw connector lines between gradient and region with anti-aliasing

radjkarl:
render the connector lines with anti-aliasing. independent from chosen pyqtgraphs CONFIG_OPTIONS.

why?
-----
it looks much better! its only 2 lines - i dont see any reason why this option should be controllable with   CONFIG_OPTIONS['antialias']

without anti-aliasing:
image

with anti-aliasing:
image


j9ac9k:
Resolving merge conflict, and letting the CI chew on this.

j9ac9k:
Thanks for the PR @radjkarl sorry it took us so long to get to this.  It's in the library now!

Issue #490, opened 2017-05-30
TypeError in isosurface


draustin:
I recently transitioned my development environment to another computer, and found that the isosurface function is broken. It reports 

  File "d:\git\pyqtgraph\pyqtgraph\functions.py", line 2142, in isosurface
    faceTableI[faceTableInds[:,0]] = np.array([triTable[j] for j in faceTableInds])
TypeError: only integer scalar arrays can be converted to a scalar index

A quick Google reveals that the problem is very likely related to a recent change in Numpy which makes it an error to use a single-element array for indexing as described here:
https://stackoverflow.com/questions/42128830/typeerror-only-integer-scalar-arrays-can-be-converted-to-a-scalar-index

The following trivial change to line 2142 
faceTableI[faceTableInds[:,0]] = np.array([triTable[j[0]] for j in faceTableInds])
seems to fix the problem.

I am happy to put this into a pull request, but I am a Git newbie so it might be easier for the pyqtgraph developers to make this change directly...

edbarnard:
I've put your fix in as a pull request. Hopefully I did it right!

QuLogic:
A similar fix is part of #356.

sarbabi:
Hi, any ideas why this error again comes up?
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-37-d8d6ec1d9d2a> in <module>
----> 1 pg.isosurface(my_array, 0.5)

c:\python\lib\site-packages\pyqtgraph\functions.py in isosurface(data, level)
   2067             faceTableI = np.zeros((len(triTable), i*3), dtype=np.ubyte)
   2068             faceTableInds = np.argwhere(nTableFaces == i)
-> 2069             faceTableI[faceTableInds[:,0]] = np.array([triTable[j] for j in faceTableInds])
   2070             faceTableI = faceTableI.reshape((len(triTable), i, 3))
   2071             faceShiftTables.append(edgeShifts[faceTableI])

c:\python\lib\site-packages\pyqtgraph\functions.py in <listcomp>(.0)
   2067             faceTableI = np.zeros((len(triTable), i*3), dtype=np.ubyte)
   2068             faceTableInds = np.argwhere(nTableFaces == i)
-> 2069             faceTableI[faceTableInds[:,0]] = np.array([triTable[j] for j in faceTableInds])
   2070             faceTableI = faceTableI.reshape((len(triTable), i, 3))
[...]

PR #498, opened 2017-06-12
Fixes an issue where isosurface fails on later versions of numpy. See Issue #490

edbarnard:
Fixes an issue where isosurface fails on later versions of numpy. See Issue #490


QuLogic:
This should be fixed by #356 also.

PR #501, opened 2017-06-21
Add some compatibility for Python3

ausk:
Mainly add some compatibility for Python3. More exactly, it is change some division operations from "/" to "//" for indexing list.


j9ac9k:
Looks like the majority of these were implemented since you made the PR.  Thanks for the submission regardless.

PR #502, opened 2017-06-21
fix some small indexing bug.

ausk:


j9ac9k:
This PR contained one fix and an unneeded change, but the fix was already incorporated.  Thanks for the PR.

PR #503, opened 2017-06-22
Fix bug in RawImageWidget.py

kiwi0fruit:
For example: it prevents integration of this widget to Enaml.

Actually I'm not sure that this wasn't intentional... But for some reason `parent` was not used and set to `None`.

New:

1. Bug was in the `def paintGL(self)` method (at least with PySide1): image was mirrored upside down.
2. Added support for `setConfigOptions(imageAxisOrder='row-major')`
3. Small cosmetic changes


kiwi0fruit:
Actually I'm not sure that this wasn't intentional... But for some reason `parent` was not used and set to None.

j9ac9k:
Merging even w/ CI failures, as CI failures were a result of an issue fixed with #1241

PR #504, opened 2017-06-28
Make two class names into links

enterprisey:


ixjlyons:
Thanks for the PR and sorry it sat around for so long. The graphicswindow page was reworked a bit to include the API information in #1142

PR #505, opened 2017-06-30
Avoid imprecission errors in siScale function

alfonnews:
With this change avoid precission errors on multiply operation


2xB:
Note: As this Pull Request is not generated from a seperate branch, it contains more commits than intended. It is currently a combination of PR #507 and PR #508 with a fix for a bug introduced (but not yet resolved) in #508.

j9ac9k:
As this is a combination of two other open PRs, let's close this one out and deal with the other two individually.

PR #507, opened 2017-07-03
Fix Dock close event QLabel still running with no parent

alfonnews:


alfonnews:
Explicit calls Dock Label QLabel close function to avoid that QLabels keep running with no parent

j9ac9k:
Thanks for the PR @alfonnews sorry it took so long for us to get to this.  I had to modify your PR slightly to avoid any merge conflicts.

PR #508, opened 2017-07-03
Fix siScale imprecision errors

alfonnews:
Fix siScale imprecision errors


pyqtgraph/functions.py
```diff
@@ -75,9 +75,8 @@ def siScale(x, minVal=1e-25, allowUnicode=True):
             pref = SI_PREFIXES[m+8]
         else:
             pref = SI_PREFIXES_ASCII[m+8]
-    p = .001**m
-    
-    return (p, pref)
```
2xB: The `return` statement being removed looks like a mistake?

2xB: To fix this, just commit the fix to your branch `alfonnews:Fix_siScale_imprecision_errors`. Thank you!

pyqtgraph/functions.py
```diff
@@ -75,9 +75,8 @@ def siScale(x, minVal=1e-25, allowUnicode=True):
             pref = SI_PREFIXES[m+8]
         else:
             pref = SI_PREFIXES_ASCII[m+8]
-    p = .001**m
-    
-    return (p, pref)    
+    m1 = -3*m
+    p = float('1E{0}'.format(m1))
```
2xB: Somewhat cleaner is
```suggestion
    p = 10.**m1
```
, as it abstains from any type casting and formatting, while ensuring the same amount of precision.

j9ac9k:
Grr...how did that other commit get in here...

PR #513, opened 2017-07-26
Making flowchart saving compatible with Python 3

npyoung:
- Replaces some calls to `unicode` made by `Flowchart` as it is being saved with calls to `asUnicode`, which is Python 2/3 compatible.
- Fixes an attempt to merge two OrderedDict values as if they were lists. In Python 3, `OrderedDict.values()` returns a generator. These objects have been wrapped in `list()` to make them mergeable in Python 2 or 3.


Ryytikki:
Can confirm that these issues still exist and that this successfully fixes them

ksunden:
Did a quick Grep and this does seem to be the only file which has improper use of the `unicode` type (others are all in Py2 only blocks)

pyqtgraph/flowchart/Flowchart.py
```diff
@@ -519,7 +520,8 @@ def loadFile(self, fileName=None, startDir=None):
             self.fileDialog.fileSelected.connect(self.loadFile)
             return
             ## NOTE: was previously using a real widget for the file dialog's parent, but this caused weird mouse event bugs..
-        fileName = unicode(fileName)
+            #fileName = QtGui.QFileDialog.getOpenFileName(None, "Load Flowchart..", startDir, "Flowchart (*.fc)")
```
ksunden: Extra comment left in place here

j9ac9k:
Implemented with #886

PR #519, opened 2017-07-29
Expose the `clickable` property of `PlotDataItem`.

chrismullins:
Currently if you attempt to set the `clickable` property of a PlotDataItem,
this property is silently ignored. The expected behavior is to set the
`clickable` property of the underlying PlotCurveItem, and to return the 
property from the PlotCurveItem.

This commit allows us to set the `clickable` property in these ways:
```
pdi = pq.PlotDataItem(...,clickable=True)
pdi.setClickable(True)
pdi.clickable = True
```
and "get" the property in this way:
```
print(pdi.clickable)
```
without referencing `pdi.curve`.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -174,7 +174,7 @@ class PlotDataItem(GraphicsObject):
             
             'data': None,
         }
-        self.setClickable(kargs.get('clickable', False))
+        self.setCurvesClickable(kargs.get('clickable', False))
         self.setData(*args, **kargs)
     
     def implements(self, interface=None):
@@ -186,17 +186,12 @@ class PlotDataItem(GraphicsObject):
     def name(self):
         return self.opts.get('name', None)
 
-    def setClickable(self, s, width=None):
+    def setCurvesClickable(self, s, width=None):
         self.curve.setClickable(s, width)
 
-    @property
-    def clickable(self):
+    def curvesClickable(self):
         return self.curve.clickable
 
-    @clickable.setter
-    def clickable(self, s):
-        self.setClickable(s)
-    
     def boundingRect(self):
         return QtCore.QRectF()  ## let child items handle this
 
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -174,6 +174,7 @@ class PlotDataItem(GraphicsObject):
             
             'data': None,
         }
+        self.setClickable(kargs.get('clickable', False))
         self.setData(*args, **kargs)
     
     def implements(self, interface=None):
@@ -184,6 +185,17 @@ class PlotDataItem(GraphicsObject):
     
     def name(self):
         return self.opts.get('name', None)
+
+    def setClickable(self, s, width=None):
+        self.curve.setClickable(s, width)
+
+    @property
+    def clickable(self):
+        return self.curve.clickable
+
+    @clickable.setter
+    def clickable(self, s):
+        self.setClickable(s)
     
     def boundingRect(self):
         return QtCore.QRectF()  ## let child items handle this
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -184,6 +185,17 @@ def implements(self, interface=None):
...
         return self.opts.get('name', None)
+
+    def setClickable(self, s, width=None):
+        self.curve.setClickable(s, width)
+
+    @property
+    def clickable(self):
+        return self.curve.clickable
+
+    @clickable.setter
```
campagnola: Do we need both the `setClickable` method and the property setter? In general I have avoided setters in pyqtgraph, in part to be consistent with Qt, and also to avoid the issue where setting a misspelled property is silently ignored.

chrismullins: Good point, I hadn't thought about that.  My goal was to mirror the interface of `PlotCurveItem` so that users don't need to know about the underlying implementation in order to set that property outside the initialization. Especially since the `clickable` property is `False` by default.

An alternative would be to have line 177 read:
```
self.curve.setClickable(kargs.get('clickable', False))
```
if you prefer.  That way the example would still work interchangably with `PlotDataItem`, but if someone is keeping track of what's returned from `PlotItem.plot()` they will have to know to use the `curve` to change that attribute.

j9ac9k:
Hi @chrismullins 

Sorry we haven't followed up in so long.  We've been chipping away at the PR queue, and some-how this PR gets postponed; but I'm motivated to get this feature merged, as you are absolutely right, `PlotDataItem` should expose the `clickable` property from the `PlotCurveItem`'s it holds.

I recognize it's been almost 4 years and you've likely moved on to other projects, and this PR is likely no longer relevant to you.  I don't mind making some changes to it and merging it; but figure I would give you the option first. 

I would recommend instead of trying to emulate the `PlotCurveItem` API (which is somewhat unfortunate that we have `PlotCurveItem.clickable` a property and not a method from how Qt API typically goes), we introduce a `PlotDataItem.setCurvesClickable` and `PlotDataItem.curvesClickable` (this is analogous to `QHeaderView.setSectionsClickable() and QHeaderView.sectionsClickable()`)

In a separate PR we can worry about making `PlotCurveItem.clickable` have a `FutureWarning` about it changing from a method to a property.

chrismullins:
Hi there @j9ac9k !  

It's as much my fault for not following up on this 😆 I've updated to be more in line with your suggestions, let me know what you think!

j9ac9k:
> It's as much my fault for not following up on this 😆 I've updated to be more in line with your suggestions, let me know what you think!

As I've gone through the backlog or PRs, I've been blown away by how many people are not only still active on github, but are willing to make recommended changes to PRs submitted _years_ ago (for the record, there is no expectation of this on our end, before I take over someone else's PR, I try to give them the opportunity to make the changes themselves).  I always feel so conflicted about it, on one hand, I'm super thankful, and just hope that this is a reflection of the importance of this project, on the other, this situation is clearly a reflection of a failure on our part.  So, please pardon us for the nearly 4 year silence, I promise it will be less than 4 years before we resolve this PR 😆 

As this introduces a new API hook I'm going to ask some of the other maintainers to chime in with what they think.  @NilsNemitz @ixjlyons, I should really remove the two of you from my clipboard 😆

NilsNemitz:
I would argue that a PlotDataItem has precisely one curve, and that the methods should therefore be the singular
`setCurveClickable()` and `curveClickable()`.
API-wise, this seems reasonable; I guess at some point we might want to have a `setPointsClickable()` method, too.

j9ac9k:
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/PlotDataItem.py#L158-L160

Blerg, you're right, there is just one curve, it's `PlotItem` that has curves (plural).  Sorry @chrismullins would you mind changing the method to singular?

j9ac9k:
I'm going to close/open this PR to trigger the new CI pipelines... assuming nothing blows up, I'll merge after.

j9ac9k:
Huh, that didn't work, ok, I'll just merge.  Thanks @chrismullins and sorry again for the delayed response.

j9ac9k:
and I merged to the wrong branch because I'm not paying attention....

PR #526, opened 2017-08-04
Added essential if.

kokokostation:
When autorange is enabled and there is no dots of `ScatterPlotItem` on current scene, `np.nanmin` (line 655) throws exception as `d` is empty.


campagnola:
Awesome, thanks!

PR #528, opened 2017-08-08
Fixed downsample inf image issue

HashSplat:
The fn.downsample function can cause an image array size to go to 0. This will cause an error in makeQImage.


campagnola:
Thanks!

PR #529, opened 2017-08-08
Fixed dataBounds issue for a line with inf values.

justengel:
Fixed dataBounds issue for an inf line.


campagnola:
Looks like this is already fixed upstream, agreed?

justengel:
It looks like this is fixed in the upstream develop version using np.nanmin.

PR #541, opened 2017-08-31
ImageItem render error if graph is too small.

justengel:
Fixed issue where render would error because 'mapToDevice' would return None if the view size was too small.


campagnola:
..and thanks!

PR #553, opened 2017-09-12
Fixed legend size after remove item

HashSplat:
When the legend removed an item with a large label name the legend would not shrink in width. This fix uses the sample and label minimum width to allow the legend to shrink when an item is removed.


campagnola:
Looks good to me. Thanks!

PR #565, opened 2017-09-20
Fixed AxisItem mouse drag issue

justengel:
Issue where MouseDragEvent would only work for 1 AxisItem. I made it so the MouseDragEvent could propagate to other AxisItems. I had this issue by setting the ViewBox background color and changing the AxisItem ZValue to make the AxisItems visible. This made one of the AxisItems receive all of the MouseDragEvents and only one Axis would actually allow dragging. ViewBox.mouseDragEvent sets event.accept() preventing the mouse event to propagate to other AxisItems.


campagnola:
..and thanks again!  I might implement a better fix for this in the future, but it's an improvement for now.

justengel:
Awesome! Thank you so much. Can we get an updated version of pyqtgraph on pypi?

PR #571, opened 2017-09-23
Fixed: items are not removed from legend

MichaelSchreier:
LegendItem.removeItem requires the name of the item to be removed as an input. However, here the entire GraphicsItem was passed instead.


campagnola:
@MichaelSchreier I'm going to close this, but please reopen if you think this is still an issue.

Issue #576, opened 2017-09-28
BarGraphItem does not support horizontal bars


acburigo:
[This line](https://github.com/pyqtgraph/pyqtgraph/blob/5195d9dd6308caee87e043e859e7e553b9887453/pyqtgraph/graphicsItems/BarGraphItem.py#L123) breaks when `x0` is a scalar, which disagrees with [this line](https://github.com/pyqtgraph/pyqtgraph/blob/5195d9dd6308caee87e043e859e7e553b9887453/pyqtgraph/graphicsItems/BarGraphItem.py#L129).

PR #577, opened 2017-09-28
BarGraphItem can plot horizontal bars.

acburigo:
Proposed fix to https://github.com/pyqtgraph/pyqtgraph/issues/576


campagnola:
Good catch, thanks.

PR #598, opened 2017-10-20
fix: some unreadable items in dark themes

sem-geologist:
The problem:
on KDE5 breeze dark theme few pyqtgraph items are not readable. I think the problem is not only on KDE5, but apparently could be seen on win10? It is caused by statically set background color, supposing that default font color is black, but that is not the case when OS uses dark theme/skins.
Two items were identified: 
1. subgroup items in parameter tree:
parameter_tree1
2. console output:
console_foreground

Fix:
The foreground color of these items is statically set.


sem-geologist:
and this is how it looks after fix with dark theme:
after_fix

and with light theme it looks ok as well:
after_fix_light_breeze

campagnola:
Looks good, thanks!

PR #600, opened 2017-10-22
Update ImageItem.py

miranis:
Functions nanmin and nanmax are defined in the numpy module and cannot be accessed from the global namespace. An exception is raised each time the code in ImageItem:331 is executed.


campagnola:
Thanks!

PR #606, opened 2017-11-11
Avoid raising a general Exception for Dock

satarsa:
It is not very pythonic to raise just a general Exception whenever an error appears. Here, for Dock, we define a DockError class, which is always raised whenever there is an exception, which makes it much easier to catch.


campagnola:
In general I am for replacing `Exception` with something more specific, but:
* Replacing `TypeError` and `ValueError` is not backward compatible (and I think `DockError` is less specific in those cases anyway)
* Usually I prefer to use the builtin exception classes (so the user doesn't need an extra import just to catch) unless the exception is meant to carry some unique data structure.

Issue #603, opened 2017-11-06
ReferenceError on closing plot in pyqtgraph 0.10.0 & python 3.6 


jphdotam:
I have a plot as follows:

```
        app = QtGui.QApplication([])

        win = pg.GraphicsWindow(title="ETT: {} / Stage: {} / Lead: {}".format(self.patient_id,self.stage,self.lead))
        win.setWindowTitle("ETT: {} / Stage: {} / Lead: {}".format(self.patient_id,self.stage,self.lead))

        plot = win.addPlot(title=self.lead, x=self.x, y=self.y, pen='b')
        plot.setYRange(0, 35, padding=0)
        plot.setXRange(300, 1200, padding=0)

        marker_qrs_onset.sigPositionChanged.connect(update_qrs)
        marker_qrs_offset.sigPositionChanged.connect(update_qrs)

        app.instance().exec_()

        print("App closed")
```

When I close the plotting window, the print() statement never runs, and instead I get:

```
Traceback (most recent call last):
  File "/anaconda/lib/python3.6/site-packages/pyqtgraph/__init__.py", line 312, in cleanup
    if isinstance(o, QtGui.QGraphicsItem) and isQObjectAlive(o) and o.scene() is None:
ReferenceError: weakly-referenced object no longer exists
```

If I add an exception for this ReferenceError to the cleanup() function in pyqtgraph's __init__.py then it works perfectly, as it now reads:

```
    for o in gc.get_objects():
        try:
            if isinstance(o, QtGui.QGraphicsItem) and isQObjectAlive(o) and o.scene() is None:
                if getConfigOption('crashWarning'):
                    sys.stderr.write('Error: graphics item without scene. '
                        'Make sure ViewBox.close() and GraphicsView.close() '
                        'are properly called before app shutdown (%s)\n' % (o,))
                
                s.addItem(o)
        except RuntimeError:  ## occurs if a python wrapper no longer has its underlying C++ object
            continue
        except ReferenceError: ##ADDED
            continue
    _cleanupCalled = True
```

This is tested on pyathon 3.5 & 3.6 on Windows and on Mac.

toinsson:
could also be:

```
except (RuntimeError, ReferenceError):
```

I get this issue when running pyqtgraph with tensorflow. This is referenced on [SO](https://stackoverflow.com/questions/41542571/pyqtgraph-tries-to-put-tensorflow-stuff-onto-a-qgraphicsscene-on-cleanup).

[...]

PR #613, opened 2017-12-18
Prevent ReferenceErrors on cleanup

jphdotam:
Prevents a ReferenceError being thrown when PyQtGraph tries to put TensorFlow stuff onto a QGraphicsScene() on cleanup - see https://github.com/pyqtgraph/pyqtgraph/issues/603 and https://stackoverflow.com/questions/41542571/pyqtgraph-tries-to-put-tensorflow-stuff-onto-a-qgraphicsscene-on-cleanup


campagnola:
Thanks!

PR #614, opened 2017-12-18
Add custom gradients and name to GradientEditorItem

JoeXinfa:
The first commit is because I always see color-bar names in commercial software. I add a simple name label in front of the color-bar. I think it helps user selection and communication with colleagues. In future, when the list grows even longer, we may add search option.

The second one builds on top of the first one, providing user the freedom to add their favorite gradients.

BTW. I've been using PyQtGraph for several months and I really like it. Thank you very much @campagnola . I just started using GitHub and this is my first pull :)


campagnola:
There are a lot of unrelated changes in this PR (GLTextItem, GLBoxItem, GLAxisItem, histogram orientation, etc.). Generally each PR should focus on solving one problem, otherwise it becomes very difficult to review. You're welcome to re-open these as individual PRs.

PR #615, opened 2017-12-21
Limit wheel mouse scroll to zoom limits if set

jwasys:
Especially with pinch gestures the zoom level gets too deep easily. Some logic from updateViewRange is duplicated into the wheelEvent handler.


campagnola:
When I run `examples/ViewBoxFeatures.py`, the view at middle-right obeys range limits when I roll the mouse wheel. Can you give an example of what behavior is fixed by your changes?

j9ac9k:
Thanks for the PR @jwasys ; I'm unable to replicate the issue that this PR addresses just like @campagnola   I'm going to close out this issue as a result.

If you're still having the issue, could you provide us with the code that you were running, as well as what kind of hardware/interaction you were partaking with to cause the issue you had?

PR #616, opened 2018-01-09
Play image along 't' axis instead of first axis

gram526:
The play function under timeout iterates along axis 0. This could be changed to axis 't'.


campagnola:
Agreed! Thanks.

Issue #292, opened 2016-02-29
PolyLineROI getArrayRegion works wrong


OSjerick:
If I have to get an array region using a `PolyLineROI`, the result is always an array of zeros. This thing doesn't happen if I use a `RectROI`.

2xB:
This is still an issue with `develop`. I tried using the latest pyqtgraph version from pip and from `develop` with the `imageAnalysis.py` example, where I switched the line

```python3
roi = pg.ROI([-8, 14], [6, 5])
```

with

```python3
roi = pg.PolyLineROI(([-8, 14], [6, 5]))
```

and deleted

```python3
roi.addScaleHandle([0.5, 1], [0.5, 0.5])
roi.addScaleHandle([0, 0.5], [0.5, 0.5])
```

. Note that if you do not delete `addScaleHandle`, the result seems to be somewhat similar to the rectangular ROI that is shipped with the example as a default.

PR #618, opened 2018-01-20
ROI.py, getArrayRegion: Fix return mapped coordinates

malikolivier:
The *getArrayRegion* method is defined as returning a tuple of the points
in the selected region and the mapped coordinates if the
*returnMappedCoords* keyword argument is set to True in the parent class
*ROI*.

In the *EllipseROI* and *PolyLineROI* classes, *getArrayRegion* was overriden,
however it ignores the *returnMappedCoords* keyword argument, leading to
unintended bugs because of the change in interface between the parent class and
the subclass.

This patch fixes the above bug.
If *returnMappedCoords* is set to False, then only *arr* containing the
array region is returned. If *returnMappedCoords* is set to True, a
tuple of the array region and the mapped coordinates is returned.

EDIT: Fixed the same bug for PolyLineROI.

@campagnola Do you require unit tests? I wanted to write some but I gave up for now as I cannot get unit tests to succeed in my environment (maybe I missed something in the setup):

```
$ python --version
Python 3.6.3
$ pip --version
pip 9.0.1 from /home/malik/workspace/pyqtgraph/venv/lib/python3.6/site-packages (python 3.6)
$ pip freeze
apipkg==1.4
attrs==17.4.0
coverage==4.4.2
execnet==1.5.0
numpy==1.14.0
pkg-resources==0.0.0
pluggy==0.6.0
py==1.5.2
PyQt5==5.9.2
pytest==3.3.2
pytest-cov==2.5.1
pytest-forked==0.2
pytest-xdist==1.22.0
scipy==1.0.0
sip==4.19.6
six==1.11.0
$ py.test
7 failed, 24 passed, 129 skipped in 4.04 seconds
```

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -1631,16 +1631,25 @@ class EllipseROI(ROI):
         
         p.drawEllipse(r)
         
-    def getArrayRegion(self, arr, img=None, axes=(0, 1), **kwds):
+    def getArrayRegion(self, arr, img=None, axes=(0, 1),
+                       returnMappedCoords=False, **kwds):
         """
         Return the result of ROI.getArrayRegion() masked by the elliptical shape
         of the ROI. Regions outside the ellipse are set to 0.
         """
         # Note: we could use the same method as used by PolyLineROI, but this
         # implementation produces a nicer mask.
-        arr = ROI.getArrayRegion(self, arr, img, axes, **kwds)
+        if returnMappedCoords:
+           arr, mappedCoords = ROI.getArrayRegion(self, arr, img, axes,
+                                                  returnMappedCoords, **kwds)
+        else:
+           arr = ROI.getArrayRegion(self, arr, img, axes,
+                                    returnMappedCoords, **kwds)
         if arr is None or arr.shape[axes[0]] == 0 or arr.shape[axes[1]] == 0:
-            return arr
+            if returnMappedCoords:
+                return arr, mappedCoords
+            else:
+                return arr
         w = arr.shape[axes[0]]
         h = arr.shape[axes[1]]
 
@@ -1653,7 +1662,10 @@ class EllipseROI(ROI):
         shape = [(n if i in axes else 1) for i,n in enumerate(arr.shape)]
         mask = mask.reshape(shape)
         
-        return arr * mask
+        if returnMappedCoords:
+            return arr * mask, mappedCoords
+        else:
+            return arr * mask
     
     def shape(self):
         if self.path is None:
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -1926,7 +1926,8 @@ class PolyLineROI(ROI):
         p.lineTo(self.handles[0]['item'].pos())
         return p
 
-    def getArrayRegion(self, data, img, axes=(0,1), **kwds):
+    def getArrayRegion(self, data, img, axes=(0,1),
+                       returnMappedCoords=False, **kwds):
         """
         Return the result of ROI.getArrayRegion(), masked by the shape of the 
         ROI. Values outside the ROI shape are set to 0.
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/ROI.py
```diff
@@ -1653,7 +1662,10 @@ def getArrayRegion(self, arr, img=None, axes=(0, 1), **kwds):
         shape = [(n if i in axes else 1) for i,n in enumerate(arr.shape)]
         mask = mask.reshape(shape)
         
-        return arr * mask
+        if returnMappedCoords:
+            return arr * mask, mappedCoords * mask
```
campagnola: Doesn't make sense to me to mask the coordinates as well--each element in the returned array has a corresponding location in the original data, even if that element value was masked.

malikolivier: You're right. I pushed a fix. Let me know your thoughts.

campagnola:
Try running `examples/ROIExamples.py` -- these changes break the triangular ROI in the upper-left panel.

pyqtgraph/graphicsItems/ROI.py
```diff
@@ -1914,15 +1926,23 @@ def shape(self):
...
         """
         br = self.boundingRect()
         if br.width() > 1000:
             raise Exception()
-        sliced = ROI.getArrayRegion(self, data, img, axes=axes, fromBoundingRect=True, **kwds)
+        if returnMappedCoords:
+            sliced, mappedCoords = ROI.getArrayRegion(self, data, img,
+                                                      axes=axes,
+                                                      fromBoundingRect=True,
+                                                      **kwds)
```
campagnola: do you need `returnMappedCoords=True` here?

malikolivier: You are right. This is utterly wrong. I got confused and I mixed up the use of  *fromBoundingRect* and *returnMappedCoords*.

I pushed a fix.

malikolivier:
3bc971e fixes the triangular ROI in the upper-left panel of `examples/ROIExamples.py`.

malikolivier:
Up. Is it Ok to merge?

paulmueller:
likely related to #292

j9ac9k:
Closing and Opening to trigger CI run.

j9ac9k:
CI Failure is unrelated to this PR.

j9ac9k:
Closing/Opening to trigger new CI run....

j9ac9k:
I just checked out this issue, I was thinking it might address #292 but it does not appear to do so...I'm still a little puzzled with the intended behavior so I'm going to investigate a bit more with `ImageAnalysis.py` example.

j9ac9k:
Also CI passed for all pipelines

pyqtgraph/graphicsItems/ROI.py
rwalroth: I disagree, makes more sense to mask out coordinates. I implemented a different fix that uses a flag value to mark which coordinates were masked in the returned array, this also takes into account that coords is a 3 dimensional array:
```
def maskCoordinates(coords, mask, flag=-1):
    """
    Applies the mask to the coordinates, setting any 0 values in mask
    to -1.

    Parameters
    ----------
    coords : numpy.ndarray, coordinates returned from getArrayRegion
    mask : numpy.ndarray, mask to be applied
    flag : int, value to represent 0 values in mask

    Returns
    -------
    m_coords : numpy.ndarray, masked coordinates

    """
    m_coords = coords.copy()
    m_coords[0] *= mask
    m_coords[1] *= mask
    m_coords[0][mask == 0] = flag
    m_coords[1][mask == 0] = flag

    return m_coords
```

outofculture:
This had merge conflicts with the recent TriangleROI work, which I resolved on [a copy of this branch](https://github.com/outofculture/pyqtgraph/tree/fix-ROI-getArrayRegion). @malikolivier you can merge from my branch, resolve them yourself, or I can reopen this PR against the copy. However you'd prefer.

This branch behaves identically to master with respect to #292, but I'm in @j9ac9k's boat of being confused about that bug.

This branch definitely fixes #1054 

I couldn't see any other negative impact from this code. Ellipses, triangles and poly ROIs all report seemingly accurate values.

outofculture:
Superceded by #1700

PR #623, opened 2018-01-28
Scatter plot

termim:
In PyQt5 QPainterPath is not hashable anymore which causes SymbolAtlas to fail accept it as a custom symbol, use id instead.
Add corresponding test and an example of using arbitrary text strings as a custom symbols for ScatterPlotItem.

### DIFF
--- a/examples/ScatterPlot.py
+++ b/examples/ScatterPlot.py
@@ -11,6 +11,7 @@ import initExample
 from pyqtgraph.Qt import QtGui, QtCore
 import pyqtgraph as pg
 import numpy as np
+from collections import namedtuple
 
 app = QtGui.QApplication([])
 mw = QtGui.QMainWindow()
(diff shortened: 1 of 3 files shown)


examples/ScatterPlot.py
```diff
@@ -57,21 +58,41 @@ def clicked(plot, points):
...
 
-## 2) Spots are transform-invariant, but not identical (top-right plot). 
-## In this case, drawing is almsot as fast as 1), but there is more startup 
-## overhead and memory usage since each spot generates its own pre-rendered 
+## 2) Spots are transform-invariant, but not identical (top-right plot).
+## In this case, drawing is almsot as fast as 1), but there is more startup
+## overhead and memory usage since each spot generates its own pre-rendered
 ## image.
 
+PlotLabel = namedtuple("PlotLabel", "label symbol scale")
```
campagnola: Was confused by this at first--how about the name `TextSymbol` instead?

termim:
You are right, TextSymbol is better.

campagnola:
Looks good, thanks!

PR #626, opened 2018-01-31
Fix issue # 366

feketeimre:
Set the right texture for rendering, otherwise a previously set texture is overwritten.


campagnola:
I wasn't able to reproduce the problem (which is not surprising; opengl drivers have widely inconsistent implementations). However your explanation makes perfect sense, so: merge!
Thanks for all the detail.

PR #631, opened 2018-02-05
Improve Documentation

Billy4195:
Improve documentation by following changes:
* Use bold font in `mouse_interaction.rst`
* Add `pip3` installation method in `installation.rst`
* Add `Graphics Windows` content in API Reference

### DIFF
--- a/doc/source/apireference.rst
+++ b/doc/source/apireference.rst
@@ -9,6 +9,7 @@ Contents:
     config_options
     functions
     graphicsItems/index
+    graphicsWindows/index
     widgets/index
     3dgraphics/index
     colormap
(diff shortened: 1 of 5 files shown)


campagnola:
I am going to merge your style and installation changes (plus more of my own), but will probably reject the `graphicsWindow` docs since I want to deprecate those classes. I wrote those over a decade ago when it was necessary to manually place widgets inside windows; now (and for many years now) you can just call `show()` on a widget and it will make its own window, so the `graphicsWindow` classes have not been necessary for a long time.

Billy4195:
I got it, but the `graphicsWinodw` is widely used in the examples, would you like to provide a sample code to replace `graphicsWindow` class such that I can help to modify the examples.

campagnola:
Good point! The only window class I see in the examples is `GraphicsWindow`, so you could replace
```
win = pg.GraphicsWindow(title="Scatter Plot Symbols")
```
with
```
win = pg.GraphicsLayoutWidget(show=True, title="Scatter Plot Symbols")
```
(note the `title` argument was just added)

PR #633, opened 2018-02-07
Cylinder now generates the correct amount of sides

feketeimre:
Before the first and the last vertex was overlapping.


campagnola:
Thanks!

PR #634, opened 2018-02-16
Replace deprecate class in examples

Billy4195:
Using class GraphicsLayoutWidget to replace the deprecated class GraphicsWindow, cc #631.


campagnola:
Thanks!

PR #641, opened 2018-03-01
SVGExporter: fix axes position and scale

markotoplak:
Fixes problems with non-aligning axes on Qt5 svg exports.  In the output svg, the axes were (individually) scaled for physical/logical ration.

The export still works well in on Qt4. I suspect there were no problems there, because, at least on my machine (Ubuntu 14.04), logical and physical dpis return the same value.

### DIFF
--- a/pyqtgraph/exporters/SVGExporter.py
+++ b/pyqtgraph/exporters/SVGExporter.py
@@ -169,7 +169,7 @@ def _generateItemSvg(item, nodes=None, root=None):
         buf = QtCore.QBuffer(arr)
         svg = QtSvg.QSvgGenerator()
         svg.setOutputDevice(buf)
-        dpi = QtGui.QDesktopWidget().physicalDpiX()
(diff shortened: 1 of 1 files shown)


sem-geologist:
I can't understand the "internal error" on Travis (PyQt4 Python2.7). For me it looks unrelated with this PR. This is important fix, at current state no working Svg export when using PyQt5 is an outstanding major stopper.

campagnola:
@sem-geologist, you are correct that the "internal" Travis errors have nothing to do with this PR. Usually just restarting the job makes the error go away, and I have not yet been able to fix the underlying problem.

campagnola:
Looks good in my tests. Many thanks, @markotoplak! This one has been on my list for a while.

markotoplak:
Thanks, @campagnola. We have been using pyqtgraph for a while and its my pleasure to contribute.

PR #653, opened 2018-03-29
ScatterPlotItem: Fix a GC memory leak due to numpy issue 6581

ales-erjavec:
Accessing ``ScatterPlotItem.points` causes that instance to persist in memory indefinitely

```python
import gc

from PyQt5.QtWidgets import QApplication
from pyqtgraph import ScatterPlotItem

app = QApplication([])

sp = ScatterPlotItem(x=[1, 2], y=[1, 2], data=[42, 43])
sp.points()  # fill in data["items"]

del sp

gc.collect()

print([o for o in gc.get_objects() if isinstance(o, ScatterPlotItem)])
```

The cause is that `data` numpy array with object fields contains reference cycles back to itself (via a SpotItem._plot), but numpy does not support cyclic gc for object arrays ([numpy issue 6581](https://github.com/numpy/numpy/issues/6581)).

Fixed by taking a weak reference to `plot` in `SpotItem`

### DIFF
--- a/pyqtgraph/graphicsItems/ScatterPlotItem.py
+++ b/pyqtgraph/graphicsItems/ScatterPlotItem.py
@@ -857,10 +857,17 @@ class SpotItem(object):
     def __init__(self, data, plot):
         #GraphicsItem.__init__(self, register=False)
         self._data = data
-        self._plot = plot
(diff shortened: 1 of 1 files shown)


campagnola:
Thanks @ales-erjavec!

PR #657, opened 2018-03-30
When a floating Dock window is closed, the dock is now returned home

edbarnard:
When a Dock is floated by double clicking on the label, it pops into a new window. However when that window is closed the dock is closed rather than restored to its original DockArea. This can be a problem especially for Docks that are not supposed to be closed (and have the closable flag False).
This PR places the floated Dock back to its original area upon TempAreaWindow close.


campagnola:
Looks great! Thanks.

PR #660, opened 2018-04-04
Refactor PlotItem.py and test_GraphicsItem.py

Billy4195:
* Remove some outdated comments
* Extract function to reduce the redundancy


campagnola:
Thanks @Billy4195!

PR #684, opened 2018-05-08
update version for new development and PlotDataItem stepMode fix

danielhrisca:
### Version 
PyPI version is 0.10.0 and uses the release code from november 2016, however the documentation (http://www.pyqtgraph.org/documentation/index.html) has the same version but is based on the development branch. This is very confusing (take for example _addMarker_ method of InfiniteLine).

### PlotDataItem fix
If _stepMode_ is True we get an exception in the _addPoints_ method of _ScatterPlotItem_ since X and Y axis length is not the same. The last X point is in this case excluded.

### DIFF
--- a/pyqtgraph/__init__.py
+++ b/pyqtgraph/__init__.py
@@ -4,7 +4,7 @@ PyQtGraph - Scientific Graphics and GUI Library for Python
 www.pyqtgraph.org
 """
 
-__version__ = '0.10.0'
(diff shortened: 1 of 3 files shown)


danielhrisca:
@campagnola 
did you have time to look over this PR?

campagnola:
Thanks @danielhrisca, I have merged your version changes and resolved the scatter plot bug a different way.

Issue #699, opened 2018-06-01
ImageItem.setData does not handle NaNs properly


polwel:
In `ImageItem.setData`, `autoLevels` will not ignore `NaN`s in the data. While this will not be fatal, drawing the histogram will be incorrect. Also, `HistogramLUTItem.paint` takes a ridiculous amount of CPU time (100 ms per call) when the histogram range includes `NaN`. In itself this probably constitutes a Qt bug.

In the same way, `ImageItem.getHistogram` also fails to account for `NaN` in the image.

polwel:
Thanks for merging my PR. Shall we close the issue?

PR #700, opened 2018-06-01
Fix issue #699

polwel:
See issue #699 for context.

I ditched `numpy.min` and `numpy.max` for their `nanmin` and `nanmax` equivalent.

I suppose this does have a small impact on performance as each element needs to be compared to `NaN`. However, it is largely offset by the gain in robustness.


campagnola:
Looks good to me, thanks! (the one travis issue is unrelated)

PR #716, opened 2018-06-26
Add appveyor CI

campagnola:


test.py
```diff
@@ -19,6 +19,6 @@
 import pyqtgraph as pg
 pg.systemInfo()
 
-pytest.main(args)
+os.exit(pytest.main(args))
```
j9ac9k: should this by `sys.exit` ?

j9ac9k:
Looks like we're going to try and implement azure pipelines and maybe maintain our travis-ci, I think we should close this guy out.

PR #718, opened 2018-07-06
SVGExport: Initialize option.exposedRect for items that use it

ales-erjavec:
Ref: http://doc.qt.io/qt-5/qstyleoptiongraphicsitem.html#exposedRect-var


campagnola:
Hi @ales-erjavec, can you explain what this PR is for, or perhaps provide a code example that demonstrates the effect of the change?

campagnola:
Tests look good on my end.
Thanks @ales-erjavec!

PR #720, opened 2018-07-09
Fix axis autoscaling with disabled mouse axes

andyclegg:
This PR modifies wheelEvent to put None in the scaling array s where the mouseEnabled mask is not set. It further fixes scaleBy, which would crash with a scaling array containing None (at the point where it calls scale=Point(s)). As s is unconditionally turned into a Point later, this can be deferred.


campagnola:
Thanks @andyclegg! We fixed the same bug at the same time, but I'm going to go with mine because it contains a bit of proximal code cleanup as well.

PR #723, opened 2018-07-11
AxisItem: Fix tick elision with self.scale set

klickverbot:
Previously, only the highest level ticks would be displayed
for e.g. `scale == 1e3`.


campagnola:
Thanks @klickverbot!

PR #725, opened 2018-07-12
Fix X axis values when the axis scale is inverted.

speixoto-plaxis:


j9ac9k:
This should be fixed with resolving issue #811  with PR #815.  If you don't feel that's the case, let us know.

PR #727, opened 2018-07-15
InfiniteLine: Fix bounding box for differing x and y ranges

dnadlinger:
Previously, the same pixel -> local transform was used for both
direction, which breaks <del>e.g. for `AxisItem.scale`</del> if the
displayed coordinate range is not the same for both axes. This also
fixes the computation of label coordinates in that situation.

### DIFF
--- a/pyqtgraph/graphicsItems/InfiniteLine.py
+++ b/pyqtgraph/graphicsItems/InfiniteLine.py
@@ -300,28 +300,28 @@ class InfiniteLine(GraphicsObject):
         vr = self.viewRect()  # bounds of containing ViewBox mapped to local coords.
         if vr is None:
             return QtCore.QRectF()
-        
-        ## add a 4-pixel radius around the line for mouse interaction.
-        
-        px = self.pixelLength(direction=Point(1,0), ortho=True)  ## get pixel length orthogonal to the line
-        if px is None:
-            px = 0
-        pw = max(self.pen.width() / 2, self.hoverPen.width() / 2)
-        w = max(4, self._maxMarkerSize + pw) + 1
-        w = w * px
-        br = QtCore.QRectF(vr)
-        br.setBottom(-w)
-        br.setTop(w)
 
-        length = br.width()
-        left = br.left() + length * self.span[0]
-        right = br.left() + length * self.span[1]
-        br.setLeft(left)
-        br.setRight(right)
+        # Compute width of area around line to use for mouse interaction, in
+        # pixels (at least 4 for thin lines).
+        penWidth = max(self.pen.width() / 2, self.hoverPen.width() / 2)
+        padding = max(4, self._maxMarkerSize + penWidth) + 1
+
+        br = QtCore.QRectF()
+
+        orthoPadding = padding * (self.pixelLength(direction=Point(1,0), ortho=True) or 0)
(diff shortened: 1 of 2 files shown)


campagnola:
Hi @klickverbot, your code looks much nicer. Can you give me an example (with minimal code if necessary) of the situation that is fixed by this? Ideally, you or I can write a unit test to cover that case.

j9ac9k:
@dnadlinger Thanks for the PR, sorry we're taking a while to get to it, we're going through a lot of house-keeping and getting our test suite in order has been the priority.

I'll try and get to this in the coming few days; feel free to ping me if I haven't followed up in a week or so.

j9ac9k:
hi @dnadlinger, I attempted to do the conflict resolution and blew up your PR, sorry about that, I'll straighten this out.

j9ac9k:
So the CI system doesn't like this PR...  failure pattern is the oddest I've ever seen.  All Linux builds pass, all Windows builds fail, most macOS builds fail, but some pass.  I'm on macOS and replicated one of the failed builds, and replicated the passed build as well.

I'll dive into this a bit more and see if I can turn anything up.

j9ac9k:
Now Windows images are failing on

```
>       assert not contains(hbr, 37, 185)
E       assert not True
E        +  where True = <function test_InfiniteLine_scaled_bounding_box.<locals>.contains at 0x0000018852A023A0>(<PyQt5.QtGui.QPolygonF object at 0x0000018852A00660>, 37, 185)
```

If I had more time I would get the custom pytest `assertrepr` functionality working so I could at least see what the bounding rect was for these polygons and see how close these points are...

j9ac9k:
CI is about to wrap up, two of the `assert contains(...)` checks are commented out.  I'll see if anyone else has  thoughts on how we should modify them (or remove those checks entirely).

dnadlinger:
No idea, sorry – I just modeled the tests after the existing ones.

j9ac9k:
Hi @dnadlinger figured as much due to thinking about how I would programmatically determine good points to verify.  Wasn't expecting you to come back after 2 years, again, sorry its taken us so long to look into this.  I'm still going to want to merge but I'll think a bit more about relevant test conditions.

dnadlinger:
Yep, it's a bit non-trivial, because ideally you wouldn't want to depend on the same piece of math used to compute the bounding box to verify the result as well.

I thought the scene coordinates should be quite stable, but perhaps there is a platform-dependent offset or something to take into consideration?

j9ac9k:
From my first re-run on this, it's not even platform dependent (which may just be a CI related issue, after all, I'm running these tests on headless systems with virtual displays, who knows how things render), but the test results varied within versions of Qt, so I'm not comfortable w/ even just experimenting until we find known good values, as things could change.

I do want to get this merged at some point, but I'm going to leave it open for now until I can try and come up with a more robust test method.

dnadlinger:
A simple regression test would just be to inspect `_bounds` – this would still show the problem with the old implementation.

dnadlinger:
And yes, experimenting is definitely not the way to go. I was just hoping someone who has the various coordinate transforms in pyqtgraph paged into their working memory could find a more appropriate coordinate space in which to do the comparison.

j9ac9k:
Will be fixed in #2407 🎊

PR #733, opened 2018-07-25
Py2/3 MetaArray adjustments, first pass

pbmanis:
Suggested approach for making metaarray work in Py2 and 3.
This addresses the issue that the ".ma" file contains bytestrings as generated by hdf5 in py3.
Tested under py2.7.14 and py3.6.5 with small test script (on google groups for pyqtgraph).
The PR is against the current "develop" branch on github.com/pyqtgraph


j9ac9k:
This was incorporated as part of #742 Thanks for the PR @pbmanis

PR #738, opened 2018-07-30
Undo regression in wheelEvent of ViewBox

achennu:
Due to a missing line, the scroll wheel event on the axis caused an exception: UnboundLocal error for the variable "mask". Restored from previous commit.


campagnola:
Thanks @achennu, I am going to fix this a slightly different way..

PR #740, opened 2018-08-02
Added clipItem option to LinearRegionItem

achennu:
Added a clipItem option to LinearRegionItem which can be used to limit the rectangle of the region to the clipItem's boundingRect instead of the ViewBoxes. This is useful when placing a linearRegionItem on top of an ImageItem and we do not want the region rectangle to go beyond the image.


achennu:
@campagnola I also tried to see if  the infinite lines could also be limited similarly. I tried tt play with the ``bounds`` parameter, but the docstrings don't specify what coordinates the min, max should be in, etc. 

If that can be done, I think it should be a separate option flag though, because it can actually be useful to have the infinite lines go beyond the image in some circumstances -- linking to labels on edge of viewBox, etc.

j9ac9k:
Closing/Re-Opening to trigger a CI run

j9ac9k:
CI comes back all green, just need to physically test it out (maybe create an example?)

j9ac9k:
Thanks for the PR @achennu.  I like this feature, and I want to merge it but it needs a bit more work.

I tapped into your branch, and am attempting to use it with the "Image Analysis" example, where I added the following code after the ImageItem is added ot the plot.

```python
lr = pg.LinearRegionItem(values=(1.5, 2.5), clipItem=img)
p1.addItem(lr)
```

So this blows up due to the check of the viewbox for LinearRegionItem is the same as the viewbox for the ImageItem, which it is not, as I haven't added the LinearRegionItem to the plot yet.

So I can skirt that issue by adding it to the plot, and manually calling `setClipItem` later.

```python
lr = pg.LinearRegionItem(values=(1.5, 2.5))
p1.addItem(lr)
lr.setClipItem(item=img)
```

Unfortunately it appears I drag the linear region item past the ImageItem.

image

I know it's been quite some time since you made this PR and we completely understand if you've moved on.  But if you could address these issues (and ideally add to the examples the linear region item clipping behavior) I would happily get this merged right away.

Thanks again for the PR!

achennu:
@j9ac9k Thanks for catching that.

It's simple to handle the case when a viewBox for the LinearRegionItem is not yet available.

It has indeed been a long time. I'm not able to easily figure out how to enforce the coordinates for clipping in `boundingRect()`.

Could somebody provide guidance on what the difference in coordinates would be between `self.getViewBox().itemBoundingRect(self.clipItem)` and `self.viewRect()`?

j9ac9k:
Got superseded by #1834  Thanks @achennu for your proposal here!

Issue #747, opened 2018-08-21
HDF5Exporter: throws error when curves aren't the same length


jdchristesen:
When trying to save data from a graph as an hdf5 file, the HDF5Exporter throws and error when you have multiple curves with differing lengths. This looks to be because the numpy.array(data).astype('double') can't handle lists with different lengths. Below is a traceback from the error. This occurs when trying to save data from the "Multiple curves" graph in the "Basic Plotting" example.

````
Traceback (most recent call last):
  File "/home/jchrist/PycharmProjects/lib/python3.5/site-packages/pyqtgraph/exporters/Exporter.py", line 77, in fileSaveFinished
    self.export(fileName=fileName, **self.fileDialog.opts)
  File "/home/jchrist/PycharmProjects/lib/python3.5/site-packages/pyqtgraph/exporters/HDF5Exporter.py", line 55, in export
    fdata = numpy.array(data).astype('double')
ValueError: setting an array element with a sequence.
````

campagnola:
Agreed; this was originally written with the intent of supporting only equal-length plots. For plots of different length, you would have to store each plot in its own dataset. Probably not too difficult to add an exporter option for that?

PR #748, opened 2018-08-21
HDF5Exporter handling of ragged curves

jdchristesen:
Set up the HDF5Exporter to save ragged curves into different datasets based on their names. They are grouped under a datagroup which is named like the dataset for not ragged curves.


ixjlyons:
Just want to add a note that this is a fix for #747. It looks good to me at a glance, but I will try to return and review it more thoroughly soon. Sorry for the wait.

j9ac9k:
Going to close this and merge #1072 thanks for the PR @jdchristesen  and thanks for the review @ixjlyons

PR #755, opened 2018-08-31
Check lastDownsample in viewTransformChanged

### LINKED ISSUE #752: ImageItem: autoDownsample forces qimage recalculation every render
The way autoDownsample is currently implemented forces the entire Image to be downsampled, the lut/levels to be applied and the ARGB and QImage to be recalculated every single time the image is updated. This happens even if the calculated downsample is 1 (ie. no downsampling needed). This leads to using autoDownsample often making drawing images slower than not downsampling. That is kind of counterproductive.

This happens because in viewTransformChanged, the qimage is simply set to None if autoDownsampling is enabled, and an update is forced which calls render. There everything is realculated every time. viewTranformChanged is called everytime the image is moved or zoomed in/out on.

However, since the entire image is downsampled and the LUT applied to it, a new recalculation would only be needed if the downsampling factors change (because of zooming in or out) or if the LUT/levels are changed (which works with a different event pathway).

[Edit] A relatively easy fix is to do the check for the xds, yds in viewTransformChanged and only delete the qimage if that is different from before.

ChristophRose:
Add a check in the viewTransformChanged function to only force a rerender when the downsampling factor changed.
Previously simply moving the image around or zooming in/out without changing the downsampling factor would force a complete rerendering of the image, which was very slow with large images. This way, the expensive rerender is only forced if necessary.

Fixes #752

### DIFF
--- a/pyqtgraph/graphicsItems/ImageItem.py
+++ b/pyqtgraph/graphicsItems/ImageItem.py
@@ -51,6 +51,7 @@ class ImageItem(GraphicsObject):
         self.levels = None  ## [min, max] or [[redMin, redMax], ...]
         self.lut = None
         self.autoDownsample = False
+        self._lastDownsample = (1, 1)
 
         self.axisOrder = getConfigOption('imageAxisOrder')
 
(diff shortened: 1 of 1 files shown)


j9ac9k:
bah, in attempting to update this PR and fix the merge conflict, I messed it up, sorry @ChristophRose  ... going to try and straighten it out right now...

j9ac9k:
CI Passed, looks good to me, thanks for the PR @ChristophRose and sorry it's taken us so long to follow up.

Issue #726, opened 2018-07-13
TreeWidget.topLevelItems is broken on Python3


polwel:
As per the title. The method uses `xrange`, which obviously is not available in Python 3.
I haven't tried it, but I assume the regression was introduced with 6c7e0fa, where an `from ..python2_3 import xrange` was removed from the corresponding file.

I can see two possible fixes for the issue.

1. Change `xrange` to `range`, which is available in both languages.
2. Revert the removal of the `python2_3` compatibility import.

@campagnola Let me know which one you prefer, and I'll submit a PR.

PR #757, opened 2018-09-07
Fixed issue #726

polwel:
I restored the compatibility layer.

Also, I the `weakref` import is not used anywhere in the file.


j9ac9k:
This was fixed with #888

PR #761, opened 2018-09-21
added optional Gradient argument to HistogramLUTItem

behzadba:
In case the image lut was different than gray it can get the same gradient as the imageItem.

### DIFF
--- a/pyqtgraph/graphicsItems/HistogramLUTItem.py
+++ b/pyqtgraph/graphicsItems/HistogramLUTItem.py
@@ -54,7 +54,7 @@ class HistogramLUTItem(GraphicsWidget):
     sigLevelsChanged = QtCore.Signal(object)
     sigLevelChangeFinished = QtCore.Signal(object)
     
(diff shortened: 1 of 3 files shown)


pyqtgraph/graphicsItems/HistogramLUTItem.py
```diff
@@ -69,9 +69,12 @@ def __init__(self, image=None, fillHistogram=True, rgbHistogram=False, levelMode
...
         self.vb.setMouseEnabled(x=False, y=True)
-        self.gradient = GradientEditorItem()
-        self.gradient.setOrientation('right')
-        self.gradient.loadPreset('grey')
+        if gradient==None:
+          self.gradient = GradientEditorItem()
+          self.gradient.loadPreset('grey')
+        else:
+          self.gradient = gradient
+          self.gradient.setOrientation('right')
```
j9ac9k, line 77: I think this needs to be un-indented

j9ac9k:
Hi @behzadba Thanks for the PR, I think I'm going to close this out as this PR contains unrelated commits.  Feel free to re-submit with just the relevant commit(s) and the suggested change.

PR #763, opened 2018-10-05
FIX: Protection over downsampling calculation.

hhslepicka:
This PR adds a couple of checks to avoid OverflowError when using Auto Downsampling.


campagnola:
Looks good, thanks @hhslepicka!

PR #767, opened 2018-10-10
fixed bug in graphicsItems/ImageItem.py: degenerate images (max==min) 

jimofthecorn:
… would raise exception in getHistogram(). Just added two lines:

if mx == mn:
    mx += 1


Swati707:
Thanks for fix.

israel-segundo:
Thanks for the fix, I have successfully tested it with a local modiciation, but I would really like this to be merged with production.

j9ac9k:
Merging this; sorry it took us so long to get to this PR.

Issue #481, opened 2017-05-09
Parameter 'pos' of GLViewWidget.setCameraPosition() is ignored


cklb:
It seems like the mentioned method ignores its first argument `pos` .

As you can see in:
https://github.com/pyqtgraph/pyqtgraph/blob/5195d9dd6308caee87e043e859e7e553b9887453/pyqtgraph/opengl/GLViewWidget.py#L224
all other arguments are handled correctly.

Adding:
```
if pos is not None:
      self.opts['center'] = pos
```
fixes the behavior for me.

If the solution is really that simple, I'm gladly opening a PR for this.

ParisNeo:
Thanks, this work for me too.
I don't want to mess around with the pyqtgraph library so I added a workaround function in my project:
`
def setCameraPosition(w, pos=None, distance=None, elevation=None, azimuth=None):
    if pos is not None:
        w.opts['center'] = pos
    if distance is not None:
        w.opts['distance'] = distance
    if elevation is not None:
        w.opts['elevation'] = elevation
    if azimuth is not None:
        w.opts['azimuth'] = azimuth
    w.update()       
`

j9ac9k:
Closing due to the merging of #770

PR #770, opened 2018-10-12
Fix issue #481

cklb:
The method `setCameraPosition` of the `GLViewWidget` seems the ignore its first argument `pos` (see #481)
By adding the attached lines the position is set correctly.

### DIFF
--- a/pyqtgraph/opengl/GLViewWidget.py
+++ b/pyqtgraph/opengl/GLViewWidget.py
@@ -236,6 +236,8 @@ class GLViewWidget(QtOpenGL.QGLWidget):
                     glPopMatrix()
             
     def setCameraPosition(self, pos=None, distance=None, elevation=None, azimuth=None):
+        if pos is not None:
+            self.opts['center'] = pos
         if distance is not None:
             self.opts['distance'] = distance
         if elevation is not None:


j9ac9k:
Thanks for the PR, looks good

Cocco17:
Hi! Had the same problem with the current version of pyqtgraph 0.10.0 - apparently the change hasn't been merged to release version?  @j9ac9k could you check that please?

cklb:
Hey there, since the next release is not finished the changes are not yet in master.
If you are willing to take the risk however, you can try the current release candidate
`pip install pyqtgraph==0.11.0rc0`
which should include the fix.

j9ac9k:
@Cocco17 We are still gearing up for a stable release (0.11), and as @cklb said, you can install the version with this patch by installing the current rc build.  There are a few more outstanding issues we want to address before we do the next stable release. 

One of the reasons this upcoming release is taking so long is that it will be the last to support qt4/py2, so we're trying to identify all those relevant issues and get fixes in.

Cocco17:
@cklb and @j9ac9k thanks for the quick response! Alright, I'll wait for the 0.11 release, in the meantime I added the missing lines to the setCameraPosition function in the  GLViewWidget.py file

j9ac9k:
We highly recommend installing the rc builds over 0.11,.as it includes hundreds of bug fixes and performance improvments.  Also we would like to know if you run into any issues.  I use the rc for all my current pyqtgraph projects.

Cocco17:
@j9ac9k Okay, I will switch to 0.11 - Thanks!

PR #771, opened 2018-10-14
No warning for arrays with zeros in logscale

2xB:
NumPy evaluates log10(0) to -inf, so there is no reason to show the user a RuntimeWarning. Before, if visualizing data arrays containing zeros in logscale, a RuntimeWarning was shown. Additionally: Removed some invisible whitespace.


j9ac9k:
Looks good, thanks!

Issue #654, opened 2018-03-29
Legend items creeping to the right


reubengann:
My code manages the items in the legend manually. For creation
```
vb = pq.ViewBox()
vb.setMaximumWidth(100)
self.ui.theplts.addItem(vb, row = row, col = 1)
legend = pq.LegendItem()
legend.setParentItem(vb)
```
When a new series is plotted, it calls
```legend.addItem(self.pointerToPlot, self.legendtext)```
and when removing 
```legend.removeItem(self.legendtext)```
After creation, I do not interact with the legend in any other way. This works, but if I repeat this process (adding and removing the same item), the legend item moves to the right a couple of pixels each time. If I print legend.layout.geometry(), this effect can be seen
```PyQt5.QtCore.QRectF(0.0, 0.0, 96.0, 120.0)
PyQt5.QtCore.QRectF(0.0, 0.0, 98.0, 120.0)
PyQt5.QtCore.QRectF(0.0, 0.0, 100.0, 120.0)
PyQt5.QtCore.QRectF(0.0, 0.0, 102.0, 120.0)
PyQt5.QtCore.QRectF(0.0, 0.0, 104.0, 120.0)
PyQt5.QtCore.QRectF(0.0, 0.0, 106.0, 120.0)
PyQt5.QtCore.QRectF(0.0, 0.0, 108.0, 120.0)
```

I suspect the issue lies inside the LegendItem.updateSize function, though I do not immediately see a reason for the behavior I observe.

I am using pyqtgraph v 0.10.0 and PyQt 5.6.2 on Windows 7/Anaconda/Python 3.6.4

reubengann:
I have narrowed this down to an increasing sample width. A workaround is

```
legend.addItem(self.pointerToPlot, self.legendtext)
if legend.size is None:
    height = 0
    width = 0
    for sample, label in legend.items:
        height += max(sample.height(), label.height()) + 3
        width = max(width, 11+label.width())
        legend.setGeometry(0,0,width + 25, height)
```

Where I've substituted a constant (11) for the sample.width().

reubengann:
Issue #578 noted this problem.

NilsNemitz:
I think this should have been fixed by #958.
Does anybody still see these issues?

j9ac9k:
[...]

Issue #625, opened 2018-01-30
LegendItem formatting


falkkyburz:
Hi pyqtgraph contributers, 

As one can examine in the examples, the legend-formatting is not very good. Two issues:
1. Width of sample isn't set/configured correctly.
2. Label text is center justified.

Fix for 2. is easy: Just add justify='left' parameter to LabelItem instantiation call.
The first problem has somehow to do with Qt's way of determining width. Since size is hardcoded anyways, you could use self.setFixedWidth(20) in the constructor of the ItemSample class.

I am no GUI expert!

Before/After of the two modifications is shown below:
![capture](https://user-images.githubusercontent.com/30900572/35569847-a10e3ee6-05cd-11e8-8998-2af9e11e438d.PNG) ![capture](https://user-images.githubusercontent.com/30900572/35569760-55db5558-05cd-11e8-87c7-9ea2b6057b80.PNG)

This is also discussed in #578

NilsNemitz:
I think this should have been fixed in #958.
If anybody is still seeing this issue, please let us know!

j9ac9k:
I think this was indeed solved, if you can still recreate the issue, let us know @falkkyburz and I'll re-open.  Sorry fro the long-time to follow up.

Issue #578, opened 2017-09-28
Need a few spaces between LegendItem symbols and their names


Felix-neko:
When I add a legend to a plot, the legend marks (for example, scatter plot marks) are too close to the names i wrote to them. Cannot you fix it?

https://imgur.com/a/uqJPg — here is it.

P.S. Your library is really cool, it's much better than slow matplotlib = )

gdotmueller:
i had the same problem and i think i found the bug. it is a bug in **legenditem.py line 112 & 113**.
that is in **method updateSize()**.
The same bug also moves the text further away from the symbol each time a legend row is appended!

my Fix:
instead of **sample.height()** and **sample.width()** it should be:
**sample.boundingRect().height()** & **sample.boundingRect().width()**

```
    def updateSize(self):
        if self.size is not None:
            return
            
        height = 0
        width = 0
        #print("-------")
        for sample, label in self.items:
            height += max(sample.boundingRect().height(), label.height()) + 3
            width = max(width, sample.boundingRect().width()+label.width())
            #print(width, height)
        #print width, height
        self.setGeometry(0, 0, width+25, height)
```

i think it was intented to use the boundingRect but it was forgotten to use it here!
when i make these changes it works flawlessy!!

FalkMusick:
Thanks for the tip gdotmueller! 
Additionaly, the legend text should be 'left' aligned instead of the default 'center'.
````
def addItem(self, item, name):
        #label = LabelItem(name)
        label = LabelItem(name, justify='left')
        if isinstance(item, ItemSample):
            sample = item
        else:
            sample = ItemSample(item)        
        row = self.layout.rowCount()
        self.items.append((sample, label))
        self.layout.addItem(sample, row, 0)
        self.layout.addItem(label, row, 1)
        self.updateSize()`

Best Regards
Falk

tomer2141:
I was able to fix it by adding horizontalSpacing for the LegendItem layout
`GraphicsWidget.__init__(self)
        GraphicsWidgetAnchor.__init__(self)
        self.setFlag(self.ItemIgnoresTransformations)
        self.layout = QtGui.QGraphicsGridLayout()
        self.layout.setHorizontalSpacing(15) # custom space between the legend shape and label
        self.setLayout(self.layout)
        self.items = []
        self.size = size
        self.offset = offset
        if size is not None:
            self.setGeometry(QtCore.QRectF(0, 0, self.size[0], self.size[1]))`

[...]

PR #772, opened 2018-10-19
Legend labels should align left with some spacer

behzadba:
As discussed here: #578 #654 #625

### DIFF
--- a/pyqtgraph/graphicsItems/LegendItem.py
+++ b/pyqtgraph/graphicsItems/LegendItem.py
@@ -70,7 +70,7 @@ class LegendItem(GraphicsWidget, GraphicsWidgetAnchor):
         title           The title to display for this item. Simple HTML allowed.
         ==============  ========================================================
         """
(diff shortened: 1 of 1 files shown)


Achilles1515:
If you are adding another item (the spacer) to the layout, then you need to make sure to remove it as well in the LegendItem.removeItem() method. Otherwise, the spacer lingers and continues to use up the legend row when removeItem() is called.

I don't believe a spacer is even necessary. The real issue is in the updateSize() method where sample.width() has the potential to return 0, which messes up all the spacing and pushes the label left into the line icon. Changing from sample.width() to sample.boundingRect().width() seems to fix the issue.

There are a number of other issues with the legend...I may submit a PR later.

j9ac9k:
Closing/Re-Opening to trigger new CI run.

j9ac9k:
Looks like #958 implemented this feature.  Thanks for the PR @behzadba sorry it's taken so long for us to follow up.

PR #776, opened 2018-10-25
Updated according to NumPy FutureWarning

2xB:
FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.


2xB:
Note: It has to be checked if this destroys compatibility with regular python arrays as data.

2xB:
Closes #842.
Closed as #859 does the same.

PR #782, opened 2018-10-29
Terminate FileForwarder thread on process end

### LINKED ISSUE #781: FileForwarder thread doesn't terminate on child process end
On windows, when the child process is ended, the FileForwarder's that are created to forward output don't detect the process is close, and spin, using up a core's worth of CPU, making it impossible to restart remote plot windows on an error.

To see this error:
```py3
proc = mp.QtProcess()
proc.join()
```

then, a CPU core is pinned at 100%.

spauka:
Fixes #781 

Previously on windows the FileForwarder threads would continue to run and eat up a lot of CPU once the child process they were forwarding dies. This commit shuts down those threads when the child process is joined.

### DIFF
--- a/pyqtgraph/multiprocess/processes.py
+++ b/pyqtgraph/multiprocess/processes.py
@@ -166,6 +166,14 @@ class Process(RemoteEventHandler):
                     raise Exception('Timed out waiting for remote process to end.')
                 time.sleep(0.05)
         self.conn.close()
+
+        # Close remote polling threads, otherwise they will spin continuously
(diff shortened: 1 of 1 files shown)


spauka:
Bump ^^

j9ac9k:
Hi @spauka thanks for the PR; I'll likely look into this in the coming days.  Can you give me a code example that I can easily recreate your issue and check against your fix?

Thanks

spauka:
The following code snippet will reproduce the error on my PC:

```py3
proc = mp.QtProcess()
proc.join()
```

Issue #792, opened 2018-12-02
Wrong default value for scale parameter in makeARGB


titusjan:
In the doc string of `makeARGB` it says about the `scale` parameter that: _By default this will be set to the length of the lookup table._ However since commit e2f43ce4 this no longer the case. The scale is now set to the length of the lookup table - 1. 

> scale = lut.shape[0] - 1

I think this is wrong and the following code demonstrates that.

```
import numpy as np
import pyqtgraph as pg

lut1 = np.array([(0, 0, 0), (1, 1, 1), (2, 2, 2)])

vector = np.linspace(0.0, 1.0, 11)
data = np.expand_dims(vector, axis=0)
print("data: {}".format(data))

levels = np.amin(data), np.amax(data)

for scale in [None, 2, 3]:
    res, _ = pg.makeARGB(data, lut1, levels=levels, scale=scale)
    print("\nScale: {}".format(scale))
    for idx, val in enumerate(vector):
        color = res[0, idx, :]
        print("    Value {:5.2f} ->  RGBA {} ".format(val, color))
```

This applies a look up table with three colors to an array that ranges from 0.0 to 1.0 with steps of 0.1. It does this three times, first with the default `scale` parameter, then with `scale = 2` and then with `scale = 3`. 
```
Scale: None
    Value  0.00 ->  RGBA [  0   0   0 255] 
    Value  0.10 ->  RGBA [  0   0   0 255] 
    Value  0.20 ->  RGBA [  0   0   0 255] 
    Value  0.30 ->  RGBA [  0   0   0 255] 
    Value  0.40 ->  RGBA [  0   0   0 255] 
    Value  0.50 ->  RGBA [  1   1   1 255] 
    Value  0.60 ->  RGBA [  1   1   1 255] 
    Value  0.70 ->  RGBA [  1   1   1 255] 
    Value  0.80 ->  RGBA [  1   1   1 255] 
    Value  0.90 ->  RGBA [  1   1   1 255] 
    Value  1.00 ->  RGBA [  2   2   2 255] 

Scale: 2
    Value  0.00 ->  RGBA [  0   0   0 255] 
    Value  0.10 ->  RGBA [  0   0   0 255] 
    Value  0.20 ->  RGBA [  0   0   0 255] 
    Value  0.30 ->  RGBA [  0   0   0 255] 
    Value  0.40 ->  RGBA [  0   0   0 255] 
    Value  0.50 ->  RGBA [  1   1   1 255] 
[...]

PR #793, opened 2018-12-02
Fixes incorrect default value for scale parameter in makeARGB.

titusjan:
Fixes issue #792

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1073,7 +1073,7 @@ def makeARGB(data, lut=None, levels=None, scale=None, useRGBA=False):
     # Decide on maximum scaled value
     if scale is None:
         if lut is not None:
-            scale = lut.shape[0] - 1
+            scale = lut.shape[0]
         else:
             scale = 255.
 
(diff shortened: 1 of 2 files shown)


j9ac9k:
CI fails on this PR as it currently stands:

https://dev.azure.com/pyqtgraph/pyqtgraph/_build/results?buildId=259

campagnola:
Looks good to me; I agree with @titusjan's assessment in #792. Unit tests in `pyqtgraph/tests/test_functions.py : test_makeARGB()` will need to be updated to reflect the change, and we definitely want a CHANGELOG entry noting the behavior change.

Dschoni:
https://dev.azure.com/pyqtgraph/pyqtgraph/_build/results?buildId=347

j9ac9k:
I fixed the test cases so they now pass; not exactly sure what should be stated in `CHANGELOG` specifically.  I'm going to leave the CHANGELOG update to be handled by #1005

titusjan:
Sorry for my lack of response on this and thanks for fixing the tests.

j9ac9k:
Don't sweat it, thanks for the PR and the analysis you did in #792

Dschoni:
Does that indeed fix #792? As it is still open. If so, I would put exactly that in CHANGELOG.

j9ac9k:
Yeah I meant to close it, my mistake.

PR #801, opened 2018-12-21
Remove use of GraphicsScene._addressCache in translateGraphicsItem

ales-erjavec:
Use `QGraphicsItem.toQGrapicsObject` on the item instead.

(This probably is not even needed since PyQt 4.9)

The problem with `_addressCache` is that it can map a plain QGraphicsItem that was not 'registered' with a `registerObject` to an invalid (already deleted on the C++ side) wrapper object. For instance:

1. An item registered with registerObject and added to a scene.
2. The item is deleted from the C++ side, but the PyQt wrapper remains live (its entry remains in `_addressCache`).
3. A plain QGraphicsItem that is not registered with `registerObject` is created that reuses the memory that the first item occupied (it is located at the same address) and is added to the scene.
Then calling `scene.items()` will replace the new item with a wrapper for the old invalid item in the output.

I have used `item.toQGraphicsObject()` in the past to workaround the same bug (I am pretty sure it is no longer needed since PyQt 4.9)

### DIFF
--- a/pyqtgraph/GraphicsScene/GraphicsScene.py
+++ b/pyqtgraph/GraphicsScene/GraphicsScene.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 import weakref
+import warnings
+
 from ..Qt import QtCore, QtGui
 from ..Point import Point
 from .. import functions as fn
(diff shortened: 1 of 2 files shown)


j9ac9k:
Pretty sure I've actually run into this exact issue on some of my projects, which my quick workaround was to just remove the names from my objects.  I'm going to try this diff on my project and see if it addresses the issue.

Thanks for submitting this PR @ales-erjavec sorry it's taken us so long to get to this.

j9ac9k:
I'm going to close and re-open this issue to trigger a new CI run.

j9ac9k:
CI came back all green, I'm inclined to merge this, will solicit input from other maintainers.

Thanks for the PR @ales-erjavec sorry it's taken us so long to get to this.

PR #809, opened 2019-01-14
Fix selection of FlowchartWidget input/output nodes from issue #808

### LINKED ISSUE #808: Selecting flowchart input or output nodes in Flowchart widget gives KeyError
The Flowchart widget has a flowchart visualisation and a tree view. The input and output nodes are in the visualisation within FlowchartWidget but not the tree of FlowchartCtrlWidget. When the user clicks on the input or output node to select it in the FlowchartWidget, an attempt it made to highlight it in the FlowchartCtrlWidget. This results in a KeyError, and therefore stops the data on those nodes being displayed in the FlowchartWidget.

To reproduce, run the Flowchart example: _examples/Flowchart.py_. Click Flowchart in the first window, and then click on the input or output node in the flow chart of the second window. A KeyError is raised.

christuart:
Fixes #808.

Whenever they were clicked on, FlowchartWidget was trying to highlight the input and output nodes in the tree widget FlowchartCtrlWidget. These nodes are not put into the tree, so a KeyError was being raised and stopped the information about these nodes from being populated in FlowchartWidget.

FlowchartWidget now always checks if a node is found in FlowchartCtrlWidget before calling 'select' on that node. If it is not found, it calls a new method of FlowchartCtrlWidget to clear the selection in the tree. In this way, the selection in the tree does not become out of sync with the selection in FlowchartWidget.


j9ac9k:
Hi @christuart thanks for the PR, sorry about the long delay on this, the change looks good to me, I'll go ahead and merge.

PR #813, opened 2019-01-16
AxisItem performance improvements

danielhrisca:
### Avoid calling _setLabel_ repeatedly on AxisItem
* add _setLabel_ arguments to __init__
* in _setLabel_ only call _showLabel_ once if either _text_ or _units_ have changed

For example in the case of creating 400 AxisItem objects and customizing the axis label and units, the _setLabel_ was previously called 3300 times, and with this implementation the call count went down to 920.

### improve \_\_init\_\_ speed
call _showLabel(False)_ earlier to avoid unnecessary _setLabel_ calls via _setRange->updateAutoSIPrefix_


j9ac9k:
Thanks for the PR, looks good!

PR #814, opened 2019-01-16
enforce enableMenu in ViewBox

danielhrisca:
enforce _enableMenu_ when False:
* set _menu_ attribute to None at __init__
* populate _menu_ as necessary is the state is changed

For example in case of 400 ViewBox objects, the time spent creating the _menu_ attributes was 1.2s. If no menu is needed then this can improve the performance.


j9ac9k:
Looks good to me.

Issue #811, opened 2019-01-15
"X Axis --> Invert Axis" unexpected behavior


ZanSara:
This issue was found in the pyqtgraph examples. I refer specifically to the plots in the "Basic Plotting" example.

By right-clicking on the plot area, in the context menu there are the entries "X Axis" and "Y Axis". In each of their submenus there is an entry labeled "Invert Axis". I personally expected it to mirror the values on the axis. When ticking "Y Axis --> Invert Axis", this is exactly what happens: values on the axis are mirrored, the labels are updated and the plot gets mirrored too. However, when ticking "X Axis --> Invert Axis" the plot gets mirrored correctly, but the X Axis labels are not affected, which is very confusing and looks like a bug. In addition, the Y Axis range gets changed into the X Axis one. For example, in the "Multiple curves" plot, the Y Axis range changes from [0, 10] into [0, 100]. Note that only the labels are changing, because the plot is not resized. This looks purely like a labeling issue.

I attach some screenshot to clarify my point. It should be easy to reproduce though.

"Expected" behavior: changes the Y Axis labels from [0, 10] to [10, 0] and mirrors the plot along the horizontal axis.
![pyqtgraph-yaxis](https://user-images.githubusercontent.com/5703634/51182723-5aab8d80-18cf-11e9-9fc8-7415318bdd5b.png)

"Unexpected" behavior: leave the X Axis labels unchanged, changes Y Axis labels, and mirrors the plot along the vertical axis.
![pyqtgraph-xaxis](https://user-images.githubusercontent.com/5703634/51182721-5a12f700-18cf-11e9-8b7c-74a0656a6a5b.png)

I could not find any reference to this behavior and I am not experienced with pyqtgraph either, so if there is any simple fix please let me know. I'm running pyqtgraph==0.10.0 (from PyPI), Python 3.6.6, PyQt5==5.11.3. 

------------

EDIT: Apparently the labels are redrawn correctly as soon as the plot is redrawn. For example, the "Updating plot" in this same window does not display the issue. For the other plots, panning or zooming is enough to fix the issue. I guess this can be fixed quickly, will check.

ZanSara:
Here is it:
https://github.com/pyqtgraph/pyqtgraph/blob/2e69b9c5e6bd8008c2d7e9f384b9c9b32ff2f5a7/pyqtgraph/graphicsItems/ViewBox/ViewBox.py#L1017
[...]

PR #815, opened 2019-01-17
Fix issue #811

ZanSara:


HappyTreeBeard:
This change looks good to me. I confirmed this does fix the problem described in #811.

j9ac9k:
Looks good to me as well.

PR #816, opened 2019-01-18
Avoid constructing shadow pens when no shadow pen is set

fake-name:
Basically, profiling has pointed me to the fact that a fair bit of code time (3%, actually) is spent in `setShadowPen()` (actually, it's in `mkPen()`, which `setShadowPen()` calls), even when no shadow pen is specified.

In my application, I'm calling `pyqtgraph.PlotDataItem.setdata()`, which calls through PlotDataItem->setData, PlotDataItem->updateItems. At some point in the call stack, the default value for `shadowPen` is being inserted into the kwargs, which then causes the specious calling of setShadowPen.

Anyways, if we check if shadowPen is a value other then none, this doesn't happen.


j9ac9k:
if we will do these kinds of checks, any reason not to do instead, so instead of checking against `None` we are actually ensuring that the pen is the correct type?

```python
isinstance(kargs.get("shadowPen"), QtGui.QPen)
```

fake-name:
I don't see why that wouldn't work just as well.

The issue I encountered is that the `self.setShadowPen(kargs['shadowPen'])` call is *expensive*, even if it's called with `None` as it's argument. If you're plotting lots of data, it's a non-trivial overhead that's trivial to remove. 

Anything that avoids calling `setShadowPen()` would fix the issue.

j9ac9k:
Hi @fake-name 

First of all, I should apologize, I only looked over the PR briefly earlier and I didn't thank you for your submission.  Thanks for the PR!  Second, I didn't take enough time to adequately indicate what I was trying to do with my suggested change, so sorry about that.

Currently, there is an if check that checks if there is a key `shadowPen` in `kwargs` and if so, regardless of what the value of that dictionary is, to call `setShowPen(kwargs['shadowPen'])`.  You point out that this call, even when the value is `None` is expensive.

You're proposing that in addition for checking that the key is present in `kwargs`, we don't call `setShadowPen` if the value of `kwargs['shadowPen']` is `None`.

What I was proposing is we ensure that the value in `shadowPen` is actually an instantiated `QPen` object, and if so, then call `setShadowPen`.   From looking at the rest of the code, It doesn't make sense to call `setShadowPen` with anything other than aa `QPen` object as an argument. 

The difference in what I'm suggesting is that it enforces the type check; you're not just passing values that aren't `None` to `setShadowPen`, you're only passing values that are of the correct type.  Doing it this way in my opinion addresses the issue most clearly, don't call `setShadowPen` unless you have a valid argument to pass to it.

```python
if isinstance(kwargs.get('shadowPen'), QtGui.QPen):
   self.setShadowPen(kwargs['shadowPen'])
```

If you wanted to keep going with that route, might as well do `kwargs.get('shadowPen') is not None` instead, which would be one boolean evaluation, not two.

Lastly, I apologize for us taking so long to get back to you on this, development of pyqtgraph is picking up again, and our first order of business has been restoring the test suite and getting our CI system to pass.  Performance is very important to pyqtgraph, but we've been tending to the broken things for the most part so far.

Let me know if I'm missing something or if something I said didn't make sense.

fake-name:
@j9ac9k - Your summary pretty much is what I assumed you meant. It does seem to me assuming that `shadowPen` is present in `kwargs` is not totally ideal (It seems like it'd be possible to call `PlotCurveItem.updateData()` directly if you really tried, and assuming a set of kwargs that aren't enumerated in the declaration makes me uncomfortable), but that's a flavor thing.

> Lastly, I apologize for us taking so long to get back to you on this, development of pyqtgraph is picking up again, and our first order of business has been restoring the test suite and getting our CI system to pass. Performance is very important to pyqtgraph, but we've been tending to the broken things for the most part so far.

Not a big deal. I basically patched my PyQtGraph version, shipped the vendored library, and moved on.

j9ac9k:
> It does seem to me assuming that `shadowPen` is present in `kwargs` is not totally ideal 

`dict.get()` makes no assumption about a key being present, if the key is not present, it will return `None` (not raise a `KeyError`)

fake-name:
> `dict.get()` makes no assumption about a key being present, if the key is not present, it will return None (not raise a KeyError)

Well, huh. I could have sworn you had to pass a default to avoid the keyerror. 

I learn something new every day.

j9ac9k:
> I learn something new every day.

That's how you know you're doing it right.

If you make the `isinstance()` change, and resolve the conflict, and CI doesn't blow it up, I'll merge the PR.

Thanks again for the PR!

fake-name:
Ok, digging into the code, I think checking if the `shadowPen` is a `QtGui.QPen` isn't the right solution. 

The value passed to `self.setShadowPen()` is forwarded to `functions.mkPen(*args, **kargs):`, which supports a large variety of argument formats (dict, QPen, color-string). 

I think, given the variety of ways this could potentally be called, `if kargs.get("shadowPen") is not None:`  is about as good as it can get.

It's less then ideal, since even `None` is a theoretically valid value:

```    
    if len(args) == 1:
        arg = args[0]
        if isinstance(arg, dict):
            return mkPen(**arg)
        if isinstance(arg, QtGui.QPen):
            return QtGui.QPen(arg)  ## return a copy of this pen
        elif arg is None:
            style = QtCore.Qt.NoPen
        else:
            color = arg
    if len(args) > 1:
        color = args
        
    if color is None:
        color = mkColor('l')
    if hsv is not None:
        color = hsvColor(*hsv)
    else:
        color = mkColor(color)
```

You call `setShadowPen(None)` to clear a shadow pen, so it's theoretically possible for someone to *intentionally* pass `None` as the `kargs["shadowPen"]` value.

On the other hand, supporting that might be a esoteric enough case that we don't feel it's worth the overhead during normal plotting.

j9ac9k:
Oh, I ran into an issue with `functions.mkPen` taking `True` but not `False` or something like that a few weeks back...granted that's an unrelated issue. 

Let me investigate what all `mkPen` takes, if it's indeed a long list of stuff (3-tuples, QPen's, booleans, etc etc), yeah let's filter for `None` and call it a day.  If the list is reasonably short, we can have `isinstance(kwargs.get('shadowPen', boolean, QPen, ...)`

fake-name:
Rolled into upstream, but the merge is being really messy for some reason.

At some point, *more* invocations to mkPen were added, [so I put checking around those too](https://github.com/pyqtgraph/pyqtgraph/pull/816/commits/8b6466c7e924689f35986eb113c3a7001cdae859#diff-b064b448cfe1000597d68d33cc5781b7R497).

j9ac9k:
> Rolled into upstream, but the merge is being really messy for some reason.
> 
> At some point, _more_ invocations to mkPen were added, [so I put checking around those too](https://github.com/pyqtgraph/pyqtgraph/pull/816/commits/8b6466c7e924689f35986eb113c3a7001cdae859#diff-b064b448cfe1000597d68d33cc5781b7R497).

Oof, that's one head of a diff, 142 commits; you may want to undo that merge; fetch upstream, checkout develop branch, pull the upstream develop branch, checkout your branch, and rebase develop, ...that will likely clear things up (I've been there with blowing up my git branches though, it happens!)

fake-name:
Well, that was a mess.

j9ac9k:
Thanks for the update and cleaning up that PR, always a nice feeling when you can undo a git mess.

I'm going to have to look at this a bit more carefully in the next day or so.  I have some (likely unfounded) concerns, but as I'm still getting familiar with the library, I should test this out and ensure things are working as intended.

On your end, especially if this has been a major performance bottleneck, I would make sure that this effectively addresses your problem.

Thanks again for the PR and the update.

fake-name:
> I'm going to have to look at this a bit more carefully in the next day or so. I have some (likely unfounded) concerns, but as I'm still getting familiar with the library, I should test this out and ensure things are working as intended.

It would 100% not surprise me if there's a major derp on my part in there *somewhere*.

> On your end, especially if this has been a major performance bottleneck, I would make sure that this effectively addresses your problem.

It was never a major performance anything. It was more "I'm trying to plot 56 4000 point lines that update at 10 hertz, what can I do to make things go faster", and this was a particular low hanging fruit.

IIRC, the profiler reported spending ~3% in the mkPen calls, and since I wasn't using them, it was a easy tweak to make.

I don't actually even have the hardware I'd need to run the associated code (it was a realtime graphing frontend for some custom hardware).

Dschoni:
https://dev.azure.com/pyqtgraph/pyqtgraph/_build/results?buildId=346

j9ac9k:
Resolving merge conflict, re-running CI.

j9ac9k:
CI passed, merging.  Thanks for the PR @fake-name sorry it's taken me so long to get to this.

PR #817, opened 2019-01-18
avoid double call to mkPen when creating PlotCurveItem objects

danielhrisca:
Avoid double call to _mkPen_ if _'pen'_ keyword argument is provided


j9ac9k:
LGTM, Thanks for the PR!

PR #797, opened 2018-12-18
ensure QPoint.__sub__ doesn't get called

pijyoi:
In PySide2 5.12.0, the following code snippet: 
```python
import pyqtgraph
import PySide2
from PySide2.QtCore import QPoint

print(PySide2.__version__)

x = QPoint()
y = pyqtgraph.Point()
z = x - y
```
gives the following error output:
```text
5.12.0
Traceback (most recent call last):
  File "test_qpoint.py", line 9, in <module>
    z = x - y
TypeError: 'PySide2.QtCore.QPoint.__sub__' called with wrong argument types:
  PySide2.QtCore.QPoint.__sub__(Point)
Supported signatures:
  PySide2.QtCore.QPoint.__sub__(PySide2.QtCore.QPoint)
```

Previously, (tested on 5.11.2) Point.\_\_rsub\_\_ method was getting invoked.
This fix ensures that Point.\_\_sub\_\_ gets invoked rather than QPoint.\_\_sub\_\_

PR #818, opened 2019-01-18
Fixed incorrect evaluation of Point in mouseMoveEvent.

mliberty1:
ev.pos() is a QT Point, but self.lastMousePos was converted to pyqtgraph Point.


ksunden:
Duplicate of #797

j9ac9k:
Closing, as this should have been fixed with #824 

Let us know if that wasn't the case.

PR #823, opened 2019-01-31
Added PySide2 support

justengel:
PySide2 now works with pyqtgraph. It looks like there are a lot of deprecation warnings for numpy. PySide2 also seems to have trouble with similar types. QPoint() - QPointF() raises a TypeError. QMatrix4x4(QMatrix4x4()) does not work as well, but would work for QMatrix4x4(QMatrix()). I was able to get all of the examples running for PySide2.

Also be careful when installing PyQt4, PyQt5, and PySide in a Python environment that has PySide2. PySide2 seems to raise odd errors when another Qt library is installed.


justengel:
I think I need to updated to the latest version before I do the pull request.

PR #824, opened 2019-02-01
Fix for PySide2; QtCore.QPoint.__sub__ no longer works with tuples

ronpandolfi:
Otherwise, hovering over GraphicsViews produces error:
```
  File "C:\Users\rp\PycharmProjects\pyqtgraph\pyqtgraph\widgets\GraphicsView.py", line 365, in mouseMoveEvent
    delta = Point(ev.pos() - self.lastMousePos)
TypeError: 'PySide2.QtCore.QPoint.__sub__' called with wrong argument types:
  PySide2.QtCore.QPoint.__sub__(Point)
Supported signatures:
  PySide2.QtCore.QPoint.__sub__(PySide2.QtCore.QPoint)
```

### DIFF
--- a/pyqtgraph/widgets/GraphicsView.py
+++ b/pyqtgraph/widgets/GraphicsView.py
@@ -362,7 +362,7 @@ class GraphicsView(QtGui.QGraphicsView):
     def mouseMoveEvent(self, ev):
         if self.lastMousePos is None:
             self.lastMousePos = Point(ev.pos())
-        delta = Point(ev.pos() - self.lastMousePos)
+        delta = Point(ev.pos() - QtCore.QPoint(*self.lastMousePos))
         self.lastMousePos = Point(ev.pos())
 
         QtGui.QGraphicsView.mouseMoveEvent(self, ev)


tacaswell:
Would it be better store `lastMousePos` as the Qt type to avoid the back-and-forth conversion?

It looks like this is only used in this module:

```
jupiter@18:08  ➤  git grep lastMousePos
pyqtgraph/widgets/GraphicsView.py:        self.lastMousePos = None
pyqtgraph/widgets/GraphicsView.py:        self.lastMousePos = Point(ev.pos())
pyqtgraph/widgets/GraphicsView.py:        if self.lastMousePos is None:
pyqtgraph/widgets/GraphicsView.py:            self.lastMousePos = Point(ev.pos())
pyqtgraph/widgets/GraphicsView.py:        delta = Point(ev.pos() - self.lastMousePos)
pyqtgraph/widgets/GraphicsView.py:        self.lastMousePos = Point(ev.pos())
```

on the other hand people may be grabbing this and using it so this is the more conservative option.

tacaswell:
restarted the failed test to see if it is something transient...

ronpandolfi:
@tacaswell, fully agreed; Qt types could have been used, but lets not change it now. 

The appveyor tests haven't been working for some time, I think.

campagnola:
Thanks @ronpandolfi ! 

What version of pyside2 are you using? I was not able to reproduce this, and I wonder what else might be broken at the same time..

ronpandolfi:
@campagnola, The issue still seems present on latest pipable PySide2 version, `5.12.3`. Example (`Point` is `pyqtgraph.Point`):

```
In [5]: QPoint(0,0)-Point(0,0)                                                                                                      
TypeError                                 Traceback (most recent call last)
<ipython-input-5-f7a3b0083e74> in <module>
----> 1 QPoint(0,0)-Point(0,0)

TypeError: 'PySide2.QtCore.QPoint.__sub__' called with wrong argument types:
  PySide2.QtCore.QPoint.__sub__(Point)
Supported signatures:
  PySide2.QtCore.QPoint.__sub__(PySide2.QtCore.QPoint)

```
compared with PyQt5:
```
In [6]: QPoint(0,0)-Point(0,0)                                                                                                      
Out[6]: Point(0.000000, 0.000000)
```

ronpandolfi:
@campagnola: Actually, the following works with PySide2:
```
In [8]: QPointF(0,0)-Point(0,0)                                                                                                     
Out[8]: Point(0.000000, 0.000000)
```
Since `Point` subclasses `QPointF`, this makes sense. PySide2 is just a bit less forgiving than PyQt5, and doesn't accept operations between `int`'s and `float`'s.

PR #832, opened 2019-02-18
Bug fix in test_csv.py

HappyTreeBeard:
Fixed bug where the temp file ```test_CSVExporter``` generates remains open and ```os.unlink``` raises a ```PermissionError```. This commit closes the TextIO connection after the data is read.


j9ac9k:
Thanks for this PR! A lesson to always use the context manager!

PR #836, opened 2019-03-03
tickSpacing bug fix

SamSchott:
Fixed a bug where `tickSpacing` would return `None` if `style['maxTickLevel'] < 2`, resulting in the axis not being drawn.


j9ac9k:
Thanks for the PR, LGTM, merging!

PR #839, opened 2019-03-09
fix a misplaced )

long-pham:
This pull request fixes a misplaced parenthesis that cause exception when mouse moves over a plot


Birdulon:
Current version of PySide2 requires this change to interact with plots. Earlier versions and PyQt5 didn't seem to complain about the implicit cast but this change should make it more correct regardless.

ksunden:
Duplicate of #797

j9ac9k:
Closing, as this should have been fixed with #824 

Let us know if that wasn't the case.

PR #840, opened 2019-03-14
Clipping: don't assume that x-values have uniform spacing

SamSchott:
Do not assume that x-values have uniform spacing -- this can cause problems especially with large datasets that have non-uniform x-value spacing.

Use `np.searchsorted` instead to find the first and last data index in the view range. This only assumes that x-values are in ascending order and prevents potentially too strong clipping.


j9ac9k:
Thanks for the PR, I love this feature. My main concern is the performance hit.  Let me do some testing on my end.

SamSchott:
Unfortunately, I expect that there will be a significant performance hit if the the x-axis range is clipped close to the middle (`searchsorted` uses a binary search). Maybe it makes more sense to calculate the clipping indices the old way, with O(1) performance, and only refine if this fails. Something similar to:

```python
    dx = float(x[-1]-x[0]) / (len(x)-1)
    x0 = np.clip(int((range.left()-x[0])/dx)-1*ds, 0, len(x)-1)
    x1 = np.clip(int((range.right()-x[0])/dx)+2*ds, 0, len(x)-1)

    if x[x0] > range.left():
        x0 = np.searchsorted(x, range.left()) - 1*ds
        x0 = np.clip(x0, a_min=0, a_max=len(x))
    if x[x1] < range.right():
        x1 = np.searchsorted(x, range.right()) + 2*ds
        x1 = np.clip(x1, a_min=0, a_max=len(x))

    x = x[x0:x1]
    y = y[x0:x1]
```

The code is less elegant but will preserve performance for uniformly spaced data while preventing errors otherwise.  What do you think?

j9ac9k:
I think going to `searchsorted` in case there is a failure the way you show makes the most sense, I would approve (but I'll ask one of the other members to take a look over as well as I'm new in these parts).

Given you have PR #841 that also has these changes, should we migrate that discussion there?

SamSchott:
Ah, that was a mistake. Somehow, the code from other pull requests got all combined in PR #841. I intended to keep separate PRs for separate issues.

j9ac9k:
@SamSchott thanks for updating your PRs.  Also sorry we've taken an eternity to follow up with you.  You may or may not have noticed, but there are some new members to this repo (myself included) and we're trying to patch up our test suite/CI system first before we look at much of anything else...

SamSchott:
No worries, I much appreciate the renewed activity! Some of my PRs are purely for style / personal taste but others are for functionality, such as #836 and #840. So we have ended up relying on our own forked version for some time now...

j9ac9k:
> No worries, I much appreciate the renewed activity! Some of my PRs are purely for style / personal taste but others are for functionality, such as #836 and #840. So we have ended up relying on our own forked version for some time now...

I'll admit for a while I had a `Makefile` that patched my pyqtgraph install for my application, luckily for my use case this hasn't been necessary any more.  We have addressed most issues in testing (some issues still are evasive, including a segfault on test tear-down), and some issues with our remote plotting example.

If there is other functionality that your forked version has that you'd like to get embedded into the main library, open up and issue with the `[Proposal]` tag and make a case for it, we'll give it a look over.

We do think 0.11 release is around the corner (once we patch up our test suite, address python 3.8 deprecations, and expand our documentation).

Thanks again for the PR.

j9ac9k:
@SamSchott Did you still want to incorporate the O(1) check first before implementing `np.searchsorted`?

SamSchott:
Done.

j9ac9k:
Thanks! I'll try and review this later today.

j9ac9k:
LGTM thanks for the PR including the test; merging

PR #845, opened 2019-03-22
Bugfix: ValueError for histogram of pictures with all pixels of the same value

mschu:
When mx == mn the following Exception was thrown:
"pyqtgraph\graphicsItems\ImageItem.py", line 497, in getHistogram
    bins = np.arange(mn, mx+1.01*step, step, dtype=np.int)
ValueError: arange: cannot compute length

Fixed by handling this edge case


j9ac9k:
Should be resolved with merging #767`

Issue #846, opened 2019-03-28
Parameter: hasDefault() always returns true after initializing a new Parameter even without passing in default value (and potential fix for this in comments)


colehurwitz:
I recently ran into an issue when trying to instantiate Parameter objects with and without default parameters. I include the description of the issue and a potential solution below.

I used these two dictionaries to create new Parameter objects:

```python
{'name': 'stick', 'type': 'int', 'value':1., 'default':None, 'title': "stick"},
{'name': 'file_name', 'type': 'str', 'title': "filename"}
```

When I created new Parameters with these dictionaries, both Parameters returned True for hasDefault(). I looked into the code of Parameter and found these two snippets:

```python
def __init__(self, **opts):
     .
     .
     .
    if 'default' not in self.opts:
              self.opts['default'] = None
              self.setDefault(self.opts['value'])
```

```python
def hasDefault(self):
        """Returns True if this parameter has a default value."""
        return 'default' in self.opts
```

In the initialization it sets self.opts['default'] to None if there is no default passed in so _hasDefault will always return True regardless of if a default is given by the user._

I would modify this behavior (if this is unintended) by adding a boolean flag like this:

```python
def __init__(self, **opts):
     .
     .
     .
     if 'default' not in self.opts:
          self.has_default = False
```

```python
def hasDefault(self):
        """Returns True if this parameter has a default value."""
        return self.has_default
```

```python
def setDefault(self, val):
        """Set the default value for this parameter."""
        if  'default' not in self.opts:
            self.opts['default'] = val
            self.has_default = True
            self.sigDefaultChanged.emit(self, val)
       elif self.opts['default'] == val:
            return
       else:
            self.opts['default'] = val
            self.sigDefaultChanged.emit(self, val)
```

There would have to be some more changes since some other functions expect a default value. If you would like, I would be happy to add this in a branch and make a pull request.

PR #847, opened 2019-03-29
Parameter: hasDefault, setDefault, and other default functions behaviour fixed

colehurwitz:
Closes: #846 

This changes the behaviour of 'default' in the Parameter class. Basically, if you don't pass in a default value, then Parameter.hasDefault returns False. This will be set to Parameter.hasDefault will return True when the default value is changed (setDefault).

### DIFF
--- a/pyqtgraph/parametertree/Parameter.py
+++ b/pyqtgraph/parametertree/Parameter.py
@@ -21,31 +21,31 @@ def __reload__(old):
 class Parameter(QtCore.QObject):
     """
     A Parameter is the basic unit of data in a parameter tree. Each parameter has
-    a name, a type, a value, and several other properties that modify the behavior of the 
+    a name, a type, a value, and several other properties that modify the behavior of the
     Parameter. Parameters may have parent / child / sibling relationships to construct
     organized hierarchies. Parameters generally do not have any inherent GUI or visual
     interpretation; instead they manage ParameterItem instances which take care of
     display and user interaction.
-    
-    Note: It is fairly uncommon to use the Parameter class directly; mostly you 
+
+    Note: It is fairly uncommon to use the Parameter class directly; mostly you
     will use subclasses which provide specialized type and data handling. The static
     pethod Parameter.create(...) is an easy way to generate instances of these subclasses.
-       
+
     For more Parameter types, see ParameterTree.parameterTypes module.
-    
+
     ===================================  =========================================================
     **Signals:**
-    sigStateChanged(self, change, info)  Emitted when anything changes about this parameter at 
+    sigStateChanged(self, change, info)  Emitted when anything changes about this parameter at
                                          all.
-                                         The second argument is a string indicating what changed 
+                                         The second argument is a string indicating what changed
                                          ('value', 'childAdded', etc..)
-                                         The third argument can be any extra information about 
+                                         The third argument can be any extra information about
                                          the change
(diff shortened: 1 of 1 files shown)


2xB:
From a quick look, it seems as if your changes would also work if you did not introduce has_default, as you track self.opts["default"] to be synced with has_default. Is that right?

Btw. for the next time it would probably be a good idea to configure your editor not to remove whitespace, as that would make your Pull Request easier to read.

ixjlyons:
I agree -- I think we *either* want `has_default` with `self.opts['default']` falling back to `None` *or* `'default' in self.opts` logic if a `'default'` key isn't added implicitly. I tend to have a slight preference for avoiding adding another attribute.

As a note for @2xB: there's an option in the github "files changed" tab to hide whitespace changes, though I tend to agree we should keep linting-type changes minimal outside dedicated PRs for things like that.

colehurwitz:
Thanks for the feedback @2xB and @ixjlyons . For my intended usage, I actually needed the has_default boolean. I was assuming that a user could pass in None as a default value so in that case self.opts['default']=None but has_default=False. Is that an incorrect assumption?

Sorry about the white space! I will figure that out.

ixjlyons:
Hmm, I considered the case where the user actually wants the default to be `None`, so explicitly passing `default=None` will work (`self.opts['default']` gives `None`, `hasDefault` returns `True`). I'm confused why you'd pass `default=None` but still want `hasDefault` to return `False` though -- can't you just not pass the `default` keyword argument if you don't want a default? I may be missing something here; I haven't really used the parametertree stuff much.

colehurwitz:
Ah, sorry, I screwed up that explanation (it has been a while since I made the change). I meant to say that I wanted to differentiate between when default was set to None and when default is set to None because not default value was passed in. 

User passes in None for default --> has_default=True and default=None
User does not pass in default value --> has_default=False and default=None

So if you just tracked whether or not default=None, then you can not differentiate between these cases. This is why I introduced the has_default boolean.

ixjlyons:
Right, so the approach I was suggesting is to *not* add a `'default'` key at all to `self.opts` when no default has explicitly been set. Then `hasattr(self.opts, 'default')` accurately reflects whether or not the user explicitly passed a default value (could be `None` or whatever) and `defaultValue` returns `None` as a fallback if there isn't a default value. So:

- user passes `default=None`:
    - `hasattr(self.opts, 'default')` -> `True`, so `hasDefault` -> `True`
    - `defaultValue` -> `self.opts['default']` which is `None`
- user does not pass `default` keyword arg:
    - `hasattr(self.opts, 'default')` -> `False`, so `hasDefault` -> `False`
    - `defaultValue` -> `None` as fallback

pyqtgraph/parametertree/Parameter.py
```diff
@@ -408,14 +417,17 @@ def setToDefault(self):
 
     def hasDefault(self):
         """Returns True if this parameter has a default value."""
-        return 'default' in self.opts
-        
+        return self.has_default
```
2xB: How about just returning ˋself.defaultValue() is not Noneˋ?

2xB:
I would not differenciate between no default and a default of ˋNoneˋ, as the latter is the most common to explicitly describe the former.

pyqtgraph/parametertree/Parameter.py
2xB: (Instead of using has_default etc.?)

PR #850, opened 2019-04-13
Implemented PlotDataItem.appendData()

awkbr549:
Simple implementation of PlotDataItem.appendData() based on PlotDataItem.setData().

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -634,7 +634,120 @@ class PlotDataItem(GraphicsObject):
         self.scatter.setData([])
             
     def appendData(self, *args, **kargs):
-        pass
+        """
+        Append new data to data already displayed by this item.
+        See :func:`__init__() <pyqtgraph.PlotDataItem.__init__>` for details; it accepts the same arguments.
+        """
+        profiler = debug.Profiler()
+        y = None
+        x = None
+        if len(args) == 1:
+            data = args[0]
+            dt = dataType(data)
+            if dt == 'empty':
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for the PR; would you mind adding some test code to ensure that this method behaves as intended?   The tests should go in `pyqtgraph/graphicsItems/tests/test_PlotDataItem.py`

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -634,7 +634,120 @@ def clear(self):
...
+        for k in list(self.opts.keys()):
+            if k in kargs:
+                self.opts[k] = kargs[k]
+                
+        if y is None:
+            self.updateItems()
+            profiler('update items')
+            return
+        if y is not None and x is None:
+            x = np.arange(len(y))
```
j9ac9k, line 726: I was just playing with this via the ViewBox example, and I don't think this should be correct; 

in the ViewBox example, adding this line:

```python
l1.appendData(np.flip(y))
```

will result in this   

image


If appending the data, I would imagine it would continue the x-axis from where it left off...

j9ac9k:
Looking over this diff, looks like most of the code is just copied from the `setData()` method; I'm good with implementing this feature, but we shouldn't have so much duplicated code, and should refactor out much of it into other methods to parse the input arguments at the very least.

j9ac9k:
I'm closing out this issue, but I created #1430 as I do think this is a good feature to add, so progress ideally can be tracked there.

Thanks again for the PR @awkbr549 if you can address the issues I brought up, feel free to reopen.

Issue #860, opened 2019-05-02
ErrorBarItem with no data crashes PlotWidget.addItem


jaxankey:
pyqtgraph v0.10.0 installed via conda on windows or linux (same behavior)

Hello,

To get fast plot updates, I currently add a `PlotCurveItem` to a `PlotWidget` once (via `addItem()`), then keep updating the `PlotCurveItem`'s data with `setData()`. This allows me to initially specify no data, which is helpful when I initially build the GUI.

I now wish to do the same with an `ErrorBarItem`, but it will crash unless the data when the ErrorBarItem is created: for example,

```
import pyqtgraph as pg
import numpy     as np

x = np.array([1,2,3,4,5])
y = np.array([1,2,1,3,1])
e = np.array([1,0.1,0.5,0.1,3])

plt = pg.plot()
err = pg.ErrorBarItem(x=x, y=y, top=e, bottom=e)
cur = pg.PlotCurveItem(x=x, y=y)
plt.addItem(err)
plt.addItem(cur)
```
works just fine, but
```
import pyqtgraph as pg
import numpy     as np

x = np.array([1,2,3,4,5])
y = np.array([1,2,1,3,1])
e = np.array([1,0.1,0.5,0.1,3])

plt = pg.plot()
err = pg.ErrorBarItem() #x=x, y=y, top=e, bottom=e)
cur = pg.PlotCurveItem(x=x, y=y)
plt.addItem(err)
plt.addItem(cur)
```
raises the following exception:
```
  File "/home/jack/Anaconda3/lib/python3.7/site-packages/pyqtgraph/graphicsItems/ErrorBarItem.py", line 147, in boundingRect
    return self.path.boundingRect()

AttributeError: 'NoneType' object has no attribute 'boundingRect'
```

As a hack, it is possible to simply set some dummy x and y data with top=0 and bottom=0.

Cheers,
Jack

ixjlyons:
Should be fixed now

PR #861, opened 2019-05-04
Set path attr in case ErrorBarItem initialized without data

ixjlyons:
This aims to fix issue #860.

In case the user wants to defer setting `ErrorBarItem` data, we can set the `path` attribute to the plain `QPainterPath` which has a bounding rect with zero width/height. Example (borrowed from #860):

```python
import pyqtgraph as pg
import numpy as np

x = np.arange(5)
y = np.array([1, 2, 1, 3, 1])
e = np.array([1, 0.1, 0.5, 0.1, 3])

plt = pg.plot()
err = pg.ErrorBarItem()
cur = pg.PlotCurveItem(x=x, y=y)
plt.addItem(err)
plt.addItem(cur)

err.setData(x=x, y=y, top=e, bottom=e)

pg.mkQApp().exec_()
```

This avoids the error shown in the issue, but one potential shortcoming of this approach is the bounding rect is at the origin, so it could cause scaling/range issues until the `ErrorBarItem` data is set:

```python
import pyqtgraph as pg
import numpy as np

x = np.arange(5) + 100  # offset x data by 100 to demonstrate
y = np.array([1, 2, 1, 3, 1])
e = np.array([1, 0.1, 0.5, 0.1, 3])

plt = pg.plot()
err = pg.ErrorBarItem()
cur = pg.PlotCurveItem(x=x, y=y)
plt.addItem(err)
plt.addItem(cur)

pg.mkQApp().exec_()
```

issue_860_2

### DIFF
--- a/pyqtgraph/graphicsItems/ErrorBarItem.py
+++ b/pyqtgraph/graphicsItems/ErrorBarItem.py
@@ -45,9 +45,11 @@ class ErrorBarItem(GraphicsObject):
         
         This method was added in version 0.9.9. For prior versions, use setOpts.
         """
-        if 'x' in opts and 'y' in opts:
-            self.setVisible(True)
         self.opts.update(opts)
+        if self.opts['x'] is not None and self.opts['y'] is not None:
+            self.setVisible(True)
+        else:
+            self.setVisible(False)
         self.path = None
         self.update()
         self.prepareGeometryChange()
--- a/pyqtgraph/graphicsItems/ErrorBarItem.py
+++ b/pyqtgraph/graphicsItems/ErrorBarItem.py
@@ -46,10 +46,10 @@ class ErrorBarItem(GraphicsObject):
         This method was added in version 0.9.9. For prior versions, use setOpts.
         """
         self.opts.update(opts)
-        if self.opts['x'] is not None and self.opts['y'] is not None:
-            self.setVisible(True)
-        else:
+        if self.opts['x'] is None or self.opts['y'] is None:
             self.setVisible(False)
+        else:
+            self.setVisible(True)
         self.path = None
         self.update()
         self.prepareGeometryChange()
--- a/pyqtgraph/graphicsItems/ErrorBarItem.py
+++ b/pyqtgraph/graphicsItems/ErrorBarItem.py
@@ -46,10 +46,7 @@ class ErrorBarItem(GraphicsObject):
         This method was added in version 0.9.9. For prior versions, use setOpts.
         """
         self.opts.update(opts)
-        if self.opts['x'] is None or self.opts['y'] is None:
-            self.setVisible(False)
-        else:
-            self.setVisible(True)
+        self.setVisible(all(self.opts[ax] is not None for ax in ['x', 'y']))
         self.path = None
         self.update()
         self.prepareGeometryChange()
--- a/pyqtgraph/graphicsItems/ErrorBarItem.py
+++ b/pyqtgraph/graphicsItems/ErrorBarItem.py
@@ -23,6 +23,7 @@ class ErrorBarItem(GraphicsObject):
             beam=None,
             pen=None
         )
+        self.setVisible(False)
         self.setData(**opts)
 
     def setData(self, **opts):
@@ -44,6 +45,8 @@ class ErrorBarItem(GraphicsObject):
         
         This method was added in version 0.9.9. For prior versions, use setOpts.
         """
+        if 'x' in opts and 'y' in opts:
+            self.setVisible(True)
         self.opts.update(opts)
         self.path = None
         self.update()
--- a/pyqtgraph/graphicsItems/ErrorBarItem.py
+++ b/pyqtgraph/graphicsItems/ErrorBarItem.py
@@ -59,6 +59,7 @@ class ErrorBarItem(GraphicsObject):
         
         x, y = self.opts['x'], self.opts['y']
         if x is None or y is None:
+            self.path = p
             return
         
         beam = self.opts['beam']
--- a/pyqtgraph/graphicsItems/tests/test_ErrorBarItem.py
+++ b/pyqtgraph/graphicsItems/tests/test_ErrorBarItem.py
@@ -1,12 +1,10 @@
-import pytest
-from pyqtgraph.Qt import QtGui, QtCore
 import pyqtgraph as pg
 import numpy as np
 
 app = pg.mkQApp()
 
 
-def test_errorbaritem_defer_data():
+def test_ErrorBarItem_defer_data():
     plot = pg.PlotWidget()
     plot.show()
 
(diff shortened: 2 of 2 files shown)


2xB:
As far as I understand this issue, 

https://github.com/pyqtgraph/pyqtgraph/blob/eb90616ae2cd2c26ec543865b8414a999b34b73e/pyqtgraph/graphicsItems/ErrorBarItem.py#L144

tries to get the bounding box of the ErrorBarItem by calling `self.path.boundingRect()`. If there are no entries in the item, it defaults with your suggestion to a boundingRect of width and height 0 at x=0, y=0. This destroys autorange. Therefore I would suggest not to modify `drawPath` but `boundingRect` and if necessary the occurences of `boundingRect`.

ixjlyons:
Well, I guess the question then is what do we return from `boundingRect` when there is no data? Falling back to some default seems reasonable to me.

2xB:
How about returning None if there is no bounding rectangle, and skipping every bounding box that is None when calculating autorange? This seems the most intuitive solution for me.

j9ac9k:
I'm a fan of returning `None` where appropriate, however care must be given that we don't break something else, as . now suddenly that function is returning an `Optional`; so everyone that calls that function should be prepared to handle the case it returns `None` (mypy is fantastic for this sort of thing)

ixjlyons:
The problem with returning `None` from `boundingRect` is that we have to be pretty careful to avoid that method from being called from within Qt. I think I figured out a solution that still always returns a valid `QRectF` but doesn't affect autorange until data is added. I got into the weeds there a bit but I eventually realized that toggling visibility would be straightforward and might work. Looks like it does. Tested the commit above with PyQt5 and PySide2 and this little demo:

```python
import pyqtgraph as pg
import numpy as np
import time

app = pg.mkQApp()

x = np.arange(5) + 100
y = np.array([1, 2, 1, 3, 1])
e = np.array([1, 0.1, 0.5, 0.1, 3])

plt = pg.plot()

cur = pg.PlotCurveItem(x=x, y=y)
plt.addItem(cur)

err = pg.ErrorBarItem()
plt.addItem(err)

app.processEvents()
time.sleep(1)

# set data, plot should auto-range
err.setData(x=x, y=y, top=e, bottom=e)

app.processEvents()
time.sleep(1)
```

pyqtgraph/graphicsItems/ErrorBarItem.py
```diff
@@ -46,10 +46,10 @@ def setData(self, **opts):
         This method was added in version 0.9.9. For prior versions, use setOpts.
         """
         self.opts.update(opts)
-        if self.opts['x'] is not None and self.opts['y'] is not None:
-            self.setVisible(True)
-        else:
+        if self.opts['x'] is None or self.opts['y'] is None:
```
j9ac9k: Don't meant to get nitpickey; but should we not do something like

```
self.setVisible(True) if all(self.opts[axis] is not None for axis in {'x', 'y'}) else set.Visible(False)
```

ixjlyons: Isn't that equivalent? Or are you just talking stylistically? While we're at it, this is pretty concise:

```python
self.setVisible(all(self.opts[ax] is not None for ax in {'x', 'y'}))
```

j9ac9k: oh yeah these are equivalent; I've recently been on a ternary operator kick ...and I'm seeing many opportunities to use them and don't ...

Also, if there is a more likely axis to have `self.opts[axis] is None`, we should likely specify the order in list form, instead of the set notation I used.

j9ac9k: I like your method even more

ixjlyons: The check in `drawPath` is `if x is None or y is None`, which will short-circuit if x is None. Similarly, if x is None here, the second compare won't happen if we specify `all(self.opts[ax] for ax in ['x', 'y'])` :+1:

ixjlyons:
Great, down to +3 lines. That's much better than some of the deep hacks I was trying out last night

j9ac9k:
> Great, down to +3 lines. That's much better than some of the deep hacks I was trying out last night
> 
> edit: maybe I should add a test?

Our coverage is at ~3X% so probably not the worst idea

ixjlyons:
Yeah I might as well tack on some test cases (should bump coverage a bit since there are none for `ErrorBarItem` at the moment). This is a pretty low priority PR so I might focus on something else and get back to this later.

j9ac9k:
> Yeah I might as well tack on some test cases (should bump coverage a bit since there are none for `ErrorBarItem` at the moment). This is a pretty low priority PR so I might focus on something else and get back to this later.

Eh, putting something down to pick it up later is a pretty expensive proposition... if it's something that impacts you, even if minor, probably best wrap this up now (especially if you already have a fix and just need some test cases)

pyqtgraph/graphicsItems/ErrorBarItem.py
2xB: Just to have it mentioned: Readability should be more important than line count, as it directly affects how good Pull Requests from less experienced users are. - Having said that, I also really like to play Code Golf myself ^^

j9ac9k: > Just to have it mentioned: Readability should be more important than line count, as it directly affects how good Pull Requests from less experienced users are. - Having said that, I also really like to play Code Golf myself ^^

Oh I'm totally down for readability; but an `if/elif` check in this case where was pretty verbose, and while capturing the necessary logic, I do think using the `all` keyword is appropriate here.

Then again... https://i.imgur.com/ZgQHMHo.png

ixjlyons:
> putting something down to pick it up later is a pretty expensive proposition...

Good point. Added a test. I thought about adding more tests for `ErrorBarItem` but I was considering we could probably write some general test cases and run them on a bunch of different item classes to cover generic things like setting color, adding the item to a plot, etc.

pyqtgraph/graphicsItems/tests/test_ErrorBarItem.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+from pyqtgraph.Qt import QtGui, QtCore
+import pyqtgraph as pg
+import numpy as np
+
+app = pg.mkQApp()
+
+
+def test_errorbaritem_defer_data():
```
j9ac9k: Make camelCase ?  `test_errorBarItemDeferData` perhaps?

ixjlyons: Seems to be mixed convention in the code base. One thing that seems pretty consistent is `test_ClassName_what_this_test_does`. So `test_ErrorBarItem_defer_data`?

j9ac9k: Given we don't have any consistent naming scheme, I would write whatever you want in haha; renaming to a consistent convention should happen at some-point, but not with this PR :)

PR #865, opened 2019-05-11
Implement azure ci

j9ac9k:
This pull request is for implementing azure pipelines on the pyqtgraph library.  

The configuration supports the following operating system 

- macOS 10.13
- ubuntu 16.04 
- Windows Server 2016

Under each operating system, the following Qt bindings are tested

- conda based pyqt4
- conda based pyside
- conda based pyside2 (5.6)
- conda based PyQt5 (5.9)
- pip basedd PyQt5 (5.12)
- pip based PySide2 (5.12)

For each configuration, it runs `python -m pytest --cov pyqtgraph -sv`.  The only configuration that actually passes all tests is Ubuntu-pip-PyQt5

Should I look to add other steps besides that test?


azure-pipelines.yml
```diff
@@ -0,0 +1,59 @@
...
+# Build only for PRs for master branch
+pr:
+  autoCancel: true
+  branches:
+    include:
+      - master
+      - develop
+
+variables:
+  OFFICIAL_REPO: 'j9ac9k/pyqtgraph'
```
campagnola: I guess this should change before merge?

j9ac9k: yup, update made (also cleaned up the comment code top of the non-templates file)

PR #873, opened 2019-05-18
Fix Travis CI

### LINKED ISSUE #869: Travis fails because support system_site_packages is in conflict with Conda
See https://github.com/travis-ci/travis-ci/issues/2219#issuecomment-41804942 , which includes a solution to the problem. Should that be applied here?

2xB:
Travis CI currently does not work because system_site_packages do not work with Conda installations. As system_site_packages are not needed, this pull request aims at removing them.

Note that now the build fails because of real bugs in the code, and not because of Travis.
Also note that some tests pass although there are errors in the Travis log that need attention.

Closes #869

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,6 @@ sudo: false
 notifications:
   email: false
 
-virtualenv:
-    system_site_packages: true
-
-
 env:
     # Enable python 2 and python 3 builds
     # Note that the 2.6 build doesn't get flake8, and runs old versions of


2xB:
@j9ac9k Should I quickly open the same pull request for 'develop', or will 'master' be merged into 'develop' at one point?

j9ac9k:
@2xB last we talked on slack, we're thinking of doing away w/ `develop` so I doubt we'll merge master into develop; I think doing the same PR into the `develop` branch would be the way to go.

PR #876, opened 2019-05-21
Typecast Levels to be float

Dschoni:
This circumvents cases in which "levels" is a boolean array and therefore the substraction fails due to deprecation. It seems to be, that "levels" is only used for min/max calculation further down, so this change should not affect other code.


j9ac9k:
I was just looking at this exact error a few days ago and wondering what the best way to go about addressing this.  This looks good to me, will run this by folks more knowledgable w/ the codebase.

PR #879, opened 2019-05-22
Add h5py dependency to Conda in Travis

2xB:
test_examples.py needs h5py, which it hereby gets.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ install:
     - conda update conda --yes
     - conda create -n test_env python=${PYTHON} --yes
     - source activate test_env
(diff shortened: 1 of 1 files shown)


2xB:
Note that the HDF5 test takes long for Travis to finish, so this probably needs some more thoughts.

2xB:
This needs more thoughts. Closing.

Dschoni:
What about adding a `try-except` block around the `import h5py` statement instead, so in the future, the test is run automatically if h5py is present and passes if not present?

j9ac9k:
@Dschoni discussion is at #882

PR #441, opened 2017-02-10
flowchart: python3: replaced calls to unicode function with asUnicode

giantamoeba:
necessary for running flowcharts on python3

PR #886, opened 2019-05-24
Fix flowchart save on Python 3

### LINKED ISSUE #512: Unicode calls when saving and loading flowcharts fail on Python 3
[`Flowchart.saveFile`](https://github.com/pyqtgraph/pyqtgraph/blob/develop/pyqtgraph/flowchart/Flowchart.py#L538) and [`Flowchart.loadFile`](https://github.com/pyqtgraph/pyqtgraph/blob/develop/pyqtgraph/flowchart/Flowchart.py#L517) both contain raw calls to `unicode` that have not been updated to use the python 2/3 compatibility library. These calls should be replaced by [`python2_3.asUnicode`](https://github.com/pyqtgraph/pyqtgraph/blob/develop/pyqtgraph/python2_3.py#L6).

`FlowchartCtrlWidget` has similar issues in its `fileSaved` and `setCurrentFile` methods too.

ixjlyons:
I took the work started in #231 and implemented the additional requested change to only append the `.fc` extension if the user didn't add it in the file dialog.

If this is merged, we can close duplicate pull requests #231, #423, #435, #441, #513.

Also fixes #512

### DIFF
--- a/pyqtgraph/flowchart/Flowchart.py
+++ b/pyqtgraph/flowchart/Flowchart.py
@@ -526,11 +526,6 @@ class Flowchart(Node):
         self.viewBox.autoRange()
         self.sigFileLoaded.emit(fileName)
 
-    def saveFileSelected(self, fileName):
-        if not fileName.endswith('.fc'):
-            fileName += '.fc'
-        self.saveFile(fileName=fileName)
-
     def saveFile(self, fileName=None, startDir=None, suggestedFileName='flowchart.fc'):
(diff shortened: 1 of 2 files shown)


ixjlyons:
Actually better to just let Qt handle adding the extension (now user can use their own extension if they really want to) https://doc.qt.io/qt-5/qfiledialog.html#defaultSuffix-prop

j9ac9k:
Looks good to me!

PR #888, opened 2019-05-24
TreeWidget.topLevelItems Python 3 fix

### LINKED ISSUE #726: TreeWidget.topLevelItems is broken on Python3

ixjlyons:
Fixes #726 

Opted to use the same pattern as in `TreeWidgetItem.childItems` rather than importing `xrange` from `python2_3`.

### DIFF
--- a/pyqtgraph/widgets/TreeWidget.py
+++ b/pyqtgraph/widgets/TreeWidget.py
@@ -201,7 +201,7 @@ class TreeWidget(QtGui.QTreeWidget):
         return item
 
     def topLevelItems(self):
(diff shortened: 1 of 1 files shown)


j9ac9k:
good call on returning a list instead of a map object.

Issue #511, opened 2017-07-13
hoverEvent with parentItem None


georgedimitriadis:
Hello,

I am developing with python 3.5 pyqtgraph 0.10.0 on windows 10

I am making a custom ViewBox that allows PolyLineROIs to be created live through mouse clicks.

I am using this to find the points inside the ROI that are on a ScatterPlotItem. It works fine when I am adding points to the ROI.

I also want to call a function to update the point detection when I move a ROI handle.
To do this I am calling a function connected to a sigRegionChangeFinished event. Every now and then I get the following error
File "E:\InstalledSoftware\Languages\Pythons\Miniconda3_3p5\lib\site-packages\pyqtgraph\graphicsItems\ROI.py", line 2155, in hoverEvent
    |      if self.parentItem().translatable:
    |  AttributeError: 'NoneType' object has no attribute 'translatable'

It of course goes away if I add a self.parentItem() is not None in the if clause in line 2155

Also what I realized is the sigRegionChangeFinished is being called multiple times as I am dragging the handle. I was hoping it would be called only when I released it (which would make life much easier). It is not a serious problem though for what I am doing. I am just mentioning it here.

I am happy to give you more details of the code if you need to

Thanks

j9ac9k:
Closing this issue as it's a duplicate of #884 which we're diagnosing.

Issue #884, opened 2019-05-23
[CI-fail] pyqtgraph/graphicsItems/tests/test_ROI.py::test_PolyLineROI


Dschoni:
Could be reproduced locally
```
pyqtgraph/graphicsItems/tests/test_ROI.py::test_PolyLineROI [07:10:13]  Error sending hover exit event:
    |==============================>>
    |  Traceback (most recent call last):
    |    File "/home/travis/mc/envs/test_env/bin/py.test", line 11, in <module>
    |      sys.exit(main())
    |    File "/home/travis/mc/envs/test_env/lib/python2.7/site-packages/_pytest/config/__init__.py", line 79, in main
    |      return config.hook.pytest_cmdline_main(config=config)
    |    File "/home/travis/mc/envs/test_env/lib/python2.7/site-packages/pluggy/hooks.py", line 289, in __call__
    |      return self._hookexec(self, self.get_hookimpls(), kwargs)
    |    File "/home/travis/mc/envs/test_env/lib/python2.7/site-packages/pluggy/manager.py", line 68, in _hookexec
    |      return self._inner_hookexec(hook, methods, kwargs)
    |    File "/home/travis/mc/envs/test_env/lib/python2.7/site-packages/pluggy/manager.py", line 62, in <lambda>
    |      firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
    |    File "/home/travis/mc/envs/test_env/lib/python2.7/site-packages/pluggy/callers.py", line 187, in _multicall
    |      res = hook_impl.function(*args)
    |    File "/home/travis/mc/envs/test_env/lib/python2.7/site-packages/_pytest/main.py", line 242, in pytest_cmdline_main
    |      return wrap_session(config, _main)
    |    File "/home/travis/mc/envs/test_env/lib/python2.7/site-packages/_pytest/main.py", line 205, in wrap_session
    |      session.exitstatus = doit(config, session) or 0
    |    File "/home/travis/mc/envs/test_env/lib/python2.7/site-packages/_pytest/main.py", line 249, in _main
    |      config.hook.pytest_runtestloop(session=session)
    |    File "/home/travis/mc/envs/test_env/lib/python2.7/site-packages/pluggy/hooks.py", line 289, in __call__
    |      return self._hookexec(self, self.get_hookimpls(), kwargs)
    |    File "/home/travis/mc/envs/test_env/lib/python2.7/site-packages/pluggy/manager.py", line 68, in _hookexec
[...]

PR #889, opened 2019-05-24
Check if Item actually has a parent before checking for collision

Dschoni:
Trying a fix for #884

### DIFF
--- a/azure-test-template.yml
+++ b/azure-test-template.yml
@@ -12,8 +12,8 @@ jobs:
     matrix:
       Python27-Qt4:
         python.version: '2.7'
-        install.method: "conda"
         qt.bindings: "pyqt=4"
+        install.method: "conda"
       Python27-PySide:
         python.version: '2.7'
         qt.bindings: "pyside"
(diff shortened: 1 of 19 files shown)


Dschoni:
If this works, would also close #511

j9ac9k:
I think this PR may need to be rebased, some of the commits here appear unrelated.

As discussed in the issue, I'm reluctant to accept this fix as I'm not seeing how this case is possible to begin with and we may have another larger issue in play.

I want to walk through this test in the debugger and see if I can figure out why this is happening

j9ac9k:
@Dschoni can you rebase this PR?

Dschoni:
@j9ac9k My git is a mess right now. I did a rebase, leading me to an error ("Patch already has been applied"), so I had to merge remote tracking branch and.... well.... it should be kind of good now.

j9ac9k:
> @j9ac9k My git is a mess right now. I did a rebase, leading me to an error ("Patch already has been applied"), so I had to merge remote tracking branch and.... well.... it should be kind of good now.

Ha, we've all been there with the git mess.  I'm going to close this PR, it's showing as having ~40+ commits with a bunch of file changes.

Checking for the parentItem() not being None I don't think is the way we want to solve this issue.  If we're going to do a check like this we should check if `self.scene() is None`; but I really do think this is a bug elsewhere in the `removeHandle` or `removeSegment` code.

PR #890, opened 2019-05-24
Show failing examples as "FAILED" in pytest

2xB:
Catching everything and then just outputting the words "FAILED" does not help keeping track of errors. With this pull request, pytest is noticed via assert when an error occurred. This also replaces further additional text output of whether the test was failing.


j9ac9k:
Thanks for the PR.  I'm going to close this one out as I think #911 handled the same thing, plus it gave some additional benefits.

Now that we have more failures, we really have our work cut out for us :|

PR #891, opened 2019-05-24
Only execute hdf5 example if module h5py exists

### LINKED ISSUE #882: [CI-fail] HDF5 big data
This test fails due to h5py missing. Probably we can change requisites to include h5py or skip this test if h5py is not found.
```
examples/test_examples.py::test_examples[PyQt4-f17] frontend = PyQt4. f = ('HDF5 big data', 'hdf5.py')
HDF5 big data.....................................FAILED
Python 2.7.16 |Anaconda, Inc.| (default, Mar 14 2019, 21:00:58) 
[GCC 7.3.0] on linux2
  File "hdf5.py", line 19, in <module>
    import h5py
ImportError: No module named h5py
>>> 
```

2xB:
Introduces a dictionary containing requirements of chosen examples. If a requirement is not met, the example will not be executed.

To test whether this really also works if h5py exists, I created a new branch with a similar commit but also let travis import h5py (not the identical, it contains one unnecessary import statement more - sorry):
branch: https://github.com/2xB/pyqtgraph/commits/hdf5_test
log: https://travis-ci.com/2xB/pyqtgraph/jobs/203002184

Closes #882

### DIFF
--- a/examples/utils.py
+++ b/examples/utils.py
@@ -120,7 +120,7 @@ def testFile(name, f, exe, lib, graphicsSystem=None):
             try:
                 __import__(requirement)
             except ImportError:
-                print(" Requirement {} of this test not satified ".format(requirement))
+                print(" Requirement \"{}\" of this test not satified ".format(requirement))
                 pytest.skip()
 
     import1 = "import %s" % lib if lib != '' else ''
--- a/examples/utils.py
+++ b/examples/utils.py
@@ -117,11 +117,7 @@ def testFile(name, f, exe, lib, graphicsSystem=None):
     
     if f in example_requirements:
         for requirement in example_requirements[f]:
-            try:
-                __import__(requirement)
-            except ImportError:
-                print(" Requirement \"{}\" of this test not satified ".format(requirement))
-                pytest.skip()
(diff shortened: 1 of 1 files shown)


2xB:
This suffers from the same problem @j9ac9k mentioned in #892: It makes ˋpytestˋ a dependency of ˋpyqtgraph.examplesˋ.

tacaswell:
In Matplotlib, we have a private `matplotlib._called_from_pytest` attribute that we use to flag when we are running in the context of pytest (see https://github.com/matplotlib/matplotlib/blob/7b100001f4a2049fa1a492f551b3322409f90e22/lib/matplotlib/testing/conftest.py#L7-L10).  If pyqtgraph adopted this, then we could hide the pytest dependancy here via

```python
try:
     __import__(requirement)
except ImportError:
    if getattr(pyqtgraph, '_called_from_pytest', False):
        import pytest
        pytest.markskip()
     else:
        raise

```

or something like that.

2xB:
HDF5 example should not be tested at all, which @campagnola states in issue #882. Therefore, PR #905 is a better solution.

PR #892, opened 2019-05-25
Skip testing example if ImportError occurs

### LINKED ISSUE #882: [CI-fail] HDF5 big data

2xB:
This is an alternative to #891 with the advantage of less code that needs to be maintained (no extra dict) and therefore less bug probability. Note that this does produce more ugly output, because stdout and stderr are both printed, so one can reproduce what caused the ImportError.

Closes #882

### DIFF
--- a/examples/utils.py
+++ b/examples/utils.py
@@ -170,8 +170,6 @@ except:
         if output.endswith('test failed'):
             fail = True
             break
-        if output.endswith('test skipped'):
-            break
     time.sleep(1)
     process.kill()
     #res = process.communicate()
--- a/examples/utils.py
+++ b/examples/utils.py
@@ -4,6 +4,7 @@ import time
 import os
 import sys
 import errno
+import pytest
 from pyqtgraph.pgcollections import OrderedDict
 from pyqtgraph.python2_3 import basestring
 
@@ -129,6 +130,9 @@ try:
     while True:  ## run a little event loop
         pg.QtGui.QApplication.processEvents()
         time.sleep(0.01)
+except ImportError:
+    print("test skipped")
+    raise
 except:
     print("test failed")
     raise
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for the PR, I am uncertain about this implementation, because if I'm reading this right, this would make `pytest` a dependency to run `python -m pyqtgraph.examples`; as `./examples/__main__.py` imports the utils module in that same directory.

Whatever solution we need to make, we need to embed it inside `./examples/test_examples.py`

2xB:
I'm unshure if I get you right - isn't what you are criticizing already the case without the pull request?

2xB:
That is right. Does it make sense to have testFile inside utils.py in the first place (instead of test_examples.py)?

j9ac9k:
> That is right. Does it make sense to have testFile inside utils.py in the first place (instead of test_examples.py)?

That's a good question, and I'm inclined to agree with you, we should get @campagnola 's opinion on this.

campagnola:
Agreed -- it makes sense to have `testFile` inside `test_examples.py`.
These changes look good, but a proper fix for #882 is just to ensure that the h5py example is never run as a test, even if h5py can be imported. That example requires user interaction and creates a large file to read from, all things we don't want running in tests. So I'd also add to `test_examples` a list of the examples that we know we want to skip.

2xB:
That is a good point. In that case, skipping examples on ImportError might lead to things skipped because of misconfigured testing environments, I would therefore drop this PR and adapt PR #891 to the suggestion of @campagnola .

PR #897, opened 2019-05-26
Add PyQt5 and PySide2 to test_example.py

2xB:
Before, example tests were skipped for Python 3 on Travis because it used PyQt5, which was not used for testing examples.


j9ac9k:
Should we enable PySide2 while we're at it?

j9ac9k:
Looks great to me, thanks for the PR.

PR #900, opened 2019-05-27
Fix for setOrientation 

Oscar438:
Proposed change to setOrientation function of Dock.py.

Should fix issue pyqtgraph#833


2xB:
Did you also look into why this happens? From my intuitive understanding, `container` should be an instance of one of the classes defined in https://github.com/pyqtgraph/pyqtgraph/blob/develop/pyqtgraph/dockarea/Container.py, and therefore have a `type` method.

j9ac9k:
Thanks for the PR @Oscar438 , we've decided to opt for the solution in #1268 which should fix this issue this PR is meant to address.  Thanks for the submission.

PR #902, opened 2019-05-28
New Templates, without coverage, using screen resolution utility

j9ac9k:
This is a new azure-pipelines template that makes use of the Windows screen resolution utility.  I also cleaned up a number of the conda configurations so more of the steps are cross-platform compatible.

I also removed the coverage check within pytest as that caused a segmentation fault on Windows - PySide2 5.12 configuration, and we'll likely want to patch that up at some point.

On my fork, this template shows 5/18 configurations passing.

### DIFF
--- a/azure-test-template.yml
+++ b/azure-test-template.yml
@@ -36,96 +36,32 @@ jobs:
         install.method: "pip"
 
   steps:
-  - powershell: Write-Host "##vso[task.prependpath]$env:CONDA\Scripts"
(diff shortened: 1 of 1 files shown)


j9ac9k:
wait, I messed something up, azure is saying it's not a valid config file...

j9ac9k:
actually I'm going to cancel this PR as I think I just discovered a way to clean up the templates file some more.

PR #903, opened 2019-05-28
More Robust CI

### LINKED ISSUE #898: [CI-fail] on azure-pipelines, windows pip builds aren't running...
When azure-pipelines attempts to run the tests on Windows-pip builds (PyQt5-5.12, and PySide2-5.12), the following error is generated:

```
    import fcntl
ModuleNotFoundError: No module named 'fcntl'
```

likely the template needs to be modified to include that package.

j9ac9k:
This CI configuration provides a few benefits over the existing one:

* Steps are cross-platform compatible and cross install method (pip/conda) compatible as well, so the configuration is far easier to read, and there is no duplicated code
* Make use of Screen Resolution Utility so tests can now pass on Windows
* Make use of the miniconda and python built in tasks
* Test reporting tools (including coverage) are uploaded and displayed properly In the Code Coverage tab
* Screen shots from failed Image tests are saved as artifacts and can be downloaded for review
* `pyqtgraph` is installed from a built wheel, also that built wheel is available to download as an artifact
* Uses the native test reporting
* This PR fixes #898 

What is not added is the ability to attach files to specific tests.  The idea would be that for every test, we grabbed a screen shot of the active widget and stored it along with the test; but to my knowledge you cannot do that with their current setup.  I created a feature request, if anyone thinks this is a good idea, please upvote it here:

[Azure Pipelines Docs for Attachments](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=azure-devops&tabs=yaml#attachments-support)
[Feature Requesst](https://developercommunity.visualstudio.com/idea/589590/allow-for-attachments-to-junitxml-tests.html)

### DIFF
--- a/azure-test-template.yml
+++ b/azure-test-template.yml
@@ -12,8 +12,8 @@ jobs:
     matrix:
       Python27-Qt4:
         python.version: '2.7'
-        install.method: "conda"
         qt.bindings: "pyqt=4"
+        install.method: "conda"
       Python27-PySide:
         python.version: '2.7'
         qt.bindings: "pyside"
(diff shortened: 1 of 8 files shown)


j9ac9k:
Adding `WIP` label as I've been making more changes to the template on my branch, and now that #904 is merged, the next iteration of the CI should take advantage of the artifacts generated from failed tests.

j9ac9k:
removed `[WIP]` label, updated OP with summary of changes

j9ac9k:
Hmm...the faulthandler is having problems with windows CI runs, need to investigate how to handle, would recommend holding off on merging until . I sort this out.

j9ac9k:
pytest devs not wasting any time, just a little while after opening an [issue with pytest-faulthandler](https://github.com/pytest-dev/pytest-faulthandler/issues/32#) there is a [PR submitted](https://github.com/pytest-dev/pytest-faulthandler/pull/33) to address it.

j9ac9k:
With pytest-faulthandler 1.6 released, we're back in business.

2xB:
Is testing only Python 3.7 reasonable, as many Linux distros still ship with 3.6?

j9ac9k:
> Is testing only Python 3.7 reasonable, as many Linux distros still ship with 3.6?

I have no problem modifying one of the configurations to run on python 3.6 instead of 3.7 (this change would propagate to all OSs).  Is there a specific configuration you think should be changed?

2xB:
The last Ubuntu LTS ships Python 3.6 and has PyQt 5.10.1 in its repositories: https://packages.ubuntu.com/bionic/python3-pyqt5

One could therefore just add this combination as a test case, what do you think?

j9ac9k:
> The last Ubuntu LTS ships Python 3.6 and has PyQt 5.10.1 in its repositories: https://packages.ubuntu.com/bionic/python3-pyqt5
> 
> One could therefore just add this combination as a test case, what do you think?

I think that's a good configuration to test for.  I'm not thrilled to add another test configuration (we're already at 18!), but I can see this being a common enough of a use case that we should check for it.

So I make sure I understand correctly, the idea would be to install `numpy`, `scipy`, `pyqt`, `pip` via `apt` and then install the test dependencies via `pypi` and go from there?

This would complicate the CI system, as I would need to install another template (as we won't be installing pyqt via pypi, but via apt). but it should be doable.  Unfortunately azure pipelines only has an ubuntu 16.04 image, they don't have an 18.04 image yet (supposedly it's coming at some point).

Dschoni:
I would just add Python 3.6 with PyQt 5.10.1. but not necessarily on a Ubuntu 18.04 but rather on the same 16.04 image. There "should" be no incompatabilities.

j9ac9k:
I'm not sure how the apt packages for pyqt/numpy/etc are built but I have to imagine there are differences vs the pypi packages and if we want to test that kind of setup, we should make an effort to do so.


In terms of adding a python3.6, pyqt5 5.10.1 setup, we can certainly do that, but it would get applied to all OSs, there would be far too much code duplication to make it worthwhile to do just Linux.

j9ac9k:
I really like the idea of doing the native configurations via non-pypi install methods ...the more I think about it, the better of an idea this is.  `xenial` universe has python3 (3.5.1), numpy (1.11), scipy (0.17), pyqt5 (5.5.1) pyqt4 (4.11.?)

Upon thinking about it more, these versions are probably are a great representation of the oldest version of dependencies we should test/support.

Sounds like 18.04 images are coming to azure pipelines in summer/early fall (per: https://github.com/Microsoft/azure-pipelines-image-generation/issues/506#issuecomment-490508153 ), I would ask we punt on trying to recreate the native 18.04 environment via pypi (or conda-forge) installs, and just wait until the official image is released and test on that.  In the meantime we implement a "native" xenial install via apt.

@Dschoni If you want python 3.6 to be a test configuration, how about I modify the `pyqt5-conda` configuration (which installs pyqt5 5.9.x) instead, and run that for all OSs?

ksunden:
One option for that problem is to use Travis for "native" and have Azure as the version compatibility matrix.

PR #904, opened 2019-05-29
Take screenshots + Show Resolution

j9ac9k:
This PR consists of a few changes.

1. Add `--pyside2` option to `examples/__main__.py
2. Take screenshots of failed tests, and optionally upload them to `data.pyqtgraph.org`
  * We will use this capability to extract failed images from CI systems via Artifact upload
3. Have `test_ViewBox.py` show us the available resolution of the desktop being tested on.
  * This may assist with troubleshooting the `assertMapping` failed tests on macOS Azure pipeline

Once this PR is merged, I will update the azure-pipeline templates to add the ability to capture the stored images from the failed tests and upload them as an artifact for further review.

### DIFF
--- a/pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
@@ -31,8 +31,6 @@ def init_viewbox():
     
     g = pg.GridItem()
     vb.addItem(g)
-    desktop = app.desktop().screenGeometry()
-    print("\n\nDesktop Resolution: {} x {}\n\n".format(desktop.width(), desktop.height()))
     app.processEvents()
     
 def test_ViewBox():
--- a/pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
@@ -31,7 +31,8 @@ def init_viewbox():
     
     g = pg.GridItem()
     vb.addItem(g)
-    
+    desktop = app.desktop().screenGeometry()
+    print("\n\nDesktop Resolution: {} x {}\n\n".format(desktop.width(), desktop.height()))
     app.processEvents()
     
 def test_ViewBox():
--- a/pyqtgraph/tests/image_testing.py
+++ b/pyqtgraph/tests/image_testing.py
@@ -191,6 +191,7 @@ def assertImageApproved(image, standardFile, message=None, **kwargs):
         if os.getenv('PYQTGRAPH_AUDIT_ALL') == '1':
             raise Exception("Image test passed, but auditing due to PYQTGRAPH_AUDIT_ALL evnironment variable.")
     except Exception:
+
         if stdFileName in gitStatus(dataPath):
             print("\n\nWARNING: unit test failed against modified standard "
                   "image %s.\nTo revert this file, run `cd %s; git checkout "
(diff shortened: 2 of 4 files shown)


pyqtgraph/tests/image_testing.py
```diff
@@ -306,15 +309,25 @@ def saveFailedTest(data, expect, filename):
     img[2:2+diff.shape[0], -diff.shape[1]-2:-2] = diff
 
     png = makePng(img)
+    directory, _, _ = filename.rpartition("/")
```
campagnola: FYI this is `os.path.dirname`

campagnola: .. also, not safe to use `"/"` because on windows it's `"\"`; that's what `os.path.sep` is for.

pyqtgraph/tests/image_testing.py
```diff
@@ -306,15 +309,25 @@ def saveFailedTest(data, expect, filename):
...
+    if not os.path.isdir(directory):
+        os.makedirs(directory)
+    with open(filename + ".png", "wb") as png_file:
+        png_file.write(png)
+    print("\nImage comparison failed. Test result: %s %s   Expected result: "
+        "%s %s" % (data.shape, data.dtype, expect.shape, expect.dtype))
+    if upload:
+        uploadFailedTest(filename)
     
+def uploadFailedTest(filename):
```
campagnola: I guess this is soon-to-be dead code anyway, but `png` is not defined in this function.

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
```diff
@@ -31,7 +31,8 @@ def init_viewbox():
     
     g = pg.GridItem()
     vb.addItem(g)
-    
+    desktop = app.desktop().screenGeometry()
+    print("\n\nDesktop Resolution: {} x {}\n\n".format(desktop.width(), desktop.height()))
```
campagnola: It actually seems like this is useful enough to go into `pg.systemInfo()`, in which case it would be printed whenever we run `pyqtgraph/test.py`

pyqtgraph/tests/image_testing.py
j9ac9k: you're right, normally I use `pathlib` to get around this... but `os.path.dirname` is the correct call.

pyqtgraph/tests/image_testing.py
j9ac9k: you're right; I need to pass the png in there; also regarding the dead code, I'll leave that up to you, as you are the one maintaining the server.

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
j9ac9k: I considered doing this in `pg.systemInfo()` but it doesn't have a `QApplication` running at that point from what I could tell.  Should I create a `QApplication` instance?

pyqtgraph/tests/image_testing.py
j9ac9k: Fixed with c4e295c

pyqtgraph/tests/image_testing.py
j9ac9k: fixed with c4e295c

pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
j9ac9k: Migrated this to `test.py` per offline discussion in 4b26519

PR #907, opened 2019-05-30
Fixed ViewBox.updateViewRange to apply transformation sooner

mliberty1:
With PySide2, I traced updateViewRange triggering the sigXRangeChanged and sigYRangeChanged before _matrixNeedsUpdate was set to True.  This results in an incorrect mapViewToScene call in my application.  By assigning _matrixNeedsUpdate to True before these signals, the transform is correctly updated within the slot's code giving the correct mapViewToScene result.

### DIFF
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -1496,14 +1496,13 @@ class ViewBox(GraphicsWidget):
         changed = [(viewRange[i][0] != self.state['viewRange'][i][0]) or (viewRange[i][1] != self.state['viewRange'][i][1]) for i in (0,1)]
         self.state['viewRange'] = viewRange
 
-        # emit range change signals
-        if changed[0]:
-            self.sigXRangeChanged.emit(self, tuple(self.state['viewRange'][0]))
-        if changed[1]:
-            self.sigYRangeChanged.emit(self, tuple(self.state['viewRange'][1]))
-
         if any(changed):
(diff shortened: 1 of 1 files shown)


j9ac9k:
@mliberty1 I'm curious if this is the same issue I've been seeing hanging up PySide2 5.6 tests...


Which version of PySide2 are you using?

mliberty1:
I am using PySide 5.12.0 with Python 3.7.3 on Win 10 x64.  However, the Joulescope application is deployed to Linux & Mac OS X, too.  I did not try to see if the problem is the same on those platforms yet.

j9ac9k:
I'm interested in this issue as I've experienced similar issues in the past.

I'll try and clone into your branch and run it through tox in the next day or so...

mliberty1:
I just fixed this one case, and I did not audit the pyqtgraph codebase.  If pyqtgraph was developed thinking that all signals go to the Qt Event Queue, then it's likely that other similar issues are scattered throughout pyqtgraph.  If so, then it would result in a lot of strange behaviors that are highly dependent upon how each individual application uses pyqtgraph.

j9ac9k:
While testing this change, I think i came across a `tox` bug....which I'm 99.9% sure has nothing to do with this PR; but as our CI/test suite is a  bit of a mess (also something I'm working on, see PR #903 ), I do want to run this through my `tox` suite which has been fairly reliable.

Sorry it's taking me a while to follow up.

j9ac9k:
Thanks for the PR, going to merge, looks good to me.

PR #908, opened 2019-06-01
Add pytest config file specifying colordepth when using pytest-xvfb

j9ac9k:
This PR is meant to address some of the failures on Linux, as @campagnola pointed out he was suspicious of the color depth setting.  I ran the tests on one of the head less servers I have access to and confirmed 6 tests that failed now pass.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,3 @@
+[pytest]
+
+xvfb_colordepth = 24


j9ac9k:
not sure why Azure pipelines isn't running on this PR, it's running on my branch, you can see the results here:

https://dev.azure.com/j9ac9k/pyqtgraph/_build/results?buildId=154

The short of it is that PySide2-5.6/PyQt5-5.9 now pass on Linux as well, we're up to 5 passing builds (of 18)

PR #909, opened 2019-06-01
Fixed GLViewWidget Gimbal Lock with Quaternion Parametrization

yycho0108:
**Problem**
Currently, the `GLViewWidget` is parametrized with a spherical coordinate (`azimuth` and `elevation`). However, this results in a [gimbal lock](https://en.wikipedia.org/wiki/Gimbal_lock), which can lead to an unnatural interaction with the GUI.

**Solution**
GLViewWidget has been refactored to use a `QtGui.QQuaternion` that can sufficiently represent the orientation of the camera with 3DoF.

This is a close-to drop-in replacement for the previous implementation, but any code that relied on features such as  `GLViewWidget.opt['azimuth']`  might be broken. As this introduces a minor break in the API, I'm not 100% sure if everyone would welcome this change -- but I was certainly quite frustrated when I encountered this problem.


j9ac9k:
Thanks for the PR @yycho0108 I'm inclined to agree that using spherical coordinates is not ideal here for the reason you described.  I'm not familiar with this part of the code base, so I'm going to want to experiment with it some (as well with testing out your PR/branch).

j9ac9k:
I cloned into your branch; but when running the examples I keep getting the following errors:

```
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/opengl/GLViewWidget.py", line 187, in paintGL
    self.setModelview()
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/opengl/GLViewWidget.py", line 140, in setModelview
    m = self.viewMatrix()
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/opengl/GLViewWidget.py", line 147, in viewMatrix
    u, h = self.opts['rotation'].getAxisAndAngle()
AttributeError: 'PySide2.QtGui.QQuaternion' object has no attribute 'getAxisAndAngle'
```

When I ran this using PyQt5 it worked (but I have some concerns).

This is strange given that the method is [documented](https://doc.qt.io/qtforpython/PySide2/QtGui/QQuaternion.html).

Any suggestions?

ixjlyons:
That method was introduced in Qt 5.5, so we wouldn't be able to use it as long as we're supporting Qt 4 anyway: https://doc.qt.io/qt-5/qquaternion.html#getAxisAndAngle

We can always wrap `QQuaternion` with a `Quaternion` class (like `Vector`) and add the method ourselves, or just compute the axis-angle representation locally.

j9ac9k:
> That method was introduced in Qt 5.5, so we wouldn't be able to use it as long as we're supporting Qt 4 anyway: https://doc.qt.io/qt-5/qquaternion.html#getAxisAndAngle

Ahh, I just looked that the object was introduced in Qt 4.6, didn't see that the method hadn't come around until later.

> We can always wrap `QQuaternion` with a `Quaternion` class (like `Vector`) and add the method ourselves, or just compute the axis-angle representation locally.

Certainly possible; how nasty are the math operations here with respect to numpy operations?

yycho0108:
Sorry for the late response and thanks for the feedback! @ixjlyons is correct - there is no `getAxisAngle` for Qt4. However, the method should not be necessary as the rotation could simply be achieved by calling `tr.rotate()` with the Quaternion argument directly. @j9ac9k I haven't extensively tested against Qt4, so let me know if other concerns or problems arise.

j9ac9k:
@yycho0108 I'll do another run through tomorrow and take a look.  Thanks for the update!

On a slightly unrelated note, if you want an easy way to run the test suite on a variety of our supported configurations, you only need `conda` in your path, `tox` and `tox-conda` (a plugin for `tox` to create test environments using conda environments).  Once created you can run `tox` on it's own and it will run the entire unit test suite (this will take a while), alternatively you can run `tox -- -k "testExamples" and it will run through all the examples (must be on the current dev branch) in an automated fashion (note, they don't all pass right now!).

j9ac9k:
I can confirm this works w/ pyside2 now, thanks for the update.  I'm going to test this a little more thoroughly on my machine and ask some of the other contributors to see if they have any issues.

campagnola:
FWIW the gimbal lock is deliberate and desirable in situations where the model data have a natural "up" direction; in this case the user can become disoriented if it is possible to tilt the z axis sideways. So in addition to the API change, it is a major UI behavior change as well. 

I would only recommend merging this if the default API and behavior are preserved, and an option is introduced to allow switching between different user interaction modes. We did this for vispy; it worked reasonably well there.

j9ac9k:
> I would only recommend merging this if the default API and behavior are preserved, and an option is introduced to allow switching between different user interaction modes. We did this for vispy; it worked reasonably well there.

@yycho0108 would this work for you? I do like the mechanism you proposed but if there was a deliberate design decision to use spherical coordinates, we should certainly keep that as the default.

I also still need to test with qt4 bindings.

yycho0108:
@j9ac9k Sorry for the radio silence on my part. I agree with @campagnola that we should probably keep the existing behavior as there are good reasons to keep this as the default. I'll close this PR for now, but perhaps others could reference the implementation here if they seek to have all three degrees of freedom for rotation.

j9ac9k:
@yycho0108 I was actually just thinking about modifying this PR to maintain the default behavior but have this method available as an option, because I do think this change is useful.  If you don't mind I'll re-open this PR; I don't mind taking over the changes.

j9ac9k:
ugh, sorry, I messed up your branch.  I'll straighten this out momentarily

yycho0108:
Happy that you still had this PR in mind a full year after I initially opened it!
Let me know I can do anything for you to help, @j9ac9k .

j9ac9k:
It's a good feature you implemented, it's compatible with our entire dependency matrix, and while I'm with @campagnola that I don't think this should be the default behavior, it should be an option for users to implement. 

I actually had it working locally (having rotationMethod optionally be set to `quaternion`), but got confused between my forks' develop branch, and your fork's develop branch, hence messing up your branch...so I'll redo it; it wasn't that much work...

Not being in this space, can you recommend me some suggested vocabulary since I don't do much work in 3D space.   I was considering having `rotationMethod` by the optional keyword parameter in `QGLWidget.__init__`, which defaults to `gimbal` but would also take `quaternion`.  Would that naming convention make the most sense?

yycho0108:
Sounds good. Perhaps `euler` instead of `gimbal` might be a more recognizable term as `euler` and `quaternion` both state canonical representational forms of rotation, while `gimbal` specifically refers to a specific rotational system or (in case of gimbal lock) a particular problem inherent to a euler representation.

j9ac9k:
I'm super unfamiliar with OpenGL related issues, but is there a reason why the volumetric example would behave in this manner?

output

yycho0108:
@j9ac9k I'm not too familiar with other parts of the code that touch on the opengl pipeline,
but based on a quick look it looks like the issue is likely somewhere from line 95 to line 103 in `GLVolumeItem.py`. I'll have to figure out exactly how `GLVolumeItem` works to know precisely which logic is being inconsistent, though.

j9ac9k:
Thanks, if you do decide to add more to the PR, would you mind doing it in another branch?  Having your branch be named `develop` makes hoping onto it and manipulating/testing a little trickier.

And for the record, I don't think the behavior I show above is a show-stopper, I posted it to verify I didn't do something wrong w/ my attempts to make the rotation methods optional.  If there is a bug elsewhere in the library, I would consider it outside the scope of this PR.

j9ac9k:
Curse you Qt4!

```
AttributeError: 'PySide.QtGui.QQuaternion' object has no attribute 'toEulerAngles'
```

@yycho0108 in your workflow do you use Qt4 at all? ... reason I'm asking is if I want to hold off on merging this PR until after we adopt NEP-29 (and drop py2/qt4 support).

j9ac9k:
seeing this PR makes use of `.fromEulerAngles()` and `.toEulerAngles()` I am recommending we merge this post NEP-29 adoption.

j9ac9k:
Closing/Re-Openning to trigger on new CI system.

j9ac9k:
remaining CI failures are unrelated, merging.  Thanks @yycho0108 for the PR and the follow up changes.  Sorry this took us so long to merge.

yycho0108:
Thanks for the tenacity with which you have held onto this PR, and for making this feature a reality @j9ac9k . Happy New Year!

PR #910, opened 2019-06-01
Moved emits after all method state updates

mliberty1:
PySide2 immediately executes signals without passing through the Qt Event Queue. 

Pull request #907 addressed a specific case where a signal was emitted before a state update.
If an application's slot then calls back into the instance, the instance was in an inconsistent
state.  This pull request audits and fixes similar issues throughout the pyqtgraph library.  This
commit fixes several latent issues:

* SignalProxy: flush -> sigDelayed -> signalReceived would have incorrectly resulted in timer.stop().
* ViewBox: resizeEvent -> sigStateChange -> background state
* ViewBox: setRange -> sigStateChange -> autoranging not updated correctly
* ViewBox: updateMatrix -> sigTransformChanged -> any _matrixNeedsUpdate = True -> ignored
* Parameter: Child may have missed state tree messages on insert or received extra on remove
* GraphicsView: updateMatrix -> sigDeviceRangeChanged/sigDeviceTransformChange -> before propagated to locked viewports.

### DIFF
--- a/pyqtgraph/SignalProxy.py
+++ b/pyqtgraph/SignalProxy.py
@@ -67,11 +67,11 @@ class SignalProxy(QtCore.QObject):
         """If there is a signal queued up, send it now."""
         if self.args is None or self.block:
             return False
-        #self.emit(self.signal, *self.args)
-        self.sigDelayed.emit(self.args)
-        self.args = None
+        args, self.args = self.args, None
         self.timer.stop()
         self.lastFlushTime = time()
+        #self.emit(self.signal, *self.args)
+        self.sigDelayed.emit(args)
         return True
         
     def disconnect(self):
--- a/pyqtgraph/dockarea/Dock.py
+++ b/pyqtgraph/dockarea/Dock.py
@@ -346,9 +346,9 @@ class DockLabel(VerticalLabel):
         ev.accept()
             
     def mouseReleaseEvent(self, ev):
+        ev.accept()
         if not self.startedDrag:
             self.sigClicked.emit(self, ev)
-        ev.accept()
         
     def mouseDoubleClickEvent(self, ev):
         if ev.button() == QtCore.Qt.LeftButton:
--- a/pyqtgraph/flowchart/Flowchart.py
+++ b/pyqtgraph/flowchart/Flowchart.py
@@ -503,8 +503,8 @@ class Flowchart(Node):
         finally:
             self.blockSignals(False)
             
-        self.sigChartLoaded.emit()
         self.outputChanged()
+        self.sigChartLoaded.emit()
         self.sigStateChanged.emit()
             
     def loadFile(self, fileName=None, startDir=None):
--- a/pyqtgraph/graphicsItems/HistogramLUTItem.py
+++ b/pyqtgraph/graphicsItems/HistogramLUTItem.py
@@ -205,8 +205,8 @@ class HistogramLUTItem(GraphicsWidget):
     def regionChanging(self):
         if self.imageItem() is not None:
             self.imageItem().setLevels(self.getLevels())
-        self.sigLevelsChanged.emit(self)
         self.update()
+        self.sigLevelsChanged.emit(self)
 
     def imageChanged(self, autoLevel=False, autoRange=False):
         if self.imageItem() is None:
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -711,10 +711,10 @@ class ROI(GraphicsObject):
                 
         if hover:
             self.setMouseHover(True)
-            self.sigHoverEvent.emit(self)
             ev.acceptClicks(QtCore.Qt.LeftButton)  ## If the ROI is hilighted, we should accept all clicks to avoid confusion.
             ev.acceptClicks(QtCore.Qt.RightButton)
             ev.acceptClicks(QtCore.Qt.MidButton)
+            self.sigHoverEvent.emit(self)
         else:
             self.setMouseHover(False)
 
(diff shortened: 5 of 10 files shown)


j9ac9k:
Thanks for the PR; do you feel like you caught the relevant use cases and if now, is there a way you would recommend for seeking more of these out?

mliberty1:
I think that I have addressed all of the simple cases, but more likely remain.  I will discuss the problem, what I did in this pull request, and what future work remains.

### The Problem

The problem fundamentally reduces to mutable state consistency and reentrancy.  For this work, the reentrancy happens in a single thread, and the concern does not include [Reentrancy and Thread Safety](https://doc.qt.io/qt-5/threads-reentrancy.html).  I am also using the term reentrancy loosely, as I mean any code that operates on the same shared state (usually the class instance members which are mutated by methods of the class instance).

Before finding #907, my mental model was that signals were put onto the QT Event Queue and dispatched to slots at a later time.  This implementation choice would have allowed all current code to run to completion but with higher overhead.  However, this is not the case with QT.  From [Signals & Slots](https://doc.qt.io/qtforpython/overviews/signalsandslots.html):

> When a signal is emitted, the slots connected to it are usually executed immediately, just like a normal function call. When this happens, the signals and slots mechanism is totally independent of any GUI event loop. Execution of the code following the emit statement will occur once all slots have returned. The situation is slightly different when using queued connections ; in such a case, the code following the emit keyword will continue immediately, and the slots will be executed later.

When investigating my application issues that led to #907, it was very clear that the ViewBox call stack was emitting a signal that then called a different ViewBox method, but the ViewBox state was not yet fully updated resulting in incorrect data being returned to the slot code.

### This Pull Request

This pull request addresses the simplest cases.  I ensured that all methods containing "emits" made no further changes to the instance state, which typically means moving the emits to be the last code in the method body.  I searched all pyqtgraph code for "\.emit\s*\(" which yielded about 190 hits.  I manually inspected each.  When the emit had other code after it, I rearranged the order of execution to ensure all state changes were made before the emit.  

### Future work

This pull request does not address methods that part of the way through their body invoke other methods that then emit.  It also does not handle callbacks or direct method calls.  The safest approach is to treat any non-functional call (function or method with side-effects) as potentially reentrant.  

We should be able to perform some type of static analysis.  The goal would be to prove that all member state changes occur before invocation anything that is not purely functional (no side effects).  Since signal emits can call any arbitrary code, they are certainly not purely functional.

I did a quick search of static analysis tools for Python, and my google-foo has not found any such tool yet.  I may not be using the right terms to search for this issue.  We could use a call graph tool (I found a few, no recommendation yet, but [pyan](https://github.com/davidfraser/pyan) looks interesting) with manual inspection, but that would be tedious.  We may be able to hack the call graph code for automatically detecting our cases of interest.

Are you aware of any suitable tools?

j9ac9k:
Constructing a static analysis test for this sort of thing would be great. I can't think of any tools off the top of my head. 

I will be wanting to merge this PR but I want to see if there are any other easily identifiable cases.

j9ac9k:
Was on my phone earlier, so I apologize for the short reply.  Thank you for the detailed reply, and what you describe is content I have observed within my own applications.  If I was still in my grad program I would have proposed the creation of some kind of static analysis tool to serve this purpose.

@campagnola may have more to say about this; I'm going to ask he review/comment these changes before merging.

mliberty1:
@j9ac9k I agree that creating a static analysis test for this issue would be the best way to be certain.  I could also use it on my own code Qt/pyqtgraph code!  Unfortunately, this is likely couple of days work for me just to see how difficult it is going to be.  I don't have that amount of time now, but I may later in  July/August.  Also fixing these more complicated cases would likely take more time.  If you and @campagnola agree, we can merge in these easy cases and follow up with the more complicated ones as we find the time.

j9ac9k:
@mliberty1 We're currently trying to triage failures in our test suite (and if you look at our CI, we have no shortage of them!) and get ready for another stable release.  What you are describing is a bug no doubt, but not something we're in a position to actively track down.  I agree that the "easy cases" you identified should be merged (provided Luke doesn't have an issue) and then go from there.  I have no intention on holding this PR until _all_ cases are addressed.

campagnola:
These changes look good to me, but they definitely need to be tested carefully (manually, probably by me unless someone else feels comfortable with that). There's a good chance that something downstream of these signals also has assumptions baked in about the order in which updates occur. ViewBox should be considered especially brittle..

j9ac9k:
> These changes look good to me, but they definitely need to be tested carefully (manually, probably by me unless someone else feels comfortable with that). There's a good chance that something downstream of these signals also has assumptions baked in about the order in which updates occur. ViewBox should be considered especially brittle..

Would it be worth the effort in attempting to write tests that ensure only "one update" is made on the viewbox for various inputs?

mliberty1:
@campagnola I agree that ViewBox seems pretty brittle.  #907 fixed a specific case for my application, and this should fix a few other potential issues.  However, I don't know what assumptions even exist in other code, so that likely needs some of your attention.  Thanks, and thanks to both you and @j9ac9k for your work on this project!

j9ac9k:
Merging, thanks for the PR @mliberty1

j9ac9k:
@mliberty1 Figured I would post here, but if you're curious about number of signals being emitted, PyQt5 makes `QtTest.QSignalSpy` available which tests for this exact thing (this module is not available on PySide2 currently).

You can read more about it here, but a further down the road feature I would like to have is to implement his signal for testing various ViewBox operations to ensure that signals are only emitted once (or whatever is appropriate)

https://doc.qt.io/qt-5/qsignalspy.html#details

PR #905, opened 2019-05-29
Exclude selected examples from tests (such as HDF5)

2xB:
This pull request is following from the discussions to PR #891 and #892 and issue #882.
It adds an array of examples that are excluded from testing.

Closes #882.

PR #911, opened 2019-06-02
Overhaul testExamples

j9ac9k:
This PR filters down the example tests run by not even creating parametrized tests for frontends we don't have installed (so no needless skipping), and allowing to skip testing examples if some non-standard dependency is not installed.  For example, `hdf5.py` requires `h5py` to be installed, and with this PR, when the testing gets to `hdf5.py` example, and sees that `h5py` is not installed, the test will be skipped.  What separates this PR from #905 is that `hdf5.py` will run if `h5py` is installed.  This represents a desired behavior IMO.

Furthermore, this PR makes the testing of example code better integrated with pytest, and by extension, our CI systems (similar to that of #890).  Failures in the examples are now registered by our CI (as seen by the [results in my fork](https://dev.azure.com/j9ac9k/pyqtgraph/_build/results?buildId=157)).  

Also some code has been shuffled around such that test code now resides in test files.  Prior to this PR, a user could test all the examples by running `python -m pyqtgraph.examples --test` (as far as I'm aware, this was an undocumented feature).  This functionality is now removed, in part that if anyone wants to do testing, they should utilize `pytest` as with the rest of our suite.  Now, the equivalent command is `python -m pytest -k "testExamples" -sv`.

Lastly, duplicate entries were also identified in the example application, and have been removed.

Tagging @2xB as they have done some work in this area of the library, I'd like their input.   Also tagging @asottile to thank him for showing me how to skip certain parametrized tests.


j9ac9k:
Also throwing it out there, I'm not a huge fan of how i'm doing the criteria for skipping certain tests (where I list any weird dependencies and then check if they are present).  I am thinking we may want a more robust method of deciding to skip tests.  If anyone has a suggestion I'm all ears.

j9ac9k:
Added `[WIP]` tag as I wanted to restructure a bit more, but not sure if I'm going to get the time today...

2xB:
Note that campagnola wrote:
https://github.com/pyqtgraph/pyqtgraph/pull/892#issuecomment-496588932
> These changes look good, but a proper fix for #882 is just to ensure that the h5py example is never run as a test, even if h5py can be imported. That example requires user interaction and creates a large file to read from, all things we don't want running in tests. So I'd also add to `test_examples` a list of the examples that we know we want to skip.

So it may be good to exclude the hdf5 example from the parametrization?

j9ac9k:
Oh, that's right he posted about that earlier, maybe I should have looked that over, I'll make sure that test is consistently skipped with respect to testing.

2xB:
Or you might exclude it completely as it is currently? Then it is not even listed as a test in pytest. But that's a very minor complaint ^^

j9ac9k:
You mean not even skip it with `reason="Requires user interaction"`, but explicitly exclude it from pytest seeing it?

2xB:
Yes, exactly

j9ac9k:
> Yes, exactly.

I am of the mind we should be skipping it (regardless of hdf5 presence in the path) primarily because the example should likely be modified such that it doesn't require user interaction.  That said I don't want to modify the example now but as long as we put a `reason` parameter I think it's fairly safe to just skip.

j9ac9k:
I just made another commit which has two large changes.  The first one, which I think is more important, is what pytest reports during the failure (with verbose logging).

Prior to this commit it would do something like 

```
$ pytest -k "testExamples" -svv
================================================================= test session starts =================================================================
platform darwin -- Python 3.7.3, pytest-4.5.0, py-1.8.0, pluggy-0.12.0 -- /Users/ognyan/.pyenv/versions/3.7.3/envs/pyqtgraph-main/bin/python
cachedir: .pytest_cache
rootdir: /Users/ognyan/Developer/pyqtgraph, inifile: pytest.ini
plugins: forked-1.0.2, faulthandler-1.6.0, xdist-1.28.0, cov-2.7.1
collected 109 items / 47 deselected / 62 selected

# passing tests...
examples/test_examples.py::testExamples[ RemoteSpeedTest.py - PySide2 ] Remote Plotting

Python 3.7.3 (default, Apr 28 2019, 09:33:41)
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 919, in __call__
    return self._handler.callObj(obj=self, args=args, kwds=kwds, **opts)
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 564, in callObj
    return self.send(request='callObj', opts=dict(obj=obj, args=args, kwds=kwds), byteData=byteMsgs, **opts)
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 469, in send
    return req.result()
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 639, in result
    raise ClosedError()
pyqtgraph.multiprocess.remoteproxy.ClosedError
Error in atexit._run_exitfuncs:
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/widgets/RemoteGraphicsView.py", line 168, in close
    self.shm.close()
BufferError: cannot close exported pointers exist

FAILED

# this is the part we care about!!!
# more passing tests. ...
examples/test_examples.py::testExamples[ scrollingPlots.py - PySide2 ] Scrolling plots PASSED
examples/test_examples.py::testExamples[ hdf5.py - PySide2 ] SKIPPED
...


# Followed by completely irrelevant traceback
====================================================================== FAILURES =======================================================================
____________________________________________________ testExamples[ RemoteSpeedTest.py - PySide2 ] _____________________________________________________


frontend = 'PySide2', f = ('Remote Plotting', 'RemoteSpeedTest.py'), graphicsSystem = ''

    @pytest.mark.parametrize(
    	"frontend, f",
    	[
    		pytest.param(
    			frontend,
                f,
                marks=pytest.mark.skipif(conditionalExampleTests[f[1]]["conditional"] is False,
                                         reason=conditionalExampleTests[f[1]]["reason"]) if f[1] in conditionalExampleTests.keys() else (),
    		)
    		for frontend, f, in itertools.product(installed, files)
    	],
        ids = [" {} - {} ".format(f[1], frontend) for frontend, f in itertools.product(installed, files)]
    )
    def testExamples(frontend, f, graphicsSystem=None):
        # runExampleFile(f[0], f[1], sys.executable, frontend)

        name, file = f
        global path
        fn = os.path.join(path,file)
        os.chdir(path)
        sys.stdout.write("{} ".format(name))
        sys.stdout.flush()
        import1 = "import %s" % frontend if frontend != '' else ''
        import2 = os.path.splitext(os.path.split(fn)[1])[0]
        graphicsSystem = '' if graphicsSystem is None else "pg.QtGui.QApplication.setGraphicsSystem('%s')" % graphicsSystem
        code = """
    try:
        %s
        import initExample
        import pyqtgraph as pg
        %s
        import %s
        import sys
        print("test complete")
        sys.stdout.flush()
        import time
        while True:  ## run a little event loop
            pg.QtGui.QApplication.processEvents()
            time.sleep(0.01)
    except:
        print("test failed")
        raise

    """ % (import1, graphicsSystem, import2)
        if sys.platform.startswith('win'):
            process = subprocess.Popen([sys.executable],
                                        stdin=subprocess.PIPE,
                                        stderr=subprocess.PIPE,
                                        stdout=subprocess.PIPE)
        else:
            process = subprocess.Popen(['exec %s -i' % (sys.executable)],
                                       shell=True,
                                       stdin=subprocess.PIPE,
                                       stderr=subprocess.PIPE,
                                       stdout=subprocess.PIPE)
        process.stdin.write(code.encode('UTF-8'))
        process.stdin.close() ##?
        output = ''
        fail = False
        while True:
            try:
                c = process.stdout.read(1).decode()
            except IOError as err:
                if err.errno == errno.EINTR:
                    # Interrupted system call; just try again.
                    c = ''
                else:
                    raise
            output += c

            if output.endswith('test complete'):
                break
            if output.endswith('test failed'):
                fail = True
                break
        time.sleep(1)
        process.kill()
        #res = process.communicate()
        res = (process.stdout.read(), process.stderr.read())
        if fail or 'exception' in res[1].decode().lower() or 'error' in res[1].decode().lower():
            print(res[0].decode())
            print(res[1].decode())
>           assert False, "Example Test Failed"
E           AssertionError: Example Test Failed
E           assert False

test_examples.py:145: AssertionError
```

The trackback at the end was completely not relevant, as it showed the function used to run the test example file, not the traceback that was in-line earlier (which is the part we want).

Now, instead we get the meaningful information


```
$ pytest -k "testExamples" -svvq
================================================================= test session starts =================================================================
platform darwin -- Python 3.7.3, pytest-4.5.0, py-1.8.0, pluggy-0.12.0 -- /Users/ognyan/.pyenv/versions/3.7.3/envs/pyqtgraph-main/bin/python
cachedir: .pytest_cache
rootdir: /Users/ognyan/Developer/pyqtgraph, inifile: pytest.ini
plugins: forked-1.0.2, faulthandler-1.6.0, xdist-1.28.0, cov-2.7.1
collected 49 items / 47 deselected / 2 selected

examples/test_examples.py::testExamples[ hdf5.py - PySide2 ] SKIPPED
examples/test_examples.py::testExamples[ RemoteSpeedTest.py - PySide2 ] Remote Plotting

Python 3.7.3 (default, Apr 28 2019, 09:33:41)
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 919, in __call__
    return self._handler.callObj(obj=self, args=args, kwds=kwds, **opts)
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 564, in callObj
    return self.send(request='callObj', opts=dict(obj=obj, args=args, kwds=kwds), byteData=byteMsgs, **opts)
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 469, in send
    return req.result()
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 639, in result
    raise ClosedError()
pyqtgraph.multiprocess.remoteproxy.ClosedError
Error in atexit._run_exitfuncs:
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/widgets/RemoteGraphicsView.py", line 168, in close
    self.shm.close()
BufferError: cannot close exported pointers exist

FAILED

====================================================================== FAILURES =======================================================================
____________________________________________________ testExamples[ RemoteSpeedTest.py - PySide2 ] _____________________________________________________


Python 3.7.3 (default, Apr 28 2019, 09:33:41)
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 919, in __call__
    return self._handler.callObj(obj=self, args=args, kwds=kwds, **opts)
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 564, in callObj
    return self.send(request='callObj', opts=dict(obj=obj, args=args, kwds=kwds), byteData=byteMsgs, **opts)
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 469, in send
    return req.result()
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 639, in result
    raise ClosedError()
pyqtgraph.multiprocess.remoteproxy.ClosedError
Error in atexit._run_exitfuncs:
  File "/Users/ognyan/Developer/pyqtgraph/pyqtgraph/widgets/RemoteGraphicsView.py", line 168, in close
    self.shm.close()
BufferError: cannot close exported pointers exist

Failed Remote Plotting Example Test Located in RemoteSpeedTest.py
================================================= 1 failed, 1 skipped, 47 deselected in 3.39 seconds ==================================================
```

pytest reports in the "failure" the traceback we actually care about.


The second part of the commit that is of relevance is determining what tests to skip, specifically these lines of code:

```python
exceptionCondition = namedtuple("exceptionCondition", ["condition", "reason"])
conditionalExampleTests = {
    "hdf5.py": exceptionCondition(False, reason="Example requires user interaction and is not suitable for testing")
}
```

I propose having a dictionary of filenames (for the examples), followed by a value of a `namedtuple` I call `exceptionCondition` which contains a boolean expression to be evaluated that determines if this file should be skipped, and a `reason` parameter.

In the case of `hdf5.py`, we want to skip this test always because this test requires user interaction.

Generally speaking, this method will give us a consistent interface for saying "hey, we don't want to run this test on Windows, (replace `False` with `platform.system() != "Windows"`) or something like that.

j9ac9k:
I removed the `[WIP]` tag, however folks here should know that the `RemoteGraphicsView.py` example that fails now actually registers as a failure in the test suite.  It appears to pass on some platforms, but not on others (so the CI looks much worse).

examples/test_examples.py
```diff
@@ -50,16 +49,101 @@ def flaky_print(*args):
         frontends[frontend] = True
     except ImportError:
         pass
+    except ModuleNotFoundError:
```
2xB: ˋModuleNotFoundErrorˋ is a subclass of ˋImportErrorˋ, so excepting ˋImportErrorˋ should also catch ˋModuleNotFoundErrorˋ. More problematic, ˋModuleNotFoundErrorˋ is not available in Python 2.7, so this will actually raise a NameError for ModuleNotFoundError once an exception is handled.
It does not fail currently because Python is so dynamic it would allow you to define the class ModuleNotFoundError inside the try block and excepting it then directly after.

examples/test_examples.py
```diff
@@ -50,16 +49,101 @@ def flaky_print(*args):
...
+            break
+    time.sleep(1)
+    process.kill()
+    #res = process.communicate()
+    res = (process.stdout.read(), process.stderr.read())
+    if fail or 'exception' in res[1].decode().lower() or 'error' in res[1].decode().lower():
+        print(res[0].decode())
+        print(res[1].decode())
+        pytest.fail("{}\n{}\nFailed {} Example Test Located in {} ".format(res[0].decode(), res[1].decode(), name, file), pytrace=False)
+    assert True
```
2xB:
```suggestion
```
ˋassert Trueˋ has no effect at all

examples/test_examples.py
j9ac9k: Wasn't sure if catching `ImportError` would catch `ModuleNotFoundErrors`s or not...regardless, each enough to remove if the `ImportError` is sufficient to catch it.

examples/test_examples.py
j9ac9k: I'm a pytest noob, haven't used it much before working on here...does pytest report a test as passing if it gets through a function with no assert failures?

2xB: Yes, as long as no other error is thrown. Pytest does not even notice passing asserts - `assert expression` is an unlatered Python keyword that just throws an `AssertionError` if `expression` is evaluated to be `False`.

examples/test_examples.py
2xB: Have to admit that I also stumbled across this when making the hdf5 PRs, but turns out that ModuleNotFoundError is a fancier ImportError (it has an attribute "name" containing the name of the module that was not found): https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError

j9ac9k: be0e95a incorporates this change

examples/test_examples.py
j9ac9k: be0e95a incorporates this change

j9ac9k:
> Looks good!
> The duplicate example entries were deliberate, but I think it's fine either way.

Thanks for the look over Luke, is it deliberate with the intention of belonging in multiple categories on the examples app/window, or some other reason?

If you still want the entries duplicated on the examples app, I could easily convert all the test files to a set to weed out duplicates and not effect the example app itself.

campagnola:
The duplicates were just because I wanted those examples to be prominently displayed in the menu, but I also wanted the longer `Widgets` and `Graphics Items` menus to be complete.

j9ac9k:
In that case, I'll undo the changes to the example app, but have the tests go off of a set, not a list (to remove the duplicates).

j9ac9k:
Removed `[WIP]` tag as I now restored the previous duplicate entries and modified test_examples to test the ordered list of a set of the example files (to exclude duplicates, but have the test order be reproducible).

At this point I think this is ready for merging.

j9ac9k:
I force-pushed some undoing to extra whitespaces that somehow snuck in.  Will be merging this momentarily; just as a heads up everyone, a lot more of our CI builds will break due to some of the examples actually failing.

PR #917, opened 2019-06-09
Allow AxisItem linked view to be unlinked.

mliberty1:
The AxisItem.linkToView() method allows the Axis instance to be associated with a ViewBox instance.  The linked view is initially None, but the current implementation does not support unlinking the view (returning the linked view to None) after one view is linked.  This pull request adds support for linkToView(None) to unlink the active view.

**What is changed**
The linkToView() method checks the parameter view, and if None sets _linkedView to None (no weakref).  Where view would normally be connected to signals, it is first checked to not be None.

**Motivation**
In my application, I have a single x-axis AxisItem essentially serving multiple ViewBoxes which are added and removed dynamically.  The AxisItem is linked to one "primary" ViewBox, and all other ViewBoxes are then linked to that "primary" through setXLink().  If the user wants to remove all ViewBoxes, the existing AxisItem implementation has no clean way of unlinking the "primary" ViewBox from the AxisItem.


j9ac9k:
I think this is a good idea; would want some other folks to chime in as well before merging, but in the meantime would you be willing to write some tests to cover this case and ensure everything is behaving as it should?

mliberty1:
I'll take a look at writing some unit tests on AxisItem in the next couple days.

mliberty1:
Quick status update: I have been too busy working on shipping a product to get back to this.  I do intend to implement some unit tests and update this PR, but it will likely not be until mid-July.

j9ac9k:
> Quick status update: I have been too busy working on shipping a product to get back to this. I do intend to implement some unit tests and update this PR, but it will likely not be until mid-July.

Take your time, we have quite the backlog ourselves and our first priority is getting our CI suite to be green across the board (we're passing on 9/18 pipelines right now).  

This is a good feature that I do want to get it merged.

mliberty1:
After 10 months, I finally got back to this feature!  I added unit tests, which fully exercise both the view link and view unlink conditions.

mliberty1:
I added the docstring to note that `linkToView(None)` unlinks the current view.  As for `unlinkView`, that is a matter of style.  I don't care either, but I can certainly add it if you want.

j9ac9k:
Hi @mliberty1 looks like you merged `develop` into this branch, effectively making it so there is no diff for your PR.  I'm assuming this was accidental, in which case would you mind either backing up that merge or creating a new PR?

mliberty1:
Thanks for noticing!  I had forgotten about this pull request.  I think that the full intent of this pull request is already merged along with https://github.com/pyqtgraph/pyqtgraph/pull/1154 and https://github.com/pyqtgraph/pyqtgraph/pull/1179.  Closing.

PR #921, opened 2019-06-17
Prevent element-wise string comparison

### LINKED ISSUE #835: Numpy FutureWarning when setting bins in ImageItem.getHistogram()
If I set the `bins` parameter of the `ImageItem.getHistogram()` method to a Numpy array, I get the following Future warning...

```
/Users/kenter/prog/py/pyqtgraph/pyqtgraph/graphicsItems/ImageItem.py:493: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  if bins == 'auto':
```

The solution seems to me to replace `if bins == 'auto'` with `if bins is 'auto'`. 

I didn't make a pull request since it's only one line, but if you like I can make a PR. Just let me know.

I use Numpy 1.15.4 and PyQtGraph 0.10.

2xB:
Issue #835 shows that comparing `bins`, which may be a numpy array, with a string `'auto'` leads to element-wise comparison. This is because the `==` operator for numpy arrays is used. With this Pull Request, potential array and string are switched, so the `==` operator for strings is used, which does no element-wise comparison.

Closes #835

### DIFF
--- a/pyqtgraph/graphicsItems/ImageItem.py
+++ b/pyqtgraph/graphicsItems/ImageItem.py
@@ -488,7 +488,7 @@ class ImageItem(GraphicsObject):
             step = (step, step)
         stepData = self.image[::step[0], ::step[1]]
 
-        if bins == 'auto':
+        if 'auto' == bins:
             mn = np.nanmin(stepData)
             mx = np.nanmax(stepData)
             if mx == mn:


j9ac9k:
This looks good to me, I'll merge.

PR #922, opened 2019-06-17
Curve fill: draw line around patch 

SamSchott:
This is a matter of personal taste, but it sometimes looks a lot more refined when drawing a line around the fill-patch. For instance this example, created with pyqtgraph:

mercurygui screenshot

### DIFF
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -507,7 +507,6 @@ class PlotCurveItem(GraphicsObject):
         else:
             p.drawPath(path)
         profiler('drawPath')
-        profiler('drawPath')
         
         #print "Render hints:", int(p.renderHints())
         #p.setPen(QtGui.QPen(QtGui.QColor(255,0,0)))
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -644,9 +644,9 @@ class AxisItem(GraphicsWidget):
             if dif / intervals[minorIndex] <= maxTickCount:
                 levels.append((intervals[minorIndex], 0))
             return levels
-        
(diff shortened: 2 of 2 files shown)


2xB:
Would this prevent you from different fill and line colors? Then I would suggest to make it optional.

Btw. if you can, it would be great if you created patches fresh from the develop branch, because then only the commits really associated with the pull request show up.

SamSchott:
> Would this prevent you from different fill and line colors? Then I would suggest to make it optional.

No, it wouldn't. Fill and line colors will still be independent, but the line will be drawn all around the fill in addition to the plot itself. This is really a personal preference, so it may be a good idea it make it optional anyways.

> Btw. if you can, it would be great if you created patches fresh from the develop branch, because then only the commits really associated with the pull request show up.

Of course, sorry about that.

pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -502,7 +502,11 @@ def paint(self, p, opt, widget):
             p.setPen(sp)
             p.drawPath(path)
         p.setPen(cp)
-        p.drawPath(path)
+        if self.fillPath is not None:
+            p.drawPath(self.fillPath)
+        else:
+            p.drawPath(path)
+        profiler('drawPath')
```
j9ac9k: Mind removing the duplicate call to `profiler('drawPath')`?

SamSchott: Done.

j9ac9k:
Thanks for the PR @SamSchott LGTM, merging

2xB:
Since I saw this now "in the real world": For scientific visualizations a plotted line has a very precise meaning and multiple lines in the same color can e.g. mean a bifurcation. Therefore, I'd like to go with @SamSchott s suggestion of making it optional.

PR #932, opened 2019-06-18
Fix incorrect clipping of horizontal axis when stopAxisAtTick is set

### LINKED ISSUE #930: stopAxisAtTick doesn't work for horizontal axes
Settings axes like this

```python
axis = self.plot.getAxis("bottom")
axis.setStyle(stopAxisAtTick=(True, True))

axis = self.plot.getAxis("left")
axis.setStyle(stopAxisAtTick=(True, True))
```

results in:

Screenshot 2019-06-18 at 17 56 34

The bottom axis starts at the left margin and ends at 0.

I suppose the bug is in `AxisItem.generateDrawSpecs`:

```python
        ...
        if self.style['stopAxisAtTick'][0] is True:
            stop = max(span[0].y(), min(map(min, tickPositions)))
            if axis == 0:
                span[0].setY(stop)
            else:
                span[0].setX(stop)
        ...
```

If `axis == 1`, it should consider `span[0].x()` instead of `span[0].y()` (and same for the other side).

janezd:
Fixes #930.

Horizontal axes are incorrectly clipped when `stopAxisAtTick` is set to `(True, True)`. This happens because the code always takes the vertical coordinate of the span even if the axis is horizontal.

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -900,16 +900,20 @@ class AxisItem(GraphicsWidget):
 
 
         if self.style['stopAxisAtTick'][0] is True:
-            stop = max(span[0].y(), min(map(min, tickPositions)))
+            minTickPosition = min(map(min, tickPositions))
             if axis == 0:
+                stop = max(span[0].y(), minTickPosition)
                 span[0].setY(stop)
             else:
+                stop = max(span[0].x(), minTickPosition)
                 span[0].setX(stop)
         if self.style['stopAxisAtTick'][1] is True:
-            stop = min(span[1].y(), max(map(max, tickPositions)))
+            maxTickPosition = max(map(max, tickPositions))
             if axis == 0:
+                stop = min(span[1].y(), maxTickPosition)
                 span[1].setY(stop)
             else:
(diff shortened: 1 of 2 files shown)


janezd:
Build fails, but it doesn't look like my fault. If I need to fix something, please tell me so.

j9ac9k:
@janezd Thanks for the pull request.  As you noticed, CI builds are failing for unrelated reasons, which is where the bulk of our attention is going toward currently.

The code change looks fine, I am curious if it is possible to have a test that programmatically detects the incorrect clipping; where it would fail currently, but would pass with this merge request.  If you have an idea about that, would you feel comfortable adding a test case to this PR?

janezd:
No, I'm not comfortable with writing pytests (I always use unittest instead), but I wrote them anyway. :)

Everything happens within `paint`, so I had to monkeypatch the drawing function to see where it puts the axis. Please check whether my mapping of coordinates is correct. It passes the tests, but I'm not sure this is the most appropriate mapping function to call.

j9ac9k:
Thanks for doing this! I will run this myself in the coming days (have some major work related deadlines right now).

j9ac9k:
Thank you again for the PR, this looks great, merging.

PR #935, opened 2019-06-20
PlotItem.addLine now accepts 'pos' and 'angle' parameters

### LINKED ISSUE #70: Bug in PlotItem.py: addLine(pos=user_pos, angle=user_angle) cannot work
Compare 
https://github.com/pyqtgraph/pyqtgraph/blob/develop/pyqtgraph/graphicsItems/PlotItem/PlotItem.py#L570
specifying the "angle" kwd on the addLine() call will lead to `InfiniteLine.__init__(pos, angle, **kwds)` being called with two angle keywords (the first, explicit one, and the one in **kwds).

Suggestion: 
instead of 

```
        angle = kwds.get('angle', 0 if x is None else 90)
```

do 

```
        kwds['angle'] = kwds.get('angle', 0 if x is None else 90)
```

2xB:
The issue and this solution are discussed in issue #70:
> specifying the "angle" kwd on the addLine() call will lead to `InfiniteLine.__init__(pos, angle, **kwds)` being called with two angle keywords (the first, explicit one, and the one in **kwds).

Closes #70

### DIFF
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -545,9 +545,9 @@ class PlotItem(GraphicsWidget):
         :func:`InfiniteLine.__init__() <pyqtgraph.InfiniteLine.__init__>`.
         Returns the item created.
         """
-        pos = kwds.get('pos', x if x is not None else y)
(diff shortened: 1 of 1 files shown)


j9ac9k:
Merging

Issue #881, opened 2019-05-23
[CI-fail] RemoteSpeedTest.py on macOS and Linux (Windows passes)


Dschoni:
On TravisCI remote plotting fails on python 2.7 with pyqt 4 with the following stacktrace:

```
Python 2.7.16 |Anaconda, Inc.| (default, Mar 14 2019, 21:00:58) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> >>> ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... Traceback (most recent call last):
  File "/home/travis/repo-clone/pyqtgraph/multiprocess/remoteproxy.py", line 919, in __call__
    return self._handler.callObj(obj=self, args=args, kwds=kwds, **opts)
  File "/home/travis/repo-clone/pyqtgraph/multiprocess/remoteproxy.py", line 564, in callObj
    return self.send(request='callObj', opts=dict(obj=obj, args=args, kwds=kwds), byteData=byteMsgs, **opts)
  File "/home/travis/repo-clone/pyqtgraph/multiprocess/remoteproxy.py", line 469, in send
    return req.result()
  File "/home/travis/repo-clone/pyqtgraph/multiprocess/remoteproxy.py", line 637, in result
    while not self.hasResult():
  File "/home/travis/repo-clone/pyqtgraph/multiprocess/remoteproxy.py", line 655, in hasResult
    self.result(block=False)
  File "/home/travis/repo-clone/pyqtgraph/multiprocess/remoteproxy.py", line 648, in result
    self._result = self.proc.getResult(self.reqId)  ## raises NoResultError if result is not available yet
  File "/home/travis/repo-clone/pyqtgraph/multiprocess/remoteproxy.py", line 486, in getResult
    self.processRequests()
  File "/home/travis/repo-clone/pyqtgraph/multiprocess/processes.py", line 371, in processRequests
    RemoteEventHandler.processRequests(self)
  File "/home/travis/repo-clone/pyqtgraph/multiprocess/remoteproxy.py", line 126, in processRequests
    while self.conn.poll():
IOError: poll() gave POLLNVAL or POLLERR
Traceback (most recent call last):
  File "/home/travis/repo-clone/pyqtgraph/multiprocess/processes.py", line 371, in processRequests
    RemoteEventHandler.processRequests(self)
[...]

PR #937, opened 2019-06-20
Skip some test examples

j9ac9k:
This PR skips two tests

1. The remote plotting test is skipped, this is failing pretty consistently, and will likely take a fair amount of work/investigation to diagnose.
2. `optics_demo.py` is skipped when running with pyside1, this is due to a known bug that is even linked as a comment in the code.

### DIFF
--- a/examples/test_examples.py
+++ b/examples/test_examples.py
@@ -56,7 +56,7 @@ exceptionCondition = namedtuple("exceptionCondition", ["condition", "reason"])
 conditionalExampleTests = {
     "hdf5.py": exceptionCondition(False, reason="Example requires user interaction and is not suitable for testing"),
     "RemoteSpeedTest.py": exceptionCondition(False, reason="Test is being problematic on CI machines"),
-    "optics_demos.py": exceptionCondition(frontends[Qt.PYSIDE], reason="Test fails due to PySide bug: https://bugreports.qt.io/browse/PYSIDE-671")
+    "optics_demos.py": exceptionCondition(not frontends[Qt.PYSIDE], reason="Test fails due to PySide bug: https://bugreports.qt.io/browse/PYSIDE-671")
 }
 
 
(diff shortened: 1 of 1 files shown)


j9ac9k:
Huh, for some reason `optics_demos.py` was not being skipped, do not merge until this is resolved.

j9ac9k:
had the skip logic backwards, I think this is ready for merging, but will leave up for anyone else to comment

also how about that, 5 successful checks!

j9ac9k:
I'm on the fence here, I hate skipping `RemoteSpeedTest.py`"because it's always failing" but I have no knowledge of multiprocessing code to to try and address it.

Also, should we not be making use of `QThread` for any kind of multithreading as it is, and thus consider deprecating `RemoteSpeedTest.py`?

Pinging @2xB @ixjlyons and @campagnola for thoughts (anyone else should feel free to chime in as well).

2xB:
I'd suggest to skip it for now but to keep issue #881 open containing the remark that it is now skipped but one can access old logs like https://dev.azure.com/pyqtgraph/pyqtgraph/_build/results?buildId=168&view=logs to see it. This does not seem like a central feature of pyqtgraph, and having a succeeding CI pipeline on `develop` is very helpful. Nevertheless it should at some point be tackled.

For the `QThread` suggestion, I think the point with this example is to show that one is able to generate data and plot in two different processes with different namespaces, not only threads - Probably the idea was to demonstrate that one could have one application generating data and sending it asynchronously to a pyqtgraph application?

j9ac9k:
I'll preface by saying I really don't understand the multiprocessing code within pyqtgraph so I could be way off base, but doesn't `QThread` effectively create a new namespace as well, where the only way of getting data across the threads is using signals?

I'll merge this for now, as @campagnola has suggested we skip this test due to its consistent failures, and I think leaving #881 open is the right call but we _should_ get this example working.

Issue #928, opened 2019-06-18
[CI-fail] Windows fails on GLImageItem.py


j9ac9k:
On windows images, we get failures for `GLImageItem.py`

```
[gw0] win32 -- Python 3.6.8 C:\Miniconda\envs\test-environment-3.6\python.exe



Traceback (most recent call last):

  File "D:\a\1\s\pyqtgraph\opengl\GLViewWidget.py", line 60, in addItem

    item.initializeGL()

  File "D:\a\1\s\pyqtgraph\opengl\items\GLImageItem.py", line 34, in initializeGL

    glEnable(GL_TEXTURE_2D)

  File "C:\Miniconda\envs\test-environment-3.6\lib\site-packages\OpenGL\platform\baseplatform.py", line 409, in __call__

    return self( *args, **named )

  File "C:\Miniconda\envs\test-environment-3.6\lib\site-packages\OpenGL\error.py", line 232, in glCheckError

    baseOperation = baseOperation,

OpenGL.error.GLError: GLError(

	err = 1282,

	description = b'invalid operation',

	baseOperation = glEnable,

	cArguments = (GL_TEXTURE_2D,)

)



During handling of the above exception, another exception occurred:



Traceback (most recent call last):

  File "<stdin>", line 7, in <module>

  File "D:\a\1\s\examples\GLImageItem.py", line 41, in <module>

    w.addItem(v1)

  File "D:\a\1\s\pyqtgraph\opengl\GLViewWidget.py", line 62, in addItem

    self.checkOpenGLVersion('Error while adding item %s to GLViewWidget.' % str(item))

  File "D:\a\1\s\pyqtgraph\opengl\GLViewWidget.py", line 429, in checkOpenGLVersion

    ver = glGetString(GL_VERSION).split()[0]

  File "C:\Miniconda\envs\test-environment-3.6\lib\site-packages\OpenGL\platform\baseplatform.py", line 409, in __call__

    return self( *args, **named )

  File "C:\Miniconda\envs\test-environment-3.6\lib\site-packages\OpenGL\error.py", line 232, in glCheckError

    baseOperation = baseOperation,

OpenGL.error.GLError: GLError(

	err = 1282,

	description = b'invalid operation',

	baseOperation = glGetString,

	cArguments = (GL_VERSION,)

)


Failed Image Example Test Located in GLImageItem.py 
```

2xB:
[...]

PR #940, opened 2019-06-20
[WIP] Fix checkOpenGLVersion raising exception on gl call

2xB:
As can be seen in this stackoverflow post, seemingly OpenGL remembers errors and throws them as soon as it can, e.g. on the next function call: https://stackoverflow.com/a/36377085/8575607
That might cause errors to be raised in weird places. Fetching errors beforehand circumvents that problem.

(Hopefully) Fixes second error of #928.

This is a WIP and still a PR because I would like to see the according Azure tests. If they prove my assumption, this will be a regular PR.


2xB:
Does not work. Continue with different test in another PR.

PR #942, opened 2019-06-21
Warn if visible GraphicsView is garbage collected

### LINKED ISSUE #936: GraphicsWindow may be garbage collected too early
Consider the following code:

```python3
import pyqtgraph as pg

def plot():
    pg.GraphicsWindow()

plot()

pg.QtGui.QApplication.exec_()
```
I would expect a GraphicsWindow to show up, but it does not. It does when preventing garbage collection:

```python3
import pyqtgraph as pg

def plot():
    return pg.GraphicsWindow()

a = plot()

pg.QtGui.QApplication.exec_()
```

So there is an issue with garbage collection of GraphicsWindows.

2xB:
Python Garbage Collection and Qt windows sometimes produce unexpected behaviour, causing windows to close. This PR aims at providing a warning if this happens so users can react properly.

Test code:
```python
import pyqtgraph as pg

def plot():
    pg.GraphicsWindow()

plot()
```

Warning:
```
C:\[...]\pyqtgraph\__init__.py
test.py:4: RuntimeWarning: Visible window deleted by Python Garbage Collection. To prevent this, store a reference to the window object.
  pg.GraphicsWindow()
```

Closes #936

### DIFF
--- a/pyqtgraph/graphicsItems/tests/test_ImageItem.py
+++ b/pyqtgraph/graphicsItems/tests/test_ImageItem.py
@@ -121,7 +121,7 @@ def test_ImageItem(transpose=False):
     assertImageApproved(w, 'imageitem/resolution_with_downsampling_y', 'Resolution test with downsampling across y axis.')
     assert img._lastDownsample == (1, 4)
     
-    view.hide()
+    w.hide()
 
 def test_ImageItem_axisorder():
     # All image tests pass again using the opposite axis order
--- a/pyqtgraph/graphicsItems/tests/test_ROI.py
+++ b/pyqtgraph/graphicsItems/tests/test_ROI.py
@@ -147,6 +147,7 @@ def check_getArrayRegion(roi, name, testResize=True, transpose=False):
     # allow the roi to be re-used
     roi.scene().removeItem(roi)
 
+    win.hide()
 
 def test_PolyLineROI():
     rois = [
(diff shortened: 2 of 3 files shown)


2xB:
Updated warning:
```
C:\[...]\pyqtgraph\__init__.py
test.py:6: RuntimeWarning: Visible window deleted. To prevent this, store a reference to the window object.
  pg.GraphicsWindow()
```

j9ac9k:
I'm not opposed to merging this, but I would like to get more input, I feel like there may be a consequence here that I'm not seeing.

campagnola:
I do like the extra friendliness added here, but there are a couple of issues:

1. There are plenty of situations where it is totally normal to garbage collect a visible GraphicsView; I personally would be annoyed to have to track these down and silence the warnings.
2. Up until Python 3.4, the presence of a `__del__` method could prevent objects being garbage collected, leading to memory leaks (see https://www.python.org/dev/peps/pep-0442/). This would be a pretty serious break in backward compatibility.

The basic problem here is that for newcomers to python/Qt/pyqtgraph, it is not intuitively obvious that you need to keep a reference to a widget in order for it to remain visible (especially given functions like `pg.plot`, which circumvent this with global variables). I wonder if we can come up with other ways to make this easier, without affecting existing users/projects?

j9ac9k:
> Up until Python 3.4, the presence of a __del__ method could prevent objects being garbage collected, leading to memory leaks (see https://www.python.org/dev/peps/pep-0442/). This would be a pretty serious break in backward compatibility.

Do we even support python 3.4 currently?  We don't test for it (even in `tox`)

2xB:
1. is a good catch, I changed it slightly so it now only warns if the item has no parent widget, therefore being a standalone window. Such a window should not be intended to be garbage collected when visible.

For 2., I made a commit changing `__del__` to only be defined if a recent python version >= 3.4 is used. As far as I understand it, beginning from Python 3.4 this problem does no longer persist.

PR #944, opened 2019-06-21
Update Travis according to new xvfb syntax

2xB:
Travis updated their syntax to start xvfb. This is further described here: https://benlimmer.com/2019/01/14/travis-ci-xvfb/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ env:
     # - PYTHON=3.4 QT=pyside TEST=standard # pyside isn't available for 3.4 with conda
     #- PYTHON=3.2 QT=pyqt5 TEST=standard
 
+services:
+    - xvfb
 
 before_install:
     - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then wget http://repo.continuum.io/miniconda/Miniconda-3.5.5-Linux-x86_64.sh -O miniconda.sh; else wget http://repo.continuum.io/miniconda/Miniconda3-3.5.5-Linux-x86_64.sh -O miniconda.sh; fi
@@ -74,7 +76,6 @@ install:
 before_script:
     # We need to create a (fake) display on Travis, let's use a funny resolution
     - export DISPLAY=:99.0
-    - "sh -e /etc/init.d/xvfb start"
     - /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -screen 0 1400x900x24 -ac +extension GLX +render
 
     # Make sure everyone uses the correct python (this is handled by conda)


2xB:
Notice that Travis still fails, because pytest.ini includes `--faulthandler-timeout`, which is only available when using pytest-faulthandler. This also means that currently just using pytest on the repository leads to the exact same failure.

ixjlyons:
We might as well go ahead and install pytest-faulthandler at the same time pytest-xdist and pytest-cov are installed right?

j9ac9k:
Do we even want to maintain using travis?

ixjlyons:
I don't see a strong reason to keep it. I haven't looked at the Azure pipelines stuff in detail -- are we doing all the same checks there that are being done on Travis currently? From comments in the config (not necessarily saying we should keep all of them):

```
    # Run unit tests
    # check line endings
    # Check repo size does not expand too much
    # Check for style issues
    # Check install works
    # Check double-install fails
    # Check we can import pg
```

The other thing I can think of is that we have Travis configured with a coverage service that integrates with Github to show coverage deltas in pull requests. I personally think that has value and we should be a bit better about adding tests for things as we fix them. Obviously we've been in the "get the current set of tests passing first" mode -- just looking forward a bit.

j9ac9k:
When creating azure pipelines config, I used the unit-testing check as the basis of the pipeline, but didn't implement the linting checks and some of the other checks.  If we're seriously considering ditching travis; that would be a good thing to implement first.

That said, macOS on azure-pipelines is being problematic (with context to the viewbox mapping test); I'm not sure if that same error is persistent with travis or not (AFAIK our travis config just tests linux), but if not, we may want to use Travis just for macOS checks until we get Azure to behave as we intend.

j9ac9k:
Thanks for the PR @2xB I'll merge this, given that we are currently using travis we should maintain it's configuration...the discussion on whether we should ditch travis altogether should likely not reside in this PR.

2xB:
* I like our Travis CI e.g. for the mentioned linting, the flake8 syntax checker and for the fact that it is trivially simple to get own forks of this repository using the same CI, also for the fact that it can simply be used to automate pypi deployment, if one wants to. And for the fact that it currently supports OpenGL ^^
 * No further package (such as pytest-faulthandler) should in my opinion be needed to run pytest on this project on ones own PC - if that is disagreed with, it should at least be documented in the README.

j9ac9k:
You can setup azure pipelines on your own fork easily enough, I have azure pipelines running on my fork... I think it's the same level of difficulty as setting up travis.

[Azure pipelines also supports automated pypi deployment](https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/pypi?view=azure-devops&tabs=yaml)

In terms of the lint checker, we just don't do that in azure pipelines as my goal was to get unit-testing up, we can (and should) add lint-checking to the pipeline build, but we should likely only do that check on 1 of our 18 pipelines (although after we release 0.11, I'm likely going to push to deprecate anything Qt < 5.9 and python < 3.5, so our number of pipelines will decrease).

I can see OpenGL support being a reason to keep using Travis, especially if implementing OpenGL support on macOS/Windows on Azure pipelines becomes problematic.

Regarding test requirements/processes, they're [spelled out in `CONTRIBUTING.md`](https://github.com/pyqtgraph/pyqtgraph/blob/develop/CONTRIBUTING.md#testing-setting-up-a-test-environment).  No pytest plugins are _required_ to test (considering proposing future versions to use pytest-qt, but we're not there yet); but I highly recommend using tox/tox-conda so you can test on a variety of qt-bindings with ease.

2xB:
Well, it seems I my opinion is mainly based on me having most of my experience with Travis...

For the requirements, pytest-faulthandler is at the moment required to test, that's what I try to say. See e.g. [the last Travis build on `develop`](https://travis-ci.org/pyqtgraph/pyqtgraph/jobs/548847482), line 643.

j9ac9k:
> Well, it seems like my opinion is mainly based on me having most of my experience with Travis...
> 
> For the requirements, pytest-faulthandler is at the moment required to test, that's what I try to say. See e.g. [the last Travis build on `develop`](https://travis-ci.org/pyqtgraph/pyqtgraph/jobs/548847482), line 643.

My reluctance w/ travis is their post buy-out engineer layoff a few months ago: https://twitter.com/ReinH/status/1098663375985229825

In terms of feature-set, azure-pipelines does offer more parallel jobs, and conda-forge migrated over right away due to speed related issues with what appears to be pretty good success.

Regarding pytest-faulthanlder, the error you site on travis is indeed what happens when you run pytest with-out pytest-faulthandler as you rightly identified, it's [due to this line right here](https://github.com/pyqtgraph/pyqtgraph/blob/5238c097d59fd57f7e819484ce0ba80af4fc7972/pytest.ini#L6).  To my knowledge, pytest-faulthandler has no other way to set its options; so I'm thinking we should make that argument be explicit when calling pytest within CI systems (so otherwise you don't need it to install it).

2xB:
This is new to me and very sad.

For pytest-faulthandler, I created a feature request to fix this: https://github.com/pytest-dev/pytest-faulthandler/issues/34. As you already got a very fast response on that repository, this might be the easiest solution with which also others can profit for their projects.

PR #945, opened 2019-06-21
Add pytest-faulthandler to Travis

2xB:
Follow-up to PR #944. Currently our `pytest.ini` requires to have `pytest-faulthandler` installed, and as it makes sense to have it installed in a CI, this PR adds it.
This is independent from the question whether it is good to require further packages for running pytest on the project in general.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,6 +63,7 @@ install:
       fi;
     - pip install pytest-xdist # multi-thread py.test
     - pip install pytest-cov # add coverage stats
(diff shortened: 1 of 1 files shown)


j9ac9k:
thanks for this PR, ...honestly I really don't like the addition I made to `pytest.ini` that requires the fault handler; I _really_  dislike the fact that pytest-faulthandler is now a required dependency.

It should be present in `.travis.yml` no doubt, but if someone has a suggestion on how to specify a default value in `pytest.ini` without requiring it to be installed, I'm all ears.

PR #949, opened 2019-06-22
Allow multiline parameters in configparser

### LINKED ISSUE #491: Bug when reading arrays in configfile
Arrays (longer than a few elements) can be written out using configfile.writeConfigFile, but then cannot be read back in using configfile.readConfigFile.

This seems to be because `repr(array)` produces a string with line breaks in it. While `eval(repr(array))` works just fine, configfile chokes while reading it in because it parses each line separately, and we end up with a SyntaxError about unexpected EOF. 

To reproduce:
```
>>> from numpy import *
>>> from pyqtgraph import configfile
>>> arr = arange(20)
>>> arr
array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15, 16,
       17, 18, 19])
>>> configfile.writeConfigFile({'arr':arr}, 'arraytest.cfg')
>>> configfile.readConfigFile('arraytest.cfg')
  File "pyqtgraph/configfile.py", line 64, in readConfigFile
    data = parseString(s)[1]
  File "pyqtgraph/configfile.py", line 168, in parseString
    raise ParseError("Error evaluating expression '%s': [%s: %s]" % (v, ex.__class__.__name__, str(ex)), (ln+1), l)
pyqtgraph.configfile.ParseError: Error parsing config file '/Users/mbkratz/Code/pyqtgraph/arraytest.cfg' at line 1:
arr: array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15, 16,
Error evaluating expression 'array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15, 16,': [SyntaxError: unexpected EOF while parsing (<string>, line 1)]
>>> 
```

2xB:
Currently parameters may be written to multiple lines in config files but cannot be read, as described in the linked issue.

Fixes #491.

### DIFF
--- a/pyqtgraph/configfile.py
+++ b/pyqtgraph/configfile.py
@@ -33,7 +33,7 @@ class ParseError(Exception):
             msg = "Error parsing string at line %d:\n" % self.lineNum
         else:
             msg = "Error parsing config file '%s' at line %d:\n" % (self.fileName, self.lineNum)
-        msg += "%s\n%s" % (self.line, self.message)
+        msg += "%s\n%s" % (self.line, Exception.__str__(self))
         return msg
         #raise Exception()
         
--- a/pyqtgraph/configfile.py
+++ b/pyqtgraph/configfile.py
@@ -35,7 +35,6 @@ class ParseError(Exception):
             msg = "Error parsing config file '%s' at line %d:\n" % (self.fileName, self.lineNum)
         msg += "%s\n%s" % (self.line, Exception.__str__(self))
         return msg
-        #raise Exception()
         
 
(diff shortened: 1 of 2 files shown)


j9ac9k:
@2xB this fix looks right, but I think this is a good opportunity to add a test/increase our test coverage;  

Ideally would be a file`./pyqtgraph/tests/test_configfile.py` and write and then read a long-ish array and assert that it's the same.  Would you mind doing that?

Issue #950, opened 2019-06-22
Moving a Scale Handle Does Not Fire a sigRegionChangeStarted Signal


Ben-Mathews:
Moving a scale handle on a ROI object does not fire a sigRegionChangeStarted signal.  It is possible to create a ROI object and use the scale handle to manipulate the object, and only receive sigRegionChanged and sigRegionChangeFinished signals, but no sigRegionChangeStarted signal.

Possibly similar to [https://github.com/pyqtgraph/pyqtgraph/issues/463](https://github.com/pyqtgraph/pyqtgraph/issues/463)

Example to reproduce:

```python
import pyqtgraph as pg

y = [2, 4, 6, 8, 10, 12, 14, 16, 18, 20]
x = range(0, 10)

plt = pg.plot(x, y, pen='r')
roi = pg.RectROI([5, 5], [10, 10], pen=(0, 9), sideScalers=True)
roi.sigRegionChangeFinished.connect(lambda x: print("sigRegionChangeFinished"))
roi.sigRegionChangeStarted.connect(lambda x: print("sigRegionChangeStarted"))
roi.sigRegionChanged.connect(lambda x: print("sigRegionChanged"))
plt.addItem(roi)

## Start Qt event loop.
if __name__ == '__main__':
    import sys

    if sys.flags.interactive != 1 or not hasattr(pg.QtCore, 'PYQT_VERSION'):
        pg.QtGui.QApplication.exec_()
```

j9ac9k:
Huh, is `roi._moveStarted()` not ever called?  That method should be called during the mouseDragEvent method of the MouesDragHandler.

j9ac9k:
I agree that the PR in question will fix the issue as you identified, but I would like to investigate why the MouseDragHandler isn't handling this case; because if it's not, there may be other issues we are unaware of.

ixjlyons:
Mouse events on the handles go to the `Handle` objects themselves, not `MouseDragHandler`. I think the PR is a legitimate solution.

PR #951, opened 2019-06-22
Fire a sigRegionChangeStarted Signal When Moving a Scale Handle

Ben-Mathews:
Fixes [https://github.com/pyqtgraph/pyqtgraph/issues/950](https://github.com/pyqtgraph/pyqtgraph/issues/950)

Moving a scale handle on a ROI object does not fire a sigRegionChangeStarted signal.  This patch adds the signal emit to handleMoveStarted() to cause the signal to be fired.


j9ac9k:
Thanks for the PR!

Ben-Mathews:
Cool!  Thanks for the great software!

PR #952, opened 2019-06-23
Fix RotateFree handle not rotating freely

### LINKED ISSUE #174: RotateFree handle possible bug
The handle of type 'rf' (look ROI.addRotateFreeHandle) is not changing distance from center of rotation. It works like rotate-only handle (look ROI.addRotateHandle).
###### How to reproduce
- launch the following code:

``` python
import pyqtgraph as pg

QtCore.qInstallMsgHandler(lambda *args: None)   # to suppress "Qpicture: invalid format version 0" warning
                                                #   see: https://groups.google.com/forum/#!msg/pyqtgraph/dYDRWVORxWg/NVV84upli1IJ

app = QtGui.QApplication([])
w = pg.GraphicsWindow(size=(800,800), border=True)

vb = w.addViewBox(row=1, col=0, lockAspect=True)

g = pg.GridItem()
vb.addItem(g)

roi = pg.LineSegmentROI( [[0,0], [1,0]], movable=False  )
roi.addRotateFreeHandle( [2,0], [1,0] )


vb.addItem(roi)
vb.disableAutoRange('xy')
vb.autoRange()


if __name__ == '__main__':
    import sys

```
- try to use last handle (at position (2,0), with circle shape)  as free handle to move it away from position (1,0)
###### Expected result:

the handle can be placed at any distance from center point (1,0)
###### Actual result:

the handle can be moved only around center point
###### Possible error location:

method ROI.movePoint, end of processing types 'r' and 'rf':

``` python
    def movePoint(self, handle, pos, modifiers=QtCore.Qt.KeyboardModifier(), finish=True, coords='parent'):
...
        elif h['type'] in ['r', 'rf']:
...
            if h['type'] == 'rf':
                h['item'].setPos(self.mapFromScene(p1))  ## changes ROI coordinates of handle
```

it's line 678 in version 0.9.8
###### Possible fix:

replace lines 

``` python
            if h['type'] == 'rf':
                h['item'].setPos(self.mapFromScene(p1))  ## changes ROI coordinates of handle
```

with lines:

``` python
            if h['type'] == 'rf':
                h['item'].setPos(self.mapFromParent(p1))  ## changes ROI coordinates of handle
                h['pos'] = self.mapFromParent(p1)
```
###### Possible workaround:

subclassing of ROI class (or it's descendent) and overwrite method movePoint like this:

``` python
    def movePoint(self, handle, pos, modifiers=QtCore.Qt.KeyboardModifier(), finish=True, coords='parent'):
        pg.ROI.movePoint(self, handle, pos, modifiers, finish, coords )

        index = self.indexOfHandle(handle)
        h = self.handles[index]
        if h['type'] == 'rf':
            p0 = self.mapToParent(h['pos'] * self.state['size'])
            p1 = Point(pos)
            if coords == 'parent':
                pass
            elif coords == 'scene':
                p1 = self.mapSceneToParent(p1)

            h['item'].setPos(self.mapFromParent(p1))  ## changes ROI coordinates of handle
            h['pos'] = self.mapFromParent(p1)

```

2xB:
@alguryanow made a very good issue #174 explaining this. In short, handles added via `ROI.addRotateFreeHandle` behaved as if added via `ROI.addRotateHandle`, in that they did not allow the user to drag them towards or away from the center. This fixes that.

Closes #174

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -928,6 +928,7 @@ class ROI(GraphicsObject):
             
             if h['type'] == 'rf':
                 h['item'].setPos(self.mapFromScene(p1))  ## changes ROI coordinates of handle
+                h['pos'] = self.mapFromParent(p1)
                 
         elif h['type'] == 'sr':
             if h['center'][0] == h['pos'][0]:


j9ac9k:
tested this out, LGTM!

PR #953, opened 2019-06-23
Fix HistogramLUTWidget with background parameter

### LINKED ISSUE #175: HistogramLUTWidget kargs bug?
# lut_widget = HistogramLUTWidget(background='w')

 File "/usr/local/lib/python3.4/dist-packages/pyqtgraph-0.9.8-py3.4.egg/pyqtgraph/widgets/HistogramLUTWidget.py", line 18, in **init**
    self.item = HistogramLUTItem(_args, *_kargs)
# TypeError: **init**() got an unexpected keyword argument 'background'

I can fix it by:

class HistogramLUTWidget(pg.GraphicsView):

```
def __init__(self, parent=None,  *args, **kargs):
    # background = kargs.get('background', 'default')
    background = kargs.pop('background', 'default')
```

...

2xB:
HistogramLUTWidget cannot be initialized with the `background` parameter, because all parameters are also passed to the constructor of HistogramLUTItem which does not have a `background` parameter. This pull request fixes that issue by defining `background` explicitly as parameter in the function header.

Closes #175

### DIFF
--- a/pyqtgraph/tests/test_histogramlutwidget.py
+++ b/pyqtgraph/tests/test_histogramlutwidget.py
@@ -41,5 +41,4 @@ def testHistogramLUTWidget():
     w.setImageItem(img)
     
     QtGui.QApplication.processEvents()
-    pg.exit()
     
--- a/pyqtgraph/tests/test_histogramlutwidget.py
+++ b/pyqtgraph/tests/test_histogramlutwidget.py
@@ -1,0 +1,45 @@
+"""
+HistogramLUTWidget test:
+
+Tests the creation of a HistogramLUTWidget.
+"""
+
+import pyqtgraph as pg
+from ..Qt import QtGui
+import numpy as np
+
+def testHistogramLUTWidget():
+    pg.mkQApp()
+    
+    win = QtGui.QMainWindow()
+    win.show()
+
+    cw = QtGui.QWidget()
+    win.setCentralWidget(cw)
+
+    l = QtGui.QGridLayout()
+    cw.setLayout(l)
+    l.setSpacing(0)
+
+    v = pg.GraphicsView()
+    vb = pg.ViewBox()
+    vb.setAspectLocked()
+    v.setCentralItem(vb)
+    l.addWidget(v, 0, 0, 3, 1)
+
+    w = pg.HistogramLUTWidget(background='w')
+    l.addWidget(w, 0, 1)
+
+    data = pg.gaussianFilter(np.random.normal(size=(256, 256, 3)), (20, 20, 0))
+    for i in range(32):
+        for j in range(32):
+            data[i*8, j*8] += .1
+    img = pg.ImageItem(data)
+    vb.addItem(img)
+    vb.autoRange()
+
+    w.setImageItem(img)
+    
+    QtGui.QApplication.processEvents()
+    pg.exit()
+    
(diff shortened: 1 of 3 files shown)


j9ac9k:
Regarding the code change directly, the placement between the background='default' and *args should probably be switches (so *args and **kwargs are at end together).

That said is this the right fix, as opposed to the suggestion of the issue, where we just pop the background item from the kwargs dict?

I haven't used these widgets before, but would it be inappropriate to pass the background color to HistogramLUTItem?

Also, we should add a test case for this as well, I'm not seeing any kind of existing test for these widgets, maybe just to initialize one without an error (and with a background passed to it).

2xB:
> Regarding the code change directly, the placement between the background='default' and *args should probably be switches (so *args and **kwargs are at end together).
That would make it necessary to explicitely state `background`, even if the default is used.
Consider this code:
```python3
def test1(a, background='default', *args, **kwargs):
    print(a, b, args, kwargs)

def test2(a, *args, background='default', **kwargs):
    print(a, b, args, kwargs)
```
Now if you want `args` to be `("abc", "def")`, the shortest function call looks like this:
```python3
test1(1, 'default', "abc", "def")
test2(1, "abc", "def")
```

This is why I chose the same parameter order as seen here in `test2`.

2xB:
> That said is this the right fix, as opposed to the suggestion of the issue, where we just pop the background item from the kwargs dict?

None of both approaches is wrong, they do exactly the same thing (at least with this order of `*args` and `background='default'`). The only technical difference is how this occurs in the documentation, where for this approach the parameter and its default values are printed in the function header. I would call this approach more pythonic, as it uses directly the function parameter default mechanism from Python.

2xB:
> I haven't used these widgets before, but would it be inappropriate to pass the background color to HistogramLUTItem?

`HistogramLUTItem` currently simply can't be initialized with a `background` parameter, as its `__init__` has a fixed set of parameters that do not include this.

2xB:
> Also, we should add a test case for this as well, I'm not seeing any kind of existing test for these widgets, maybe just to initialize one without an error (and with a background passed to it).

Well there exists the example `HistogramLUT.py`, which could be called a test for `HistogramLUTWidget`, using the default background. I can of course add a test case for setting the background color - just added that to my TODO list here.

j9ac9k:
> > That said is this the right fix, as opposed to the suggestion of the issue, where we just pop the background item from the kwargs dict?
> 
> None of both approaches is wrong, they do exactly the same thing (at least with this order of `*args` and `background='default'`). The only technical difference is how this occurs in the documentation, where for this approach the parameter and its default values are printed in the function header. I would call this approach more pythonic, as it uses directly the function parameter default mechanism from Python.

That's a really good point that I hadn't considered; I don't have much experience with `*args` and `**kwargs` so I don't have a strong feel for when one should be explicit with a key word argument or not.

j9ac9k:
> > Also, we should add a test case for this as well, I'm not seeing any kind of existing test for these widgets, maybe just to initialize one without an error (and with a background passed to it).
> 
> Well there exists the example `HistogramLUT.py`, which could be called a test for `HistogramLUTWidget`, using the default background. I can of course add a test case for setting the background color - just added that to my TODO list here.

I don't have a strong preference toward modifying the example to include a background vs. having a specific test for it, I'll leave it up to you to decide what's best.

2xB:
Oh, Python2 does not support `*args` before `background='default'`. Well, then we will have to use pop...

2xB:
I noticed that the test could be placed in `pyqtgraph/widgets/tests`, I'll move it.

j9ac9k:
closing and opening to trigger CI tests

j9ac9k:
CI came back good, thanks for the PR @2xB Merging

Issue #927, opened 2019-06-18
[CI-Warning] Invalid value encountered in multiply


j9ac9k:
Tox (and our CI) generates a warning for invalid value
```
pyqtgraph/functions.py:1108
pyqtgraph/imageview/tests/test_imageview.py::test_nan_image
  /Users/ognyan/Developer/pyqtgraph/pyqtgraph/functions.py:962: RuntimeWarning: invalid value encountered in multiply
    d2 *= scale
```

This issue, like #926 both occur during `test_nan_image`.  This can be reproduced by running `pytest . -sv -k "test_nan_image"`

ixjlyons:
Resolved by #955

Issue #926, opened 2019-06-18
[CI-Warning] Divide by zero encountered in double_scalars


j9ac9k:
Our CI generates the following output:

```
pyqtgraph/functions.py:1108
  /home/vsts/work/1/s/pyqtgraph/functions.py:1108: RuntimeWarning: divide by zero encountered in double_scalars
    data = rescaleData(data, scale/(maxVal-minVal), minVal, dtype=dtype)
```

While the CI system doesn't show it (it should), i think this error occurs during the `test_nan_image` and you can recreate this warning by running `pytest . -sv -k "test_nan_image"`

ixjlyons:
Resolved by #955

PR #955, opened 2019-06-23
Fix infinite scale in makeARGB

ixjlyons:
In case `minVal == maxVal`, the small constant used currently 1e-16 can be too small to effectively make `maxVal - minVal` nonzero. Choosing the next available floating point value should be a robust method of avoiding this, regardless of the magnitude of `maxVal`.

Fixes both #926 and #927


j9ac9k:
huh, TIL I learned about `np.nextafter`...here I had been using `np.finfo(float).eps` like some caveman.

Will merge

PR #838, opened 2019-03-08
`_updateMaxTextSize` to reduce text size when no longer needed

SamSchott:
Currently `_updateMaxTextSize ` will increase the current space required for axis labels, if necessary, but not decrease it when the extra space is no longer needed. The proposed change will release no longer needed space again.

PR #403, opened 2016-11-19
Axis fixes

lesauxvi:
Modification of PR #235 proposed by JenSte taking into account the modification suggested in the discussion of PR #235.

PR #959, opened 2019-06-24
[WIP] Update max text size

ixjlyons:
This takes over from #838 which was reverted for causing the `ErrorBarItem` test to fail. Not sure if there's a cleaner way to re-implement with fixes, but hopefully this works.

I've found that I have to call `app.processEvents()` twice after initializing the plot window to get the test to pass but I'm not sure why (hence the WIP).

Here's a script that's similar to the test case with some sleeps added in so you can see what's happening. You can see that the second `app.processEvents` call causes the view rect to change because the y axis shifts to the left. It makes some sense that this would be caused by the `_updateMaxTextSize` change, but I haven't determined why it takes two cycles.

```python
import pyqtgraph as pg
import numpy as np
import time

app = pg.mkQApp()

x = np.arange(5) + 100
y = np.array([1, 2, 1, 3, 1])
e = np.array([1, 0.1, 0.5, 0.1, 3])

plt = pg.plot()

cur = pg.PlotCurveItem(x=x, y=y)
plt.addItem(cur)
app.processEvents()
print(plt.viewRect())
time.sleep(1)

app.processEvents()
print(plt.viewRect())

err = pg.ErrorBarItem()
plt.addItem(err)
app.processEvents()
print(plt.viewRect())
time.sleep(1)

err.setData(x=x, y=y, top=e, bottom=e)
app.processEvents()
print(plt.viewRect())
time.sleep(1)

err.setData(x=None, y=None)
app.processEvents()
print(plt.viewRect())
time.sleep(1)
```

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -301,18 +301,16 @@ class AxisItem(GraphicsWidget):
     def _updateMaxTextSize(self, x):
         ## Informs that the maximum tick size orthogonal to the axis has
         ## changed; we use this to decide whether the item needs to be resized
-        ## to accomodate.
+        ## to accommodate.
         if self.orientation in ['left', 'right']:
-            mx = max(self.textWidth, x)
-            if mx > self.textWidth or mx < self.textWidth-10:
-                self.textWidth = mx
+            if x > self.textWidth or x < self.textWidth-10:
+                self.textWidth = x
                 if self.style['autoExpandTextSpace'] is True:
                     self._updateWidth()
                     #return True  ## size has changed
         else:
-            mx = max(self.textHeight, x)
-            if mx > self.textHeight or mx < self.textHeight-10:
-                self.textHeight = mx
+            if x > self.textHeight or x < self.textHeight-10:
+                self.textHeight = x
                 if self.style['autoExpandTextSpace'] is True:
                     self._updateHeight()
                     #return True  ## size has changed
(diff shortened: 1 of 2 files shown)


ixjlyons:
Another note: the leftward shift noted in the script above doesn't happen if I run with xvfb-run but it does if I use my usual display or one created by Xephyr.

j9ac9k:
Having views shift around like this is usually an indication of resize events/signals not happening at the correct order....

@mliberty1 identified some places where this was a problem, and kindly made #910 as a follow up to #907 to address this case.  The understanding was that there were likely other cases in the library where this occurred; and `ErrorBarItem` is like a place that's happening would be my guess.

A way to diagnose this would be to implement `QSignalSpy`, but that's somewhat problematic as it's only accessible via PyQt5 bindings (maybe PyQt4, I'm not sure), and not PySide2.

Another option is to set some signals with `Qt.QueuedConnection`, so instead of having the slot code run immediately, it will wait until the event loop finishes and then run the slot code (at least from my understanding).

I'll try and look at this more carefully soon.

2xB:
This seems to be partly a duplicate of #235, which has some discussion including a comment by campagnola why this was not done yet. These thoughts were incorporated into #403.

j9ac9k:
Tagging @samschott to get their input as this was originally they're PR.

SamSchott:
Thank you all for following up on this. Not knowing enough about pyqtgraph, I would have never found the issue in the tests.

Yes, this PR does overlap with #235 and #403. I had also overlooked the issue raise by @campagnola about the axis labels potentially becoming unreadable with rapidly auto-scaling plots. 

A small issue that I have with the proposed solution in #403: Introducing a keyword argument `autoScale` next to the style option `autoExpandTextSpace` can make the settings somewhat confusing. For instance, if  `autoScale == True` but `self.style['autoExpandTextSpace'] == False`, the text size will not be updated, despite expectations. How about an additional style option `autoReduceTextSpace` instead?

PR #960, opened 2019-06-24
Work around PySide setOverrideCursor bug in BusyCursor

ixjlyons:
This should fix failures observed in CI with PySide and `examples/VideoSpeedTest.py` on Linux. See https://bugreports.qt.io/browse/PYSIDE-243. It doesn't look like the issue was ever resolved though at least it doesn't produce a segfault like it did at one time.

I selectively use the workaround for PySide, but passing the [CursorShape](https://doc.qt.io/qt-5/qt.html#CursorShape-enum) directly to [setOverrideCursor](https://doc.qt.io/qt-5/qguiapplication.html#setOverrideCursor) seems to work for all Qt implementations despite not being documented.

### DIFF
--- a/pyqtgraph/widgets/BusyCursor.py
+++ b/pyqtgraph/widgets/BusyCursor.py
@@ -1,4 +1,4 @@
-from ..Qt import QtGui, QtCore
+from ..Qt import QtGui, QtCore, QT_LIB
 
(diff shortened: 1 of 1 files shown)


j9ac9k:
And now we're at 10/18 passing builds, huzza!

PR #961, opened 2019-06-24
[WIP] Fix pg.exit test in case pyqtgraph is not installed

ixjlyons:
Stole the implementation from the other test in `test_exit_crash.py` even though it's skipped. This test should now work on Travis where we don't install pyqtgraph until after the tests are run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,9 +128,14 @@ before_script:
 script:
 
     - source activate test_env
-    
+
     # Check system info
     - python -c "import pyqtgraph as pg; pg.systemInfo()"
+
+    # Check install works
+    - start_test "install test";
+      python setup.py --quiet install;
+      check_output "install test";
 
     # Run unit tests
     - start_test "unit tests";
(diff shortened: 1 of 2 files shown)


ixjlyons:
Alright this isn't working at all. If anyone has an idea, let me know :sweat_smile:

j9ac9k:
I think I got it on my PR here https://github.com/pyqtgraph/pyqtgraph/pull/962 . Feel free to glance over the code changes, but there are two main factors.

1. We actually install the pyqtgraph lib
2. We do not pass `shell=True` to `subprocesss.Popen()`

I'm not sure if `close_fds=True` is necessary or not; I added it in because 🤷‍♂

ixjlyons:
It's not really important, but just to check my understanding I verified you don't need the insert pyqtgraph's root dir into the path if pyqtgraph is installed before running the tests. I guess we don't really want the test code to be making lots of assumptions about how CI is set up, but I think assuming you can import pyqtgraph in a subprocess is reasonable.

@j9ac9k Going to close this since you made a couple other changes to the travis config in the other PR.

j9ac9k:
Sounds good; I think there were a few minor tweaks I wanted to make; will get that merged later today.

PR #962, opened 2019-06-25
[WIP] x2 Ixjlyons fix exit oserror

j9ac9k:
Don't have travis setup on my fork, so need a PR to evaluate on Travis...


ixjlyons:
Looks good to me. I thought about installing before running the tests, but I figured that shouldn't be necessary.

j9ac9k:
Got tangled in my own git mess; looks like things are cleared up now.

I'll wait for CI to finish and ensure it doesn't have a cow after all those rebases and then I'll merge.

PR #963, opened 2019-06-25
Allow last image in stack to be selected by slider in ImageView

### LINKED ISSUE #298: Can't select last image in image stack by drag and drop of slider
Minimum non-working example:

``` python
import numpy as np
import pyqtgraph as pg

a = np.random.rand(3,10,10)
a[2] = 0
pg.image(a)
```

Now, try to drag the slider to the last image (the "0" one). This won't work.
However, navigating there by using the arrow keys just works.

### LINKED ISSUE #751: ImageView not showing the last frame
Quite self explanatory i think. If you load a 3 dimension array into imageView you cant seek to the last frame. However playing the sequence ends up in the last frame.

### LINKED ISSUE #912: Z-axis in pyqtgraph.image() is too short in some examples
```
import pyqtgraph as pg
import numpy as np
from scipy import misc

test_image = np.zeros((3,512,512))
test_image[0] = misc.ascent()
test_image[2] = misc.ascent()
pg.image(test_image)
```

This code should lead to an image stack, with the first and the last image being scipy's ascent test image and the middle image being blank. In the associated viewer, the z-slider at the bottom can be used to scroll through the image stack. The last image in the image stack can never be reached, because the slider stops shortly before the last image is displayed.

ixjlyons:
Fixes #298 and fixes #912 and fixes #751

### DIFF
--- a/pyqtgraph/imageview/ImageView.py
+++ b/pyqtgraph/imageview/ImageView.py
@@ -740,7 +740,7 @@ class ImageView(QtGui.QWidget):
             return (0,0)
         
         t = slider.value()
-        
+
         xv = self.tVals
         if xv is None:
             ind = int(t)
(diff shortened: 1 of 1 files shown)


j9ac9k:
Looks like azure pipelines is having a cow about miniconda builds...

j9ac9k:
You mind merging/rebasing develop?  I have the python2 conda environment fix in place now.

j9ac9k:
LGTM

PR #968, opened 2019-06-27
Fix for AxisItem using old scale to create label

ecejeff:
The self.setLabel() call in updateAutoSIPrefix() was being made before the autoSIPrefixScale was written, which caused the label string to be generated with an incorrect unit in some cases.


j9ac9k:
Thanks for the PR @ecejeff !

PR #970, opened 2019-06-28
Always update transform when setting angle of a TextItem

### LINKED ISSUE #871: Rotating TextItem dosen't work
I want to dynamically rotate TextItem but cannot get it to work. Changing position or anchor with setPos and setAnchor updates the item, but wanting to change angel with setAngle doesn't update the text. The strangest thing is that it does update once I drag the canvas. I am wondering what signal or function to call so that the item updates immediately. 
Below is some code that should move and rotate a TextItem.
```python
import pyqtgraph as pg

app = QtGui.QApplication([])

w = pg.GraphicsView()
w.show()
w.resize(800,800)

view = pg.ViewBox()
w.setCentralItem(view)
view.setAspectLocked(True)
view.setRange(QtCore.QRectF(0, 0, 200, 200))

anchor = pg.TextItem()
anchor.setText("hey")
anchor.setColor(QtGui.QColor(255, 255, 255))

view.addItem(anchor)


def rotate():
    x, y = anchor.pos()
    anchor.setPos(x + 1, y + 1)
    anchor.setAngle(anchor.angle + 10)


timer = QtCore.QTimer()
timer.timeout.connect(rotate)
timer.start(1000)


if __name__ == '__main__':
    import sys
```

ixjlyons:
As demonstrated in #871, calling `TextItem.setAngle()` with a new angle doesn't actually rotate the `TextItem` visually because `updateTransform` uses a cached transform to avoid extra computation if it hasn't changed, but setting the `angle` attr of the `TextItem` doesn't invalidate it.

I decided to use a keyword arg instead of setting `self._lastTransform = None` right before calling `self.updateTransform()` for readability. I don't know if `force` is the right name (maybe `use_cache` with inverted logic?), but it's not really a user method so maybe it doesn't matter much.

Fixes #871

### DIFF
--- a/pyqtgraph/graphicsItems/TextItem.py
+++ b/pyqtgraph/graphicsItems/TextItem.py
@@ -111,7 +111,7 @@ class TextItem(GraphicsObject):
         
     def setAngle(self, angle):
         self.angle = angle
-        self.updateTransform()
+        self.updateTransform(force=True)
         
     def setAnchor(self, anchor):
         self.anchor = Point(anchor)
@@ -169,7 +169,7 @@ class TextItem(GraphicsObject):
             p.setRenderHint(p.Antialiasing, True)
             p.drawPolygon(self.textItem.mapToParent(self.textItem.boundingRect()))
         
-    def updateTransform(self):
+    def updateTransform(self, force=False):
         # update transform such that this item has the correct orientation
         # and scaling relative to the scene, but inherits its position from its
         # parent.
@@ -181,7 +181,7 @@ class TextItem(GraphicsObject):
         else:
             pt = p.sceneTransform()
         
-        if pt == self._lastTransform:
+        if not force and pt == self._lastTransform:
             return
 
         t = pt.inverted()[0]
--- a/pyqtgraph/graphicsItems/TextItem.py
+++ b/pyqtgraph/graphicsItems/TextItem.py
@@ -110,9 +110,16 @@ class TextItem(GraphicsObject):
         self.updateTextPos()
         
     def setAngle(self, angle):
+        """
+        Set the angle of the text in degrees.
+
+        This sets the rotation angle of the text as a whole, measured
+        counter-clockwise from the x axis of the parent. Note that this rotation
+        angle does not depend on horizontal/vertical scaling of the parent.
+        """
         self.angle = angle
         self.updateTransform(force=True)
-        
+
     def setAnchor(self, anchor):
         self.anchor = Point(anchor)
         self.updateTextPos()
--- a/pyqtgraph/graphicsItems/tests/test_TextItem.py
+++ b/pyqtgraph/graphicsItems/tests/test_TextItem.py
@@ -1,0 +1,22 @@
+import pytest
+import pyqtgraph as pg
+from numpy.testing import assert_almost_equal, assert_raises
+
+app = pg.mkQApp()
+
+
+@pytest.mark.parametrize("lock_aspect", [True, False])
+def test_TextItem_setAngle(lock_aspect):
+    plt = pg.plot()
+    plt.setXRange(-1, 1)
+    plt.setYRange(-10, 10)
+    plt.setAspectLocked(lock_aspect)
+    item = pg.TextItem(text="test")
+    plt.addItem(item)
+
+    assert item.transformAngle() == 0.0
+
+    for angle in [10, -30, 90]:
+        item.setAngle(angle)
+        # negative because transformAngle measures from item to parent
+        assert_almost_equal(-item.transformAngle() % 360, angle % 360)
--- a/pyqtgraph/graphicsItems/tests/test_TextItem.py
+++ b/pyqtgraph/graphicsItems/tests/test_TextItem.py
@@ -5,18 +5,20 @@ from numpy.testing import assert_almost_equal, assert_raises
 app = pg.mkQApp()
 
 
-@pytest.mark.parametrize("lock_aspect", [True, False])
-def test_TextItem_setAngle(lock_aspect):
+def test_TextItem_setAngle():
     plt = pg.plot()
-    plt.setXRange(-1, 1)
-    plt.setYRange(-10, 10)
-    plt.setAspectLocked(lock_aspect)
+    plt.setXRange(-10, 10)
+    plt.setYRange(-20, 20)
     item = pg.TextItem(text="test")
     plt.addItem(item)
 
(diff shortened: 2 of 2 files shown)


j9ac9k:
Should there be a test case to catch this?

ixjlyons:
Not sure how to test this well to be honest. One possibility might be to verify that it's position in the scene changes before vs. after `setAngle`. Any ideas to be a little more precise?

j9ac9k:
I suspect we have to accept a certain amount of imprecision in automated testing for things that are normally evaluated virtually.

We don't have any mechanism to ensure a signal has fired do we?

ixjlyons:
pytest-qt makes it a lot nicer but you can always do something like this:

```python
import pyqtgraph as pg

class SigGen(pg.Qt.QtCore.QObject):
    signal = pg.Qt.QtCore.Signal()

class SigTest(object):

    def __init__(self):
        self.received = False

    def slot(self, *args):
        self.received = True

siggen = SigGen()
sigtest = SigTest()

assert not sigtest.received
siggen.signal.connect(sigtest.slot)
siggen.signal.emit() # or other code causing signal to be fired
siggen.signal.disconnect(sigtest.slot)
assert sigtest.received
```

I'm not seeing how that helps in this case though.

j9ac9k:
I definitely think we should make an attempt to get pytest-qt interested.  Can we get some kind of angle information from qyryibg

ixjlyons:
There's `GraphicsItem.transformAngle()` which makes this test pretty straightforward actually. Unfortunately, the `TextItem` rotation doesn't work perfectly unless the aspect ratio of the `ViewBox` is locked to 1. I think this is because the transformation matrix in that case involves a non-uniform scaling.

Demo:

```python
import pyqtgraph as pg

app = pg.mkQApp()

plt = pg.plot()
plt.setXRange(-1, 1)
plt.setYRange(-1, 1)
plt.setAspectLocked(False)

text = pg.TextItem("------------------", anchor=(0.5, 0.5))
line = pg.InfiniteLine(angle=0)
plt.addItem(text)
plt.addItem(line)

angle = 0

print((3*"{:>8s}").format("set", "text", "line"))
print(24*"-")

def rotate():
    global angle
    text.setAngle(angle)
    line.setAngle(angle)
    print((3*"{:8.2f}").format(angle, text.transformAngle(),
                               line.transformAngle()))
    angle += 5
    angle %= 360

timer = pg.Qt.QtCore.QTimer()
timer.timeout.connect(rotate)
timer.start(100)

app.exec_()
```

At this juncture, I'm inclined to merge this and create a new issue for testing/fixing the angle setting. Or maybe I could just add a known-good test case here and we can open an issue to fix rotation when the aspect ratio isn't 1.

j9ac9k:
I'm good with merging and creating a new issue for testing

denivyruck:
One way for testing would be at least ensuring that the transformation forwarded to Qt asserts to `isRotating()`. But for sure an image test would be better

j9ac9k:
@ixjlyons I think I got `transformAngle` to behave as it should.

replace this line:
https://github.com/pyqtgraph/pyqtgraph/blob/3a863fff9ab711ba7db67e9ef83fc9dce77e1448/pyqtgraph/graphicsItems/GraphicsItem.py#L411

with this

```python
tr = self.deviceTransform(relativeItem.viewTransform())
```

I could use a sanity check as it's getting late here.

ixjlyons:
Ok, it turns out my example above comparing an `InfiniteLine` to the `TextItem` missed an important point. For a line (i.e. "data" in the plot), an angle of 45 degrees maps to a slope of 1, so the angle it makes with respect to the edge of your screen should depend on the aspect ratio of the plot. For text, it makes more sense to maintain an "on-screen" angle of 45 degrees regardless of the aspect ratio. Otherwise, the text will be unreadable most of the time. In revisiting this, I've realized `TextItem.updateTransform` was designed specifically to keep the text un-scaled regardless of zoom, so I think it makes sense to disregard scene/parent scaling when it comes to the angle as well.

That sound reasonable? I've added a test that verifies this behavior.

ixjlyons:
Ugh of course it works locally and fails in CI... @j9ac9k does this work locally for you?

ixjlyons:
Huh, something in the develop branch must have changed to cause this to fail. I'm not sure what that would be glancing through the commit history.

ixjlyons:
Looks to me like #1017 must be it. @2xB any insight?

2xB:
@ixjlyons Yes - so the issue here is that before my PR, `ViewBox.updateMatrix` was called through `ViewBox.paint`, which was **not** called before your assert statements. Now, it is called through `ViewBox.update`, which now **is** called before your statements. Therefore, before my PR, your asserts only seemingly have worked, but in reality if you would wait for one single call of `paint`, you would get a different result.

Here is an example to demonstrate this, just run it on both versions of `pyqtgraph`:
```python3
import pyqtgraph as pg

app = pg.mkQApp()

plt = pg.plot()
plt.setXRange(-1, 1)
plt.setYRange(-10, 10)
item = pg.TextItem(text="test")
plt.addItem(item)

item.setAngle(10)
print(item.transformAngle())

app.processEvents()
print(item.transformAngle())
```

You see that after `app.processEvents` is called, the result is the same on every version.

What you probably want to do is to compare against `item.angle` instead of `item.transformAngle()`, as the former one is the transformation that is used for constructing a fresh matrix on `paint` and the latter one is that transformation with the scene transformation applied (that is what causes these numbers there). Always constructing a new transformation in `TextItem.paint` through calling `TextItem.updateTransform` btw. seems to be very unnecessary and hacky to circumvent the scene transformation to being applied, but as the code of TextItem.py states:

https://github.com/pyqtgraph/pyqtgraph/blob/584c4516f0c4386ba01edf6fee2b1f392f88cfaf/pyqtgraph/graphicsItems/TextItem.py#L176-L177

2xB:
PS: The example is meant to be run with this PR. Without it, it obviously does not work.

ixjlyons:
Ok, I don't really see the point in testing the value of `textitem.angle` so I can just take the test out and we can move on. Sound ok?

2xB:
Or you just override `TextItem.transformAngle` to `return self.angle`, which would probably also be cleaner since `TextItem.transformAngle` would then tell the truth. Then the test can stay as a test for the new function.

j9ac9k:
I'm good with removing the test if you feel it's unnecessary

j9ac9k:
Merging this, thanks @ixjlyons !

PR #973, opened 2019-06-29
FIX: RecursionError when instanciating CtrlNode

### LINKED ISSUE #296: RecursionError with custom node
In a custom Node, I try to initialize with `CtrlNode.__init__(self, name, terminals=terminals)`, but I get a RecursionError:

``` python
  File "/usr/lib/python3.5/site-packages/pyqtgraph/flowchart/Flowchart.py", line 871, in nodeMenuTriggered
    self.chart.createNode(nodeType, pos=pos)
  File "/usr/lib/python3.5/site-packages/pyqtgraph/flowchart/Flowchart.py", line 174, in createNode
    node = self.library.getNodeType(nodeType)(name)
  File "/data/libs.git/JML/python/TBA/nodes.py", line 37, in __init__
    CtrlNode.__init__(self, name, terminals=terminals)
  File "/usr/lib/python3.5/site-packages/pyqtgraph/flowchart/library/common.py", line 89, in __init__
    if hasattr(self, 'uiTemplate'):
  File "/usr/lib/python3.5/site-packages/pyqtgraph/flowchart/Node.py", line 193, in __getattr__
    if attr not in self.terminals:
  File "/usr/lib/python3.5/site-packages/pyqtgraph/flowchart/Node.py", line 193, in __getattr__
    if attr not in self.terminals:
(...)
  File "/usr/lib/python3.5/site-packages/pyqtgraph/flowchart/Node.py", line 193, in __getattr__
    if attr not in self.terminals:
RecursionError: maximum recursion depth exceeded while calling a Python object
```

The problem is that `__getattr__` checks for `self.terminals` which is not yet defined, so  `__getattr__` is called again and so on.

I think putting the `if ui is None:` block after `Node.__init__` in  `CtrlNode.__init__` would do the trick.

2xB:
CtrlNode checks this:
https://github.com/pyqtgraph/pyqtgraph/blob/23b4e174f073bb1d444ff8390641dc2c7c52e0a0/pyqtgraph/flowchart/library/common.py#L95

Problem is: If 'uiTemplate' does not exist (it could exist via subclassing CtrlNode), it calls `Node.__getattr__`, with the important line
https://github.com/pyqtgraph/pyqtgraph/blob/23b4e174f073bb1d444ff8390641dc2c7c52e0a0/pyqtgraph/flowchart/Node.py#L192

, before `Node.__init__` is called. Therefore `self.terminals` does not yet exist (it is initialized in `Node.__init__`), which causes a new call to `Node.__getattr__` - you spot the recursion.

This pull request fixes this by calling `Node.__init__` first, as the lines before do not have any influence on that call.

Fixes #296.

### DIFF
--- a/pyqtgraph/flowchart/library/common.py
+++ b/pyqtgraph/flowchart/library/common.py
@@ -91,6 +91,8 @@ class CtrlNode(Node):
     sigStateChanged = QtCore.Signal(object)
     
     def __init__(self, name, ui=None, terminals=None):
+        Node.__init__(self, name=name, terminals=terminals)
+        
         if ui is None:
             if hasattr(self, 'uiTemplate'):
                 ui = self.uiTemplate
@@ -98,7 +100,6 @@ class CtrlNode(Node):
                 ui = []
         if terminals is None:
             terminals = {'In': {'io': 'in'}, 'Out': {'io': 'out', 'bypass': 'In'}}
-        Node.__init__(self, name=name, terminals=terminals)
         
         self.ui, self.stateGroup, self.ctrls = generateUi(ui)
         self.stateGroup.sigChanged.connect(self.changed)


2xB:
As can be seen by these failing tests, this seemingly introduces a bug.

j9ac9k:
You can leave this PR open and try and debug the issue, having it in one PR is a bit easier to see what was attempted already.

PR #975, opened 2019-06-30
Remove pytest-faulthandler from test dependencies

ixjlyons:
As of a few days ago, the pytest-faulthandler plugin was merged into pytest: https://github.com/pytest-dev/pytest/pull/5441

Along with this, the `--faulthandler-timeout` command line option was moved to an ini option `faulthandler_timeout`, causing most CI environments to fail.

This PR introduces `pytest>=5` as a requirement so we can just assume the new ini option is available and remove `pytest-faulthandler`.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -4,7 +4,6 @@ xvfb_height = 1080
 # use this due to some issues with ndarray reshape errors on CI systems
 xvfb_colordepth = 24
 xvfb_args=-ac +extension GLX +render
-addopts = --faulthandler-timeout=15
 
 filterwarnings =
     # comfortable skipping these warnings runtime warnings
--- a/pytest.ini
+++ b/pytest.ini
@@ -4,6 +4,7 @@ xvfb_height = 1080
 # use this due to some issues with ndarray reshape errors on CI systems
 xvfb_colordepth = 24
 xvfb_args=-ac +extension GLX +render
+faulthandler_timeout = 15
 
 filterwarnings =
     # comfortable skipping these warnings runtime warnings
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,6 +64,11 @@ install:
     - pip install pytest-xdist # multi-thread pytest
     - pip install pytest-cov # add coverage stats
 
+    # faulthandler support not built in to pytest for python 2.7
+    - if [ "${PYTHON}" == "2.7" ]; then
+        pip install pytest-faulthandler;
+      fi;
+
     # Debugging helpers
     - uname -a
     - cat /etc/issue
@@ -139,7 +144,7 @@ script:
 
     # Run unit tests
     - start_test "unit tests";
-      PYTHONPATH=. pytest --cov pyqtgraph -sv;
+      PYTHONPATH=. pytest --cov pyqtgraph -sv -o faulthandler_timeout=15;
       check_output "unit tests";
     - echo "test script finished. Current directory:"
     - pwd
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,6 @@ install:
       fi;
     - pip install pytest-xdist # multi-thread pytest
     - pip install pytest-cov # add coverage stats
-    - pip install pytest-faulthandler # activate faulthandler
 
     # Debugging helpers
     - uname -a
--- a/.travis.yml
+++ b/.travis.yml
@@ -144,7 +144,7 @@ script:
 
     # Run unit tests
     - start_test "unit tests";
-      PYTHONPATH=. pytest --cov pyqtgraph -sv -o faulthandler_timeout=15;
+      PYTHONPATH=. pytest --cov pyqtgraph -sv;
       check_output "unit tests";
     - echo "test script finished. Current directory:"
     - pwd
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,7 @@ install:
     # faulthandler support not built in to pytest for python 2.7
     - if [ "${PYTHON}" == "2.7" ]; then
         pip install pytest-faulthandler;
+        export PYTEST_ADDOPTS="--faulthandler-timeout=15";
       fi;
 
     # Debugging helpers
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,6 +47,10 @@ Please use the following guidelines when preparing changes:
 * pytest-xdist
 * Optional: pytest-xvfb
 
+If you have pytest < 5, you may also want to install the pytest-faulthandler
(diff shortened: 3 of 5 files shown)


ixjlyons:
Uh-oh, pytest version 5 is also the "no more Python 2" release :/ https://docs.pytest.org/en/latest/changelog.html#important

j9ac9k:
@ixjlyons I don't think we need to specify, pip should grab the latest compatable version by default, we can't assume that will be pytest 5.0 is all.

That said we should remove pytest-faulthandler from being required. We can do that by removing it's entry in `pytest.ini`

ixjlyons:
Alright that makes it pretty straightforward. I suppose we could add logic to the CI configs to install `pytest-faulthandler` with Python 2.7 if we still want it to be available. Latest update just removes it.

j9ac9k:
On mobile, please forgive the lack of formatting will be on a machine later today...

You should add the faulthandler timeout arg to the pytest call in the azure template file.

Also probably want to explicitly install it still on .travis config file

ixjlyons:
It is an ini-only option with pytest 5, which is why the tests are currently failing -- should've mentioned that in the initial comment.

j9ac9k:
Hmmm this is actually problematic... We should request a change to pytest-faulthandler plugin to accept the same .ini commands as pytest 5+ handles.

Last time I requested a change to the plugin it was done inside an hour.

ixjlyons:
It actually was already added: https://github.com/pytest-dev/pytest-faulthandler/issues/34

Also from that issue I found out it's possible to use `-o` to override ini options. I looked through the help output for something like that but missed it.

We'll see if CI is happy now, but now we should have faulthandler support across the board in CI without the option in the ini file so it's not a strict requirement for testing. Though realistically I think we could add it back in the ini file now that it's an option built into pytest.

2xB:
There is no need to use `-o` instead of writing `faulthandler_timeout = 15` in the ini - unused parameters in the ini are just ignored, so that does not cause a hidden dependency, as opposed to having the old command line argument `--faulthandler-timeout`, which was not ignored in that case.
As writing this parameter in the ini makes test results more consistent across different test setups, I think we should add it there.
But: The new ini parameter is NOT included in the 'old' `pytest-faulthandler` plugin, the referenced issue was closed when pytest itself got the parameter option, as the respective pull request made the plugin depreciated. It is therefore only available in pytest 5.0.

2xB:
I personally think we can just ignore that this parameter is not recognized for pytest < 5.0, as it really only enhances debug information, but does not really affect testing further (it e.g. does not abord any tests after the timeout). It's just a nice-to-have that we can use when pytest >= 5.0.

ixjlyons:
> There is no need to use -o instead of writing faulthandler_timeout = 15 in the ini - unused parameters in the ini are just ignored

That's good. Didn't know that.

> But: The new ini parameter is NOT included in the 'old' pytest-faulthandler plugin, the referenced issue was closed when pytest itself got the parameter option, as the respective pull request made the plugin depreciated. It is therefore only available in pytest 5.0.

Yeah this is what I was thinking with the most recent commit. The goal was to enable faulthandler with the timeout in CI across all environments without making pytest-faulthandler or pytest>5 required for testing locally.

> I personally think we can just ignore that this parameter is not recognized for pytest < 5.0, as it really only enhances debug information, but does not really affect testing further

I agree. @j9ac9k what do you think?

j9ac9k:
I'm good with ignoring the < 5.0 case.

pytest.ini
```diff
@@ -4,12 +4,11 @@ xvfb_height = 1080
 # use this due to some issues with ndarray reshape errors on CI systems
 xvfb_colordepth = 24
 xvfb_args=-ac +extension GLX +render
-addopts = --faulthandler-timeout=15
```
j9ac9k: Place whatever the new `pytest.ini` entry pytest 5.0+ is looking for and I think we're good.

ixjlyons:
> I just realized that the issue in pytest-faulthandler plugin referenced the merging of that plugin with pytest 5.0+, which still means we have the issue w/ pytest <5.0.

As it is now, pytest-faulthandler is explicitly installed in case pytest < 5, so the ini option is available in all cases and we can either specify it in the pytest command with `-o` or actually put it in the ini file. If we add the option to the ini file, you can test locally with pytest >= 5 and get the benefit of the timeout (overwhelming majority of cases I would assume) or run without faulthandler at all and the extra ini option will have no effect. With a note in CONTRIB

I think we've converged to that being the best compromise.

j9ac9k:
One last question, will pytest < 5.0 (aka python2 pipelines) with pytest-faulthandler installed manually respect that entry in `pytest.ini`?

ixjlyons:
Ah I've misunderstood something. I thought the ini option was also added to the pytest-faulthandler project but it wasn't. So, no, the timeout does not seem to be applied for pytest < 5 unless we add a special case to run pytest with the old command line arg for pytest < 5.

Not having the timeout for just the Python 2.7 pipeline seems ok to me given how much of a mess this is.

j9ac9k:
> Ah I've misunderstood something. I thought the ini option was also added to the pytest-faulthandler project but it wasn't. 

Yeah I got confused by this too!

> So, no, the timeout does not seem to be applied for pytest < 5 unless we add a special case to run pytest with the old command line arg for pytest < 5.

Doesn't the command line argument `--faulthandler-timeout=x` remain consistent all versions of pytest (assuming you have pytest-faulthandler installed on pytest < 5)?  If so, we should add that command line argument to our CI given that we have such tight control there and are very interested in its output.

If the command line argument syntax isn't consistent, I'm thinking this may be worth an extra if-statement in the CI.

ixjlyons:
Yeah the command line argument is no longer supported. I added the timeout using the `PYTEST_ADDOPTS` environment variable for Python 2.7. Pretty sure this is everything we want and reasonably clean.

j9ac9k:
> Yeah the command line argument is no longer supported. I added the timeout using the `PYTEST_ADDOPTS` environment variable for Python 2.7. Pretty sure this is everything we want and reasonably clean.

Ooof didn't realize the command line arguments weren't backwards compatible.... yeah that makes it definitely tougher.  Setting the env variable for py27 is likely the way to go for sure.

j9ac9k:
Oh, also need to remove the dependency from `tox.ini`

ixjlyons:
Looks like it was removed from my end

j9ac9k:
:facepalm: sorry you had, I'm just scatter brained

ixjlyons:
Oh hello, codecov. Funny to look back through this because I have no recollection of it whatsoever. Time flies! 😅

PR #976, opened 2019-07-01
Add issue template

2xB:
Using an issue template improves the amount of information given when an issue is opened and provides guidance. I tried to create an issue template that fits PyQtGraph, that I hereby propose.

I would be glad about feedback regarding the tone and style of this template. It can be tested here: https://github.com/2xB/pyqtgraph/issues/new/choose

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,42 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+<!-- In the following, please describe your issue in detail! -->
+<!-- If some of the sections do not apply, just remove them. -->
+
+### Short description
+<!-- This should summarize the issue. -->
+
+### Code to reproduce
+```python
+import pyqtgraph as pg
+import numpy as np
+# This is a placeholder for a minimal working example that reproduces the issue.
+```
+
+### Expected behavior
+<!-- What should happen? -->
+
+### Real behavior
+<!-- What happens? -->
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,42 @@
...
+
+---
+
+<!-- In the following, please describe your issue in detail! -->
+<!-- If some of the sections do not apply, just remove them. -->
+
+### Short description
+<!-- This should summarize the issue. -->
+
+### Code to reproduce
```
ixjlyons: Maybe instead of the code comment below add a template comment like

> Please provide a minimal working example that reproduces the issue in the code block below.
> Ideally, this should be a full example someone else could copy+paste and run without additional setup.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Expected behavior
+<!-- What should happen? -->
+
+### Real behavior
+<!-- What happens? -->
+
+```
+An error occurred?
+Post it inside these 'code fences'!
```
ixjlyons: I think it's worth specifying that the *full traceback* (not just the final exception message) should be provided.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+### Tested environment(s)
```
ixjlyons: One more bullet point I'd add is installation method (pip, conda, system packages, etc.)

PR #977, opened 2019-07-02
Add a faster method for computing pseudoscatter

campagnola:
The current `pseudoScatter` function (used for beeswarm plots) has very slow performance on large arrays. This PR introduces an alternative method that produces an approximate result much more quickly. By default, the original (exact) method is still used.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -2311,13 +2311,61 @@ def invertQTransform(tr):
             raise Exception("Transform is not invertible.")
         return inv[0]
     
+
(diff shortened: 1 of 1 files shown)


j9ac9k:
@campagnola Do you think this is something that's worth adding a test for?  I know the new method doesn't ensure exact results, but we could use `np.allclose` or something similar if necessary.

j9ac9k:
Closing/Re-Opening to rerun CI pipeline

j9ac9k:
CI passed, ...after playing w/ this, yeah, no way `np.allclose()` will work to compare the two.  Merging.

PR #978, opened 2019-07-02
Reduce test window size for OSX compatibility

campagnola:
May fix some broken OSX tests. The main problem is that OSX reports its window size in _logical_ pixels rather than _physical_ pixels, which means that on some systems we can have unexpectedly small maximum window sizes. This causes the test window to shrink vertically, changing the aspect ratio and breaking the test. (solution is just to use a smaller window with the same aspect ratio)


j9ac9k:
Look at all that green....

PR #981, opened 2019-07-02
Merge master into develop

2xB:
This PR merges `master` into `develop`, so the branches are no longer diverging. That effectively only causes one spelling error to be corrected.

`pyqtgraph/graphicsItems/PlotItem/PlotItem.py` and `pyqtgraph/graphicsWindows.py` had merge conflicts in documentation strings that I resolved in favor of the version from `develop`.


j9ac9k:
LGTM

Issue #967, opened 2019-06-26
[CI-fail] Intermittent Segfault on GLIsosurface and GLScatterPlotItem


j9ac9k:
https://dev.azure.com/pyqtgraph/pyqtgraph/_build/results?buildId=246&view=logs

Running tests on these two files generated a segmentation fault on Py37/PyQt5/macOS configuration

* GLIsosurface.py
* GLScatterPlotItem.py

This does not happen often, but it's not the first time I've seen it.  I've been able to get failures to be more consistent when I upgraded the image to macOS 10.14, but for other reasons I rolled it back.

Perhaps we should consider rolling the macOS image forward to 10.14 and try and fix the failures there?

j9ac9k:
Been trying to replicate locally but I haven't been able to.

j9ac9k:
segfaults occur on this line:

https://github.com/pyqtgraph/pyqtgraph/blob/73c440a4db6cef93fa47e8aefd7d0f502e281e7e/pyqtgraph/opengl/GLViewWidget.py#L192

j9ac9k:
Fixed with #1131

PR #982, opened 2019-07-03
[WIP] Fix segfaults in GLViewWidget on Mac

2xB:
This is a work in progress. `glClear` is said to fail if there is no valid OpenGL context, so the first step is to check if that is the case.

Aims at fixing #967

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,4 +35,4 @@ jobs:
   - template: azure-test-template.yml
     parameters:
       name: MacOS
-      vmImage: 'macOS-10.14'
+      vmImage: 'macOS-10.13'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,4 +35,4 @@ jobs:
   - template: azure-test-template.yml
     parameters:
       name: MacOS
-      vmImage: 'macOS-10.13'
+      vmImage: 'macOS-10.14'
--- a/examples/GLSurfacePlot.py
+++ b/examples/GLSurfacePlot.py
@@ -82,11 +82,14 @@ p4.shader()['colorMap'] = np.array([0.2, 2, 0.5, 0.2, 1, 1, 0.2, 0, 2])
 p4.translate(10, 10, 0)
 w.addItem(p4)
 
+w.update() # Only necessary if running headless
+
 index = 0
 def update():
     global p4, z, index
     index -= 1
     p4.setData(z=z[index%z.shape[0]])
+    w.update() # Only necessary if running headless
     
 timer = QtCore.QTimer()
 timer.timeout.connect(update)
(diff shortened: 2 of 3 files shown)


j9ac9k:
Saw that this passed on macOS, but due to the inconsistent nature of this issue, I queued up this PR to run again (will likely queue it up a 3rd time too even if it passes this second time).

j9ac9k:
@2xB can you do me a favor on this PR, and change the `10.13` on this line

https://github.com/pyqtgraph/pyqtgraph/blob/b0bc5b8931b8430b82e7fab4114811176272a994/azure-pipelines.yml#L38

to 10.14?

2xB:
If these tests also pass, the issue seems to be that `glClearColor(*bgcolor)` raises an error that I catch with `glGetError()`. Therefore, what needs to be done to understand this, is to investigate what `glGetError()` returns and possibly whether `bgcolor` is invalid.
I'll have to do a break here now, will look at this later.

2xB:
And one test failed with the same segfault again. Well then...

j9ac9k:
if macOS 10.14 is still being problematic, we can definitely revert back, but should probably open an issue about it to investigate more later...

2xB:
Well, that this is failing shows that the error is indeed happening in `glClear`, and not seeing it previously was just luck.

2xB:
As the tests do not output the BG color, although they should as soon as `paintGL` is run, might it be that always when `paintGL` is run in the test suite, it fails?
To check whether `paintGL` is really only run in this macOS configuration, I will now add a debug commit that causes a RuntimeError everytime `paintGL` is run.

2xB:
This is interesting: All Windows and MacOS configurations do NOT run `paintGL` in the tests (specifically the `GLSurfacePlot.py` example), while all Linux configurations do. This is especially interesting because it means currently at least the GLSurfacePlot example seemingly never calculates graphics besides on Linux.

2xB:
I created issue #983 for this and now ignore this.

2xB:
The CI is not failing, meaning `paintGL` is not executed. The only way to re-run succeeding tests that I am aware of is by closing and re-opening the pull request, so I try it once. If it does not work, I will not be able to ignore #983 but have to wait until it is fixed...

2xB:
Okay, Mac tests passed - debugging this therefore does not make sense before issue #983 is resolved.

2xB:
Since calling paintGL explicitly worked, the next question is whether the MacOS tests reproducibly fail now even if no artifical RuntimeError is raised.

2xB:
Okay, so the issue does NOT just occur if `paintGL` is called. It seemingly only occurs, when `paintGL` is called by Qt. And it obviously only occurs at random for Python 3.7. The last commit aims at testing what happens, when Qt calls `paintGL`. A difference might be the `viewport` variable.

2xB:
Retriggering checks by reopening...

2xB:
MacOS Python 3.6 PyQt-5.12 and PyQt-5.9 are seemingly also affected.

2xB:
If `WaitForWindowShown` does the expected, MacOS not failing here demonstrates that this is NOT (only?) a problem of paintGL not being called, but also has some additional randomness. Therefore, the next commit tests whether paintGL is now reliably called.

2xB:
Tests show that `WaitForWindowShown` does not cause paintGL to be reliably called on MacOS.

j9ac9k:
Fixed in #1131

PR #987, opened 2019-07-05
FIX: Curves are automatically set visible when one is deleted

### LINKED ISSUE #985: PlotWidget (and other plotting widgets) unhides all Items after removing single item
### Short description
I am plotting some spectral curves which overlaps. For my application it is practical to pre-add all spectras (curves) into plot and `hide()`, "unhide"/`show()` them with some checkboxes in `QListView`. Now I want to add and remove some preview vertical lines, for fast visual identification of peaks. When I finalize the preview and want the vertical lines to be removed from Plot with `removeItem`, it somehow triggers all spectral curves to unhide (probably with `show()` method). I made very simple minimal test to isolate the problem from my software code. See the minimal example below.

### Code to reproduce
```python
import pyqtgraph as pg

pw = pg.PlotWidget()
pw.show()
curves = {i: pg.PlotCurveItem([j ** i for j in range(10)]) for i in range(7)}

# 7 curves should be generated.
for i in curves:
    curves[i].hide()
    pw.addItem(curves[i])
curves[0].show()
curves[1].show()
# there should be only two lines on the plot visible at this point.
pw.removeItem(curves[5])
# but after this all items (except removed curve) are unhided()
```

### Expected behavior
`removeItem` should only remove the item without triggering sibling items into show() state.

### Real behavior
all sibling items are triggered to become visible.

### Tested environment(s)

 * PyQtGraph version: 0.11.0.dev0+g73c440a
 * Qt Python binding: PyQt5 5.11.3 Qt 5.11.2
 * Python version: 3.6.8 (default, Jan 14 2019, 11:02:34)
                             [GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]
 * isolated in python-virtual-environment without inclusion of system site packages. 
 * Operating system: KDE neon
 * Installation method: `pip git+https://pyqtgraph/pyqtgraph.git`

2xB:
On call of `PlotItem.removeItem`, `PlotItem.updateDecimation` is called, which sets invisible curves visible. This behavior seems unintended.

Fixes #985.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -925,9 +925,7 @@ class PlotItem(GraphicsWidget):
         curves = self.curves[:]
         split = len(curves) - numCurves
         for i in range(len(curves)):
-            if numCurves == -1 or i >= split:
-                curves[i].show()
-            else:
+            if numCurves != -1 and i >= split:
                 if self.ctrl.forgetTracesCheck.isChecked():
                     curves[i].clear()
                     self.removeItem(curves[i])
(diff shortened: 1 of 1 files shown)


2xB:
That is a good hint, thank you! I only slightly changed the iteration from your example.

j9ac9k:
even better change!

PR #990, opened 2019-07-10
Fix typo

ziqiangxu:


ixjlyons:
Woops this shouldn't have gone to the master branch. I don't think it's necessary to revert but I'll re-open #989 and merge that as well

PR #991, opened 2019-07-10
Expand CI + pre-commit

j9ac9k:
Current azure-pipelines only does unit-testing on the 18 configurations we have set.  With this pull request, azure-pipelines does more checks in line with the `extra` flag on used in Travis CI.

To be more specific, azure-pipelines now has multiple stages, `pre-test` and `test`.  During the `pre-test` stage, the following things happen in parallel

* `python.setup.py style` is run
  * checks flake8 required on entire codebase
  * checks flake8 optional on files that have diffs
  * checks for correct style line-endings
* diff size is checked
* pyqtgraph bdist_wheel is built (and then uploaded as an artifact)

During the `test` stage, all the unit tests run as before, with the exception that they download the bdist_wheel artifact and install that (so instead of creating 18 wheels and installing each one for on its own pipeline, we create one wheel, and use that wheel for all 18 pipelines)

The latest build can be seen here:

https://dev.azure.com/j9ac9k/pyqtgraph-fork/_build/results?buildId=284
In addition, this PR includes a few other features

* `.flake8` file is added to better support flake8 integration with editors  and `pre-commit`
* `.pre-commit-config.yml` is added allowing for support for `pre-commit` checks to run at commit time.

A note about `pre-commit` is added to `CONTRIBUTING.md`; it is strictly optional, however it will be helpful for ensuring code passes quality checks at commit time.

Also some changes were made to some files with diffs to make them conform to the new flake8 checks that are run against the CI.


azure-pipelines.yml
```diff
@@ -21,18 +17,77 @@ pr:
...
+        git gc --aggressive &&
+        MERGE_SIZE=`du -s . | sed -e "s/\t.*//"` &&
+        if [ "${MERGE_SIZE}" != "${TARGET_SIZE}" ]; then
+          SIZE_DIFF=`expr \( ${MERGE_SIZE} - ${TARGET_SIZE} \)`;
+        else
+          SIZE_DIFF=0;
+        fi;
+        echo -e "Estimated content size difference = ${SIZE_DIFF} kB" &&
+        test ${SIZE_DIFF} -lt 100;
+        # python setup.py mergesize
```
campagnola: Why have the merge-size check implemented both here and in setupHelpers?

j9ac9k: right, it was implemented in setupHelpers; but before this gets merged, it should really be tested via both methods... not sure if I should have the CI run the bash or setupHelper method

j9ac9k:
Added `[WIP]` tag in an attempt to get the `flake8 --diff` checks going, not sure . that's going to be feasible though, will likely have to remove altogether.

j9ac9k:
Started working on this on another branch, the flake8 -diff checks are still tough to integrate at the commit time.  I am not a fan of having this check strictly in the CI (this takes too long to hear back that you have an extra space, or your line is too long).

I've created https://github.com/pyqtgraph/pre-commit-hooks and have been trying too make a hook to run the flake8 diff, but so far I haven't been able to get it working as I intended.

If anyone has any suggestions on how to integrate that check at commit time, let me know.

j9ac9k:
ugh....what on earth is happening w/ creating conda environments in the CI process?

j9ac9k:
something has changed on azure pipelines with respect to how macOS and conda task work; this will take be a little while to debug.

j9ac9k:
Gave up on trying to get linting checks to happen at the pre-commit time (doing a flake8 check on a file basis is fine, but doing it on the `--diff` I'm not sure can be done).

I had to modify the CI script substantially for a few reasons

* `conda 4.7.X` has problems with python2 environments
  * fixed by manually downloading miniconda2 4.6.14 and installing it over the existing miniconda setup
* `-c conda-forge pyside` on linux has some missing symbol issue that's an indication of a build related problem, but given its age, it won't be fixed (luckily -c defaults pyside` doesn't have this issue

azure-pipelines.yml
```diff
@@ -12,7 +8,7 @@ trigger:
 
 # Build only for PRs for master branch
 pr:
-  autoCancel: true
+  autoCancel: "true"
```
2xB: Intuitively (and from https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml ), this change should not be necessary? This also applies for further booleans in this document.

j9ac9k: Shouldn't be necessary but my vscode azure pipeline validator plugin was having a cow about it, so this was more about appeasing my editor 😆

j9ac9k: Actually just changed that, editor isn't complaining, either it got an update or I remembered wrong... will update the PR will that change.

j9ac9k:
Throwing WIP label as i realized i never actually tested the 100kb size limit on the size check...will want to do that before this is merged, I'll test on another branch

j9ac9k:
Thanks @campagnola I'll go ahead and merge as you gave approval

Thanks for everyone's patience with me on this,...this took me way too long to sort out, and there will likely be changes coming in the future with pytest-nunit, updates to pyqtgraph-core and doc builds.

Issue #283, opened 2016-02-12
GLGridItem ignores color argument


flutefreak7:
If I try to create a GLGridItem with a different color grid (like for gray grid on white background), it doesn't do anything with the color argument.  In the paint method, It appears the color is set with `glColor4f(1, 1, 1, .3)`.

flutefreak7:
Thank you ixjlyons for adding this functionality!

Since this change isn't in the version I have installed I implemented ixjlyons fix using a subclass:

    class Grid(gl.GLGridItem):

        def __init__(self, color=None, **kwargs):
            super().__init__(self, **kwargs)

            if color is None:
                color = (255, 255, 255, 80)
            self.setColor(color)

        def setColor(self, args):
            """
            Set the color of the grid. Arguments are the same as those accepted by
            :func:`glColor <pyqtgraph.glColor>`.
            """
            self.color = fn.glColor(args)

        def paint(self):
            self.setupGLState()

            if self.antialias:
                glEnable(GL_LINE_SMOOTH)
                glEnable(GL_BLEND)
                glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA)
                glHint(GL_LINE_SMOOTH_HINT, GL_NICEST)

            glBegin(GL_LINES)

            x, y, z = self.size()
            xs, ys, zs = self.spacing()
            xvals = np.arange(-x / 2., x / 2. + xs * 0.001, xs)
            yvals = np.arange(-y / 2., y / 2. + ys * 0.001, ys)
            glColor4f(*self.color)
            for x in xvals:
                glVertex3f(x, yvals[0], 0)
                glVertex3f(x, yvals[-1], 0)
            for y in yvals:
                glVertex3f(xvals[0], y, 0)
                glVertex3f(xvals[-1], y, 0)

            glEnd()

ixjlyons:
Fixed by #648

MeechaelA:
I think it would be beneficial to allow this attribute to be set by the user, of course with a default.

ixjlyons:
[...]

PR #992, opened 2019-07-11
Adding setter for GLGridItem.color

2xB:
`GLGridItem` has a color attribute that has no setter. This Pull Requests adds such a setter function. It originates from a cherry-picked commit from PR #419.

### DIFF
--- a/pyqtgraph/opengl/items/GLGridItem.py
+++ b/pyqtgraph/opengl/items/GLGridItem.py
@@ -14,7 +14,7 @@ class GLGridItem(GLGraphicsItem):
     Displays a wire-frame grid. 
     """
     
-    def __init__(self, size=None, color=(1, 1, 1, .3), antialias=True, glOptions='translucent'):
+    def __init__(self, size=None, color=(255, 255, 255, 76.5), antialias=True, glOptions='translucent'):
         GLGraphicsItem.__init__(self)
         self.setGLOptions(glOptions)
         self.antialias = antialias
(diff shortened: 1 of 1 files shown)


2xB:
It can e.g. be tested running the `GLMeshItem` example interactively and executing `g.setColor((1,1,0,1))`.

2xB:
PR #397 aims at providing the same function. Differences are: 
* That PR function accepts everything `QColor()` also accepts, this PR function accepts an `(r,g,b,a)` tuple. PR #397 does not accept the tuples currently only accepted by `GLGridItem`, so this would change the API.
* That PR is missing a call to `self.update`.
If desired, I can add a call to QColor if no tuple is given.

ixjlyons:
I wrote a few notes recently about this here: https://github.com/pyqtgraph/pyqtgraph/issues/283#issuecomment-507730765

I'm not sure how consistent the rest of the code is, but I think it makes sense to "hide" the color attribute as `self._color` when there's an explicit getter/setter.

2xB:
Thank you, the last two commits made `GLGridItem` consistent with `GLBoxItem`. That includes making the `color` attribute private, adding a getter and using `pyqtgraph.function.Color` to parse the color arguments with a corresponding docstring similar to that one in GLGridItem.

j9ac9k:
@2xB  thanks for the PR.  I'm good with merging this; when merging this does this mean we can close #397 and #419 ?

2xB:
Yes, this PR contains the part of #419 that still applies, and has the same goal as #397, although a slightly different approach (a comparison is done in a comment further above).

2xB:
Btw. I created this PR as an updated version of #419, which is why I already closed #419 when creating this PR. In retrospect, I am not sure if that was a wise decision.

j9ac9k:
Closing/Opening to trigger new CI run.

PR #994, opened 2019-07-16
Disable the downsample ratio spinbox when you have autodownsampling enabled

fake-name:
Basically, right now, the plotitem context menu has the option to configure downsampling. However, when you turn on automatic downsampling, you can still manually set the downsampling ratio (which is then ignored).

This can be very confusing, so this sets the downsampling spinbox to disabled when you have autodownsampling enabled.

I'd like to have the actual downsampling ratio percolate back up and change the value in the read-only spinbox, but doing that involves a lot of changes, and if you have two datasets with different x-axis spacings, there is no single value you can put in the spinbox anyways, so this just does the simpler thing.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -898,6 +898,12 @@ class PlotItem(GraphicsWidget):
             
         auto = self.ctrl.downsampleCheck.isChecked() and self.ctrl.autoDownsampleCheck.isChecked()
             
+        if auto:
+            self.ctrl.downsampleSpin.setEnabled(True)
+        else:
+            self.ctrl.downsampleSpin.setEnabled(False)
+
+
         if self.ctrl.subsampleRadio.isChecked():
             method = 'subsample' 
         elif self.ctrl.meanRadio.isChecked():


fake-name:
Yeah, it's because I initially had it also attach/detatch `downsampleSpin.valueChanged`, so it could be programmatically tweaked without firing events.

I'd still like to figure out how to make the value of `downsampleSpin` show the downsampling ratio that the `auto` arrives at, but I haven't figured out a way to do so yet.

j9ac9k:
Looking at this PR, I think the boolean logic is backwards, I think you want

```python
        if auto:
            self.ctrl.downsampleSpin.setEnabled(False)
        else:
            self.ctrl.downsampleSpin.setEnabled(True)
```

But I think this is easier to read

```python
self.ctl.downsampleSpin.setEnabled(not auto)
```

If you update the PR accordingly, I'll gladly merge it.

fake-name:
I totally don't remember the context where I write this, and I'm no longer at the job where I was using pyqtgraph any more. 

 ¯\\\_(ツ)\_/¯

j9ac9k:
Haha, don't sweat it, I know it's been a while.  Sorry I didn't take more time with this when you submitted it.

PR #996, opened 2019-07-17
Update LegendItem.py

kevinanewman:
Propose adding a clear() method (or equivalent) for easier legend re-use with dynamically updated plots...

### DIFF
--- a/pyqtgraph/graphicsItems/LegendItem.py
+++ b/pyqtgraph/graphicsItems/LegendItem.py
@@ -101,6 +101,15 @@ class LegendItem(GraphicsWidget, GraphicsWidgetAnchor):
                 label.close()
                 self.updateSize()                       # redraq box
 
+    def clear(self):
+        """
+        Removes all items from the legend.
+
+        Useful for reusing and dynamically updating charts and their legends.
+        """
+        while self.items != []:
+            self.removeItem(self.items[0][1].text)
+                
     def updateSize(self):
         if self.size is not None:
             return


2xB:
@kevinanewman With the `pyqtgraph 0.11.0` version (as currently on branch `develop`), items removed from a plot will also be removed from the legend so if a plot is cleared, its legend is also cleared. Does this solve the problem you address with this Pull Request?

sharon92:
clear() method works well on 0.11.0

But it would also be nice to update legends by setting name in setData.

for eg. 

`PlotDataItem.setData(x=X,y=Y, name = 'New Name')`

2xB:
@sharon92 This should work, yes. Might you please open an issue regarding this?

sharon92:
@2xB Done!

kevinanewman:
In regards to 

> @kevinanewman With the `pyqtgraph 0.11.0` version (as currently on branch `develop`), items removed from a plot will also be removed from the legend so if a plot is cleared, its legend is also cleared. Does this solve the problem you address with this Pull Request?

The way my code works right now is that I create the chart and the curves and then dynamically update them... I'm trying to reuse all the chart elements and just set new data.  So, I'm not removing the curves, I just give them new data.  I needed a way to clear the legend so that I could create new labels for the new curves.

Maybe there's a better way to do this, I'm open to suggestions, otherwise it's working well for my live simulation outputs.

Thank you,
Kevin

j9ac9k:
@kevinanewman thanks for the PR, sorry it's taken us a while to review it.  Change looks good to me, i'll be merging shortly.

Issue #461, opened 2017-03-12
QtProcess proxy cannot handle numpy arrays with dtype uint8


peendebak:
A minimal example is this:

```
import numpy as np
import pyqtgraph.multiprocess as mp

class test:
    def set_attribute(self, name, value):
        setattr(self, name, value)

if __name__=='__main__':
    proc = mp.QtProcess()
    testpg = proc._import('testpg')
    t = testpg.test()
    t.set_attribute('dummy', 10) # fine
    im = np.zeros((3, 5, 6))
    t.set_attribute('dummy', im ) # fine
    im = np.zeros((3, 5, 6), dtype=np.uint8)

    # the next line generated a: ValueError: cannot reshape array of size 3 into shape (3,5,6)
    t.set_attribute('dummy', im ) 
```

2xB:
I can reproduce this using pyqtgraph 0.10.0 from `pip`, Python 3.6.7, PyQt 5.12.2 under KDE Neon 5.15.

Note that I had to add `pg.mkQApp()` at the beginning and that the test is required to be saved with the filename "testpg.py" .

The error is:
```
error: <pyqtgraph.multiprocess.processes.QtProcess object at 0x7f3c378e9128>_child 8 cannot reshape array of size 3 into shape (3,5,6)
/[...]/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py:506: RemoteExceptionWarning: ===== Remote process raised exception on request: =====
  warnings.warn("===== Remote process raised exception on request: =====", RemoteExceptionWarning)
/[...]/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py:507: RemoteExceptionWarning: Traceback (most recent call last):
  File "/[...]/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 235, in handleRequest
    fnargs[i] = np.fromstring(byteData[ind], dtype=dtype).reshape(shape)
ValueError: cannot reshape array of size 3 into shape (3,5,6)

  warnings.warn(''.join(excStr), RemoteExceptionWarning)
/[...]/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py:508: RemoteExceptionWarning: ===== Local Traceback to request follows: =====
  warnings.warn("===== Local Traceback to request follows: =====", RemoteExceptionWarning)
Traceback (most recent call last):
  File "test_.py", line 21, in <module>
    t.set_attribute('dummy', im ) 
[...]

PR #998, opened 2019-07-20
Fix QtProcess proxy transmitting small multidimensional arrays

2xB:
This fixes issue #461 as a workaround for https://bugs.python.org/issue37637 by transforming all objects into bytes objects before transmitting them using `send_bytes`.

Note: I did not notice any performance decrease using the example `RemoteSpeedTest.py` with this workaround.


j9ac9k:
Thanks for the PR @2xB LGTM

PR #999, opened 2019-07-20
Make outline around fillLevel optional

2xB:
PR #922 introduced outlines surrounding fillLevel areas. This is a design element that is not always wanted: In some scientific use cases, multiple lines in the same color are commonly interpreted as a bifurcation or hysteresis. This PR aims at providing a boolean option `fillOutline` to be used together with `fillLevel` to activate outlines.

An example of a confusing outline can be seen in the bottom left plot of example `Plotting.py`:

This Pull Request demonstrates the new option in the example `histogram.py`:

Further, API documentation is added.

### DIFF
--- a/examples/histogram.py
+++ b/examples/histogram.py
@@ -22,7 +22,7 @@ y,x = np.histogram(vals, bins=np.linspace(-3, 8, 40))
 
 ## Using stepMode=True causes the plot to draw two lines for each sample.
 ## notice that len(x) == len(y)+1
-plt1.plot(x, y, stepMode=True, fillLevel=0, brush=(0,0,255,150))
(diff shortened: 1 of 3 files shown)


Dschoni:
Re-run to see if the segfault on windows could be reproduced.

j9ac9k:
@2xB thanks for the PR.  This looks good to me, I'm curious if we should update one of the examples to reflect this change, thoughts?

2xB:
This Pull Request demonstrates the new option in the example `histogram.py`, there is a screenshot of this above!

j9ac9k:
Totally shouldn't have looked this just before bedtime, sorry about that.  Thanks for the PR @2xB

PR #1012, opened 2019-08-07
fix: circular texture was slightly off-center

Aikhjarto:


j9ac9k:
@Aikhjarto   Thanks for the PR, can you provide a code sample where this is an issue so i can verify the solution does indeed fix that problem? 

Thanks!

Aikhjarto:
import numpy as np
def fn_old(x,y):
    r = ((x-w/2.)**2 + (y-w/2.)**2) ** 0.5
    return 255 * (w/2. - np.clip(r, w/2.-1.0, w/2.))

def fn_new(x,y):
    r = ((x-(w-1)/2.)**2 + (y-(w-1)/2.)**2) ** 0.5
    return 255 * (w/2. - np.clip(r, w/2.-1.0, w/2.))

w=8
plt.figure()
plt.title('old circle function')
plt.imshow(np.fromfunction(fn_old, (w,w)))
plt.figure()
plt.title('new circle function')
plt.imshow(np.fromfunction(fn_new, (w,w)))

j9ac9k:
Thanks for the PR @Aikhjarto and thanks for the matplotlib code to highlight the issue.  I'll merge momentarily.

PR #1014, opened 2019-08-07
Fix: colormap: Support various arguments as color

### LINKED ISSUE #1009: ColorMap getLookupTable fails when alpha=None
### Short description
The `getLookupTable` method of `pyqtgraph.ColorMap` fails to automatically determine whether alpha values should be included in a lookup table.

### Code to reproduce

```python
import pyqtgraph as pg

colors = [
    (0, 0, 0),
    (45, 5, 61),
    (84, 42, 55)
]
cmap = pg.ColorMap(pos=(0, 1, 2), color=colors)

cmap.getLookupTable(alpha=False)  # Works
cmap.getLookupTable(alpha=True)  # Works, but returned lookup table does not contain alpha values.
cmap.getLookupTable()  # Does not work
```

### Expected behavior
A lookup table is returned.

### Real behavior
The following exception is thrown:

```
  File "test.py", line 13, in <module>
    cmap.getLookupTable()  # Does not work
  File "...\venv\lib\site-packages\pyqtgraph\colormap.py", line 223, in getLookupTable
    alpha = self.usesAlpha()
  File "...\venv\lib\site-packages\pyqtgraph\colormap.py", line 236, in usesAlpha
    return np.any(self.color[:,3] != max)
IndexError: index 3 is out of bounds for axis 1 with size 3
```

### Tested environment(s)

 * PyQtGraph version: 0.11.0.dev0+gad7453c
 * Qt Python binding: 0.11.0.dev0+gad7453c
 * Python version: 3.7.3
 * NumPy version: 0.11.0.dev0+gad7453c
 * Installation method: pip

2xB:
Currently, `colormap`s may be provided with RGB values, but internally treat `self.color` as an array of RGBA values. Examples are provided in issue #1009.

This PR fixes that by allowing to pass any type of color to `colormap` that `mkColor` accepts.

Fixes #1009

### DIFF
--- a/pyqtgraph/colormap.py
+++ b/pyqtgraph/colormap.py
@@ -69,10 +69,11 @@ class ColorMap(object):
         self.pos = np.array(pos)
         order = np.argsort(self.pos)
         self.pos = self.pos[order]
-        self.color = []
-        for color in np.array(color)[order]:
-            self.color.append(mkColor(color).getRgb())
-        self.color = np.array(self.color)
+        self.color = np.apply_along_axis(
+            func1d = lambda x: mkColor(x).getRgb(),
+            axis   = -1,
+            arr    = color,
+            )[order]
         if mode is None:
             mode = np.ones(len(pos))
         self.mode = mode
--- a/pyqtgraph/colormap.py
+++ b/pyqtgraph/colormap.py
@@ -1,6 +1,7 @@
 import numpy as np
 from .Qt import QtGui, QtCore
 from .python2_3 import basestring
+from .functions import mkColor
 
 
 class ColorMap(object):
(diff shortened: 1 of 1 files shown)


pyqtgraph/colormap.py
```diff
@@ -68,7 +69,10 @@ def __init__(self, pos, color, mode=None):
         self.pos = np.array(pos)
         order = np.argsort(self.pos)
         self.pos = self.pos[order]
-        self.color = np.array(color)[order]
+        self.color = []
+        for color in np.array(color)[order]:
+            self.color.append(mkColor(color).getRgb())
+        self.color = np.array(self.color)
```
j9ac9k: not to get all code-golfey, but this should likely be done as a list comprehension

```python
self.color = np.array([mkColor(color).getRgb() for color in np.array(color)[order]])
```

Reason I'm preferential to this method is that under the proposed implementation, not only is it a bit slower (because list-comprehensions are amazing like that), but at one point in the code `self.color` is a list, but at another point in the code, `self.color` is a np.array

j9ac9k:
thanks for the PR @2xB just had one comment/suggestion.

pyqtgraph/colormap.py
j9ac9k: Thinking more about this, you can get even better performance (this may not exactly quite right).

```python
self.color = np.apply_along_axis(lambda x: mkColor(x).getRgb(), 0, np.array(color)[order])
```

2xB:
Thank you, I incorporated this with the last commit!

PR #1015, opened 2019-08-08
Always convert PlotDataItem data to NumPy array

2xB:
Currently, the following code does not work:

```python3
import pyqtgraph as pg

pg.plot(range(3), range(3))
```

This is because only lists are converted to NumPy arrays, not e.g. iterators like `range(...)`. This PR fixes this by converting everything that is not a NumPy array.


j9ac9k:
Thanks for the PR @2xB, LGTM

PR #1016, opened 2019-08-08
Fix: ImageView sigTimeChanged was only emitted on mouse interaction

### LINKED ISSUE #862: Synchronize multiple pg.ImageView objects
Hello,

I encountered some difficulties two synchronize two (or more) pg.ImageView objects:

```python
import numpy as np
import pyqtgraph as pg


pg.mkQApp()
imv1 = pg.ImageView()
imv1.show()
imv1.setImage(np.random.random((100,8,8)))

imv2 = pg.ImageView()
imv2.show()
imv2.setImage(np.random.random((100,8,8)))

imv2.sigTimeChanged.connect(imv1.setCurrentIndex)
imv1.sigTimeChanged.connect(imv2.setCurrentIndex)
```

Both image views are synchronized when I set the position of the time axis via the mouse. However if I use keys, e.g. <space> for <play>, only the selected pg.ImageView is reacting. I found a small workaround, but I wonder, if this is the intended usage?

```python
import numpy as np
import pyqtgraph as pg

class PatchedImageView(pg.ImageView):
    def jumpFrames(self, n):
        super().jumpFrames(n)
        _, time = self.timeIndex(self.timeLine)
        self.sigTimeChanged.emit(self.currentIndex, time)


pg.mkQApp()
imv1 = PatchedImageView()
imv1.show()
imv1.setImage(np.random.random((100,8,8)))

imv2 = PatchedImageView()
imv2.show()
imv2.setImage(np.random.random((100,8,8)))

imv2.sigTimeChanged.connect(imv1.setCurrentIndex)
imv1.sigTimeChanged.connect(imv2.setCurrentIndex)
```

2xB:
Signal `ImageView.sigTimeChanged` was only emitted in method `ImageView.timeLineChanged`. It was skipped via
```
        if self.ignoreTimeLine:
            return
```
when `ImageView.setCurrentIndex` changed the timeline. This PR moves the update of `ImageView.currentIndex` from `ImageView.setCurrentIndex` to `ImageView.timeLineChanged`, so the latter function is always executed and therefore the call to `ImageView.sigTimeChanged` always happens when the index is changed, therefore fixing the original issue.

Fixes #862

### DIFF
--- a/pyqtgraph/imageview/ImageView.py
+++ b/pyqtgraph/imageview/ImageView.py
@@ -131,7 +131,7 @@ class ImageView(QtGui.QWidget):
         self.scene = self.ui.graphicsView.scene()
         self.ui.histogram.setLevelMode(levelMode)
         
-        self.ignoreTimeLine = False
+        self.ignorePlaying = False
         
         if view is None:
             self.view = ViewBox()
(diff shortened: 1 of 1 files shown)


j9ac9k:
LGTM

PR #1017, opened 2019-08-09
Fix: Item on ViewBox causes duplicate paint calls

### LINKED ISSUE #7: GridItem causes duplicate view udpates
When a GridItem is added to a ViewBox, any change to the view range results in the view being repainted twice.

2xB:
Analysis leads to the reason for issue #7:

Via `scene.sigPrepareForPaint.connect(self.prepareForPaint)` in `ViewBox.checkSceneChange()`, the method `ViewBox.prepareForPaint()` is called whenever a paint event happens. This calls `ViewBox.updateMatrix`, which - if the transformation was not yet applied - calls `ViewBox.childGroup.setTransform()`, that issues a repaint. This brings a total of two `paint` calls.

By moving the task of detecting scene changes to the method `update`, that is there to be notified on scene changes before `paint`, this PR fixes the issue and cleans up code.

Fixes #7

### DIFF
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -281,23 +281,10 @@ class ViewBox(GraphicsWidget):
             #if scene is not None and hasattr(scene, 'sigPrepareForPaint'):
                 #scene.sigPrepareForPaint.connect(self.prepareForPaint)
         #return ret
-
-    def checkSceneChange(self):
-        # ViewBox needs to receive sigPrepareForPaint from its scene before
-        # being painted. However, we have no way of being informed when the
-        # scene has changed in order to make this connection. The usual way
-        # to do this is via itemChange(), but bugs prevent this approach
-        # (see above). Instead, we simply check at every paint to see whether
-        # (the scene has changed.
-        scene = self.scene()
-        if scene == self._lastScene:
-            return
-        if self._lastScene is not None and hasattr(self._lastScene, 'sigPrepareForPaint'):
-            self._lastScene.sigPrepareForPaint.disconnect(self.prepareForPaint)
-        if scene is not None and hasattr(scene, 'sigPrepareForPaint'):
-            scene.sigPrepareForPaint.connect(self.prepareForPaint)
+        
+    def update(self, *args, **kwargs):
         self.prepareForPaint()
-        self._lastScene = scene
+        GraphicsWidget.update(self, *args, **kwargs)
 
     def prepareForPaint(self):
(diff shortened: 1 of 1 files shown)


2xB:
The CI shows that this brings test_InfiniteLine and test_mouseInteraction to fail. I do not understand why, as InfiniteLines seem to work as intended. If someone has an insight here or would like to spend time getting one, that would be very welcome!

2xB:
It turns out this raises the amount of calls to updateMatrix. I did not figure out why this breaks the mentioned tests though.

2xB:
Btw. logging calls of `ViewBox.prepareForPaint()` for the `Plotting.py` example shows that updating plotting data formerly caused a call of that function even when not needed, this PR also fixes that.

2xB:
The bug can be reproduced by this simplification of one failing test:
```python3
import pyqtgraph as pg

pg.mkQApp()

plt = pg.plot()
plt.scene().minDragTime = 0  # let us simulate mouse drags very quickly.
vline = plt.addLine(x=0, movable=True)
hline = plt.addLine(y=0, movable=True)
hline2 = plt.addLine(y=-1, movable=False)
plt.setXRange(-10, 10)
plt.setYRange(-10, 10)
#plt.setYRange(-10, 10)
```

It can be seen that `setYRange` has no effect. If you call it twice (uncomment the last row), it does. Therefore this PR introduces a new issue, or it shows a bug that already exists for `setYRange`. As switching `setXRange` and `setYRange` still sets the x range properly and the y range not, this seems to be an issue with that function.

2xB:
It actually was an issue with `ViewBox.setRange`, which implicitly called `updateAutoRange` before the autorange for the specific axes were disabled. The call went through `updateViewRange` -> `update` -> `prepareForPaint` (this is why it was not a problem before - prepareForPaint was called later). Disabling autorange as soon as possible in `ViewBox.setRange` therefore fixes this error.

2xB:
The other commit assured that `updateMatrix` is called when a `resizeEvent` happens. Otherwise items inside a ViewBox would not be rescaled on resize.

j9ac9k:
ViewBox stuff is pretty fragile, the changes look good to me @2xB but I'm going to ask @campagnola to look this over.

campagnola:
TBH I don't remember the reason for using `scene.sigPrepareForPaint` vs `ViewBox.update`, but I can't find anything broken by this PR, and it improves performance in some common cases.

Super happy any time we can reduce the code complexity in ViewBox!

j9ac9k:
@2xB Thanks for the PR, now that @campagnola has given it his blessing, I'll go ahead and merge!

Issue #1021, opened 2019-08-17
PR 942 introduces RuntimeError: wrapped C/C++ object of type GraphicsLayoutWidget has been deleted


titusjan:
Since commit 08c0de7, which merges PR #942, I get the following error when I close my application window.

`RuntimeError: wrapped C/C++ object of type GraphicsLayoutWidget has been deleted`

The following program reproduces the issue.
```

import sys
import pyqtgraph as pg
from pyqtgraph.Qt import QtWidgets
from PyQt5.Qt import PYQT_VERSION_STR


class DemoWindow(QtWidgets.QMainWindow):

    def __init__(self, parent=None):
        super().__init__(parent=parent)

        self.mainWidget = QtWidgets.QWidget()
        self.setCentralWidget(self.mainWidget)

        self.mainLayout = QtWidgets.QVBoxLayout()
        self.mainWidget.setLayout(self.mainLayout)

        self.plotItem = pg.PlotItem()
        self.plotItem.plot([3.5, 5, -4, 0, 0, 3])

        self.graphicsLayoutWidget = pg.GraphicsLayoutWidget()
        self.graphicsLayoutWidget.addItem(self.plotItem, 0, 0)

        self.mainLayout.addWidget(self.graphicsLayoutWidget)


def main():
    print(sys.version)
    print("PyQt version: {}".format(PYQT_VERSION_STR))
    print("PyQtGraph version: {}".format(pg.__version__))
    app = QtWidgets.QApplication([])
    win = DemoWindow()
    win.setGeometry(400, 100, 700, 600)
    win.show()
    app.exec()


if __name__ == '__main__':
    main()
```

I tested with PyQt 5.9 on OS-X 10.14.6.

```
3.6.7 | packaged by conda-forge | (default, Feb 28 2019, 02:16:08)  [GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
PyQt version: 5.9.2
PyQtGraph version: 0.11.0.dev0
```

and also with PyQt 5.6

```
3.6.8 |Anaconda, Inc.| (default, Dec 29 2018, 19:04:46) [GCC 4.2.1 Compatible Clang 4.0.1 tags/RELEASE_401/final)]
PyQt version: 5.6
PyQtGraph version: 0.11.0.dev0
```

[...]

PR #1042, opened 2019-09-22
Close windows at the end of test functions

### LINKED ISSUE #1040: Warnings output for visible windows after test functions run
When running the tests (I think as of merging #942), I get the following:

```
============================================================================================= warnings summary =============================================================================================
pyqtgraph/exporters/tests/test_svg.py::test_plotscene
pyqtgraph/graphicsItems/tests/test_AxisItem.py::test_AxisItem_stopAxisAtTick
pyqtgraph/graphicsItems/tests/test_ErrorBarItem.py::test_ErrorBarItem_defer_data
pyqtgraph/graphicsItems/tests/test_PlotCurveItem.py::test_PlotCurveItem
pyqtgraph/graphicsItems/tests/test_PlotDataItem.py::test_clipping
  /home/kenny/src/pyqtgraph/.venv-pyqt5/lib/python3.7/site-packages/_pytest/python.py:165: RuntimeWarning: Visible window deleted. To prevent this, store a reference to the window object.
    testfunction(**testargs)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
============================================================================= 50 passed, 5 skipped, 5 warnings in 9.77 seconds =============================================================================
Exception ignored in: <function GraphicsView._del at 0x7f8a7df8bc20>
  File "/home/kenny/src/pyqtgraph/pyqtgraph/widgets/GraphicsView.py", line 414, in _del
TypeError: 'NoneType' object is not callable
```

The `TypeError` is happening because `warnings.warn` is not available in some circumstances [[ref]](https://bugs.python.org/issue29174).

I think we just need to make sure any windows are closed at the end of the test functions so the warning isn't thrown during the tests. There aren't that many. The `TypeError` is kind of confusing so maybe we could catch that let it go in `GraphicsView._del` as well.

ixjlyons:
Fixes #1040

### DIFF
--- a/pyqtgraph/exporters/tests/test_svg.py
+++ b/pyqtgraph/exporters/tests/test_svg.py
@@ -28,6 +28,7 @@ def test_plotscene():
     ex.export(fileName=tempfilename)
     # clean up after the test is done
     os.unlink(tempfilename)
+    w.close()
 
 def test_simple():
     tempfilename = tempfile.NamedTemporaryFile(suffix='.svg').name
--- a/pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/tests/test_ViewBox.py
@@ -71,6 +71,8 @@ def test_ViewBox():
     size1 = QRectF(0, h, w, -h)
     assertMapping(vb, view1, size1)
 
+    win.close()
+
 
 skipreason = "Skipping this test until someone has time to fix it."
 @pytest.mark.skipif(True, reason=skipreason)
--- a/pyqtgraph/graphicsItems/tests/test_AxisItem.py
+++ b/pyqtgraph/graphicsItems/tests/test_AxisItem.py
@@ -28,3 +28,5 @@ def test_AxisItem_stopAxisAtTick(monkeypatch):
     monkeypatch.setattr(left, "drawPicture", test_left)
 
     plot.show()
+    app.processEvents()
+    plot.close()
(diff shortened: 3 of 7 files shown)


j9ac9k:
LGTM

I'm on mobile so I can't see CI results easily, if they passed, I say go ahead and merge.

ixjlyons:
Yep the warnings are gone. @2xB do you want to take a quick look at this? The `GraphicsView._del` function is getting pretty messy but I think we knew there'd be some corner cases to deal with.

2xB:
I think at this point we should really decide whether we want to keep the warning in `GraphicsView._del`. It is still an open question as written by @j9ac9k in https://github.com/pyqtgraph/pyqtgraph/issues/1021#issuecomment-522288478. Since this seems to have way more implications than expected and as far as I know there is nobody except from me really advocating this, it might be reasonable to just remove it.

Besides that, this PR looks good!

ixjlyons:
Yep, I thought about that too. I'm generally in favor of the nicer warning and the problems we've run into so far aren't all that bad. The code *looks* a little ugly but it's not all that complicated at this point (especially compared to other stuff going on in pyqtgraph). I think we can keep moving forward and fixing things like this if/when they come up. With our testing to get 0.11 out and with users relying on the develop branch, I'm pretty sure we will or already have hit the edge cases.

j9ac9k:
I'm not strongly opinionated one way or the other, but I agree with @ixjlyons that the code is a little ugly and for the time being we can patch things up as they arise.  Anyway, I'll merge this.

Thanks for the PR @ixjlyons

PR #1048, opened 2019-09-28
Fix: InfiniteLine bounding box wrong

### LINKED ISSUE #1037: Line Labels not visible for small graph values
Hey Everyone,

I found a bug where InfLinLables attached to InfiniteLines are not properly shown, if the total y-range of the graph is too small. Simple example below.

If the label position is big enough (e.g. 0.5) the label sometimes comes into view. But this is not always the case.

The red line label also moves out of sight if you zoom in too much.

### Short description
InfLineLabels are not visible if the Y-Values of a graph are too small

### Code to reproduce
```python
import sys
from PyQt5 import QtWidgets
from PyQt5.QtCore import Qt
import pyqtgraph as pg
from pyqtgraph import mkPen


    def __init__(self):

        self.resize(800, 600)

        self.setCentralWidget(glw)

        x_vals = [0, 1, 2, 3, 4]
        y_vals1 = [0, 1, 2, 3, 4]
        y_vals2 = [0.000, 0.001, 0.002, 0.003, 0.004]

        plot1 = pg.PlotItem()
        plot2 = pg.PlotItem()

        red_line = plot1.plot(x=x_vals, y=y_vals1, pen="r")
        green_line = plot2.plot(x=x_vals, y=y_vals2, pen="g")

        label_kwargs = {"position": 0.05,
                         "color": "w"}

        plot1.addLine(x=2.5,
                      pen=mkPen('r', style=Qt.DashLine),
                      label="{:.2f}".format(2.5),
                      labelOpts=label_kwargs)

        plot2.addLine(x=2.5,
                      pen=mkPen('g', style=Qt.DashLine),
                      label="{:.2f}".format(2.5),
                      labelOpts=label_kwargs)

        glw.addItem(plot1, 0, 0)
        glw.addItem(plot2, 1, 0)


if __name__ == '__main__':
    ex = Example()
    ex.show()
    sys.exit(app.exec_())

```

### Expected behavior
Label of green Infinite line should be at the bottom of the graph, as it is with the red line

### Real behavior
Label is not in the view range and only becomes visible once you zoom out so y=-1 becomes visible

### Tested environment(s)

 * PyQtGraph version: Current development branch
 * Qt Python binding: PyQt 5.12.3
 * Python version: 3.7.4
 * NumPy version: 1.16.3
 * Operating system: Win10
 * Installation method: pip

2xB:
The bounding box of `InfiniteLine`s is wrong: Currently the pen width is translated into a width `w` relative to the plot range in one direction. But `w` is then used in both directions, x and y. Since `w` is only necessary in one direction, this PR removes its influence in the bounding box in the other direction.

Fixes #1037

### DIFF
--- a/pyqtgraph/graphicsItems/InfiniteLine.py
+++ b/pyqtgraph/graphicsItems/InfiniteLine.py
@@ -314,8 +314,8 @@ class InfiniteLine(GraphicsObject):
         length = br.width()
         left = br.left() + length * self.span[0]
         right = br.left() + length * self.span[1]
-        br.setLeft(left - w)
-        br.setRight(right + w)
+        br.setLeft(left)
+        br.setRight(right)
         br = br.normalized()
         
         vs = self.getViewBox().size()


j9ac9k:
Failures in CI are unrelated, PR LGTM, merging.

Thanks for the PR @2xB

Issue #1011, opened 2019-08-06
Plots with large numbers invisible


2xB:
### Short description
If a plot contains an array with large numbers (e.g. 1e27), it is not drawn at all.

### Code to reproduce

```python
import numpy as np
import sys
import pyqtgraph as pg

app = pg.mkQApp()


x = [1,2,3,4]
y = [1,2,3,1e27]

pg.plot(x, y)


if (sys.flags.interactive != 1) or not hasattr(QtCore, 'PYQT_VERSION'):
    app.exec_()
```

### Expected behavior
A plot should be drawn.

### Real behavior
No plot is drawn.

### Tested environment(s)

 * PyQtGraph version: develop commit ad7453c7818eddb3b18354c65370f416cb668b29
 * Qt Python binding: PyQt5 5.13.0 Qt 5.13.0
 * Python version: 3.7.3
 * NumPy version: 1.16.4
 * Operating system: Manjaro
 * Installation method: pyqtgraph: Run in git repository; Dependencies: pip

jongra:
same problem observed for `np.nan` values on windows system.

2xB:
@jongra Could you please post an example and screenshot? I tried the following on Windows and it looks fine.

```python3
import numpy as np
import sys
import pyqtgraph as pg

app = pg.mkQApp()

x = [1,2,3,4]
y = [1,2,3,np.nan]

pg.plot(x, y)

if (sys.flags.interactive != 1) or not hasattr(QtCore, 'PYQT_VERSION'):
    app.exec_()
```

jongra:
Hi @2xB ,
thank you for your reply.
I just copied your example, but on my system (Win10 Education 64bit, i3-2120, Dell U2412M onboard graphics, python version 3.7.4, pygraph v. 0.10.0, PyQt5 v. 5.13.1, numpy v. 1.17.2) the plot is empty (cf. screenshot) and python doesn't raise any errors.
If I replace the `np.nan` with a `4`, it works fine and the graph shows a nice line plot.

![screen_1](https://user-images.githubusercontent.com/12088604/66753620-81b1e280-ee94-11e9-871c-979f0713526d.PNG)

2xB:
@jongra Interesting! My Windows system does not show this. Environment:

 * PyQtGraph version: both current develop and 0.10.0
 * Qt Python binding: PyQt5 5.12.2 Qt 5.12.3
 * Python version: 3.6.5
[...]

PR #1058, opened 2019-10-17
Avoid Qt receiving nan values

### LINKED ISSUE #1057: plots not showing with NaN
### Short description

Plotting data with np.nan results in an empty plot.

### Code to reproduce

```python
import numpy as np
import pyqtgraph as pg

data2 = np.random.normal(size=1000)
data2[0] = np.nan

pg.plot(data, title="no NaN")
pg.plot(data2, title="one NaN")

pg.QtGui.QApplication.exec_()
```

### Expected behavior

Two plots with random data.

### Real behavior

First plot shows as normal.
Second plot is empty.
```
An error occurred?
```

### Tested environment(s)

 * PyQtGraph version: 0.11.0.dev0+ged6586c
 * Qt Python binding: PySide2 5.13.1 Qt 5.13.1
 * Python version: 3.7.3
 * NumPy version: 1.17.2
 * Operating system: Windows 10, x64
 * Installation method: python from python.org installer. numpy from pip. pyqtgraph from git and / "python setup.py install" in pyqtgraph directory

 * PyQtGraph version: 0.11.0.dev0+ged6586c
 * Qt Python binding: PySide2 5.13.1 Qt 5.13.1
 * Python version: 3.7.3
 * NumPy version: 1.17.2
 * Operating system: Centos 7
 * Installation method: python built from source. numpy from pip. pyqtgraph from git and / "python setup.py install" in pyqtgraph directory

### Additional context

2xB:
Since Qt 5.13.1, paths containing non-finite values are not drawn. This PR avoids passing non-finite values to Qt. Also, as long as `connect` is neither `'all'` nor an array, it also draws paths unconnected at places where a non-finite value occurs.

Fixes #1057.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1492,18 +1492,29 @@ def arrayToQPath(x, y, connect='all'):
     arr[1:-1]['x'] = x
     arr[1:-1]['y'] = y
 
+    non_finite_values = np.logical_not(np.isfinite(x) & np.isfinite(y))
+
+    arr[1:-1]['x'][non_finite_values] = np.roll(x,-1)[non_finite_values]
+    arr[1:-1]['y'][non_finite_values] = np.roll(y,-1)[non_finite_values]
+    
     # decide which points are connected by lines
     if eq(connect, 'all'):
         arr[1:-1]['c'] = 1
-    elif eq(connect, 'pairs'):
-        arr[1:-1]['c'][::2] = 1
-        arr[1:-1]['c'][1::2] = 0
-    elif eq(connect, 'finite'):
-        arr[1:-1]['c'] = np.isfinite(x) & np.isfinite(y)
+        if non_finite_values[-1] and len(non_finite_values) > 1:
+            arr[1:-1]['c'][-2] = 0
     elif isinstance(connect, np.ndarray):
         arr[1:-1]['c'] = connect
     else:
-        raise Exception('connect argument must be "all", "pairs", "finite", or array')
+        if eq(connect, 'pairs'):
+            arr[1:-1]['c'][::2] = 1
+            arr[1:-1]['c'][1::2] = 0
+        elif eq(connect, 'finite'):
+            arr[1:-1]['c'] = 1
+        else:
+            raise Exception('connect argument must be "all", "pairs", "finite", or array')
+        
+        arr[1:-1]['c'][non_finite_values] = 0
+        arr[1:-1]['c'][np.roll(non_finite_values,-1)] = 0
 
     #profiler('fill array')
     # write last 0
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1494,8 +1494,8 @@ def arrayToQPath(x, y, connect='all'):
 
     non_finite_values = np.logical_not(np.isfinite(x) & np.isfinite(y))
 
-    arr[1:-1]['x'][non_finite_values] = np.roll(x,-1)[non_finite_values]
-    arr[1:-1]['y'][non_finite_values] = np.roll(y,-1)[non_finite_values]
+    arr[1:-1]['x'][non_finite_values] = np.interp(np.flatnonzero(non_finite_values), np.flatnonzero(~non_finite_values), arr[1:-1]['x'][~non_finite_values])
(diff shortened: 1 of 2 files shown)


2xB:
Btw. as @j9ac9k noted in #1011, this might be a third-party bug (not in PyQt5, as it was observed in both PySide2 and PyQt5), so it might also be a good idea to instead of using this workaround, fixing the original bug. Now I sadly can't afford spending that time currently, so I will leave it as it is now.

j9ac9k:
I'm reluctant to say this is a 3rd party "bug"; as the `np.NaN` behavior was always sort of a happy accident from what I can remember (as that was the most efficient way to plot a bunch of lines at once).

Looking at the Qt Changelog from 5.13.0 to 5.13.1, we have the following:
https://code.qt.io/cgit/qt/qtcharts.git/tree/dist/changes-5.13.1/?h=v5.13.1

```
****************************************************************************
*                                Bug Fixes                                 *
****************************************************************************

- QCandlestickSeries: remove sets on destroy to prevent leaks
- [QTBUG-70987] Fix crash on ChartView component destroy
- [QTBUG-76271] Revert "Fix rendering lines with Antialiasing render hint"
- [QTBUG-60384] Force entire chart update when points label clipping changes

```

There is also an update in QGraphicsView

```
 - QGraphicsView:
   * Ignore disabled items when setting the mouse cursor.
```

Not sure if any of those changes are what's causing this behavior.

2xB:
Somehow tests are failing for Python 2.7/Qt 4.8...

j9ac9k:
@2xB if failing for an unrelated reason, would you mind opening a new issue?

Maintaining the python2-qt4 pipeline has been a real chore...

j9ac9k:
I checked the CI builds, Python 2.7 builds are fine otherwise so something in this diff is breaking the test.  It may be worth pinging @campagnola on slack to take a look at this one.

2xB:
Turns out: The reason this test fails only for PyQt4 is probably that there is a larger tolerance in `assertImageMatch` for images generated by PyQt4 than for PyQt5.

2xB:
The higher tolerance for PyQt5 is not intended to be permanent, therefore I created issue  #1064 to remember that something has to be done there.

It seems btw. that one pixel is off with this change in Python2/PyQt4. If someone is interested in tracking down the reason, feel free to do so!

spauka:
I'm also interested in getting this fixed in the latest version, as it's causing some issues with my plots as well. I will highlight that, while this solution works, it is causing a performance regression for plots with 10000-1000000 points.

j9ac9k:
Closing/Opening to re-trigger CI pipelines

j9ac9k:
Here is the CI run, looks like all python2 pipelines failed:

https://dev.azure.com/pyqtgraph/pyqtgraph/_build/results?buildId=633&view=results

/shakes-fist-at-python2 

We could put this PR on ice until we ship out 0.11 and drop py2 support.... thoughts @2xB ?

j9ac9k:
Closing/Opening again, CI pipeline should be more robust now...

pyqtgraph/functions.py
```diff
@@ -1492,18 +1492,29 @@ def arrayToQPath(x, y, connect='all'):
...
+        if eq(connect, 'pairs'):
+            arr[1:-1]['c'][::2] = 1
+            arr[1:-1]['c'][1::2] = 0
+        elif eq(connect, 'finite'):
+            arr[1:-1]['c'] = 1
+        else:
+            raise Exception('connect argument must be "all", "pairs", "finite", or array')
+        
+        arr[1:-1]['c'][non_finite_values] = 0
+        arr[1:-1]['c'][np.roll(non_finite_values,-1)] = 0
```
campagnola, line 1517: I think there's a bug here--roll(-1) moves the first element to the end of the array ?

2xB: The `roll` was used because the last entry of `arr[1:-1]['c']` doesn't matter. Code to test:
```python3
import numpy as np
import pyqtgraph as pg

app = pg.mkQApp()

data = np.random.normal(size=10)
data2 = np.copy(data)
data2[0] = np.nan

pg.plot(data, title="no NaN")
pg.plot(data2, title="one NaN")

app.exec_()
```

j9ac9k:
Merged #1287 which takes care of this issue, did some timing information and documented it in #1057 showing that implementation being faster.  If this PR addresses a use case that #1287 does not, let me know and we'll re-address.  In the meantime, going to close this out.

PR #1063, opened 2019-10-28
Fix: AttributeError in ViewBox.setEnableMenu

ales-erjavec:
### Issue

Executing
```python
from PyQt5.QtWidgets import QApplication
import pyqtgraph as pg

app = QApplication([])

view = pg.PlotWidget()
viewbox = view.plotItem.vb
viewbox.setMenuEnabled(False)
viewbox.setMenuEnabled(False)  # <- error here
```
raises a 
```
  File "<stdin>", line 9, in <module>
  File "/Users/aleserjavec/workspace/pyqtgraph/pyqtgraph/graphicsItems/ViewBox/ViewBox.py", line 388, in setMenuEnabled
    self.menu.setParent(None)
AttributeError: 'NoneType' object has no attribute 'setParent'
```
in the second call to setMenuEnabled(False)

Fixed by validating the current menu state and only deleting it when necessary.

### DIFF
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -324,7 +324,8 @@ class ViewBox(GraphicsWidget):
         if self.state['enableMenu'] and self.menu is None:
             self.menu = ViewBoxMenu(self)
             self.updateViewLists()
-        else:
+        elif not self.state['enableMenu'] and self.menu is not None:
+            self.menu.setParent(None)
             self.menu = None
 
         self.updateViewRange()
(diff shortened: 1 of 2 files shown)


2xB:
Okay, that looks reasonable to me. Could we avoid that code duplication there, e.g. by calling `setMenuEnabled` in `setState`?

ales-erjavec:
> Could we avoid that code duplication there, e.g. by calling setMenuEnabled in setState?

That would change the order of `sigStateChanged` and `sigRangeChanged` (via `updateViewRange`) signal emission.

2xB:
How about a method `applyMenuEnabled` with that code? An alternative would be to just call `setMenuEnabled` after `updateViewRange`, but that would be a bit less clean as `self.state['enableMenu']` would then be overwritten with itself in the call to `setMenuEnabled`.

markotoplak:
I've been testing orange3 (https://github.com/biolab/orange3) with pyqtgraph 0.11rc0. Merging this PR into the development branch fixed some regressions with the current release candidate.

@2xB, could you consider merging this PR?

2xB:
Sounds good to me. Tagging @ixjlyons, @j9ac9k and @ksunden, as they merged to `develop` recently.

j9ac9k:
Eek, so sorry I missed this @2xB @markotoplak  .... I'll close/re-open to trigger a CI run, if it passes, I'll merge.

j9ac9k:
CI had a hick-up on one pipeline,  passed on re-run, merging, sorry for the wait folks.

PR #1069, opened 2019-11-06
Allow actions to display title instead of name

rwalroth:
ActionParameterItem changed so that if there is
a title it will be displayed, otherwise displays name.

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes.py
+++ b/pyqtgraph/parametertree/parameterTypes.py
@@ -612,7 +612,10 @@ class ActionParameterItem(ParameterItem):
         self.layout = QtGui.QHBoxLayout()
         self.layout.setContentsMargins(0, 0, 0, 0)
         self.layoutWidget.setLayout(self.layout)
-        self.button = QtGui.QPushButton(param.name())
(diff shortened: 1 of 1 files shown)


pyqtgraph/parametertree/parameterTypes.py
```diff
@@ -612,7 +612,10 @@ def __init__(self, param, depth):
         self.layout = QtGui.QHBoxLayout()
         self.layout.setContentsMargins(0, 0, 0, 0)
         self.layoutWidget.setLayout(self.layout)
-        self.button = QtGui.QPushButton(param.name())
+        title = param.opts.get('title', None)
```
j9ac9k: would `title = param.opts.get('title', param.name())` not work

j9ac9k:
Thanks for the PR @rwalroth I had just one comment on the change.

pyqtgraph/parametertree/parameterTypes.py
rwalroth: I was following a convention used for setting titles in other parameters, I have not tested that but it should work.

PR #1071, opened 2019-11-07
Update Data.py

patricev:
Python eval not working with python 3 - bug fix with the exec() part


j9ac9k:
@patricev thanks for the PR, sorry it's taken so long for someone to get back to you.

I didn't realize we had `exec` anywhere in the codebase ... do you know if there is a reason that `exec()` is being used instead of `ast.literal_eval()`?

patricev:
Hi, I don't know about the reason...
Regards,

j9ac9k:
hi @patricev 

can you import the `ast` module and replace the `exec()` calls with `ast.literal_eval()`?  Provided CI doesn't have a gripe about that, I'll merge the PR then.

Thanks again!

j9ac9k:
Sorry this has taken me so long to get to.  Merging. Thanks for the PR @patricev

PR #1072, opened 2019-11-09
HDF5Exporter handling of ragged curves with tests

### LINKED ISSUE #747: HDF5Exporter: throws error when curves aren't the same length

ixjlyons:
Builds from #748, which I thought could use some testing.

Fixes #747

### DIFF
--- a/azure-test-template.yml
+++ b/azure-test-template.yml
@@ -116,9 +116,9 @@ jobs:
           conda install -c conda-forge $(qt.bindings) --yes
         fi
         echo "Installing remainder of dependencies"
-        conda install -c conda-forge numpy scipy six pyopengl --yes
+        conda install -c conda-forge numpy scipy six pyopengl h5py --yes
       else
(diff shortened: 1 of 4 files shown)


j9ac9k:
LGTM thanks @jdchristesen  and @ixjlyons  for the PR!

PR #1076, opened 2019-11-13
Reset currentRow and currentCol on GraphicsLayout.clear()

paulmueller:
If a GraphicsLayout has multiple columns and GraphicsLayout.clear() is called, then when adding new plot items, there will be an offset in the first row which propagates to all remaining rows.

This is a simple fix.

actual behavior:
image

desired behavior:
image


j9ac9k:
@paulmueller Thanks for the PR!  Sorry it took me so long to get to it, merging shortly.

PR #1079, opened 2019-11-19
improve performance of updateData PlotCurveItem (saves about 2us per call)

danielhrisca:


j9ac9k:
@danielhrisca Pull requests like this one totally make my day, thanks!

Issue #176, opened 2015-06-18
SVG Exporter does not place symbols on correct place


CPrescher:
Hei there,

i have found a strange bug, when creating a plot having a line and markers for a specific dataset the SVG Exporter does not place the marker symbols at the expected positions but somehow below.

an example script:

```
import pyqtgraph as pg
from pyqtgraph.exporters.SVGExporter import SVGExporter
from PyQt4 import QtGui

if __name__ == '__main__':
    app = QtGui.QApplication([])
    win = pg.GraphicsWindow("title SVG Exporter bug")
    win.resize(1000, 600)
    p1 = win.addPlot()
    p1.plot(x=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10], y=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10], pen=(200, 200, 200),
            symbolBrush=(255, 0, 0), symbolPen='w')
    QtGui.QApplication.processEvents()
    QtGui.QApplication.processEvents()
    exporter = SVGExporter(p1)
    exporter.export("test.svg")``
```

This produces an svg file were the markers are clearly at the wrong positions...

![plotitem_1](https://cloud.githubusercontent.com/assets/4086557/8243952/9269077e-15e1-11e5-859b-3c80fdd4bd67.png)

markotoplak:
The same thing happens with an ImageExporter (to .png).

j9ac9k:
Confirmed that this is still the case with the current developer build.  Thanks for the example code/error.

paulmueller:
I can also confirm this bug.

When explicitly plotting the scatter plot with "pxMode=False", the offset goes away:
```
import pyqtgraph as pg
import numpy as np
from pyqtgraph.exporters.SVGExporter import SVGExporter
from PyQt5 import QtGui

if __name__ == '__main__':
    app = QtGui.QApplication([])
    win = pg.GraphicsWindow("title SVG Exporter bug")
    win.resize(1000, 600)
    p1 = win.addPlot()
    p1.plot(x=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10], y=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
    scatter = pg.ScatterPlotItem(pen=(200, 200, 200),
                                 symbolBrush=(255, 0, 0),
                                 symbolPen='w',
                                 size=.5,
                                 pxMode=False,
                                 )
    p1.addItem(scatter)
[...]

PR #1085, opened 2019-11-25
fix wrong offset when drawing symbol

paulmueller:
I believe this was not correctly implemented in #88.
Note that this fix has nothing to do with #176.

### DIFF
--- a/pyqtgraph/graphicsItems/ScatterPlotItem.py
+++ b/pyqtgraph/graphicsItems/ScatterPlotItem.py
@@ -781,7 +781,7 @@ class ScatterPlotItem(GraphicsObject):
                 pts = pts[:,viewMask]
                 for i, rec in enumerate(data):
                     p.resetTransform()
-                    p.translate(pts[0,i] + rec['width'], pts[1,i] + rec['width']/2)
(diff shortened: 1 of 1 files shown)


j9ac9k:
@paulmueller Thanks for the PR!  I'll merge momentarily.  Let me know if you would like an invite to our slack group to help us coordinate development efforts!

paulmueller:
@j9ac9k thanks for the offer. But I am only trying to fix a couple of issues I am having at the moment :)

j9ac9k:
> @j9ac9k thanks for the offer. But I am only trying to fix a couple of issues I am having at the moment :)

No problem; you've been killing it on the PR front and making detailed issues, all us maintainers appreciate it. 

Also, to be fair, most maintainers hopped onto the project because of a small subset of issues that impacted each of us, and we've largely stayed on to try and push out the next release to pypi. Fact is the codebase is so large that no one maintainer (with the exception of Luke probably) has a good grasp on all of it.

Feel free to reach out if you change your mind and want to help get the next stable release out the door.  This upcoming release will be the last to support Qt4 and Py2, after that, we will be adopting NEP-29 ( https://numpy.org/neps/nep-0029-deprecation_policy.html ), so the support matrix will be getting a lot simpler before long.

PR #980, opened 2019-07-02
remove deprecated call to time.clock

danielhrisca:

PR #1088, opened 2019-11-27
Update ptime.py for 3.8 compatibility

valenmgm:
systime.clock() is no longer supported for Python 3.8. perf_counter works well. Tested in Windows 10


ixjlyons:
This was fixed with #980 (note the conditional import at the top so `clock` refers to `perf_counter` for Python > 3). Thank you for the PR though!

PR #1090, opened 2019-12-03
dump ExportDialog.exporterParameters

### LINKED ISSUE #1087: Propose removal of `GraphicsScene.ExportDialog.ExporterParameters`
I propose to remove `ExportDialog.exporterParameters`. They are used to remember the export parameters of an exporter when the user switches between exporters (e.g. SVGExporter and ImageExporter).

https://github.com/pyqtgraph/pyqtgraph/blob/db483f8ec22a0d53bd0e71fac69b0045a96c6637/pyqtgraph/GraphicsScene/exportDialog.py#L127-L136

### Reasons

1. I doubt that anyone actually needs this.

2. I would like to create a PR that automatically sets the correct resolution in the ImageExporter and also keeps the correct aspect ratio when the user changes width or height in the ImageExporter (This feature was intended but is currently broken). The problem is that a simple fix would invalidate the rationale behind `ExportDialog.exporterParameters`, because changing the scene item would also update the ImageExporter parameters (and the stored parameters are not used). An elaborate fix would add a lot of clutter code (creating a dictionary of dictionaries for each item and each exporter).
The simple fix is highlighted here:
https://github.com/paulmueller/pyqtgraph/blob/b81f6d66d2b3cf4d334aa31b75b71e68cc7a3796/pyqtgraph/GraphicsScene/exportDialog.py#L101-L111

### Related Bug

As mentioned in my second point, the ImageExporter currently only displays the height and width of the scene and it does not adjust when e.g. a PlotItem is selected (which has different height, widht, aspect ratio). Furthermore, width and height are not updated to match the PlotItem aspect ratio if I change one of width or height.

### Question

Would you accept a PR that gets rid of  `ExportDialog.exporterParameters` and fixes  the related bug?

paulmueller:
fixes #1087

### DIFF
--- a/pyqtgraph/GraphicsScene/exportDialog.py
+++ b/pyqtgraph/GraphicsScene/exportDialog.py
@@ -22,8 +22,6 @@ class ExportDialog(QtGui.QWidget):
         self.shown = False
         self.currentExporter = None
         self.scene = scene
-            
-        self.exporterParameters = {}
 
         self.selectBox = QtGui.QGraphicsRectItem()
         self.selectBox.setPen(fn.mkPen('y', width=3, style=QtCore.Qt.DashLine))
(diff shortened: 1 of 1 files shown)


j9ac9k:
Hi @paulmueller sorry this has taken so long for me to get to.  Thanks for the detailed issue report.

PR #1092, opened 2019-12-06
avoid App Nap on macOS >= 10.9

caenrigen:
This solves several errors rising up in an application that uses realtime plotting based on pyqtgraph.
The problem is that macOS will activate App Nap for an application when it is minimised or not displayed on the screen. The moment when it happens it is up to the OS.
If this happens it is not possible to communicate with the application anymore and errors like timeout will occur. Also as a side effect code might get slow.
Screenshot 2019-12-06 at 16 22 38

This might not be the right way of doing it but is the solution I found. It avoids runtime errors which are cryptic and hard to debug.

Might also not be the correct file to put this lines of code.
Might also not be a good idea to do it at all as it defeats the purpose of App Nap.
I just wanted to share this with the world. Maybe it solves other issues.

Maybe this should be a concern for each application that make use of pyqtgraph. In my case I cannot fix it on the level of my application because it is being built on top another python module that uses pyqtgraph under the hood.

Related readings:
- [https://github.com/msorvig/qt-and-app-nap](url)
- [https://codereview.qt-project.org/c/qt/qtbase/+/202515](url)

P.S. Ignore lines below 35, just whitespaces that my editor removed

### DIFF
--- a/pyqtgraph/__init__.py
+++ b/pyqtgraph/__init__.py
@@ -29,6 +29,10 @@ if sys.version_info[0] < 2 or (sys.version_info[0] == 2 and sys.version_info[1] 
 ## helpers for 2/3 compatibility
 from . import python2_3
 
+## avoid App Nap in macOS >= 10.9
+import appnope
+appnope.nope()
+
 ## in general openGL is poorly supported with Qt+GraphicsView.
 ## we only enable it where the performance benefit is critical.
 ## Note this only applies to 2D graphics; 3D graphics always use OpenGL.
@@ -39,10 +43,10 @@ elif 'darwin' in sys.platform: ## openGL can have a major impact on mac, but als
     if QtGui.QApplication.instance() is not None:
         print('Warning: QApplication was created before pyqtgraph was imported; there may be problems (to avoid bugs, call QApplication.setGraphicsSystem("raster") before the QApplication is created).')
     if QtGui.QApplication.setGraphicsSystem:
-        QtGui.QApplication.setGraphicsSystem('raster')  ## work around a variety of bugs in the native graphics system 
+        QtGui.QApplication.setGraphicsSystem('raster')  ## work around a variety of bugs in the native graphics system
 else:
-    useOpenGL = False  ## on windows there's a more even performance / bugginess tradeoff. 
-                
+    useOpenGL = False  ## on windows there's a more even performance / bugginess tradeoff.
+
 CONFIG_OPTIONS = {
     'useOpenGL': useOpenGL, ## by default, this is platform-dependent (see widgets/GraphicsView). Set to True or False to explicitly enable/disable opengl.
     'leftButtonPan': True,  ## if false, left button drags a rubber band for zooming in viewbox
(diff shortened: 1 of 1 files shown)


caenrigen:
The tests will fail due to the new `import appnope`
Is there a `requirements` file to include that?

j9ac9k:
Thanks for the PR @caenrigen,

Some of the other maintainers may have more to say, I'm reluctant but willing, to add this change, provided it's wrapped inside a `try` /`except ImportError` block, so pyqtgraph does not error if `appnope` isn't installed.

This PR as it stands would break `pyqtgraph` on Windows and Linux systems altogether as I assume the `appnope` package is not installable on those platforms?

`pyqtgraph` specifies the required dependencies in `setup.py`, but I do not think this should be placed there, and instead a blurb to `README.md` can be added.

j9ac9k:
Upon further reflection I'm leaning towards not merging this.  A dependency like appnope should reside in the end application, not in the library IMO.

I'll leave the PR open in case other maintainers have thoughts.

caenrigen:
> Thanks for the PR @caenrigen,
> 
> Some of the other maintainers may have more to say, I'm reluctant but willing, to add this change, provided it's wrapped inside a `try` /`except ImportError` block, so pyqtgraph does not error if `appnope` isn't installed.
> 
> This PR as it stands would break `pyqtgraph` on Windows and Linux systems altogether as I assume the `appnope` package is not installable on those platforms?

@j9ac9k, It is :) the `appnope` makes sure to only have any effects on macOS and it makes the appropriate checks on other operating systems.
From the source:
```python
if sys.platform != "darwin" or V(platform.mac_ver()[0]) < V("10.9"):
    from ._dummy import *
else:
    from ._nope import *
```
Therefore, I believe it should not be treated a special case.
> 
> `pyqtgraph` specifies the required dependencies in `setup.py`, but I do not think this should be placed there, and instead a blurb to `README.md` can be added.

@j9ac9k I would say it is safe to include it in the dependencies and have no one dealing with this issue over and over again. A note about it in the README should indeed be included for reference and for who might want to look into the problem deeper.

j9ac9k:
I installed appnope on a linux VM, so the build complexity isn't there.... but I'm still thinking this is an application level dependency, not a library level one...

I'll leave this PR open for a while longer to give other maintainer a chance to chime in.

j9ac9k:
Hi @caenrigen 

Sorry for the long time to follow up.

I talked this over w/ some of the other maintainers, we decided we're not going to integrate `appnope` into `pyqtgraph`.  For one, even though app nap may create head aches for pyqtgraph users, it's not for pyqtgraph to decide that app nap _should_ be disabled, that's a decision for the application developer.

That said, we would accept a PR if there is some functionality where pyqtgraph emits a warning when app nap is engaged as pyqtgraph is running; along with a blurb in the README about symptoms that app nap may cause and how to work around them.

Thanks again for the PR, and I apologize for taking so long to get back to you.  Let me know if you have any other questions.

caenrigen:
Hey!

Thank you for the reply.

That sounds like a good decision.

If I dive deeper into developing pyqtgraph applications myself I will work on your suggestion :)

Thank you for maintaining this package, it keeps our science running!!

PR #1097, opened 2019-12-16
pretty-print log-scale axes labels

paulmueller:
This patch pretty-prints log-scale axes tick labels.

before:

after:

This probably does not work for Python2 because of the unicode characters in the labels.


paulmueller:
I added a check such that this only applies to Python 3+.

j9ac9k:
Closing/Re-Opening to trigger new CI run

j9ac9k:
CI is green across the board, thanks for the PR @paulmueller sorry I didn't get to this for so long.  Merging.

PR #1101, opened 2019-12-24
Fix small oversight in LegendItem

### LINKED ISSUE #1094: LegendItem.setParentItem fails when no offset is provided
### Short description
In the current development version, calling setParentItem() on a LegendItem without offset results in an exception and crash.
This is because the offset value seems now to be saved in self.opts["offset"] while it was in self.offset before.
Right now setParentItem() checks if an offset exists by checking in the wrong place but uses the right offset afterwards. If no offset is set, a crash ensues
```python
if self.offset is not None:
            offset = Point(self.opts['offset'])
```
Changing to the following code fixes the problem:
```python
if self.opts['offset'] is not None:
            offset = Point(self.opts['offset'])
```

### Code to reproduce

```python
import pyqtgraph as pg

plt = pg.plot()
plt.setWindowTitle('pyqtgraph example: Legend')
l = pg.LegendItem((100, 60))
l.setParentItem(plt.graphicsItem())


# Start Qt event loop unless running in interactive mode or using pyside.
if __name__ == '__main__':
    import sys
```

### Expected behavior
The plot should be displayed normally

### Real behavior
The following exception occurs:

```
  File "/home/vin/test/testlegend.py", line 7, in <module>
    l.setParentItem(plt.graphicsItem())
  File "/home/vin/miniconda3/lib/python3.7/site-packages/pyqtgraph/graphicsItems/LegendItem.py", line 128, in setParentItem
    offset = Point(self.opts['offset'])
  File "/home/vin/miniconda3/lib/python3.7/site-packages/pyqtgraph/Point.py", line 35, in __init__
    QtCore.QPointF.__init__(self, *args)
  QPointF(): too many arguments
  QPointF(float, float): argument 1 has unexpected type 'NoneType'
  QPointF(QPoint): argument 1 has unexpected type 'NoneType'
  QPointF(QPointF): argument 1 has unexpected type 'NoneType'

```

### Tested environment(s)

 * PyQtGraph version: 0.11.0.dev0+gdb483f8
 * Qt Python binding: PyQt5 5.13.0 Qt 5.13.0 (same problem with PySide2 5.12.2 Qt 5.12.2)
 * Python version: 3.7
 * NumPy version: 1.17.3
 * Operating system: Linux Mint 19.1
 * Installation method: pip from latest github version

ixjlyons:
Fixes #1094

### DIFF
--- a/pyqtgraph/graphicsItems/LegendItem.py
+++ b/pyqtgraph/graphicsItems/LegendItem.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from .GraphicsWidget import GraphicsWidget
 from .LabelItem import LabelItem
 from ..Qt import QtGui, QtCore
@@ -123,7 +124,7 @@ class LegendItem(GraphicsWidget, GraphicsWidgetAnchor):
 
     def setParentItem(self, p):
         ret = GraphicsWidget.setParentItem(self, p)
-        if self.offset is not None:
+        if self.opts['offset'] is not None:
             offset = Point(self.opts['offset'])
             anchorx = 1 if offset[0] <= 0 else 0
             anchory = 1 if offset[1] <= 0 else 0
@@ -251,7 +252,3 @@ class ItemSample(GraphicsWidget):
 
             p.translate(10, 10)
             path = drawSymbol(p, symbol, size, pen, brush)
-
-
-
-


j9ac9k:
meant to get to this a while ago, sorry about the wait.  LTGM, merging.

PR #1112, opened 2020-01-09
Add opengl text item and example

JoeXinfa:


danielhrisca:
Why not follow the API of [TextItem](https://github.com/pyqtgraph/pyqtgraph/blob/develop/pyqtgraph/graphicsItems/TextItem.py)

PR #1117, opened 2020-01-16
Updated for aligned csv data output with separtor of spaces

dpzwick:
This feature would be useful to have a space separated csv-like output where the headers and data are aligned.

### DIFF
--- a/pyqtgraph/exporters/CSVExporter.py
+++ b/pyqtgraph/exporters/CSVExporter.py
@@ -13,7 +13,7 @@ class CSVExporter(Exporter):
     def __init__(self, item):
         Exporter.__init__(self, item)
         self.params = Parameter(name='params', type='group', children=[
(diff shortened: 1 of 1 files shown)


j9ac9k:
I'm pretty reluctant on this PR, although I'll certainly welcome input from other maintainers.  Introducing spaces into a CSV output may mess with any kind of CSV importing tools as CSVs represent coma separated values, parsers will likely thing the spaces that were inserted by pyqtgraph are actually part of the value they need to read in.

I don't imagine other libraries that write csv outputs (pandas) have an option to do this?

ksunden:
I am also inclined to say out of scope... For the record, pandas does allow spaces (or _any_ one character string) to be the separator, but they do _not_ do alignment.

j9ac9k:
I'm going to close this out, as I haven't heard anyone else mention it.  Thanks for the PR @dpzwick if there is a use case for the CSV exporter needs to implement for functional improvements, please do let us know in the issue tracker.

PR #1123, opened 2020-02-09
Embed pyside2uic from pyside2 5.13

### LINKED ISSUE #1102: pyqtgraph.Qt.loadUiType broken for PySide2 >= 5.14
https://dev.azure.com/pyqtgraph/pyqtgraph/_build/results?buildId=557&view=logs&j=1d023af0-a188-58ef-16b4-2e987dd32dd1&t=addebc46-66b7-583c-f559-d6f453020f3e&l=308

There's a note here, though I'm not sure what it means precisely: https://wiki.qt.io/Qt_for_Python_Development_Notes#24._October_2019

Location of the issue for reference: https://github.com/pyqtgraph/pyqtgraph/blob/6e9a48558c010803883f0f83a7170581098b978a/pyqtgraph/Qt.py#L91

campagnola:
This fixes #1102 by copying pyside2uic out of pyside2 v5.13 and embedding it inside pyqtgraph.util. 
I have also copied the license file from https://code.qt.io/cgit/pyside/pyside-tools.git/tree/LICENSE-uic, which is dual BSD/GPL. I have no idea what legal repercussions this has (if any) for pyqtgraph users.

This solution makes me squirm, so I am leaving it here for discussion and review. I don't immediately see anything better, though. As far as I can tell, there is no replacement in pyside2 5.14 for the `loadUiType` functionality that was lost when they dropped pyside2uic (and furthermore the new ui compiler chokes on `examples/designerExample.ui`).

Another possible fix is just to declare pyqtgraph's incompatibility with 5.14 (actually the last version I could find that worked at all was 5.11) in some way. Maybe this can be done in setup.py, or maybe we can just raise an exception recommending 5.11.

### DIFF
--- a/pyqtgraph/Qt.py
+++ b/pyqtgraph/Qt.py
@@ -104,7 +104,10 @@ def _loadUiType(uiFile):
     if QT_LIB == "PYSIDE":
         import pysideuic
     else:
-        import pyside2uic as pysideuic
+        try:
+            import pyside2uic as pysideuic
+        except ImportError:
+            from .util import pyside2uic
     import xml.etree.ElementTree as xml
 
     parsed = xml.parse(uiFile)
--- a/pyqtgraph/Qt.py
+++ b/pyqtgraph/Qt.py
@@ -107,7 +107,7 @@ def _loadUiType(uiFile):
         try:
             import pyside2uic as pysideuic
         except ImportError:
-            from .util import pyside2uic
+            from .util import pyside2uic as pysideuic
     import xml.etree.ElementTree as xml
 
     parsed = xml.parse(uiFile)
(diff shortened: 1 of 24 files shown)


cmaureir:
Oh, I didn't know you had this dependency.
In any case, I think it will be really nice if you can open a Suggestion [here](https://bugreports.qt.io/projects/PYSIDE/issues)  to implement similar functionality in PySide2, regarding the `loadUiType` you are currently using, in that way you can have a similar behavior and avoid copying over the files from `pysideuic` (which was really broken and outdated).

If there are other requests regarding what PySide2 provides, feel free to open more suggestions, so we can look into it.

Regarding the tools and licenses, they were GPLv2 (AFAIK), so I don't know how that might affect the `pyqtgraph` license, but since you are already using PyQt stuff (which is also GPL) I guess you are fine, but IANAL.

j9ac9k:
Hi @cmaureir, thanks for being open to the a modification in the pyside2 library.

@campagnola if pyside2 folks are willing to restore functionality like in previous versions I would vote this issue no longer be considered a show stopper for 0.11.

campagnola:
@cmaureir if we could have loadUiType back in pyside2, that would be awesome!

j9ac9k:
I have created [PYSIDE-1223](https://bugreports.qt.io/browse/PYSIDE-1223)

PR #1125, opened 2020-02-10
GraphicsObject.py, GraphicsObject().itemChange():

jakimowb:
fixes "TypeError: unable to convert a QVariant back to a Python object" by replacing value == QtCore.QVariant() with None

### DIFF
--- a/pyqtgraph/graphicsItems/GraphicsObject.py
+++ b/pyqtgraph/graphicsItems/GraphicsObject.py
@@ -18,6 +18,8 @@ class GraphicsObject(GraphicsItem, QtGui.QGraphicsObject):
         GraphicsItem.__init__(self)
         
     def itemChange(self, change, value):
+        if value == QtCore.QVariant():
+            value = None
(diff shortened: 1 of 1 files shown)


j9ac9k:
Closing and Opening PR to trigger a new CI run, ...for debugging an issue unrelated to this PR.

j9ac9k:
Closed and Opened this PR with the new CI system which should pass 100% green....

j9ac9k:
PySide/PySide2 bindings _really_ don't like this change... you can view results here:

https://dev.azure.com/pyqtgraph/pyqtgraph/_build/results?buildId=626&view=results

Looks like a test on pyside1 just hangs, and on pyside2 we actually segfault.

@jakimowb can you verify this fix works locally for you if you're running pyside bindings of some kind?

j9ac9k:
I'm going to close out this PR as I can't replicate the issue that @jakimowb is facing, and this fix blows up our PySide pipelines.

@jakimowb if you can provide some code that replicates this issue, please open up an issue so we can try and figure out a fix that will work for all bindings.

PR #1131, opened 2020-02-21
Update CI/tox and Enable More Tests

j9ac9k:
Currently azure pipelines is starting to get a little dated (using macOS 10.13, older windows and ubuntu images); and I'm not making use of the `free` conda channel for python2 (instead I'm downloading/installing an old version of miniconda presently.

This should be addressed, along with the fragile `tox.ini` file, and some tests should be enabled.

Furthermore, the CI should be able to compensate for PySide2 losing the loadUIFile functionality as well.


j9ac9k:
I think I'm ready to remove the WIP, you can see the results at this link; for some reason, azure is choking on the git checkout step on one of the windows tasks, once it fails, I will restart that one.

https://j9ac9k.visualstudio.com/pyqtgraph-fork/_build/results?buildId=416

This PR includes a handful of changes

1. Change to CI system
  * I now use latest images for all OSs, and have removed PySide2=5.9 configuration.  That configuration was tested before due to that being the newest version of pyside that could be installed via conda/conda-forge, but that is no longer the case.
  * No more having to downgrade conda, instead py2 now uses the `free` channel
  * All conda-versions across platforms are now the same guaranteed by running `conda update --all --yes --quiet` early on in the process.  This is done due to the versions of conda being available on each platform was different, causing for funky if statements in the CI pipeline
2. Change to the test suite
  * New skips are added for the examples to compensate for PySide2 removing the loadUiType functionality from 5.14.x
  * Some tests have been un-skipped.  A few tests I was able to identify what the issue was, make the corrective action, and enable them.
  * All warnings are now either fixed, or filtered.  If you run the test suite and see a warning, you should look closely from this point on.
3.  Change to tox
  * `tox` working before was some miracle, I really goofed with my use of `-` in the names of the dependencies.  I renamed the dependencies to not use `-` (which is a special character when specifying what dependencies `tox` needs to install
  * `tox` on macOS 10.15 should pass with zero warnings and zero errors, I could not verify on other OSs, but I encourage other people to give it a go.
4.  Fix to test_eq, `dict1 = deepcopy(dict2)` on python2 does not guarantee that the insertion order of `dict1` is the same as the insertion order of `dict2`.  On some configurations, `test_eq` would fail, sometimes it would pass.  Now it passes reliably.
5.  Updates to readme/contributing pages, not a lot here; made a note of pyside2's lack of loadUiType, and what that means for pyqtgraph users.

pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -1078,7 +1078,7 @@ def drawPicture(self, p, axisSpec, tickSpecs, textSpecs):
             p.setFont(self.tickFont)
         p.setPen(self.textPen())
         for rect, flags, text in textSpecs:
-            p.drawText(rect, flags, text)
+            p.drawText(rect, flags.__int__(), text)
```
campagnola: Just curious--is this somehow different from `int()`?

j9ac9k: i don't think so, I actually noticed this shortly after commenting above ... I'll test to make sure `int()` works, and update the PR.

j9ac9k:
I removed some unused files, will merge momentarily.  After this is merged, CI should pass reliably.

PR #1133, opened 2020-03-01
compatibility-fix for py2/pyside

Zorkator:
replaced py3-specific string construction.
compatible with pyside/pyside2/PyQt#.


j9ac9k:
Hi @Zorkator thanks for the PR.

I'm not opposed to merging this, but I would like to have a test here to make sure this code works across all configurations (which has been generally tricky to do).

j9ac9k:
Closing/Opening to trigger new CI run

ixjlyons:
There's currently no test coverage of `ImageExporter`. I confirmed locally that this works for py2-pyside, py2-pyqt4, py3-pyside2, and py3-pyqt5 and that `ImageExporter` is currently broken for py2-pyside. I submitted #1151 to add a regression test.

Issue #31, opened 2014-02-07
PlotCurveItem disappears with large ViewBox scale factor


campagnola:
Example:

```
import pyqtgraph as pg
x = pg.np.linspace(0, 1, 1000)
plt = pg.plot(x, x**20)
plt.setYRange(0, 1e-8)
```

Drag the plot downward and it begins to flicker when the plot crosses the viewbox boundary.

This appears to be due to a Qt bug; report filed here: 
https://bugreports.qt-project.org/browse/QTBUG-36720

Original report discussion is here: 
https://groups.google.com/forum/#!msg/pyqtgraph/oSLHg8N-0JE/-7kT0Vgg4DoJ

petermoran:
I'm encountering this issue a lot with time series plots.  It still seems to exist in Qt5, but the QT bug was closed because it was for Qt4.

I don't know enough Qt to be able to translate the example bug script from 4 to 5.  Any chance someone could please resubmit the report for Qt5?

Anyone have any ideas for a work around?  I'm wondering if I could rescale the axes in pyqtgraph in some way...

petermoran:
Managed to port the script to Qt5.  I've uploaded the example and requested that they reopen the issue.

campagnola:
Thanks @petermoran !

petermoran:
Thanks for the awesome package @campagnola 

For anyone else encountering this I've managed to work around it by using some custom clipToView functionality.  I override the viewRangeChanged method so that whenever the user is zoomed in very close it clips the data to only plot what is in the current view.  This seems to work well.

petermoran:
I had a request to provide my work around.  First we have to enable clipToView for the plot:

```
import pyqtgraph as pg

x = pg.np.linspace(0, 1, 1000)
plt = pg.plot(x, x**20)
plt.setYRange(0, 1e-8)

# enable clipToView
# TODO for some reason passing clipToView=True to pg.plot() doesn't apply the option correctly?
plt.setClipToView(True)
```

Then I use an alternate clipToView algorithm in PlotDataItem.py.  Here's the diff:

```
diff --git a/pyqtgraph/graphicsItems/PlotDataItem.py b/pyqtgraph/graphicsItems/PlotDataItem.py
index ab2eef9..1341f19 100644
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
[...]

PR #1140, opened 2020-03-10
dynamic range limiting in PlotDataItem

NilsNemitz:
Workaround for issue #31 
PlotCurveItem disappears with large ViewBox scale factor

(Code editor ate trailing whitespace, sorry.)

Test code:
```
from PyQt5 import QtWidgets
import sys
import pyqtgraph as pg

class MainWindow(QtWidgets.QMainWindow):
        super(MainWindow, self).__init__(*args, **kwargs)
        self.graphWidget = pg.PlotWidget()
        self.setCentralWidget(self.graphWidget)

        y_data_a = [  10,  10, -1,  1,  0,  10,  10]
        y_data_b = [ 2e9, 2e9,  0,  1, -1, 2e9, 2e9]
        y_data_c = [-2e9,-2e9,  0, -1,  1,-2e9,-2e9]
        x_data = [idx for idx, val in enumerate(y_data_a)]
        self.graphWidget.setYRange( -1.1, 1.1)
        item0 = pg.PlotDataItem( x=x_data, y=y_data_a, pen='g')
        item1 = pg.PlotDataItem( x=x_data, y=y_data_b, pen='r')
        item2 = pg.PlotDataItem( x=x_data, y=y_data_c, pen='m')

        disable = False
        for item in [item0, item1, item2]:
            self.graphWidget.addItem( item )
            if disable: item.setLimitDynamicRange(False)

def main():
    main = MainWindow()
    main.show()
    sys.exit(app.exec_())

if __name__ == '__main__':
    main()
```


j9ac9k:
Hi @NilsNemitz thanks for the PR!

A quick question, is `self.opts['limitDynamicRange']` intended to just take a boolean?  if so does the method `setLimitDynamicRange` take a boolean as the input?  

```python
    def setLimitDynamicRange(self, limit):
        if self.opts['limitDynamicRange'] == limit:
            return
        self.opts['limitDynamicRange'] = limit
        self.xDisp = self.yDisp = None
        self.updateItems()
```

If so, can you replace the `==` with the `is` operator?

Also, can you add a docstring on that method?

Thanks for the PR!

NilsNemitz:
Hi @j9ac9k !
I am taking your quick reply as encouragement and have updated pull request #1140 . 
1) Operator is changed
2) docstring is added to what is now "setDynamicRangeLimit()" and to the \__init\__() function's argument parsing.
3) setDynamicRangeLimit(limit) now takes an argument that exposes the actual limit to the user.
4) added some checks to bypass clipping the data in the 99% of cases that do not need it.
For the last point, I have added a function dataRect() parallelling the existing viewRect(). This stores the range of the data and is only calculated once per data set (cleared in setData()). The functionality is a little too close to dataBounds(), but *that* relies on functionality of the scatter and curve children... which see the data only after clipping.
I hope we can find a solution with minimal computational overhead and enable the workaround to the disappearing plots by default. Those who seek ultimate performance will probably call ScatterPlotItem and PlotCurveItem directly.
There is still some improvement available by modifying viewRangeChanged() to not invalidate xDisp and yDisp unless necessary, but that is a larger change than I want to propose right now.
Let me know what else I can do :)

Extended testing code:
```
from PyQt5 import QtWidgets
import sys
import pyqtgraph as pg

class MainWindow(QtWidgets.QMainWindow):
        super(MainWindow, self).__init__(*args, **kwargs)
        self.graphWidget = pg.PlotWidget()
        self.setCentralWidget(self.graphWidget)

        y_data_a = [  10,  10, -1,  1,  0,  10,  10]
        y_data_b = [ 2e9, 2e9,  0,  1, -1, 2e9, 2e9]
        y_data_c = [-2e9,-2e9,  0, -1,  1,-2e9,-2e9]
        x_data = [idx for idx, val in enumerate(y_data_a)]
        self.graphWidget.setYRange( -1.1, 1.1)
        item0 = pg.PlotDataItem( x=x_data, y=y_data_a, pen='g')
        item1 = pg.PlotDataItem( x=x_data, y=y_data_b, pen='r')
        item2 = pg.PlotDataItem( x=x_data, y=y_data_c, pen='m')
        # white curve demonstrates bug and disappears with viewport change
        # also demonstrates magic argument parsing
        item3 = pg.PlotDataItem( x=x_data, y=y_data_c, pen='w', dynamicRangeLimit=None)

        disable = False
        reconfigure = False
        for item in [item0, item1, item2, item3]:
            self.graphWidget.addItem( item )
            if disable:
                item.setDynamicRangeLimit(None) # curves will disappear with viewport change
            if reconfigure:
                item.setDynamicRangeLimit(2) # does not make any sense, but shows up as visible distortion
                item.setDynamicRangeLimit(2) # check that repeated setting works

def main():
    main = MainWindow()
    main.show()
    sys.exit(app.exec_())

if __name__ == '__main__':
    main()
```

j9ac9k:
my activity here is indeed feast or famine with nothing in between, I know that's totally not fair to contributors like yourself, and for that I apologize... just not sure what else I can do on a volunteer effort.

Anyway, back to the PR!

Looks like the CI system doesn't like your changes for the python2 builds:

https://dev.azure.com/pyqtgraph/pyqtgraph/_build/results?buildId=647

```
SyntaxError: Non-ASCII character '\xc2' in file pyqtgraph\graphicsItems\PlotDataItem.py on line 357, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

NilsNemitz:
Your work is very appreciated! I did not expect to get such a quick reply.
Sorry about the errors. I am not familiar with the CI system and it took me too long to get to the actual error. "bash exited with code 1" is not very helpful by itself.
Thank you for the hint. All checks pass now. Python 2 does not share my love for unicode.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -323,28 +329,42 @@ def setDownsampling(self, ds=None, auto=None, method=None):
...
-        
-        
+
+    def setDynamicRangeLimit(self, limit):
+        """
+        Limit the off-screen positions of data points at large magnification
+        This avoids errors with plots not displaying because their visibility is incorrectly determined. The default setting repositions far-off points to be within +-1E+06 times the viewport height.
+        =============== =============================================================
+        **Arguments:**
+        limit           (float or None) Maximum allowed vertical distance of plotted points in units of viewport height.
```
j9ac9k: wait, I thought limit was of type `bool`, if it can be `float`, then we should most definitely not be using the `is` operator

j9ac9k: unless explicitly checking for `None`

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -69,59 +69,66 @@ def __init__(self, *args, **kargs):
...
+                              multiple line segments per pixel. This can improve performance when
+                              viewing very high-density data, but increases the initial overhead
+                              and memory usage.
+            clipToView        (bool) If True, only plot data that is visible within the X range of
+                              the containing ViewBox. This can improve performance when plotting
+                              very large data sets where only a fraction of the data is visible
+                              at any time.
+            dynamicRangeLimit (float or None) Limit off-screen positions of data points at large
+                              magnification to avoids display errors. Disabled if None.
+        =============== =============================================================
```
j9ac9k: Looks like you need to indent this line over another 4 spaces (as well as the next couple of lines)

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -323,28 +329,42 @@ def setDownsampling(self, ds=None, auto=None, method=None):
...
+    def setDynamicRangeLimit(self, limit):
+        """
+        Limit the off-screen positions of data points at large magnification
+        This avoids errors with plots not displaying because their visibility is incorrectly determined. The default setting repositions far-off points to be within +-1E+06 times the viewport height.
+        =============== =============================================================
+        **Arguments:**
+        limit           (float or None) Maximum allowed vertical distance of plotted points in units of viewport height.
+                        'None' disables the check for a minimal increase in performance. Default is 1E+06.
+        """
+        if self.opts['dynamicRangeLimit'] is limit:
```
j9ac9k: should this be checking `if None`?

j9ac9k: maybe it should be `if self.opts['dynamicRangeLimit'] is None or limit == self.opts['dynamicRangeLimit']`?

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -564,10 +585,24 @@ def getData(self):
...
-                    
+
+            if self.opts['dynamicRangeLimit'] is not None:
+                view_range = self.viewRect()
+                if view_range is not None:
+                    data_range = self.dataRect()
+                    if data_range is not None:
+                        view_height = view_range.height()
+                        lim = self.opts['dynamicRangeLimit']
+                        # print('data:', data_range.height(), '  view:', view_height, '  limit:', lim )
```
j9ac9k: remove commented out print statements.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -586,12 +621,32 @@ def getData(self):
...
-                
-                    
+
             self.xDisp = x
             self.yDisp = y
         return self.xDisp, self.yDisp
 
+    def dataRect(self):
+        """
+        Returns a bounding rectangle (as QRectF) for the full set of data
```
j9ac9k: If this function can return `None`, that should be indicated in the docstring (as well as a description of why that might be the case)

j9ac9k:
Thanks for the update! 

Azure Pipelines isn't the most UI friendly interface, I think if you click on the "bash exited with code 1" it takes you to the portion... 

I made some inline comments, I could be off-base on some of them, if so, let me know.

NilsNemitz:
Thank you for the helpful comment @j9ac9k !
I have applied your suggestions to PR  #1140 

One additional change: I have moved the limiting code after downsampling (instead of before). 
a) Shifting even far-offscreen points might give visible effects with downsampling in mean mode.
b) limiting is faster after downsampling reduces the number of datapoints.

Two observations:
a) Should self.xClean and self.yClean be removed? I ran a full-text search over the entire pyqtgraph folder, and the only hits were three lines in PlotDataItem.py where they get set to None. I think they are never used, never set to an actual value.
b) Is it intentional that PlotItem.addItem() (in line 507) calls PlotDataItem.setDownsampling() and overwrites any earlier setting? It seems that this makes the documented way of setting downsampling through \__init__()\ parameters of PlotDataItem quite useless. In my tests, any downsampling I set before the .additem() call gets reverted to the default of "no downsampling". The same is probably true for ClipToView, FFT, Alpha, and PointMode (but I did not test that.) Is there a use case that does not involve adding to some variant of PlotItem?

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -586,12 +601,45 @@ def getData(self):
...
+    def dataRect(self):
+        """
+        Returns a bounding rectangle (as QRectF) for the full set of data.
+        Will return None if there is no data or if all values (x or y) are NaN.
+        """
+        if self._dataRect is not None:
+            return self._dataRect
+        if self.xData is None or self.yData is None:
+            return None
+        ymin = np.nanmin(self.yData)
```
j9ac9k, line 630: we should catch the runtime warning being emitted here as we account for the `nan` case just below.

This can be done with the following code

```python
import warnings

...

with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    np.nanmin(self.yData)
```

j9ac9k:
Hi @NilsNemitz sorry it's taken way too long for me to follow up.  I changed the base of where this PR should be pointed at, resolved a merge conflict, and added one more comment... enough time has passed by I would understand if you don't want to implement that last suggestion, and I don't mind implementing it myself.

Thanks again for this PR!

NilsNemitz:
Dear @j9ac9k ,
I wish I was more fluid with the collaborative aspects of git, but I think I have managed to update the pull request to suppress the unneeded "all-NaN" warnings. Sorry for the persistent "trailing whitespace" changes. I fixed my IDE now, but the damage in PlotDataItem.py is already done. :(

I tried my previous test cases; The changes still seem to fix the issue of disappearing plots.

I tested with all-NaN x-data, y-data, and both. No problem was observed. I still got an "all-NaN" warning from the code in PlotCurveItem.py. Since there is only a single line with a np.nanmin / np.nanmax call, I suppressed the warning there, too.

Now the code is silent for an all-NaN (curve) plot. I suspect that scatter plots will still throw a warning, but I did not want to touch more files than necessary for now.

Let me know if you have more suggestions. Any idea why the "pre_build build_docs" test is failing?

j9ac9k:
Hey @NilsNemitz thanks for following up; again apologize for the long time that elapsed.

Regarding the docs the error message is this:

```
/home/vsts/work/1/s/doc/source/../../pyqtgraph/graphicsItems/PlotDataItem.py:docstring of pyqtgraph.PlotDataItem.setDynamicRangeLimit:6:Unexpected indentation.
```

I'm not that familiar w/ sphinx, but I can try and debug this for you as I have my local machine setup to build documentation and it's much easier to debug locally than on the CI level.  I'll hop onto your branch, and see if I can figure out the necessary modification to the docstring to make sphinx happy.

NilsNemitz:
@j9ac9k: Let's see if this change fixes it, otherwise it would be great if you could have a look. Thank you for your time!

j9ac9k:
LGTM, thanks for the PR @NilsNemitz  merging!

PR #1142, opened 2020-03-14
Miscellaneous doc fixups

### LINKED ISSUE #686: Missing documentation for BarGraphItem
I suppose it should be described [here](http://www.pyqtgraph.org/documentation/graphicsItems), shouldn't it?

ixjlyons:
Iterated until the docs built without warnings and made some drive-by improvements along the way.

Other things of note:

- Reworked installation page a little bit.
- Added representation of classes in graphicsWindows.py even though they're deprecated. Remove when the classes are actually removed.
- Added `BarGraphItem` (fixes #686)
- Removed extra parametertree doc from widgets TOC (it's already documented at the top level & apireference). widgets/parametertree rendered an empty page previously.
- Moved dockarea to top level of apireference (similar to parametertree, flowchart). Eventually this should be expanded I think
- Made sphinx_rtd_theme default. It seems to be rendered with that theme on readthedocs and pyqtgraph.org/documentation

### DIFF
--- a/doc/source/apireference.rst
+++ b/doc/source/apireference.rst
@@ -13,6 +13,7 @@ Contents:
     3dgraphics/index
     colormap
     parametertree/index
+    dockarea
     graphicsscene/index
     flowchart/index
     graphicswindow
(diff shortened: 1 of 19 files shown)


j9ac9k:
Thanks for the PR @ixjlyons

PR #1145, opened 2020-03-24
Pass showAxRect keyword arguments to setRange

lcmcninch:
Subclasses of ViewBox may wish to, for example, set padding when executing a scale or translation.


j9ac9k:
Thanks for the PR @lcmcninch , looks good to me, merging.

PR #1147, opened 2020-03-29
Update ptime.py

Derrick56007:
The function time.clock() has been deprecated since Python 3.3


ixjlyons:
Thanks for the pull request, but this should already be working in the current develop branch (see the line `from time import perf_counter as clock` added as part of #980)

PR #1151, opened 2020-04-03
ImageExporter py2-pyside fix with test

ixjlyons:
Adds a test on top of #1133. It pretty much only hits the code that #1133 fixes, but it's better than nothing. We should really have more coverage of `ImageExporter` - maybe for a future PR?

### DIFF
--- a/pyqtgraph/exporters/ImageExporter.py
+++ b/pyqtgraph/exporters/ImageExporter.py
@@ -47,10 +47,7 @@ class ImageExporter(Exporter):
     
     def export(self, fileName=None, toBytes=False, copy=False):
         if fileName is None and not toBytes and not copy:
-            if QT_LIB in ['PySide', 'PySide2']:
-                filter = ["*."+str(f, encoding='utf-8') for f in QtGui.QImageWriter.supportedImageFormats()]
-            else:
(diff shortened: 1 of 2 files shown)


j9ac9k:
hmm...that was a weird CI failure, going to rerun that one failed test...if it's good, I'll merge

j9ac9k:
still failing, re-running CI on an unrelated PR and see if I get the same failure....

j9ac9k:
rerunning CI on just the dev branch to see if unrelated.... can follow along here:

https://dev.azure.com/pyqtgraph/pyqtgraph/_build/results?buildId=660&view=results

ixjlyons:
Just a blind guess, but I didn't explicitly instantiate a `QApplication` instance at the module level though other test modules seem to do so.

j9ac9k:
oh, that's probably it!  Mind adding that into the test? ...not sure if this is a Qt 5.9 issue, a macOS issue or a python 3.6 issue.... (or all 3).

j9ac9k:
I ran the CI on the develop branch as it stands today, and it doesn't have this issue, so this error is definitely a result of a change in this PR.

ixjlyons:
Done. The funny thing is I originally had it in there but removed it because it didn't seem necessary :sweat_smile: Hopefully that's it

j9ac9k:
It passed! Merging!

PR #1098, opened 2019-12-19
#406 does not work for 3D data

jkotan:
Without the PR on gets the following error for RGB data

```
Traceback (most recent call last):
  File "/home/jkotan/sources/lavue/lavuelib/imageDisplayWidget.py", line 64, in paint
    _pg.ImageItem.paint(self, p, *args)
  File "/home/jkotan/testinst/lib/python2.7/site-packages/pyqtgraph/graphicsItems/ImageItem.py", line 439, in paint
    self.render()
  File "/home/jkotan/testinst/lib/python2.7/site-packages/pyqtgraph/graphicsItems/ImageItem.py", line 431, in render
    argb, alpha = fn.makeARGB(image, lut=lut, levels=levels)
  File "/home/jkotan/testinst/lib/python2.7/site-packages/pyqtgraph/functions.py", line 1187, in makeARGB
    imgData[nanMask, 3] = 0
IndexError: too many indices for array
```

PR #1152, opened 2020-04-04
Fnanfix test

ixjlyons:
Follow-up for #1098. Adds a couple tests to make sure both grayscale and rgb(a) input images with NaNs work.


j9ac9k:
Thanks @jkotan  and @ixjlyons for the PR!  All tests pass, merging!

PR #1159, opened 2020-04-08
Fix: Update axes after data is set

### LINKED ISSUE #1144: Plots after event loop has started does not update axes immediately
### Short description

When plotting data after the event loop has started, the axes are not updated immediately.

Results of git bisection:
d77ad273c71870881c3fda3d2b49d3488f100425 is the first bad commit
commit d77ad273c71870881c3fda3d2b49d3488f100425
Author: 2xB <31772910+2xB@users.noreply.github.com>
Date:   Sat Aug 17 20:52:45 2019 +0200

### Code to reproduce

```python
import pyqtgraph

QtCore = pyqtgraph.Qt.QtCore
app = pyqtgraph.mkQApp()

plotwgt = pyqtgraph.PlotWidget()
curve = pyqtgraph.PlotCurveItem()
plotwgt.addItem(curve)

def draw_later():
    curve.setData(list(range(100)))
        
timer = QtCore.QTimer.singleShot(0, draw_later)
plotwgt.show()
app.exec_()
```

### Expected behavior

plot_good

### Real behavior

plot_bad

### Tested environment(s)

 * PyQtGraph version: 0.11.0rc0
 * Python version: Python 3.7.6
 * NumPy version: 1.18.1
 * Operating system: Windows 10 64-bits
 * Installation method: extract to local directory

2xB:
On data changes on PlotCurveItems, bounds are updated before the new data is set, leading to axes not performing autoscaling directly. This PR changes the order of execution for setting data and updating bounds to fix that.

Fixes #1144

### DIFF
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -365,11 +365,12 @@ class PlotCurveItem(GraphicsObject):
 
         #self.setCacheMode(QtGui.QGraphicsItem.NoCache)  ## Disabling and re-enabling the cache works around a bug in Qt 4.6 causing the cached results to display incorrectly
                                                         ##    Test this bug with test_PlotWidget and zoom in on the animated plot
+        self.yData = kargs['y'].view(np.ndarray)
+        self.xData = kargs['x'].view(np.ndarray)
+        
         self.invalidateBounds()
         self.prepareGeometryChange()
         self.informViewBoundsChanged()
-        self.yData = kargs['y'].view(np.ndarray)
-        self.xData = kargs['x'].view(np.ndarray)
 
         profiler('copy')
 


j9ac9k:
Hi @2xB Thanks for the PR

This CI failure that occurred is something that we've seen from minor changes, using having to do w/ the use of a QApplication (or lack-there-of).

I'm really on the verge of skipping that specific test.... it's failing just on macOS/Qt5.9/Python3.6 pipeline....

j9ac9k:
I really hate that this test failure is getting in the way of this PR, @ixjlyons as you were the last person to run into this, figure you may have an opinion?

2xB:
Let's test if this failure is consistent:

j9ac9k:
Good call on closing/opening ... I just reran the failed pipeline, which failed a second time for the same test... anyway, as it all passes now, I'm inclined to merge.

PR #1161, opened 2020-04-09
import numpy as np for lines 44 and 51

cclauss:
1. Undefined name `np` in pyqtgraph/PlotData.py
~2. To fix CI, make sure that `call_with_timeout()` polls at least once.~

@j9ac9k @2xB Your reviews please.

### DIFF
--- a/pyqtgraph/PlotData.py
+++ b/pyqtgraph/PlotData.py
@@ -1,3 +1,4 @@
+import numpy as np
 
 
 class PlotData(object):
@@ -50,7 +51,3 @@ class PlotData(object):
             mn = np.min(self[field])
             self.minVals[field] = mn
         return mn
-    
-    
-    
-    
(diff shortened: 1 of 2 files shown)


j9ac9k:
The CI doesn't like the change you made to test_exit_crash

```
>       for i in range(max(timeout // wait_per_poll, 1)):  # Poll at least once
E       TypeError: range() integer end argument expected, got float.
```

Fired up my own interpretter

```
>>> max(5.5 // 2, 1)
2.0
```

Probably should wrap that `max` statement as an `int`

j9ac9k:
Thanks for this PR @cclauss , sort of amazing we haven't noticed this before.

cclauss:
It can happen that another file has already imported numpy so things work but then if code get refactored and then things break.  This is safer.

PR #1164, opened 2020-04-09
Ensure call_with_timeout will call p.poll

cclauss:
Ensure `call_with_timeout()` will call `p.poll()` as discussed in #1163


cclauss:
Passes every time.  Closing.

PR #1166, opened 2020-04-11
pg.mkQApp: Pass non-empty string array to QApplication() as default

### LINKED ISSUE #1165: pg.mkQApp generates warning
### Short description
`pg.mkQApp()` generates warning `QSettings::value: Empty key passed`

### Code to reproduce
```python
import pyqtgraph as pg
pg.mkQApp()
```

### Expected behavior
No output

### Real behavior

Output to console:
```
QSettings::value: Empty key passed
QSettings::value: Empty key passed
```

### Tested environment(s)

 * PyQtGraph version: Latest commit from `develop` ( 6f4048596b080f71eaec544555159cb7c40c075f )
 * Qt Python binding: PyQt5 5.14.1 Qt 5.14.1
 * Python version: 3.8.2
 * NumPy version: 1.18.2
 * Operating system: Linux 5.5.7-1-MANJARO x86_64 19.0.2 Kyria
 * Installation method: Executed from source package

### Additional context
This is because `mkQApp` calls `QApplication([])`, which according to its documentation should never be called without arguments: See warning in https://doc.qt.io/qt-5/qapplication.html#QApplication

2xB:
According to Qt documentation, the `QApplication` initializer should never be called with an empty array: See warning in https://doc.qt.io/qt-5/qapplication.html#QApplication .
This manifests in a warning being printed to the console:
```
QSettings::value: Empty key passed
QSettings::value: Empty key passed
```

This PR changes `pg.mkQApp()` to pass "pyqtgraph" as application name and adds parameters to pass a different name and Qt command line arguments.
Further, some documentation to the function is added.

I do not know what the status of this function is, but I find it very convenient. Most examples use `QApplication([])` though, generating said warning. Therefore if this idea finds support, I'd extend this PR to replace every call to `QApplication([])` by `pg.mkQApp`, removing unnecessary warnings arising when launching the examples.

Fixes #1165.

### DIFF
--- a/pyqtgraph/Qt.py
+++ b/pyqtgraph/Qt.py
@@ -329,9 +329,18 @@ if m is not None and list(map(int, m.groups())) < versionReq:
 
 
 QAPP = None
-def mkQApp():
-    global QAPP    
+def mkQApp(name="pyqtgraph", qt_args=[]):
+    """
+    Creates new QApplication or returns current instance if existing.
+    
+    ==============  =================================================================================
+    **Arguments:**
+    name            Application name, passed to Qt
+    qt_args         Array of command line arguments passed to Qt
+    ==============  =================================================================================
+    """
+    global QAPP
     QAPP = QtGui.QApplication.instance()
     if QAPP is None:
-        QAPP = QtGui.QApplication([])
+        QAPP = QtGui.QApplication([name] + qt_args)
     return QAPP


j9ac9k:
Thanks for the PR @2xB was sort of ignoring this section of the test code base, hoping to eventually migrate to pytest-qt which would handle this case for us...

chrisjbillington:
My complaint in PR form: #1199

2xB:
Discussion continues in #1199.

PR #1168, opened 2020-04-12
GroupParameterItem: Did not pass changed options to ParameterItem

### LINKED ISSUE #788: Hiding nested group parameters
Hiding groups nested in groups does not work after the widget is shown. Look at this minimal example:

```
import os
import pyqtgraph.parametertree.ParameterTree as ParameterTree
from pyqtgraph.parametertree import Parameter, parameterTypes

class MyGroup(parameterTypes.GroupParameter):

    def __init__(self, *args):
        super().__init__(name='GroupA')
        self.addChild(parameterTypes.ListParameter(name='idx', values=[0, 1, 2, 3], value=0))
        for i in range(4):
            self.addChild(Parameter.create(name=str(i), type='group', children=[{
                'name' : 'member', 'type' : 'str'
            }]))
           # self.addChild(Parameter.create(name=str(i), type='str')) # works fine

    def treeStateChanged(self, param, changes):
        super().treeStateChanged(param, changes)
        if param is self.child('idx'):
            # Show the element at this index
            self.update_value()
        
    def update_value(self):
        # Destroy the old child
        idx = self.child('idx').value()
        for child in self.children():
            if child.name() != 'idx' and child.name() != str(idx):
                child.hide()
        self.child(str(idx)).show()

p = MyGroup()
tree.addParameters(p)
win = QtGui.QWidget()
layout = QtGui.QGridLayout()
win.setLayout(layout)
layout.addWidget(tree)
win.show()

if __name__ == '__main__':
    import sys
```

I create a Group that holds a list parameter, the value of which determines, which of the subgroups should be visible. When setting up the tree, all works well, only the group with name '0' is visible. However when changing the value of the list parameter 'idx', no changes appear whatsoever. Note that when using any parameter type other than group, (that is, instead of adding group parameters as children appending e.g. action, str, etc.) everything works just as intended.

2xB:
`ParameterItem` handles visibility changes in `optsChanged`. 
`GroupParameterItem` overrides this function, but never calls
the super function, resulting in visibility changes not being
applied. This PR fixes this by calling said function.

Fixes #788

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes.py
+++ b/pyqtgraph/parametertree/parameterTypes.py
@@ -437,8 +437,10 @@ class GroupParameterItem(ParameterItem):
         else:
             ParameterItem.addChild(self, child)
             
-    def optsChanged(self, param, changed):
-        if 'addList' in changed:
+    def optsChanged(self, param, opts):
+        ParameterItem.optsChanged(self, param, opts)
+        
+        if 'addList' in opts:
             self.updateAddList()
                 
     def updateAddList(self):


j9ac9k:
Somehow missed this one, sorry about that folks.  LGTM, merging.

PR #1169, opened 2020-04-12
Replace default list arg with None

ixjlyons:
Didn't catch #1165 in time. Just a suggestion to avoid issues with a mutable default argument.

I had also considered changing the function to accept a single optional list and use `['pyqtgraph']` if it's None, but either way seems fine.


j9ac9k:
bah, you're right, ... I should have known better, thanks for the fix @ixjlyons

2xB:
@ixjlyons My assumption was that `qt_args` is only read from and therefore it is safe to use a mutable default?

PR #1172, opened 2020-04-13
Fix: AxisItem tickFont is defined in two places while only one is used

2xB:
To set the tick font of `AxisItem`s, there are two options:
`setStyle({"tickFont":...})` and `setTickFont(...)`.
The first option sets `AxisItem.style['tickFont']`, the second
sets `self.tickFont`. Only `self.tickFont` is actually used.
This PR replaces all occurrences of the second variable with the first
variable, so both options work again. Also, documentation from
setStyle is copied to setTickFont.

[DRAFT]: This is a draft, since this PR collides with one commit suggested for the `DateAxisItem` PR #1154 that uses `self.tickFont` in a new location. To avoid conflicting PRs, this is held as a draft and updated once PR #1154 is merged.

### DIFF
--- a/doc/source/graphicsItems/dateaxisitem.rst
+++ b/doc/source/graphicsItems/dateaxisitem.rst
@@ -1,0 +1,8 @@
+DateAxisItem
+============
+
+.. autoclass:: pyqtgraph.DateAxisItem
+    :members:
+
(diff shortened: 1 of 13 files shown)


2xB:
Updated PR: #1180

PR #1174, opened 2020-04-14
Fixed a possible race condition with linked views

Ma-Shell:
The `linkedView()`-function of the `AxisItem` uses the attribute `_linkedView`, which is a weakref (or `None`). The current implementation of the changed event-handlers leaves a small timeframe for the corresponding viewbox to be deleted after the check against `None` but before being used. This is remedied by introducing a temporary local variable `lv`, which holds a strong reference, thereby preventing the deletion while the function is running.

I did not actually experience any crashes due to that issue and producing an error-case would be probably very hard, since the timeframe is very small and the problem can only appear if the gc is working on a different thread. However, the problem exists in theory and this fix is the clean solution to it.

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -1098,23 +1098,26 @@ class AxisItem(GraphicsWidget):
             self._updateHeight()
 
     def wheelEvent(self, ev):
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for the PR @Ma-Shell sorry for the delay with following up.  The changes look good to me.

PR #1175, opened 2020-04-14
Fix: Parameter tree ignores user-set 'expanded' state

### LINKED ISSUE #1130: ParameterTree, GroupParameter, saveState() and loadState() do not work / do not exist
Edited to clarify:

`ParameterTree` does not have a means of saving and restoring the collapsed/expanded state of its items. `GroupParameter` and `SimpleParameters` do have `saveState()` and `restoreState()`, but they do not affect the collapsed state of the tree, despite having these keywords in opts; changing the state of the GUI with my mouse (collapsing / expanding items) also does not change the value stored in `opts` or `saveState()`, nor does `setOpts()` or setting `opts` directly affect the collapse state state.

### Code to reproduce

```python
import pyqtgraph

app = pyqtgraph.QtGui.QApplication([])
tree = pyqtgraph.parametertree.ParameterTree()
tree.show()

p1 = pyqtgraph.parametertree.Parameter.create(name='test',  type='bool',  value=True, expanded=True)
p2 = pyqtgraph.parametertree.Parameter.create(name='test2', type='float', value=37.2, expanded=True)

tree.addParameters(p1)
p1.addChild(p2)

p1.setOpts(value    = False) # Works
p1.setOpts(expanded = False) # Does not update the GUI

state = p1.saveState() # Works based on ops, not GUI state
print('Should be True:', state['expanded'] == False)

p1.restoreState(state) # Should collapse the first parameter, but does not. Does update opts.

# Additionally, toggling the collapsed state of the GUI (via mouse, e.g.)
# has no effect on the output of p1.saveState()
```

Basically, 'expanded' seems to be unlinked with the GUI state. Is there a way to control whether a tree item is collapsed?

Thanks in advance! 
Jack

### Tested environment(s)

 * PyQtGraph version: 0.10.0 and 0.11.0rc0
 * Python version: 3.7
 * NumPy version: 1.15.4
 * Operating system: Linux Mint 19 & Windows 10
 * Installation method: conda

2xB:
When setting the 'expanded' state of parameters, this change is not applied
in the graphically visible tree. This commit changes that behaviour by
adding a clause in `ParameterItem.optsChanged` to react to that.

Also, an old and trivial out-commented debugging code line is removed.

Fixes #1130

### DIFF
--- a/pyqtgraph/parametertree/Parameter.py
+++ b/pyqtgraph/parametertree/Parameter.py
@@ -135,9 +135,12 @@ class Parameter(QtCore.QObject):
                                      (default=False)
         removable                    If True, the user may remove this Parameter.
                                      (default=False)
-        expanded                     If True, the Parameter will appear expanded when
-                                     displayed in a ParameterTree (its children will be
-                                     visible). (default=True)
+        expanded                     If True, the Parameter will initially be expanded in
+                                     ParameterTrees: Its children will be visible.
+                                     (default=True)
+        syncExpanded                 If True, the `expanded` state of this Parameter is
+                                     synchronized with all ParameterTrees it is displayed in.
+                                     (default=False)
         title                        (str or None) If specified, then the parameter will be 
                                      displayed to the user using this string as its name. 
                                      However, the parameter will still be referred to 
@@ -159,6 +162,7 @@ class Parameter(QtCore.QObject):
             'removable': False,
             'strictNaming': False,  # forces name to be usable as a python variable
             'expanded': True,
+            'syncExpanded': False,
             'title': None,
             #'limits': None,  ## This is a bad plan--each parameter type may have a different data type for limits.
         }
@@ -453,7 +457,7 @@ class Parameter(QtCore.QObject):
         Set any arbitrary options on this parameter.
         The exact behavior of this function will depend on the parameter type, but
         most parameters will accept a common set of options: value, name, limits,
-        default, readonly, removable, renamable, visible, enabled, and expanded.
+        default, readonly, removable, renamable, visible, enabled, expanded and syncExpanded.
         
         See :func:`Parameter.__init__ <pyqtgraph.parametertree.Parameter.__init__>`
         for more information on default options.
--- a/pyqtgraph/parametertree/ParameterItem.py
+++ b/pyqtgraph/parametertree/ParameterItem.py
@@ -146,10 +146,12 @@ class ParameterItem(QtGui.QTreeWidgetItem):
     def optsChanged(self, param, opts):
         """Called when any options are changed that are not
         name, value, default, or limits"""
-        #print opts
         if 'visible' in opts:
             self.setHidden(not opts['visible'])
         
+        if 'expanded' in opts:
+            self.setExpanded(opts['expanded'])
+    
     def editName(self):
         self.treeWidget().editItem(self, 0)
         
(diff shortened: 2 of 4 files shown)


jaxankey:
While we're at it, is there a similar action to take on the other options, e.g., "enabled" (I've wanted this in the past), "readonly" (maybe useful), "renamable", "strictNaming", and (maybe?!) "type"?

Also, I think the same should be applied to `GroupParameterItem`, perhaps by calling the `ParameterItem.optsChanged` from within `GroupParameterItem.optsChanged`, so as not to duplicate code.

jaxankey:
(Also, this is great, thanks!)

2xB:
@jaxankey You are right, there is something fundamentally wrong with the update mechanism. I was not aware that there are that many properties that are not handled on update. I will have a look at this tomorrow or the day after, that definitely needs to be fixed.

jaxankey:
Yeah, there are more than I listed, too, e.g., 'tip'.

Note this seems to fix `setOpts` and `restoreState`, but does not fix the inverse problem of someone expanding/collapsing a parameter in the GUI, which currently does not update `opts`.

jaxankey:
(And I can't find a signal that triggers when an expand happens. But I give up for today.)

2xB:
@jaxankey The inverse problem will actually be more difficult to solve: If states are synced with the fundamental `Parameter`, and as we just implemented that change in the `Parameter` signals all associated `ParameterItem`s, effectively the collapsed/expanded state of all `ParameterItem`s is always synchronized. But users should be able to collapse something in one view and leave it expanded in another view.

2xB:
@jaxankey Made three changes, namely:
 * As a default, synchronizing the "expanded" would lead to users being forced to always have multiple `ParameterTree`s to be expanded in the exact same way. To work around that, an option `syncExpanded` is added to customize whether synchronization of the "expanded" state should happen. This reverts the initial idea of this PR to always pass changes in `expanded` to `ParameterItem`s, making the synchronization bidirectional once activated.
 * Reflect changes in `Parameter` options `renamable` and `removable` in UI.
 * Reflect changes in `Parameter` option `tip` in UI.

Additionally, since changing `Parameter` types after making them visible is not supported, I tried generating errors when a `Parameter`s type is changed, but reverted the corresponding commit with a force push, since that generated errors in the example `relativity_demo.py`.

For the other parameters:
* `strictNaming`, `readonly`: From my tests, changes in that `Parameter` option are already reflected in UI.
* `enabled`: From my tests, this option does not seem to do anything, so no change has to be reflected.

j9ac9k:
@2xB I merged another PR a bit ago, which appears to have caused some conflicts with your branch.  Would you mind resolving those, which should trigger another CI run (which I was about to do anyway).

Thanks.

j9ac9k:
@2xB I accidentally merged a PR that broke  the `test_pg_edit` ...I'm working on a PR to skip that test on pyqt 5.9 right now.

2xB:
@j9ac9k Isn't that the `test_pg_exit` test that is failing randomly on macOS Py3.6 PyQt5.9 since quite a while now and that you just managed to replicate locally, as you reported in #1163? I don't think that is new since a specific recent PR ^^

j9ac9k:
That `test_pg_exit` test started failing on a variety of PRs, but not all of them.  I was doing my best last night to merge PRs that were not affected, but either I merged one that was impacted, or the combination of a PRs that I merged, where each passed on their own caused this failure.  I am able to reproduce locally, but I suspect this is a Qt specific issue, not a pyqtgraph one, so decided to just skip this test on Qt 5.9 builds (not just macOS).

Anyway, now that the dev build is failing this test, now your PR is going to show that test failing regardless of your code change (until I merge #1215)

2xB:
Retriggering CI

j9ac9k:
Green across the board 🎊 Merging, thanks for the PR @2xB

PR #1184, opened 2020-04-29
improve SymbolAtlas.getSymbolCoords performance

danielhrisca:
I've done some tests plotting ~900 ScatterPlotItem objects each with ~500 samples.

Profiling the current code I get about 4.98s

image

and profiling the patched code I get 2.51s

image


j9ac9k:
@danielhrisca Do you happen to have some test code handy that I could benchmark this on?  It's not that I don't believe your PR makes the performance improvement that it does; but I would like to poke around it, usually pre-defining a empty numpy array is faster than appending to list.

danielhrisca:
Hello, 

there are some steps to follow:

1. install asammdf with GUI suppport (https://github.com/danielhrisca/asammdf) ``pip install asammdf[gui]``
2. download some demo files (you can use the zip from [here](https://github.com/danielhrisca/asammdf/issues/14#issuecomment-339546372))
3. plot the first channel group 
asammdf

The problem is that iterating over an np.array is slower that converting the array to list and iterating over the list

image

danielhrisca:
Ping @j9ac9k

j9ac9k:
Thanks for keeping me honest.  If I don't get to this this evening, I'll merge tomorrow.

j9ac9k:
Thanks for sharing @danielhrisca while I'm not convinced there isn't a vectorized way of doing that function, figuring out what it is is well outside the scope of this PR.  I'm merge, thanks for the PR and thanks for sharing your application.

danielhrisca:
I would love to see an accelerated version to speed up scatter plots.

Are you working in the automotive domain?

j9ac9k:
No, I work in the speech recognition domain, but before I went into a software career profession, I was a mechanical engineer; the automotive industry is not particularly strong in the parts of the country I wanted to live in so never got to touch that domain (for better or for worse).

PR #1188, opened 2020-05-06
Added support for plot curve to handle both fill and connect args.

millen1m:
Added option to handle both fill and connect options in plot curve to enclose shapes without having to plot them all separately.

Example use and screenshot of output attached.

[plot_example_using_fills_and_connect_py.txt](https://github.com/pyqtgraph/pyqtgraph/files/4584781/plot_example_using_fills_and_connect_py.txt)
Screen Shot 2020-05-06 at 4 09 56 PM

.


j9ac9k:
Thanks for the PR @millen1m , looks good to me, merging!

PR #1198, opened 2020-05-11
improve SymbolAtlas.getSymbolCoords and ScatterPlotItem.plot performance

danielhrisca:
* only use the required fields from the opts
* better handling of recarray views in ScatterPlotItem.paint


j9ac9k:
hey @danielhrisca thanks for the PR, looks like the commits got into a bit of a mess, I'm going to squash and merge.

PR #1199, opened 2020-05-14
Pass non-empty sys.argv to QApplication

chrisjbillington:
This is important to preserve the default behaviour of Qt that
applications may expect, for example on Linux setting the X WM_CLASS
attribute based on sys.argv[0], or setting themes as command line
options to the application.

This is a better fix for #1165 than #1169

### DIFF
--- a/pyqtgraph/Qt.py
+++ b/pyqtgraph/Qt.py
@@ -330,21 +330,19 @@ if m is not None and list(map(int, m.groups())) < versionReq:
 
 
 QAPP = None
-def mkQApp(name="pyqtgraph", qt_args=None):
+def mkQApp(name=None):
     """
     Creates new QApplication or returns current instance if existing.
     
-    ==============  =================================================================================
+    ============== ========================================================
     **Arguments:**
-    name            Application name, passed to Qt
-    qt_args         Array of command line arguments passed to Qt
-    ==============  =================================================================================
+    name           (str) Application name, passed to Qt
+    ============== ========================================================
     """
     global QAPP
     QAPP = QtGui.QApplication.instance()
     if QAPP is None:
-        args = [name]
-        if qt_args is not None:
-            args.extend(qt_args)
-        QAPP = QtGui.QApplication(args)
+        QAPP = QtGui.QApplication(sys.argv or ["pyqtgraph"])
+    if name is not None:
+        QAPP.setApplicationName(name)
     return QAPP


2xB:
Thank you for the notice! In the following, I'll provide some arguments for the solution that was just merged:

`sys.argv` is empty for interactive Python sessions launched via `python`/`python3` (as opposed to `ipython`/`ipython3`). `pg.mkQApp` is used in many examples that are designed to also work in interactive sessions. So in case of such an interactive session, Qt would receive an empty array.

Further, if python scripts are invoked with e.g. `python3 launch.py`, `argv[0]` is `launch.py`, and if you are developing multiple applications the same way, you can end up with windows being unintentinally grouped. Of course this is not a problem if you take the time to write a proper `setup.py` and install your application before you run it, but we are talking about a library also used for rapid prototyping.

Lastly, I have to confess that so far I have always percieved Qt receiving `argv` and `argc` and being able to provide common command line arguments for every Qt application as a conceptually cool idea that does not really provide a use case outside of advanced debugging, for which one probably uses a seperate entry point into ones application anyways.

My idea with the PR was that people could use `pg.mkQApp("AppName", sys.argv)` if they wish to pass `sys.argv` to Qt.

I'd be happy to discuss this though!

chrisjbillington:
Ah, actually I was wrong that Linux DEs will group windows that have the same `WM_CLASS` - they will only do so if the WM_CLASS corresponds to the name of a properly installed launcher. So many `launch.py` scripts won't be grouped unless there is for example a launcher installed at `/usr/share/applications/launch.py.desktop`

Avoiding an empty argv makes perfect sense, I didn't know that was a problem in interactive sessions.

Might I suggest the following?

```python
QAPP = None
def mkQApp():
    """ Creates new QApplication or returns current instance if existing."""
    global QAPP
    QAPP = QtGui.QApplication.instance()
    if QAPP is None:
        QAPP = QtGui.QApplication(sys.argv or ["pyqtgraph"])
    return QAPP

```

Then it's setting it to `"pyqtgraph"` rather than allowing it to be empty.

Since the user may set the application name with `QApplication.setApplicationName()`, I don't think it's worth wrapping setting the name in this function. Also, since this function is called internally by pyqtgraph code, the user can't rely on any arguments they pass to it actually being used - they would have to call it before any other pyqtgraph code to make sure their arguments are used in construction, otherwise they would silently have no effect. I actually wonder if this should be a private function - if the user wants to customise QApplication creation they can simply instantiate one themselves.

If you do want to retain custom args that can be passed to this function though, I'll be happy so long as when it is called with no arguments (as it is throughout pyqtgraph), it passes in `sys.argv`. Otherwise a 'properly made and installed' application will need to make sure it beats pyqtgraph to creating a `QApplication`. Such an application can always call `setApplicationName()` itself to override whatever pyqtgraph does, but relying on `sys.argv[0]` is more universal - the fact that most GUI toolkits automatically set the `WM_CLASS` from it is why so many Linux apps get correct taskbar behaviour 'for free' without their developers necessarily knowing about how to make it work.

Also, I'm currently working on a [project](https://github.com/chrisjbillington/desktop-app) that tries to ensure Python applications behave properly in the desktop environment, regardless of what GUI toolkit the application uses. It does this by ensuring the application is run with `sys.argv[0]` matching the name of the launcher the library creates, since that works across GUI toolkits (being an [old standard](https://books.google.com/books?id=d8tByjvMmIwC&lpg=PA418&ots=mpXLgs49H9&dq=WM_CLASS%20argv%5B0%5D&pg=PA418#v=onepage&q=WM_CLASS%20argv%5B0%5D&f=false)). So whilst a Qt app can always call `setApplicationName()`, it's unfortunate if it has to just to undo something that would have been working correctly if pyqtgraph hadn't modified it.

Finally, if I reordered some function calls in my Qt app such that pyqtgraph beats me (or some other library that needs a QApplication singleton) to the punch instantiating a QApplication, and this modified my app's taskbar behaviour, it would be pretty tricky thing to track down - that's a pretty subtle side effect.

2xB:
I honestly wrote my PR primarely having the use in custom scripts in mind. You have a good point in stating that this might lead to the user providing function parameters that are ignored because a `QApplication` already exists. I think for the users actually having `mkQApp` as a not configurable shortcut for quickly getting a `QApplication` instance whether it exists or not, and the `QApplication` constructor as the customizable variant, is a good split. In that case, we probably want to reflect this in the documentation.

> I actually wonder if this should be a private function - if the user wants to customise QApplication creation they can simply instantiate one themselves.

Well, to me, this has three aspects: Two practical and a "religious" one.
The religious one: Coming from a non-Python background, I also like exposing only a limited set of well curated functions. But making things `private` in Python is actually generally considered bad, see e.g.: https://stackoverflow.com/a/1547163/8575607 .

The first practical: Writing `pg.mkQApp()` in your own scripts is just as short as it gets. If you do rapid prototyping, or just quickly want to interactively visualize something, it is great to have one singleton factory method that does not need any additional import.

The second practical: From Qt4 to Qt5, `QApplication` moved from `QtGui` to `QtWidgets`. `pyqtgraph` currently provides a modded `QtGui` module that contains `QApplication` whatever version of PyQt you use. In many examples, `pyqtgraph` suggests `QtGui.QApplication` to be used. But what happens when `pyqtgraph` drops backwards compatibility with Qt4, which will happen after the next release? `mkQApp` is independent of this. No other module needed, nothing version specific. It just works.

chrisjbillington:
That's all fair enough - it is good for very quick, particularly interactive use. How about the following then? Use args if given, otherwise use sys.argv, and if they're empty use ["pyqtgraph"]. Then, set the name if given. This means the name will be set even if the user calls this function after the singleton was already created.

```python
QAPP = None
def mkQApp(name=None, qt_args=None):
    global QAPP
    QAPP = QtGui.QApplication.instance()
    if QAPP is None:
        QAPP = QtGui.QApplication(qt_args or sys.argv or ["pyqtgraph"])
    if name is not None:
        QAPP.setApplicationName(name)
    return QAPP
```

I agree that setting command line args to Qt is a pretty niche use case, and you can't insert them after the fact - if the singleton was already created they will have no effect. Whereas the name can be changed at any time. So I would probably even go for ditching them and doing:

```python
QAPP = None
def mkQApp(name=None):
    global QAPP
    QAPP = QtGui.QApplication.instance()
    if QAPP is None:
        QAPP = QtGui.QApplication(sys.argv or ["pyqtgraph"])
    if name is not None:
        QAPP.setApplicationName(name)
    return QAPP
```

This checks the boxes of allowing a customisable name, but using sys.argv if called with no arguments, unless that would be empty, then it makes sure to set a sensible default. And it sets the name regardless of whether the QApplication was already existing or not, so the name argument will always have an effect instead of mysteriously sometimes not depending on the order of things.

2xB:
The last suggestion also looks very good to me! Especially since it doesn't potentially ignore any function parameter.

chrisjbillington:
Excellent :). Updated this branch to reflect the latest suggestion.

j9ac9k:
Thanks for the PR @chrisjbillington and thanks for being so descriptive with the potential issues in play.

@2xB I'm good with merging if you are.

2xB:
Me too! Just mentioning that it is always good to terminate files with an empty line, but that's just nit picking.
@chrisjbillington Thank you for the PR and the discussion! Seeing a new face reviewing a PR and even providing commits is great!

j9ac9k:
nice catch, @chrisjbillington mind adding a new line to the end of the file, we'll merge right after that 👍 

thanks again!

chrisjbillington:
Thanks, newline added!

j9ac9k:
I'm still not seeing a new line at the end of the file, the last line of the file is `    return QAPP
` not an empty line...

chrisjbillington:
I believe it's there. I don't think github prints trailing blank lines, instead to indicate the lack of a newline character at the end of a file it appends a little icon saying so.

You can the lack of a newline indicated in the previous version [here](https://github.com/pyqtgraph/pyqtgraph/commit/cc411449e91798ec33056bf94e4dfed6a7bdfc05)

j9ac9k:
oh right!

Thanks for pointing that out, merging.  Thanks again!

PR #1201, opened 2020-05-17
Get readthedocs working

ixjlyons:
Results can be viewed here: https://pyqtgraph.readthedocs.io/en/rtd_build/index.html

The main thing that wasn't working previously was the API documentation because autodoc wasn't able to import the code without the dependencies installed. All that was really needed to fix it was adding the requirements.txt and readthedocs could figure out the rest. I added a config file [as they recommend](https://docs.readthedocs.io/en/stable/config-file/index.html) to be a little more explicit about the configuration and keep it in the repo itself.

Adding `fail_on_warning` may become annoying, but I think it's worth it.

One (possibly separate) issue I haven't quite figured out is how to get the `stable` docs working correctly to point to the latest tagged release. I *think* it's not parsing the `pyqtgraph-x.y.z` tag names the way it would parse `x.y.z`, but I haven't confirmed that. We can always manually activate versions as needed.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -2,5 +2,7 @@ pyside2
 numpy
 scipy
 h5py
+matplotlib
+pyopengl
 sphinx
 sphinx_rtd_theme
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,8 +1,5 @@
 pyside2
 numpy
-scipy
-h5py
-matplotlib
 pyopengl
 sphinx
 sphinx_rtd_theme
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,0 +1,6 @@
+pyside2
+numpy
+scipy
+h5py
+sphinx
+sphinx_rtd_theme
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,12 @@
+# Read the Docs configuration file
+# https://docs.readthedocs.io/en/stable/config-file/v2.html
+
+version: 2
+
+python:
+  version: 3
+  install:
+    - requirements: doc/requirements.txt
+
+sphinx:
+  fail_on_warning: true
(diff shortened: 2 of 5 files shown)


j9ac9k:
I'm 👍 on `fail_on_warning`.  Thanks for doing this @ixjlyons

doc/requirements.txt
```diff
@@ -0,0 +1,8 @@
+pyside2
+numpy
+scipy
+h5py
+matplotlib
```
j9ac9k, line 5: Just out of curiosity why is `matplotlib` needed here?

j9ac9k:
Should we link the readthedocs url from `README.md`?

ixjlyons:
> Should we link the readthedocs url from `README.md`?

That's a good point, I hadn't thought of that. Although now that I do, I think there are still some decisions to be made regarding the website. We could keep the URL (which I'm sure has been used all over the place)  and [set up readthedocs and DNS](https://docs.readthedocs.io/en/stable/custom_domains.html#custom-domain-support) or at least redirect. The alternative I guess would be pyqtgraph.org/documentation for stable and readthedocs for dev.

doc/requirements.txt
ixjlyons: MatplotlibWidget

ixjlyons: We could [mock it](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports)

j9ac9k:
I'm definitely not of the opinion of having stable docs in one place, and dev docs elsewhere.  I'm not sure what the uptime has been like for `pyqtgraph.org` historically, but probably not as good as readthedocs. 

@campagnola do you have an opinion on this?

doc/requirements.txt
j9ac9k: Oh right, forgot about that, I'm good without mocking unless you think we should.

ixjlyons: I went ahead and trimmed it down to what's necessary, if only to speed up builds.

campagnola:
The documentation on pyqtgraph.org is pretty frequently accessed, especially from search engines. My inclination is to move all versions of the documentation to readthedocs and set up a redirect such that all old references will still point to the correct location.

j9ac9k:
@ixjlyons given now that pyqtgraph.org now redirects to readthedocs, we should probably update the link on `README.md`

ixjlyons:
Alright, link in the README updated and I added a badge which should be helpful together with `fail_on_warning`

j9ac9k:
Thanks for getting this working @ixjlyons !

Issue #1189, opened 2020-05-06
Potential issues with GraphicsItem.pixelVectors caching


erikmansson:
I found some potential issues in the caching in GraphicsItem.pixelVectors.

In both the local cache lookup:

https://github.com/pyqtgraph/pyqtgraph/blob/720fa5f3c2e77dd1b5d18af26f65de41188b282e/pyqtgraph/graphicsItems/GraphicsItem.py#L189-L190

and global cache lookup:

https://github.com/pyqtgraph/pyqtgraph/blob/720fa5f3c2e77dd1b5d18af26f65de41188b282e/pyqtgraph/graphicsItems/GraphicsItem.py#L194-L198

the cached value is used if `direction is None` (implicitly `QtCore.QPointF(1, 0)`). However, the cached value may have been calculated for any direction, and the direction does seem not affect the key obtained from `dt`. This may result in returning an incorrect value in when `pixelVectors` has been called with an explicit direction prior to calling it with `direction=None`.

Furthermore, when `pixelVectors` is called with `direction=Point([1, 0])`, as done here:

https://github.com/pyqtgraph/pyqtgraph/blob/720fa5f3c2e77dd1b5d18af26f65de41188b282e/pyqtgraph/graphicsItems/ROI.py#L2266

and here

https://github.com/pyqtgraph/pyqtgraph/blob/720fa5f3c2e77dd1b5d18af26f65de41188b282e/pyqtgraph/graphicsItems/InfiniteLine.py#L306

through here:

https://github.com/pyqtgraph/pyqtgraph/blob/720fa5f3c2e77dd1b5d18af26f65de41188b282e/pyqtgraph/graphicsItems/GraphicsItem.py#L262-L268

No cache lookup is done since direction is not None, but should be since it's set to that value implicitly if direction is None. This slows down updating e.g. InfiniteLines _a lot_, which is an issue for a project I'm currently working on.

I'd be happy to help out fixing this, but I'd like to know what the intended behavior is first. Is including the direction in the cache key and then not checking if `direction is None` a potential good fix/improvement?

erikmansson:
Can anyone have a look at this? I have a potential fix ready at https://github.com/erikmansson/pyqtgraph/commit/4cae76bd29fc63b670aeba3bfb2a68993d8dfc91. I'll create a PR if it looks ok.

j9ac9k:
Hi @erikmansson sorry for the late response and thank you for the detailed issue creation. 

[...]

PR #1212, opened 2020-05-28
Fix PixelVectors cache

erikmansson:
Suggested fix for issue #1189 where I stumbled upon an issue with the PixelVectors cache. See the issue for more details.

### DIFF
--- a/pyqtgraph/graphicsItems/GraphicsItem.py
+++ b/pyqtgraph/graphicsItems/GraphicsItem.py
@@ -188,24 +188,23 @@ class GraphicsItem(object):
         ## (such as when looking at unix timestamps), we can get floating-point errors.
         dt.setMatrix(dt.m11(), dt.m12(), 0, dt.m21(), dt.m22(), 0, 0, 0, 1)
         
+        if direction is None:
+            direction = QtCore.QPointF(1, 0)
+        elif direction.manhattanLength() == 0:
+            raise Exception("Cannot compute pixel length for 0-length vector.")
+
+        key = (dt.m11(), dt.m21(), dt.m12(), dt.m22(), direction.x(), direction.y())
+
         ## check local cache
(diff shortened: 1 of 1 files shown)


j9ac9k:
Hmm, all linux CI runs are failing, but not just on this PR, but on another one as well... wonder if there is some upstream issue.

2xB:
The linux CI tests fail in pyvirtualdisplay, and that was recently updated, i.e. yesterday and the day before yesterday: https://pypi.org/project/PyVirtualDisplay/#history

It causes an issue in `pytest-xvfb`, already reported here:
https://github.com/The-Compiler/pytest-xvfb/issues/22

We could either ignore it for now, awaiting the update in one of the above modules, or pin `pyvirtualdisplay` to version 0.2.5.

### More info:
#### End of traceback:
```
  File "/usr/share/miniconda/envs/test-environment-3.6/lib/python3.6/site-packages/pyvirtualdisplay/abstractdisplay.py", line 313, in stop
    raise XStartError("stop() is called before start().")
pyvirtualdisplay.abstractdisplay.XStartError: stop() is called before start().
```
#### End of internal pytest error stack:
```
INTERNALERROR>   File "/usr/share/miniconda/envs/test-environment-3.7/lib/python3.7/site-packages/pytest_xvfb.py", line 46, in start
INTERNALERROR>     self._virtual_display.cmd.extend(self.args)
INTERNALERROR> AttributeError: 'Display' object has no attribute 'cmd'
```

j9ac9k:
@2xB figured it had to do w/ the virtual display, but hadn't had a chance to do anything more than glance at it... I'll read a bit more when I get the kids to bed and try and figure out a course of action....on one hand we should ignore, on the other hand, I hate red from the CI.

j9ac9k:
Closing/Reopening the PR to trigger a new CI run.

campagnola:
Looks great to me. Thanks @erikmansson !

PR #1214, opened 2020-05-29
Pin PyVirtualDisplay Version

j9ac9k:
@2xB found another issue describing the same problem we're seeing with our CI system where we have failures with xvfb.  A dependency of pytest-xvfb was updated, this commit merely pins the version back.


j9ac9k:
Not to be that guy that merges their own PR....but yeah when CI is broke and the PR fixes it... Merging.

PR #1217, opened 2020-05-30
Get docs version and copyright year dynamically

ixjlyons:
https://pyqtgraph.readthedocs.io/en/rtd/

The version in the sidebar shows the RTD build version - for dev docs it'll show "latest" where it's implicit that it's up to date with github, and for versioned docs it'll show the tag name. This fixes the page title so it shows the pyqtgraph version string when the docs were built.

I considered adding "...and pyqtgraph contributors" to the copyright string, but that doesn't really feel like my decision to make. @campagnola any thoughts?


2xB:
Might it be that now both `version` and `release` are `0.11.0.dev0`, while it should be `0.11.0` and `0.11.0.dev0`? ( see also https://www.sphinx-doc.org/en/master/usage/configuration.html )

ixjlyons:
They are very commonly set to the same value, so that was my main reasoning. Looking into it a bit further, `version` is used in sphinx_rtd_theme for the navbar, which seems to be overridden by readthedocs anyway ([ref](https://github.com/readthedocs/sphinx_rtd_theme/blob/48a9c21b674815c2482e00538a34ee79af46d766/sphinx_rtd_theme/layout.html#L132-L142)). The title comes from `html_title` which [defaults to using `release`](https://www.sphinx-doc.org/en/master/usage/configuration.html?highlight=title#confval-html_title) (docs say revision, but the comment in the conf template says release).

For versioned docs, the tag (shown in the navbar) and release (shown in the title) will match. For dev docs, the navbar will show `latest` and the title will show `x.y.z.dev0`, which seems fine to me. I don't see a strong reason to differentiate, but I could be missing something.

j9ac9k:
LGTM, thanks @ixjlyons for the PR!

PR #1223, opened 2020-06-04
Remove workaround for memory leak in QImage

### LINKED ISSUE #724: PySide QImage workaround should only be used where bug is present
The current workaround would probably cause a segfault if the upstream bug is fixed.
Reference: https://github.com/pyqtgraph/pyqtgraph/commit/f6819dda28e9f725215efb5f198257a43b827d38#commitcomment-29552713

j9ac9k:
Fixes #724 - PySide2 < 5.12 should not be considered stable, and due to its age, pyqtgraph should not try and compensate with it by embedding workarounds.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1261,30 +1261,10 @@ def makeQImage(imgData, alpha=None, copy=True, transpose=True):
         
     if QT_LIB in ['PySide', 'PySide2']:
         ch = ctypes.c_char.from_buffer(imgData, 0)
-        
-        # Bug in PySide + Python 3 causes refcount for image data to be improperly 
-        # incremented, which leads to leaked memory. As a workaround, we manually
-        # reset the reference count after creating the QImage.
-        # See: https://bugreports.qt.io/browse/PYSIDE-140
-        
-        # Get initial reference count (PyObject struct has ob_refcnt as first element)
-        rcount = ctypes.c_long.from_address(id(ch)).value
         img = QtGui.QImage(ch, imgData.shape[1], imgData.shape[0], imgFormat)
-        if sys.version[0] == '3':
-            # Reset refcount only on python 3. Technically this would have no effect
-            # on python 2, but this is a nasty hack, and checking for version here 
-            # helps to mitigate possible unforseen consequences.
-            ctypes.c_long.from_address(id(ch)).value = rcount
     else:
-        #addr = ctypes.addressof(ctypes.c_char.from_buffer(imgData, 0))
         ## PyQt API for QImage changed between 4.9.3 and 4.9.6 (I don't know exactly which version it was)
         ## So we first attempt the 4.9.6 API, then fall back to 4.9.3
-        #addr = ctypes.c_char.from_buffer(imgData, 0)
-        #try:
-            #img = QtGui.QImage(addr, imgData.shape[1], imgData.shape[0], imgFormat)
-        #except TypeError:  
-            #addr = ctypes.addressof(addr)
-            #img = QtGui.QImage(addr, imgData.shape[1], imgData.shape[0], imgFormat)
         try:
             img = QtGui.QImage(imgData.ctypes.data, imgData.shape[1], imgData.shape[0], imgFormat)
         except:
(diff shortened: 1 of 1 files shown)


2xB:
Could the if block also be deleted, leaving only the code from the current `else` block?

j9ac9k:
> Could the if block also be deleted, leaving only the code from the current `else` block?

Suppose only one way to find out... let me make the change and see what the CI says....

j9ac9k:
@2xB looks like some failures are starting to trickle in through the CI, specifically on the PySide2 builds (I'll let it finish out to see if there is more of a pattern besides pyside1/2 fail, pyqt4/5 pass.

pijyoi:
PySide2 doesn't support "QImage(imgData.ctypes.data, ...)"

Both PySide2 and PyQt5 support "QImage(imgData, ...)", i.e. just use the numpy buffer directly.

Don't know about PySide1 and PyQt4 though.

j9ac9k:
> PySide2 doesn't support "QImage(imgData.ctypes.data, ...)"
> 
> Both PySide2 and PyQt5 support "QImage(imgData, ...)", i.e. just use the numpy buffer directly.
> 
> Don't know about PySide1 and PyQt4 though.

Thanks for pointing that out @pijyoi let's try it out!

2xB:
Good to know!
The `imgData.ctypes.data` was introduced in  https://github.com/pyqtgraph/pyqtgraph/commit/adda8ae24d3db24cf933ddd959055afa82c47994 in which Luke writes:
> This fixes memory leaks with PyQt 4.10 _except_ when using makeQImage(copy=False).
Tested on 4.9.3 and 4.10.2; need to be tested against other versions.

Well - It would look even more compact otherwise.

This experimenting shows that the fix for old PyQt4 versions in the `except` clause is Python2 specific though:
```
>               img = QtGui.QImage(buffer(imgData), imgData.shape[1], imgData.shape[0], imgFormat)
E               NameError: name 'buffer' is not defined
```

j9ac9k:
can't kill Py2/Qt4 support soon enough...

j9ac9k:
Looks like this is failing for all PyQt4/5 versions (still CI runs are going).  I think I'll revert back to how it was and leave it at that, or is there another change I should try?

2xB:
Okay, so technically this works, but image comparison tests in PyQt fail though...

j9ac9k:
> Okay, so technically this works, but image comparison tests in PyQt fail though...

Ahh, I should have looked at the nature of the failures... you caught me being lazy  ... you think it's worth updating the images being compared?

j9ac9k:
have to call it a night; I reverted to the state where CI passed across the board.

j9ac9k:
I'm going to merge this as it stands, I suspect after we limit the number of supported libraries we're going to be taking a closer look at all these workarounds for the different bindings.

2xB:
I am sorry for somehow missing your question. I think your plan sounds good. 
Also, I think this shows that we should start to require notes like `# Using imgData.ctypes.data needed to fix memory leak in 4.10 for makeQImage(copy=True)` in the source code for obscure fixes, having this information only in commit messages is obviously impractical.

j9ac9k:
@2xB I think that's a great idea (making better use of comments to describe workarounds).

pijyoi:
>Both PySide2 and PyQt5 support "QImage(imgData, ...)", i.e. just use the numpy buffer directly.

I found some old comments in my code regarding PyQt5:
"If QImage is created with a python object (including ndarray and ctypes object), then it will make a copy of the data when you call setColorTable()"

So looks like PySide and PyQt codepaths can't be safely unified here as there may be some unknown/unwanted side effects.

j9ac9k:
> > Both PySide2 and PyQt5 support "QImage(imgData, ...)", i.e. just use the numpy buffer directly.
> 
> I found some old comments in my code regarding PyQt5:
> "If QImage is created with a python object (including ndarray and ctypes object), then it will make a copy of the data when you call setColorTable()"
> 
> So looks like PySide and PyQt codepaths can't be safely unified here as there may be some unknown/unwanted side effects.

Do you think it would be worth testing again to see if this is an issue with Qt6  (or PyQt5 5.15)?

pijyoi:
>Do you think it would be worth testing again to see if this is an issue with Qt6 (or PyQt5 5.15)?

I retested PyQt5 5.15 while adding PyQt6 support and documented it more thoroughly in the comments.
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L1259-L1279

In short, PyQt5 5.15 needs its own special case.

The copying effect in PyQt5 can be demonstrated in the following code..
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/widgets/RemoteGraphicsView.py#L230-L241
If you modify the PyQt5 code to be the same as PyQt6, no plot will result. i.e. calling (non-const) QImage::fill() triggers the copy-on-write mechanism and further renders are to this new memory rather than to the shared memory.

PR #1225, opened 2020-06-06
ParameterTree: Fix custom context menu

2xB:
This issue was introduced in merging develop into #1175.
While refactoring for the merge, I did not attribute the change in namespace correctly, leading to the parameter `opts` to be assumed in local namespace when it isn't.


2xB:
The last commit adds `opts` as alias for `self.param.opts` in the function `contextMenuEvent`, following the style of the function `updateFlags`.

j9ac9k:
This LGTM, love your simple patch 👍

PR #1229, opened 2020-06-06
Wrap text in tables in docs

ixjlyons:
Before:

image

After:

image

Live example: https://pyqtgraph.readthedocs.io/en/rtd/graphicsItems/plotitem.html#pyqtgraph.PlotItem.setLabel


j9ac9k:
Nice!

PR #1241, opened 2020-06-10
Use older pytest-xvfb for py2 configs

j9ac9k:
This PR handles the case with support py2 with the updated pytest-xvfb.  Thanks to @The-Compiler for maintaining the `pytest-xvfb` package we're dependent on.

### DIFF
--- a/azure-test-template.yml
+++ b/azure-test-template.yml
@@ -135,7 +135,12 @@ jobs:
       then
         source activate test-environment-$(python.version)
       fi
-      pip install PyVirtualDisplay==0.2.5 pytest-xvfb
+      if [ $(python.version) == "2.7" ]
+      then
+        pip install PyVirtualDisplay==0.2.5 pytest-xvfb==1.2.0
+      else
+        pip install pytest-xvfb
+      fi
     displayName: "Virtual Display Setup"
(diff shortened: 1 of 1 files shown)


The-Compiler:
FWIW if you have a new enough pip (pip 9 or newer), I'd expect `pip install pytest-xvfb` to pick up an older version automatically on Python 2, because it sets the appropriate [`python_requires` value](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires).

j9ac9k:
> FWIW if you have a new enough pip (pip 9 or newer), I'd expect `pip install pytest-xvfb` to pick up an older version automatically on Python 2, because it sets the appropriate [`python_requires` value](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires).

I think the problem in this case wasn't the correct selection of `pytest-xvfb` but the fact that your current release requires a recent version of PyVirtualDisplay; and we had pinned `PyVirtualDisplay` to 0.2.5 (as I think previous versions of `pytest-xvfb` installed the most recent version of PyVirtualDisplay.

Your library has saved us a ton of head aches when it comes to our CI testing; thanks for maintaining it!

PR #1251, opened 2020-06-12
DateAxisItem: Parameterize utcOffset during construction

dearkarm:
Allows for control over timezone offset in a simple, optional manner


j9ac9k:
Thanks for the PR @dearkarm LGTM, merging

PR #1257, opened 2020-06-15
AxisItem: Make painter (tick) font dependent

VesnaT:
##### Issue
When setting a tick font manually, texts may collide since not considering the font.

##### Description of changes
Set painter's font before generating ticks specifications.


j9ac9k:
LGTM, thanks for the PR @VesnaT 

Merging.

j9ac9k:
Also thank you for adding the test code!

PR #1265, opened 2020-06-19
Make the documentation reproducible

lamby:
Whilst working on the [Reproducible Builds](https://reproducible-builds.org/) effort we noticed that pyqtgraph could not be built reproducibly. This is because it generates copyright years from the current build date and therefore will vary on when you build it.

This commit uses [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/specs/source-date-epoch/) for the "current" build date.

This was originally filed in Debian as [#963124](https://bugs.debian.org/963124).

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -11,6 +11,7 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
+import time
 import sys
 import os
 from datetime import datetime
@@ -46,7 +47,10 @@ master_doc = 'index'
 
 # General information about the project.
 project = 'pyqtgraph'
-copyright = '2011 - {}, Luke Campagnola'.format(datetime.now().year)
+now = datetime.utcfromtimestamp(
+    int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
+)
+copyright = '2011 - {}, Luke Campagnola'.format(now.year)
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the


j9ac9k:
Hi @lamby thanks for the PR, this looks good to me, but before merging, I'm going to ask @ixjlyons to take a look at it as he's been the person that's got our documentation system up and running.

doc/source/conf.py
```diff
@@ -46,7 +47,10 @@
 
 # General information about the project.
 project = 'pyqtgraph'
-copyright = '2011 - {}, Luke Campagnola'.format(datetime.now().year)
+now = datetime.utcfromtimestamp(
```
ixjlyons, line 50: I noticed a note in the [`datetime.utcfromtimestamp`](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp) docs that `datetime.fromtimestamp(timestamp, timezone.utc)` may be preferred to `utcfromtimestamp`. It seems like the `year` attribute would be safe from the issues they're warning against, so I'd defer to Debian folks on this -- if this is the usual approach for python projects, I'm fine with it. 

Also, it might be good to have a brief comment above this line indicating that getting the date in this way is for reproducible builds. Just a link to [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/) would probably suffice. I see it's already in the commit message, but I think having it here in the code wouldn't hurt.

j9ac9k:
@ksunden would be the person to talk to about releases.  I'm good w/ doing another release, a decent number of PRs have been merged since 0.11.0, and one segfault has been fixed.

doc/source/conf.py
lamby: This is how I've been fixing it for a long time now with no issue, if that helps. :+1:

ksunden:
Happy to do a release whenever we want

j9ac9k:
Merging! Thanks for the PR @lamby

PR #1268, opened 2020-06-20
Check for container before setting dock orientation

### LINKED ISSUE #833: if self.container().type() == 'tab': AttributeError: 'NoneType' object has no attribute 'type'
Not sure why I was seeing this or why no one else had. But I seem to have solved the problem by just adding a check that container is not None on line 155 of Dock.py. 

            if self.container() is not None and self.container().type() == 'tab':

ixjlyons:
Fixes #833.

This is an alternative to #900. Yet another alternative would be `if self.container() is not None and self.container().type() == 'tab':`, but I don't see a reason to do any of the stuff in `setOrientation` until the container is set. Once the container *is* set, `setOrientation` is called thereafter anyway.

I played around with `examples/dockarea.py` for a while as well as the example in #833 and I haven't found any problems.

### DIFF
--- a/pyqtgraph/dockarea/Dock.py
+++ b/pyqtgraph/dockarea/Dock.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from ..Qt import QtCore, QtGui
 
 from .DockDrop import *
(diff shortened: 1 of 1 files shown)


j9ac9k:
LGTM, thanks for the PR @ixjlyons

PR #1270, opened 2020-06-23
Implement pr 403

### LINKED ISSUE #449: Axis values
After zooming in/out a lot, numbers of axis values are so big that axis is moved. Calling auto scale after that do not restart axis position
In attachment there are  images that show use cases.

moved axis

after autoscaling

### LINKED ISSUE #821: AxisItem: text space is not reduced when tick labels get smaller
By default, the text space needed for tick labels is automatically increased upon zoom when the labels get longer (e.g. when showing more digits). When zooming back to normal size, though, the space reserved for tick labels stays large. This can result in a huge gap between the axis label and the tick labels.

The issue occurs because the `textWidth` and `textHeight` properties of AxisItem can never decrease.
To see this, let's have a look at the method `_updateMaxTextSize`:
https://github.com/pyqtgraph/pyqtgraph/blob/2e69b9c5e6bd8008c2d7e9f384b9c9b32ff2f5a7/pyqtgraph/graphicsItems/AxisItem.py#L290-L307
With line 295, `mx = max(self.textWidth, x)`, and line 297, `self.textWidth = mx`, we observe that `self.textWidth` is set to the maximum of itself and `x`. So `self.textWidth` can only increase.

In fact, the code should be modified as follows:
Line 295 should be deleted (as the maximum is not needed here).
Line 296 should be replaced by `if x > self.textWidth or x < self.textWidth - 10:`.
Line 297 should be replaced by `self.textWidth = x`.

The same modifications should be done in line 302, 303 and 304 for `self.textHeight`.

### LINKED ISSUE #1053: Axis takes space required but does not yield it
Hey Guys,

i am currently looking to create a plot that shows sizes from 1Å to several µm. so 1E-10 to 10E-6 maybe. Obviously I chose a logarithmic axis to do so. On this axis, I don't want to change the unit away from nm and I also don't want scientific notation for my numbers. So 1µm should not show up as 1E-6 or other scaled factors but as 1000 on my nanometer scale. And yes I am aware this will lead to number like 100000 on my graph. Thats fine and wanted!

Down below is my solution which works, but has a the issue mentioned below. Although not strictly related to the issue itself: is my approach the right one? Is there an easier, better or more stable way to do this?

I tried confining the y-axis by setting the viewbox limits but that caused all sorts of weird bugs. Viewbox limits and log axes don't seem to work together very well. But that for another issue.

Onto the issue itself:

### Short description
If tick labels become large, the axis taxes more and more space to show them. To do so the axis grows. But once labels get smaller again, the axis does not "shrink" again so the plot does not increase in size again.

To see this: use the example and zoom the y-axis (I locked the x-axis for convenience). As numbers get bigger the axis grows. If you zoom back in and numbers get smaller, the axis does not get smaller. In extreme cases this leads to the graph no longer visible at all. I couldn't find a way to recover from this.

### Code to reproduce
```python
import sys
from PyQt5 import QtWidgets
import pyqtgraph as pg
import numpy as np


    def __init__(self):

        self.resize(800, 600)

        self.setCentralWidget(glw)

        x_vals = [0, 1, 2, 3,]
        y_vals1 = [1e-10, 6e-9, 10e-9, 3e-6]

        plot1 = pg.PlotItem(axisItems={'left': NanoMeterScaledLogAxis(orientation='left')})
        plot1.setLogMode(False, True)
        plot1.setMouseEnabled(x=False)
        left_axis = plot1.getAxis("left")
        left_axis.setLabel("Radius [nm]")
        left_axis.enableAutoSIPrefix(False)

        red_line = plot1.plot(x=x_vals, y=y_vals1, pen="r")

        glw.addItem(plot1, 0, 0)


class NanoMeterScaledLogAxis(pg.AxisItem):
        super(NanoMeterScaledLogAxis, self).__init__(*args, **kwargs)

    def tickStrings(self, values, scale, spacing):
        return ["{:.2f}".format(np.power(10, value)*1e9) for value in values]


if __name__ == '__main__':
    ex = Example()
    ex.show()
    sys.exit(app.exec_())
```

### Expected behavior
In all scenarios, independent of the the zoom factor the y-axis should only take the space necessary.

### Real behavior
Y-Axis takes space from the graph, but does not yield it once it is no longer required.

### Tested environment(s)

* PyQtGraph version: Current development branch
* Qt Python binding: PyQt 5.12.3
* Python version: 3.7.4
* NumPy version: 1.16.3
* Operating system: Win10
* Installation method: pip

### LINKED ISSUE #1183: Graph permanently shifts to the right when having long tick strings
### Short description
A plot with long tick names is squashed horizontally to make room for the tick names. That is good. However, even after changing the plot and the tick names, the plot stays squashed. Is there some way to prevent that from happening?

### Code to reproduce

```python
import pyqtgraph as pg

plot_widget = pg.plot(title="test")
plot_widget.getPlotItem().getAxis("left").setTicks([[(0, "a loooooooooooooooooooooooooooooong tick string")]])
input("Press enter to change the tick string to a short string")
plot_widget.getPlotItem().getAxis("left").setTicks([[(0, "short")]])
input("Press enter to quit")
```

### Expected behavior
Optimally I expect the plot to automatically adjust to tick strings whenever they are changed.

### Real behavior
The plot gets smaller when adding long tick strings, but it won't expand back again when having short tick strings.

### Tested environment(s)
 * PyQtGraph version: 0.11.0.dev0+ga76d9da
 * Qt Python binding: PyQt5 5.14.2 Qt 5.14.2
 * Python version: 3.8.2
 * NumPy version: 1.17.4
 * Operating system: Linux x64, KDE
 * Installation method: pip

### Additional context
For now I will circumvent this problem be completely replacing the entire PlotWidget on every graph update.

j9ac9k:
This PR implements PR 403, with the suggested change by @SamSchott here: https://github.com/pyqtgraph/pyqtgraph/pull/959#issuecomment-507217802

I would have updated PR 403, but github is saying it's come from an `unknown repository` so I cannot update the PR accordingly.

Fixes #403 
Fixes #235
Fixes #1053 
Fixes #1183 
Fixes #959 
Fixes #821 
Fixes #449

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -56,6 +56,7 @@ class AxisItem(GraphicsWidget):
             'tickTextWidth': 30,  ## space reserved for tick text
             'tickTextHeight': 18,
             'autoExpandTextSpace': True,  ## automatically expand text space if needed
+            'autoReduceTextSpace': True,
             'tickFont': None,
             'stopAxisAtTick': (False, False),  ## whether axis is drawn to edge of box or to last tick
             'textFillLimits': [  ## how much of the axis to fill up with tick text, maximally.
@@ -125,6 +126,7 @@ class AxisItem(GraphicsWidget):
         tickTextHeight      (int) Vertical space reserved for tick text in px
         autoExpandTextSpace (bool) Automatically expand text space if the tick
                             strings become too long.
+        autoReduceTextSpace (bool) Automatically shrink the axis if necessary 
         tickFont            (QFont or None) Determines the font used for tick
                             values. Use None for the default font.
         stopAxisAtTick      (tuple: (bool min, bool max)) If True, the axis
@@ -314,19 +316,26 @@ class AxisItem(GraphicsWidget):
         ## changed; we use this to decide whether the item needs to be resized
         ## to accomodate.
         if self.orientation in ['left', 'right']:
-            mx = max(self.textWidth, x)
-            if mx > self.textWidth or mx < self.textWidth-10:
-                self.textWidth = mx
-                if self.style['autoExpandTextSpace'] is True:
-                    self._updateWidth()
-                    #return True  ## size has changed
+            if self.style["autoReduceTextSpace"]:
+                if x > self.textWidth or x < self.textWidth - 10:
+                    self.textWidth = x
+            else:
+                mx = max(self.textWidth, x)
+                if mx > self.textWidth or mx < self.textWidth - 10:
+                    self.textWidth = mx
+            if self.style['autoExpandTextSpace']:
+                self._updateWidth()
+        
         else:
-            mx = max(self.textHeight, x)
-            if mx > self.textHeight or mx < self.textHeight-10:
-                self.textHeight = mx
-                if self.style['autoExpandTextSpace'] is True:
-                    self._updateHeight()
-                    #return True  ## size has changed
+            if self.style['autoReduceTextSpace']:
+                if x > self.textHeight or x < self.textHeight - 10:
+                    self.textHeight = x
+            else:
+                mx = max(self.textHeight, x)
+                if mx > self.textHeight or mx < self.textHeight - 10:
+                    self.textHeight = mx
+            if self.style['autoExpandTextSpace']:
+                self._updateHeight()
 
     def _adjustSize(self):
         if self.orientation in ['left', 'right']:
@@ -1001,6 +1010,7 @@ class AxisItem(GraphicsWidget):
             #textHeight = self.style['tickTextHeight'] ## space allocated for horizontal text
 
         textSize2 = 0
+        lastTextSize2 = 0
         textRects = []
         textSpecs = []  ## list of draw
 
@@ -1062,6 +1072,8 @@ class AxisItem(GraphicsWidget):
                         break
                 if finished:
                     break
+            
+            lastTextSize2 = textSize2
 
             #spacing, values = tickLevels[best]
             #strings = self.tickStrings(values, self.scale, spacing)
@@ -1097,7 +1109,7 @@ class AxisItem(GraphicsWidget):
         profiler('compute text')
 
         ## update max text size if needed.
-        self._updateMaxTextSize(textSize2)
+        self._updateMaxTextSize(lastTextSize2)
 
         return (axisSpec, tickSpecs, textSpecs)
 
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -359,7 +359,7 @@ class AxisItem(GraphicsWidget):
             if self.fixedHeight is None:
                 if not self.style['showValues']:
                     h = 0
-                elif self.style['autoExpandTextSpace'] is True:
+                elif self.style['autoExpandTextSpace']:
                     h = self.textHeight
                 else:
                     h = self.style['tickTextHeight']
@@ -390,7 +390,7 @@ class AxisItem(GraphicsWidget):
             if self.fixedWidth is None:
                 if not self.style['showValues']:
                     w = 0
-                elif self.style['autoExpandTextSpace'] is True:
+                elif self.style['autoExpandTextSpace']:
                     w = self.textWidth
                 else:
                     w = self.style['tickTextWidth']
--- a/pyqtgraph/graphicsItems/tests/test_ErrorBarItem.py
+++ b/pyqtgraph/graphicsItems/tests/test_ErrorBarItem.py
@@ -21,19 +21,19 @@ def test_ErrorBarItem_defer_data():
     app.processEvents()
     r_empty_ebi = plot.viewRect()
 
-    assert r_no_ebi == r_empty_ebi
+    assert r_no_ebi.height() == r_empty_ebi.height()
 
     err.setData(x=x, y=x, bottom=x, top=x)
     app.processEvents()
     r_ebi = plot.viewRect()
 
-    assert r_empty_ebi != r_ebi
+    assert r_ebi.height() > r_empty_ebi.height()
 
     # unset data, ErrorBarItem disappears and view rect goes back to original
     err.setData(x=None, y=None)
     app.processEvents()
     r_clear_ebi = plot.viewRect()
 
-    assert r_clear_ebi == r_no_ebi
+    assert r_clear_ebi.height() == r_empty_ebi.height()
 
     plot.close()


j9ac9k:
and I inherited some commits that haven't been merged upstream yet, whoops...

j9ac9k:
@ixjlyons I know we've had this discussion in few places, including #959 I'm trying to look into this a bit more, sorry to be a broken record about this.

I took screenshots of the plots to see if I can visually spot the difference

This is at the beginning of the text
Screen Shot 2020-06-23 at 9 30 57 PM

This is at the end of the test (right before the assert failure)
image

Visually, the only thing I can notice is the gap between the numbers of the vertical axis, and the edge of the window.  In the first image, you can see the leading `1` is _right_ up against the edge of the window, but at the end, there is a bit more of a gap between the leading `1` and the vertical axis.

I took a screenshot of the two images, yellow line shows the left edge is lined up, right line shows the vertical axis is thicker on the second image.

Screen Shot 2020-06-23 at 9 37 01 PM

I'll keep poking at this some more.

j9ac9k:
ok, this is getting weird now, I'm inserting breakpoints's throughout `test_ErrorBarItem` and just hitting continue seems to be sufficient to make the test pass

```
❯ pytest -k "test_ErrorBarItem"
========================================================= test session starts =========================================================
platform darwin -- Python 3.8.3, pytest-5.4.3, py-1.8.2, pluggy-0.13.1
rootdir: /Users/ogi/Developer/pyqtgraph, inifile: pytest.ini
plugins: cov-2.10.0
collected 138 items / 137 deselected / 1 selected

pyqtgraph/graphicsItems/tests/test_ErrorBarItem.py
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/ogi/Developer/pyqtgraph/pyqtgraph/graphicsItems/tests/test_ErrorBarItem.py(22)test_ErrorBarItem_defer_data()
-> err = pg.ErrorBarItem()
(Pdb) c

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB continue (IO-capturing resumed) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/ogi/Developer/pyqtgraph/pyqtgraph/graphicsItems/tests/test_ErrorBarItem.py(30)test_ErrorBarItem_defer_data()
-> err.setData(x=x, y=x, bottom=x, top=x)
(Pdb) c

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB continue (IO-capturing resumed) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/ogi/Developer/pyqtgraph/pyqtgraph/graphicsItems/tests/test_ErrorBarItem.py(37)test_ErrorBarItem_defer_data()
-> err.setData(x=None, y=None)
(Pdb) c

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB continue (IO-capturing resumed) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/ogi/Developer/pyqtgraph/pyqtgraph/graphicsItems/tests/test_ErrorBarItem.py(41)test_ErrorBarItem_defer_data()
-> assert r_clear_ebi == r_no_ebi
(Pdb) c

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB continue (IO-capturing resumed) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
.                                                                            [100%]

========================================================== warnings summary ===========================================================
pyqtgraph/graphicsItems/tests/test_ErrorBarItem.py: 62 tests with warnings
  /Users/ogi/Developer/pyqtgraph/pyqtgraph/widgets/GraphicsView.py:373: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
    delta = Point(ev.pos() - QtCore.QPoint(*self.lastMousePos))

-- Docs: https://docs.pytest.org/en/latest/warnings.html
=========================================== 1 passed, 137 deselected, 62 warnings in 5.57s ============================================

~/Developer/pyqtgraph implement-pr-403*
```

j9ac9k:
Got it to pass!  I implemented `autoReduceTextSpace`, and then set it to false for `test_ErrorBarItem.py`

I'm feeling good about this solution, @ixjlyons you have any input?

ixjlyons:
That seems ok to me, though I can't shake the feeling the test itself is just flaky. Is it even reasonable to assume everything settles correctly with one call to `app.processEvents()`? My understanding is that's [not generally the case](https://doc.qt.io/qt-5/qcoreapplication.html#processEvents):

> events that are posted while the function runs will be queued until a later round of event processing

My guess at this point is that this feature is effectively adding another event to the queue to readjust things after the window is initially displayed, so the first rect (`r_no_ebi`) doesn't have the adjustment and the second one (`r_empty_ebi`) does. Does that sound about right? I think that aligns with your debug session observations.

In any case, the new `autoReduceTextSpace` kwarg seems ok to me. I'm good with this if you add a comment to the test why it's disabled. Would replacing it with a second `app.processEvents` like in #959 be a more obvious/general band-aid fix? It'd be cool to eventually get pytest-qt going to use waitUntil, thought to be honest that might not even be that helpful in this particular case (what would we be waiting for? Could just use it for the timeout to allow the event loop to run for a bit I suppose).

Also as a side note, I believe this PR will fix #1053 and #1183 as well - maybe you could check the code to reproduce?

j9ac9k:
I'm definitely 👍 for the comment; I think I'll also set the same parameter for the bottom axis (for consistently), I greatly prefer this method to the 2x call of `app.processEvents`.  The test is likely flakey, but I'm not sure what a better test would be, and given the poor state of our coverage, I lean on preserving tests that may be flakey than outright doing away w/ them.

ixjlyons:
Sounds good. The point of the test was just to make sure you can construct an `ErrorBarItem` without data and calling `setData` at a later time works. The test really doesn't need to be strict about the viewbox range, just that it changes in a predictable way (i.e. translates, grows, shrinks, etc.). Just the fact that you can now construct an `ErrorBarItem` without data and add it to a plot without an exception forms the bulk of the test anyway.

j9ac9k:
> Sounds good. The point of the test was just to make sure you can construct an `ErrorBarItem` without data and calling `setData` at a later time works. The test really doesn't need to be strict about the viewbox range, just that it changes in a predictable way (i.e. translates, grows, shrinks, etc.). Just the fact that you can now construct an `ErrorBarItem` without data and add it to a plot without an exception forms the bulk of the test anyway.

You think we should scrap the asserts?

ixjlyons:
I think there's value in asserting *something* about the viewbox ranges as an approach to verifying the error bars are present/absent, but I don't really feel strongly about it. It wouldn't surprise me if there's already some test function for rect equivalence with tolerance, or the test could just check that the rect height increases after setting data and decreases after clearing it.

j9ac9k:
I thought rect height stayed the same throughout this test, do you mean scale of the vertical axis?

ixjlyons:
I believe `vb.viewRect()` represents the plot range but I could actually be mistaken. The intention is the y range in plot data space.

j9ac9k:
> I believe `vb.viewRect()` represents the plot range but I could actually be mistaken. The intention is the y range in plot data space.

Ahh right!  I'll modify this test accordingly as part of this PR.

PR #1272, opened 2020-06-23
LegendItem: fix clear() not closing widgets

maxkl:
`LegendItem.clear()` did previously not call `.close()` on the child widgets after removing them from the layout. This caused them to still be rendered even after calling `.clear()`. It does the same as `LegendItem.removeItem()` now.

The issue I had was that after clearing all items and adding new ones, the old items were still visible underneath the newly added ones. My changes fixed this issue. Because `LegendItem.removeItem()` also calls `.close()`, I assume this is the correct way to do it.

I didn't add or modify and unit tests because I don't know how this could be tested. If somebody knows a way, let me know and I will implement it.

### DIFF
--- a/pyqtgraph/graphicsItems/LegendItem.py
+++ b/pyqtgraph/graphicsItems/LegendItem.py
@@ -189,7 +189,9 @@ class LegendItem(GraphicsWidget, GraphicsWidgetAnchor):
         """Remove all items from the legend."""
         for sample, label in self.items:
             self.layout.removeItem(sample)
+            sample.close()
             self.layout.removeItem(label)
+            label.close()
 
         self.items = []
         self.updateSize()


j9ac9k:
Changing target, as we're having PRs aim at `master` now... also the current location of where `develop` has a broken pipeline for Py3/Linux configs.

maxkl:
Ah ok, I'm sorry, I was following CONTRIBUTING.md which still says that pull requests should be created against `develop`.

Do I need to incorporate the changes that happened on `master`? I don't have a lot of experience with contributing on GitHub, so I'm not entirely sure what's still needed (if anything).

j9ac9k:
You're fine, those instructions are a little out of date, I'll update them.  Thanks for pointing that out!

j9ac9k:
Weird failure on just a single pipeline; rerunning as a sanity check.

j9ac9k:
now an openGL error on the debug stage....what is happening....

j9ac9k:
Thanks for the PR @maxkl this LGTM, merging.

PR #1273, opened 2020-06-23
Implemented pColorMeshItem

### LINKED ISSUE #146: 2D colormesh plot
Good evening! Cannot you please support a 2D colormesh plot? For example, if I want to plot something in polar coordinates, I have to somehow prepare an image data first. 

In matplotlib it is made really simply:

```
import matplotlib
import numpy as np

r_array = np.linspace(0, 2, 1000)
phi_array = np.linspace(0, 2 * np.pi, 1000)

r_grid, phi_grid, = np.meshgrid(r_array, phi_array)


z_grid = r_grid + phi_grid
x_grid = r_grid * np.cos(phi_grid)
y_grid = r_grid * np.sin(phi_grid)

plt.pcolormesh(x_grid, y_grid, z_grid)
plt.show()
```

Cannot you do something like that?

### LINKED ISSUE #1262: Proposition of equivalent item to pcolormesh from matplotlib
### Short description
Hi all,

This is not an issue but I didn't know where to post it...
I was working on an item equivalent to the [pcolormesh ](https://matplotlib.org/3.2.1/api/_as_gen/matplotlib.pyplot.pcolormesh.html) function from matplotlib.
In a sense this is extremely similar to the imageitem with the crucial difference that the tiles of the image don't have to be rectangles but can be polygons.
This is extremely handy in my work where such cases occur quite often.

I came up with the following class that I am giving as an executable file containing an example.
This is heavily inspired by the "customGraphicsItem.py" examples.

```python
# This Python file uses the following encoding: utf-8
import pyqtgraph as pg
import numpy as np
from matplotlib.colors import Normalize




class pColorMesh(pg.GraphicsObject):


    def __init__(self, x, y, z, cmap=None):
        
        self.x = x
        self.y = y
        self.z = z

        if cmap is None:
            self.cmap = plt.cm.viridis

        self.generatePicture()
    
    def generatePicture(self):
        ## pre-computing a QPicture object allows paint() to run much more quickly, 
        ## rather than re-drawing the shapes every time.
        self.picture = QtGui.QPicture()
        p = QtGui.QPainter(self.picture)
        p.setPen(pg.mkPen('w'))
        
        xfn = self.z.shape[0]
        yfn = self.z.shape[1]

        norm = Normalize(vmin=self.z.min(), vmax=self.z.max())
        for xi in range(xfn):
            for yi in range(yfn):
                
                p.drawConvexPolygon(QtCore.QPointF(self.x[xi][yi],     self.y[xi][yi]),
                                    QtCore.QPointF(self.x[xi+1][yi],   self.y[xi+1][yi]),
                                    QtCore.QPointF(self.x[xi+1][yi+1], self.y[xi+1][yi+1]),
                                    QtCore.QPointF(self.x[xi][yi+1],   self.y[xi][yi+1]))

                c = self.cmap(norm(self.z[xi][yi]))[:-1]
                p.setBrush(QtGui.QColor(c[0]*255, c[1]*255, c[2]*255))
        p.end()
    
    def paint(self, p, *args):
        p.drawPicture(0, 0, self.picture)
    
    def boundingRect(self):
        ## boundingRect _must_ indicate the entire area that will be drawn on
        ## or else we will get artifacts and possibly crashing.
        ## (in this case, QPicture does all the work of computing the bouning rect for us)


# The example

x = np.array([[1,1,1,1],
              [2,2,2,2],
              [3,3,3,3],
              [4,4,4,4],
              [5,5,5,5]])
y = np.array([[4,8,12,16],
              [2,4,6,8],
              [3,6,9,12],
              [5,10,15,20],
              [6,12,18,24]])
z = np.array([[1,2,3],
              [5,6,7],
              [9,10,11],
              [13,14,15]])

app = QtGui.QApplication([])
mw = QtGui.QMainWindow()
view = pg.GraphicsLayoutWidget()
mw.setCentralWidget(view)
mw.show()
mw.setWindowTitle('pyqtgraph example: pColorMesh')

v = view.addPlot()


item = pColorMesh(x, y, z)
v.addItem(item)

if __name__ == '__main__':
    import sys
```

This file produces the following plot:
a

While this class is working I was wondering if some people would be interested to see it integrated in pyqtgraph, I know I would!
Also while this is working for a reasonable number of points, it fails to keep up with the original pcolormesh and lag for large number of polygons >10,000.

### Tested environment(s)

 * PyQtGraph version:  0.10.0
 * Python version: 3.7
 * NumPy version: 1.17.4
 * Operating system: windows 10
 * Installation method: conda

edumur:
As discussed in the issue #1262 I proposed here a minimal implementation of an item mimicking the `pcolormesh` function of [matplotlib](https://matplotlib.org/3.2.1/api/_as_gen/matplotlib.pyplot.pcolormesh.html).
As I didn't know what to do to make the pull request valid I decided to make it as minimal as possible.
I am willing to integrate it more into pyqtgraph if someone gives me some directions.
In particular a I think a kind of pColorMeshWidget with axis and Histogram would be an excellent end result of this work.

Fixes #146
Fixes #1262


j9ac9k:
Updating the base to merge to master; I need to update `CONTRIBUTING.md`

j9ac9k:
Ooof...this CI is blowing up due to matplotlib not being installed... looking at the diff, matplotlib is wanted for the `Normalize` function, and for the viridis colormap.

An approximation of viridis is available in `pygraph/graphicsItems/GradientEditorItem.py` but I've been wanting to move the colormaps from matplotlib into this library in a way other libraries such as bokeh do.  You can see how they make the color maps available here:

https://github.com/bokeh/bokeh/blob/7cc500601cdb688c4b6b2153704097f3345dd91c/bokeh/palettes.py

I understand that this is outside the scope of this PR, but having matplotlib as a test dependency is rough, especially for just those methods.

ixjlyons:
Ideally we'd have test coverage of the matplotlib exporter/widget, so it could be required for tests eventually anyway. That said, I agree it would be better to avoid requiring it for users for the color map.

Also the class name might be more consistent with the library as PseudoColorMeshItem or PColorMeshItem (latter might be more intuitive/recognizable).

2xB:
For this PR, I did not have time for a reasonable review, just from first glance I noticed that the naming `pColorMeshItem` - especially with the lowercase prefix - does not fit the naming scheme used for `pyqtgraph`, which should probably be changed.

For matplotlib colormaps, I've found this repository by @honkomonk some time ago: https://github.com/honkomonk/pyqtgraph_sandbox It might not be the ideal way though, but maybe a good start if someone wants to look into that.

2xB:
Also, if the last lines of the first comment are the following two, then GitHub automatically links the merge of this PR to closing the according issues:

Fixes  #146
Fixes #1262

honkomonk:
_Regarding the colormaps (so a bit off-topic):_
**tl;dr**: In maptplotlib Viridis is just implemented as a list of 256 pre-computed colors, while in pyqtgraph's Gradient it's defined with only 5 colors. The easiest way for a more accurate Viridis colormap would be to just add more reference points.

_The longer answer:_
Matplotlib offers multiple methods for defining colormaps (1. equally spaces color values, like Viridis; 2. color values with positional argument, like in pyqtgraph; 3. functionally generated maps, like Cubehelix). In the code that @2xB cited I tried to convert all of them to pyqtgraph gradients. That's why it got so messy and I never got to clean it up for a PR.
I could image to integrate it into the ColorMap class as an importer for matplotlib cmaps.

Bokeh just copied matplotlibs colormaps and converted them into their format. So no magic is happening there.

PS: It's an interesting story how Viridis was created:  https://bids.github.io/colormap/

edumur:
Hi all,

I removed all dependencies to matplotlib so that we can focus on the Item itself.
I also changed the name as suggested.

j9ac9k:
all pipelines failed with:

```
  File "<string>", line 1, in <module>
  File "/home/vsts/work/1/s/pyqtgraph/__init__.py", line 221, in <module>
    from .graphicsItems.PColorMeshItem import * 
ModuleNotFoundError: No module named 'pyqtgraph.graphicsItems.PColorMeshItem'
display information
  File "/opt/hostedtoolcache/Python/3.8.3/x64/lib/python3.8/runpy.py", line 185, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/opt/hostedtoolcache/Python/3.8.3/x64/lib/python3.8/runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "/home/vsts/work/1/s/pyqtgraph/__init__.py", line 221, in <module>
    from .graphicsItems.PColorMeshItem import * 
ModuleNotFoundError: No module named 'pyqtgraph.graphicsItems.PColorMeshItem'
openGL information
  File "<string>", line 1, in <module>
  File "/home/vsts/work/1/s/pyqtgraph/__init__.py", line 221, in <module>
    from .graphicsItems.PColorMeshItem import * 
ModuleNotFoundError: No module named 'pyqtgraph.graphicsItems.PColorMeshItem'
```

edumur:
This is weird, I renamed all the file with an uppercase "P":
image

And the example file works on my laptop and give this:
image

Is there a way for me to test these "pipeline" online before pushing anything?

j9ac9k:
you could try running `pytest`, which will run through the test suite and see what happens... (fair warning, if you have a hidpi display on windows, you'll see some failures that are unrelated).  Since Windows is not case sensitive, you likely have some shenanigans at play.

I would recommend renaming the file to lower case in windows, then run

`git mv pyqtgraph/graphicsItems/pColorMesh.py pyqtgraph/graphicsItems/PColorMesh.py`

edumur:
Ok, will try the pytest thing.
I followed your advice and I am going to bed now.
See you ;) .

j9ac9k:
FYI that looks to have done the trick, some pipelines have passed (some are still running).

2xB:
Just out of curiosity: Is the `P` there at the beginning out of a reason? I would imagine just `ColorMeshItem` to be simpler and more intuitive, so there is probably a reason against it?

edumur:
> Just out of curiosity: Is the `P` there at the beginning out of a reason? I would imagine just `ColorMeshItem` to be simpler and more intuitive, so there is probably a reason against it?

The idea was to copy the matplotlib function `pcolormesh`, see [doc](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.pcolormesh.html).
The "p" stands for pseudocolor and I don't really know it's meaning, in the doubt I would leave it.

@j9ac9k What will be the next step for me? I would like to have a final item with axis and histogram but would like your advice on the road to take to go there.

doc/source/graphicsItems/make
```diff
@@ -11,6 +11,7 @@ GraphicsWidget
 GridItem
 HistogramLUTItem
 ImageItem
+pColorMeshItem
```
j9ac9k, line 14: Should be upper case?

examples/pColorMeshItem.py
```diff
@@ -0,0 +1,65 @@
+# -*- coding: utf-8 -*-
+"""
+Demonstrates very basic use of pColorMeshItem
```
j9ac9k, line 3: Make upper case

examples/pColorMeshItem.py
```diff
@@ -0,0 +1,65 @@
...
+
+def updateData():
+    global pcmi, x, y, z, i
+
+    ## Display the data
+    pcmi.setData(x-i, y, z)
+
+    QtCore.QTimer.singleShot(fps*1000, updateData)
+    i += 1
+    print(i)
```
j9ac9k, line 56: is the print statement here needed?

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,125 @@
...
+    from collections import Callable
+
+__all__ = ['PColorMeshItem']
+
+
+class PColorMeshItem(GraphicsObject):
+    """
+    **Bases:** :class:`GraphicsObject <pyqtgraph.GraphicsObject>`
+
+    TODO
```
j9ac9k, line 26: Fill out docstring

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,125 @@
...
+        GraphicsObject.__init__(self)
+
+        self.x = x
+        self.y = y
+        self.z = z
+        self.qpicture = None  ## rendered image for display
+
+        self.axisOrder = getConfigOption('imageAxisOrder')
+
+        if cmap in list(Gradients.keys()):
```
j9ac9k, line 46: Do you need to convert to a list here?  doing `in` lookups in the `dict.keys()` I know works in python3, does it not work in python2? (added bonus of not converting to list, is that `in` lookups in the hashed objects are significantly faster.

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,125 @@
...
+        return len(self.y)
+
+
+
+
+    def boundingRect(self):
+
+        if self.z is None:
+            return QtCore.QRectF(0., 0., 0., 0.)
+        return QtCore.QRectF(0., 0., float(self.width()), float(self.height()))
```
j9ac9k, line 125: new line at the end of the file

j9ac9k:
@edumur this PR looks good to me, I made some in-line comments for suggested changes and questions.  I haven't worked w/ these graphical objects before so I'm going to let some other folks around here give things a look over.

Thanks for the PR!

edumur:
@j9ac9k ok, I will work on your comments and follow with a PColorMeshView that will offer axis and histogram as ImageView.

edumur:
Hi,

This is where I am now.
I still have work to do but I thought people may already have comments about it.

j9ac9k:
Thanks for the updated PR.  There is a python 2 syntax error on line 32 of PColorMeshItem.py ... Sorry I'm on mobile and can't better link it, but if you click through the CI checks it should eventually show you the error.

j9ac9k:
I would guess it doesn't like the trailing coma, I don't think that syntax is supported with Python 2.7

```
2020-06-28T20:00:25.1757554Z Traceback (most recent call last):
2020-06-28T20:00:25.1758170Z   File "<string>", line 1, in <module>
2020-06-28T20:00:25.1758641Z   File "pyqtgraph/__init__.py", line 222, in <module>
2020-06-28T20:00:25.1759041Z     from .graphicsItems.PColorMeshItem import * 
2020-06-28T20:00:25.1759442Z   File "pyqtgraph/graphicsItems/PColorMeshItem.py", line 32
2020-06-28T20:00:25.1760546Z     cmap='viridis', edgecolors=None, ):
2020-06-28T20:00:25.1760968Z        ^
2020-06-28T20:00:25.1762039Z SyntaxError: invalid syntax
```

edumur:
Nope it's weird...
I also didn't get how I could run these test locally (without commit and push), if someone could give me some hint.

j9ac9k:
> Nope it's weird...
> I also didn't get how I could run these test locally (without commit and push), if someone could give me some hint.

The way I do it is configure a python2 virtual environment.  To setup pyside1 or qt4, you would probably need to use a conda environment; however if you're just checking for syntax errors, you may be able to find another tool to do it that won't need you to install the dependencies.

j9ac9k:
@edumur I have a python2 environment configured on my machine, I'll test it out when I wrap up work later today and see if I can identify the fix.

2xB:
This might be late, but the problem is:
Python2 does not allow any arguments other than variable-length keyword arguments after variable-length positional parameters (which is broken in this case by `cmap='viridis', edgecolors=None` after `*args`). So `*args, **kwargs` would be the way to go here to preserve Python2 compatibility.

j9ac9k:
Finally setup my local python2 dev environment (got a new machine)

Sure enough, I can replicate the CI issue with a python2 environment:

```
$ python -c "import pyqtgraph as pg;pg.systemInfo()"
  File "<string>", line 1, in <module>
  File "pyqtgraph/__init__.py", line 221, in <module>
    from .graphicsItems.PColorMeshItem import *
  File "pyqtgraph/graphicsItems/PColorMeshItem.py", line 31
    def __init__(self, *args, cmap='viridis', edgecolors=None):
                                 ^
SyntaxError: invalid syntax
```

I can skip this error by moving `*args` to the end of the argument list in the `__init__` method:

```python
def __init__(self, cmap='viridis', edgecolors=None, *args):
```
 
There are a few other  issues, `examples/pColorMeshItem.py` should be renamed to `examples/PColorMeshItem.py` (this can  be done w/ `git mv` command).

also, your example should be added to the app!  This can be done by adding the entry in `examples/utils.py` I would add it under the `GraphicsItems` section

```
        ('Labeled Graph', 'CustomGraphItem.py'),
        ('PColorMeshItem', 'PColorMeshItem.py')
    ])),
```

After I made those changes (And some other for unrelated Py2 example app errors outside the scope of this PR) your example worked wonderfully.

ixjlyons:
I was thinking of advocating for fully explicit arguments e.g. `PColorMeshItem(x=None, y=None, z=None, cmap='viridis'...)`, but `*args, **kwargs` would make transitioning to `*args, cmap='viridis', ...` after dropping Python 2 simpler I guess. The docstring as it is looks pretty good, so I'm fine with that.

For what it's worth, I generally don't care for throwing everything in `*args, **kwargs` unless it's obvious they're passed along to some other function/parent or the possible argument combinations are really flexible (and in the latter case, classmethods/factory functions may be easier to understand/document).

j9ac9k:
The `*args` and `**kwargs` everywhere reminds me of matplotlib, it can definitely get disorienting with what one _should_ be passing.  I have no strong preference in this case on if `*args` should be preserved as is or not; I was mostly just trying to get it to run on python 2...

edumur:
Hi all,

I think I implemented everything which was discussed above.
I tested the example in a python 2 environment and it ran smoothly.
I went for `*args` and `**kwargs` but I do not have strong opinion here.

I will progress towards a PColorItem now.
Feedback are always welcome.

j9ac9k:
All CI pipelines failed with

```
2020-07-15T08:39:21.7360444Z =================================== FAILURES ===================================
2020-07-15T08:39:21.7364888Z _________________ testExamples[ PColorMeshItem.py - PySide2 ] __________________
2020-07-15T08:39:21.7367772Z [gw0] linux -- Python 3.8.3 /opt/hostedtoolcache/Python/3.8.3/x64/bin/python
2020-07-15T08:39:21.7368283Z 
2020-07-15T08:39:21.7368533Z 
2020-07-15T08:39:21.7368937Z Python 3.8.3 (default, Jun  9 2020, 16:31:42) 
2020-07-15T08:39:21.7369330Z [GCC 7.5.0] on linux
2020-07-15T08:39:21.7369736Z Type "help", "copyright", "credits" or "license" for more information.
2020-07-15T08:39:21.7370393Z >>> >>> ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... Traceback (most recent call last):
2020-07-15T08:39:21.7372059Z   File "<stdin>", line 6, in <module>
2020-07-15T08:39:21.7372901Z ModuleNotFoundError: No module named 'PColorMeshItem'
2020-07-15T08:39:21.7374272Z >>> 
2020-07-15T08:39:21.7374574Z 
2020-07-15T08:39:21.7375905Z Failed PColorMeshItem Example Test Located in PColorMeshItem.py 
```

I'll investigate in a bit...

j9ac9k:
oh BTW thanks @edumur for updating the PR, it's almost there!

ixjlyons:
Looks like the example script is still named with a lowercase "p".

edumur:
The file name and all occurrences of `PColorMeshItem` are written with an uppercase 'P'.
The only occurrence I find with a lowercase 'p' is in the `index.rst` file.
Did I rename the file incorrectly with git?

j9ac9k:
Took a screenshot to illustrate; you can see this in the "Files Changed" tab of this PR.

image

`git mv examples/pColorMeshItem.py examples/PColorMeshItem.py` should do the trick.

edumur:
ok so what I did is putting myself in the examples folder and then did the git rename command.
Apparently doing this doesn't work
Let's see how it goes now.

j9ac9k:
closer, all pyside pipelines are failing 

```
----------------------------- Captured stdout call -----------------------------
PColorMeshItem 

Python 2.7.18 |Anaconda, Inc.| (default, Apr 23 2020, 22:42:48) 
[GCC 7.3.0] on linux2
  File "PColorMeshItem.py", line 77, in <module>
    updateData()
  File "PColorMeshItem.py", line 72, in updateData
    new_z)
  File "/home/vsts/work/1/s/pyqtgraph/graphicsItems/PColorMeshItem.py", line 204, in setData
    QtCore.QPointF(self.x[xi][yi+1],   self.y[xi][yi+1]))
TypeError: drawConvexPolygon() takes exactly one argument (4 given)
>>> 
QPaintDevice: Cannot destroy paint device that is being painted
Fatal Python error: Segmentation fault

Current thread 0x00007fd0771a5580 <python> (most recent call first):
```

I can replicate locally with a PySide Qt bindings installed and running

```
pytest -k "PColorMeshItem"
```

(so you don't have to run the entire test suite).

j9ac9k:
Doing a little work procrastination; tinkering with this, wrapping the qpoint's with a `[]` makes it work for pyside but it fails on with pyqt5

```
  File "/Users/ogi/Developer/pyqtgraph/examples/PColorMeshItem.py", line 77, in <module>
    updateData()
  File "/Users/ogi/Developer/pyqtgraph/examples/PColorMeshItem.py", line 70, in updateData
    pcmi.setData(new_x,
  File "/Users/ogi/Developer/pyqtgraph/pyqtgraph/graphicsItems/PColorMeshItem.py", line 203, in setData
    p.drawConvexPolygon([QtCore.QPointF(self.x[xi][yi],     self.y[xi][yi]),
  drawConvexPolygon(self, Union[QPointF, QPoint], *): argument 1 has unexpected type 'list'
  drawConvexPolygon(self, QPolygonF): argument 1 has unexpected type 'list'
  drawConvexPolygon(self, QPoint, *): argument 1 has unexpected type 'list'
  drawConvexPolygon(self, QPolygon): argument 1 has unexpected type 'list'
>>>
```

will keep messing with it...

j9ac9k:
Got it working with both configurations made the change in `PColorMeshItem.py` such that `drawConvexPolygon()` took a `QPolygonF` type object.

```
                polygon = QtGui.QPolygonF(
                    [QtCore.QPointF(self.x[xi][yi],     self.y[xi][yi]),
                     QtCore.QPointF(self.x[xi+1][yi],   self.y[xi+1][yi]),
                     QtCore.QPointF(self.x[xi+1][yi+1], self.y[xi+1][yi+1]),
                     QtCore.QPointF(self.x[xi][yi+1],   self.y[xi][yi+1])]
                )

                # DrawConvexPlygon is faster
                p.drawConvexPolygon(polygon)
```

edumur:
I am thankful ! 
I will integrate the change in a few days.

j9ac9k:
I'm good with this as is, @ixjlyons any other requested changes?

doc/source/graphicsItems/make
```diff
@@ -11,6 +11,7 @@ GraphicsWidget
 GridItem
 HistogramLUTItem
 ImageItem
+PColorMeshItem
```
ixjlyons, line 14: I'm not actually sure these "make" files do anything in the current docs build. I'm not aware of something like this being part of sphinx - my guess is this was part of an older build setup to generate the stub files for the API reference. Could you copy one of the other files in the `doc/source/graphicsItems` folder?

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,253 @@
...
+        Create a pseudocolor plot with convex polygons.
+
+        Call signature:
+
+        PColorMeshItem([x, y,] z, **kwargs)
+
+        x and y can be used to specify the corners of the quadrilaterals.
+        z must be used to specified to color of the quadrilaterals.
+
+        Parameters
```
ixjlyons, line 42: Unfortunately, we haven't yet moved to NumPy style doc strings, so these will fail to build with the current sphinx config. It might be possible to just enable the `sphinx.ext.napoleon` extension and both styles would be accepted. I can look into that if you want or you can go ahead and try it.

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,253 @@
...
+        ----------
+        x, y : np.ndarray, optional, default None
+            2D array containing the coordinates of the polygons
+        z : np.ndarray
+            2D array containing the value which will be maped into the polygons
+            colors.
+            If x and y is None, the polygons will be displaced on a grid
+            otherwise x and y will be used as polygons vertices coordinates as:
+
+            (x[i+1, j], y[i+1, j])           (x[i+1, j+1], y[i+1, j+1])
```
ixjlyons, line 52: These diagrams probably need to be put in a code environment to render correctly.

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+class PColorMeshItem(GraphicsObject):
+    """
+    **Bases:** :class:`GraphicsObject <pyqtgraph.GraphicsObject>`
+    """
+
+    sigImageChanged = QtCore.Signal()
+    sigRemoveRequested = QtCore.Signal(object)  # self; emitted when 'remove' is selected from context menu
+
```
ixjlyons, line 29: Nitpick: there are extra lines between methods and such in this class that could be removed.

pyqtgraph/graphicsItems/PColorMeshItem.py
edumur: I added the napoleon extension in the conf.py file:
```python
extensions = ['sphinx.ext.autodoc', 'sphinx.ext.viewcode', 'sphinx.ext.napoleon']
```
Is that enough?
I run `sphinx-apidoc -f -o docs/source` but I do not now what to expect.

doc/source/graphicsItems/make
edumur: I do not understand what you mean by "Could you copy one of the other files in the `doc/source/graphicsItems` folder?".

pyqtgraph/graphicsItems/PColorMeshItem.py
edumur: How can I do that?

edumur:
Yep, I see the artifacts as well:
image

I tried to remove them but so far it didn't work.
My guess is that since the  y axis change over time the `height` function return a value to the `boundingRect` function which exclude part of the previous `QPicture` and so leave these artifacts.

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+    def paint(self, p, *args):
+        profile = debug.Profiler()
+        if self.z is None:
+            return
+
+        profile('p.drawPicture')
+        p.drawPicture(0, 0, self.qpicture)
```
j9ac9k, line 225: you may be able to redo this as something like 

```python
if self.qpicture is not None:
    self.qpicture.play(p)
```

The implementation taken from TargetItem.py https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/TargetItem.py#L73-L76

j9ac9k: wait, that probably does nothing, was reading the docs:

```
This function does exactly the same as QPainter::drawPicture() with (x, y) = (0, 0).
```

doc/source/graphicsItems/make
ixjlyons: For now at least, the API docs are manually created, so the .rst files in `doc/source/graphicsItems/` are all stub files with pretty much the same content. So the idea was to copy one of those files and replace the class name with `PColorMeshItem`. legenditem.rst would be a good example since it includes the top class-level doc and the `__init__` method.

pyqtgraph/graphicsItems/PColorMeshItem.py
ixjlyons: Like I mentioned above, we're currently manually creating the API reference.

The good news is adding `sphinx.ext.napoleon` does just give us the ability to use numpy style docstrings without affecting how other docstrings are parsed (the others are "pure RST" so I guess that makes sense).

I'm :+1: for going ahead and committing that change to `conf.py` and then we don't have to overhaul the `PColorMeshItem` docstrings.

pyqtgraph/graphicsItems/PColorMeshItem.py
ixjlyons: You can take a look at how matplotlib is doing it: https://matplotlib.org/3.3.0/_modules/matplotlib/axes/_axes.html#Axes.pcolormesh

Basically, you can make a double-colon `::` where you already have one and then indent the code block one level more than the text it's surrounded by.

edumur: ok, it looks like this:

```python
"""
            otherwise x and y will be used as polygons vertices coordinates as::

                (x[i+1, j], y[i+1, j])           (x[i+1, j+1], y[i+1, j+1])
                                    +---------+
                                    | z[i, j] |
                                    +---------+
                    (x[i, j], y[i, j])           (x[i, j+1], y[i, j+1])
"""
```

pyqtgraph/graphicsItems/PColorMeshItem.py
edumur: ok done

doc/source/graphicsItems/make
edumur: Done, the file looks like this:
```rst
PColorMeshItem
==============

.. autoclass:: pyqtgraph.PColorMeshItem
    :members:

    .. automethod:: pyqtgraph.PColorMeshItem.__init__

```

pyqtgraph/graphicsItems/PColorMeshItem.py
edumur: Ok, so this is clearly it because if I add a coefficient in the `boundingRect` method to return, for instance,  `QtCore.QRectF(0., 0., float(self.width())*2., float(self.height())*2.)` then the artifacts disappear.

I tried fix to make it better but so far it didn't work well...

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,253 @@
...
+                # DrawConvexPlygon is faster
+                p.drawConvexPolygon(polygon)
+
+
+        p.end()
+        self.update()
+
+        if shapeChanged:
+            self.informViewBoundsChanged()
+
```
campagnola, line 216: If this would change the result of `boundingRect()`, then you should also call `self.prepareGeometryChange()`.
See: https://doc.qt.io/qt-5/qgraphicsitem.html#boundingRect

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,253 @@
...
+    def height(self):
+        if self.y is None:
+            return None
+        return np.max(self.y)
+
+
+
+
+    def boundingRect(self):
```
campagnola, line 250: Suggestion: returning `QtCore.QRectF(self.picture.boundingRect())` here might help with the motion artifacts.

edumur:
Thanks all, I implemented your suggestions.
Unfortunately the "artifacts" are still here.
They basically appear because a polygon is drawn somewhere at one frame and not erase at the next frame because its location does not overlap other polygons of that frame...
It is clearly not an issue for static data visualization.

j9ac9k:
hey @edumur just wanted to let you know I haven't forgotten about this, I want to get this merged as this is an awesome feature.  I just started experimenting w/ `QPainter` for the first time on my own project; I feel like there must be something really small to try and capture those artifacts... I'm messing around w/ your branch locally (I moved recently so I took a bit of a break from dev work while I got unpacked).

j9ac9k:
I'm likely not telling you anything you don't know; but doing some debugging here, and I suspect the issue is that your `boundRect()` value is changing, but you're not calling `prepareGeometryChange()` beforehand (even w/ @campagnola 's changes)

I added a couple of print statements

```python
        if shapeChanged:
            self.prepareGeometryChange()
            print("shape changed")
```

and 

```python

    def paint(self, p, *args):
        profile = debug.Profiler()
        if self.z is None:
            return

        profile('p.drawPicture')
        p.drawPicture(0, 0, self.qpicture)
        print(self.boundingRect())
```

wanting the console I get the following


```
pyqtgraph-pyqt5-latest ❯ python -m examples
/Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.8.3/Python.framework/Versions/3.8/lib/python3.8/runpy.py:127: RuntimeWarning: 'examples.__main__' found in sys.modules after import of package 'examples', but prior to execution of 'examples.__main__'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
qt.qpa.fonts: Populating font family aliases took 296 ms. Replace uses of missing font family "FreeMono" with one that exists to avoid this cost.
Using PyQt5 (default graphics system)
shape changed
/Users/ogi/Developer/pyqtgraph/examples/PColorMeshItem.py:75: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  QtCore.QTimer.singleShot(1000/fps, updateData)
PyQt5.QtCore.QRectF(1.0, 0.0, 54.0, 47.0)
PyQt5.QtCore.QRectF(1.0, -1.0, 54.0, 49.0)
PyQt5.QtCore.QRectF(1.0, -1.0, 54.0, 49.0)
PyQt5.QtCore.QRectF(1.0, -1.0, 54.0, 49.0)
PyQt5.QtCore.QRectF(1.0, -2.0, 54.0, 50.0)
PyQt5.QtCore.QRectF(1.0, -2.0, 54.0, 50.0)
PyQt5.QtCore.QRectF(1.0, -2.0, 54.0, 50.0)
PyQt5.QtCore.QRectF(1.0, -2.0, 54.0, 50.0)
PyQt5.QtCore.QRectF(1.0, -2.0, 54.0, 50.0)
PyQt5.QtCore.QRectF(1.0, -2.0, 54.0, 49.0)
PyQt5.QtCore.QRectF(1.0, -2.0, 54.0, 49.0)
PyQt5.QtCore.QRectF(1.0, -2.0, 54.0, 48.0)
PyQt5.QtCore.QRectF(1.0, -2.0, 54.0, 48.0)
PyQt5.QtCore.QRectF(1.0, -1.0, 54.0, 46.0)
PyQt5.QtCore.QRectF(1.0, 0.0, 54.0, 45.0)
```

So we're having boundingRect change, but not calling `prepareGeometryChange` ahead of time... I'll see If I can see if we can determine more accurately when the boundingRect changes...

j9ac9k:
Ha, I think I got it, sure enough, a pretty dead simple change; move `self.prepareGeometryChagne()` to be outside the `if shapeChanged` block (just have that line run at the end of `setData()`)... I'm not sure some of the colors within the PColorMeshItem are quite right (maybe not?) but there are no visible artifacts away from the animated body.

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,251 @@
...
+                shapeChanged = True
+
+        profile = debug.Profiler()
+
+        self.qpicture = QtGui.QPicture()
+        p = QtGui.QPainter(self.qpicture)
+        
+        # We set the pen of all polygons once
+        if self.edgecolors is None:
+            p.setPen(QtGui.QColor(0, 0, 0, 0))
```
j9ac9k, line 173: Probably need to do `p.setPen(fn.mkPen(QtGui.QColor(0, 0, 0, 0)))` as I think `setPen` requires a `QPen` object

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,251 @@
...
+        if self.qpicture is None:
+            shapeChanged = True
+        elif len(args)==1:
+            if args[0].shape[0] != self.x[:,1][-1] or args[0].shape[1] != self.y[0][-1]:
+                shapeChanged = True
+        elif len(args)==3:
+            if np.any(self.x != args[0]) or np.any(self.y != args[1]):
+                shapeChanged = True
+
+        profile = debug.Profiler()
```
j9ac9k, line 166: Useful for debugging, but might as well remove from the version we merge.

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,251 @@
...
+        norm  = self.z - self.z.min()
+        norm = norm/norm.max()
+        norm  = (norm*(len(lut)-1)).astype(int)
+        
+        # Go through all the data and draw the polygons accordingly
+        for xi in range(self.z.shape[0]):
+            for yi in range(self.z.shape[1]):
+                
+                # Set the color of the polygon first
+                # print(xi, yi, norm[xi][yi])
```
j9ac9k, line 193: no need for commented out print statement

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,251 @@
...
+                c = lut[norm[xi][yi]]
+                p.setBrush(QtGui.QColor(c[0], c[1], c[2]))
```
j9ac9k, line 195: probably should also wrap this with a `fn.mkBrush()`

examples/PColorMeshItem.py
```diff
@@ -0,0 +1,83 @@
...
+    ## Display the new data set
+    new_x = x
+    new_y = y+wave_amplitude*np.cos(x/wave_length+i)
+    new_z = np.exp(-(x-np.cos(i*color_speed)*xn)**2/1000)[:-1,:-1]
+    pcmi.setData(new_x,
+                 new_y,
+                 new_z)
+
+    i += wave_speed
+    QtCore.QTimer.singleShot(1000/fps, updateData)
```
j9ac9k, line 75: I'm getting a deprecation warning here for using `float`

```
/Users/ogi/Developer/pyqtgraph/examples/PColorMeshItem.py:75: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  QtCore.QTimer.singleShot(1000/fps, updateData)
```

Probably should change this line to:

```python
QtCore.QTimer.singleShot(1000//fps, updateData)
```

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -0,0 +1,251 @@
...
+                shapeChanged = True
+        elif len(args)==3:
+            if np.any(self.x != args[0]) or np.any(self.y != args[1]):
+                shapeChanged = True
+
+        profile = debug.Profiler()
+
+        self.qpicture = QtGui.QPicture()
+        p = QtGui.QPainter(self.qpicture)
+
```
j9ac9k, line 170: totally optional here, but I've been experimenting on my own, but consider setting the anti-aliassing render hint

```python
        p.setRenderHint(QtGui.QPainter.Antialiasing)
```

With this set, the lines were actually visible on my screen; admittedly that may not be what you want.

edumur:
Hi @j9ac9k,

Thanks for the comments.
I believed to have implemented everything you described.
I added an optional parameter to take into account the antialiasing.

j9ac9k:
That's absolutely perfect; thanks so much for this feature @edumur , merging!

PR #1275, opened 2020-06-25
Fix Parameter.hasDefault

### LINKED ISSUE #846: Parameter: hasDefault() always returns true after initializing a new Parameter even without passing in default value (and potential fix for this in comments)

ixjlyons:
This is a slightly different version of #847 and #331. There's some discussion in #847 for background.

Essentially, it seems we got stuck deciding whether or not explicitly passing `Parameter(default=None)` or calling `param.setDefault(None)` should make `param.hasDefault() == True`.

In this version, `param.hasDefault() == False` if you set it to `None` or don't set it at all. I eventually came around to lean toward this behavior by realizing that if `Parameter.__init__` didn't hide its argument structure in `**opts`, we'd have `default=None` and it wouldn't even be possible to tell whether or not the user set it explicitly. It's also fairly natural in Python for `None` to represent "unspecified."

I also implemented an alternative version where `param.hasDefualt() == True` when you `setDefault(None)`. You can see that here to compare: https://github.com/ixjlyons/pyqtgraph/commit/6d506f27d21ef0102de62f49167c45fd86aeda98

Fixes #847 
Fixes #846 
Fixes #331

### DIFF
--- a/pyqtgraph/parametertree/Parameter.py
+++ b/pyqtgraph/parametertree/Parameter.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from ..Qt import QtGui, QtCore
 import os, weakref, re
 from ..pgcollections import OrderedDict
@@ -188,13 +189,11 @@ class Parameter(QtCore.QObject):
         
         self.addChildren(self.opts.pop('children', []))
         
-        self.opts['value'] = None
         if value is not None:
             self.setValue(value)
 
         if 'default' not in self.opts:
             self.opts['default'] = None
-            self.setDefault(self.opts['value'])
     
         ## Connect all state changed signals to the general sigStateChanged
         self.sigValueChanged.connect(lambda param, data: self.emitStateChanged('value', data))
@@ -420,7 +419,7 @@ class Parameter(QtCore.QObject):
 
     def hasDefault(self):
         """Returns True if this parameter has a default value."""
-        return 'default' in self.opts
+        return self.opts['default'] is not None
         
     def valueIsDefault(self):
         """Returns True if this parameter's value is equal to the default value."""
--- a/pyqtgraph/parametertree/tests/test_Parameter.py
+++ b/pyqtgraph/parametertree/tests/test_Parameter.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+import pytest
+from pyqtgraph.parametertree import Parameter
+
+
+def test_parameter_hasdefault():
+    opts = {'name': 'param', 'type': int, 'value': 1}
+
+    # default unspecified
+    p = Parameter(**opts)
+    assert not p.hasDefault()
+
+    p.setDefault(1)
+    assert p.hasDefault()
+    assert p.defaultValue() == 1
+
+    # default specified
+    p = Parameter(default=0, **opts)
+    assert p.hasDefault()
+    assert p.defaultValue() == 0
+
+
+@pytest.mark.parametrize('passdefault', [True, False])
+def test_parameter_hasdefault_none(passdefault):
+    # test that Parameter essentially ignores defualt=None, same as not passing
+    # a default at all
+    opts = {'name': 'param', 'type': int, 'value': 0}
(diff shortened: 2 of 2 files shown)


pyqtgraph/parametertree/Parameter.py
```diff
@@ -188,13 +189,11 @@ def __init__(self, **opts):
         
         self.addChildren(self.opts.pop('children', []))
         
-        self.opts['value'] = None
```
ixjlyons, line 191: This is just a duplicate from above

j9ac9k:
python2 strikes again:

```
2020-06-25T03:54:48.6699544Z ==================================== ERRORS ====================================
2020-06-25T03:54:48.6700102Z _______ ERROR collecting pyqtgraph/parametertree/tests/test_Parameter.py _______
2020-06-25T03:54:48.6701312Z /usr/share/miniconda/envs/test-environment-2.7/lib/python2.7/site-packages/_pytest/python.py:507: in _importtestmodule
2020-06-25T03:54:48.6701914Z     mod = self.fspath.pyimport(ensuresyspath=importmode)
2020-06-25T03:54:48.6702755Z /usr/share/miniconda/envs/test-environment-2.7/lib/python2.7/site-packages/py/_path/local.py:704: in pyimport
2020-06-25T03:54:48.6703284Z     __import__(modname)
2020-06-25T03:54:48.6703694Z E     File "/home/vsts/work/1/s/pyqtgraph/parametertree/tests/test_Parameter.py", line 18
2020-06-25T03:54:48.6704141Z E       p = Parameter(**opts, default=0)
2020-06-25T03:54:48.6704659Z E                           ^
2020-06-25T03:54:48.6704969Z E   SyntaxError: invalid syntax
```

j9ac9k:
for the record, I'm all good with `setDefault(None)` resulting in `.hasDefault() is False`

ixjlyons:
Amazing how many 2/3 differences I've forgotten about. I ought to start running tests locally at least with Py3-Qt5 and Py2-Qt4 to weed out silly issues before pushing.

j9ac9k:
> Amazing how many 2/3 differences I've forgotten about. I ought to start running tests locally at least with Py3-Qt5 and Py2-Qt4 to weed out silly issues before pushing.

you could do that, or wait 15 minutes for the CI to run .... having attempted to setup a tox setup for all the configurations, it's just too fragile ...should be easier post NEP-29 adoption.

j9ac9k:
LGTM, thanks for the PR @ixjlyons merging.

PR #1279, opened 2020-06-26
Implement PR313 Diff + Enable Infinite Line Example

j9ac9k:
This PR Implements PR 313 which cannot be merged due to coming from an `unknown repository` and there being a merge conflict.  Thank you to @lesauxvi for the PR, sorry it's taken us so long to follow up.

Fixes #313

### DIFF
--- a/examples/InfiniteLine.py
+++ b/examples/InfiniteLine.py
@@ -46,6 +46,10 @@ cursorLabel = pg.CursorItem(pos=[20,-20], radius=3, cursor='c', pen=(0, 0, 255),
                             labelOpts={'color': (200, 200, 200), 'fill': (20, 20, 20, 20)})
 p1.addItem(cursorLabel)
 
+target = pg.TargetItem()
+target.setLabel("test")
+p1.addItem(target)
+
 ## Start Qt event loop unless running in interactive mode or using pyside.
 if __name__ == '__main__':
     import sys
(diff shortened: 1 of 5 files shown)


j9ac9k:
In slack chat @campagnola pointed out this feature is _really_ similar to [TargetItem](https://github.com/pyqtgraph/pyqtgraph/blob/f76fb3ac72281b942c632ea1198e84a2be55f4ae/pyqtgraph/graphicsItems/TargetItem.py).  I think this PR should be modified to incorporate its features into the TargetItem object instead of creating a separate one

j9ac9k:
CursorItem is far more feature complete than TargetItem, ....going to collect some feedback from others, but I'm considering renaming CursorItem to TargetItem; and overwriting the current implementation.

ixjlyons:
> I'm considering renaming CursorItem to TargetItem; and overwriting the current implementation

Sounds good to me :+1:

Should `CursorItem` be made more visually compatible with `TargetItem` in that case (i.e. the crosshairs)? It makes it a little more distinct from an ROI and gives a more precise indication of the center.

j9ac9k:
Started merging the CursorItem into the existing TargetItem, while supporting the array of existing Symbols as potential entries.... hover pen isn't working, and neither is click + drag...more investigation needed.

j9ac9k:
Went for a different way, closing this PR

PR #1051, opened 2019-10-01
Matplotlib exporter scientific notation bugfix

penkod:
Fix setting labels and values in case of plots which use scientific notation.

Solves issue #1050

PR #1124, opened 2020-02-09
Syntax highlighting for examples.

dargor:
Allow bad eyesighted people like me to read and play with examples.

Tested and working with both `PyQt5` and `PySide2`, on Python `3.6` and `3.7`.

Based on https://github.com/art1415926535/PyQt5-syntax-highlighting

PR #1282, opened 2020-06-27
Handle axis SI prefix scaling in MatplotlibExporter

### LINKED ISSUE #1050: Bug: Export to Matplotlib while scientific notation is being used
Pyqgraph by default uses scientific notation, meaning that when values higher than 1000000 are present, all such values are reduced to e.g. "<reduced value> x1e+06" and the plot label is changed accordingly. However, if we then try to export the plot to matplotlib window, the original values are kept **AND** the (now wrong) label, which includes also the "(x1e+06)" part, making the plot state that the values are then **x 1e+12**.
pyqtgraph_matplotlib_bug

ixjlyons:
Finally following up #1051 

Basic test:

```python
import pyqtgraph as pg
from pyqtgraph.exporters import MatplotlibExporter

plt = pg.plot(x=[0, 1, 2], y=[1e10, 2e10, 3e10])
plt.setLabel('left', 'magnitude')

exp = MatplotlibExporter(plt.plotItem)
exp.export()

pg.mkQApp().exec_()
```

I ended up sticking with the basic idea of #1051, keeping the axis labels and scaling the data. The alternative I had considered was stripping the prefix from the label and letting matplotlib handle the axis scaling with a multiplier label, but this maintains the pyqtgraph style.

I recommend viewing the diff ignoring whitespace changes. I usually try to make minimal diffs, but removing the unnecessary indentation level seemed worthwhile.

~No test because we don't have matplotlib in CI afaik.~

Fixes #1050 
Fixes #1051

### DIFF
--- a/pyqtgraph/exporters/Matplotlib.py
+++ b/pyqtgraph/exporters/Matplotlib.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from ..Qt import QtGui, QtCore
 from .Exporter import Exporter
 from .. import PlotItem
@@ -31,6 +32,7 @@ publication. Fonts are not vectorized (outlined), and window colors are white.
 class MatplotlibExporter(Exporter):
     Name = "Matplotlib Window"
     windows = []
+
     def __init__(self, item):
         Exporter.__init__(self, item)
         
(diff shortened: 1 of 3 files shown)


j9ac9k:
FYI we can easily add matplotlib to our CI suite if you think this should be tested

j9ac9k:
Should probably have a skip if decorator for skipping the test if matplotlib isn't installed

ixjlyons:
I think it's probably worth adding some test coverage. I'm not sure how much we could/should test in the case of `MatplotlibExporter`, but I'm up for putting together some basic tests to hit the code paths.

j9ac9k:
Sounds good, I'm not at a computer right now, do you need help with wing matplotlib to the CI pipelines?

ixjlyons:
Was about to say nah I got it but of course there's a failure...

I think I can figure it out but chime in if you notice something. At the moment I see the naive approach of adding matplotlib to the dependency install line pulled in (py)qt-5 in the Python2-PySide-4.8 job.

j9ac9k:
Those Qt4/PySide1/Python2 configs are no joke.

In the "install dependencies" step of the linux/pyside1 pipeline

```
  pyqt-5.6.0                 |           py27_2         5.3 MB
    pyside-1.2.0               |           py27_0         5.6 MB
```

I won't be able to help debug this for a few hours; but the problem seems to be that installing matplotlib via conda bundles pyqt5 along with it; which is not at all what we want...

I think we may want to install matplotlib via `pip` _after_ installing the other dependencies via conda.

ixjlyons:
Wow, only the Linux Python2-PySide-4.8 job failed despite (py)qt-5 being installed for Windows/MacOS too.

https://dev.azure.com/pyqtgraph/pyqtgraph/_build/results?buildId=996&view=logs&j=fdf8a08f-d10f-57ad-9e84-c0e040fe4fde&t=50eff170-233a-5108-7962-acb3810e8052&l=190

Well, I have to say I was afraid of pretty much exactly this. There's probably a way to write the requirement spec to avoid it, but I'm pretty sure that would ugly up the CI template quite a bit.

@j9ac9k what do you think about only installing it for the pip jobs and letting pytest skip the tests otherwise (skipping alrady implemented)? I guess it would be nice to test this with a variety of Qt versions particularly because it's matplotlib with Qt backends though.

j9ac9k:
I think we can make this work by installing via pip and specifying which matplotlib dependencies we install.  I'm definitely :-1: on writing spec files.

I think either way we should use the skipif decorator and skip those tests if matplotlib is not installed

ixjlyons:
So just installing with pip in the conda environments seems to work. The PyPI package makes no assumptions about what backend you want, so I think this is a pretty stable solution.

The whole test module is skipped with `pytest.importorskip("matplotlib")`.

Going to move this out of draft but feedback still welcome obviously.

j9ac9k:
This might be outside of the scope of this PR, but does matplotlib qt widget offer any kind of dark mode recognition?  on macOS, which is running dark mode, this is what the window looks like:

image

pyqtgraph does have some dark mode recognition in the codebase that was introduced as part of #1124 

in `__main__.py`

```python

        self.paletteChanged.connect(self.onPaletteChange)
        self.onPaletteChange(self.palette())

    def onPaletteChange(self, palette):
        self.dark_mode = palette.base().color().name().lower() != "#ffffff"
```

j9ac9k:
Some googling turns up this issue:

https://github.com/matplotlib/matplotlib/issues/8590

A bit more googling, and light icons were merged, but they are slated to be in the 3.3.0 release!

https://github.com/matplotlib/matplotlib/pull/17459

I'm good with hoping this sorts itself out for matplotlib 3.3.0

pyqtgraph/exporters/Matplotlib.py
```diff
@@ -55,58 +57,72 @@ def cleanAxes(self, axl):
...
+        xscale = yscale = 1.0
+        if xax.autoSIPrefix:
+            xscale = xax.autoSIPrefixScale
+        if yax.autoSIPrefix:
+            yscale = yax.autoSIPrefixScale
+
+        ax = fig.add_subplot(111, title=title)
+        ax.clear()
+        self.cleanAxes(ax)
+        #ax.grid(True)
```
j9ac9k, line 86: delete commented out line?

PR #1284, opened 2020-06-28
Update doc strings to clarify getArrayRegion API for ROI subclasses

ixjlyons:
Re-implementation of #116. Some of the work in that PR was overcome by events - particularly adding/passing the `axes` kwarg which was implemented in the meantime. I opted to make a note that `returnMappedCoords` isn't supported rather than raising an exception. It should be pretty clear from the method signatures where it's not directly supported, but the presence of `**kwds` could make it seem like you can pass additional arguments through to the base `ROI.getArrayRegion`.

One additional thing I didn't pull from #116 is renaming `arr` -> `data`. I think the correspondence is pretty clear, and technically it would be an API change since users could be using "named arguments" (e.g. `roi.getArrayRegion(arr=imgdata, img=imgitem)`).

Fixes #116

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -1662,6 +1662,15 @@ class MultiRectROI(QtGui.QGraphicsObject):
         return pos
         
     def getArrayRegion(self, arr, img=None, axes=(0,1), **kwds):
+        """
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks @ixjlyons ! LGTM

PR #1285, opened 2020-06-29
Improve control over ROI/handle pens

ixjlyons:
Picks up from #271. I fixed merge conflicts and added the suggested `handleHoverPen` argument. It turns out the option to set the `Handle` hover pen was added with #1243 but was called `activePen`. Given that was merged a couple weeks ago, I figured it'd be acceptable to rename it.

Fixes #271


j9ac9k:
Thanks @ixjlyons  LGTM!  I remember seeing #271 and thinking that I may have an issue w/ "Active Pen" ... Definitely fine with changing it, we don't have done a release yet.

PR #1286, opened 2020-06-29
convert connect='finite' to connect=ndarray

pijyoi:
Qt no longer accepts non-finite values for plotting as seen here:
https://github.com/qt/qtbase/blob/f6b7b64ed0168038e365b936a1daea9b3bcda335/src/gui/painting/qpainterpath.cpp#L2537

This would prevent PlotCurveItem.setData(connect='finite') from working.

This commit converts a call using connect='finite' to the equivalent call using connect=ndarray with all the non-finite elements removed.

Related to PR #1058 
Related to issue #1057

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1486,11 +1486,11 @@ def arrayToQPath(x, y, connect='all'):
     if eq(connect, 'finite'):
         # convert connect='finite' to connect=ndarray
         msk = np.isfinite(x) & np.isfinite(y)
-        connect = np.ones_like(msk)
+        connect = np.ones(len(msk)+1, dtype=bool)
         connect[np.where(~msk)[0] + 1] = 0
         x = x[msk]
         y = y[msk]
-        connect = connect[msk]
+        connect = connect[0:-1][msk]
 
     #profiler = debug.Profiler()
     n = x.shape[0]
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1486,11 +1486,11 @@ def arrayToQPath(x, y, connect='all'):
     if eq(connect, 'finite'):
         # convert connect='finite' to connect=ndarray
         msk = np.isfinite(x) & np.isfinite(y)
-        connect = np.ones(len(msk)+1, dtype=bool)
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for the PR @pijyoi 

From a glance, it looks like this PR would actually implement what PR #1058 is trying to achieve?

pijyoi:
It is trying to solve a smaller scope than PR #1058.
I think PR #1058 wants to allow NaNs in the plot data; whereas this PR only wants to fix it for the use-case connect='finite', where NaNs / Infs are delimiters and not part of the plot data.

j9ac9k:
Looks like in merging #1283 we created a merge conflict

pijyoi:
I have rebased and also modified the code to output the new ndarray syntax required by #1283

summoningdark:
This fixes my problem in [#1057](https://github.com/pyqtgraph/pyqtgraph/issues/1057) with a slight modification. Line 1490:
`connect[np.where(~msk)[0] + 1] = 0` 
can cause an IndexError if the last value in msk is false, since connect and msk have the same number of elements. Making connect one longer than msk, and then cutting off the last value in line 1493 seems to solve it.
`connect = np.ones_like(msk)` => `connect = np.ones(len(msk)+1)`
`connect = connect[msk]` => `connect = connect[0:-1][msk]`

pijyoi:
@summoningdark , thanks for the fix. I have incorporated it.
It is my understanding however that PR #1287 supersedes this PR as it fixes all the connect kinds.

j9ac9k:
@pijyoi i assigned @campagnola  to these issues given he's the author of 1 of the PRs and the low level nature of these changes.  Last I heard he wanted to do some performance evaluations on the proposals (iow we haven't forgotten about this PR (or the other ones), a solution with these fixes will be in the next release.

j9ac9k:
Hi @pijyoi with merging #1287 I believe the base issue that this PR was meant to address is now resolved.  Sorry this took us so long to follow up on.  I'm going to close out this PR, but if something isn't working as expected, or this PR covers a use-case that #1287 does not, let me know and I'll re-open.

PR #1287, opened 2020-06-30
Fill in non-finite plot values for Qt versions >= 5.12.3

### LINKED ISSUE #1057: plots not showing with NaN

campagnola:
Fixes #1057 

I have only manually tested this for a few cases; it would be nice to have some test coverage here before merging.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1510,8 +1510,10 @@ def arrayToQPath(x, y, connect='all'):
             idx[mask] = -1
             np.maximum.accumulate(idx, out=idx)
             first = np.searchsorted(idx, 0)
-            idx[:first] = first
-            arr[1:-1] = arr[1:-1][idx]
+            if first < len(x):
+                # Replace all non-finite entries from beginning of arr with the first finite one
+                idx[:first] = first
+                arr[1:-1] = arr[1:-1][idx]
 
     # decide which points are connected by lines
     if eq(connect, 'all'):
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -13,7 +13,7 @@ import ctypes
 import sys, struct
 from .pgcollections import OrderedDict
 from .python2_3 import asUnicode, basestring
-from .Qt import QtGui, QtCore, QT_LIB
+from .Qt import QtGui, QtCore, QT_LIB, QtVersion
 from . import getConfigOption, setConfigOptions
 from . import debug, reload
 from .reload import getPreviousVersion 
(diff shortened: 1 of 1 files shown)


campagnola:
Wow, ok.. turns out this has been attempted before in #1058 and #1286.

pyqtgraph/functions.py
```diff
@@ -1483,20 +1483,36 @@ def arrayToQPath(x, y, connect='all'):
...
+        isfinite = np.isfinite(x) & np.isfinite(y)
+        if not np.all(isfinite):
+            # credit: Divakar https://stackoverflow.com/a/41191127/643629
+            mask = ~isfinite
+            idx = np.arange(len(x))
+            idx[mask] = -1
+            np.maximum.accumulate(idx, out=idx)
+            first = np.searchsorted(idx, 0)
+            idx[:first] = first
+            arr[1:-1] = arr[1:-1][idx]
```
abrauchli, line 1514:
```suggestion
            if first < len(x):
                # Replace all non-finite entries from beginning of arr with the first finite one
                idx[:first] = first
                arr[1:-1] = arr[1:-1][idx]
```
Fixes the case where the whole buffer contains non-finite items in x or y (or both).

abrauchli:
I'm still hitting a case where everything just freezes with my all Y values being NaN buffer, so I can't really validate the code.. I'm not sure if it's related to the `arr[2:]['c'] = isfinite` which was patched in https://github.com/pyqtgraph/pyqtgraph/commit/b61c7c1e39dbaf81071b70ce696811ff76dfd582 .. will have to investigate more. Running on Py3.8 with PySide2 5.15.1 with pyqtgraph 0.11.0 + this patch and the mentioned commit.

Also, could it be that with the fill-in solution, that for time-series, where the X-axis is generally the (finite) timestamp, filling-in values will overwrite the timestamps of the points that aren't rendered and thus cause auto-scaling to behave differently since the NaN-cleaned version of the buffer will have a different starting point?
To solve that, I suppose we'd have to clean-up only the non-finite Y value with something like the code below. Since the NaN-turned-zero values are non-connect they shouldn't appear.

```python
    if qtver >= [5, 12, 3]:
        isfinite = np.isfinite(x) & np.isfinite(y)
        isfinite_x = np.isfinite(x)
        isfinite_y = np.isfinite(y)
        if not np.all(isfinite_x):
            # credit: Divakar https://stackoverflow.com/a/41191127/643629
            mask = ~isfinite_x
            idx = np.arange(len(x))
            idx[mask] = -1
            np.maximum.accumulate(idx, out=idx)
            first = np.searchsorted(idx, 0)
            if first < len(x):
                # Replace all non-finite X-axis points from beginning of arr with the first finite one
                idx[:first] = first # np.max(first, len(x) - 1)
                arr[1:-1] = arr[1:-1][idx]
        if not np.all(isfinite_y):
            mask = ~isfinite_y
            # Replace all non-finite Y-axis points with 0
            arr[1:-1]['y'][mask] = 0
```

PR #1302, opened 2020-07-08
example app now works with Qt4 and Python2 again

### LINKED ISSUE #1301: Example Application Cannot be Launched with Python2 Environment
running `python -m examples` (or `python -m pyqtgraph.examples` depending on how you have installed pyqtgraph) will result in an error

```
pyqtgraph-pyqt ❯ python -m examples
  File "/Users/ogi/.zinit/plugins/pyenv---pyenv/versions/miniconda2-latest/envs/pyqtgraph-pyqt/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/Users/ogi/.zinit/plugins/pyenv---pyenv/versions/miniconda2-latest/envs/pyqtgraph-pyqt/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Users/ogi/Developer/pyqtgraph/examples/__main__.py", line 159, in <module>
    run()
  File "/Users/ogi/Developer/pyqtgraph/examples/__main__.py", line 154, in run
    app = App([])
  File "/Users/ogi/Developer/pyqtgraph/examples/__main__.py", line 28, in __init__
TypeError: super() takes at least 1 argument (0 given)
```

This is in `__main__.py`

```python

        self.paletteChanged.connect(self.onPaletteChange)
```

There are two issues, one is the `super()` call, which we can easily work around with `super(QtGui.QApplication, self)`.  The second line in `__init__` is a bit more troublesome, as `paletteChanged` is a signal present in `QGuiApplication` in Qt5, but as far as I can tell, there is no equivalent signal in Qt4.  This should likely be implemented by safely checking if `paletteChanged` is an attribute of the super-class.

j9ac9k:
This PR fixes an issue introduced in #1124 where the example application was not compatible with Python2 (due to a python3 method of calling `super()`) and Qt4 (due to depending on `paletteChanged` signal which only exists in Qt5).

Fixes #1301


ksunden:
Is there a reason our tests didn't catch this? We do run the examples in the test suite, pretty sure.

j9ac9k:
Our test site runs the examples, but not the example application funny enough.

Should I look to add launching the example application as part of the test suite?

Also that CI failure, I am able to intermittently reproduce locally, but not reliably.

j9ac9k:
Been thinking, I think using `hasattr` is the wrong way to go about this, I should check with `if QT_LIB in ['PyQt4', 'PySide']:` which is how the rest of the library does it... will update the PR ...I'll see about adding the example app to the test suite as well.

j9ac9k:
Hmm getting the example app to run as part of the test suite is tricky haha

j9ac9k:
@ksunden can you think of a way we can add the example app as part of the test suite?  I just added it to https://github.com/pyqtgraph/pyqtgraph/blob/master/examples/utils.py#L93 where instead of `files = []` to `files = [("Example App", "__main__.py")]` but that just ran successfully (When it should have blown up).

j9ac9k:
whoops, accidentally pushed the wrong commit; going to put this in draft commit form until ready...

j9ac9k:
While this branch now works for having the example application launch on Python2 successfully, I want to try and minimize the warnings on python3 when launching the application, specifically....

```
/Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.8.3/Python.framework/Versions/3.8/lib/python3.8/runpy.py:127: RuntimeWarning: 'pyqtgraph.examples.__main__' found in sys.modules after import of package 'pyqtgraph.examples', but prior to execution of 'pyqtgraph.examples.__main__'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
qt.qpa.fonts: Populating font family aliases took 291 ms. Replace uses of missing font family "FreeMono" with one that exists to avoid this cost.
```

I get this warning when launching examples via `python -m pyqtgraph.examples`, but I do not get it when running

```python
import pyqtgraph.examples
pyqtgraph.examples.run()
```

or

```bash
$ python examples
```

j9ac9k:
Scary warning no longer comes up; still want `python -m pyqtgraph.examples` to work when installing in editable mode (`pip install -e .`) as well as the font warning:

```
qt.qpa.fonts: Populating font family aliases took 299 ms. Replace uses of missing font family "FreeMono" with one that exists to avoid this cost.
```

j9ac9k:
Apparently running examples in editable mode is related to a long-standing setuptools issue: https://github.com/pypa/setuptools/issues/230

We can work around this by changing the directory structure, but I don't think we actually want to do that for the additional functionality... last issue is fixing the font warning...

j9ac9k:
Switched font family to Courier New, which is available everywhere by default, had to regenerate the uic templates (always annoying)... 

Did last bit of code cleanup, I think this PR is ready for merging.

j9ac9k:
just kidding, still not ready for merging, test app didn't actually get run...

j9ac9k:
ok, _now_ the test application is tested as part of the test suite.

j9ac9k:
ok now that docs/CI and so on work, I think I'll just merge this guy...

Issue #1303, opened 2020-07-08
Import failure when launching cx_frozen application with pyqtgraph


berr:
### Short description
Freezing an application on linux with cx_freeze and python 3.8 leads to an error when importing pyqtgraph:
`ImportError: cannot import name 'ThreadsafeTimer' from partially initialized module 'pyqtgraph' (most likely due to a circular import)`

### Code to reproduce
Have the 3 files in your folder. Setup the virtual environment using requirements.txt and run the following commands:

```
python3 setup.py build  # build the executable
./build/exe.linux-x86_64-3.8/repro  #run it
```

File `repro.py`
```python
from pyqtgraph.Qt import QtGui, QtCore
import numpy as np
import pyqtgraph as pg

app = QtGui.QApplication([])
win = pg.GraphicsLayoutWidget(show=True, title="Basic plotting examples")
p1 = win.addPlot(title="Basic array plotting", y=np.random.normal(size=100))

if __name__ == '__main__':
    QtGui.QApplication.instance().exec_()
```
File `setup.py`
```python
import sys
from cx_Freeze import setup, Executable
from pathlib import Path

include_files = []

if sys.platform == "linux":
    # On linux the numpy libraries are in site-packages/numpy.libs, but cx_freeze copies them to
    # lib/numpy/*.so instead of lib/numpy.libs/*.so
    import numpy
    numpy_folder = Path(numpy.__file__).parent
    for numpy_shared_library in numpy_folder.with_name("numpy.libs").iterdir():
        include_files.append((str(numpy_shared_library), f"lib/numpy.libs/{numpy_shared_library.name}"))

build_options = {
    "packages": [],
    "includes": [],
    "excludes": [],
    "zip_exclude_packages": "*",
    "zip_include_packages": "",
    "include_files": include_files,
}


base = 'Win32GUI' if sys.platform == 'win32' else None

executables = [
    Executable('repro.py', base=base)
]

setup(name='bug repro',
      version='1.0',
      description='bug repro',
      options={"build_exe": build_options},
      executables=executables)

```

File requirements.txt
```
cx-Freeze==6.1
numpy==1.19.0
pyqtgraph==0.11.0
PySide2==5.14.2.2
shiboken2==5.14.2.2
```

### Expected behavior
The application should launch normally.

### Real behavior
An exception is raised:

```
Traceback (most recent call last):
[...]

PR #1304, opened 2020-07-08
Import ThreadSafeTimer in __init__

berr:
This fixes issue #1303


j9ac9k:
Thanks for the PR @berr this looks good to me, merging.

Issue #743, opened 2018-08-10
renderToArray not rendering opaque


PaulDebus:
Hey,

I am trying to create high resolution screenshots from my 3d scene. Using `GLViewWidget.readQImage()` works as expected, though the resolution is only the one of my screen. Using `GLViewWidget.renderToArray()` with the required size produces the correct resolution, but the `setGLOptions("opaque")` is not respected. In the attached image, you can see lines, that should be covered by the object.
![Image with visible lines behind the object](https://user-images.githubusercontent.com/4587428/43946153-c1947c0e-9c84-11e8-8c36-cc6b55ce2fb2.png)

Using the `readQImage()` function results in the correct image, as seen below.
![pfosten8](https://user-images.githubusercontent.com/4587428/43946234-f7738a22-9c84-11e8-835e-591964414575.png)

The code I use is the following:

```
        size = (self.center.width()*4, self.center.height()*4)
        pixels = self.center.renderToArray(size)
        tmp = pixels[...,0].copy()
        pixels[...,0] = pixels[...,2]
        pixels[...,2] = tmp
        pixels = pixels[::-1]
        pixels = np.rot90(pixels, k=3)
```

Am I doing something wrong here?

Thanks for your time and help!

PR #1306, opened 2020-07-13
Fix incorrect rendering of overlapping object in renderToArray()

ahendriksen:
This commit adds a depth buffer in renderToArray().

This fixes the issue that overlapping objects are rendered incorrectly
when using renderToArray() on a GLViewWidget.

This might be related to issue #743.


j9ac9k:
Thanks for the PR @ahendriksen sorry it's taken us so long to follow up with you.  The changes look good to me; merging.

ahendriksen:
Fantastic! Thanks for merging, and thanks for the wonderful package!

PR #1317, opened 2020-07-16
Fix line width not being set

Dennis-van-Gils:
Added missing gl.glLineWidth(width) in PlotCurveItem

### DIFF
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -580,6 +580,7 @@ class PlotCurveItem(GraphicsObject):
                 if pen.isCosmetic() and width < 1:
                     width = 1
                 gl.glPointSize(width)
+                gl.glLineWidth(width)
                 gl.glEnable(gl.GL_LINE_SMOOTH)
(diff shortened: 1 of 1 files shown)


j9ac9k:
Hi @Dennis-van-Gils thanks for the PR.

I'm not that familiar w/ the openGL implementations, would you mind pointing me to an example where the issue setting the line width here addresses?

Dennis-van-Gils:
Hi @j9ac9k thanks for quick reply.
I have tracked down the issue to a specific case.

When setting `pg.setConfigOptions(useOpenGL=True)`, all curves will show with the correct line width as set. But they are not anti-aliased (AA). So, when also enabling `pg.setConfigOptions(antialias=True)` the line width is still fine, but there is still no AA. To finally enable AA, also `pg.setConfigOptions(enableExperimental=True)` has to be set. Now all curves look stunning and the GPU acceleration is amazing. The last setting however breaks the set line width and fixes it to 1 px. My PR fixes that.

Minimal running example to reproduce the problem follows below:

j9ac9k:
@Dennis-van-Gils 

Thanks so much for the example and for the PR! Verified your fix, merging!

PR #1318, opened 2020-07-17
Improve target item - incorporate bits from PR 313

j9ac9k:
At the time #313 was created, `TargetItem` did not exist, but it is there now; but it's options are very limited.  PR #313 added the equivalent feature, but had some other methods such as `setPen`, `setBrush`, `setHoverPen` and so on... making the item far more feature complete.  It also included a label which could be configured for various text annotation purposes.

This PR attempts to merge the existing functionality, and the functionality in PR #313, with the addition of supporting more symbols (in-fact this PR makes it so you can use any symbol in the SymbolAtlas in ScatterPlotItem).

~What does not work is the positioning of the label that goes alongside the target.  The label is positioned correctly on load, but the moment the TargetItem is clicked + dragged anywhere, the TargetLabel gets positioned somewhere very incorrectly.~

~I'm opening this PR in case anyone wants to give it another go, or try something themselves.~

From the root folder, you can interact with the TargetItem by running `python examples/InfiniteLine.py`


j9ac9k:
Took this PR off of draft status, but I will say there are a few more changes I want to make.

1. I don't like how `TargetLabel` has the `valueChanged` method is structured; primarily because if it does not give the opportunity to the user to do any kind of calculations based on x, y position and displaying those results, it does give the `str.format()` formulation...but I think this can be expanded.
2. Some of the docstrings need updating.

Regarding the first point, I'm thinking there should be the option of some kind of mapping method, that a user provides...suppose it can be a plain string, or a formatted string as well...

j9ac9k:
I addressed the two issues I listed above, but there are still a few more issues.

1. The `TargetLabel.setPos(offset)` is moving the label down vs. the target item for a positive y-value.  I could flip the sign of the y-component, but if someone has an explanation as to why this is happening, and what I can do about it, I'd appreciate it.
2. The `radius` parameter, that governs the size of the target item is not used.

Regarding the second item, I think we should make all the symbols in the ScatterPlot Symbols potentially available here; and have the `radius` just indicate what size those symbols should be.

j9ac9k:
I now have the scaling working the way I want it, I migrated the crosshair to the `ScatterPlotItem.py` symbols dict, and I also made all the symbols be accessible by name for TargetItem... 

One thing that is not working right now is setting the text for the TargetLabel, which is using the default no matter what.  I'm getting a bit tired so I may try and wrap that up later.

j9ac9k:
This PR is functional in the way I intended; and I think is ready for feedback/review.

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,352 @@
...
+        """
+        Parameters
+        ----------
+        pos : list, tuple, QPointF, or QPoint
+            Initial position of the cursor.
+        diameter : int
+            Size of the cursor in pixels.  Default is 10.
+        cursor : str
+            String that defines the shape of the cursor (can take the following
+            values for the moment : 's' (square), 'c' (circle))
```
NilsNemitz, line 43: Hasn't the 'cursor' option been replaced by 'symbol'?

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,352 @@
...
+            `~pyqtgraph.ScatterPlotItem.setData`
+        label : bool, str or callable, optional
+            Text to be displayed in a label attached to the cursor, or None to
+            show no label (default is None). May optionally include formatting
+            strings to display the cursor value, or a callable that accepts x
+            and y as inputs.  If True, the label is "x = {:0.3f}, y = {:0.3f}"
+            False or None will result in no text being displayed
+        labelOpts : dict
+            A dict of keyword arguments to use when constructing the text
+            label. See :class:`CursorLabel` and :class: `~pyqtgraph.TextItem`.
```
NilsNemitz, line 74: CursorLabel --> TargetLabel ?

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,352 @@
...
 
-    def __init__(self, movable=True, radii=(5, 10, 10), pen=(255, 255, 0), brush=(0, 0, 255, 100)):
-        GraphicsObject.__init__(self)
+    sigPositionChanged = QtCore.Signal(object)
+    sigPositionChangeFinished = QtCore.Signal(object)
+
+    def __init__(
+        self,
+        pos=None,
+        diameter=10,
```
NilsNemitz, line 24: Should this use 'size' instead, to match ScatterPlotItem?

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,352 @@
...
+        :func:`~pyqtgraph.mkPen`."""
+        self.hoverPen = fn.mkPen(*args, **kwargs)
+        if self.mouseHovering:
+            self.currentPen = self.hoverPen
+            self.update()
+
+    def setLabel(self, text=None, labelOpts=None):
+        if text is None and self.label is not None:
+            # remove the label if it's already added
+            self.label.scene().removeItem(self.label)
```
NilsNemitz, line 178: If I call `setLabel(None)` , this fails for me with
`AttributeError: 'NoneType' object has no attribute 'removeItem'`

If I call `setLabel('some text')` nothing happens, the default text does not change. No error message.

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,352 @@
...
+            self.label = None
         else:
+            if isinstance(text, bool):
+                # convert to default value or empty string
+                text = "x = {:0.3f}, y = {:0.3f}" if text else ""
```
NilsNemitz, line 183: In a perfect world, this would do something smart to show e.g. three significant digits for any scale of the plot. But that goes under "future improvements".

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,352 @@
...
-            self.label.setText(label)
+                # add a label
+                labelOpts = {} if labelOpts is None else labelOpts
+                self.label = TargetLabel(self, text=text, **labelOpts)
+            else:
+                # update the label text
+                self.label.setText(text)
             self._updateLabel()
 
     def setLabelAngle(self, angle):
```
NilsNemitz, line 193: I am not sure if this is needed, since the angle might be regarded as a property of TargetLabel now, rather than the TargetItem. If it is included, maybe have a setLabelOffset, too?

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,352 @@
...
+            self.prepareGeometryChange()
+        return self._shape
+
+    def generateShape(self):
+        dt = self.deviceTransform()
+        if dt is None:
+            self._shape = self._path
+            return None
+        v = dt.map(QtCore.QPointF(1, 0)) - dt.map(QtCore.QPointF(0, 0))
+        va = atan2(v.y(), v.x())
```
NilsNemitz, line 231: I do not understand what this set of calculations does. The fact that is uses both ArcTan and inverse transforms seems a little suspicious to me. This is the only occurrrence of atan across the entire codebase. :)

NilsNemitz (commented):
Looks nice and is fun to play with. I couldn't get the text to change through setText. Otherwise only minor comments.
I made a multiline label 'rt={:.2f}\nup={:.2f}'. That worked nicely, but made me wish for an easy way to switch to a fixed width font.

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: you're absolutely right; all instances in docs or variable names have been updated 👍

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: honestly this bit was inherited from the original PR; i probably should sketch out the math a bit more closely

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,352 @@
...
+        if self.label is None:
+            return
+        self.label.valueChanged()
+
+
+class TargetLabel(TextItem):
+    """A TextItem that attaches itself to a CursorItem.
+
+    This class extends TextItem with the following features :
+    * Automatically positions adjacent to the cursor at a fixed position.
```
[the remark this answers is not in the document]
j9ac9k: The label should remain in the same place in the coordinate space, but remain the same size in pixels regardless of zoom.

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,352 @@
...
+        elif isinstance(offset, (QtCore.QPoint, QtCore.QPointF)):
+            self.setPos(offset)
+        else:
+            raise TypeError("Offset parameter is the wrong data type")
+        self.target.sigPositionChanged.connect(self.valueChanged)
+        self.valueChanged()
 
+    def valueChanged(self):
+        x, y = self.target.position()
+        if isinstance(self.format, str):
```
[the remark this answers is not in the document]
j9ac9k: ooo ...this is a good idea; ... i'm not sure how to programmatically detect if a string is a string that can accept a format parameters, or if it just a fixed value... i should look into that.

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: hmm.... there is some si unit handling used in the spinbox code if I remember right, it's worthy of a look

pyqtgraph/graphicsItems/TargetItem.py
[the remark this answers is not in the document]
j9ac9k: hmm...that's actually a good point, the offset should be in screen coordinates ...

j9ac9k:
Thanks for the feedback @NilsNemitz I've incorporated some of the suggested changes; I do think the rest should be incorporated as well, it just might take me a few days to get to it. 👍

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: Thanks for testing, was able to replicate each issue; fixed with e3d9c078ce871d25a40567353e44dc920a9e93d4

pyqtgraph/graphicsItems/TargetItem.py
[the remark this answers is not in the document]
j9ac9k: Implemented with bab7bbc2a76373760ece37a1f5ca16cad534c4e9

pyqtgraph/graphicsItems/TargetItem.py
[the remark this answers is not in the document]
j9ac9k: Likely can achieve this with `string.Formatter.parse`  https://docs.python.org/3/library/string.html#string.Formatter.parse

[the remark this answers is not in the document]
j9ac9k: Implemented with 489a2af6d342dd5b7dfc5d90a68c874598cf5af8

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: Implemented with 9f182de53197241fcbe09ad873175d9a917d48c1

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: I re-arranged the device transformation math in hopes to give it a bit more context.

```python
        tr = QtGui.QTransform()
        tr.scale(self.scale, self.scale)
        
        devPos = dt.map(QtCore.QPointF(0, 0))
        tr.translate(devPos.x(), devPos.y())
        
        v = dt.map(QtCore.QPointF(1, 0)) - dt.map(QtCore.QPointF(0, 0))
        va = atan2(v.y(), v.x())
        tr.rotate(va * 180.0 / pi)
        
        dti = fn.invertQTransform(dt)
        return dti.map(tr.map(self._path))
```

I'm interpreting this to handle the use-case of the coordinate system being rotated with respect to the screen.  I'm not sure pyqtgraph even supports that functionality!

j9ac9k: also we do use `np.arctan2` in other parts of the library for this same bit in `ROI.py`.

examples/InfiniteLine.py
```diff
@@ -32,6 +32,11 @@
 p1.addItem(inf2)
 p1.addItem(inf3)
 
+targetItem1 = pg.TargetItem(label=True, symbol="crosshair", labelOpts={"angle": 0, "offset": QtCore.QPointF(2., 2.)})
+targetItem2 = pg.TargetItem(pos=(30, 5), diameter=20, label="Second Item", symbol="star", labelOpts={"angle": 45, "offset": QtCore.QPointF(2., 0)})
```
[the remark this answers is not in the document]
j9ac9k: Incorporated with e699d9d08c35efef7a2cf6f32d93a85399a0db8f

Thanks for the suggestion.

j9ac9k:
Functionally I think this is good, but I think some of the API methods and interaction between TargetItem and TargetLabel need some cleaning up.

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,382 @@
...
-                self.label = TextItem()
-                self.label.setParentItem(self)
-            self.label.setText(label)
+                # add a label
+                labelOpts = {} if labelOpts is None else labelOpts
+                self.label = TargetLabel(self, text=text, **labelOpts)
+            else:
+                # update the label text
+                self.label.format = text
+                self.label.valueChanged()
```
NilsNemitz, line 194: I think this does not do anything when the new label is not "formattableText".
I suspect that changing the text to something non-formattable after the initial draw might not work?

Not sure how to easily setup a test for that. Changing the text before the initial draw works fine.
Maybe add a `force_update` parameter to the valueChanged call? Or add a separate `textChanged()` function?

NilsNemitz (commented):
This seems quite functional now, I only have some minor questions on how and if updates to the label work after it has already been drawn.

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,382 @@
...
+            return None
+        v = dt.map(QtCore.QPointF(1, 0)) - dt.map(QtCore.QPointF(0, 0))
+        dti = fn.invertQTransform(dt)
+        devPos = dt.map(QtCore.QPointF(0, 0))
+        tr = QtGui.QTransform()
+        tr.translate(devPos.x(), devPos.y())
+        va = atan2(v.y(), v.x())
+        tr.rotate(va * 180.0 / pi)
+        tr.scale(self.scale, self.scale)
+        return dti.map(tr.map(self._path))
```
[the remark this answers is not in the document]
j9ac9k: no, but we have an analogous method to it in `ROI.py`

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ROI.py#L1493-L1509

This was what the PR originally looked like, but I modified it to use the `math` module instead of numpy, as that is significantly faster when dealing with single values.

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,382 @@
...
+            self.setText(text)
+            self.setProperty("formattableText", False)
+        self._format = text
+
+    def valueChanged(self):
+        x, y = self.target.position()
+        if isinstance(self.format, str) and self.property("formattableText"):
+            self.setText(self.format.format(float(x), float(y)))
+        elif callable(self.format):
+            self.setText(self.format(x, y))
```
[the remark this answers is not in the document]
j9ac9k: in this case.`self.format == self._format`.

for the property bit, I'm mostly interested in the setter, where I'm trying to have things automatically update when some kind of operation like `self.format = ...`  occurs.

I have no problem changing that to `self._format` for clarify/slight-faster lookup.

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: oh right, yeah this only will do something if the string is either formattable or callable.

Changing the text after the fact is fine due to how I have the format property presently,

> Or add a separate textChanged() function?

This is effectively why I'm using the python property/setter, when I set `label.format =  text` it's like calling `label.format(text)` which checks to see if the new text is format-able or not, and sets the `formattableText` property to True/False accordingly.

j9ac9k:
I should add that python properties like this are a bit of an anti-pattern with the Qt framework, and that's what largely bugs me about this as it stands right now... but that's why it's not merged yet 😆

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,382 @@
...
             if self.label is None:
-                self.label = TextItem()
-                self.label.setParentItem(self)
-            self.label.setText(label)
+                # add a label
+                labelOpts = {} if labelOpts is None else labelOpts
+                self.label = TargetLabel(self, text=text, **labelOpts)
+            else:
+                # update the label text
+                self.label.format = text
```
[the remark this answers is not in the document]
j9ac9k: A lot of those `kwargs` work because it's during the `TextItem` initializer, they all don't map nicely to being updated; for that you would nee do call the appropriate setters; I'm not seeing a good way that `.setLabel()` can try and pass those along in a way they can be effective.  Do you happen to have a way in mind?

j9ac9k:
I cleaned up the API some more, added a callable example.  I'm not sure how to pass `**labelOpts` after `TargetItem` has been created; but I do thing that `setLabelAngle` should likely be done away with (and have functionality merge with `setLabel`).

IOW, not quite ready to merge.

j9ac9k:
Actually as I'm trying to get the `setAngle` functionality into `setLabel` this is looking like it might be brutal.  For example, what if a user leaves the default, `text=None` but `angle is not None`; does the user intent to adjust the angle or does the user intend to remove the label?

That said, why is `angle` treated any differently from any of the other `labelOpts`... probably don't need `setLabelAngle` to begin with at all, but I should give a method to easily get the `TargetLabel` so that the user can easily call the various `TextItem` methods on it.

j9ac9k:
ok, @NilsNemitz I think this is ready for another look-over whenever you have some time.

j9ac9k:
Docs now updated, ... Will see if readthedocs will let me peek at what the TargetItem docs look like

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,429 @@
...
+
+            # provide default text if text is True
+            if isinstance(text, bool):
+                # convert to default value or empty string
+                text = "x = {: .3n}\ny = {: .3n}" if text else ""
+
+            # if there is no label, add a label
+            if self._label is None:
+                labelOpts = {} if labelOpts is None else labelOpts
+                self._label = TargetLabel(self, text=text, **labelOpts)
```
outofculture, line 327: I pared down one of the example uses to a minimal `pg.TargetItem()`, and it blows up here:
```
  File "examples/InfiniteLine.py", line 35, in <module>
    targetItem1 = pg.TargetItem()
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/TargetItem.py", line 120, in __init__
    self.setLabel(label, labelOpts)
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/TargetItem.py", line 327, in setLabel
    self._label = TargetLabel(self, text=text, **labelOpts)
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/TargetItem.py", line 375, in __init__
    self.setFormat(text)
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/TargetItem.py", line 404, in setFormat
    if parsed[0][1] is not None:
```

Maybe this method needs another top-level if case for `text is None and labelOpts is None`? We could also fix it inside of TargetLabel, and in fact we might want both.

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,429 @@
...
+
+    def pos(self):
+        """Provides the current position of the TargetItem
+
+        Returns
+        -------
+        tuple
+            ``(x, y)`` position of the TargetItem
+        """
+        return self._pos
```
outofculture, line 281: This should probably return a `Point`. Most other `pos` methods I found do as much, and when dragging, I get this error:
```
    |    File "examples/InfiniteLine.py", line 86, in <module>
    |      QtGui.QApplication.instance().exec_()
    |    File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/widgets/GraphicsView.py", line 372, in mouseMoveEvent
    |      super().mouseMoveEvent(ev)
    |    File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/GraphicsScene/GraphicsScene.py", line 218, in mouseMoveEvent
    |      if self.sendDragEvent(ev, init=init):
    |    File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/GraphicsScene/GraphicsScene.py", line 343, in sendDragEvent
    |      debug.printExc("Error sending drag event:")
    |    --- exception caught here ---
    |    File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/GraphicsScene/GraphicsScene.py", line 341, in sendDragEvent
    |      item.mouseDragEvent(event)
    |    File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/TargetItem.py", line 231, in mouseDragEvent
    |      self.symbolOffset = self.pos() - self.mapToParent(ev.buttonDownPos())
    |  TypeError: unsupported operand type(s) for -: 'tuple' and 'QPointF'
```

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,429 @@
...
+            raise TypeError
+        if self._pos != newPos:
+            self._pos = newPos
+            super().setPos(Point(self._pos))
+            self.sigPositionChanged.emit(self)
 
-    def setLabelAngle(self, angle):
-        if self.labelAngle != angle:
-            self.labelAngle = angle
-            self._updateLabel()
```
outofculture, line 44: This method didn't get re-implemented in the new code. It might be enough to pass through `angle` to the `self._label.setAngle`, though the previous code did more with it than that, so I'm not sure.

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,429 @@
...
     """
-    sigDragged = QtCore.Signal(object)
 
-    def __init__(self, movable=True, radii=(5, 10, 10), pen=(255, 255, 0), brush=(0, 0, 255, 100)):
-        GraphicsObject.__init__(self)
-        self._bounds = None
-        self._radii = radii
-        self._picture = None
+    sigPositionChanged = QtCore.Signal(object)
+    sigPositionChangeFinished = QtCore.Signal(object)
```
outofculture, line 22: These are well named signals! Do we need to do a deprecation dance with `sigDragged`?

pyqtgraph/graphicsItems/TargetItem.py
outofculture: I should rephrase: I tried the code I proposed, and things _looked like_ they worked, so it wouldn't surprise me if that implementation is sufficient.

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,429 @@
...
+            # if there is no label, add a label
+            if self._label is None:
+                labelOpts = {} if labelOpts is None else labelOpts
+                self._label = TargetLabel(self, text=text, **labelOpts)
+
+            # if there is already a label, update the text formatting
+            else:
+                # update the label text
+                self._label.setFormat(text)
+                self._label.valueChanged()
```
outofculture, line 333: `TargetLabel.setFormat` already calls `valueChanged()` on itself.

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,429 @@
...
+
+            # if there is no label, add a label
+            if self._label is None:
+                labelOpts = {} if labelOpts is None else labelOpts
+                self._label = TargetLabel(self, text=text, **labelOpts)
+
+            # if there is already a label, update the text formatting
+            else:
+                # update the label text
+                self._label.setFormat(text)
```
outofculture, line 332: We maybe also need to do something with the `labelOpts`.

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -1,132 +1,429 @@
...
+            else:
+                self.setText(text)
+                self.setProperty("formattableText", False)
+        else:
+            self.setProperty("formattableText", False)
+        self._format = text
+        self.valueChanged()
+
+    def valueChanged(self):
+        x, y = self.target.position()
```
outofculture, line 415: This should probably be `.pos()`:
```
  File "examples/InfiniteLine.py", line 45, in <module>
    "offset": QtCore.QPoint(15, 15)
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/TargetItem.py", line 120, in __init__
    self.setLabel(label, labelOpts)
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/TargetItem.py", line 333, in setLabel
    self._label = TargetLabel(self, text=text, **labelOpts)
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/TargetItem.py", line 385, in __init__
    self.setFormat(text)
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/TargetItem.py", line 423, in setFormat
    self.valueChanged()
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/TargetItem.py", line 427, in valueChanged
    x, y = self.target.position()
AttributeError: 'TargetItem' object has no attribute 'position'
```

outofculture:
UX-wise, I was a bit surprised that the labels were not interactive ( as a drag handle for the target itself, or maybe just to change the label angle ). I don't think this PR needs to address this issue, though.

pyqtgraph/graphicsItems/TargetItem.py
outofculture: Updating the label options isn't on its own already supported by the labels. We could destroy the current label and make a new one, as a cheap implementation

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: :facepalm: i just renamed that method, of course you would think I would check for any references to said method...

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: Ahh good point, should have checked what similar methods do elsewhere in the library.

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: I admit, I had to lookup `sigDragged`, didn't know of that signal off the top of my head.  I'm not opposed to that; it arguably makes a bit more sense with `ROI.py` ...but for InfiniteLine, I would think `sigPositionChanged` would be sufficient (and frankly, I think `positionChanged` would be even better as `positionChanged` is the name of a signal is some upstream objects that are more or less equivalent.

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: The reason it works is my version I leverage the use of `TextItem.anchor`, so setting the angle rotates the label around the anchor point.  I think the original version didn't use `TextItem.anchor` and likely had to do more re-implementation to get the desired behavior that way.  (I'm going off of memory here, but I think that's why it "just works" without the use of this method).

pyqtgraph/graphicsItems/TargetItem.py
outofculture: Oh, sorry, I was confusing.

I mean that the previous code used `sigDragged` instead of `sigPositionChanged`, and we should keep the previous interface around for at least one version, but issue a deprecation warning. Same could be applied to `setLabelAngle`.

outofculture:
> I'm trying to think, what should dragging the label here do, should that move the target item, should that move the label's position relative to the target item (and if we do that, do we provide a "reset" mechanism?"

I would pick "move the target itself" as the least surprising behavior ( subject to `moveable`, of course ). Yeah, as your question implies, the issues with "moving the label separate from the target" make that behavior a poor choice.

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: Ahhh right I keep forgetting that this PR replaces something, even though it was completely undocumented.

I'll go back in and insert the original methods and throw some deprecation warnings.

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: I made a modification where a text label is not generated if test resolves `False`, so that should cover, `""`, `None` and `False`.

In `TargetLabel`, instead of defaulting to `None`, I think I'll default to the empty string, just like `TextItem`

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: re-creation a TargetLabel here would actually be clever; as trying to support the mapping of all the individual options supported would be pretty tricky, and I'll likely screw up something.

If a user wanted to preserve the original item, they could likely just call `tragetItem.label().<whatevermethod>`

I think that's what I'll do...

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: I think this is resolved sufficiently; I now check that the `parsed` list at least exists; which should be sufficient for our use-case.

j9ac9k:
I think I've addressed all but two areas of concerns

1. Deprecations (even though this object wasn't documented, it did exist, so I should at get older args/methods in-place/functional again and slap deprecation warnings)
2. Handle the case of clicking + dragging the TargetLabel

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: Fixed with 5ae0175cbf0ee0622d58cd72826b6358077c30c6

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: Fixed with 03394567e87d77e4f2308ce2e12aeac876db9fd8

j9ac9k:
@outofculture @NilsNemitz I think this guy is ready again.

Works as it did before; added deprecation warnings for `radii is not None`, use of `sigDragged` and  `setLabelAngle`.  

clicking + dragging by the label works as well (I used that all of 1 time before I loved the suggestion, thanks for that @outofculture )

Also added benefit of it won't explode if you pass no arguments to it.

PR #1319, opened 2020-07-17
Acq4 merge

campagnola:
- fix reload on py3 (again) -- don't automatically reload modules without pyc
- clean up exception messages in console -- don't raise from eval syntax error
- make ptime.time on py3 return precision wall-clock time


j9ac9k:
LGTM, merging!

PR #1321, opened 2020-07-19
Revert plotcurveitem changes from 18999c

campagnola:
The commit 18999c (https://github.com/pyqtgraph/pyqtgraph/commit/18999c) says it's just simplifying code, but it changes the API and breaks some use cases.

Example, generates a stream of error messages:

    import pyqtgraph as pg
    p = pg.plot()
    c = pg.PlotCurveItem([1,2,3], pen=None)
    p.addItem(c)


campagnola:
@j9ac9k @fake-name this bug was introduced in #816. My revert takes care of the errors I was getting, but I did not check to see whether this also breaks the intended performance improvements from 816. Some more review needed before merge..

Issue #812, opened 2019-01-16
multiplotwidget curve color with pen


ghost:
When using multiplotwidget, it is not possible to pass a pen to the plot method.
This crashes:
pw = MultiPlotWidget()
pw.plot(metaarray,, pen=pg.mkPen('b', width=0.00001))

So there is no way to customize the color of the curves.

angulartist:
https://github.com/pyqtgraph/pyqtgraph/pull/1324

ixjlyons:
Implemented with #1324

PR #1324, opened 2020-07-19
Enhancement: [Issue/812]: MultiPlotItem handles pen (QPen) as a keywo…

angulartist:
In response to issue : https://github.com/pyqtgraph/pyqtgraph/issues/812

This PR:
- allows MultiPlotItem .plot() method to handles kwargs
- allows the usage of a QPen as an argument to customize curves

### DIFF
--- a/examples/MultiPlotWidget.py
+++ b/examples/MultiPlotWidget.py
@@ -30,7 +30,7 @@ ma = MetaArray(data, info=[
         ]}, 
     {'name': 'Time', 'values': linspace(0., 1., 1000), 'units': 's'}
     ])
-pw.plot(ma)
+pw.plot(ma, pen='y')
 
 ## Start Qt event loop unless running in interactive mode.
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


angulartist:
> Idea seems good to me, but I think the pen value should go through `mkPen` so you can just write `multiplot.plot(data, pen='r')` or whatever.
> 
> On second thought, I think it makes the most sense to simply let you pass anything you want to each underlying `PlotItem.plot` call. Naming it `**plotArgs` and just passing it along untouched should be pretty clear I would think.

I updated my PR.
We can call plot() like that **pi.plot(metaarray..., pen='y')**

examples/MultiPlotWidget.py
```diff
@@ -30,7 +30,7 @@
         ]}, 
     {'name': 'Time', 'values': linspace(0., 1., 1000), 'units': 's'}
     ])
-pw.plot(ma)
+pw.plot(ma, pen='y')
```
ixjlyons, line 33: Great, I was going to suggest doing something like this.

Issue #986, opened 2019-07-05
Parameter nudge of AxisItem.resizeEvent() is not passable for axis label adjustment


Maesiu:
```python
    def resizeEvent(self, ev=None):
        #s = self.size()
        
        ## Set the position of the label
        nudge = 5
        br = self.label.boundingRect()
        p = QtCore.QPointF(0, 0)
        if self.orientation == 'left':
            p.setY(int(self.size().height()/2 + br.width()/2))
            p.setX(-nudge)
        elif self.orientation == 'right':
            p.setY(int(self.size().height()/2 + br.width()/2))
            p.setX(int(self.size().width()-br.height()+nudge))
        elif self.orientation == 'top':
            p.setY(-nudge)
            p.setX(int(self.size().width()/2. - br.width()/2.))
        elif self.orientation == 'bottom':
            p.setX(int(self.size().width()/2. - br.width()/2.))
            p.setY(int(self.size().height()-br.height()+nudge))
        self.label.setPos(p)
        self.picture = None
```
It would be nice to have access to `nudge` and thereby be able to adjust the axis label offset.

Thanks.

2xB:
This sounds reasonable to me. It would make sense to expose `nudge` via `AxisItem.setStyle`.

PR #1327, opened 2020-07-20
Enhancement: [Issue/986]: Expose param "nudge" via AxisItem.setStyle()

pompierninja:
Ref issue: https://github.com/pyqtgraph/pyqtgraph/issues/986

This PR:
- exposes the parameter "nudge" via AxisItem .setStyle() method so you can adjust axis label offset (was fixed to 5, and now defaults to 5)
- updates DateAxisItem example

### DIFF
--- a/examples/DateAxisItem.py
+++ b/examples/DateAxisItem.py
@@ -17,7 +17,7 @@ app = QtGui.QApplication([])
 # Create a plot with a date-time axis
 bax = pg.DateAxisItem()
 bax.setLabel('Usage of DateAxisItem')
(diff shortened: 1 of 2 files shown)


ixjlyons:
Sorry for the delayed review.

Sounds like a good idea, though I'm not sure "nudge" is the right name for this. It's tolerable as a variable name in the code with the surrounding context and usage, but I think something like "labelOffset" might be clearer. Open to other ideas.

Also, I think this should probably be accounted for in `boundingRect()`. For example, if you set it to 20 in your example, the label disappears.

PR #1331, opened 2020-07-27
sip: switch to new PyQt5 api import, with fallback to the old impleme…

LocutusOfBorg:
…ntation if not found.

Also calls to setapi are not needed anymore, so use them only with the old implementation.

See: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=966045
for reference

### DIFF
--- a/pyqtgraph/multiprocess/bootstrap.py
+++ b/pyqtgraph/multiprocess/bootstrap.py
@@ -25,9 +25,12 @@ if __name__ == '__main__':
 
     pyqtapis = opts.pop('pyqtapis', None)
     if pyqtapis is not None:
-        import sip
-        for k,v in pyqtapis.items():
-            sip.setapi(k, v)
+        try:
+            from PyQt5 import sip
+        except ImportError:
+            import sip
+            for k,v in pyqtapis.items():
+                sip.setapi(k, v)
         
     qt_lib = opts.pop('qt_lib', None)
     if qt_lib == 'PySide':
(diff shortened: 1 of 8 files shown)


pyqtgraph/multiprocess/bootstrap.py
```diff
@@ -25,9 +25,12 @@
...
     pyqtapis = opts.pop('pyqtapis', None)
     if pyqtapis is not None:
-        import sip
-        for k,v in pyqtapis.items():
-            sip.setapi(k, v)
+        try:
+            from PyQt5 import sip
+        except ImportError:
+            import sip
+            for k,v in pyqtapis.items():
```
j9ac9k, line 32: Should this go in a "finally" clause?

j9ac9k:
Hi @LocutusOfBorg thanks for the PR (fantastic user name btw). Thanks for the PR!

I made one comment, there is a `for` loop immediately following the import, I suspect that for loop should go either in the `finally` clause for the `try`/`except` or it should be un-indented.

pyqtgraph/multiprocess/bootstrap.py
LocutusOfBorg: Hello, according to the Debian maintainer:
"Also, calls like "sip.setapi('QDate', 2)" are not needed at all with PyQt5.
They were needed only with PyQt4 and Python 2. It is safe to delete them."

This is why I'm looping on the "ImportError" case, because its where the old api is being used.

LocutusOfBorg: BTW I'm not a Python-savvy man, so please double check my statement :D

ixjlyons:
Just want to point out there's a note here that the bare `import sip` will continue to work as long as another PyQt5 module is imported first (https://www.riverbankcomputing.com/static/Docs/PyQt5/incompatibilities.html#importing-the-sip-module), so I think that's why we haven't seen any test failures from this. We should still migrate to the new design of course.

Looks good to me. Thanks @LocutusOfBorg

PR #1333, opened 2020-07-27
BarGraphItem: add missing methods

termim:
Make BarGraphItem's interface closer to the GraphItem's one.
Add setPen, setBrush and  dataBounds methods, factor out setData method.

### DIFF
--- a/pyqtgraph/graphicsItems/BarGraphItem.py
+++ b/pyqtgraph/graphicsItems/BarGraphItem.py
@@ -12,34 +12,28 @@ class BarGraphItem(GraphicsObject):
         """
         Valid keyword options are:
         x, x0, x1, y, y0, y1, width, height, pen, brush
-        
+
         x specifies the x-position of the center of the bar.
         x0, x1 specify left and right edges of the bar, respectively.
         width specifies distance from x0 to x1.
         You may specify any combination:
-            
+
             x, width
             x0, width
             x1, width
             x0, x1
-            
-        Likewise y, y0, y1, and height. 
+
+        Likewise y, y0, y1, and height.
         If only height is specified, then y0 will be set to 0
-        
+
         Example uses:
-        
+
             BarGraphItem(x=range(5), height=[1,5,2,4,3], width=0.5)
-            
-        
+
+
         """
         GraphicsObject.__init__(self)
         self.opts = dict(
-            x=None,
-            y=None,
-            x0=None,
-            y0=None,
-            x1=None,
-            y1=None,
             name=None,
             height=None,
             width=None,
@@ -51,8 +45,9 @@ class BarGraphItem(GraphicsObject):
         self._shape = None
         self.picture = None
         self.setOpts(**opts)
-        
+
     def setOpts(self, **opts):
+        self.setData(**opts)
         self.opts.update(opts)
         self.picture = None
         self._shape = None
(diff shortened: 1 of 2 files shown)


pyqtgraph/graphicsItems/BarGraphItem.py
```diff
@@ -58,6 +58,16 @@ def setOpts(self, **opts):
         self.update()
         self.informViewBoundsChanged()
         
+    def setPen(self, *args, **kargs):
+        """Set the pen used to draw the curve."""
+        self.setOpts(pen = fn.mkPen(*args, **kargs))
```
ixjlyons, line 63: Nitpick: remove the spaces around the `=`, also in `setBrush`

pyqtgraph/graphicsItems/BarGraphItem.py
```diff
@@ -58,6 +58,16 @@ def setOpts(self, **opts):
...
+
+
```
ixjlyons, line 65: Nitpick: can remove this blank like so there's only one between methods. Also after `setBrush`

ixjlyons (changes requested):
Two minor comments, otherwise looks good to me.

termim:
Done

ixjlyons:
This is kind of strange - the diff I originally reviewed only had a couple changes (seemingly the first commit only, adding `setPen` and `setBrush`), but this now looks quite a bit more significant. I think I see what the thought process was via the individual commits, so I'll try to re-review soon.

termim:
I'm sorry I've messed up something while reacting on your review :(. I'll resubmit soon.

dgoeries:
Would be great if you can put with tests some solid ground for other developers.

pyqtgraph/graphicsItems/BarGraphItem.py
```diff
@@ -58,93 +60,103 @@ def setOpts(self, **opts):
...
         
         def asarray(x):
             if x is None or np.isscalar(x) or isinstance(x, np.ndarray):
                 return x
             return np.array(x)
 
+        x = asarray(kargs.pop('x', None))
+        self.x0 = asarray(kargs.pop('x0', None))
+        x1 = asarray(kargs.pop('x1', None))
+        self.width = asarray(kargs.pop('width', None))
```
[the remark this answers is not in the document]
termim: - setOpts (and self.opts) are used in pyqtgraph for options not data most of the times.
- documentation for BarGraphItem explicitly says that any combination of x, x0, x1, y, y0, y1, width and height is allowed, so I would not guess about what one typycally sets. 
- getData was broken before (it could return unset data), fixed in the last update.

pyqtgraph/graphicsItems/BarGraphItem.py
```diff
@@ -50,119 +44,131 @@ def __init__(self, **opts):
...
+
+    def setPen(self, *args, **kargs):
+        """Set the pen used to draw the curve."""
+        self.setOpts(pen=fn.mkPen(*args, **kargs))
+
+    def setBrush(self, *args, **kargs):
+        """Set the brush used when filling the area under the curve"""
+        self.setOpts(brush=fn.mkBrush(*args, **kargs))
+
+    def setData(self, **kargs):
```
j9ac9k, line 66: There should be a docstring here with what the expected/supported arguments here are

j9ac9k:
Hi @termim 

Thanks for this PR; I've been toying with this a bit with a slight modifications to `examples/BarGraphItem.py`.   I think updating that example to reflect the use of `BarGraphItem.setData()` would be worthwhile.  I also added an inline comment about adding a docstring for accepted `**kargs` parameters that you can pass in there, I think that would be helpful.

Everything appeared to work for me just fine.  Thanks for the PR.  Sorry it's taken us a while to follow up.

pyqtgraph/graphicsItems/BarGraphItem.py
termim: Done

termim:
@j9ac9k, added docstring to setData and example of it's usage.
Also updated setData to make sure plot is updated.

j9ac9k:
Closing/Re-Opening to trigger new CI runs...

pyqtgraph/graphicsItems/BarGraphItem.py
```diff
@@ -178,4 +212,48 @@ def name(self):
         return self.opts.get('name', None)
 
     def getData(self):
-        return self.opts.get('x'),  self.opts.get('height')
+        return self.x0, self.width, self.y0, self.height
+
+    def dataBounds(self, ax, frac=1.0, orthoRange=None):
```
ixjlyons, line 217: This implementation doesn't look quite right. For example, in the `ax == 0` case, the bar width isn't taken into account.

I think both `ax == 0` and `ax == 1` cases should look basically identical with x/y swapped, and both need to take into account horizontal and vertical bar orientations.

ixjlyons (changes requested):
It took a little while to realize this is mostly just splitting some stuff to a `setData` method and adding `dataBounds` (edit: as stated in the top comment, doh!). Anyway, in the interest of getting through the PR backlog, I'd be in favor of removing this `dataBounds` implementation for now and creating an enhancement issue with it as a starting point.

pyqtgraph/graphicsItems/BarGraphItem.py
```diff
@@ -50,119 +44,159 @@ def __init__(self, **opts):
         )
         self._shape = None
         self.picture = None
+        self.bounds = [None, None]  ## caches data bounds
+        self.setData(**opts)
```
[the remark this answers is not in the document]
j9ac9k: I'm indifferent regarding passing `**opts` from `__init__` to `setData`, but I do agree that we should follow the pattern of `PlotDataItem`

pyqtgraph/graphicsItems/BarGraphItem.py
j9ac9k: Didn't even consider this; but this is right, BarGraphItem should be able to handle horizontal bars or vertical bars.

PR #1334, opened 2020-07-27
More data validity checks in PlotCurveItem

termim:
More data validity checks in PlotCurveItem

### DIFF
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -162,6 +162,8 @@ class PlotCurveItem(GraphicsObject):
             # include a percentile of data range
             mask = np.isfinite(d)
             d = d[mask]
+            if len(d) == 0:
+                return (None, None)
             b = np.percentile(d, [50 * (1 - frac), 50 * (1 + frac)])
 
         ## adjust for fill level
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -194,8 +196,10 @@ def pixelPadding(self):
     def boundingRect(self):
         if self._boundingRect is None:
             (xmn, xmx) = self.dataBounds(ax=0)
+            if xmn is None or xmx is None:
+                return QtCore.QRectF()
             (ymn, ymx) = self.dataBounds(ax=1)
-            if xmn is None or ymn is None:
```
ixjlyons, line 198: It seems to me the `dataBounds` interface is such that either both members or neither member of the tuple can be `None`, so the original check appears to be sufficient. Have you run into a case where that's not true?

termim: Yes it did happen when I used `LinearRegionItem` to zoom into some data plot. It is possible that this was just user mistake and the GUI should check for that, but it is good to check anyway.
BTW the same checks are performed in `ScatterPlotItem.boundingRect`.
Also `dataBounds` interface is somewhat vague - some items can return None as well.

ixjlyons: Fair enough.

ixjlyons:
Thanks @termim

PR #1339, opened 2020-08-05
Vector init in qt5

outofculture:
`Vector.__init__` improvements

 * add docstring
 * fix handling of QVector3D args (cannot list() them)
 * refactor to no longer need return statements

Also add Pa to the list of units.

### DIFF
--- a/pyqtgraph/Vector.py
+++ b/pyqtgraph/Vector.py
@@ -17,14 +17,14 @@ class Vector(QtGui.QVector3D):
 
         ==============  ================================================================================================
         **Arguments:**
-        *args*      Could be any of:
-                     * 3 numerics
-                     * 2 numerics (`0` assumed for z)
-                     * Either of the previous in a list-like collection
-                     * 1 QSizeF (`0` assumed for z)
-                     * 1 QPointF (`0` assumed for z)
-                     * Any other valid QVector3D init args.
-        ----------------------------------------------------------------------------------------------------------------
+        *args*          Could be any of:
+
+                         * 3 numerics (x, y, and z)
+                         * 2 numerics (x, y, and `0` assumed for z)
+                         * Either of the previous in a list-like collection
+                         * 1 QSizeF (`0` assumed for z)
+                         * 1 QPointF (`0` assumed for z)
+                         * Any other valid QVector3D init args.
         ==============  ================================================================================================
         """
         initArgs = args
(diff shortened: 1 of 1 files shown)


outofculture:
Rebased onto master; turns out the refactor had happened once before, but this one is more accurate.

pyqtgraph/Vector.py
```diff
@@ -12,35 +12,37 @@ class Vector(QtGui.QVector3D):
     """Extension of QVector3D which adds a few helpful methods."""
     
     def __init__(self, *args):
+        """
+        Handle additional constructions of a Vector
+
+        ==============  ================================================================================================
+        **Arguments:**
+        *args*      Could be any of:
```
ixjlyons, line 20: Vector isn't currently in the docs, but there should be few more spaces introduced before "Could". It should align with the `=` after the gap in the table borders. Please also add an empty line after this one - I'm not sure exactly why but RST/sphinx seems to prefer it this way for bulleted lists.

pyqtgraph/Vector.py
```diff
@@ -12,35 +12,37 @@ class Vector(QtGui.QVector3D):
...
+                     * 3 numerics
+                     * 2 numerics (`0` assumed for z)
+                     * Either of the previous in a list-like collection
+                     * 1 QSizeF (`0` assumed for z)
+                     * 1 QPointF (`0` assumed for z)
+                     * Any other valid QVector3D init args.
+        ----------------------------------------------------------------------------------------------------------------
```
ixjlyons, line 27: I think this line can be removed.

pyqtgraph/Vector.py
```diff
@@ -12,35 +12,37 @@ class Vector(QtGui.QVector3D):
     """Extension of QVector3D which adds a few helpful methods."""
     
     def __init__(self, *args):
+        """
+        Handle additional constructions of a Vector
+
+        ==============  ================================================================================================
+        **Arguments:**
+        *args*      Could be any of:
+                     * 3 numerics
```
ixjlyons, line 21: It's probably obvious, but maybe add a parenthetical `(x, y, and z)` at the end of this line.

pyqtgraph/Vector.py
outofculture: Thanks; I was bumbling about for the rules there.

outofculture:
I've acted on all the comments.

ixjlyons:
Looks good. Thanks @outofculture

PR #1344, opened 2020-08-11
rename SignalProxy.block

outofculture:
`block` is being overwritten with a boolean by QObject in PyQt5. This method was added recently enough that we should be able to rename it without breaking anyone's code (exception: ACQ4 will have a complimentary PR).


outofculture:
Ooops. This was due to a stale branch of pyqtgraph on my test machine. Disregard.

PR #1349, opened 2020-08-24
Fixed bug where removeTempArea would be called on home even if it was not defined

MitchelPaulin:
Ran into an issue where after poping a graph from a dock area I would get an error about calling removeTempArea on None. This seems to have fixed the problem.

### DIFF
--- a/pyqtgraph/dockarea/DockArea.py
+++ b/pyqtgraph/dockarea/DockArea.py
@@ -319,7 +319,7 @@ class DockArea(Container, QtGui.QWidget, DockDrop):
         #print "apoptose area:", self.temporary, self.topContainer, self.topContainer.count()
         if self.topContainer is None or self.topContainer.count() == 0:
             self.topContainer = None
-            if self.temporary:
+            if self.temporary and self.home:
                 self.home.removeTempArea(self)
                 #self.close()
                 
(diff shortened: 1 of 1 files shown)


j9ac9k:
I really should sync up develop branch w/ master, ... as the current state of develop is rather broken.  I should also update the contributing guide to suggest creating PRs to merge with master.  Anyway let me fix the target branch and let CI rerun and do its thing...

MitchelPaulin:
@j9ac9k Everything look good?

j9ac9k:
Sorry I'm a bit off the grid as I'm actually moving.  I probably won't get to look at this for at least another week.  Paging @ixjlyons or @2xB  to take a look if they have time.

j9ac9k:
Hi @MitchelPaulin sorry it's taken me so long to follow up.  Moving takes forever... now that I'm taking a closer look at this, I have one suggestion, which I'll make in-line.

Thanks again for the PR and sorry for the delayed reply.

pyqtgraph/dockarea/DockArea.py
```diff
@@ -319,7 +319,7 @@ def apoptose(self, propagate=True):
         #print "apoptose area:", self.temporary, self.topContainer, self.topContainer.count()
         if self.topContainer is None or self.topContainer.count() == 0:
             self.topContainer = None
-            if self.temporary:
+            if self.temporary and self.home:
```
j9ac9k, line 322: would suggest having `and self.home is not None` if we're checking for `None` and not the boolean condition of `self.home`

j9ac9k:
LGTM!  Thanks for the change! Merging!

Issue #1320, opened 2020-07-18
Bug: ActionParameter.hide() causes rename to empty string


meganbkratz:
Calling hide() on an ActionParameter causes it to be renamed to an empty string. Thus, it's no longer accessible in code by it's original name, and if you access it by indexing there is no text on the button (consistent with the name becoming an empty string).

### Code to reproduce
```import pyqtgraph as pg
from pyqtgraph.Qt import QtCore, QtGui

app = QtGui.QApplication([])
import pyqtgraph.parametertree.parameterTypes as pTypes
from pyqtgraph.parametertree import Parameter, ParameterTree, ParameterItem, registerParameterType

params = [
    {'name':'action_test', 'type':'action'},
    {'name':'checkbox', 'type':'bool'},
    {'name':'subgroup', 'type':'group', 'children':[
        {'name':'action2', 'type':'action'},
        {'name':'string', 'type':'str', 'readonly':True, 'value':'A string?'}
        ]}
    ]
p = Parameter.create(name='params', type='group', children=params)

t = ParameterTree()
t.addParameters(p)

win = QtGui.QWidget()
layout = QtGui.QGridLayout()
win.setLayout(layout)
layout.addWidget(t)
win.show()

### trigger the bug:
p.child('action_test').hide()
p.child('action_test').show()
```

In commits where the bug exists the above script raises an Exception:

```Traceback (most recent call last):
  File "action_test.py", line 28, in <module>
    p.child('action_test').show()
  File "/Users/meganbkratz/Code/ai/opto/pyqtgraph/pyqtgraph/parametertree/Parameter.py", line 674, in child
    raise KeyError("Parameter %s has no child named %s" % (self.name(), names[0]))
KeyError: 'Parameter params has no child named action_test'
```

It should simply show the ParameterTree without raising an exception.

### Tested environment(s)

 * PyQtGraph version: 0.11.0
 * Qt Python binding: PyQt5 5.9.2 Qt 5.9.6
 * Python version: Python 3.6.9 |Anaconda
 * NumPy version: 1.17.3
 * Operating system: OSX 10.13.6
 * Installation method: git repository

### Additional context
I used git bisect to trace the introduction of this bug to commit 7672b5b72596687295488fd1b1164ece113c71db

meganbkratz:
This bug is fixed by PR #1350

PR #1350, opened 2020-08-25
Bugfix for issue #1320 : Fix for ActionParameterItem name changing to empty string when hidden

meganbkratz:
This is a fix for issue #1320


campagnola:
The fix here looks fragile to me--it relies on `ignoreNameColumnChange` always being True, which should not be relied on since that variable is being used for other purposes. There are a few other problems surrounding this issue as well, which should be addressed in #1356.

meganbkratz:
Thank you for figuring out how to fix this better!

PR #1360, opened 2020-09-08
Add "left" and "right" step Modes

cpascual:
stepMode is currently either `True` or `False`. If it is `True`,
it requires the user to make `len(x) = len(y)+1`. This is
inconvenient because it makes it difficult to change the
stepMode on a given curve (just as one would change, e.g.,
its color).

This commit extends the current situation by introducing
two more step modes: `"lstep"` and `"rstep"`, which do not require
passing an extra x value. In turn, this modes associate each
y value to either the left or the right boundary of the step.

For example, the `"rstep"` mode is handy when plotting "live"
digital signals in which x,y data pairs are appended as they
are read.

This commit does not modify the behavior in case of `stepMode=True`

### DIFF
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -62,7 +62,7 @@ class PlotCurveItem(GraphicsObject):
             'fillLevel': None,
             'fillOutline': False,
             'brush': None,
-            'stepMode': False,
+            'stepMode': None,
             'name': None,
             'antialias': getConfigOption('antialias'),
             'connect': 'all',
@@ -315,14 +315,17 @@ class PlotCurveItem(GraphicsObject):
                         by :func:`mkBrush <pyqtgraph.mkBrush>` is allowed.
         antialias       (bool) Whether to use antialiasing when drawing. This
                         is disabled by default because it decreases performance.
-        stepMode        If True, a step is drawn using the x values as
-                        boundaries and the given y values are associated to the
-                        mid-points between the boundaries of each step.
-                        This is commonly used when drawing histograms. Note that
-                        in this case, len(x) == len(y) + 1.
+        stepMode        (str or None) If "mid", a step is drawn using the x
+                        values as boundaries and the given y values are
+                        associated to the mid-points between the boundaries of
+                        each step. This is commonly used when drawing
+                        histograms. Note that in this case, len(x) == len(y) + 1
                         If "left" or "right", the step is drawn assuming that
                         the y value is associated to the left or right boundary,
                         respectively. In this case len(x) == len(y)
+                        If not passed or an empty string or None is passed, the
+                        step mode is not enabled.
+                        Passing True is a deprecated equivalent to "mid".
         connect         Argument specifying how vertexes should be connected
                         by line segments. Default is "all", indicating full
                         connection. "pairs" causes only even-numbered segments
@@ -384,10 +387,10 @@ class PlotCurveItem(GraphicsObject):
         if 'stepMode' in kargs:
             self.opts['stepMode'] = kargs['stepMode']
 
-        if self.opts['stepMode'] is True:   ## if mode strictly **is** True
+        if self.opts['stepMode'] in ("mid", True):  ## check against True for backwards compatibility
             if len(self.xData) != len(self.yData)+1:  ## allow difference of 1 for step mode plots
                 raise Exception("len(X) must be len(Y)+1 since stepMode=True (got %s and %s)" % (self.xData.shape, self.yData.shape))
-        else:  ## mode may be `False` or `left` or `right`
+        else:
             if self.xData.shape != self.yData.shape:  ## allow difference of 1 for step mode plots
                 raise Exception("X and Y arrays must be the same shape--got %s and %s." % (self.xData.shape, self.yData.shape))
 
@@ -428,7 +431,7 @@ class PlotCurveItem(GraphicsObject):
                 x2 = np.empty((len(x) + 1, 2), dtype=x.dtype)
                 x2[1:] = x[:, np.newaxis]
                 x2[0] = x2[1]
-            else:
+            else:  # stepMode is "mid" (or True)
                 x2 = np.empty((len(x),2), dtype=x.dtype)
                 x2[:] = x[:, np.newaxis]
             if self.opts['fillLevel'] is None:
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -388,6 +388,9 @@ class PlotCurveItem(GraphicsObject):
             self.opts['stepMode'] = kargs['stepMode']
 
         if self.opts['stepMode'] in ("center", True):  ## check against True for backwards compatibility
+            if self.opts['stepMode'] is True:
+                import warnings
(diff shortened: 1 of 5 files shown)


cpascual:
Here an example:

```python
import pyqtgraph as pg
import numpy as np

app = QtGui.QApplication([])
p = pg.PlotWidget()
p.addLegend()
x = y = np.arange(5)

p.plot(x=x, y=y, symbol="t", stepMode=False, name="False", pen="r")
p.plot(x=x, y=2+y[:-1], symbol="t", stepMode=True, name="True", pen="g")
p.plot(x=x, y=4+y, symbol="t", stepMode="rstep", name="rstep", pen="c")
p.plot(x=x, y=6+y, symbol="t", stepMode="lstep", name="rstep", pen="m")

p.show()
app.exec_()
```

cpascual:
Hi, sorry for this shameless "bumping", but... can someone (@campagnola ?) comment on whether this contribution may be considered for inclusion or if there is something I should do to improve it?

(I just want to know because if it is going to be rejected I should find another way to implement this in my own project)

ixjlyons (changes requested):
This seems like a nice feature to me, and the API seems good in particular keeping backward compatibility. I'm inclined to accept but thoughts from others would be great.

A couple notes:

Using "step" in the possible values seems redundant. Maybe `stepMode="left"` and `stepMode="right"`? Just a thought, I don't feel strongly about it.

I would actually emphasize documentation of `stepMode` in the `PlotDataItem` class, and you could either copy the text to `PlotCurveItem` or refer to `stepMode` in `PlotDataItem`. The reason is the scatter plot item may be impacted by the step mode as well as the curve item. Also, people are likely (I think) to land on the `PlotDataItem` reference page from `plot` -> `PlotItem.plot` -> `PlotDataItem.__init__` in the docs.

When `stepMode == True`, it's the *x* value that's taken as the midpoint between provided x values, not y (y values are used as given).

And a last nitpick, I think it'd be prudent to add a comment above the line in `PlotCurveItem.updateData` where it says `if self.opts['stepMode'] is True`. Someone in the future might be inclined to remove `is True`, but that's actually a useful distinction now that `stepMode` can be a string where that condition doesn't apply

cpascual:
Thanks a lot @ixjlyons for the review. I just pushed 4 commits which, I think, address all your comments. 

Please note that I included a "added in version 0.12.0" for the left and right step modes  in the `PlotDataItem` docs, but of course I am not sure if the specific version number is right. I did it in a separate commit so that it can easily be reverted.

pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -315,9 +315,14 @@ def setData(self, *args, **kargs):
                         by :func:`mkBrush <pyqtgraph.mkBrush>` is allowed.
         antialias       (bool) Whether to use antialiasing when drawing. This
                         is disabled by default because it decreases performance.
-        stepMode        If True, two orthogonal lines are drawn for each sample
-                        as steps. This is commonly used when drawing histograms.
-                        Note that in this case, len(x) == len(y) + 1
+        stepMode        If True, a step is drawn using the x values as
```
j9ac9k, line 318: Should this start with `(bool or str or None)` ?

j9ac9k:
I don't have anything to add other than a thank you for submitting this feature @cpascual, I appreciate you ensuring the addition to this feature does not impact the current implementation.  I must admit the API feels a bit weird that it is an `Optional[Union[bool, str]]` but I'm not sure of a better way of doing it (unless we introduced a `stepType` (or something analogous) kwarg, that defaults to `standard` which the existing behavior, and has `lstep` and `rstep` as potential options.

I'm just throwing this idea out there to see what folks think, I'm not requesting any changes be made.

ixjlyons:
Regarding the weird mix of types accepted (particularly evident with `is True` being meaningful in the code), I was thinking there could be a "mid" option so True could be more of a legacy option (with the possibility of being deprecated and removed over time).

cpascual:
I can do the "mid" option implementation.
I didn't do it originally because I was after the minimum possible changes in order to be super-safe in respecting backwards-compat. But I think it is quite safe to do that.

cpascual:
d8b71e5 adds the "mid" option and marks usage of booleans as a legacy API

cpascual:
I also added 3997532 to prevent that any string (e.g. a misspell like "letf") ends up interpreted as "mid".

dgoeries:
I very much like this feature. I only have a nitpick and would call it `center` instead of `mid`.

cpascual:
@dgoeries :
>  would call it center instead of mid.

17cab0f does it. 

I originally called it "mid" as per @ixjlyons suggestion. Since it is a matter of taste, I leave it up to the integrator to revert this last commit if he/she prefers "mid" over "center".

dgoeries:
thanks for the updates!

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -76,10 +76,19 @@ def __init__(self, *args, **kargs):
...
+                         each step. This is commonly used when drawing
+                         histograms. Note that in this case, len(x) == len(y) + 1
+                         If "left" or "right", the step is drawn assuming that
+                         the y value is associated to the left or right boundary,
+                         respectively. In this case len(x) == len(y)
+                         If not passed or an empty string or None is passed, the
+                         step mode is not enabled.
+                         Passing True is a deprecated equivalent to "center".
                          (added in version 0.9.9)
+                         (str modes added in version 0.12.0)
```
j9ac9k, line 91: I would go ahead and remove the `modes` added in version 0.12.0; we'll likely have this in 0.11.1....

pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -379,7 +387,7 @@ def updateData(self, *args, **kargs):
         if 'stepMode' in kargs:
             self.opts['stepMode'] = kargs['stepMode']
 
-        if self.opts['stepMode'] is True:
+        if self.opts['stepMode'] in ("center", True):  ## check against True for backwards compatibility
             if len(self.xData) != len(self.yData)+1:  ## allow difference of 1 for step mode plots
```
j9ac9k, line 391: I would also consider adding a `DeprecationWarning` if `stepMode is True`

j9ac9k:
hey @cpascual thanks for the updates, sorry it's taking me a while to follow up (I'm still unpacking from my move).  The changes you made look great; I'm going to toy with it a bit more later, but I do think a deprecation warning should be emitted if `stepMode=True` is passed into the `__init__` methods.

pyqtgraph/graphicsItems/PlotDataItem.py
cpascual: 73542aa does it

pyqtgraph/graphicsItems/PlotCurveItem.py
cpascual: 2335e29 does it.


Note 1: I did the check on the `PlotDataItem.setData()` and `PlotCurveItem.updateData()` instead of the constructors, to cover the possibility of those being called directly. If stepMode is used in the constructors, these methods end up being called anyway.

Note 2: since the `warnings` module is not otherwise used in these modules, I chose to import it locally so that the whole deprecation check can be cleanly removed once it is no longer needed.

j9ac9k: Thanks for pointing this out; this looks good to me.

j9ac9k:
This looks good to me, thanks for the PR @cpascual I'll merge.

PR #1361, opened 2020-09-09
Prevent item duplication in Node context menu

bruchar1:
If you right-click on a Node in the Flowchart, there is a Add Node menu.
If you close the context menu, and open it again, there are two Add Node menu.
A new Add Node item is added each time you right-click on the Node.

This should solve this.


ixjlyons:
I haven't gotten a chance to dig into this much yet, but I did notice this issue only seems to happen with PySide2 and not PyQt5. Any insight there?

bruchar1:
I'm not sure, since I use PySide and PySide2, but not PyQt nor PyQt5. In PySide documentation, it is stated you cannot add an action twice:

> A PySide.QtGui.QWidget should only have one of each action and adding an action it already has will not cause the same action to be in the widget twice.

Maybe this is the same for adding QMenu in PyQt, but not in PySide? But this is just a blind guess...


Rethinking about this, maybe it would be better to modify the addParentContextMenus method of GraphicsScene instead. That would be more generic.

j9ac9k:
@bruchar1 this look good to me, I don't see a reason to merge, but I think you're right that modifying `addParentContextMenus` is probably the right solution here.  I'm going to leave this PR open as this provides a really handy example to test with ... if modifying `addParentContextMenus` method is straight forward to resolve this; I'll do that.

Thanks for the PR!

j9ac9k:
This is a tricky one, I got a debugger but the issue is that the QMenu being generated to add a node is different from the previous QMenu's that were added to the menu

```
(Pdb) pp menu.actions()
[<PySide2.QtWidgets.QAction(0x7fe1cdc57160 text="Add input" toolTip="Add input" menuRole=TextHeuristicRole visible=true) at 0x1235517c0>,
 <PySide2.QtWidgets.QAction(0x7fe1cdc57800 text="Add output" toolTip="Add output" menuRole=TextHeuristicRole visible=true) at 0x123551800>,
 <PySide2.QtWidgets.QAction(0x7fe1cdc33580 text="Remove node" toolTip="Remove node" menuRole=TextHeuristicRole visible=true) at 0x123551180>,
 <PySide2.QtWidgets.QAction(0x7fe1cc097fc0 text="" menuRole=TextHeuristicRole visible=true) at 0x138a5e300>,
 <PySide2.QtWidgets.QAction(0x7fe1d05fdbe0 text="Add node" toolTip="Add node" menuRole=TextHeuristicRole visible=true) at 0x122f33c40>,
 <PySide2.QtWidgets.QAction(0x7fe1ca7086a0 text="" menuRole=TextHeuristicRole visible=true) at 0x119be10c0>,
 <PySide2.QtWidgets.QAction(0x7fe1cc7e6580 text="Add node" toolTip="Add node" menuRole=TextHeuristicRole visible=true) at 0x13ac6dd40>,
 <PySide2.QtWidgets.QAction(0x7fe1ca657660 text="ViewBox options" toolTip="ViewBox options" menuRole=TextHeuristicRole visible=true) at 0x13ac673c0>,
 <PySide2.QtWidgets.QAction(0x7fe1cc7adc70 text="Export..." toolTip="Export" menuRole=TextHeuristicRole visible=true) at 0x123531e00>]
(Pdb) pp m.menuAction() in menu.actions()
False
(Pdb) m.menuAction()
<PySide2.QtWidgets.QAction(0x7fe1cc2c9010 text="Add node" toolTip="Add node" menuRole=TextHeuristicRole visible=true) at 0x13aced540>
```

I _really_ don't want to go off of string matching for the text, any other suggestions on how to filter out the (almost) duplicate entries?

j9ac9k:
So I have a "fix" in relation to modifying `GraphicsScene.addParentContextMenus` but I think I prefer your fix, I'll post it here in case there is a better suggestion:

```python
    def addParentContextMenus(self, item, menu, event):
        menusToAdd = []
        while item is not self:
            item = item.parentItem()
            if item is None:
                item = self
            if not hasattr(item, "getContextMenus"):
                continue
            subMenus = item.getContextMenus(event) or []
            if isinstance(subMenus, list): ## so that some items (like FlowchartViewBox) can return multiple menus
                menusToAdd.extend(subMenus)
            else:
                menusToAdd.append(subMenus)

        if menusToAdd:
            menu.addSeparator()

        for m in menusToAdd:
            if isinstance(m, QtGui.QMenu):
                if m.title() in [action.text() for action in menu.actions()]:
                    continue
                menu.addMenu(m)
            elif isinstance(m, QtGui.QAction):
                menu.addAction(m)
            else:
                raise Exception("Cannot add object %s (type=%s) to QMenu." % (str(m), str(type(m))))
            
        return menu
```

the new lines being toward the end

```python
                if m.title() in [action.text() for action in menu.actions()]:
                    continue
```

@ixjlyons you have any opinions on either of these fixes?

bruchar1:
I think I found a more elegant solution to the problem. See my last commit.

j9ac9k:
The CI failure I'm 99.9% sure is unrelated (this test is historically....fragile).

j9ac9k:
Fantastic fix @bruchar1 not sure how I missed that while I was tinkering with it.  Thanks for sticking with it.  I'll merge momentarily

PR #1364, opened 2020-09-10
fix for log scaling being broken 

meganbkratz:
Running examples/Plotting.py produced RuntimeWarnings and failed to display data for the middle plot (the example for log scaling). This PR fixes that.

-- log10 was being calculated twice, resulting in nan. Now it's only being calculated once.


ixjlyons:
Great, it looks like this is the correct one to remove so you can still get e.g. FFTs with log scaling. Thanks.

PR #1367, opened 2020-09-14
AxisItem: Account for empty strings in the visibility of text and units

dgoeries:
- Account for empty `text` and `units`
- Always reset the visibility of the label, in case of swap

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -283,19 +283,19 @@ class AxisItem(GraphicsWidget):
             axis.setLabel('label text', units='V', **labelStyle)
 
         """
-        show_label = False
-        # Account empty string and `None` for units and text
-        if text:
-            self.labelText = text
-            show_label = True
-        if units:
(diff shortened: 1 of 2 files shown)


j9ac9k:
Changed the base, I really should update the `CONTRIBUTING.md` instructions...

dgoeries:
Thanks! I believe I have to look at the tests. An exporter is failing now ...

dgoeries:
All tests pass through now. I had to cleanup a bit the AxisItem and disentangle the `setLabel` from internal use. There is way more potential for separation and cleanup.

There might be a slight regression potential depending on how the method was used. I put this to expected behavior (at least for me, `None` means empty label.)

Can you have a look again please.

j9ac9k:
Thanks for the PR @dgoeries , sorry this took me so long to get to; these diffs look good to me, but given I already hosed up AxisItem once by merging quickly, I'll ask @ixjlyons or @ksunden to take a look.

j9ac9k:
After taking another look, I think this is good, thanks again for the PR @dgoeries merging!

PR #1368, opened 2020-09-14
Fix disconnect() in Flowchart with wrong receiver slot

bruchar1:


ixjlyons:
Good catch! This makes me wonder if there really needs to be a try/except there in the first place, but I'm inclined to keep it since it seems like it was part of the original design (i.e. before https://github.com/pyqtgraph/pyqtgraph/commit/193367a56b0928c63d2f9e4ae6862e3204e1d756).

PR #1369, opened 2020-09-15
Bug fix in PlotCurveItem.mouseClickEvent

btonner:
Move ev.accept() to follow sigClicked.emit so event is passed with sigClicked.


btonner:
There was a change to the definition of sigClicked that eliminated passing the event, so this bug fix is irrelevant.

PR #1370, opened 2020-09-16
Fix deprecation warning in multiprocess module 

jenshnielsen:
The argument to qtimer.start should be an int not a float
and the current code raises a deprecation warning with python 3.8.
This fixes that by simply converting it to an int.


ixjlyons:
It looks like there are a number of places this needs to be fixed, but this is as good a place as any to start. Thanks.

jenshnielsen:
@ixjlyons Thanks. My test suite happened to hit this one which is why I started here

PR #1374, opened 2020-09-20
Test warnings cleanup

ixjlyons:
This should hopefully clean up most of the warnings cropping up in CI


ixjlyons:
There are a couple warnings from matplotlib related to the Qt4 backend, but I think we can live with this until Qt4 support is dropped

Also, there are some "CI warnings" as I'd call them related to the pytest.ini. I'll consider them out of scope for this PR.

PR #1375, opened 2020-09-20
Add targeted Vector test coverage

ixjlyons:
I wrote some tests for #1339 but didn't get around to finishing things up and pushing


j9ac9k:
Thanks @ixjlyons !

PR #1376, opened 2020-09-20
Fix PlotItem.setAxisItems

### LINKED ISSUE #1358: key error when setAxisItems in PlotItem
When I call the 
```python
plotItem = pg.PlotItem(viewBox=vb, name=name, axisItems={'bottom': self.axisTime})
```
I got this error:
```python
  File "/home/zhangr/anaconda3/envs/py38/lib/python3.8/site-packages/pyqtgraph/graphicsItems/PlotItem/PlotItem.py", line 159, in __init__
    self.setAxisItems(axisItems)
  File "/home/zhangr/anaconda3/envs/py38/lib/python3.8/site-packages/pyqtgraph/graphicsItems/PlotItem/PlotItem.py", line 329, in setAxisItems
    if axis != self.axes[k]["item"]:
KeyError: 'bottom'
```
There are maybe two bugs in this function, this is one solution:
```python
        # Array containing visible axis items
        # Also containing potentially hidden axes, but they are not touched so it does not matter
        visibleAxes = ['left', 'bottom']
        # visibleAxes.append(key)  key is dict_key, cannot append into a list
        for key in axisItems.keys():
            visibleAxes.append(key) # Note that it does not matter that this adds
                                    # some values to visibleAxes a second time

        for k, pos in (('top', (1,1)), ('bottom', (3,1)), ('left', (2,0)), ('right', (2,2))):
            if k in self.axes:
                if k not in axisItems:
                    continue # Nothing to do here

                # Remove old axis
                oldAxis = self.axes[k]['item']
                self.layout.removeItem(oldAxis)
                oldAxis.scene().removeItem(oldAxis)
                oldAxis.unlinkFromView()

            # Create new axis
            if k in axisItems:
                axis = axisItems[k]
                if axis.scene() is not None:
                    # check self.axes's key at first
                    if self.axes.get(k) and axis != self.axes[k]["item"]:
                        raise RuntimeError("Can't add an axis to multiple plots.")
            else:
                axis = AxisItem(orientation=k, parent=self)
```

### Code to reproduce

```python
import pyqtgraph as pg
import numpy as np
```

### Real behavior

```
An error occurred?
```

### Tested environment(s)

 * PyQtGraph version: 
 * Qt Python binding: 
 * Python version:  python3.8
 * NumPy version: 
 * Operating system:  ubuntu
 * Installation method:

ixjlyons:
Fixes #1358

I believe this is the intention of the logic that's there currently. Technically, it does appear to be possible to use a single `AxisItem` for multiple `PlotItem`s, but the conditions under which this works (and doesn't cause a segfault) aren't crystal clear to me. In the future, if someone figures out how to robustly detect when it's ok to do so, we can allow for it. For now this should make things work as intended and I'm pretty sure separate linked instances should work equivalently.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -325,7 +325,7 @@ class PlotItem(GraphicsWidget):
             if k in axisItems:
                 axis = axisItems[k]
                 if axis.scene() is not None:
-                    if self.axes.get(k) is None or axis != self.axes[k]["item"]:
+                    if k not in self.axes or axis != self.axes[k]["item"]:
                         raise RuntimeError(
                             "Can't add an axis to multiple plots. Shared axes"
                             " can be achieved with multiple AxisItem instances"
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -307,7 +307,7 @@ class PlotItem(GraphicsWidget):
         # Array containing visible axis items
         # Also containing potentially hidden axes, but they are not touched so it does not matter
         visibleAxes = ['left', 'bottom']
-        visibleAxes.append(axisItems.keys()) # Note that it does not matter that this adds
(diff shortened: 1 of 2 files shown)


j9ac9k:
Nice fix @ixjlyons Merging.

PR #1383, opened 2020-09-26
GraphicsView: Set a transparent background palette

dgoeries:
The background of the GraphicsView is blocking other features. In my example a grid in a designer. On the road to set a single viewbox background to white without affecting others:

- The background of the QPalette is set to transparent.

Before:

qt_graph_no_transparency

After:

qt_graph_transparency

### DIFF
--- a/pyqtgraph/widgets/GraphicsView.py
+++ b/pyqtgraph/widgets/GraphicsView.py
@@ -129,7 +129,12 @@ class GraphicsView(QtGui.QGraphicsView):
         self.mouseEnabled = False
         self.scaleCenter = False  ## should scaling center around view center (True) or mouse click (False)
         self.clickAccepted = False
-        
+
+        # Set a transparent background QPalette!
+        palette = self.palette()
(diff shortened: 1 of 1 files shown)


dgoeries:
It is a different data update,  that is why the Y axis has different values.

j9ac9k:
Thanks for the PR @dgoeries I'm good with this, but will wait for some more input before merging.

dgoeries:
thanks very much!

j9ac9k:
Unfortunately this PR broke some Qt4 functionality (can't wait to kill off Qt4 support!); I'm going to submit another PR putting this change under an `if QT_LIB in ["PyQt5", "PySide2"]`

j9ac9k:
Unfortunately this change has made some _significant_ performance regressions on scatter plots.

I'll dive into the Qt docs to see if there is some kind of alternative/workaround, but we may have to undo this change.

dgoeries:
That would be sad, but how? I don't see any difference in fps with the benchmark tool

dgoeries:
Testing 0.11.0 with and without this and I see a performance drop.

PR #1387, opened 2020-10-01
ScatterPlotItem: Make + and x symbols thicker

irgolic:
This is something we're overriding in Orange3 (biolab/orange3#5007), as we change the symbols' alpha_value to show selected symbols. The default + and x symbols are too thin to show a noticeable change in alpha_value.

But I thought you might agree that the symbols look nicer this way.

Before:

Screenshot 2020-10-01 at 12 56 36

After:

Screenshot 2020-10-01 at 12 55 48


j9ac9k:
I like the change myself; will wait for more input before merging.  Thanks for the PR @irgolic

dgoeries:
LGTM

PR #1388, opened 2020-10-01
Emit signal when GraphicScene.addItem() is used

cpascual:
Emit a signal (`sigItemAdded`) just after adding an item to a scene.
The item object is emited as the only signal argument.
This signal is useful for code that may want to react to newly added
items of a plot.


j9ac9k:
I'm good with this feature; anyone else have thoughts?

dgoeries:
LGTM

j9ac9k:
@dgoeries good call on `removeItem`; @cpascual would you mind adding a similar signal for when `removeItem` is called as well?

cpascual:
> adding a similar signal for when `removeItem` is called as well

Ups... sorry for the silence
I missed this comment and I was just waiting... I just pushed the required change

j9ac9k:
Looks good to me, thanks for the PR @cpascual , merging!

PR #1391, opened 2020-10-02
caching for viewRect of LinearRegionItem to reduce CPU load

jjuod:
## Issue

Having a LinearRegionItem inside a ViewBox causes large CPU loads on any mouse interaction - just moving the cursor around anywhere in the ViewBox (even outside the Item) takes up 2 threads at 100 % for me. This seems to be caused by `viewRect()` of the LinearRegionItem, which is called on every mouse event and does some heavy coordinate transforms.

## Solution
Similar issues elsewhere in Qt/pyqtgraph seem to be solved by caching the boundaries. I suspected that the `viewTransformChanged()` slot on pyqtgraph's GraphicsItem was intended specifically to allow that, so I set up a simple cache which is invalidated by `viewTransformChanged()`. This seems to work (CPU load reduced ~20x), but I'm not sure if that slot really catches every change that invalidates `viewRect`. Would be good if someone with better understanding could take a look.

### DIFF
--- a/pyqtgraph/graphicsItems/GraphicsItem.py
+++ b/pyqtgraph/graphicsItems/GraphicsItem.py
@@ -36,6 +36,7 @@ class GraphicsItem(object):
         self._viewBox = None
         self._connectedView = None
         self._exportOpts = False   ## If False, not currently exporting. Otherwise, contains dict of export options.
+        self._cachedView = None
         if register is not None and register:
             warnings.warn(
                 "'register' argument is deprecated and does nothing",
@@ -154,6 +155,10 @@ class GraphicsItem(object):
     def viewRect(self):
         """Return the visible bounds of this item's ViewBox or GraphicsWidget,
         in the local coordinate system of the item."""
+        if self._cachedView is not None:
+            return self._cachedView
+
+        # Note that in cases of early returns here, the view cache stays empty (None).
         view = self.getViewBox()
         if view is None:
             return None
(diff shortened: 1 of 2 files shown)


j9ac9k:
Thanks for this PR! As I frequently use linear region items I've been suspicious of some performance issue when using a significant number of them...

I'll try and review/test this in the next day or two.

campagnola:
This looks great! I can't think of a reason why the caching should cause any problems here.

However: why not just implement this on `GraphicsItem` instead? Then potentially many different items will benefit form it.

jjuod:
@campagnola I looked at the code for some other items briefly, and they seemed to have their own caching, albeit implemented in some slightly different way for each. (They also didn't seem to have any performance issues.) Figured I might break something if I try to harmonize that...

j9ac9k:
@jjuod I'm good w/ implementing it at `GraphicsItem`, the other objects that inherit `GraphicsItem` can re-implement those methods...but that may be outside the scope of the PR so if you're reluctant to make that implementation, I'll take this PR as it is.

jjuod:
Ok, moved the caching to `GraphicsItem`. Works fine with instances of LinearRegionItems. Haven't tested the other items, but it doesn't seem like they could be broken by this in any way.

j9ac9k:
Merged, thanks for the PR @jjuod !

j9ac9k:
Turns out we did break something!

image

this is a screenshot of the basic example app; when you zoom out, notice how the infinite lines do not scale all the way up...

I'm thinking the solution here is _not_ to undo this PR, but to add some more logic to InfiniteLine

j9ac9k:
Managed to fix this by adding the call to `GraphicsItem` for `viewTransformChanged()`

```
        """
        Called whenever the transformation matrix of the view has changed.
        (eg, the view range has changed or the view was resized)
        """
        GraphicsItem.viewTransformChanged(self)
        self._invalidateCache()
```

PR #1392, opened 2020-10-04
SignalProxy: Correct initialization without slot argument and tests

### LINKED ISSUE #1178: TypeError in SignalProxy __init__
### Short description
Found a small bug in `SignalProxy.__init__()` that occurs when creating an instance of `SignalProxy` without directly handing over a slot to connect to.
The default value for the optional init parameter `slot` is `None` and so you can not create a `weakref` from that.
The culprit is line 38 of _SignalProxy.py_ https://github.com/pyqtgraph/pyqtgraph/blob/develop/pyqtgraph/SignalProxy.py#L38

My proposal would be to get rid of the instance attribute `self.slot` entirely since you only use it for disconnection in `SignalProxy.disconnect`. This is not needed as you could simply call `self.sigDelayed.disconnect()` without any arguments. 
This would be even better since it would disconnect ALL slots (even the ones "manually" connected) from `sigDelayed` as the user would expect from calling `disconnect` on any `QObject` (see also https://doc.qt.io/qt-5/qobject.html#disconnect).

### Code to reproduce

```python
# -*- coding: utf-8 -*-
import pyqtgraph as pg
import numpy as np

class Dummy(pg.QtCore.QObject):
    sigTest = pg.QtCore.Signal()
    
    def my_slot():
        pass


if __name__ == '__main__':
    app = pg.QtCore.QCoreApplication([])
    dummy_obj = Dummy()
    proxy = pg.SignalProxy(dummy_obj.sigTest)  # raises TypeError
    proxy.sigDelayed.connect(dummy_obj.my_slot)
    # This would work instead:
    # proxy = pg.SignalProxy(dummy_obj.sigTest, slot=dummy_obj.my_slot)
```

### Expected behavior
One should be able to create instances of `SignalProxy` by just passing the sender signal.
It seems like this was intended behaviour, as it should be, since it can be beneficial to create a proxy and connect the receiving slot later.

### Real behavior
```
  File ".\pg_signal_proxy_test.py", line 15, in <module>
    proxy = pg.SignalProxy(dummy_obj.sigTest)  # raises TypeError
  File "C:\Program Files\Python37\lib\site-packages\pyqtgraph\SignalProxy.py", line 38, in __init__
    self.slot = weakref.ref(slot)
TypeError: cannot create weak reference to 'NoneType' object
```

### Tested environment(s)

 * PyQtGraph version: 0.10.0.4
 * Qt Python binding: PyQt5 5.9.2 Qt 5.9.3
 * Python version: 3.7.7
 * NumPy version: 1.18.3
 * Operating system: Debian 10 (stable) & Win10 Enterprise (10.0.18363)
 * Installation method: pip

dgoeries:
This PR will fix #1178 

- Provided a bunch of basic tests in addition
- Took the opportunity to clean up a bit the file

### DIFF
--- a/pyqtgraph/SignalProxy.py
+++ b/pyqtgraph/SignalProxy.py
@@ -1,11 +1,13 @@
 # -*- coding: utf-8 -*-
+import weakref
+
 from .Qt import QtCore
 from .ptime import time
 from . import ThreadsafeTimer
-import weakref
 from .functions import SignalBlock
 
 __all__ = ['SignalProxy']
+
 
 class SignalProxy(QtCore.QObject):
     """Object which collects rapid-fire signals and condenses them
@@ -13,11 +15,12 @@ class SignalProxy(QtCore.QObject):
     Used, for example, to prevent a SpinBox from generating multiple 
     signals when the mouse wheel is rolled over it.
     
-    Emits sigDelayed after input signals have stopped for a certain period of time.
+    Emits sigDelayed after input signals have stopped for a certain period of
+    time.
     """
-    
+
     sigDelayed = QtCore.Signal(object)
-    
+
     def __init__(self, signal, delay=0.3, rateLimit=0, slot=None):
         """Initialization arguments:
         signal - a bound Signal or pyqtSignal instance
@@ -26,7 +29,7 @@ class SignalProxy(QtCore.QObject):
         rateLimit - (signals/sec) if greater than 0, this allows signals to stream out at a 
                     steady rate while they are being received.
         """
-        
+
         QtCore.QObject.__init__(self)
         signal.connect(self.signalReceived)
         self.signal = signal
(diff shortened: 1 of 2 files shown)


dgoeries:
Sometimes ...

dgoeries:
Finally, I learned why I never used PySide :-). Maybe an integrator has a look?

j9ac9k:
Thanks for the PR @dgoeries  especially for the tests!

We have a small handful of PRs that are effectively waiting for the post-0.11 release where we no longer support Qt4 and Py2, and given the issues with PySide .... that said, given this addresses an issue that is facing all versions, I'm not opposed to merging right away.

@campagnola @ixjlyons @ksunden you all have opinions?

pyqtgraph/SignalProxy.py
```diff
@@ -26,86 +29,81 @@ def __init__(self, signal, delay=0.3, rateLimit=0, slot=None):
...
-        """Received signal. Cancel previous timer and store args to be forwarded later."""
+        """Received signal. Cancel previous timer and store args to be
+        forwarded later."""
         if self.blockSignal:
             return
         self.args = args
         if self.rateLimit == 0:
             self.timer.stop()
-            self.timer.start((self.delay*1000)+1)
+            self.timer.start((self.delay * 1000) + 1)
```
j9ac9k, line 61: `timer.start()` needs to take an `int` or else it spews a deprecation warning; can you have the starting parameter be `int(self.delay * 1000) + ``?

pyqtgraph/SignalProxy.py
```diff
@@ -26,86 +29,81 @@ def __init__(self, signal, delay=0.3, rateLimit=0, slot=None):
...
                 leakTime = 0
             else:
                 lastFlush = self.lastFlushTime
                 leakTime = max(0, (lastFlush + (1.0 / self.rateLimit)) - now)
-                
+
             self.timer.stop()
-            self.timer.start((min(leakTime, self.delay)*1000)+1)
-        
+            self.timer.start((min(leakTime, self.delay) * 1000) + 1)
```
j9ac9k, line 71: also convert to `int()`

j9ac9k:
Thanks for the PR @dgoeries  this looks good to me, only changes I request is converting the `time.start()` parameters to always be of type `int`.

dgoeries:
thanks for the review! Changes were made accordingly

j9ac9k:
Thanks for the update @dgoeries , this looks good to me, I'll merge.

PR #1396, opened 2020-10-13
GraphicsView: Add basic tests ...

dgoeries:


j9ac9k:
Thanks for expanding our test coverage @dgoeries we appreciate it!

pijyoi:
I have a query regarding the "app.deleteLater()" line in:
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/widgets/tests/test_graphics_view.py#L91

Am I understanding correctly from:
https://doc.qt.io/qt-5/qobject.html#deleteLater
that deleteLater should not be called on the QApplication instance since it is within the event loop itself that the deferred delete object would get destroyed?

For instance, on Windows 10 / Python 3.7.9, the following snippet crashes for {PySide2 5.15, PyQt5 5.15} and hangs for {PySide6 6.0, PyQt6 6.0}
```python
import pyqtgraph as pg
app = pg.mkQApp()
pg.Qt.QtCore.QTimer.singleShot(500, pg.Qt.QtWidgets.QApplication.quit)
app.deleteLater()
app.exec_()
print('DONE')
```

j9ac9k:
oof we probably don't want that one there, or the `del` line above.

dgoeries:
It does not fit in the pyqtgraph testing and can be removed. I realized too late

PR #1403, opened 2020-10-18
Do not apply transparent background in Qt4

j9ac9k:
This makes it so the change in PR #1383 is only applied to when using Qt5.  When using Qt4, this would make non-updating GraphicViews transparent/blank.


dgoeries:
many thanks, LGTM

PR #1404, opened 2020-10-19
LegendItem: Make ItemSample customizable

dgoeries:
This PR introduces `sampleType` of the LegendItem. It is possible to customize the `ItemSample` of the LegendItem


examples/Legend.py
```diff
@@ -32,14 +32,13 @@
 win.addItem(s1)
 
 # # option2: generic method------------------------------------------------
-legend = pg.LegendItem((80,60), offset=(70,20))
+legend = pg.LegendItem((80,60), offset=(70,20), itemStyle=1, labelTextSize='10pt')
```
dgoeries, line 35: I will move this itemStyle to 0

j9ac9k:
Thanks for the PR @dgoeries ... I love this feature; I've been on the fence with what the "disabled" should look like.  My first inclination is they should appear like a disabled QWidget (all greyed out) but I'm not sure that's most appropriate for a legend in a plot area... I do think the text should look somewhat different if disabled, maybe add some alpha to the color to make it a bit transparent?

dgoeries:
@j9ac9k I will try, but it will be more invasive. In my head right now, the symbol item and label would have to be merged to a single one for that.
Maybe that is anyway not a bad idea?

j9ac9k:
Feel free to hold off and wait on some more feedback; ...this kind of stuff can be subjective, so let's not have my opinion alone be a reason for further style modifications.

dgoeries:
Okay. I can also refactor later and move to a single item. :-)

2xB:
This looks great!

I agree to @j9ac9k, having more indication than the fill of these rectangles would be great. Although I would also like to see something like this with the "classic" item styles, because they are quite good at indicating to what the legend item corresponds. So yes, if an itemStyle would also cover the text, that could be useful. Also, `ToggleSample` could extend `ClassicSample` and add an "invisible" sign (e.g. a grey circle with a diagonal line) on top. That could be quite simple to interpret while keeping the default signs.

Btw. I would suggest if possible that you call the abstract base class something different than `ItemSample` and leave the latter name for what is `ClassicSample` in your suggestion, since as far as I can tell that is an easy way to assure that if an application previously extended or monkey patched `ItemSample`, it is not broken after this update.

examples/Legend.py
```diff
@@ -32,14 +32,13 @@
 win.addItem(s1)
 
 # # option2: generic method------------------------------------------------
-legend = pg.LegendItem((80,60), offset=(70,20))
+legend = pg.LegendItem((80,60), offset=(70,20), itemStyle=0, labelTextSize='10pt')
```
drewrisinger, line 35: In my opinion, it makes more sense to the user to have these as ``enum``, which has been supported in python for a while.
```suggestion
legend = pg.LegendItem((80,60), offset=(70,20), itemStyle=pg.LegendStyle.Classic, labelTextSize='10pt')
```

pyqtgraph/graphicsItems/LegendItem.py
```diff
@@ -9,7 +9,12 @@
 from .GraphicsWidgetAnchor import GraphicsWidgetAnchor
 from .BarGraphItem import BarGraphItem
 
-__all__ = ['LegendItem']
+__all__ = ['LegendItem', 'LegendStyle']
+
+
+class LegendStyle:
```
drewrisinger, line 15:
```suggestion
class LegendStyle(enum.IntEnum):
```

drewrisinger (changes requested):
Looks pretty nifty in general. My main comment is about making the LegendStyle into an enum so that more user-friendly values can optionally be passed in.

pyqtgraph/graphicsItems/LegendItem.py
j9ac9k: If we do that, we'll have to hold off on merging this until after we deprecate py2 support (0.12 release?)

drewrisinger: Ah, didn't look closely enough to see that py2 was still supported, since it's been EOL almost a year.

j9ac9k: Don't get me started (maintaining CI pipelines with python2 and old Qt bindings has been a real challenge)...we're looking at adopting NEP-29, but first order of business is merge outstanding py2 compatible PRs, the contributors should have the benefit of (eventually) seeing their PRs be in a stable release supporting versions of python they could have been using at the time of submission...

Once the PR count is down to ~5ish, we'll likely attempt to resolve any bug regressions, do another release, and then kill off py2/py < 3.7 support.

dgoeries: I also prefer enums more but have to accept python 2 here

dgoeries:
I can bring a version that covers the text, but I don't see a nice way doing this without merging the label and sample item

dgoeries:
Sorry it took so long, but I never found a good way to meet requirements without burning down the `LegendItem` completely (regression).
I changed the PR to make the ItemSample class customizable. Is that okay?

dgoeries:
One could move the toggle feature to the ItemSample and provide a different icon then? E.g. invisible eye?

j9ac9k:
> One could move the toggle feature to the ItemSample and provide a different icon then? E.g. invisible eye?

That's brilliant, I think that would make for a much more intuitive implementation IMO.

pyqtgraph/graphicsItems/LegendItem.py
```diff
@@ -73,16 +75,39 @@ def __init__(self, size=None, offset=None, horSpacing=25, verSpacing=0,
         if size is not None:
             self.setGeometry(QtCore.QRectF(0, 0, self.size[0], self.size[1]))
 
+        if sampleType is not None:
+            assert issubclass(sampleType, GraphicsWidget)
```
j9ac9k, line 79: `assert`'s in functional code are generally not a great idea as they are bypassed if someone runs `python -O script.py

```bash
$ python --help

...
-O     : remove assert and __debug__-dependent statements; add .opt-1 before
         .pyc extension; also PYTHONOPTIMIZE=x
...
```

If we want to have this check (which I have no problem with), we should have an `if not issubclass...` followed by a `RuntimeError`

I'm totally open to other checks/errors to be raised, this is just what I thought of first.

pyqtgraph/graphicsItems/LegendItem.py
```diff
@@ -73,16 +75,39 @@ def __init__(self, size=None, offset=None, horSpacing=25, verSpacing=0,
...
             'labelTextColor': labelTextColor,
             'labelTextSize': labelTextSize,
             'offset': offset,
         }
-
         self.opts.update(kwargs)
 
+    def setSampleType(self, sample):
+        """Set the new sample item claspes"""
+        if sample == self.sampleType:
```
j9ac9k, line 95: do we want an equality operator here or an `is` operator?

dgoeries: i think its a good idea

examples/Legend.py
dgoeries: Different implementation, this is not required anymore

j9ac9k:
Initially drafted a post here indicating I thought an issue with `sampleType` being a reference to a class, or an instance of a class, only to re-look over the PR and realize the issue was with me.

This PR LGTM, I'll merge momentarily.

I think the feature of having the invisible eye toggle visibility would be amazing; if you want to tackle that, that would be greatly appreciated.

dgoeries:
thanks for merging! Yes, I am on the toggle. Just want to make it nice

PR #1405, opened 2020-10-19
ScatterPlotItem: Fix name setting in options

dgoeries:


pyqtgraph/graphicsItems/tests/test_ScatterPlotItem.py
```diff
@@ -1,31 +1,37 @@
 from pyqtgraph.Qt import QtGui, QtCore
 import pyqtgraph as pg
 import numpy as np
-app = pg.mkQApp()
-app.processEvents()
-
 
 
 def test_scatterplotitem():
+    app = pg.mkQApp()
```
j9ac9k, line 7: Why make this part of the test function instead of leaving it at the top level as before?

dgoeries: For my experience it is better to create a QApp for each test and cleanly delete it after test due to events and signals.

One occurence was the test_PolyROI were instances probably collided? If you want I will be happy to change back, but If believe it is better to go this direction instead of creating an APP even module wise

j9ac9k: I'm fine with that... eventually when we go to pyqtgraph 0.12, we should probably migrate to `pytest-qt` and just use the qtbot fixture to manage the application instance.

dgoeries: yes!

j9ac9k:
LGTM, merging, thanks for the PR @dgoeries

PR #1411, opened 2020-10-22
Hide WidgetParameterItem.defaultBtn if param has no default

lidstrom83:
Currently the button is both visible and enabled when the parameter has no default. This PR hides it in this case. It could alternatively disable it - I just found this way more visually appealing.


ixjlyons (approved):

j9ac9k:
LGTM thanks @lidstrom83

PR #1413, opened 2020-10-22
Call GraphicsItem.viewTransformChanged()

j9ac9k:
Follow up to #1391 got a bug report on slack about infinite lines not extending the height of the view; this was due to some cache handling in GraphicsItem; fix was just to call the inherited viewTransformChanged method from GraphicsItem for InifiniteLine.

I'm wondering if we should take a look at some of the other places we call out `viewTransformChanged` explicitly.

Other places we call it include:

-  ImageItem.py
-  PlotCurveItem.py
-  ROI.py
-  ScatterPlotItem.py
-  TextItem.py
-  TargetItem.py (not currently used much)

### DIFF
--- a/pyqtgraph/graphicsItems/InfiniteLine.py
+++ b/pyqtgraph/graphicsItems/InfiniteLine.py
@@ -2,6 +2,7 @@
 from ..Qt import QtGui, QtCore
 from ..Point import Point
 from .GraphicsObject import GraphicsObject
+from .GraphicsItem import GraphicsItem
 from .TextItem import TextItem
 from .ViewBox import ViewBox
 from .. import functions as fn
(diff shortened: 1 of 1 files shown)


j9ac9k:
did a quick run through of these, the infinite line labels did not behave as intended, so I made the call to the GraphicsItem method from there too... I think this should do it...

j9ac9k:
Fixes #1412

pyqtgraph/graphicsItems/InfiniteLine.py
```diff
@@ -607,6 +609,7 @@ def hoverEvent(self, ev):
             ev.acceptDrags(QtCore.Qt.LeftButton)
 
     def viewTransformChanged(self):
+        GraphicsItem.viewTransformChanged(self)
```
dgoeries, line 612: I did not look into the code (example), but normally the `TextItem.viewTransformChanged` should do the job. Did you check if the order is important?

PR #1414, opened 2020-10-22
Emit event with mouse clicks for some items

lidstrom83:
Harmonizes behavior with PlotCurveItem.


j9ac9k:
LGTM, thanks for the PR @lidstrom83

Issue #1415, opened 2020-10-23
Shape of getArrayRegion


Toussaic:
Hello,

I try to extract a region with a RectROI. The ROI have a size of (20, 20) pixels. The ROI is moving along the time.
Sometimes, the getArrayRegion function returns an array with one pixel more (20, 21) or (21, 20). This is not what I expected.

So I tried to use the "shape" parameter which should be passed to "affineSlice" (I'm not even sure this can solve the problem). But as I saw in the source code, three parameters are computed and passed to affineSlice instead of using the ones passed in kwds.

```
def getArrayRegion(self, data, img, axes=(0,1), returnMappedCoords=False, **kwds):
        r"""Use the position and orientation of this ROI relative to an imageItem
        to pull a slice from an array.
        =================== ====================================================
        **Arguments**
        <...>
        \**kwds             All keyword arguments are passed to 
                            :func:`affineSlice <pyqtgraph.affineSlice>`.
        =================== ====================================================
        <...>
        All extra keyword arguments are passed to :func:`affineSlice <pyqtgraph.affineSlice>`.
        """
        # this is a hidden argument for internal use
        fromBR = kwds.pop('fromBoundingRect', False)
```

Here the values are computed, so the ones passed as parameters can't be used.
Those three parameters should be popped from "kwds" OR the parameters should replace the ones computed. I don't know which solution is more appropriate.

```
        shape, vectors, origin = self.getAffineSliceParams(data, img, axes, fromBoundingRect=fromBR)
        if not returnMappedCoords:
            rgn = fn.affineSlice(data, shape=shape, vectors=vectors, origin=origin, axes=axes, **kwds)
            return rgn
        else:
            kwds['returnCoords'] = True
            result, coords = fn.affineSlice(data, shape=shape, vectors=vectors, origin=origin, axes=axes, **kwds)
            
            ### map coordinates and return
            mapped = fn.transformCoordinates(img.transform(), coords)
            return result, mapped
```

AND maybe there is a problem in the "getAffineSliceParams" when computing the shape (rounding?)

**Tested environment(s)**

    PyQtGraph version: 0.11.0
    Qt Python binding: Qt 5.12.3
    Python version: 3.8.6
    Operating system: windows
    Installation method: conda

j9ac9k:
Fixed with merging #1416

PR #1416, opened 2020-10-23
User-defined keywords of getArrayRegion

Toussaic:
The three parameters "shape, vectors, origin" can't be passed as parameters because they are computed, which raise an error when kwds is passed to affineSlice.

Proposal for issue #1415


j9ac9k:
Hi @Toussaic thanks for the PR.  This looks good to me.

PR #1341, opened 2020-08-10
Make ScatterPlotItem.paint aware of layout offset

dmtvanzanten:
Without this addition the markers of the ScatterPlotItem are not visible in the export when the ScatterPlotItem is not located in the first panel of a Layout.

Issue #1209, opened 2020-05-26
Symbols of scatter plot disappears after exporting.  


sophie-cal2020:
### Short description

Was trying to run plotting.py in the examples folder.  And I noticed that the symbols of the scatter plots disappears in the .png after exporting.  I have installed pyqtgraph version 0.11.0.dev0 running on python 3.7.4.  I did a search on past issues and it seems that the problem should have been solved already?

j9ac9k:
By `0.11.0.dev0` do you mean `0.11.rc0` or did you install pyqtgraph by aiming pip at the github repo?

j9ac9k:
Not able to replicate from the current dev build:

![image](https://user-images.githubusercontent.com/646398/82931345-a34f6e00-9f3b-11ea-8cae-5f7bc734ee2f.png)

I installed `0.11.rc0` and could not replicate either

![image](https://user-images.githubusercontent.com/646398/82931556-f2959e80-9f3b-11ea-85f7-c9871aa13cf1.png)

sophie-cal2020:
Hi,

I installed version 0.11.0.dev0.  I just tested it again and realised that the problem only happens if I choose to export only a selected plot (e.g. the scatter plot on the top right).  Exporting the entire scene is fine.

![test](https://user-images.githubusercontent.com/65949208/82968488-03a7e380-a000-11ea-8de4-ba3c8a1013ef.png)


Thanks.

sophie-cal2020:
Just to clarify, i installed by aiming pip at the github repo.  Thanks.

j9ac9k:
Can you tell us more about the system and version of Qt bindings you're using?

sophie-cal2020:
I'm running on Windows 10 with PyQt5, version 5.13.0.  Thanks.

dmtvanzanten:
I'm able to reproduce with pyqtgraph (0.11.0) running Windows 10, Python 3.8 and pyqt 5.9.2.
I also noticed that the markers only disappear for panels that aren't the first panel in a GraphicsLayout. 
For example, after running the code below, and exporting the most left panel (using the ui) I do see the markers, When exporting the middle or most right panel, there are no markers visible. 

```
win = pg.GraphicsLayoutWidget()
layout = win.ci

for i in range(3):
    line = pg.ScatterPlotItem(x[::5], y[::5], brush='ffffff')
    panel = win.addPlot()
    panel.addItem(line)
```

[...]

PR #1420, opened 2020-10-25
Scatter Plot Improvements

### LINKED ISSUE #1020: Poor performance when changing colors of spots in a plot, without changing positions
### Short description

I have a scatterplot with 2000 points and I have a QSlider which changes the colors of the points, without changing their positions. I am experiencing poor performance and the feeling of the application is laggy. In the past I used a QSlider to change an image and the rendering time was almost 20 times faster, so I believe that the performance can be improved.

### Code to reproduce

I created a [minimal working example](https://github.com/LucaMarconato/dynamic_scatter_plot) which shows my problem and prints some benchmarks in the console.

### Expected behavior

The rendering time should be much smaller. Maybe achieving a short running time is more difficult in the general case in which also the positions of the points change, but in this case I want to replace only the colors.

### Real behavior

The rendering is slow.

### Tested environment(s)

 * PyQtGraph version: `0.10.0` 
 * Qt Python binding: `'PyQt5 5.9.2 Qt 5.9.6` 
 * Python version: `3.7.4`
 * NumPy version: `1.16.4` 
 * Operating system: macOS `10.14.6`
 * Installation method: `conda`

lidstrom83:
- Fixes various bugs and performance issues.
- Adds a hovering API to `ScatterPlotItem` which lets the user specify a separate style for the hovered points and show a tool tip containing information about them. A signal is also emitted during hovering.

### DIFF
--- a/examples/ScatterPlot.py
+++ b/examples/ScatterPlot.py
@@ -118,13 +118,9 @@ s4.sigClicked.connect(clicked)
 s4.setAcceptHoverEvents(True)
 hoverPoints = set()
 hoverPen = pg.mkPen('g')
-hoverRect = None
 
 def hoverEvent(ev):
-    global hoverPoints, hoverRect
-
-    if hoverRect is None and len(hoverPoints) > 0:
-        hoverRect = next(iter(hoverPoints))._data['sourceRect']  # avoid gc to prevent removal of glyph from cache
+    global hoverPoints
 
     newPoints = [] if ev.exit else s4.pointsAt(ev.pos())
     newPointsSet = set(newPoints)
--- a/examples/ScatterPlot.py
+++ b/examples/ScatterPlot.py
@@ -129,9 +129,10 @@ def hoverEvent(ev):
         lastHovered = points
 
         cutoff = 3
-        tip = '\n\n'.join(f'index: {pt.index()}\nx: {pt.pos().x():.3g}\ny: {pt.pos().y():.3g}' for pt in points[:cutoff])
+        tip = '\n\n'.join('index: {}\nx: {:.3g}\ny: {:.3g}'.format(pt.index(), pt.pos().x(), pt.pos().y())
+                          for pt in points[:cutoff])
         if len(points) > cutoff:
-            tip += f'\n\n({len(points) - cutoff} more...)'
+            tip += '\n\n({} more...)'.format(len(points) - cutoff)
 
         if tip:
             s4.getViewBox().setToolTip(tip)
--- a/examples/ScatterPlot.py
+++ b/examples/ScatterPlot.py
@@ -123,23 +123,25 @@ def hoverEvent(ev):
     global hoverPoints
 
     newPoints = [] if ev.exit else s4.pointsAt(ev.pos())
-    newPointsSet = set(newPoints)
-
-    for pt in newPointsSet.difference(hoverPoints):
-        pt.setPen(hoverPen)
-
-    for pt in hoverPoints.difference(newPoints):
-        pt.resetPen()
 
     cutoff = 3
-    tip = '\n\n'.join('index: {}\nx: {:.3g}\ny: {:.3g}'.format(pt.index(), pt.pos().x(), pt.pos().y())
-                      for pt in newPoints[:cutoff])
+    tip = ['index: {}\nx: {:.3g}\ny: {:.3g}'.format(pt.index(), pt.pos().x(), pt.pos().y())
+           for pt in newPoints[:cutoff]]
     if len(newPoints) > cutoff:
-        tip += '\n\n({} more...)'.format(len(newPoints) - cutoff)
+        tip.append('({} others...)'.format(len(newPoints) - cutoff))
+    s4.getViewBox().setToolTip('\n\n'.join(tip))
 
-    s4.getViewBox().setToolTip(tip)
+    newPoints = set(newPoints)
 
-    hoverPoints = newPointsSet
+    for pt in newPoints:
+        if pt not in hoverPoints:
+            pt.setPen(hoverPen)
+
+    for pt in hoverPoints:
+        if pt not in newPoints:
+            pt.resetPen()
+
+    hoverPoints = newPoints
 
 s4.hoverEvent = hoverEvent
 
--- a/examples/ScatterPlot.py
+++ b/examples/ScatterPlot.py
@@ -116,6 +116,7 @@ s4.sigClicked.connect(clicked)
 # Enable hovering and show a tool tip hovered points
 s4.setAcceptHoverEvents(True)
 lastHovered = []
+hoverPen = pg.mkPen('g')
 
 def hoverEvent(ev):
     global lastHovered
@@ -125,7 +126,7 @@ def hoverEvent(ev):
     if not ev.exit:
         points = s4.pointsAt(ev.pos())
         for pt in points:
-            pt.setPen('g')
+            pt.setPen(hoverPen)
         lastHovered = points
 
         cutoff = 3
--- a/examples/ScatterPlot.py
+++ b/examples/ScatterPlot.py
@@ -124,30 +124,33 @@ s4.sigClicked.connect(clicked)
 
 # Enable hovering and show information about hovered points in a tool tip
 s4.setAcceptHoverEvents(True)
-hovered = []
+hoverPoints = set()
 hoverPen = pg.mkPen('r', width=2)
-oldData = None
+
 
 def hoverEvent(ev):
-    global hovered
+    global hoverPoints
 
-    new = [] if ev.exit else s4.indexesAt(ev.pos())
+    newPoints = [] if ev.exit else s4.pointsAt(ev.pos())
 
     cutoff = 3
-    data = s4.loc[new[:cutoff], ['x', 'y', 'data']]
-    tip = ['x: {:.3g}\ny: {:.3g}\ndata: {}'.format(*rec) for rec in data]
-    if len(new) > cutoff:
-        tip.append('({} others...)'.format(len(new) - cutoff))
+    tip = ['x: {x:.3g}\ny: {y:.3g}\ndata={data}'.format(x=pt.pos().x(), y=pt.pos().y(), data=pt.data())
+           for pt in newPoints[:cutoff]]
+    if len(newPoints) > cutoff:
+        tip.append('({} others...)'.format(len(newPoints) - cutoff))
     s4.getViewBox().setToolTip('\n\n'.join(tip))
 
-    idx1 = np.setdiff1d(new, hovered)
-    if len(idx1) > 0:
-        s4.loc[idx1, 'pen'] = hoverPen
-    idx2 = np.setdiff1d(hovered, new)
-    if len(idx2) > 0:
-        s4.loc[idx2, 'pen', ] = None
+    newPoints = set(newPoints)
 
-    hovered = new
+    for pt in newPoints:
+        if pt not in hoverPoints:
+            pt.setPen(hoverPen)
+
+    for pt in hoverPoints:
+        if pt not in newPoints:
+            pt.resetPen()
+
+    hoverPoints = newPoints
 
 s4.hoverEvent = hoverEvent
 
--- a/examples/ScatterPlot.py
+++ b/examples/ScatterPlot.py
@@ -46,6 +46,7 @@ s1.addPoints(spots)
 w1.addItem(s1)
 
 ## Make all plots clickable
+clickedPen = pg.mkPen('b', width=2)
 lastClicked = []
 def clicked(plot, points):
     global lastClicked
@@ -53,7 +54,7 @@ def clicked(plot, points):
         p.resetPen()
     print("clicked points", points)
     for p in points:
-        p.setPen('b', width=2)
+        p.setPen(clickedPen)
     lastClicked = points
 s1.sigClicked.connect(clicked)
 
@@ -109,8 +110,14 @@ s3.sigClicked.connect(clicked)
 ## Test performance of large scatterplots
 
 s4 = pg.ScatterPlotItem(size=10, pen=pg.mkPen(None), brush=pg.mkBrush(255, 255, 255, 20))
-pos = np.random.normal(size=(2,10000), scale=1e-9)
-s4.addPoints(x=pos[0], y=pos[1])
+n = 10000
+pos = np.random.normal(size=(2, n), scale=1e-9)
+s4.addPoints(
+    x=pos[0],
+    y=pos[1],
+    # size=(np.random.random(n) * 20.).astype(int),
+    # brush=[pg.mkBrush(x) for x in np.random.randint(0, 256, (n, 3))]
+)
 w4.addItem(s4)
 s4.sigClicked.connect(clicked)
 
--- a/examples/ScatterPlot.py
+++ b/examples/ScatterPlot.py
@@ -12,6 +12,7 @@ from pyqtgraph.Qt import QtGui, QtCore
 import pyqtgraph as pg
 import numpy as np
 from collections import namedtuple
+from itertools import chain
 
 app = QtGui.QApplication([])
 mw = QtGui.QMainWindow()
@@ -115,28 +116,34 @@ s4.sigClicked.connect(clicked)
 
 # Enable hovering and show a tool tip hovered points
 s4.setAcceptHoverEvents(True)
-lastHovered = []
+hoverPoints = []
 hoverPen = pg.mkPen('g')
+hoverRect = None
 
 def hoverEvent(ev):
-    global lastHovered
-    for pt in lastHovered:
-        pt.resetPen()
+    global hoverPoints, hoverRect
 
-    if not ev.exit:
-        points = s4.pointsAt(ev.pos())
-        for pt in points:
+    if hoverRect is None and len(hoverPoints) > 0:
+        hoverRect = hoverPoints[0]._data['sourceRect']  # avoid gc to prevent removal of glyph from cache
+
+    newPoints = [] if ev.exit else s4.pointsAt(ev.pos())
+
+    for pt in chain(hoverPoints, newPoints):
+        if pt in newPoints and pt not in hoverPoints:
             pt.setPen(hoverPen)
-        lastHovered = points
 
-        cutoff = 3
-        tip = '\n\n'.join('index: {}\nx: {:.3g}\ny: {:.3g}'.format(pt.index(), pt.pos().x(), pt.pos().y())
-                          for pt in points[:cutoff])
-        if len(points) > cutoff:
-            tip += '\n\n({} more...)'.format(len(points) - cutoff)
+        elif pt in hoverPoints and pt not in newPoints:
+            pt.resetPen()
 
-        if tip:
-            s4.getViewBox().setToolTip(tip)
+    cutoff = 3
+    tip = '\n\n'.join('index: {}\nx: {:.3g}\ny: {:.3g}'.format(pt.index(), pt.pos().x(), pt.pos().y())
+                      for pt in newPoints[:cutoff])
+    if len(newPoints) > cutoff:
+        tip += '\n\n({} more...)'.format(len(newPoints) - cutoff)
+
+    s4.getViewBox().setToolTip(tip)
+
+    hoverPoints = newPoints
 
 s4.hoverEvent = hoverEvent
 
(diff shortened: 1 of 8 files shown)


j9ac9k:
Looks like one of the examples is hanging when running on pyside2/conda-forge instances.  I recently merged #1302 which did some changes on how the test application is loaded, and how the test suite is ran in context of the examples.  Can you replicate locally?

lidstrom83:
I noticed I used a format string in the example I changed and pushed a compatibility fix while the checks were still running. Could this have something to do with it?

lidstrom83:
Just found some more room for improvement. Should get a ~2x boost in pxMode in the next commit (~3x overall). It'll change a fair bit though, so should it be put in a separate PR?

lidstrom83:
I ended up pushing the extra commit as it turned out to be a pretty simple change after all.

j9ac9k:
Looks like those pipelines are hanging, I'll try to test out locally and see if I can repliace.

j9ac9k:
I can confirm the issue locally, I get a crash when running the `Plotting.py` test example...


```
pyqtgraph-pyside2-conda ❯ pytest -k "Plotting.py"
==================================================================== test session starts =====================================================================
platform darwin -- Python 3.7.7, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
rootdir: /Users/ogi/Developer/pyqtgraph, inifile: pytest.ini
plugins: forked-1.2.0, cov-2.10.0, xdist-1.32.0
collected 156 items / 155 deselected / 1 skipped

examples/test_examples.py Timeout (0:00:15)!
Thread 0x000000010d9dddc0 (most recent call first):
  File "/Users/ogi/Developer/pyqtgraph/examples/test_examples.py", line 222 in testExamples
...
```


passing `--full-trace` argument, I get the following non-helpful information:

```
    """ % (import1, graphicsSystem, import2)
        if sys.platform.startswith('win'):
            process = subprocess.Popen([sys.executable],
                                        stdin=subprocess.PIPE,
                                        stderr=subprocess.PIPE,
                                        stdout=subprocess.PIPE)
        else:
            process = subprocess.Popen(['exec %s -i' % (sys.executable)],
                                       shell=True,
                                       stdin=subprocess.PIPE,
                                       stderr=subprocess.PIPE,
                                       stdout=subprocess.PIPE)
        process.stdin.write(code.encode('UTF-8'))
        process.stdin.close()
        output = ''
        fail = False
        while True:
            try:
>               c = process.stdout.read(1).decode()
E               KeyboardInterrupt
```

Running pytest without test_examples, I get a setfault on `test_ScatterPlotItem.py`


```
$ pytest -k "not test_examples.py" -v
...
pyqtgraph/graphicsItems/tests/test_ScatterPlotItem.py::test_scatterplotitem zsh: segmentation fault  pytest -k "not test_examples.py" -v
```

These are always tricky to debug...

lidstrom83:
> These are always tricky to debug...

Oh joy. Installing Miniconda now.

BTW, has a timeline been decided yet for when to drop support for Python2.7 and Qt4?

j9ac9k:
> Oh joy. Installing Miniconda now.

I suggest looking here to see  how we configure our conda environment and just replicate that:  https://github.com/pyqtgraph/pyqtgraph/blob/master/azure-test-template.yml#L93-L113

> BTW, has a timeline been decided yet for when to drop support for Python2.7 and Qt4?

I want to do it after the number of outstanding PRs gets to be 5-10ish... feel free to join our slack channel if you have some opinions on that regard.

pyqtgraph/functions.py
```diff
@@ -386,7 +386,16 @@ def glColor(*args, **kargs):
...
-    
+
+def serializeQtData(obj):
+    """
+    Serializes any Qt data type that can be serialized into a QDataStream using the left-shift operator.
+    """
+    a = QtCore.QByteArray(b'')
+    s = QtCore.QDataStream(a, QtCore.QIODevice.WriteOnly)
+    s << obj
+    return bytes(a)
```
j9ac9k, line 397: this is the line generating the segmentation fault.

j9ac9k: I think you want to return `a.data()`

j9ac9k: after changing this to return `a.data()` the test suite passed for me 👍

lidstrom83: Thanks for digging into this.

lidstrom83:
I just ran the scatter plot speed test at various stages throughout this PR and realized the pxMode=True gains from various commits just bring us back to roughly the original performance; serializing objects rather than using their id as the key in SymbolAtlas.getSymbolCoords was initially a big performance hit, and I was accidentally comparing to speeds *after* doing this. ~Doubling of the speed test for pxMode=False still appears to be correct. Apologies for overselling this.

pyqtgraph/functions.py
j9ac9k: no problem, thanks for the PR and incorporating the fix 👍

j9ac9k:
> I just ran the scatter plot speed test at various stages throughout this PR and realized the pxMode=True gains from various commits just bring us back to roughly the original performance; serializing objects rather than using their id as the key in SymbolAtlas.getSymbolCoords was initially a big performance hit, and I was accidentally comparing to speeds _after_ doing this. ~Doubling of the speed test for pxMode=False still appears to be correct. Apologies for overselling this.

Thanks for following up on this.  Would say this issue still resolves #1020 ?

j9ac9k:
Also, can you explain the motivation behind serializing the objects and storing the serialized representation vs. just storing the  objects themselves in the cache dictionary?

lidstrom83:
> Thanks for following up on this. Would say this issue still resolves #1020 ?

Yes, this issue is sill resolved.

> Also, can you explain the motivation behind serializing the objects and storing the serialized representation vs. just storing the objects themselves in the cache dictionary?

Initially I noticed that in the hovering example I added, SymbolAtlas.buildAtlas was getting called every time ScatterPlotItem.updateSpots was called, and that SymbolAtlas.symbolMap was growing without bound. The cache was missing when it shouldn't in part due to recreating new default pen and brush objects during each update. I decided to use serializations of pens and brushes in the symbolMap lookup key rather than their id to avoid this problem.

I did consider instead to just avoid repeatedly recreating the default pen and brush, but realized taking this kind of care to avoid cache misses would extend to the user as well. For example, in the hover example I would have had to be careful to reuse the hovering QPen object during each update. Additionally, the SpotItem.setPen method would need to be changed to avoid creating a new QPen object. The approach I took is more forgiving to the user and avoids having to change a bunch of methods and the API. And the initially large performance hit was eventually recovered from by caching the serialized Qt objects while looping over the spots in SymbolAtlas.getSymbolCoords.

For the other side of the trade-off between caching approaches: keeping all these changes but switching back to keying by id appears to boost the speed test by ~10% with pxMode=True.

j9ac9k:
I would have never thought to serialize the objects like that!

I'm not suggesting you go undo the work, but if the problem with the cache hits involve multiple calls to `function.mk<whatever>()` return objects with different IDs for the same input, would it not be easier to get around the cache hit issue by using the `@lru_cache` decorator on those functions?  Not only would the return of those functions be _significantly_ faster, but they should be returning objects with the same ID as well.

lidstrom83:
>  I'm not suggesting you go undo the work, but if the problem with the cache hits involve multiple calls to `function.mk<whatever>()` return objects with different IDs for the same input, would it not be easier to get around the cache hit issue by using the `@lru_cache` decorator on those functions? 

During this PR I played around with this library's `LRUCache` after discovering Python 2.7 doesn't have one. I think I was trying to cache the serialized Qt objects. I didn't think to cache *these* functions though, and would've been scared to mess with them considering how much code they touch.

Thinking more about the serialized keying: if every spot's pen or brush is different then the caching of serialized pens and brushes doesn't help alleviate the performance hit with this approach. This makes me less comfortable with the trade-off.

I'm going to try out your idea now.

j9ac9k:
I can't think of any harm of wrapping `mkPen`, `mkBrush` and so on with the `@lru_cache` decorator (other than slightly increased memory usage depending on the size of the cache we allocate); I never really thought that they were being hit that often so never thought that the library should cache them, but for profiling purposes we should definitely explore it.

Regarding python 2.7 compatibility; ...we'll worry about that later 😆

lidstrom83:
> TypeError: unhashable type: 'QPen'

Right... We need hashable arguments when using `@lru_cache`. Often e.g. `mkPen` is called on an existing `QPen` object. We could just return QPen in this case.

j9ac9k:
> > TypeError: unhashable type: 'QPen'
> 
> Right... We need hashable arguments when using `@lru_cache`. Often e.g. `mkPen` is called on an existing `QPen` object. We could just return QPen in this case.

oof...while this wouldn't help us, we should consider reaching out to the pyside2 folks and see if they would consider implementing `__hash__` on those objects...

j9ac9k:
there is a SO post with a lru_cache that ignores the unhashable types... not suggesting we integrate this; but should help with getting some profiling data to see if it's worthwhile:

https://stackoverflow.com/a/64111268

lidstrom83:
Calling the original function on an unhashable input would still create a new copy of the input, and thus not resolve the issue. We'd need to return the argument itself in this case, but then mutating this output later could cause all kinds of bugs in existing code, since people have been relying on these functions returning copies.

I tried manually wrapping the `mkPen`, `mkBrush`, and `mkColor` functions like this. Some details are `mkBrush` can take a `QColor` as a single argument and `mkPen` can take a `dict` (unhashable too). In these cases I just bit the bullet and called the uncached function. Better than nothing. It seems to work, but I'm very hesitant to replace the public functions with these cached versions.

j9ac9k:
you're right, at the end of the day, the QPen, QBrush, etc are mutable objects, and as they are valid inputs...this is likely a bigger head ache than what we want any part of...

lidstrom83:
I ended up switching away from the serialization-based keying of the atlas as it proved to be a performance bottleneck in some cases. To ensure effective caching, the user should reuse `QPen` and `QBrush` objects as much as possible, and also maintain references to any `sourceRect` that might be used again later so they aren't garbage collected and subsequently dropped from the atlas. The hovering example show's what I mean.

Considering how much back-and-forth there is in this branch, should I squash the commits?

examples/ScatterPlot.py
```diff
@@ -113,6 +114,38 @@ def createLabel(label, angle):
 w4.addItem(s4)
 s4.sigClicked.connect(clicked)
 
+# Enable hovering and show a tool tip hovered points
+s4.setAcceptHoverEvents(True)
+hoverPoints = []
```
j9ac9k, line 119: since we are doing `in` lookups, should this be a set instead of a list (same with `newPoints`?

j9ac9k:
We'll squash on merge, so I wouldn't worry about that, but I do think we should document attempts to re-use brush/pen objects, maybe in the docstring for `ScatterPlotItem` ?

Should we modify `ScatterPlotItem` to maintain a reference to `sourceRect` so it doesn't get garbage collected (that way we don't have to push that responsibility to the user)?

examples/ScatterPlot.py
lidstrom83: I did do that at first but used lists instead because I was slicing during the construction of the tool tip. The hovered points are returned in the reverse order in which they're drawn, so those drawn on top are prioritized for tool tip. That said, you're still right. Commit incoming.

j9ac9k: hmm....if order matters, maybe you can just use a `OrderedDict`, and just use the keys, and ignore the values? 

I don't have a feel at all if this is a performance bottleneck or not, so sticking with lists may be perfectly fine (I don't know how large those containers get so maybe the difference O(n) and O(1) may not be that different.

Maybe we should embed a `debug.Profiler` in this function so can do some benchmarking?  (I'm watching my two kids during the days which is why I haven't been able to experiment with this more. will try to mess with it more tonight).

lidstrom83:
> We'll squash on merge, so I wouldn't worry about that, but I do think we should document attempts to re-use brush/pen objects, maybe in the docstring for `ScatterPlotItem` ?

Will do.

> Should we modify `ScatterPlotItem` to maintain a reference to `sourceRect` so it doesn't get garbage collected (that way we don't have to push that responsibility to the user)?

I thought about not setting the `sourceRect` to `None` to implicitly invalidate spots and instead use a `'valid'` field. This field would then be used for the mask here:

https://github.com/pyqtgraph/pyqtgraph/blob/da0d90299ad79b0b6c34ca6a88066763031b9a5a/pyqtgraph/graphicsItems/ScatterPlotItem.py#L608-L613

This would be an improvement, I think, but it wouldn't cover all cases. E.g. in the hovering example, when you move from hovering over no spots to then hovering over some spots, an atlas rebuild would still be triggered if we don't manually keep reference to the `sourceRect` for the the hovered spot style.

Generally, it's impossible to know which glyphs the user plans to reuse and which ones are safe to purge from the atlas.

A blunt approach to making this automatic for the user could be an option to use a regular dict here: https://github.com/pyqtgraph/pyqtgraph/blob/da0d90299ad79b0b6c34ca6a88066763031b9a5a/pyqtgraph/graphicsItems/ScatterPlotItem.py#L140
It would then be on the user to not let the atlas fill up with useless entries.

A safer approach could be to additionally store the `sourceRect`s in a bounded LRU cache so that they'd be purged from the `symbolMap` only after being bumped from from the LRU cache.

Any approach will degrade performance during atlas rebuilds proportional to how many never-to-be-used-again glyphs end up lingering around. It probably wouldn't be noticeable for small cache sizes, and that probably covers most use-cases.

lidstrom83:
I tried implementing the lru cache idea using the one from this library, but it proved to have significant overhead when the atlas is large due to frequent resizing. I then opted to just append the entries in a `deque`. This is simpler, has barely any overhead, and it appears to support many common use-cases such as the hovering example. I haven't explored choice of `maxsize=20` much yet, so please let me know if another choice seems a better trade-off.

lidstrom83:
I should mention, the approach I took is limited to be useful only when less than `maxlen` distinct glyphs are used. If e.g. distinct sizes are set for each one, then this approach will fail quickly.

lidstrom83:
I have to apologize again for overselling this. Issue #1020 is not actually solved.

The goal of this PR has become less ambitious, namely to add a hovering demo and the performance optimizations necessary to make it work. I've changed the title accordingly.

j9ac9k:
No need to apologize, I do want to try and help, would you mind communicating outside of this issue tracker?  Here is a link to hop onto our slack group.

Join me on Slack — it’s a faster, simpler way to work. Sign up here, from any device: https://join.slack.com/t/pyqtgraph/shared_invite/zt-je6aowiv-dWuX2bhzq8w2GsCsPlzxKw

Alternatively feel free to reach out via email.

pyqtgraph/graphicsItems/ScatterPlotItem.py
```diff
@@ -629,6 +722,25 @@ def updateSpots(self, dataSet=None):
...
+            if col.base is not None:
+                col = col.copy()
+            col[np.equal(col, _DEFAULT_STYLE[opt])] = self.opts[opt]
+
+            if opt == 'size' and scale is not None:
+                col *= scale
+
+            yield col
+
+    # deprecated
```
j9ac9k, line 743: If actually deprecated; we should emit a deprecation warning.

lidstrom83: I'm going to take that comment out. This method is replaced internally by _style, but there isn't a public replacement.

pyqtgraph/graphicsItems/ScatterPlotItem.py
```diff
@@ -103,138 +120,191 @@ def makeSymbolPixmap(size, pen, brush, symbol):
     img = renderSymbol(symbol, size, pen, brush)
     return QtGui.QPixmap(img)
 
+
+def _mkPen(*args, **kwargs):
```
j9ac9k, line 124: I know you explain in the comment section above about re-using pens for performance reasons; but there should be doc-strings here for these helpers methods explaining why they're here.

pyqtgraph/graphicsItems/ScatterPlotItem.py
```diff
@@ -103,138 +120,191 @@ def makeSymbolPixmap(size, pen, brush, symbol):
...
+        self.clear()
+        if data:
+            self._extendFromData(data)
 
-        self.atlasRows = []
+    def clear(self):
+        self.__init__()
+
+    def diagnostics(self):
+        n = len(self._coords)
```
j9ac9k, line 202: since we're implementing `__len__` this can be replaced with `len(self)`

pyqtgraph/graphicsItems/ScatterPlotItem.py
```diff
@@ -103,138 +120,191 @@ def makeSymbolPixmap(size, pen, brush, symbol):
...
+        # insert data into array
+        for key, arr in data:
+            y, x, h, w = self._coords[key]
+            self._data[x:x+w, y:y+h] = arr
+
+        self._pixmap = None
+
+    def _pack(self, data):
+        # pack each item rectangle as efficiently as possible into a larger, expanding, approximate square
+        n = len(self._coords)
```
j9ac9k, line 255: can replace with `len(self)`

lidstrom83:
Going through the open issues and pull requests, I've checked that this resolves #176, #540, #1020, #1209 (with a small change in the author's example to cache `QBrush` objects), and it obsolete #1341.

lidstrom83:
This also resolves #421 as well now.

lidstrom83:
And #420.

j9ac9k:
geez, leave some issues for the rest of us 👍

j9ac9k:
Something's not quite right:

```bash
python examples/ScatterPlot.py
```

From current master:

image

From `scatter-performance`

image

lidstrom83:
Thanks for the screenshots. I can't seem to reproduce this. I've tried on my Linux laptop with Python 3.8 using both PyQt5 and PySide2, with and without an external monitor plugged into it.

j9ac9k:
Running `git bisect` on your branch pointed the first commit with the issue being c6c404371dfa3cba71aeb3162b267c15ef97fd89

j9ac9k:
I can confirm on macOS the scatter plots behave well, as does the exporter, nice work!!  I'll try to break this a bit more a little bit later today.

lidstrom83:
#915 is also related.

I treated the missing `invalidate` as a bug in order to match the behaviour when `pxMode=True`, and because the user can specify `update=False` in the setters.

Regarding not emitting `sigPlotChanged`, that seemed odd to me too but I also worry about breaking users' code by adding it. Similar to [this suggestion](https://github.com/pyqtgraph/pyqtgraph/issues/915#issuecomment-521664487), I implemented an alternative API with `__getitem__`, `__setitem__` and `setOpts` methods where I emit this signal. I've since reverted it, but now that this PR has become monolithic, I could add it back in order to fully resolve the issue.

lidstrom83:
And another: #1118. I haven't implemented a hovering API in `PlotDataItem`, but that would be fairly easy to do so now. It would just pass the hovering opts `hoverable`, `hoverPen`, `hoverBrush`, `hoverSize`, `hoverSymbol`, and `tip`along to the underlying `PlotCurveItem`.

lidstrom83:
Does anyone have any thoughts about deprecating the `dataSet` and `mask` arguments in the setters? The former seems to be a foot gun - it silently fails unless `dataSet` is a `view` of `self.data` - and the latter is just doing e.g. `dataSet['pen'] = pens[mask]`, so why wouldn't the user just pass in `pens[mask]` in the first place? I could replace these with an `index` argument which could be a `slice`, `mask`, or `array` of indices.

j9ac9k:
> And another: #1118. I haven't implemented a hovering API in `PlotDataItem`, but that would be fairly easy to do so now. It would just pass the hovering opts `hoverable`, `hoverPen`, `hoverBrush`, `hoverSize`, `hoverSymbol`, and `tip`along to the underlying `PlotCurveItem`.

I think this would be great, but I don't think this has to be part of _this_ PR, as I anticipate merging shortly; ...

> Regarding not emitting sigPlotChanged, that seemed odd to me too but I also worry about breaking users' code by adding it. Similar to this suggestion, I implemented an alternative API with __getitem__, __setitem__ and setOpts methods where I emit this signal. I've since reverted it, but now that this PR has become monolithic, I could add it back in order to fully resolve the issue.

I think we should add this in, but like the previous comment, doesn't have to be part of this specific PR, but another one that addresses this specific issue.  If there is any kind of potential API breakage, having it on its own PR will make it easier to track down later should there be an issue made in the future.

j9ac9k:
This is a monster PR, thanks for your work on this @lidstrom83 now I have to go through and identify which issues/existing PRs I can close as a result 😆  merging.

PR #1422, opened 2020-10-27
Permit entry of non-finite values into float SpinBox

lidstrom83:
Bounds are enforced for `inf `and `-inf`, but `nan` ignores them. Attempting to step any of these values doesn't do anything.

A previously existing bug is also fixed where the text is not updated properly when out-of-bounds data is entered.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -38,7 +38,7 @@ SI_PREFIXES_ASCII = 'yzafpnum kMGTPEZY'
 SI_PREFIX_EXPONENTS = dict([(SI_PREFIXES[i], (i-8)*3) for i in range(len(SI_PREFIXES))])
 SI_PREFIX_EXPONENTS['u'] = -6
 
-FLOAT_REGEX = re.compile(r'(?P<number>[+-]?((\d+(\.\d*)?)|(\d*\.\d+))([eE][+-]?\d+)?)\s*((?P<siPrefix>[u' + SI_PREFIXES + r']?)(?P<suffix>\w.*))?$')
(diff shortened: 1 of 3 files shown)


j9ac9k:
Looks like a python2 issue:

```
    from math import isnan, isfinite
ImportError: cannot import name isfinite
```

Any reason not to use the `numpy.isfinite()`?

lidstrom83:
> Any reason not to use the numpy.isfinite()?

Just a habit of preferring standard library functions when operating on single values. Sometimes the numpy alternatives are much slower when operating on single values. Also, numpy hadn't been imported in this module yet and I didn't want to make it harder to reuse this code without it.

j9ac9k:
Truth be told I too have the habit for using the math module for single values, I made that suggestion not knowing there was a good py2/3 compatible function.

pyqtgraph/functions.py
```diff
@@ -38,7 +38,7 @@
 SI_PREFIX_EXPONENTS = dict([(SI_PREFIXES[i], (i-8)*3) for i in range(len(SI_PREFIXES))])
 SI_PREFIX_EXPONENTS['u'] = -6
 
-FLOAT_REGEX = re.compile(r'(?P<number>[+-]?((\d+(\.\d*)?)|(\d*\.\d+))([eE][+-]?\d+)?)\s*((?P<siPrefix>[u' + SI_PREFIXES + r']?)(?P<suffix>\w.*))?$')
+FLOAT_REGEX = re.compile(r'(?P<number>[+-]?((((\d+(\.\d*)?)|(\d*\.\d+))([eE][+-]?\d+)?)|((?i)(nan)|(inf))))\s*((?P<siPrefix>[u' + SI_PREFIXES + r']?)(?P<suffix>\w.*))?$')
```
j9ac9k, line 41: Just adding a comment here, you don't need to update the PR (unless you want to) but we should _really_ make use of `re.verbose` for these more complex regexes.

j9ac9k:
Thanks for the PR @lidstrom83 this looks good to me, merging.

PR #1423, opened 2020-10-31
Fix small bug in GridItem.py:setTextPen

ernierock:
Bug in GridItem.setTextPen() : on line 49 **kargs should be **kwargs. setTextPen does not work (until this simple fix is applied)


j9ac9k:
thanks for this PR @ernierock !

Issue #1407, opened 2020-10-21
Feature - Add Available Colors


j9ac9k:
PyQtGraph currently has only minimal numbers of colormaps or specified colors baked in the library, that functionality really should expand

Bokeh for example offers an extensive list of palettes and colors (some of which are sources from CSS named colors, some are sourced from [brewer](https://colorbrewer2.org/#type=sequential&scheme=BuGn&n=3):

https://docs.bokeh.org/en/latest/docs/reference/palettes.html
https://docs.bokeh.org/en/latest/docs/reference/colors.html

In addition, we should consider adding the colors from the following sources

-  QML [QtQuick.Controls.Material](https://doc.qt.io/qt-5/qtquickcontrols2-material.html)
- [Complete 2014 Material Design Color Listing](https://material.io/design/color/the-color-system.html#tools-for-picking-colors)
- [iOS System Colors](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/color/)
- [Microsoft Colors](https://docs.microsoft.com/en-us/uwp/api/windows.ui.colors?view=winrt-19041)
- [CSS Color Names](https://www.w3schools.com/colors/colors_groups.asp)
- [Color Brewer](https://colorbrewer2.org/#)
- [SVG Color Keywords](https://www.w3.org/TR/SVG11/types.html#ColorKeywords)

Now, how do we organize all these colors/colormaps?

ntjess:
How would you feel about `matplotlib` colormaps as well? My application currently has this dependency due to the lack of cyclical colormaps natively available in `pyqtgraph`. I know there is a list in `GradientEditorItem`, but these are hard to use for regular LUTs unless you are explicitly creating a `GradientEditorItem`.

E.g. a `maptlotlib` style colormap where you can call `cmap([3,4,5])` and get a 3x4 array of RGBA colors.

Would this request fall under the same issue? Thanks for considering.

j9ac9k:
Adding mpl colormaps would totally fall under this umbrella! Only thing I'm not sure about is how to organise the colors, and how to make them accessible via API.

Bokeh would likely be the best example here, that or matplotlib itself.

NilsNemitz:
I would also like to see a convenient way to handle custom color palettes and perceptually uniform color maps in pyqtgraph.
For discussion, I have created some proof-of-concept code in PR #1424.

[...]

PR #1424, opened 2020-10-31
Colormap extension

NilsNemitz:
Proof of concept for addressing 
#1407 Feature - Add Available Colors
by extending ColorMap object.

The function colormap.load() directly creates a ColorMap object from color map / palette files.
Such files can be distributed with pyqtgraph (now in colormaps/ folder), or can be supplied by the user.

Optional color names can be provided in the same files. Color can be used directly from the ColorMap object by making it indexable (`__getitem__`). Using different ColorMap objects for different palettes avoids a cluttered namespace:
`cm1=pg.colormap.load('xkcd_rgb.txt')`
`cm2 = pg.colormap.load('CET-C1s.csv')`
`pen=cm1['ugly purple']`
`symbolBrush=cm2[42]`


NilsNemitz:
Here is some example code:
[cm_demo_code.zip](https://github.com/pyqtgraph/pyqtgraph/files/5469637/cm_demo_code.zip)

j9ac9k:
Hi @NilsNemitz thanks for this PR.  Expanding on the `ColorMap` object was a great call.

How would it work for handling directories that are several layers nested like the Material colors?

Here is a subset of them:  https://doc.qt.io/qt-5/qtquickcontrols2-material.html

Note that they are defined by `Material.<color-name>` but have different colors depending on light vs. dark mode.

Also I'm not feeling great about colors being loaded into python objects at runtime; but I do think we should include the source files (along w/ various licenses). 

I'll try and experiment w/ this more in the next few days by adding the Color Brewer colors, and include more of the matplotlib colormaps as well.

NilsNemitz:
I think including a light/dark mode switch that works on the fly and is transparent to the user might be a little overambitious. That would require ColorMap to handle more meta information. 

If we want to provide the Material colors (I haven't looked at them much yet), then I would propose some method to generate a ColorMap (with names) for each of the sets. An ambitious user might then build an application that conforms to the color scheme by
`if dark_mode:`
`   cm_mat = pg.colormap.load('materials_dark.txt')`
`else:`
`  cm_mat = pg.colormap.load('materials_light.txt')`
Then the main code does not have to differentiate, and cm_mat['red'] will return the appropriate shade.

An upgrade to that would be to include some magic in the loader to check if a palette file is available in 'dark' and 'light' modes, and load the appropriate one. I wanted to add some magic to add the file extension to the loader in any case. E.g. colormap.load('xkcd') would find that 'xkcd.txt' exists and load that with the appropriate parser.

We might add some additional methods to generate ColorMap objects. Pulling the latest and greatest palette (including 'foreground', 'accent' etc.) out of a Qt style would be good. Colorcet as a library can provide its palettes as a list of hex codes. If we have code to use that directly (`cm = pg.colormap.from_hex( colorcet.fire )` , then we can leave the compilation work for exotic palettes to other projects.

For the Color Brewer sets, I would suggest simply treating them as gradients. If e.g. 8 colors are needed, appropriate specific colors can then be found by `cm_cb[idx/7.0]' (with idx running 0 to 7). We don't need to replicate the uneven steps (Tables 1-3 in https://www.cs.rpi.edu/~cutler/classes/visualization/S20/papers/colorbrewer.pdf ), and we don't need to target the "trailing edge" computers of 2003 anymore.

NilsNemitz:
I did, however, grow up with the "trailing edge computers of 2003", and as such it seems very wasteful to me to load a hundred 256-color palettes on startup just because the user might use one or two of them. That is why I prefer loading as required.

If the concern is that a user will edit the palettes and make things crash: The user can also edit any of the python code.
At least crashing specifically on colormap load will give them an idea of what they broke.

If the concern is speed, then my expectation is that the typical use case would be to load one or two ColorMap objects once and use them throughout the application. If rapid start-up is required, maybe python is not the best tool in any case.

The loader could cache color maps / palettes if we find that users tend to reload instead of keeping the ColorMap object around.

j9ac9k:
> I think including a light/dark mode switch that works on the fly and is transparent to the user might be a little overambitious. That would require ColorMap to handle more meta information.

Actually, pyqtgraph already has this functionality in the examples app, but that wasn't what I was trying to imply.  I was meaning, that when selecting a color scheme, whether it be color brewer, material, whatever, ...there could be nested properties in defining them... in the case of material (I'm just using this as an example) you can reference colors via the qtquick controls where you specify `Material.<colorname>` but depending on dark/light mode it would choose a color code; but there is also a grid of colors, based on color name, and some other identifier which I have no idea what it represents, you can see it here:

https://material.io/resources/color/#!/?view.left=0&view.right=0

> An upgrade to that would be to include some magic in the loader to check if a palette file is available in 'dark' and 'light' modes, and load the appropriate one. I wanted to add some magic to add the file extension to the loader in any case. E.g. colormap.load('xkcd') would find that 'xkcd.txt' exists and load that with the appropriate parser.

> I did, however, grow up with the "trailing edge computers of 2003", and as such it seems very wasteful to me to load a hundred 256-color palettes on startup just because the user might use one or two of them. That is why I prefer loading as required.

I'm not feeling great about a user having to explicitly load a color map, this also may cause some issues with packaging the wheel as we would have to tell `setup.py` to grab non-python files within certain directories (not saying this is a deal breaker, but just an extra step).  I would be all for having import statements (for importing color palettes/maps) within function calls so that they're not loaded needlessly, but the colors are still embedded in `.py` files.

> If the concern is that a user will edit the palettes and make things crash: The user can also edit any of the python code.
At least crashing specifically on colormap load will give them an idea of what they broke.

If we want to make the color immutable (I generally think this is a good idea) we can have the base object representing the color code be a `namedtuple`.  While someone motivated could still navigate around that, but this is python, we're all adults here...

In terms of caching, we can wrap any kind of call method with `@functools.lru_cache`, so I don't think we need to worry about that either.

NilsNemitz:
I have no experience with packaging, so I trust you on the issues with including non-python files. If we want to include license files (and raw versions of some palettes), we have that problem anyway, though.

What do you imagine the user-side code would look like if we include the palettes as python?
How would it effectively differ from
`cm = pg.colormap.get('inferno')`
`brush = cm[0.3]` ?
If we name it "get" instead of "load", will anybody even notice the difference?
It seems that we agree that (re-)loading speed can be made into a non-issue?

Disregard the comment about "crash on editing". I was just trying to guess your problem with loading at runtime. I would personally consider the option to reach into the data structures and mess with the colors/names/etc not to be a bug.

Supporting multi-dimensional color palettes as in `Material` would be a rather invasive change to how ColorMap works, and it seems that this is not needed for either color maps or conventionally named palettes.

I would therefore try to deflect your question by pointing out that the "shades" seem to be procedurally generated from the primary color, or rather from *any* color. A reply on StackOverflow states that the "lighter" and "darker" variants of the Material color are just standard steps on the CIELAB color scale:
https://stackoverflow.com/questions/30870167/convert-colorprimary-to-colorprimarydark-how-much-darker
And it seems that the numbered shades are just different step sizes. The answer is probably somewhere in here:
https://github.com/edelstone/material-palette-generator/blob/master/js/main.js
...but it makes my head hurt.

So my suggestion would be to include a palette of the Material "primary" colors and a convenience palette of desaturated dark mode colors (these seem to be shades 200 if the primary color is 500):
https://material.io/design/color/dark-theme.html#ui-application

The user can then hack their own additional lighter/darker variants by scaling RGB or HSV values until we come up with a tool that can do this properly. An elegant solution might be an extended QColor class that understands CIELAB color space. Or maybe it is simpler to let Qt do all things Material related.

Err... Could you maybe point me to the light mode / dark mode example so that I can see what level of functionality you have in mind?

Do you have any other application of a 2D color map?

NilsNemitz:
Hmm. It seems that the Materials colors are hand-tuned to "work well together". They do not seem to follow a proper pattern in any color space I can find. If you look at the official colors
https://material.io/resources/color/#!/?view.left=0&view.right=0
then they don't actually make good gradients/color maps in either direction:
- for "hue", blue and light blue are almost indistinguishable, teal is much too dark
- for "lightness", yellow and gray are very unevenly distributed, while red seems to give alternative reds, rather than a clear sequence.

So at least for the Material colors, I would do a "light mode" set, a "dark mode" set, and then also include the full named list from 
https://github.com/material-components/material-components-web/blob/master/packages/mdc-theme/_color-palette.scss
The idea seems to be "pick a few that you like", in any case.

We might add a luminance adjustment to semi-manually scale palettes for light/dark mode. 
`cm.lighten('red', 0.4)`, might take 'red' in the HSL scale and condense the luminance axis to span only 0.4-1.0
`cm.darken('red', 0.4)` might condense the axis to 0.0-0.6.
QColor already supports HSL, so that would not require much code. (Code for CIELAB conversion is available in  
https://github.com/gtaylor/python-colormath/blob/master/colormath/color_conversions.py
but definitely non-trivial.

j9ac9k:
> I have no experience with packaging, so I trust you on the issues with including non-python files. If we want to include license files (and raw versions of some palettes), we have that problem anyway, though.

Good point; I saw matplotlib just has a `License` folder in the top level directory that they put all the license files there, which admittedly makes it a bit easier to manage.

> What do you imagine the user-side code would look like if we include the palettes as python?
How would it effectively differ from

This is a great question, and honestly I'm not sure, I'm sure whatever I say now I'll be taking back later; which is why I've largely been referencing other libraries to see what the color selection looks like from their API

> It seems that we agree that (re-)loading speed can be made into a non-issue?

Yeah I think this will be fine here, we have multiple ways of managing it.

> Supporting multi-dimensional color palettes as in Material would be a rather invasive change to how ColorMap works, and it seems that this is not needed for either color maps or conventionally named palettes.

So I'm starting to think that the Material color palette here may be the wrong example, as you're talking about colormaps (whether they be in gradient form), I think of the material colors, and their respective shades as completely discrete colors; not as a color palette... 

> Err... Could you maybe point me to the light mode / dark mode example so that I can see what level of functionality you have in mind?

For the record, I'm not suggesting we bake any of this functionality into the ColorMap selection stuff, but FYI, you can see how pyqtgraph (or any QApplication for that matter) determines if the host OS is running in light/dark mode.  It also captures a signal (`QApplication.paletteChanged`) to indicate that the host OS may have changed light/dark mode.

https://github.com/pyqtgraph/pyqtgraph/blob/7e57e07068fc8a62e9ae3b68e5556b11b6ef1841/pyqtgraph/Qt.py#L348-L364

https://github.com/pyqtgraph/pyqtgraph/blob/7e57e07068fc8a62e9ae3b68e5556b11b6ef1841/examples/ExampleApp.py#L371

Other 2D colormaps, ...I can't think of any... the only other example I can think of are iOS system colors, which have slightly different color codes depending if used in light or dark mode.

https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/color/

> Hmm. It seems that the Materials colors are hand-tuned to "work well together". They do not seem to follow a proper pattern in any color space I can find. If you look at the official colors

Yeah, I don't think they would work well for any kind of color mapping like color brewer or other common themes; as they would be used individually, not in any kind of collection;... mostly was thinking it would be great if we could specify a material color by colorname-shade and have it be accessible; not just have to define RBG values (which frankly isn't the end of the world).  Storing of individual colors, that do not go necessarily together in a color map sounds to be outside the scope of this PR.

NilsNemitz:
> mostly was thinking it would be great if we could specify a material color by colorname-shade and have it be accessible; not just have to define RBG values

Well, we would get the RGB values if we include Google's _color-palette.scss.txt file. That has "light-green-800" etc. as individual entries. There is no problem with storing and accessing that through ColorMap. The obstacle was that we lose the ordering by color and lightness; But if we just want them as individual colors, that does not matter.

Just pulling the overall design palettes from the system has the disadvantage that it is not clear what colors are defined. The Material list has "amber" and "deep-orange", which are not in Apple's list.

It would be nice to have an equivalent of Apple's Dynamic System Colors (from the page you linked), but that would really need to happen on the Qt side. What we *could* do is provide `pg.colormap.getThemeColors()` that reaches into Qt or the system and makes a palette with entries for foreground, background, accent, secondary, tertiary, etc. These would not update after program start, but would make it convenient to use system colors in the same ColorMap `cm['name']` syntax I am proposing.

Let me know when you get a good idea for what a functional interface might work like. Or if you see an interesting use case that we should consider!

NilsNemitz:
Changes:
- Changed `colormap.load()` to `colormap.get()`
- Added automagic suffix for palette names, e.g. colormap.get('material_style_light') loads 'material_style_light.txt'
- Added Google's Material Style color definitions in dark and light mode versions. Both contain the full list, but default the basic colors ('red') to either shade 500 (light mode) or  shade 200 (dark-mode)
- Added a demo that toggles between two palettes to match light and dark mode.
- Changed divider character for .csv files from whitespace to ';'
- Added support for shorthand hex color codes like '#eee'

NilsNemitz:
Sorry, delayed by other work.
Changes:
Added import for named palettes from Matplotlib and Colorcet
Added all CET data sets (CC-BY license) to local 'colors/maps/' folder for import.
Added simple cache for previously loaded maps
Added examples/colorMaps.py , which loads and shows all available color maps
Support for named colors removed (to become a separate issue/addition)

To do:
Clean up some internal debugging code
Add tests
((Figure out what is going wrong in the PySide test))

j9ac9k:
thank you so much @NilsNemitz for making this happen.  LGTM, merging super experimental feature!

honkomonk:
First of all, thanks for the contribution! More colormaps was something I often missed in pyqtgraph.

I know I'm late for the show, but I had a look at the code and a few remarks for "improvements":

1.) Imho the function `get()` is just an unnecessary layer, that adds confusion. Just make the functions `_get_from_xy()` directly accessible. Their name is self-explanatory and if I have to use `get()` with the "source" parameter, I can just as well type in the specialized functions name.

2.) From a first glance the function  `_get_from_file()` looks like a CVS-parser? If that's the case, I strongly recommend to use one of the builtin parsers from:
  - Python: https://docs.python.org/3/library/csv.html or,
  - Numpy: https://numpy.org/doc/stable/reference/generated/numpy.genfromtxt.html#numpy.genfromtxt or,
  - save the data as Python-file.
Those are battle-proven and will save yourself the maintenance of this part of code.

j9ac9k:
hi @honkomonk thanks for chiming in.  Our colormap functionality is pretty limited right now, and we'll likely be changing it up.   Getting input from users that would be using this part of the library is very important to us.

PR #1428, opened 2020-11-05
default colors for enum ColorMap type

scseeman:


j9ac9k:
CI failed on a test that intermittently glitches out... re-running the failed pipeline.

pyqtgraph/widgets/ColorMapWidget.py
```diff
@@ -71,7 +71,15 @@ def addNew(self, name):
         defaults = fieldSpec.get('defaults', {})
         for k, v in defaults.items():
             if k == 'colormap':
-                item.setValue(v)
+                if mode == 'range':
+                    item.setValue(v)
+                if mode == 'enum':
```
j9ac9k, line 76: let's change this second `if` to be `elif`

j9ac9k:
Hi @scseeman thanks for the PR.  I just had the one suggestion I added as an inline comment, and besides that, I'm good with everything else.

pyqtgraph/widgets/ColorMapWidget.py
scseeman: Done!

j9ac9k:
Thanks for the PR @scseeman this looks good to me, merging!

Issue #1373, opened 2020-09-20
Adding TextItem to plot changes the axis range (appears on Ver. 0.11, the bug was not present on 0.10)


3fon3fonov:
Now that I am migrating to Version 0.11, I have noticed an annoying problem that was not present on 0.10.

When I add ArrowItem and TextItem to an existing plot, I get this:

![Ticks_wrong](https://user-images.githubusercontent.com/44244057/93709887-6474a200-fb42-11ea-84fe-a27563fa9581.png)

Whereas I would expect this (pyqtgraph ver. 0.10):

![Ticks_right](https://user-images.githubusercontent.com/44244057/93709898-77877200-fb42-11ea-9c90-0f4aff56483a.png)

I noticed that this happens only if I label the peaks by calling:

```
    def label_peaks(self, plot_wg2, pos_peaks, GLS = True, o_c = False, activity = False, MLP = False, DFT=False):

        if GLS == True and DFT == False and self.avoid_GLS_RV_alias.isChecked():
            x_peaks = pos_peaks[0][pos_peaks[0]>1.2]
            y_peaks = pos_peaks[1][pos_peaks[0]>1.2]
        else:
            x_peaks = pos_peaks[0]
            y_peaks = pos_peaks[1]


        if GLS == True:
            N_peaks = int(self.N_GLS_peak_to_point.value())
            if o_c == True:
                log = self.radioButton_RV_o_c_GLS_period.isChecked()
            elif activity == True:
                log = self.radioButton_act_GLS_period.isChecked()
            elif MLP == True:
                log = self.radioButton_RV_MLP_period.isChecked()
                N_peaks = int(self.N_MLP_peak_to_point.value())
            elif DFT == True:
                log = self.radioButton_RV_WF_period.isChecked()
                N_peaks = int(self.N_window_peak_to_point.value())

            else:
                log = self.radioButton_RV_GLS_period.isChecked()

            type_per = "GLS"
        else:
            N_peaks = int(self.N_TLS_peak_to_point.value())
            log = False
            type_per = "TLS"

        if len(x_peaks) <  N_peaks:
            N_peaks = len(x_peaks)
            print("You have reached the maximum number of %s peaks."%type_per)

        for i in range(N_peaks):

            text_arrow = pg.TextItem("", anchor=(0.5,1.9))

            if log == True:
                arrow = pg.ArrowItem(pos=(np.log10(x_peaks[i]), y_peaks[i]), angle=270)
[...]

PR #1435, opened 2020-11-12
Fix autosize not taking to correct range with TextItem in view

j9ac9k:
When pressing the auto-scale button, the scale can be taken to an incorrect scale, or if you zoom out _way_ too far, it can be taken to some bogus range.

Issue #1373 discusses a little more of the troubleshooting effort, eventually stumbled across the the first-call to ViewBox.updateMatrix in ViewBox.resizeEvent was causing the issue.

### DIFF
--- a/pyqtgraph/graphicsItems/TextItem.py
+++ b/pyqtgraph/graphicsItems/TextItem.py
@@ -150,7 +150,6 @@ class TextItem(GraphicsObject):
 
         
     def boundingRect(self):
-        # return self.textItem.mapToParent(self.textItem.boundingRect()).boundingRect()
         return self.textItem.mapRectToParent(self.textItem.boundingRect())
 
     def viewTransformChanged(self):
--- a/pyqtgraph/graphicsItems/TextItem.py
+++ b/pyqtgraph/graphicsItems/TextItem.py
@@ -4,9 +4,6 @@ from ..Point import Point
 from .. import functions as fn
 from .GraphicsObject import GraphicsObject
 
-import inspect
-from pprint import pprint
-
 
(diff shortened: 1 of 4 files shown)


j9ac9k:
There's always that one config that fails on that one test....rerunning the test didn't fix the issue unfortunately... I should look at this test a bit more carefully

j9ac9k:
This should do it, I couldn't come up with a test that worked reliably unfortunately... hopefully this doesn't break stuff, as ViewBox.py is crazy sensitive.

misantroop:
This commit causes a 30-40% slowdown on ViewBox update in my case. Something related to the removal of update() and prepareForPaint() called in paint().

j9ac9k:
> This commit causes a 30-40% slowdown on ViewBox update in my case. Something related to the removal of update() and prepareForPaint() called in paint().

You're not the first person to suggest this, and I'm working on some potential fixes, but I've had a tough time with a benchmark showing a difference that isn't a rounding error.  If you have a use-case that demonstrates a 30-40% difference in performance, can you please post it to #1602 so I can test with that?

misantroop:
> 
> 
> > This commit causes a 30-40% slowdown on ViewBox update in my case. Something related to the removal of update() and prepareForPaint() called in paint().
> 
> You're not the first person to suggest this, and I'm working on some potential fixes, but I've had a tough time with a benchmark showing a difference that isn't a rounding error. If you have a use-case that demonstrates a 30-40% difference in performance, can you please post it to #1602 so I can test with that?

Project too big to post. What about the PlotSpeedTest with autoscale example? I get 155 fps in 0.11.0 vs 95 fps in 0.11.1.

PR #1437, opened 2020-11-13
LegendItem: fix docs & minor bugs

drewrisinger:
For ``pyqtgraph.LegendItem``s:
* [x] Removed unused variables & arguments
* [x] fix typos
* [x] Document ``colCount`` usage
* [x] small test updates
* [x] update rowCount as items are added


j9ac9k:
Hi @drewrisinger  thanks for the PR.  Sorry it's taken me a bit to comment, looks like one test is failing:

../pyqtgraph/graphicsItems/tests/test_LegendItem.py:82: AssertionError

```
>       assert legend.rowCount == 2
E       assert 3 == 2
E         -3
E         +2
```

Once that issue is resolved, this PR will likely be ready to merge. 

Thanks!

pyqtgraph/graphicsItems/LegendItem.py
```diff
@@ -209,13 +214,15 @@ def _addItemToLayout(self, sample, label):
...
+        # Keep rowCount in sync with the number of rows if items are added
+        self.rowCount = max(self.rowCount, row + 1)
 
     def setColumnCount(self, columnCount):
         """change the orientation of all items of the legend
         """
         if columnCount != self.columnCount:
             self.columnCount = columnCount
-            self.rowCount = int(len(self.items) / columnCount)
+            self.rowCount = math.ceil(len(self.items) / columnCount)
```
j9ac9k, line 225: this will return a float in python2, so we should either cast as an `int()` or do `len(self.items) // columnCount + len(self.items) % columnCount`

drewrisinger: Good catch on the py2 difference. I'll cast to that. makes the logic clearer.

dgoeries:
Was on my list to tackle, thanks for doing this. Why do we keep the rowCount anyway`?

j9ac9k:
> Was on my list to tackle, thanks for doing this. Why do we keep the rowCount anyway`?

Sorry, just saw this, no clue why we do it!

j9ac9k:
Logic here isn't quite right; when you modify `examples/Legend.py` such that this is how you create the legend:

```python
legend = pg.LegendItem((80,60), offset=(70,20), colCount=2)
```

this is what you get:

image

j9ac9k:
Bug with placement earlier wasn't due to this PR, but due to some incorrect logic in a previous PR that added the column feature; ...but I just appended the fix to this PR.

Assuming test suite passes, I'm going to merge.  Thanks @drewrisinger for this PR, sorry I wasn't able to provide better help earlier on with resolving the issue.

PR #1439, opened 2020-11-16
Fix: TickSliderItem method uses function from subclass GradientEditorItem

2xB:
This fixes an issue where using `TickSliderItem.setTickValue` fails because it references a `GradientEditorItem` function. Further, since I was on it, the code is now improved do make much more use of signals and slots to avoid code duplication and weak references. This made it possible to reduce the amount of `GradientEditorItem.updateGradient` calls upon `GradientEditorItem` initialization from 34 to 2 in my tests.

Test code to reproduce error:

```python3
import pyqtgraph as pg
import numpy as np

app = pg.mkQApp()

        self.item = pg.TickSliderItem(*args, allowAdd=False, **kargs)
        tick = self.item.addTick(0.5, color="w", movable=False)
        self.item.setTickValue(tick, 0.05)

        self.setFixedHeight(31)


w = CustomWidget()
w.show()

app.exec_()

```

Test code to show that nothing broke:

```python3
import pyqtgraph as pg
import numpy as np

app = pg.mkQApp()

        self.item = pg.GradientEditorItem(*args, **kargs)
        tick = self.item.addTick(0.5)
        self.item.setTickValue(tick, 0.05)

        self.setFixedHeight(31)


w = CustomWidget()
w.show()

app.exec_()
```

### DIFF
--- a/pyqtgraph/graphicsItems/GradientEditorItem.py
+++ b/pyqtgraph/graphicsItems/GradientEditorItem.py
@@ -44,7 +44,10 @@ class TickSliderItem(GraphicsWidget):
     """**Bases:** :class:`GraphicsWidget <pyqtgraph.GraphicsWidget>`
     
     A rectangular item with tick marks along its length that can (optionally) be moved by the user."""
-        
+    
+    sigTicksChanged = QtCore.Signal(object)
+    sigTicksChangeFinished = QtCore.Signal(object)
+    
     def __init__(self, orientation='bottom', allowAdd=True, **kargs):
         """
         ==============  =================================================================================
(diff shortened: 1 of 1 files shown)


j9ac9k:
that `test_exit_crash` ...really should try and make that test rerun on failure if running on a CI platform...

j9ac9k:
@2xB I can re-run failed jobs, don't worry about closing/re-opening to trigger new CI runs.

j9ac9k:
Thanks for the PR @2xB code looks good to me, always appreciate fewer calls to update functions when they're not needed 😆   Merging.

PR #1441, opened 2020-11-16
TickSliderItem: Use Tick.removeAllowed

2xB:
Previously, `Tick.removeAllowed` was only used in `GradientEditorItem`. This PR makes `removeAllowed` a regular property of `Tick` and enables the use of `removeAllowed` with `TickSliderItem` for a consistent interface.

Test code:
```python3
import pyqtgraph as pg
import numpy as np

app = pg.mkQApp()

        self.item = pg.TickSliderItem(*args, **kargs)

        for pos in (0, 0.5, 1):
            tick = self.item.addTick(pos)
            tick.removeAllowed = False # Possibility 1

        # Possibility 2
        for tick, value in self.item.listTicks():
            tick.removeAllowed = False

        self.setFixedHeight(31)

w = CustomWidget()
w.show()

app.exec_()
```


j9ac9k:
LGTM, thanks for the PR @2xB Merging.

PR #1442, opened 2020-11-16
TickSliderItem: allowRemove property added

### LINKED ISSUE #744: TickSliderItem allows removal when allowAdd == False
A `TickSliderItem` has a `allowAdd` optional initialization argument. When it is set to False, it is not possible to add additional ticks. However, the current behavior allows the user to remove the ticks, but does not allow the user to add them back. The correct behavior probably should be to disable removal when `allowAdd` is False.

2xB:
Following #744, this PR suggests the addition of a property
`allowRemove` to `TickSliderItem` and therefore also to
`GradientEditorItem`. It sets the default of whether ticks can
be removed by the user and therefore contemplates `allowAdd`.

I would be interested in other opinions on this design decision,
as #744 suggests to reuse `allowAdd` for the prohibition of
tick removal.
I personally suggest this solution, since it does not change
the effect of the current API.

Demo code - Should this be part of an example?:
```python3
import pyqtgraph as pg
import numpy as np

app = pg.mkQApp()

        self.item = pg.TickSliderItem(*args, allowRemove=False, **kargs)
        
        for pos in (0, 0.5, 1):
            tick = self.item.addTick(pos)
        
        self.setFixedHeight(31)


w = CustomWidget()
w.show()

app.exec_()
```

Fixes #744

### DIFF
--- a/examples/GradientWidget.py
+++ b/examples/GradientWidget.py
@@ -27,15 +27,18 @@ cw.setLayout(l)
 w1 = pg.GradientWidget(orientation='top')
 w2 = pg.GradientWidget(orientation='right', allowAdd=False)
 #w2.setTickColor(1, QtGui.QColor(255,255,255))
-w3 = pg.GradientWidget(orientation='bottom')
+w3 = pg.GradientWidget(orientation='bottom', allowAdd=False, allowRemove=False)
 w4 = pg.GradientWidget(orientation='left')
 w4.loadPreset('spectrum')
 label = QtGui.QLabel("""
 - Click a triangle to change its color
 - Drag triangles to move
+- Right-click a gradient to load triangle presets
 - Click in an empty area to add a new color
-    (adding is disabled for the right-side widget)
+    (adding is disabled for the bottom-side and right-side widgets)
 - Right click a triangle to remove
+    (only possible if more than two triangles are visible)
+    (removing is disabled for the bottom-side widget)
 """)
 
 l.addWidget(w1, 0, 1)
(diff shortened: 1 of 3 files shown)


j9ac9k:
I think `allowRemove` would be a good property to add, would also definitely be 👍 on adding it to the examples; as I frequently use the examples myself (and they are a fantastic minimum example).

2xB:
I'm sorry for doing something in a completely different direction than a minimal example. If you suggest to add the minimal example from above also or not to add so much to the customPlot example, I can do that.

The last commit introduced something that both demonstrates the use of `TickSliderItem`, `allowAdd` and `allowRemove` in context and is a thing that I needed anyways: It adds markers to the x axis of the plot. This is how they look like:

image

j9ac9k:
Actually, sorry to go back, but is a minimal example even needed given we make use of this widget in `./examples/ImageView.py` and `./examples/GradientWidget.py` ...should we perhaps update one (or both) of those examples reflecting those new APIs?

2xB:
Oh yes, the `GradientWidget.py` example looks like a great place for this, I'll add it there. How about the not-so-minimal marker example?

j9ac9k:
> Oh yes, the `GradientWidget.py` example looks like a great place for this, I'll add it there. How about the not-so-minimal marker example?

Honestly I'm not strongly opinionated on this as I've never used this widget before 😆  I'm assuming you have had a use for it, so I'm good with whatever you think is best.

2xB:
Okay, in that case, I'd keep the marker example, since it is the first example with `TickSliderItem` (others use `GradientEditorItem`) and since marking points on the y axis without painting a vertical line on the whole plot is e.g. quite useful for me.

And I'll quickly extend the `GradientWidget.py`.

2xB:
Screenshot of the updated example:

image

j9ac9k:
That looks good to me. Thanks for the PR and updates @2xB Merging!

Issue #1445, opened 2020-11-17
Regression: PlotDataItem.setData([],[]) no longer works


2xB:
### Short description
PlotDataItem.setData([],[]) no longer works since merge of #1140 (Tagging @NilsNemitz ).

### Code to reproduce

```python
import pyqtgraph as pg

app = pg.mkQApp()
w = pg.PlotWidget()
p = w.plot([1,2], [2,2])
p.setData([], [])
w.show()

app.exec_()
```

### Expected behavior
PlotDataItem should be cleared.

### Real behavior
Error is thrown:

```
Traceback (most recent call last):
  File "test_.py", line 11, in <module>
    p.setData([], [])
  File "pyqtgraph/graphicsItems/PlotDataItem.py", line 530, in setData
    self.updateItems()
  File "pyqtgraph/graphicsItems/PlotDataItem.py", line 552, in updateItems
    x,y = self.getData()
  File "pyqtgraph/graphicsItems/PlotDataItem.py", line 664, in getData
    data_range = self.dataRect()
  File "pyqtgraph/graphicsItems/PlotDataItem.py", line 689, in dataRect
    ymin = np.nanmin(self.yData)
  File "<__array_function__ internals>", line 5, in nanmin
  File "/usr/lib/python3.8/site-packages/numpy/lib/nanfunctions.py", line 319, in nanmin
    res = np.fmin.reduce(a, axis=axis, out=out, **kwargs)
ValueError: zero-size array to reduction operation fmin which has no identity
```

### Tested environment(s)

 * PyQtGraph version: Since commit 65e90faec58d90a0394e0fd0a28f7007d43ef814
 * Qt Python binding: PyQt5 5.15.1 Qt 5.15.1
 * Python version: 3.8.5
 * NumPy version: 1.19.2
 * Operating system: Linux 5.8.16-2-MANJARO x86_64 20.1.2 Mikah
 * Installation method: Executed in folder

NilsNemitz:
Dear @2xB , I guess that is not supposed to happen.
Thank you for the page. I have prepared pull request #1446 that should fix this.

There is some strangeness to the setData code which allows the internal data to be set to some unexpected values. I think it would be best to clean that up a little, but I hope one of the developers can weigh in with some suggestions.

2xB:
Hello @NilsNemitz, thank you very much for this quick response, even with a PR!

[...]

PR #1446, opened 2020-11-17
Fix set empty

NilsNemitz:
Proposed fix for issue #1445 

The direct reason of the crash when setting plot data to [],[] was that dataRect assumes empty data will appear as None.
--> Added sanity check to avoid crash.

I checked why the user would set data to [], [] and found that this seems necessary to actually delete the plot on-screen.
I think this is unintended behavior:
Setting yData to None triggers a shortcut that fails to invalidata _dataRect and xClean / yClean. In my tests that seemed to prevent the graph actually getting deleted from screen.
--> removed shortcut

Calling setData() seems to try and set xData and yData to None. But then None gets sanitized to np.array(None), which is unfortunately not actually None.
--> special-cased the code to allow xData and yData to be set to None

Data can now be cleared by calling setData(). But existing code probably calls setData( [], [] ). 
--> zero-length datasets now convert to xData = None and yData = None.

While looking at the code, I found that any data passed as named parameters 'x' and 'y' never gets sanitized
--> added forced conversion of MetaArray to ndarray as used elsewhere.

I hope one of the developers can comment  :)
On my system the modified code seems to give the intended behavior without crashing any of the examples I tested.


j9ac9k:
Hey @NilsNemitz thanks for the super-fast PR, I haven't finished drinking my coffee this morning.  

Would you  mind updating `test_PlotDataItem.py` with `test_setData()` having another example of `pdi.setData([], [])`?

NilsNemitz:
Dear @j9ac9k, it is the other end of the world here and very late. I will look into it tomorrow.

j9ac9k:
> Dear @j9ac9k, it is the other end of the world here and very late. I will look into it tomorrow.

Haha, sure thing!  thanks again for jumping on this.

NilsNemitz:
Hi @j9ac9k , I have added the tests. 

pdi.setData()
--> assert xData and yData is None

pdi.setData( [],[] )
--> assert xData and yData is None

Fascinatingly, 
np.array(None) == None  --> True
np.array(None) is None --> False
Neat textbook case... I hope these won't happen anymore.

I found that there is a PlotDataItem.clear() command that was probably the intended way to clear the data.
The test for this asserted xData == None.
--> changed the assertion for pdi.clear() to "xData is None".

While checking that setData() and clear() both invalidate cached data properly, I found that xClean and yClean are never used in the entire pyqtgraph code folder, except when they get set to None.
--> Commented out all remaining references. It seems they have been added ~7 years ago, so I guess they can be removed?

Tests pass on my system with PyQt5, except for the GL stuff that I probably haven't set up correctly.

2xB:
@NilsNemitz Thank you! Regarding `xClean` and `yClean`: Yes, I'd not comment things out but remove them entirely. Since these properties are not documented and not used, I'd assume it is safe to remove them, and if this PR is found to prove an issue later, one can always look at the old code in the old Git commits.

NilsNemitz:
@2xB , thank you for the clear instructions, all remains of xClean and yClean are gone now.
Any advice on the failed test? I can only find this:
`pyqtgraph/tests/test_exit_crash.py::test_pg_exit `
`[gw0] [ 85%] FAILED pyqtgraph/tests/test_exit_crash.py::test_pg_exit`
...and I fail to see how that may be related to anything I touched.

j9ac9k:
That test is flaky, should pass on rerun which I'll trigger on a bit.

j9ac9k:
This LGTM, thanks for the PR @NilsNemitz

Issue #732, opened 2018-07-23
XLinked PlotItem issue


pojzn:
Greetings,

I have an issue with multiple PlotItems shown on a GraphicsLayoutWidget. I need the plots' X axis to be linked. 

If the plots are arranged vertically (one plot per row), and if they are linked when zooming in on the Y axis, the Y axis AxisItem's text cuts off the X axis instead of resizing/rescaling the AxisItem.

However, if the plots are arranged in a 2x2 grid for example, and if they are linked, everything seems to work fine.

Is there anything that I am missing here? Am I linking the X axis wrong? 
**I have noticed that the "bug" does not manifest on plots in vertical layouts where the setXLink method is not called - shown on the 3rd plot in the vertical layout widget in my example.**

[>>Example<<](https://github.com/pojzn/pyqtgraph_axis_issue/blob/master/plotting_issue_example.py)

I appreciate any feedback on this issue, since I am actively using pyqtgraph in my project.

![axisitemissue](https://user-images.githubusercontent.com/25216268/43070372-f6435e52-8e6f-11e8-810e-ea719bf6d1a7.png)

pojzn:
Bump, any ideas?

mliberty1:
I think that I have stumbled across this same issue or a very similar issue.  The textSpecs tick labels on my x-axis are sometimes not updated correctly, and fragments of the old text remain.  Here is an example:
![xaxis_label_redraw_incorrect](https://user-images.githubusercontent.com/5393422/99404012-22a17700-28b9-11eb-94c4-9fed34364691.gif)
Right-clicking the view box or resizing does not appear to respect the bounding box.  AxisItem.drawPicture() is not called on right-click.  Resizing is harder to isolate since other calls happen.

I added a print(self.picture.boundingRect()) before the call to [self.picture.play(p)](https://github.com/pyqtgraph/pyqtgraph/blob/020b7077cb635f0341ab432f7c53fa4d8b1eb697/pyqtgraph/graphicsItems/AxisItem.py#L603) in AxisItem.paint(), and I see two calls:
PySide2.QtCore.QRect(0, 20, 512, 304)
PySide2.QtCore.QRect(52, 0, 519, 298)

That seems odd.  Any thoughts or suggestions?

j9ac9k:
[...]

PR #1447, opened 2020-11-17
Clipped AxisItem tick value labels to prevent drawing artifacts #732

mliberty1:
See https://github.com/pyqtgraph/pyqtgraph/issues/732#issuecomment-729223974 for details.


j9ac9k:
Hi @mliberty1 thanks for the PR, As I haven't heard any reason not to merge this, and this fixes an existing issue I'll merge momentarily.  Thanks again for the PR!

PR #1448, opened 2020-11-19
Fixes for examples\customPlot.py

2xB:
Fixing three errors:

1. Copy-paste error: I referenced a specific object instead of self, circumventing object orientation.
2. Executing custom function `CustomTickSliderItem.setTicks` more than once resulted in a KeyError.
3. Did not consider padding of `TickSliderItem` for link between `ViewBox` axis limits and tick positions, resulted in ticks positioned slightly off when moved to left or right edge of plot

Errors 1. and 2. were caused because I did only test the custom class `CustomTickSliderItem` in the example, where both was not problematic.

This happened probably because of being happy too early about the implementation of markers being so straight-forward.

### DIFF
--- a/examples/customPlot.py
+++ b/examples/customPlot.py
@@ -39,9 +39,9 @@ class CustomTickSliderItem(pg.TickSliderItem):
         self._range = [0,1]
     
     def setTicks(self, ticks):
-        for tick in self.listTicks():
+        for tick, pos in self.listTicks():
             self.removeTick(tick)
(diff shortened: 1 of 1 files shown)


j9ac9k:
Really getting fed up w/ `test_exit_crash`...  should consider making some changes in our CI pipeline to make it automatically rerun on failure or something..

Anyway back to this PR, this LGTM, thanks for patching this up @2xB , merging.

PR #1452, opened 2020-11-20
InfiniteLine caching calls merged (fixes #1450)

### LINKED ISSUE #1450: Regression: InfiniteLines became finite and movable
### Short description

`InfiniteLine`s are movable and finite along their "infinite" direction since PR #1391 (Tagging @jjuod )
. Moving them also moves their labels. PR #1413 only fixes their initial positions.

### Code to reproduce
`examples\InfiniteLine.py`

### Expected behavior

Lines should be infinite and dragging them should not move the label into their "infinite" direction.

image

### Real behavior

Lines are finite and moving them also moves their labels.

image

### Tested environment(s)

 * PyQtGraph version: Since commit 39f9c6a
 * Qt Python binding: PyQt5 5.15.1 Qt 5.15.1
 * Python version: 3.8.5
 * NumPy version: 1.19.2
 * Operating system: Linux 5.8.16-2-MANJARO x86_64 20.1.2 Mikah
 * Installation method: Executed in folder

jjuod:
This should fix the `InfiniteLine` problems. Since `viewTransform` caching was added at the `graphicsItem` level, `InfiniteLine` ended up having two different caches and cache cleanup functions, and one of them wasn't called when needed. (Turns out that these lines actually get dragged in the "infinite" direction as well, they just update the view box to draw a different part of the line...) 

This merges both cleanup functions. Seems like `examples/InfiniteLine.py` works as expected now.

### DIFF
--- a/pyqtgraph/graphicsItems/InfiniteLine.py
+++ b/pyqtgraph/graphicsItems/InfiniteLine.py
@@ -249,7 +249,7 @@ class InfiniteLine(GraphicsObject):
 
         if self.p != newPos:
             self.p = newPos
-            self._invalidateCache()
+            self.viewTransformChanged()
             GraphicsObject.setPos(self, Point(self.p))
             self.sigPositionChanged.emit(self)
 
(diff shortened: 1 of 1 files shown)


j9ac9k:
Tested this out, and this works beautifully.  Thanks for following up @jjuod !

PR #1459, opened 2020-11-27
Fix TickSliderItem: Avoid ghost ticks | Improved customPlot.py code

2xB:
If `TickSliderItem.setTickValue` was called when a full repaint of the `TickSliderItem` was not already scheduled, the tick was visible at the old and the new position. This could e.g. be seen when using the autoscale button in the `customPlot.py`, as seen here, zooming out and using autoscale before the fix using PyQ5 - only three of the markers are real:
image

Further, code from `customPlot.py` is improved to make use of `Tick.setVisible` instead of adding and removing ticks based on their visibility.

### DIFF
--- a/examples/customPlot.py
+++ b/examples/customPlot.py
@@ -34,16 +34,16 @@ class CustomTickSliderItem(pg.TickSliderItem):
     def __init__(self, *args, **kwds):
         pg.TickSliderItem.__init__(self, *args, **kwds)
         
-        self.all_ticks = []
-        self.visible_ticks = {}
+        self.all_ticks = {}
         self._range = [0,1]
     
     def setTicks(self, ticks):
         for tick, pos in self.listTicks():
             self.removeTick(tick)
-        self.visible_ticks = {}
         
-        self.all_ticks = ticks
+        for pos in ticks:
+            tickItem = self.addTick(pos, movable=False, color="333333")
+            self.all_ticks[pos] = tickItem
         
         self.updateRange(None, self._range)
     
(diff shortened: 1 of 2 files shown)


examples/customPlot.py
```diff
@@ -56,23 +56,16 @@ def updateRange(self, vb, viewRange):
...
-            if pos not in self.visible_ticks:
-                self.visible_ticks[pos] = self.addTick(pos, movable=False, color="333333")
-            
-            tick = self.visible_ticks[pos]
-            
             tickValueIncludingPadding = (pos - viewRange[0]) / (viewRange[1] - viewRange[0])
             tickValue = (tickValueIncludingPadding*lengthIncludingPadding - origin) / length
             
-            visible = tickValue >= 0 and tickValue <= 1
+            visible = bool(tickValue >= 0 and tickValue <= 1)
```
j9ac9k, line 62: This should already be a `bool` without having to cast it, no?

2xB: It's a `np.bool_` because of the arithmethics with `viewRange` above. And `PyQt5` throws a warning when one uses `np.bool_` instead of `bool`... Wasn't intuitive for me either. Maybe adding a comment would help?

2xB:
More considerations: I think that the reason for this bug is that in the example, when clicking on the autorange button, `setTickValue` is called through `ViewBox.enableAutoRange` via `ViewBox.paint`, and my guess is that when already painting, `update` is not automatically called when the `Tick`s are moved via `QGraphicsWidget.setPos`.

One could also call `self.update` along calling `setTickValue`, so not putting that call in `TickSliderItem` code but in `customPlot.py` in this case. But `self.update()` only schedules a full repaint, and that has to be done anyways when using `setTickValue`, so I don't think this can pose any harm, but it should avoid the possibility that anyone else has issues with ghost ticks.

Btw. I am sorry for making this a [WIP], I became afraid of the performance implications of calling `self.update()` and wanted to read the [API documentation for `update`](https://doc.qt.io/qt-5/qgraphicsitem.html#update) carefully before making this decision. Now that I did read again that it only triggers a repaint when back to the control loop, and it therefore has basically no performance impact on `setTickValue` when calling it multiple times, I removed the [WIP] again.

examples/customPlot.py
j9ac9k: ahh; yeah that got me....yeah definitely add a comment for that!

2xB: Added a comment, is that understandable?

j9ac9k: LGTM

j9ac9k:
My general rule of thumb is that library users shouldn't have to worry about calling methods like `QGraphicsItem.update()` and such; this library should handle that in all but some extreme cases; so I'm glad that call isn't in the example.

This PR LGTM, Merging.  Thanks @2xB

PR #1461, opened 2020-11-27
arrayToPath: revert to old meaning of connect=ndarray parameter

### LINKED ISSUE #1460: PlotCurveItem connect=ndarray behaves differently on master
### Short description
#1283 introduced a regression for users passing ndarray into PlotCurveItem. See @pijyoi's last comment on #1283.

The i-th position of that array used to define if points (i) and (i+1) were connected, now it defines whether points (i-1) and (i) are connected. It broke https://github.com/biolab/orange3/blob/master/Orange/widgets/visualize/owlineplot.py

Both new and old interpretations of the connect parameter are equally expressive. Therefore, I suggest changing the meaning back. That would maintain backward compatibility and, as code with the new meaning was not released yet, there are essentially no downsides to reverting to old behavior.

A possible solution in arrayToPath:
```
     elif isinstance(connect, np.ndarray):
-        arr[1:-1]['c'] = connect
+        arr[2:-1]['c'] = connect[:-1]
     else:

```
With this suggested fix the code should perform equally fast.

### Code to reproduce
See @pijyoi's last comment on #1283.

### Tested environment(s)

 * PyQtGraph version: master, 0.11.0
 * Qt Python binding: 5.12.2, 5.12.3
 * Python version: 3.8

### Additional context

For some context, here is what test_PlotCurveItem draws. But I do not understand why tests pass, they should not. 

Here are the test-data. Order (all, array, finite, pairs)
connectallconnectarray
connectfiniteconnectpairs

master on Qt 5.12.3
m_connectallm_connectarray
m_connectfinitem_connectpairs

0.11.0 on Qt 5.12.2 (to avoid Qt drawing bugs later fixed in #1456 (fine):
m_connectallm_connectarray
m_connectfinitem_connectpairs

master on Qt 5.12.2 (to ensure that #1456 did not break anything):
m_connectallm_connectarray
m_connectfinitem_connectpairs

markotoplak:
The i-th position of that array used to define if points (i) and (i+1) were connected, but in master it defines whether points (i-1) and (i) are connected. This PR reverts to (i) and (i+1) interpretation.

Fixes #1460 (more therein).

A test drawing from test_PlotCurveItem

0.11.0 (as in 0.10.0, as in tests)
m_connectarray

master (different)
m_connectarray

this PR (back to what 0.11.0 draws)
m_connectarray

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1529,7 +1529,7 @@ def arrayToQPath(x, y, connect='all'):
             isfinite = np.isfinite(x) & np.isfinite(y)
         arr[2:]['c'] = isfinite
     elif isinstance(connect, np.ndarray):
-        arr[1:-1]['c'] = connect
+        arr[2:-1]['c'] = connect[:-1]
     else:
         raise Exception('connect argument must be "all", "pairs", "finite", or array')
 


j9ac9k:
CI failures are unrelated; thanks for trying out the `master` branch right away and catching this issue before we released 0.11.1.

LGTM, merging

PR #1462, opened 2020-12-07
Added option to use "None" for adj in setData

Ryytikki:
This allows you to easily create code to toggle lines on or off by setting to the appropriate array if on, or None if off. Currently you have to have 2 seperate setData calls, one with adj and one without

### DIFF
--- a/pyqtgraph/graphicsItems/GraphItem.py
+++ b/pyqtgraph/graphicsItems/GraphItem.py
@@ -55,7 +55,7 @@ class GraphItem(GraphicsObject):
         """
         if 'adj' in kwds:
             self.adjacency = kwds.pop('adj')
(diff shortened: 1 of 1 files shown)


j9ac9k:
hi @Ryytikki thanks for the PR.

I think you want to change the order of your if statment from

```python
            if self.adjacency.dtype.kind not in 'iu' and self.adjacency is not None:
```

to 

```python
            if  self.adjacency is not None and self.adjacency.dtype.kind not in 'iu':
```

The reason for this is that if `self.adjacency` is indeed `None` you're going to get an error when checking `self.adjacency.dtype.kind`.

j9ac9k:
Thanks for the PR @Ryytikki I've made some changes to it a bit, so I can get it in a state to be merged.

PR #1464, opened 2020-12-10
Use conda-forge on pyside2+linux

j9ac9k:
Attempt to patch up the CI to use the correct version of pyside2


j9ac9k:
there is one pipeline that is still hanging w/ an unrelated issue, so I think this is good; fix was to enforce strict channel ordering when installing `pyside2` via conda-forge.

PR #669, opened 2018-04-25
[wip] Imageitem performance

campagnola:
Working on improving update performance of ImageItem:
* ensure that ubyte data can be passed through to QPixmap without any extra overhead
* optimize levels / lookup tables / clipping / channel ordering where possible (maybe we can use numba to combine multiple steps into a single loop)

Fixes #667

PR #1466, opened 2020-12-12
Performance enhancement: use CUDA in ImageItem

outofculture:
[cupy](https://cupy.dev/) is a mostly-compatible drop-in replacement for numpy that performs its work on a CUDA-enabled GPU. This PR lets `ImageItem.setImage` accept an image on either substrate, and otherwise behaves identically.

Also, windows memory allocation requests are slow, so pre-allocating buffers for processing improves performance ( for both cpu and gpu ). Since writing this, I've identified other places that could benefit from this treatment, but every bit helps.

The `examples/VideoSpeedTest.py` script was improved to accept command line arguments and to test the cuda-enabled processing.

Only one cupy function was found to be anything other than identical ( `cupy.take` does not support `mode="clip"` and needed to be clipped explicitly ), but behavior will of course need to be verified everywhere. Thankfully, this PR shouldn't significantly alter the way numpy-based images are processed, so we can safely leave cuda as an optional/experimental feature while we feel out how it behaves on more diverse systems than I was able to use for my tests.

In testing, we found one issue. On windows systems with CUDA Toolkit < 11.1, an int16-dtype image with a lookup table will be incorrectly processed ( the result gets entirely mapped to 255; it whites out ). I didn't know where to document or enforce this requirement.

This also begs the question of _what else_ would benefit from becoming cupy/numpy-agnostic, but we should leave that until we've proven this small slice of functionality.


j9ac9k:
Qt4 <shakes-fist-at-cloud>

outofculture:
Oh, testing has happened on 3 systems so far.

Linux, Pop!OS, cuda:10.2, GeForce GTX 1660 Ti Mobile

Windows 10 Pro, cuda:11.1, Quadro P1000

j9ac9k:
@outofculture thanks for the PR!  I am 100% onboard w/ making cupy an optional dependency.  

There is another PR that @campagnola has been working on to improve image item performance ( #669 ) ...which it appears you are the author of some commits on that PR.  Can you expand on how the two PRs compare?

outofculture:
Yeah, @j9ac9k I originally forked off of Luke's work there, but then I sorta ignored what was there already and went off in a different direction. I was able to rebase and isolate this changeset to just the cupy and pre-alloc stuff. I would totally be willing to group them together into a single PR to ease the testing burden; I would just need to confer with @campagnola to move #669 out of wip status.

j9ac9k:
I'll try and ping him; but I don't want to hold off for too long in case we don't hear back.   Thanks again for this PR!

dgoeries:
If this make it throught, can you please keep the `qimage` variable. 

The performance drain is truly the color map application, which can be tackled as well with `QImage.setColorTable`, if pyqtgraph would move away from the very generic behavior (similar to PythonQwt).

j9ac9k:
Haven't heard from @campagnola and we've killed off Qt4/Py2 support so I'm going to take a closer look at this soon, I welcome additional feedback from other users, as I personally do not use ImageItem in my work often, so I may not recognize potential issues such as what @dgoeries brought up.

@outofculture potentially stupid question here, but as we can now depend on numpy 1.17+, do we need the the `empty_cupy` functionality, or can we rely on `__array_function__` [link](https://numpy.org/doc/stable/reference/arrays.classes.html#numpy.class.__array_function__)?  I may have a complete misunderstanding on how `__array_function__` is supposed to work.

Also, is there a preferred order with merging this vs. merging #669 ?

pijyoi:
I use the qimage variable mentioned by @dgoeries to bypass conversion of an 8-bit intensity image to RGBA float.
Granted, it's not part of the public API. 

```python
import pyqtgraph as pg
import numpy as np
import matplotlib.cm
QtGui = pg.QtGui

def make_qimage(data, grayscale=False):
    if grayscale:
        qimg = QtGui.QImage(data, data.shape[1], data.shape[0], QtGui.QImage.Format_Grayscale8)
    else:
        qimg = QtGui.QImage(data, data.shape[1], data.shape[0], QtGui.QImage.Format_Indexed8)
        cmap = matplotlib.cm.get_cmap('Spectral')
        lut = cmap(np.linspace(0, 1, 256), bytes=True)
        qtlut = [QtGui.qRgba(*x) for x in lut]
        qimg.setColorTable(qtlut)
    return qimg

def main():
    app = pg.mkQApp()

    wgt = pg.GraphicsLayoutWidget()
    wgt.resize(800, 600)
    wgt.show()
    vb = wgt.addViewBox()
    image = pg.ImageItem()
    vb.addItem(image)

    W = 4000
    data = np.random.randint(255, size=(W, W), dtype=np.uint8)
    image.setImage(data, autoLevels=False)
    image.qimage = make_qimage(data)
    
    app.exec_()

if __name__ == '__main__':
    main()
```

outofculture:
> as we can now depend on numpy 1.17+, do we need the the `empty_cupy` functionality, or can we rely on `__array_function__` [link](https://numpy.org/doc/stable/reference/arrays.classes.html#numpy.class.__array_function__)?

Yes! Thank you for noticing this. Cupy implements `__array_function__`, and relying on that would leave the code looking much cleaner. So much so that I worry that the need for interoperability would be completely hidden thereafter. I will happily switch over to using that now.

My worry does highlight a possible need for CI integration. I started researching it, [and it only looks possible](https://stackoverflow.com/questions/56861413/running-cuda-unit-tests-on-azure-pipelines) for self-hosted runners. I would need to work with someone to set that up, if that's even an option.

Notably, I hadn't written any unit tests for this; I just used the VideoSpeedTest as my testing playground. I can write a couple of those regardless, being sure to skip when cuda isn't available.

> Also, is there a preferred order with merging this vs. merging #669 ?

Or merging the two together first before review? I'm not currently able to speak to the intent of the #669 changes, but @campagnola has communicated with me and is interested in getting both of them grouped together. I'll see if that's something we can get done in the next couple of days, but otherwise this PR can be safely considered on its own.

> `image.qimage = make_qimage(data)`

`qimage` is still on the object and gets used, but yeah, this PR might break variations of that code. Hrm... The problem being that `setImage` now needs to access the qimage's data directly ( through `self._displayBuffer` ) to prevent allocs. A possible interface to support this would be to use a qimage factory somehow, e.g.:

```python
image = pg.ImageItem(qimageFactory=make_qimage)
```

Would that suffice?

pijyoi:
Actually, I am fine with simply modifying my code to:
```python
image.setImage(data, autoLevels=False)
image.qimage = make_qimage(data)
image._renderRequired = False
image._unrenderable = False
```

j9ac9k:
> Yes! Thank you for noticing this. Cupy implements `__array_function__`, and relying on that would leave the code looking much cleaner. So much so that I worry that the need for interoperability would be completely hidden thereafter. I will happily switch over to using that now.

If this can clean-up code in this case, that would be fantastic!  Especially if there are other places within the library that CuPy can help speed things up.

> My worry does highlight a possible need for CI integration. I started researching it, and it only looks possible for self-hosted runners. I would need to work with someone to set that up, if that's even an option.

Considering we don't have any self-hosted runners, this does present a challenge.  I do have linux server in-house with the nvidia docker runtime, so I _could_ setup a local runner that runs on a cron job or something; but this is not ideal. While I should explore this issue more, I will  _happily_ throw an "experimental" label on this functionality and leave it out of the CI for now until we can figure out something else...

dgoeries:
`_renderRequired` is only a translation for `qimage is None` or not?

outofculture:
> `_renderRequired` is only a translation for `qimage is None` or not?

That would have been the previous code, yeah, though `qimage is None` was overloaded to mean both "render required" and "unrenderable", depending on when the test was performed.

outofculture:
@j9ac9k I'm partway through reverting some of the `xp`/`np`/`cp` code, but I'm having a really uncomfortable feeling about the resulting implicit split. All allocation *has* to be done on the appropriate substrate, but function calls can always happen on numpy. That's not at all obvious once the `xp` incursion is minimized. So much so that I'm tempted to push even further in the opposite direction and explicitly put `np = None  # use xp for substrate compatibility` at the top of each of these bi-substrate functions. Some of the functions never allocate anything, and they're much cleaner to be rid of `xp`, but the rest need a reminder.

I _have_ determined that the following would probably look cleaner and better match other probative imports in pyqtgraph:
```python
try:
    import cupy as cp
    has_cupy = True
except ImportError:
    cp = None
    has_cupy = False
```

Then use `has_cupy` to gate attempts to use `cp`. ( Ah, my ruby background shows itself again. I so often fall into making duck-typed stand-ins for things when an `if` would perform just as well. )

j9ac9k:
@outofculture you've worked with this kind of code far more than I have, so I'll trust your judgment here regarding `xp`/`np`/`cp` code; and given you're the one making this PR, I have a feeling you'll be the biggest user at least from the get-go.

Regarding `has_cupy`; I've been actually scratching my head about that very issue, but with `numba` (I'm considering having `numba` be an optional dependency to speed up some of the plots)  I'm wondering if the best place to store those parameters is with the global configuration options: https://pyqtgraph.readthedocs.io/en/latest/config_options.html

this show allow the user to specify to _not_ use `cupy` (or `numba, or whatever) even if it imports successfully; in various places in the code-base where we want to use an if-check, we can compare against the config option there, instead of having `try/except ImportError`'s throughout.

outofculture:
I put in an option gate on the import, but * _grumble grumble import star_ * it's back to having a cupy util. I did it without the duck-typing this time, though, so things remain pretty simple.

@campagnola has been swamped, it seems, so we may as well move forward on this as is.

outofculture:
Woo! Checks pass! I have no idea why macos as failing there for a few runs, and I did nothing to try to fix anything, but the build fails went away on their own. 🤷🏽‍♀️

j9ac9k:
hi @outofculture we've had some intermittent macOS failures on openGL related examples/tests, that I'm pretty sure are CI system related; so yeah, don't sweat those...

I'll try and dive into this tonight... I'm likely going to have some pretty simple/dumb questions, is this your preferred medium to communicate with, or would you like to migrate the discussion elsewhere?

outofculture:
@j9ac9k github is fine. Depending on exactly when you look at it, I could be available on a higher-bandwidth chat platform ( live text, audio or video ) of your choosing. I'll dm my contact info.

pyqtgraph/functions.py
```diff
@@ -947,73 +951,48 @@ def rescaleData(data, scale, offset, dtype=None, clip=None):
...
-            import scipy.weave
-        except ImportError:
-            raise Exception('scipy.weave is not importable; falling back to slower version.')
-        
-        ## require native dtype when using weave
-        if not data.dtype.isnative:
-            data = data.astype(data.dtype.newbyteorder('='))
-        if not dtype.isnative:
-            weaveDtype = dtype.newbyteorder('=')
+    # p = np.poly1d([scale, -offset*scale])
```
j9ac9k, line 954: can we remove commented out lines?

pyqtgraph/functions.py
```diff
@@ -947,73 +951,48 @@ def rescaleData(data, scale, offset, dtype=None, clip=None):
...
     Uses values in *data* as indexes to select values from *lut*.
     The returned data has shape data.shape + lut.shape[1:]
     
     Note: color gradient lookup tables can be generated using GradientWidget.
+
+    Parameters
+    ----------
+    lut : ndarray
+        Either cupy or numpy arrays are accepted, though this function has
+        problems in cupy on windows with some versions of the cuda toolkit.
```
j9ac9k, line 985: is this with cuda versions < 11.0? ...probably should try and be a bit more specific here (doesn't have to be exhaustive, but documenting what's been observed would be good).

pyqtgraph/functions.py
outofculture: I didn't know if this project had a strong preference to keep old bits of code like this around, but I would be happy to drop unused comments.

j9ac9k: I don't mind leaving some old bits like this in, but probably should add a note as to why you're wanting to keep them around

outofculture:
All comments should be addressed.

j9ac9k:
There is something the CI doesn't like on this PR...given it's failing on multiple platforms I need to investigate further.  @outofculture can you replicate the issue on ubuntu/pyside2/python3.9?

outofculture:
Yeah, the tests fail for me in 3.9, but I'm getting failures in master as well.

On master, `test_Viewbox` is failing intermittently, and `test_ErrorBarItem_defer_data` is failing consistently.

On my branch, the segfault and timeout reproduced the first time I ran the tests, but since then, I can't get it to happen again. I do also get the same failures as master.

j9ac9k:
Those errors I'm not too concerned about (`test_ErrorBarItem_defer_data` has been intermittent if I remember right, haven't had any failures with `test_Viewbox before); the issues I'm concerned with are the timeouts/segfaults; historically, some segfaults have come up with small/unrelated changes in the code.  Annoyingly, I have never been able to replicate them locally, they just happen on CI platforms periodically.  

I'll run locally on some various platforms (due to my hidpi laptop display the test suite fails horribly on windows), and see if I'm able to replicate...

outofculture:
Arg; I can't reproduce the segfault. I've tried repeatedly re-running the tests, switching branches, rebuilding the env, and rebooting. It happened the first time for me, but never since. The stack in the logs doesn't look familiar to me at all; it's not in any of the code I worked on.

j9ac9k:
For the record, I'm not suggesting the issue is one of your code, but with segfaults, from what I understand, changes in one place can reveal issues elsewhere.

j9ac9k:
Thanks so much for this feature @outofculture !

PR #1468, opened 2020-12-16
Zero step in np.arange

miranis:
The following code will fail if step is 0 (np.arange raises ValueError: arange: cannot compute length):
bins = np.arange(mn, mx+1.01*step, step, dtype=np.int)

The value of step is set to 0 for all images of uniform intensity (all pixels in the image have equal intensities, e.g. a fully saturated or underexposed image).

This solution does not call np.arrange if step is 0, but instead assigns an empty list that will be later replaced by [mn, mx].


j9ac9k:
LGTM, thanks for the PR @miranis

Issue #347, opened 2016-05-28
Parameter class Issue - Can't remove and add a child parameter back 


aziesemer:
Hello,
I'm having the following problem with PyQtGraph0.9.10+PySide1.2.2+Python2.7. I'm getting a segmentation fault after remove a Parameter, add it back, and try change its value. This error happens in the valueChange method of the WidgetParameterItem class:

`self.widget.sigChanged.disconnect(self.widgetValueChanged)
`
If we call self.widget.show() in the line before, this exception is raised:

`RuntimeError: Internal C++ object (SpinBox) already deleted.
`
This is a code example that reproduces the problem I'm having:

```
from pyqtgraph.parametertree import Parameter, ParameterTree
from pyqtgraph.Qt import QtGui

app = QtGui.QApplication([])
t2 = ParameterTree()
t2.show()
root = Parameter.create(name='Root', type='group')
t2.addParameters(root)
par = Parameter.create(name='Par1', type='group')
par.addChildren([{'name':'test', 'type':'int', 'limits':(500, 2000000), 'value':38400}])
root.addChild(par)
par.remove()
root.addChild(par)

QtGui.QApplication.instance().exec_()
```

The problem is that the signals in the Parameter class remain connected to the destroyed ParameterItem. My workaround was to capture the destroyed signal of the widget objects and disconnect the ParameterItem signals:

```
# Disconnect the signals when the widget is gone
def __init___fix(self, param, depth):
 ParameterItem.__init__(self, param, depth)

 self.hideWidget = True ## hide edit widget, replace with label when not selected
 ## set this to False to keep the editor widget always visible

 ## build widget into column 1 with a display label and default button.
 w = self.makeWidget()
 self.widget = w

 self.widget.destroyed.connect(lambda: self.param.sigValueChanged.disconnect(self.valueChanged))
 self.widget.destroyed.connect(lambda: self.param.sigChildAdded.disconnect(self.childAdded))
 self.widget.destroyed.connect(lambda: self.param.sigChildRemoved.disconnect(self.childRemoved))
 self.widget.destroyed.connect(lambda: self.param.sigNameChanged.disconnect(self.nameChanged))
 self.widget.destroyed.connect(lambda: self.param.sigLimitsChanged.disconnect(self.limitsChanged))
 self.widget.destroyed.connect(lambda: self.param.sigDefaultChanged.disconnect(self.defaultChanged))
[...]

PR #1469, opened 2020-12-18
Various improvements to parametertree

### LINKED ISSUE #339: bounds / limits not always respected in tree parameters
If I set the value of a parameter in a parameter tree using the setValue(x) method, its value in the GUI is correctly updated to respect the "bounds" and/or "limits" option, but subsequent calls to getValue() will simply return x, even if it is outside the bounds (!). You can check this behavior quickly as follows:

``` python
>>> import pyqtgraph as _g
>>> p = _g.parametertree.Parameter.create(name='pants', value=32, type='int', limits=(0,10), bounds=(0,10))
>>> p.value()
32
>>> p.setValue(-200)
-200
>>> p.value()
-200
```

Does anyone else find this behavior? If so, this one is a serious bug, because the user looking at the GUI has no of knowing it's not working, which can damage lab equipment (but didn't in our case! :).

### LINKED ISSUE #760: ParameterTree list change value in GUI
I would like back at the before value if a condition.

I have a list with values in a parameterTree : 

my parameterTree is connected on TreeStateChanged signal

In my "change" method, I know how change value in my list but it's not show in my interface. 

How can I take this ?

```import sys
import pyqtgraph as pg


class UnExemple(QtWidgets.QWidget):
    def __init__(self):
        super().__init__()
        self.params = [{'name': 'Basic',
                        'type': 'group',
                        'children': [{'name': 'List',
                                      'type': 'list',
                                      'values': [1, 2, 3],
                                      'value': 2}]}]

        # Create tree of Parameter objects
        self.p = Parameter.create(name='params',
                                  type='group',
                                  children=self.params)
        self.p.sigTreeStateChanged.connect(self.change)

        # Create two ParameterTree widgets, both accessing the same data
        self.t = ParameterTree(self)
        self.t.setParameters(self.p, showTop=False)
        self.t.setWindowTitle('pyqtgraph example: Parameter Tree')
        self.resize(200, 100)
        self.show()

    # If anything changes in the tree, print a message
    def change(self, _, changes):
        print("tree changes:")
        for param, change, data in changes:
            path = self.p.childPath(param)
            if path is not None:
                childName = '.'.join(path)
            else:
                childName = param.name()
            print('  parameter: %s' % childName)
            print('  change:    %s' % change)
            print('  data:      %s' % str(data))
            print('  ----------')
            if data == 3:
                print('change')
                self.p.param('Basic', 'List').setValue(1)


#  Start Qt event loop unless running in interactive mode or using pyside.
if __name__ == '__main__':
    APP = pg.mkQApp()
    WIN = UnExemple()
    sys.exit(APP.exec_())
```

### LINKED ISSUE #1006: ParameterTree always reports default sizeHint instead of content dependent sizeHinte
### Short description
I tried to resize a window to fit to a ParameterTree. However, the sizeHint() of the ParameterTree always reports the default of 256 x 192, regardless of its conentet.

### Code to reproduce
```python
from pyqtgraph import QtGui, QtCore
import sys

# new QtApp
app = QtGui.QApplication([])
    
t = ParameterTree()
print('sizeHint of empty ParameterTree', t.sizeHint()) #  PyQt5.QtCore.QSize(256, 192)

# add lengthy parameter
t.addParameters(Parameter(name='This is a very, very long name to make the Widget expand.',
                         value=0.0,
                         default=0.0))

print('sizeHint of ParameterTree', t.sizeHint()) #  PyQt5.QtCore.QSize(256, 192), but should be larger

# spawn a window
win = QtGui.QMainWindow()
win.setCentralWidget(t)
win.show() # still 256 x 192

```

### Expected behavior
The window spawned by the example should math the size of the ParameterTree, i.e. should receive a sizeHint() corresponding to the content of the ParameterTree.

### Real behavior
The windows spawned by the example is sized to a default/empty ParameterTree.

#### Windows test system
* PyQtGraph version: 0.11.0.dev0+g51b3201
* NumPy version: 1.16.3 
* Python version:  3.7
* Operating system: Windows 7
* Installation method: develop branch from git via pip
#### Linux test system
* PyQtGraph version: 0.10.0
* Qt Python binding: PyQt5 5.13.0 Qt 5.13.0
* NumPy version: 1.17.0
* Python version: 3.6.5
* Operating system: openSUSE Leap 15.1
* Installation method: pip

### LINKED ISSUE #1385: Inconsistent behavior between 'changed' and 'changing' signals for color picker
### Short description

I expect the returned value for `sigChanging` and `sigChanged` signals to be the same. E.g. if a text box is used, both signals should emit strings. In the case of a color picker, `sigChanged` emits a color (as expected) while `sigChanging` emits the color square that was picked. Either both should emit a color or both should emit the selected square, I think.

### Code to reproduce

```python
from pyqtgraph import parametertree
from pyqtgraph import mkQApp

app = mkQApp()
p = parametertree.Parameter.create(name='test', type='color', value='fff')

t = parametertree.ParameterTree()
t.setParameters(p, showTop=True)

p.sigValueChanging.connect(lambda _p, val: print(type(val)))
p.sigValueChanged.connect(lambda _p, val: print(type(val)))

t.show()
app.exec_()
```

### Expected behavior
Both types should match

### Real behavior
```
sigChanging vs. sigChanged:
<class 'pyqtgraph.widgets.ColorButton.ColorButton'>
<class 'PySide2.QtGui.QColor'>
```
### Tested environment(s)

 * PyQtGraph version: 0.11.0
 * Qt Python binding: 5.15
 * Python version: 3.8
 * Operating system: win10
 * Installation method: pip

### LINKED ISSUE #1393: Parameters do not respect their 'enabled' option
Parameters cannot be disabled by passing `enabled=False` during initialization or by calling `param.setOpts(enabled=False)` after initialization. In the docstring of the `Parameter.__init__` function we say `enabled` is a valid option to pass, but actually passing it has no effect.

### Code to reproduce

```import pyqtgraph as pg

app = pg.mkQApp()
import pyqtgraph.parametertree.parameterTypes as pTypes
from pyqtgraph.parametertree import Parameter, ParameterTree, ParameterItem, registerParameterType

params = [
    {'name': 'Basic parameter data types', 'type': 'group', 'children': [
        {'name': 'Integer', 'type': 'int', 'value': 10, 'enabled':False},
        {'name': 'Float', 'type': 'float', 'value': 10.5, 'step': 0.1, 'enabled':False},
        {'name': 'String', 'type': 'str', 'value': "hi", 'enabled':False},
        {'name': 'List', 'type': 'list', 'values': [1,2,3], 'value': 2, 'enabled':False},
        {'name': 'Named List', 'type': 'list', 'values': {"one": 1, "two": "twosies", "three": [3,3,3]}, 'value': 2, 'enabled':False},
        {'name': 'Boolean', 'type': 'bool', 'value': True, 'tip': "This is a checkbox", 'enabled':False},
        {'name': 'Color', 'type': 'color', 'value': "FF0", 'tip': "This is a color button", 'enabled':False},
        {'name': 'Gradient', 'type': 'colormap', 'enabled':False},
        {'name': 'Subgroup', 'type': 'group', 'enabled':False, 'children': [
            {'name': 'Sub-param 1', 'type': 'int', 'value': 10, 'enabled':False},
            {'name': 'Sub-param 2', 'type': 'float', 'value': 1.2e6, 'enabled':False},
        ]},
        {'name': 'Text Parameter', 'type': 'text', 'value': 'Some text...', 'enabled':False},
        {'name': 'Action Parameter', 'type': 'action', 'enabled':False},
    ]}]

## Create tree of Parameter objects
p = Parameter.create(name='params', type='group', children=params)
t = ParameterTree()
t.setParameters(p, showTop=False)

win = QtGui.QWidget()
layout = QtGui.QGridLayout()
win.setLayout(layout)
layout.addWidget(QtGui.QLabel("All of these parameters were initialized with enable=False"), 0,  0, 1, 1)
layout.addWidget(t, 1, 0, 1, 1)
win.show()
win.resize(800,800)

def attempt_disable():
    for param in p.child('Basic parameter data types').children():
        param.setOpts(enabled=False)

def print_enabled():
    for param in p.child('Basic parameter data types').children():
        print('\t', param.name(), "opts['enabled']:", param.opts['enabled'])

print("After initialization:")
print_enabled()
attempt_disable()
print("After calling param.setOpts(enabled=False):")
print_enabled()


if __name__ == '__main__':
    import sys
```

### Expected behavior
The parameter tree should show up with disabled parameters & widgets.

### Real behavior
All the parameters and widgets are enabled regardless of whether `param.opts['enabled']` is True or False. 

### Tested environment(s)

 * PyQtGraph version: 0.11.0
 * Python version: 3.6.10
 * NumPy version: 1.19.1
 * Operating system: Mac OSX 10.13
 * Installation method: git repository master branch

lidstrom83:
Edited by @j9ac9k to add subtask list

The following is a list of existing PRs/issues that relate to Parameter Tree's that this pull request aims to resolve.

- [x] Fixes #1393 
- [x] Fixes #1385 
- [x] Fixes #1006 
- [x] Fixes #760 
- [x] Fixes #339


lidstrom83:
This should resolve #1006

lidstrom83:
#339 is now resolved, provided the Parameter has been added to a `ParameterTree`, as the widget is used for bounds enforcement. Is this a reasonable expectation of the user? Note @campagnola 's [comment](https://github.com/pyqtgraph/pyqtgraph/blob/cd40562121b843a0da76bc407283f9967d4c2c4a/pyqtgraph/parametertree/Parameter.py#L170).

#347 I can't seem to reproduce it. Old Qt problem maybe?

#760 is now resolved.

lidstrom83:
Resolves #1393 and the immediate issue raised in #1385.

j9ac9k:
Marking this PR ready for review, as I feel that even though it's not addressing all the issues, there are enough fixes in here that we should merge and then address the remaining subtasks later.

j9ac9k:
Looks like there are some Qt6 specific issues in this PR that need to be resolved first.

j9ac9k (commented):
@pijyoi while working on getting this PR ready for merging, I encountered another set of enums that had type mismatches, specifically `QFontMetric.size(QtCore.Qt.TextFlag.TextSingleLine,...)`   With your code in `Qt.py`, the enum got promoted to `QtCore.Qt.TextSingleLine` but unfortunately it didn't like the type...

I added the extra `setattr` calls on the `TextFlag` enums; but I'm wondering if this is going to be something consistent, and if there is a better way to address this that you can think of.

Do you think this would be worthwhile to do with all the enums in the `QtCore.Qt` namespace?

pijyoi:
You are encountering the same problem as here:
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/AxisItem.py#L1120-L1126

The problem is that there are some Qt enums from different classes that were meant to be or-ed together. TextFlag being one of them.

I didn't shim them because:
1) I don't know if PyQt6 will eventually change them back to IntEnum. I seem to have read somewhere that something was going to change back to IntEnum.
2) There was only one place being affected by it
3) strong typing had already caught one bug: https://github.com/pyqtgraph/pyqtgraph/commit/c0336acc11bcebc8be961c345ee29e24bc31d31d

j9ac9k:
> You are encountering the same problem as here:
> https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/AxisItem.py#L1120-L1126
> 
> The problem is that there are some Qt enums from different classes that were meant to be or-ed together. TextFlag being one of them.
> 
> I didn't shim them because:
> 
> 1. I don't know if PyQt6 will eventually change them back to IntEnum. I seem to have read somewhere that something was going to change back to IntEnum.
> 2. There was only one place being affected by it
> 3. strong typing had already caught one bug: [c0336ac](https://github.com/pyqtgraph/pyqtgraph/commit/c0336acc11bcebc8be961c345ee29e24bc31d31d)
> 4. I was hoping a better solution could be found.

Moments after I posted that, I saw the code you are referencing, and promptly decided to do the same method for parameter tree... I'm fine w/ these if checks; they're not all over the place.

Thanks for chiming in.

j9ac9k:
I'm going to merge this PR now that it handles Qt6 accordingly.  Thanks @lidstrom83 for creating this!

pijyoi:
The following syntax now works in PyQt6 6.0.1, which should allow us to remove the special case for PyQt6 if we specify in README.md that the minimum supported PyQt6 is 6.0.1:
```int(QtCore.Qt.TextFlag.TextDontClip | QtCore.Qt.Alignment.AlignVCenter)```
```int(QtCore.Qt.TextFlag.TextDontClip)```
```int(QtCore.Qt.Alignment.AlignVCenter)```
i.e. TextFlag can or with Alignment, either can be cast to int.
The ability to cast to int seems to be a special case for these flags.

somewhat related to #1559

PR #1471, opened 2020-12-19
add unicode units

outofculture:
* just "µ" and "Ω"
* keep "u" and "Ohm" for compatibility
* add c, d, da and h prefixes
* lint

### DIFF
--- a/pyqtgraph/units.py
+++ b/pyqtgraph/units.py
@@ -12,8 +12,8 @@
 import unicodedata
 
 # All unicode identifiers get normalized automatically
-SI_PREFIXES = unicodedata.normalize("NFKC", "yzafpnµm kMGTPEZY")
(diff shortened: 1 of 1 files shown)


ksunden:
This will have to wait until after we actually drop 2.7 support... which is coming literally tomorrow, i think (I'm about to sit down and gather the changelog for a release, and once that is tagged, py2.7/qt4 is going away). (i.e. not worth reorganizing to work with 2.7)

outofculture:
@ksunden Oh, right. Here; now it runs in py2 and py3. The py2 globals won't be unicode, but everything else behaves the same in both.

j9ac9k:
@outofculture Thanks for the PR, we killed off py2 support, the looks good to me.  Before this PR I didn't even know of the units file...

Going to close/re-open to trigger a new CI run.

PR #1473, opened 2020-12-20
MAINT: Post 0.11.1 release, drop py2, qt4

ksunden:
This focuses on distribution, tests, and docs. This may not be comprehensive, but covers the cases I found by looking and a few greps

Noteably, this does not change any actual internal code yet, to avoid merge conflicts with pending PRs.

PR #1474, opened 2020-12-20
Migrate To Github Actions

j9ac9k:
Migrate our CI from Azure Pipelines to Github Actions due to 20 parallel jobs and potentially better web UI.

This PR addresses some other issues.

1. Resolves all warnings in the CI suite, now no-warnings should be emitted.
2. Introduce codeQL analysis checks
3. Updates the README with more badges 💯
4. No longer does it run `python setup.py style`, it just runs flake8 on the entire repo (which currently fails). 
5. No longer does it do the diff size check (this was problematic in re-implementing, and serves of minimal value IMO)
6. Add skips for pyopengl tests on macOS big sur for python < 3.9.1; due to those not working.

With not having to deal with miniconda, the CI runs _substantially_ faster than our previous azure pipelines config.  On a push, the CI runs in ~5 minutes (or less).  On a pull-request, due to the codeQL analyzer, it may take 15 minutes or so for it to finish.

One thing github actions does not support is passing artifacts from one stage to the next.  What I would do in azure pipelines is build a universal wheel, pass the wheel as an artifact, and install it in for all the test configurations.  Unfortunately that's not really feasible with the current capability of github actions, but for 20 free parallel jobs, I'm not about to complain.


j9ac9k:
CI is mostly working, the test suite is hanging on ubuntu/pyside2/python3.9 and I'm not seeing any terminal output.  Also I'm having to skip `test_reload` on all OSs for pyside2/python3.9.

I also had to not run any xvfb stuff before the test suite (like querying debug info, or the display resolution) as that was causing a segfault...still more to do.

j9ac9k:
Ok folks, I think this guy is ready for review... given the hour, I'm in no condition to actually merge anything; but all tests pass now, and it runs _fast_

j9ac9k:
that said, I probably should setup the screenshot collection/artifacts directory... and maybe i should evaluate the registry add for the windows mesa drivers to be done using powershell instead of `cmd`

ksunden:
Already breaking 2.7 I see (no worries there, #1473 will make it not install by default)

do you wish to do something to the style check for now (maybe just mark as okay to fail?) I'd rather not have red "CI failures" abound, even though there is nothing unexpected about the failure.

ksunden:
also, we can remove the azure configs, I think (easy to get them back if we feel the need)

j9ac9k:
> do you wish to do something to the style check for now (maybe just mark as okay to fail?) I'd rather not have red "CI failures" abound, even though there is nothing unexpected about the failure.

Some googling here is showing that there may not be a feature to allow a workflow to fail and still have a green checkmark... which begs the question, should we just ignore the style check for the time being?

Also the other thing I need to do is point to the new badge status...

ksunden:
https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/enabling-required-status-checks

in github settings you can say that only certain checks are required

j9ac9k:
> https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/enabling-required-status-checks
> 
> in github settings you can say that only certain checks are required

Looks like that's a setting that I will have to make after this PR is merged, which leaves badge + AZURE environment variable/screenshot artifacts...

Also, looking at the pytest github actions, I think I should merge all the CI stuff into one file, as I cannot pass artifacts from one stage to another, there is little value in having entirely separate stages...

j9ac9k:
only thing left to do is handle the AZURE environment variable, ... also I'm curious what these warnings are all about:

```
Calculated fingerprint of 6018441acaa95d75:1 for file pyqtgraph/flowchart/__init__.py line 2, but found existing inconsistent fingerprint value 7463d9a2b6ef8a3:1 
```

j9ac9k:
ok, I think this PR is now ready for merging

ksunden (approved):

ksunden:
LGTM, thanks for putting in the work @j9ac9k

PR #1477, opened 2020-12-23
Remove linting check, add pytest test publishing

j9ac9k:
Doing this not from my fork as debugging CI config stuff from a fork can be tricky...

This Pull Request removes the lint-check step (we know it fails, and it will keep failing until we can resolve these issues w/ some automatic formatting tool such as `black`). 

Furthermore, I now publish the unit test results jobs so it's a bit easier to see the results

Made some slight changes to `README.md` as well, fixing broken badge paths/links.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,4 +1,4 @@
-name: analyze pyqtgraph
+name: codeql
 
 on: pull_request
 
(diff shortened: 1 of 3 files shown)


j9ac9k:
Ok, I'm good with this for now; I want to remove some of the artifacts from displaying on the webUI, but no easy way to do that as of right now, so I'll revisit later.

PR #1479, opened 2020-12-25
minor adjustment on when test runs

j9ac9k:
the publish unit test results action we want doesn't work with `on: pull_request` for reasons described here: https://github.com/EnricoMi/publish-unit-test-result-action#support-fork-repositories

Instead we want to run on `pull_request_target`.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,6 +1,6 @@
 name: codeql
 
-on: pull_request
+on: [push, pull_request]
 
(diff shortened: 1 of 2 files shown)


j9ac9k:
Actually with this change, when someone submits a PR, we cannot see the CI results here.... the CI runs in the owners repo, which defeats the purpose IMO.... I must have something misconfigured...

j9ac9k:
Closing/Re-Opening to trigger new CI run.

PR #1487, opened 2021-01-03
Fix #798 ('InfLineLabel' object has no attribute 'updateText')

### LINKED ISSUE #798: AttributeError: 'InfLineLabel' object has no attribute 'updateText'
When I'm setting the `InfLineLabel` visibility, an attribute error was raised.

It is the function of `InFlineLabel` `setVisible` Line 404, I solved this by comment
```
self.updateText()
self.updatePosition()
```
and use the already exist function `valueChanged`. 

the code is:
```
def setVisible(self, v):
    TextItem.setVisible(self, v)
    if v:
        self.valueChanged()
        # self.updateText()
        # self.updatePosition()
```

ltirrell:
Implements the fix described in issue #798

### DIFF
--- a/pyqtgraph/graphicsItems/InfiniteLine.py
+++ b/pyqtgraph/graphicsItems/InfiniteLine.py
@@ -550,8 +550,7 @@ class InfLineLabel(TextItem):
     def setVisible(self, v):
         TextItem.setVisible(self, v)
         if v:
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for the PR @ltirrell this looks good to me!

Issue #1488, opened 2021-01-03
Regression: LinearRegionItem span parameter broken


2xB:
### Short description
The blue area inside `LinearRegionItem`s does no longer correctly apply `span` since PR #1391 ( tagging @jjuod ). Maybe a bounding box issue?

Happy new year btw.!

### Code to reproduce

Minimal version of `Plotting.py`:
```python
import numpy as np
import pyqtgraph as pg

app = pg.mkQApp()

x2 = np.linspace(-100, 100, 1000)
data2 = np.sin(x2) / x2
p8 = pg.plot(data2, pen=(255,255,255,200))

lr = pg.LinearRegionItem([400,700], span=(0.5,1))
p8.addItem(lr)

app.exec_()
```

### Expected behavior
Blue region should be in top half of the view, also according to the documentation of `span`: https://pyqtgraph.readthedocs.io/en/latest/graphicsItems/linearregionitem.html?highlight=LinearRegionItem .

![image](https://user-images.githubusercontent.com/31772910/103490021-3f6a1d80-4e19-11eb-91b1-9372febe35c0.png)

This screenshot was done with the last working commit 2c02d58134b3b9ef55e762de6410ce3edadcd9c4 .

### Real behavior
Blue region vanishes since merge of #1391 in 39f9c6a6aa3327ffee8bbf353ac5248237bf2ebe .

![image](https://user-images.githubusercontent.com/31772910/103490028-53ae1a80-4e19-11eb-8caf-0d14a172fc7f.png)

### Tested environment(s)

 * PyQtGraph version: Current `master` branch ( e94f53ac20eca89465c0e6853e0bb28e158c3c46 ). First occuring with commit 39f9c6a6aa3327ffee8bbf353ac5248237bf2ebe .
 * Qt Python binding: PyQt5 5.15.1 Qt 5.15.1
 * Python version: Python 3.8.5
 * NumPy version: 1.19.2
 * Operating system: Linux 5.8.16-2-MANJARO x86_64 20.1.2 Mikah
 * Installation method: Executed in folder

j9ac9k:
I keep thinking that this issue gets fixed.... welcome PRs to help address this. One attempt was made here:

https://github.com/pyqtgraph/pyqtgraph/pull/1452

jjuod:
[...]

PR #1491, opened 2021-01-06
setSpan broke LinearRegionItem (#1488)

jjuod:
setSpan was calling `setLeft()` etc on a rectangle retrieved from cache, thus modifying the cached one as well. A copy should be made instead.

### DIFF
--- a/pyqtgraph/graphicsItems/LinearRegionItem.py
+++ b/pyqtgraph/graphicsItems/LinearRegionItem.py
@@ -191,8 +191,8 @@ class LinearRegionItem(GraphicsObject):
         self.update()
 
     def boundingRect(self):
-        br = self.viewRect()  # bounds of containing ViewBox mapped to local coords.
-        
+        br = QtCore.QRectF(self.viewRect())  # bounds of containing ViewBox mapped to local coords.
+
         rng = self.getRegion()
         if self.orientation in ('vertical', LinearRegionItem.Vertical):
             br.setLeft(rng[0])


j9ac9k:
Thanks so much for this PR @jjuod !  PR failure is something we've seen intermittently; unrelated to this PR.  Will wait for things to finish and then merge.

2xB:
Thank you for the quick fix @jjuod! Along those lines, might it be reasonable if the function `viewRect()` already gave a copy of the `QRectF`? This - at least to my understanding, without having too much knowledge about the caching - seems like something that otherwise could cause problems in other places as well.

jjuod:
Sorry for the late reply... @2xB yes, totally. But I checked all the other functions that call `viewRect()` (there were only 2 or 3), and they make their own copies, so I didn't change that.
(In theory, the current way also allows saving a tiny amount of performance, since it is not always necessary to make a copy - only if the return is modified by `set*`.)

2xB:
@jjuod Don't worry about time, thank you for the response anyways! I'd argue that loosing this little amount of performance for something that might cause significant confusion to developers in the future is a reasonable trade-off. This might be a nitpick though. It's just that the time debugging something that should not be broken on paper according to all known documentation - to my knowledge the caching is currently not exposed in any documentation - always feels somewhat unsatisfying. E.g. I did not understand the source of this at all during debugging, I just found that it was not broken in commits pre-dating the caching. And that's me as someone debugging code in this library since some time. As said, probably a nitpick though.

PR #1216, opened 2020-05-30
Add cache for mapRectFromView

erikmansson:
In some cases, repeated calls to invertQTransform add up to become quite time consuming. This PR introduces a cache for it to speed things up.

PR #1493, opened 2021-01-09
invert QTransform using adjoint() and determinant()

pijyoi:
This PR is a result of discussions made in #1418.

It speeds up QTransform inversion using Qt routines adjoint() and determinant() rather than inverted().
The issue with inverted() is that it uses fuzzy floating point to classify matrix types in order to take faster inversion codepaths.

In addition, this PR would make #1216 redundant..
With the speedup in inversion, cache lookup runs slower than always computing the inversion.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -2324,26 +2324,32 @@ def isosurface(data, level):
         
     return vertexes, faces
 
-
     
+def _pinv_fallback(tr):
+    arr = np.array([tr.m11(), tr.m12(), tr.m13(),
+                    tr.m21(), tr.m22(), tr.m23(),
+                    tr.m31(), tr.m32(), tr.m33()])
+    arr.shape = (3, 3)
+    pinv = np.linalg.pinv(arr)
+    return QtGui.QTransform(*pinv.ravel().tolist())
+
+
 def invertQTransform(tr):
     """Return a QTransform that is the inverse of *tr*.
-    Rasises an exception if tr is not invertible.
+    A pseudo-inverse is returned if tr is not invertible.
     
     Note that this function is preferred over QTransform.inverted() due to
     bugs in that method. (specifically, Qt has floating-point precision issues
     when determining whether a matrix is invertible)
     """
     try:
-        import numpy.linalg
-        arr = np.array([[tr.m11(), tr.m12(), tr.m13()], [tr.m21(), tr.m22(), tr.m23()], [tr.m31(), tr.m32(), tr.m33()]])
-        inv = numpy.linalg.inv(arr)
-        return QtGui.QTransform(inv[0,0], inv[0,1], inv[0,2], inv[1,0], inv[1,1], inv[1,2], inv[2,0], inv[2,1])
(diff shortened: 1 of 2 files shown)


j9ac9k:
> With the speedup in inversion, cache lookup runs slower than always computing the inversion.

Well, that's interesting ... do you suggest effectively undoing #1216 ?

pijyoi:
I get the following results with the script included below:
Finished cache lookup: 16.0494 ms
Finished direct compute: 7.2663 ms

```python
import fileinput

import numpy as np
import PySide2
from PySide2 import QtGui

from pyqtgraph.util.lru_cache import LRUCache
from pyqtgraph.debug import Profiler

_mapRectFromViewGlobalCache = LRUCache(100, 70)


def invertQTransform(tr):
    try:
        det = tr.determinant()
        detr = 1.0 / det
        inv = tr.adjoint()
        inv *= detr
        return inv
    except ZeroDivisionError:
        raise


def cache_lookup(vt):
    cache = _mapRectFromViewGlobalCache
    k = (
        vt.m11(), vt.m12(), vt.m13(),
        vt.m21(), vt.m22(), vt.m23(),
        vt.m31(), vt.m32(), vt.m33(),
    )

    try:
        inv_vt = cache[k]
    except KeyError:
        inv_vt = invertQTransform(vt)
        cache[k] = inv_vt
    return inv_vt


transforms = [eval(line.strip()) for line in fileinput.input("transforms.txt")]

# use a size that results in always getting a cache hit
transforms = transforms[:50]

# populate cache
for tr in transforms:
    cache_lookup(tr)

profiler = Profiler(disabled=False, delayed=False)

for _ in range(100):
    [cache_lookup(tr) for tr in transforms]
profiler("Finished cache lookup")
for _ in range(100):
    [invertQTransform(tr) for tr in transforms]
profiler("Finished direct compute")
```

j9ac9k:
...hard to believe the unpacking of the QTransform takes so long....thanks for the code above, I'll try and benchmark on the macbook air tomorrow.

j9ac9k:
Ok, that was unexpected; 

```
  Finished cache lookup: 8.2791 ms
  Finished direct compute: 6.0091 ms
```

I think this PR should strip the cache lookup functionality unless we can identify some circumstance where it is beneficial.

pijyoi:
Maybe that should be a separate PR, ideally by the author (@erikmansson) of #1216.

j9ac9k:
I'll give @erikmansson some time to respond, but I think removing the cache hit can be in the same PR, provided it's in a different commit (and we don't squash on merge); of course alternatively it can just be a different PR.

erikmansson:
Please feel free to remove the cache again if the new method is quicker. I have no objections. 😊

j9ac9k:
@pijyoi I was right about to merge, you pushed the commit when I was ~5 seconds away from pushing the button :P no matter... thanks for updating the PR ... I'll merge when CI finishes up.

j9ac9k:
Thanks for the PR @pijyoi thanks for chiming in @erikmansson , merging

PR #1495, opened 2021-01-10
some fixes for PySide6 future compatibility

pijyoi:
fixup some code so that it will be future-compatible with PySide6

### DIFF
--- a/examples/ExampleApp.py
+++ b/examples/ExampleApp.py
@@ -16,6 +16,8 @@ if QT_LIB == 'PySide':
     from exampleLoaderTemplate_pyside import Ui_Form
 elif QT_LIB == 'PySide2':
     from exampleLoaderTemplate_pyside2 import Ui_Form
+elif QT_LIB == 'PySide6':
+    from exampleLoaderTemplate_pyside6 import Ui_Form
 elif QT_LIB == 'PyQt5':
     from exampleLoaderTemplate_pyqt5 import Ui_Form
 else:
--- a/examples/ExampleApp.py
+++ b/examples/ExampleApp.py
@@ -362,9 +362,7 @@ class ExampleLoader(QtGui.QMainWindow):
         self.ui.codeView.setLayout(self.codeLayout)
         self.hl = PythonHighlighter(self.ui.codeView.document())
         app = QtGui.QApplication.instance()
-        if QT_LIB in ['PyQt5', 'PySide2']:
-            # Qt4 does not have a paletteChanged signal
-            app.paletteChanged.connect(self.updateTheme)
+        app.paletteChanged.connect(self.updateTheme)
         self.codeLayout.addItem(QtGui.QSpacerItem(100,100,QtGui.QSizePolicy.Expanding,QtGui.QSizePolicy.Expanding), 0, 0)
         self.codeLayout.addWidget(self.codeBtn, 1, 1)
         self.codeBtn.hide()
--- a/examples/ExampleApp.py
+++ b/examples/ExampleApp.py
@@ -12,16 +12,9 @@ path = os.path.abspath(os.path.dirname(__file__))
 sys.path.insert(0, path)
 app = pg.mkQApp()
 
-if QT_LIB == 'PySide':
-    from exampleLoaderTemplate_pyside import Ui_Form
-elif QT_LIB == 'PySide2':
-    from exampleLoaderTemplate_pyside2 import Ui_Form
-elif QT_LIB == 'PySide6':
-    from exampleLoaderTemplate_pyside6 import Ui_Form
-elif QT_LIB == 'PyQt5':
-    from exampleLoaderTemplate_pyqt5 import Ui_Form
-else:
-    from exampleLoaderTemplate_pyqt import Ui_Form
+import importlib
+ui_template = importlib.import_module(
+    f'exampleLoaderTemplate_{QT_LIB.lower()}')
 
 examples = OrderedDict([
     ('Command-line usage', 'CLIexample.py'),
@@ -352,7 +345,7 @@ class PythonHighlighter(QSyntaxHighlighter):
 class ExampleLoader(QtGui.QMainWindow):
     def __init__(self):
         QtGui.QMainWindow.__init__(self)
-        self.ui = Ui_Form()
+        self.ui = ui_template.Ui_Form()
         self.cw = QtGui.QWidget()
         self.setCentralWidget(self.cw)
         self.ui.setupUi(self.cw)
--- a/pyqtgraph/GraphicsScene/GraphicsScene.py
+++ b/pyqtgraph/GraphicsScene/GraphicsScene.py
@@ -207,7 +207,7 @@ class GraphicsScene(QtGui.QGraphicsScene):
                     now = ptime.time()
                     init = False
                     ## keep track of which buttons are involved in dragging
-                    for btn in [QtCore.Qt.LeftButton, QtCore.Qt.MidButton, QtCore.Qt.RightButton]:
+                    for btn in [QtCore.Qt.LeftButton, QtCore.Qt.MiddleButton, QtCore.Qt.RightButton]:
                         if int(ev.buttons() & btn) == 0:
                             continue
                         if int(btn) not in self.dragButtons:  ## see if we've dragged far enough yet
--- a/pyqtgraph/GraphicsScene/GraphicsScene.py
+++ b/pyqtgraph/GraphicsScene/GraphicsScene.py
@@ -3,7 +3,7 @@ import time
 import weakref
 import warnings
 
-from ..Qt import QtCore, QtGui
+from ..Qt import QtCore, QtGui, isQObjectAlive
 from ..Point import Point
 from .. import functions as fn
 from .. import ptime as ptime
@@ -298,7 +298,9 @@ class GraphicsScene(QtGui.QGraphicsScene):
         for item in prevItems:
             event.currentItem = item
             try:
-                if item.scene() is self:
+                # NOTE: isQObjectAlive(item) was added for PySide6 where
+                #       verlet_chain_demo.py triggers a RuntimeError.
+                if isQObjectAlive(item) and item.scene() is self:
                     item.hoverEvent(event)
             except:
                 debug.printExc("Error sending hover exit event:")
--- a/pyqtgraph/colormap.py
+++ b/pyqtgraph/colormap.py
@@ -359,7 +359,13 @@ class ColorMap(object):
         
         pos, color = self.getStops(mode=self.BYTE)
         color = [QtGui.QColor(*x) for x in color]
-        g.setStops(list(zip(pos, color)))
+        stops = zip(pos, color)
+        if hasattr(g, 'setStops'):
+            g.setStops(list(stops))
+        else:
+            # PySide6 has a missing setStops binding
+            for pos, col in stops:
+                g.setColorAt(pos, col)
         return g
 
     def getColors(self, mode=None):
--- a/pyqtgraph/exporters/PrintExporter.py
+++ b/pyqtgraph/exporters/PrintExporter.py
@@ -45,7 +45,11 @@ class PrintExporter(Exporter):
         #res = printer.resolution()
         sr = self.getSourceRect()
         #res = sr.width() * .4 / (self.params['width'] * 100 / 2.54)
-        res = QtGui.QDesktopWidget().physicalDpiX()
+        try:
+            res = QtGui.QDesktopWidget().physicalDpiX()
+        except AttributeError:
+            # This is available since Qt 5
+            res = QtGui.QGuiApplication.primaryScreen().physicalDotsPerInchX()
         printer.setResolution(res)
         rect = printer.pageRect()
         center = rect.center()
--- a/pyqtgraph/exporters/PrintExporter.py
+++ b/pyqtgraph/exporters/PrintExporter.py
@@ -45,11 +45,7 @@ class PrintExporter(Exporter):
         #res = printer.resolution()
         sr = self.getSourceRect()
         #res = sr.width() * .4 / (self.params['width'] * 100 / 2.54)
-        try:
-            res = QtGui.QDesktopWidget().physicalDpiX()
-        except AttributeError:
-            # This is available since Qt 5
-            res = QtGui.QGuiApplication.primaryScreen().physicalDotsPerInchX()
+        res = QtGui.QGuiApplication.primaryScreen().physicalDotsPerInchX()
         printer.setResolution(res)
         rect = printer.pageRect()
         center = rect.center()
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,19 +9,34 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
-        qt-lib: [pyqt5, pyside2]
+        os: [ubuntu-20.04 , windows-latest, macos-latest]
+        qt-lib: [pyqt, pyside]
         python-version: [3.7, 3.8, 3.9]
         include:
           - python-version: "3.7"
-            qt-version: "~=5.12.0"
+            qt-lib: "pyqt"
+            qt-version: "PyQt5~=5.12.0"
+            numpy-version: "~=1.17.0"
+          - python-version: "3.7"
+            qt-lib: "pyside"
+            qt-version: "PySide2~=5.12.0"
             numpy-version: "~=1.17.0"
           - python-version: "3.8"
-            qt-version: "~=5.15.0"
+            qt-lib: "pyqt"
+            qt-version: "PyQt5~=5.15.0"
+            numpy-version: "~=1.19.0"
+          - python-version: "3.8"
+            qt-lib: "pyside"
+            qt-version: "PySide2~=5.15.0"
             numpy-version: "~=1.19.0"
           - python-version: "3.9"
-            qt-version: ""
-            numpy-version: ""
+            qt-lib: "pyqt"
+            qt-version: "PyQt5~=5.15"
+            numpy-version: "~=1.19.0"
+          - python-version: "3.9"
+            qt-lib: "pyside"
+            qt-version: "PySide6"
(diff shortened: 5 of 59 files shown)


pyqtgraph/GraphicsScene/GraphicsScene.py
```diff
@@ -207,7 +207,7 @@ def mouseMoveEvent(self, ev):
                     now = ptime.time()
                     init = False
                     ## keep track of which buttons are involved in dragging
-                    for btn in [QtCore.Qt.LeftButton, QtCore.Qt.MidButton, QtCore.Qt.RightButton]:
+                    for btn in [QtCore.Qt.LeftButton, QtCore.Qt.MiddleButton, QtCore.Qt.RightButton]:
```
j9ac9k, line 210: eek, looks like we should have made this change long ago!

pyqtgraph/colormap.py
```diff
@@ -359,7 +359,13 @@ def getGradient(self, p1=None, p2=None):
         
         pos, color = self.getStops(mode=self.BYTE)
         color = [QtGui.QColor(*x) for x in color]
-        g.setStops(list(zip(pos, color)))
+        stops = zip(pos, color)
+        if hasattr(g, 'setStops'):
+            g.setStops(list(stops))
+        else:
+            # PySide6 has a missing setStops binding
```
j9ac9k, line 366: is this documented anywhere?  the method is referenced in the docs:

https://doc-snapshots.qt.io/qt6-dev/qgradient.html#setStops

j9ac9k:
Hi @pijyoi thanks for this PR and tackling this issue!

I'm curious about the changes in all the templates files, for Qt6, we should likely be generating new template files, not modifying the existing ones.

If it would help, I can create new template files and add them to this PR, since they can be a bit annoying to generate.

pyqtgraph/colormap.py
pijyoi: It's present in Qt6 but the binding is missing in PySide6. It could just be an unintended omission.
https://doc.qt.io/qtforpython/PySide6/QtGui/QGradient.html

In any case, hasattr(QtGui.QLinearGradient, 'setStops') is True for PySide2 but False for PySide6.

pijyoi:
>I'm curious about the changes in all the templates files, for Qt6, we should likely be generating new template files, not modifying the existing ones.

The intention for these fixes is only to get the code "future"-ready. i.e. Minor fixes to convert usage of deprecated and obsolete APIs.

In this case, setWeight() in PySide6 no longer takes an integer, and in any case, the enum value 75 was no longer correct. Since setBold(True) already does the same thing as setWeight(Bold), removing the "weight" property altogether was the simpler solution.

>If it would help, I can create new template files and add them to this PR, since they can be a bit annoying to generate.

I did try re-generating a template using pyside2-uic, but this resulted in numerous differences from the existing file, which I thought would clutter the git history.
As the change in the .ui files was minor, I manually edited the auto-generated template files.

In any case, the manual edits are localized to a single commit.

I noticed that examples/VideoTemplate_pyside2.py, compared to the files for the other bindings, imports RawImageGLWidget unconditionally, such that the example will fail to run if pyopengl isn't installed. Any ideas why?

pijyoi:
On my local machine, I have manually "generated" all the *_pyside6.py files by making a copy of the corresponding *_pyside2.py files and changing PySide2 to PySide6.

Almost all the examples are able to launch and display something with PySide6
- Remote Plotting example has a "Request timed out" error
- HDF5 big data example was not tested

MatplotlibWidget.py was not updated.

ScatterPlotItem.py seems to have a PySide2-specific codepath:
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ScatterPlotItem.py#L37
I don't know if it should apply to PySide6.

j9ac9k:
Hi @pijyoi would you like to hop on our slack (so we can communicate outside of issue trackers)?  Is interested, feel free to email me, the email is in my github profile.

pijyoi:
Hi @j9ac9k , I am good to remain on github for now.

examples/ExampleApp.py
```diff
@@ -360,7 +362,7 @@ def __init__(self):
         self.ui.codeView.setLayout(self.codeLayout)
         self.hl = PythonHighlighter(self.ui.codeView.document())
         app = QtGui.QApplication.instance()
-        if QT_LIB in ['PyQt5', 'PySide2']:
+        if QT_LIB in ['PyQt5', 'PySide6']:
```
j9ac9k, line 365: this should be `QT_LIB in [PyQt5, PyQt6, PySide2, PySide6]`; but as we have deprecated Qt4 support, we can actually remove this if-statement/check, and un-indent the `paletteChanged.connect` call.

pyqtgraph/exporters/PrintExporter.py
```diff
@@ -45,7 +45,11 @@ def export(self, fileName=None):
         #res = printer.resolution()
         sr = self.getSourceRect()
         #res = sr.width() * .4 / (self.params['width'] * 100 / 2.54)
-        res = QtGui.QDesktopWidget().physicalDpiX()
+        try:
```
j9ac9k, line 48: Probably no need for a try/except now that we've killed off Qt4, just use `QtGui.QGuiApplication.primaryScreen().physicalDotsPerInchX()`

j9ac9k:
TIL about `tools/rebuildUi.py`, here I've been regenerating those template files like a chump...

> I did try re-generating a template using pyside2-uic, but this resulted in numerous differences from the existing file, which I thought would clutter the git history.
As the change in the .ui files was minor, I manually edited the auto-generated template files.

I've manually made small changes to these template files only to regret it.  I've been tempted to migrate the creation of these template files to be part of the CI process ....but for the time being I decided that is more trouble than its worth.

The reason for this change is due to a complete reimplementation of `uic`.  I actually had to regenerate a `VideoExample_pyside2` template as part of #1466 (which I know isn't merged yet, but will be shortly).



Should this PR also attempt to get PyQt6 compatibility as well now that its' on PyPi?

examples/ExampleApp.py
pijyoi: Right, this was a leftover from replacing PySide2 with PySide6 for testing.

pyqtgraph/exporters/PrintExporter.py
pijyoi: ok done

pijyoi:
>Should this PR also attempt to get PyQt6 compatibility as well now that its' on PyPi?

I have tried to make the changes in this PR to be less PySide6 specific.

I attempted to add PyQt6 support but it appears that it is going to require more effort than PySide6. Many enums values are not in their old places.

pijyoi:
For example, for the enum TransformationType.TxRotate, in PySide2, PySide6 and PyQt5, it can be accessed as 
1) QtGui.QTransform.TxRotate
2) QtGui.QTransform.TransformationType.TxRotate

In PyQt6, it is only accessible from the latter.

j9ac9k:
Thanks for looking into this @pijyoi 

I'll have to figure out how to redo the matrix of test dependencies defined here: 

https://github.com/pyqtgraph/pyqtgraph/blob/master/.github/workflows/main.yml#L11-L24

such that PySide6 is added to the test suite.  If you have any suggestions, I would say feel free to modify the python-3.9 config such that it uses Qt6.

Also, I'd be curious to get your take on incorporating QtPy (I know they don't currently support Qt6 but they will at some point).

pijyoi:
>such that PySide6 is added to the test suite. If you have any suggestions, I would say feel free to modify the python-3.9 config such that it uses Qt6.

I am not at all familiar with configuring the CI. Adding it is for sure going to find more errors in the less used parts of the library. Remote Plotting for instance has tracebacks and doesn't exit cleanly.

>Also, I'd be curious to get your take on incorporating QtPy (I know they don't currently support Qt6 but they will at some point).
 
As a first step, we would need to stop the emulation of Qt4 API and this is likely to require changes all over the place.
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/Qt.py#L250-L300

From what I can see, QtPy serves:
1) to allow code importing QtWidgets to still run on Qt4.
2) to prevent usage of Qt5 classes that do not exist in Qt4.

If we are dropping Qt4, these functionalities are not needed.

j9ac9k:
> I am not at all familiar with configuring the CI. Adding it is for sure going to find more errors in the less used parts of the library. Remote Plotting for instance has tracebacks and doesn't exit cleanly.

The remote plotting example is actually skipped during the test suite:

https://github.com/pyqtgraph/pyqtgraph/blob/d3049e41d77fb45f49113eae33d4712aa123553d/examples/test_examples.py#L61-L64

> As a first step, we would need to stop the emulation of Qt4 API and this is likely to require changes all over the place.

I'm wondering if it would be easier to emulate the Qt6 API instead of the Qt4 one...

Regarding QtPy, it was merely an idea to try and offload our need to maintain `Qt.py` with shims, but since we have dropped Qt4 support already (🎊), if you don't think it will provide any kind of tangible benefit; then no need to add the extra dependency.

Would you be ok with me pushing a commit to this PR adding PySide6 as one of the CI configs?

pijyoi:
>I'm wondering if it would be easier to emulate the Qt6 API instead of the Qt4 one...

For this project, Qt6 API is about the same as Qt5 API. The only shim needed was for QOpenGLWidget.

>Would you be ok with me pushing a commit to this PR adding PySide6 as one of the CI configs?

Yes, please do.

pijyoi:
>>I'm wondering if it would be easier to emulate the Qt6 API instead of the Qt4 one...

>For this project, Qt6 API is about the same as Qt5 API. The only shim needed was for QOpenGLWidget.

Strangely, QAction went from Qt4.QtGui to Qt5.QtWidgets and back to Qt6.QtGui.

j9ac9k:
I meant to push to my own branch, and then merge with this branch, but I messed that up, be prepared for numerous `git commit --amend` operations ...dialing in CI can be tricky as you can't easily test locally...

j9ac9k:
@pijyoi Seeing the results here, matplotlib exporter is going to have issues as that does not support a qt6 backend.... should we perhaps not allow the opening of the matplotlib widget if running Qt6? I'll ping @ixjlyons as he last worked on the matplotlib widget as he may have input.

pijyoi:
Isn't matplotlib only used for one of the possible exporters? Early adopters of PySide6 could just avoid using that functionality? I.e. we could just skip matplotlib testing. 

Ubuntu is failing fast with failure to load libOpenGL.so. I haven't actually tried PySide6 on Linux. Will try it out later.

j9ac9k:
> Isn't matplotlib only used for one of the possible exporters? Early adopters of PySide6 could just avoid using that functionality? I.e. we could just skip matplotlib testing.

yeah, when you export to matplotlib, you get a widget window with the plot that has the standard matplotlib interface/export tools... I'm good with skipping the test for the time being; would you like me to make the code change for that?

> Ubuntu is failing fast with failure to load libOpenGL.so. I haven't actually tried PySide6 on Linux. Will try it out later.

It's possible we may need to install `libopengl-dev` as part of the CI process.

ixjlyons:
> Seeing the results here, matplotlib exporter is going to have issues as that does not support a qt6 backend.... should we perhaps not allow the opening of the matplotlib widget if running Qt6? I'll ping @ixjlyons as he last worked on the matplotlib widget as he may have input.

I'd just skip it in the tests, at least for now. I imagine they'll get Qt6 support released soon enough that it shouldn't really be an issue.

pijyoi:
>standard matplotlib interface/export tools... I'm good with skipping the test for the time being; would you like me to make the code change for that?

Yes, please.

pijyoi:
>It's possible we may need to install libopengl-dev as part of the CI process.

PySide2 links to libGL.so.1 while PySide6 links to libOpenGL.so.0.


```console
$ ldd PySide2/QtOpenGL.abi3.so | grep GL\.so
	libQt5OpenGL.so.5 => .../lib/python3.8/site-packages/PySide2/Qt/lib/libQt5OpenGL.so.5 (0x00007feb519d4000)
	libGL.so.1 => /lib/x86_64-linux-gnu/libGL.so.1 (0x00007feb4f01d000)

$ ldd PySide6/QtOpenGL.abi3.so | grep GL\.so
	libQt6OpenGL.so.6 => .../lib/python3.8/site-packages/PySide6/Qt/lib/libQt6OpenGL.so.6 (0x00007f829e7be000)
	libOpenGL.so.0 => not found
	libOpenGL.so.0 => not found
	libEGL.so.1 => /lib/x86_64-linux-gnu/libEGL.so.1 (0x00007f829c338000)
	libOpenGL.so.0 => not found
```
It appears that libOpenGL.so.0 is installed by the proprietary NVidia driver. NVidia will also install libGL.so.1.
My Ubuntu 20.04 system without NVidia graphics only has libGL.so while my other system with NVidia graphics has both. 

I can however get pyqtgraph to run on Ubuntu 20.04 by providing a symbolic link from libGL.so to libOpenGL.so.
```console
ln -s /lib/x86_64-linux-gnu/libGL.so.1.7.0 /tmp/libOpenGL.so.0
LD_LIBRARY_PATH=/tmp PYQTGRAPH_QT_LIB=PySide6 python -m examples
```

I think PySide6 wheel may have been linked wrongly?

j9ac9k:
> I think PySide6 wheel may have been linked wrongly?

Perhaps? It's easy enough to add `apt-get install libopengl-dev` as part of the CI process to see what happens. I'll make that change (along w/ the matplotlib export skip) shortly

j9ac9k:
Sorry for spamming some CI change commits to your PR, feel free to rebase/squash those commits together.  Looks like Qt6 effectively needs ubuntu 20.04, github actions was supposed to migrate to that for `ubuntu-latest` by now, but as they haven't, I've have now specified it.  Also, it wasn't `libopengl-dev` that needed installing, but `libopengl0`.

A couple of notes:

`test_reload` was failing on pyside2/python3.9, whatever the reason it was failing there is likely present on pyside6. 

For the image tests, we probably want to add `PySide6` to this line here:
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/tests/image_testing.py#L260

pijyoi:
Is it possible to run the tests directly from the development directory without installing it?

I can run those under the tests subdir
```console
$ pytest pyqtgraph/tests/test_reload.py
```

but those under the graphicsItems subdir will complain about "No module named pyqtgraph"
```console
$ pytest pyqtgraph/graphicsItems/tests/test_ROI.py
```

pijyoi:
> > I think PySide6 wheel may have been linked wrongly?
> 
> Perhaps? It's easy enough to add `apt-get install libopengl-dev` as part of the CI process to see what happens. I'll make that change (along w/ the matplotlib export skip) shortly

https://bugreports.qt.io/plugins/servlet/mobile#issue/PYSIDE-1461

j9ac9k:
> Is it possible to run the tests directly from the development directory without installing it?
> 
> I can run those under the tests subdir
> 
> ```
> $ pytest pyqtgraph/tests/test_reload.py
> ```
> 
> but those under the graphicsItems subdir will complain about "No module named pyqtgraph"
> 
> ```
> $ pytest pyqtgraph/graphicsItems/tests/test_ROI.py
> ```

In the case of `test_reload` the pyqtgraph directory is added to the path, so you're able to import the pyqtgraph module by directory name

```
pgpath = os.path.join(os.path.dirname(pg.__file__), '..')
pgpath_repr = repr(pgpath)
```

The other directories don't attempt to do the same, and assume the pyqtgraph module is installed.

Is there a reason you're trying to run tests w/o installing the library?  When making a bunch of changes, I often install pyqtgraph in editable mode `pip install -e .` that way differences from branches, take affect right away.

pijyoi:
>Is there a reason you're trying to run tests w/o installing the library? When making a bunch of changes, I often install pyqtgraph in editable mode pip install -e . that way differences from branches, take affect right away.

Oh, I didn't know about "pip install -e"
Normally, my "install" of pyqtgraph is to just copy the pyqtgraph/pyqtgraph subdir.

j9ac9k:
> > Is there a reason you're trying to run tests w/o installing the library? When making a bunch of changes, I often install pyqtgraph in editable mode pip install -e . that way differences from branches, take affect right away.
> 
> Oh, I didn't know about "pip install -e"
> Normally, my "install" of pyqtgraph is to just copy the pyqtgraph/pyqtgraph subdir.

Highly recommend using editable installs as it best simulates an end-user interaction with the library.  There is one difference I've found with editable installs vs. non-editable pip installs; and that is accessing the example app can be different.

j9ac9k:
@pijyoi I think I know how to fix that git error that comes up on the macOS builds, I'm going to try and get the fix merged shortly... if CI here comes back all green after that, this PR ready to merge?

pijyoi:
>if CI here comes back all green after that, this PR ready to merge?

Yes, ready for merging.

j9ac9k:
> > if CI here comes back all green after that, this PR ready to merge?
> 
> Yes, ready for merging.

Sorry, forgot one last requested change, can you update the Qt Bindings test matrix in `README.md`, to add another row:

```
| Qt-Bindings    | Python 3.7         | Python 3.8         | Python 3.9         |
| :------------- | :----------------: | :----------------: | :----------------: |
| PySide2-5.12   | :white_check_mark: | :x:                | :x:                |
| PyQt5-5.12     | :white_check_mark: | :x:                | :x:                |
| PySide2-5.15   | :x:                | :white_check_mark: | :white_check_mark: |
| PyQt5-5.15     | :x:                | :white_check_mark: | :x:                |
| PySide6-6.0    | :x:                | :x:                | :white_check_mark: |
```

As well as the supported Qt binding up higher.

That's the last change I got...

j9ac9k:
@pijyoi do you have strong opinions if this should be squashed or merged with all the commits?

pijyoi:
Well, at the very least, the commit that adds the generated files should be separate.

But otherwise, I have already tried to squashed related things together. I try to keep each commit to only perform one type of fix, so that it can be cherry-picked or reverted. i.e I have a mild preference to keep the commits separate.

j9ac9k:
> I have a mild preference to keep the commits separate.

Good enough for me.

ksunden:
Woot

PR #1496, opened 2021-01-11
PlotItem: Fix enableMenu setting when passing a viewbox

dgoeries:
When passing a viewBox with **enableMenu=False** in the **PlotItem**, the setting is later overwritten and configured by the **PlotItem**.

This is a backward compatible fix that only the **enableMenu** is set when the viewbox is not passed. Otherwise, the setting of the viewBox is taken.


j9ac9k:
Thanks so much for the PR @dgoeries (the extra tests are 🥇 ), merging.

if you'd like a larger role in maintaining pyqtgraph, please reach out to me, you can find my email on my github profile page.

j9ac9k:
huh, weird, github isn't letting me merge...

PR #1500, opened 2021-01-19
bug: AlignCenter should have been AlignHCenter

pijyoi:
from the Qt documentation,
- "AlignCenter = AlignVCenter | AlignHCenter"
- "You can use at most one horizontal and one vertical flag at a time.
  Qt::AlignCenter counts as both horizontal and vertical"
- "Conflicting combinations of flags have undefined meanings"


j9ac9k:
Nice find; LGTM, merging.

PR #1502, opened 2021-01-20
Pyqt6 compat

pijyoi:
Building on #1495, this PR attempts to add some support for PyQt6. As the changes are more intrusive, it is preferred to keep this PR separate.

The main changes are due to:
- enum values are one level deeper
  * shimmed away in Qt.py
- enums and flags no longer decay to int
- classes overriding a virtual Qt method and wanting to call the parent method need to use super()
  * otherwise a recursion error will occur
- mouse events no longer have methods returning integer coords
  * convert to use float coord methods
  * shim to use older float coord method names

The CI has been modified to test Python 3.9 + PyQt6 and the changes here are sufficient to pass it.


j9ac9k:
> classes overriding a virtual Qt method and wanting to call the parent method need to use super()

If you tried this before removing py2 support, this would have been a real chore...

Thanks for tackling this!

j9ac9k:
CI failures on macOS I'm 99% sure have nothing to do with this PR, but our attempts to grab our image test data from another repository; and some race condition involving pytest-xdist.  I attempted to address this is issue in #1503 but looks like that patch doesn't completely solve the issue.  Likely the best thing to do is merge test-data with the main repository, but that's _well_ outside the scope of this PR.

j9ac9k:
Thanks for this PR @pijyoi !

This is a ton of work;...only thing I noticed is the `test_examples` bit skipped testing the examples for the following reason

```
SKIPPED [1] examples/test_examples.py:158: got empty parameter set ['frontend', 'f'], function testExamples at /Users/ogi/Developer/pyqtgraph/examples/test_examples.py:157
```

I should have caught this on the PySide6 PR you submitted a bit ago; but you need to add the relevant Qt bindings here as well:

https://github.com/pyqtgraph/pyqtgraph/blob/master/examples/test_examples.py#L37-L42

Test suite came back all green on my end with the change above; so after that change, and updating README.md; this is ready for merging IMO.

Thanks again!

pijyoi:
>.only thing I noticed is the test_examples bit skipped testing the examples for the following reason

Oh, I had always been running "pytest pyqtgraph", so never actually executed test_examples!

Although the CI passes, many codepaths (and thus bugs) only get activated while actually interacting with the examples.

j9ac9k:
> Although the CI passes, many codepaths (and thus bugs) only get activated while actually interacting with the examples.

Yup, that's why the test suite runs through all the test examples too... which I should point out, all the examples passed when I ran it, nice work 😎

j9ac9k:
diff LGTM, will you try and squash any of these commits together, or should I merge as-is (like last time, I won't squash on merge).

pijyoi:
I will do one more round of squashing.

j9ac9k:
If you're be interested in a greater role in maintaining pyqtgraph, please let us know.  I'm fairly sure you've written more lines in the library than I have; we can certainly use the help.

j9ac9k:
Thank you so much for this @pijyoi  pyqtgraph now fully supports Qt6!

pijyoi:
>>Although the CI passes, many codepaths (and thus bugs) only get activated while actually interacting with the examples.

>Yup, that's why the test suite runs through all the test examples too... which I should point out, all the examples passed when I ran it

Hmm, looks like I broke mouse interactions in RemoteGraphicsView.py for Qt 5.12. (This is what I meant by interacting with the examples.) I had previously only tested manually against Qt 5.15 and Qt 6.0 because this example fails to run under a venv. 
I will try to fix it.

pijyoi:
> > classes overriding a virtual Qt method and wanting to call the parent method need to use super()
> 
> If you tried this before removing py2 support, this would have been a real chore...

https://www.riverbankcomputing.com/pipermail/pyqt/2021-February/043607.html
A recent (2021/02/09) post by the PyQt author says that needing to use super() is a bug: "_That's a SIP code generator bug that is fixed in the next snapshot._"

Roughly, the following commits would have been unnecessary:
607db15949c5d7cc689417acf2c3f1e6bf8d9258
6dbda78231b10d4aa9be526569ec75e06564fd42
eec411a3c67d077768bb443daf20f5c9dd12780c

PR #1509, opened 2021-01-23
have mkQApp configure QApplication such that it handles HIDPI displays on Windows

### LINKED ISSUE #119: Incorrect scaling on high-resolution displays
This is particularly common on some OSX machines.

Example:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/pyqtgraph/nBeb9R3kkRo/Bh75r18MJC0J

### LINKED ISSUE #197: console "Exceptions" layout is broken
See "Exception Handling" text in screenshot below.

console_broken

### LINKED ISSUE #422: PyQt5, Retina Display, OpenGL windows wrong size.
Using PyQtGraph 0.10 with PyQt4 and an OpenGL widget (either imageview or GLViewWidget) results in appropriately sized images and widgets; but, as with all Qt4 things a bit blurry.

I have just moved my codebase to Qt5. With this, all the widgets seem to work and are at 'retina' resolution, apart from the GL widgets, which are at 'retina' resolution, but a quater of the expected size.

To fix you need to sprinkle *self.devicePixelRatio(), in places were there is self.width() and self.height() on things relating to the GL view port to make them the correct size.

i.e.

(0, 0, self.width(), self.height()) becomes

(0, 0, self.width()*self.devicePixelRatio(), self.height()*self.devicePixelRatio())

There are three places: 

opengl/GLViewWidget.py
def projectionMatrix()
def getViewPort()

widgets/RawImageWidget.py
def paintGL()

devicePixelRatio is present in PyQt4 (according to the docs), but I haven't checked if these changes are backward compatible.

BW

Matt

### LINKED ISSUE #756: axis and glyph scaling issue with Multiple DPI monitors
I am happy to check this out, but I don't really know where to start.

I am running Windows10, Anaconda x64, python 3.6.6, and the below images were created using the latest git version of pyqtgraph, and PyQt5.9

I have two monitors, one at 1080p (set to 100% scaling) and the other at 4K (set to 150% scaling), and when switching between the two monitors windows will attempt to adjust their scaling accordingly.  Some applications do this better than others, and this may in fact be an issue in Qt, but in case we can address it in the pyqtgraph code, that would probably be good.

The first image is on the 1080p monitor and the second is just from moving the window over to the 4K monitor.  The problem is most easily seen in the gray plots in the upper left and lower right.  You can see that the scatter clouds are centered at 0,0 in the 1080p plots, but are shifted to lower x and higher y coordinates in the 4k 150% plots.  This is the major issue because it means the plots aren't accurate in presenting the coordinates of the data.

Weirdly, it also seems that some of the glyphs are shrinking while others are growing when the dpi change happens

Lastly, I can confirm that if I switch the 1st monitor to 150% scaling before starting python, this issue does not happen.  It is only an issue when switching from a monitor at one dpi setting to a different one

## 1080p, 100% Scaling
image

## 4K, 150% Scaling
image

### LINKED ISSUE #1200: Some widget sizes (e.g. combo box) are not good on high dpi displays
### Code to reproduce

Parameter Tree pyqtgraph exaple:
image

### Expected behavior
The whole number should be displayed

### Real behavior
The top and bottom of the list parameter is truncated during editing

### Tested environment(s)

 * PyQtGraph version: '0.11.0.dev0+g720fa5f'
 * Qt Python binding: 'PyQt5 5.9.2 Qt 5.9.6'
 * Python version: 3.7
 * Operating system: Windows 10 Home build 1909
 * Installation method: pip

### LINKED ISSUE #1250: High DPI strikes again: Incorrect axis positions on dual-monitors with different resolution
### Short description

Creating a plot widget when you have two monitors with different resolutions causes the axes to be in the wrong locations / not scaled correctly with ImageItems.
### Code to reproduce

```python
import pyqtgraph as pg
from skimage import data


app = pg.mkQApp()
pg.setConfigOption('imageAxisOrder', 'row-major')
imgNp = data.chelsea()
imgItem = pg.ImageItem(image=imgNp)

plt = pg.PlotWindow()
plt.getViewBox().invertY(True)
plt.addItem(imgItem)
plt.show()
app.exec()
```

### Expected behavior

Axes should look the same on both monitors
### Real behavior

Axis scaling is correct on high-dpi monitor:
image

Axis scaling is strange on the low-dpi monitor:
image

### Tested environment(s)

 * PyQtGraph version: 0.11.0.dev0+g720fa5f
 * Qt Python binding: 5.13.0
 * Python version: 
 * NumPy version: 
 * Operating system: 
 * Installation method:

### LINKED ISSUE #1278: Various tests failing on Windows when running locally (CI platform is all green)
When running the test suite on Windows locally, I also got a variety of errors, despite the test suite passing on both macOS and on the CI pipelines.

The tests that failed specifically were

```
test_ViewBox
test_ErrorBarItem_defer_data
test_ImageItem_axisorder
test_PlotCurveItem
test_getArrayRegion
test_getArrayRegion_axisorder
test_PolyLineROI
```

Current config is PyQt5 5.15.0, numpy 1.18.5, python 3.7.7

### LINKED ISSUE #1417: Wrong position of plots on second display when using different windows:"Scale and layout" settings on both displays.
Hello,

i plotted some data with the following code:
`import pyqtgraph as pg
pg.plot([1,2,3,4],[3,6,4,5]) `

Using a computer with two desktops and windows 10.
On the first screen the windows setting "Scale and layout" is set to 125%, on the second screen the windows setting Scale and layout is set to 150%.

On the main screen the data is displayed right, on the second screen the data is plotted with some offset.
The offset seems visually to be the same independent of how much the plot is zoomed but the offset in axis values changes with the zoom.

Probably not relevant: main screen is full hd, second screen is 3840x2160.

### Short description
Plotting using displays with different windows setting "Scale and layout" the data gets displayed wrong on the second screen.

### Code to reproduce
```python
import pyqtgraph as pg
pg.plot([1,2,3,4],[3,6,4,5]) 
```

### Expected behavior
A plot where the line starts at x=1 y=3.

### Real behavior
A plot where the line starts somewhere else, dependant of the zoom and display.

### Tested environment(s)

 * PyQtGraph version:pyqtgraph       0.11.0
 * Qt Python binding: PyQt5.15.1
 * Python version: 3.8.6
 * NumPy version: 1.19.2
 * Operating system: windows 10
 * Installation method: pip

j9ac9k:
- [x] Fixes #1278 
- [x] Fixes #1250
- [x] Fixes #1200 
- [x] Fixes #756
- [x] Fixes #197
- [x] Fixes #1417
- [x] Fixes #422 
- [x] Fixes #119 (I think this was actually fixed long ago by upstream changes, but given the context I'll say this PR fixes it)

Still need to do some more testing, looks like Qt6 has great support for hidpi on windows from the get-go.  On Qt5, you need to set some environment variables, which this PR is meant to address.

Leaving this PR in draft form as it needs more testing.

Also need to verify fractional scaling; for reference here is how fractional scaling was addressed in matplotlib: https://github.com/matplotlib/matplotlib/pull/15656


j9ac9k:
This seems to address the issues with a single monitor; I have the equipment to check w/ multiple monitors with different DPI properties...will try verify against those.

j9ac9k:
Experimenting w/ this locally; I'm having a tough time getting all the above

- Text to remain not blurry on hidpi and non-hidpi displays
- `QCheckBox` 's  to _not_ be clipped (easily visible in `examples/parametertree.py`) on the non-hidpi display

Still have more experimentation to make...

j9ac9k:
🤦‍♂️ here I'm testing examples, wondering why nothing is behaving different, only to realize when running an example by itself, it spins up it's own `QApplication` and does not use `pg.mkQApp()` ... editing all the examples to use `mkQApp()` now...  There may be some caveats with Qt < 5.14; but for 5.14+ I think this will work reliably...

j9ac9k:
Special thanks for @QuLogic for creating https://github.com/matplotlib/matplotlib/pull/19123 which was the implementation I needed to see to fix #422

j9ac9k:
On Qt < 5.14, I would get the `QCheckBox` clipped in the parameter tree example on a non-scaled display.  That was really the only visual issue I could identify but not resolve.  The text in my scaled and non-scaled display was always sharp.

In the comments of #756 there was an interesting comment from @Androwei about using `ctypes` and on windows setting the process to be DPI aware, no matter how I used that example, the text on my second, non-hidpi display was _always_ fuzzy.  @t-sommer I see you linked to that issue and incorporated that solution in your app.  I would suggest taking a look to see if you notice the same fuzziness I saw on Windows, and if so, I would suggest either migrating to Qt5.14+ and incorporating the arguments in this PR (or better yet, use Qt6).

2xB:
This concept is great! Just wondering: You are not removing old High DPI-fixes there, so this should not change anything on old code, right? Is there any reason for using PyQtGraph without these flags you set there? In any case, it might be a good idea to write this also in the documentation for people that do not rely on mkQApp?

2xB:
PS: Can't stress it enough - Finally being able to just code for PyQtGraph without worrying about HighDPI screens is a really awesome sight!

j9ac9k:
Fwiw this only works if you user mkqapp, many applications using pyqtgraph
may instantiate qapplication themselves do the benefits here wouldn't do
anything.

j9ac9k:
> This concept is great! Just wondering: You are not removing old High DPI-fixes there, so this should not change anything on old code, right? Is there any reason for using PyQtGraph without these flags you set there? In any case, it might be a good idea to write this also in the documentation for people that do not rely on mkQApp?

I'm not aware of any dpi related fixes that were in the library before.

Also I'm not opposed to adding something to some documentation, but I'm not sure where it would go.

2xB:
@j9ac9k You are of course right. I think the great thing is that if this issue is brought up in the future, one can just point people to the documentation or the definition of `mkQApp` and ask them to use these lines even if they are not using `mkQApp`, instead of adapting further DPI related fixes. Such fixes would currently btw. be #668 and https://github.com/pyqtgraph/pyqtgraph/blob/ea08dda62dade523f2193ec353b9bacac6d8d35d/pyqtgraph/graphicsItems/DateAxisItem.py#L317 . For the documentation, I think 
https://pyqtgraph.readthedocs.io/en/latest/how_to_use.html#embedding-widgets-inside-pyqt-applications
 could be reasonable?

j9ac9k:
Oh right! I forgot about the date axis... I'll take a look there.

examples/ProgressDialog.py
```diff
@@ -9,7 +9,7 @@
 import pyqtgraph as pg
 from pyqtgraph.Qt import QtCore, QtGui
 
-app = QtGui.QApplication([])
+app = pg.mkQApp("Process Dialog Example")
```
pijyoi, line 12: s/Process/Progress/g

j9ac9k: thank you!

pijyoi: I am not sure about this calling of mkQApp() with an application name parameter. Where does it get shown?

2xB: This PR might actually be the first usage of that feature, but it was introduced in #1166 and further discussed in #1199 (don't judge the PR numbers - those are really the right ones!).

j9ac9k: with name parameter: 
image

without name parameter:
image

j9ac9k:
Regarding the fontScaleFactor bit; I'm unable to "break" it ... but I'm struggling to recall the condition that warranted its usage to begin with:

on non-hidpi external display: 
image

on hidpi primary display:
image

Looks much the same on Qt6.... I'm going to remove its usage and say it's good... if someone has a hidpi monitor that can verify `examples\DateAxisItem.py` works that would be great.

j9ac9k:
@2xB (or anyone else for that matter) ... I'm having trouble getting a link to work in the documentation:

In the section I just wrote, I have this line, wanting to link to `mkQApp`

```
PyQtGraph has a method :func:`mkQApp <pyqtgraph.Qt.mkQApp>` that by default...
```

Problem is that it does not appear that the contents of `Qt.py` are in the documentation, so the link doesn't work... so there is a docstring for `mkQApp`; but it doesn't look like sphinx picks it up?... Keep in mind I know next to nothing about sphinx, how do I go about getting that added in there?

j9ac9k:
Ok, I think this is ready to merge, I think @ksunden said he would try and verify on his hidpi display, I'll give him another day.  I think this guy is ready to merge unless there are further suggestions.

j9ac9k:
Alright, given I haven't heard anyone saying they've had issues, I'll merge, and await the hidpi issues to come in later 😆

andybarry:
Congratulations, this is fantastic!

PR #1511, opened 2021-01-24
Add almost all examples to ExampleApp

pijyoi:
Quite a number of examples in the examples folder are not accessible through the ExampleApp, whether intentionally or not. Adding them all has the benefit of having more CI coverage..

Not added are the examples multiprocess.py and parallelize.py as they seem to be out of the scope of a plotting library.

examples/utils.py seems to be unused, should it be deleted?


j9ac9k:
Ha, I was going to look to do this after getting my hidpi PR merged.

I have no idea why are some excluded.  I thought the `PColorMesh` example was in at one point but my memory could be wrong.

I don't know why so many examples are not listed (do they not work, are they no longer relevant, I have no idea).

j9ac9k:
> examples/utils.py seems to be unused, should it be deleted?

I remember something about this; I think we can get rid of it.... I think stuff got shuffled around in an attempt to make the ExampleApp itself part of the test suite, and part of that was effectively deprecating this, guess I missed removing the file outright.

pijyoi:
>I have no idea why are some excluded. I thought the PColorMesh example was in at one point but my memory could be wrong.

PColorMesh and NonUniformImage got added into utils.py.

>I don't know why so many examples are not listed (do they not work, are they no longer relevant, I have no idea).

A few examples are trivial and are already demonstrated in other more complex examples.

All the examples that I re-added seem to work, or at least don't error out.

Besides, multiprocess.py and parallelize.py, I also left out DateAxisItem_QtDesigner.py which is hardcoded to use PyQt5. I tried modifying it to run on the other bindings. It ran on PySide{2,6} but crashed on PyQt6, probably because of the multi-inheritance. @axil, you added it in #1154, could you comment on it?

pijyoi:
DateAxis_QtDesigner.py works now, crash not to do with the multi-inheritance, but seems to be a bug with PyQt6.

examples/DateAxisItem_QtDesigner.py
```diff
@@ -34,8 +37,19 @@ def __init__(self):
         self.plotWidget.setAxisItems({'bottom': pg.DateAxisItem()})
         self.plotWidget.showGrid(x=True, y=True)
 
+# Qt docs have the following warning regarding the following function:
+#   "To ensure that the application's style is set correctly, it is best
+#   to call this function before the QApplication constructor, if possible"
+QtWidgets.QApplication.setStyle('Fusion')
 app = QtWidgets.QApplication(sys.argv)
-app.setStyle(QtWidgets.QStyleFactory.create('Fusion'))
+if 0:
```
j9ac9k, line 45: Should this be an if statement something like `if pg.Qt.QT_LIB != "PyQt6":` ?

pijyoi: setStyle has been moved to be called before the constructor of QApplication, valid for all bindings, and recommended by the Qt docs.

I.e. the contents of "if 0:" are not needed for any binding. It's only left behind as documentation for the PyQt6 bug. But I see that this coding style triggers a static analyser warning about unreachable code, so maybe I should fully comment it out.

j9ac9k: Should the call to `setStyle` be made in `mkQApp`?

pijyoi: So I could:
1) add a style keyword to mkQApp
Or
2) don't set the style altogether

Because in the whole library, there is only this one usage of setStyle and only inside a small example. setStyle doesn't really add any functional value.

j9ac9k: I'm good with not setting the style altogether (assuming that this example would still work w/o the style being set).  `mkQApp` contains contents for the pyqtgraph examples to run the way they should, but we have no control over the `QApplication` instantiation end-users create.

2xB:
I remember that we e.g. have multiple `DateAxisItem` examples that are very similar and therefore did not include all in the ExampleApp. There is one issue with the app, which is: It is already quite an effort to find a specific example in the app if one does not remember its exact name. Having more similar-looking examples certainly does not help.

A solution to that problem would be to add a search bar to the ExampleApp and also maybe somehow highlight a few especially beautiful/interesting examples. I failed trying that though, because when I looked into it the last time, one needed pyuic4, pyuic5, pyside-uic and pyside2-uic to update all .py files when changing the ExampleApp ui file.

pijyoi:
I think the problem here is that the ExamplesApp.examples dictionary is being doubly used as a list of showcase examples and also for CI. In fact, the main motivation of this PR was to test all the examples to check for any Qt6 issues.

Would a possible solution be to have a separate list for examples that don't need to be in the ExamplesApp but should be run through the CI? test_examples.py would then be modified to read from both lists.

j9ac9k:
Testing the examples was never meant to expand test coverage, but a way to check to make sure the examples we provided to users would work; as that is how we recommend to users experiment with the library.

j9ac9k:
Now that I'm not on my mobile device; I can write out a bit longer answer.  Testing the examples was just a way to verify that the examples we were sending to our users would run (when I first hoped on the project, and started making changes, I was a bit embarrassed to discover I was breaking some of the examples).  

I thought about adding the code it went over to our "coverage" but in truth is the test code only looks for unhandled exceptions; and doesn't really have a check for anything; so while the test suite does indeed run through the test suite, I've never felt comfortable with saying the example is "correct" just because it ran.

Now, for the purposes of the Qt6 bindings, I would imagine they're helpful as more code-paths are being executed and hopefully more problematic mapping has become apparent.

> Would a possible solution be to have a separate list for examples that don't need to be in the ExamplesApp but should be run through the CI? test_examples.py would then be modified to read from both lists.

I have no problem with this, and maybe as an interim until the test suite is expanded a bit more, that may be a good solution.

pijyoi:
>A solution to that problem would be to add a search bar to the ExampleApp and also maybe somehow highlight a few especially beautiful/interesting examples.

A search bar is probably beyond me. As for the highlighting, I pushed a commit to allow bolding those examples considered to be interesting. It could be expanded to allow marking those examples not meant for showcasing, then a single combined list could be used.

j9ac9k:
I like a search-box too, but that's well outside the scope of this PR.

@pijyoi I'm good leaving things as bolded now, but I also have no problem removing examples where functionality is shown in another more-interesting example (that isn't one of the demos).  

Let's see how things look as they are a bit more options, and if we decide it is too much, we can trim as needed.

pijyoi:
I have categorized the examples into "examples", "others", "trivial".
"examples" is the original list found in utils.py and gets used by Example App.
"examples" and "others" are used by test_examples.py

j9ac9k:
@pijyoi this ready to merge?

pijyoi:
Yes ready for merging. Please squash all.

j9ac9k:
thanks for the PR @pijyoi

j9ac9k:
Somewhere in the middle of this PR, all macOS pipelines started exhibiting a timeout when testing examples.  Annoyingly, a timeout is not registered as a failure, so I didn't notice the issue until just now and started backtracking.

pijyoi:
I had just noticed it too and was trying to find which test it was. Another
variation that I had noticed was that MacOS CI runs on three workers while
Windows and Linux run on two.

pijyoi:
@j9ac9k , should be fixed by #1547

PR #1513, opened 2021-01-26
Attempt i18n localization

### LINKED ISSUE #649: How to translate the text on the context menu into local language?(i18n)
I want to internationalize the context menu and the export dialog, but there is no documens about that in pyqtgraph's documentation! How can I do it?
image

j9ac9k:
Fixes #649 

This PR is a _very_ early attempt to try and implement i18n localization to various strings that pyqtgraph shows to the user.  At the initial draft of this PR, this includes context menus in ImageView, ImageItem, ViewBox and PlotItem, as well as the Image Exporter and SVG Exporter.  Given that I do not have a system with a different localization, or know what the process here is, I could use some assistance to test the functionality here.

The process as shown in the PyQt5 docs here seems to indicate the following:

https://www.riverbankcomputing.com/static/Docs/PyQt5/i18n.html

* run `pylupdate5` to create a `.ts` file on a per-language basis
* a translator updates the `.ts` file with the relevant translations
* run `lrelease` on the `.ts` files.  In my case `lrelease` was not bundled with the PyQt5 distribution, but was part of the homebrew Qt package, can I could run it via `/usr/local/opt/qt/bin/lrelease`

More information on running `lrelease` can be found here: https://doc.qt.io/qt-5/linguist-manager.html#using-lrelease

### DIFF
--- a/pyqtgraph/canvas/Canvas.py
+++ b/pyqtgraph/canvas/Canvas.py
@@ -16,6 +16,7 @@ import gc
 from .CanvasManager import CanvasManager
 from .CanvasItem import CanvasItem, GroupCanvasItem
 
+translate = QtCore.QCoreApplication.translate
 
 class Canvas(QtGui.QWidget):
     
@@ -78,7 +79,7 @@ class Canvas(QtGui.QWidget):
             self.ui.redirectCombo.setHostName(self.registeredName)
             
         self.menu = QtGui.QMenu()
-        remAct = QtGui.QAction("Remove item", self.menu)
+        remAct = QtGui.QAction(translate("Context Menu", "Remove item"), self.menu)
         remAct.triggered.connect(self.removeClicked)
         self.menu.addAction(remAct)
         self.menu.remAct = remAct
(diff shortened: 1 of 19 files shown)


j9ac9k:
I think I'm pretty happy with these changes; I was debating adding a blurb in the documentation about this; but I'm likely the last person that _should_ do that given that I've never had to do it.

I think this PR is ready to merge as is.  I'll let it sit open for a bit in case anyone has any other feedback.

j9ac9k:
I took a glance at the docs, and in none of the places did it feel appropriate to add a blurb about this.  Fact is that pyqtgraph is just using the existing Qt functionality, and responsibility for doing the actual translations and such falls on the app developers, so I think I'm going to merge this as is.  `pyqtgraph` should certainly use the `translate` calls to make it so text _can_ be translated, which this PR achieves (I may have missed a spot, if that's the case, please open an issue, or better yet, submit a pull request addressing the issue).

If someone wants to add documentation on how to use Qt's translation tools for localizations, we'd welcome a PR, but I am not sure any of the current maintainers are well positioned to write a documentation to this effect.

Issue #1512, opened 2021-01-25
HistogramLUTItem sigLevelsChanged problem


Fokatu:
When I try to change the levels of an image of size 2048x2048 with HistogramLUTItem, there is obvious lag. Is this caused by too many “sigLevelsChanged” signals? Can we adjust the time interval between adjacent two signals?

j9ac9k:
We have SignalProxy can do rate limiting:

https://github.com/pyqtgraph/pyqtgraph/blob/23a46b5fb98b4d0cfadbb2cdb74a08c6b06f667d/pyqtgraph/SignalProxy.py#L12-L20

I don't use HistogramLUTItem much so I have no opinion if this should be implemented or not; if you can provide a minimal example (this is usually easiest done by modifying one of the examples) that would be helpful.

Fokatu:
@j9ac9k ，Thanks for your reply. The SignalProxy class works for me. But for bigger image, the problem appears again.

Here is a small example
`import numpy as np
from pyqtgraph.Qt import QtGui
import pyqtgraph as pg
pg.setConfigOptions(imageAxisOrder='row-major')

app = QtGui.QApplication([])

win = QtGui.QMainWindow()
win.resize(800,800)
imv = pg.ImageView()
win.setCentralWidget(imv)
win.show()

data = np.random.normal(size=(4096, 4096))
imv.setImage(data)

QtGui.QApplication.instance().exec_()
`

j9ac9k:
For this sort of thing in my application I've put the emitting of a signals into a queue of just _one_ entry and use a `QThread` to grab what's in the queue and do the computation there.  I'm not sure if that's an appropriate implementation for the library; but it might be.

Fokatu:
I have tried to update image in another thread. There will be no lag in the histogram widget, but it feel quite strange if the image update is not synchronous.

I have an idea, but I am not sure if it should be added into the library. You know, the generated qimage have only 256 gray scales, so I think for most of the level change signal, there should be no actual change in the qimage. Maybe we can add a extra if to deside whether sending the singal or not

j9ac9k:
@outofculture @ixjlyons  do you all have thoughts on this?  I don't work w/ the LUTs much so I'm not sure what a solution here would look like.

outofculture:
[...]

PR #1518, opened 2021-01-28
some optimizations for rescaleData

pijyoi:
1) use float32 instead of float64 for low precision inputs
2) use in-place operations to avoid unnecessary copies

Setting as draft for discussion

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -951,7 +951,12 @@ def rescaleData(data, scale, offset, dtype=None, clip=None):
     else:
         dtype = np.dtype(dtype)
     
-    d2 = data.astype(np.float) - float(offset)
+    if np.can_cast(data, np.float32):
+        work_dtype = np.float32
+    else:
+        work_dtype = np.float64
+    d2 = data.astype(work_dtype, copy=True)
+    d2 -= offset
     d2 *= scale
 
     # Clip before converting dtype to avoid overflow
@@ -959,13 +964,14 @@ def rescaleData(data, scale, offset, dtype=None, clip=None):
         lim = np.iinfo(dtype)
         if clip is None:
             # don't let rescale cause integer overflow
-            d2 = np.clip(d2, lim.min, lim.max)
+            np.clip(d2, lim.min, lim.max, out=d2)
         else:
-            d2 = np.clip(d2, max(clip[0], lim.min), min(clip[1], lim.max))
+            np.clip(d2, max(clip[0], lim.min), min(clip[1], lim.max), out=d2)
     else:
         if clip is not None:
-            d2 = np.clip(d2, *clip)
-    data = d2.astype(dtype)
+            np.clip(d2, *clip, out=d2)
+    # don't copy if no change in dtype
+    data = d2.astype(dtype, copy=False)
     return data
 
 


j9ac9k:
today I learned about `np.can_cast`; I'm glad we're doing in-place copies, is there a performance benefit here or is this strictly for memory saving operation?

outofculture:
On windows, at least, in-place operations (or otherwise using pre-allocated memory) are definitely a performance gain.

outofculture:
I ran some performance tests, a la `python examples/VideoSpeedTest.py --dtype uint16 --lut --size 2048x2048 --levels 0,7000`, with various flags, on my linux laptop:

<table>
<tr><th>notable flags</th><th>master</th><th>rescale_opt</th></tr>

<tr><td>--dtype uint8</td><td>20</td><td>23</td></tr>
<tr><td>--dtype uint8 --cuda</td><td>135</td><td>107</td></tr>

<tr><td>--dtype uint16</td><td>17</td><td>23</td></tr>
<tr><td>--dtype uint16 --cuda</td><td>136</td><td>107</td></tr>

<tr><td>--dtype float</td><td>17</td><td>18</td></tr>
<tr><td>--dtype float --cuda</td><td>90</td><td>72</td></tr>
</table>

( Side note: this would probably be a good time to use asv, if only I could get it to build the envs properly )

So not unambiguously beneficial on linux. Windows will probably see a more dramatic effect, but I don't have access to a windows environment right now.

j9ac9k:
> So not unambiguously beneficial on linux. Windows will probably see a more dramatic effect, but I don't have access to a windows environment right now.

I do; with cuda 11.1; ...but probalby won't get to testing this out later tonight.

pijyoi:
>I'm glad we're doing in-place copies, is there a performance benefit here or is this strictly for memory saving operation?

Doing in-place has the following performance benefits:
* avoids _**allocating**_ memory which is not a free operation.
* using **_less_** memory reduces pressure on memory bandwidth.
  * This is also the reason for using float32 over float64 when possible

j9ac9k:
on macOS I see some performance improvement

| notable flags | master | rescale_opt |
| :------------ | :------ | :---------- |
|  --dtype uint8 | 12 | 14 |
|  --dtype uint16 | 10 | 11 |
| --dtype float | 10 |  10 |

Will check windows later (with CuPy)

j9ac9k:
On my windows platform (granted not exactly the strongest machine), I saw minimal improvement

| notable flags 	 	| master | rescale_opt 	|
| :---    				| :--	 | :---		   	|
| --dtype uint8 		| 7 	 | 7		 	|
| --dtype uint8 --cuda 	| 33     | 33			|
| --dtype uint16 		| 6 	 | 6			|
| --dtype uint16 --cuda | 32     | 33			|
| --dtype float 		| 5      | 5 			|
| --dtype float --cuda  | 28     | 28 			|

pijyoi:
With the example in #1512 

**master**
> Entering functions.makeARGB
  check inputs: 0.0200 ms
  **apply levels: 791.3606 ms**
  apply lut: 0.0191 ms
  allocate: 0.0041 ms
  reorder channels: 85.1755 ms
  alpha channel: 11.0657 ms
< Exiting functions.makeARGB, total time: 889.6992 ms

**this PR**
> Entering functions.makeARGB
  check inputs: 0.0138 ms
  **apply levels: 489.8183 ms**
  apply lut: 0.0134 ms
  allocate: 0.0029 ms
  reorder channels: 66.6134 ms
  alpha channel: 10.1454 ms
< Exiting functions.makeARGB, total time: 568.9564 ms

outofculture:
@pijyoi what OS was that under? I ran the #1512 code with profiling, and saw a lot of variation - 50% - as I slid the histogram slider around, at least on a cpu. ( In contrast, the profiling while running the VideoSpeedTest only ever varied by a few percent. )

[master](https://github.com/pyqtgraph/pyqtgraph/files/5890940/master.txt)
[rescale_opt.txt](https://github.com/pyqtgraph/pyqtgraph/files/5890941/rescale_opt.txt)

At a glance, I do think rescale_opt averaged 10% faster.

In any case, this highlights a huge problem with trying to optimize in this space: code that's optimized for one scenario may be meaningless or even detrimental to other scenarios. Maybe we have to break these functions up to no longer be general, but to instead handle each specific shape+dtype+levels+lut+substrate(+os?)(+who knows what else) separately. When I tried to move some of this code to numba, that's what my (partial, broken) solution looked like. That sounds like a nightmare to maintain, but maybe things are too complex to handle any other way. What I really want is for a couple us to pair on this for a few days of devoted effort. We also need a diverse, automated, pan-platform benchmark suite, to even know what impact our changes are having.

( As an aside, using the VideoSpeedTest, I observe a 3fps/13% _improvement_ from increasing the resolution from `2048x2048` to `2048x2049`. Does anyone want to explain that? )

pijyoi:
>@pijyoi what OS was that under? I ran the #1512 code with profiling, and saw a lot of variation - 50% - as I slid the histogram slider around, at least on a cpu.

Windows 10 x64, Intel i5 dual-core 

>As an aside, using the VideoSpeedTest, I observe a 3fps/13% improvement from increasing the resolution from 2048x2048 to 2048x2049. Does anyone want to explain that? )

With power-of-twos sizes in the last dimension, you will experience associative cache evictions and TLB misses / thrashing when you access memory along the first dimension. (i.e. down the column)
So if you were to use 2049x2048, you would get the same performance as 2048x2048.
But 2048x2049 gets better performance as you don't get so many associative cache evictions.

It's a known trick to over-allocate memory, such that a 2048x2048 _logical_ array sits within a _physical_ 2048x2049 array. That could be achieved with numpy views. However, that would not be appropriate for a general library.

j9ac9k:
Should i try and get that benchmarking CI service set up?

pijyoi:
@outofculture , is it intentional that VideoSpeedTest.py maps command line parameter "--dtype float" to xp.float?
np.float is 64-bit double precision.
I would have thought that np.float32 would be the data type of choice for image display.

outofculture:
> I would have thought that np.float32 would be the data type of choice for image display.

You're probably right, @pijyoi! My change to the VideoSpeedTest code was mostly only `s/np/xp/g`, so I'm not a substantive author of what's happening in that file, but I fully expect your understanding to be correct. None of the image sources I've ever used in real life have been 64 bit, that much I can corroborate.

j9ac9k:
I think this PR looks good, you planning on making further changes @pijyoi ?

j9ac9k:
The only other thing I can think of, which is outside of the scope of this PR IMO is having `rescaleData` handle the case when `data` is 2D; to potentially allow to vectorize this bit of code:

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L1107-L1114

pijyoi:
>I think this PR looks good, you planning on making further changes @pijyoi ?

I have dropped the live update menu option from this PR. It was meant for testing only.

j9ac9k:
CI failures appear to be unrelated; ...really wish I could get to the bottom of those a bit more reliably ...anyway merging this, thanks @pijyoi

Issue #1519, opened 2021-01-28
Wrong usage of the translate function


speixoto:
@j9ac9k I think the final implementation will introduce some bugs. Check for example
on https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/exporters/ImageExporter.py#L104
this will not work anymore when the 'invertValue' is translated to a different language.

My suggestion would be to use the 'title' property of the Parameter as the translatable string while keeping the name in the english version

_Originally posted by @speixoto in https://github.com/pyqtgraph/pyqtgraph/issues/649#issuecomment-769045555_

speixoto:
#1520

PR #1520, opened 2021-01-28
Bugfix/wrong translate implementation

speixoto:


speixoto:
This PR should cover fix for #1519

j9ac9k:
This is certainly a _much_ cleaner way than my attempt at it, thank you for this PR @speixoto

PR #1525, opened 2021-01-30
switch to use of QOpenGLWidget in GraphicsView.py

pijyoi:
For Qt5, this switches from use of legacy QGLWidget to QOpenGLWidget.
For Qt6, this enables the option of using OpenGL, also using QOpenGLWidget.

Experimental plotting in PlotCurveItem gets broken by this switch and I do not know how to fix it, so we allow Qt5 users to opt back to using QGLWidget with the enableExperimental option.

There is some discussion here about issues with switching from QGLWidget to QOpenGLWidget when using them as a viewport.
https://stackoverflow.com/questions/31658176/using-qopenglwidget-with-qgraphicsview-viewport-not-updating

### DIFF
--- a/pyqtgraph/Qt.py
+++ b/pyqtgraph/Qt.py
@@ -354,7 +354,8 @@ if QT_LIB in [PYQT6, PYSIDE6]:
     # We're using Qt6 which has a different structure so we're going to use a shim to
     # recreate the Qt5 structure
 
-    QtWidgets.QOpenGLWidget = QtOpenGLWidgets.QOpenGLWidget
+    if not isinstance(QtOpenGLWidgets, FailedImport):
+        QtWidgets.QOpenGLWidget = QtOpenGLWidgets.QOpenGLWidget
 
 
 # Common to PySide, PySide2 and PySide6
--- a/pyqtgraph/Qt.py
+++ b/pyqtgraph/Qt.py
@@ -212,10 +212,6 @@ elif QT_LIB == PYQT5:
     except ImportError as err:
         QtSvg = FailedImport(err)
     try:
-        from PyQt5 import QtOpenGL
-    except ImportError as err:
-        QtOpenGL = FailedImport(err)
-    try:
         from PyQt5 import QtTest
         QtTest.QTest.qWaitForWindowShown = QtTest.QTest.qWaitForWindowExposed
     except ImportError as err:
@@ -249,10 +245,6 @@ elif QT_LIB == PYSIDE2:
         from PySide2 import QtSvg
     except ImportError as err:
         QtSvg = FailedImport(err)
-    try:
-        from PySide2 import QtOpenGL
-    except ImportError as err:
-        QtOpenGL = FailedImport(err)
     try:
         from PySide2 import QtTest
         QtTest.QTest.qWaitForWindowShown = QtTest.QTest.qWaitForWindowExposed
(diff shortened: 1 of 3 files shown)


j9ac9k:
Do you think it is worth attempting to try and get `QOpenGLWidget` to work with `PlotCurveItem` or do you think that's a lost cause until there is some upstream fix?

Looking at the Qt Bucktracker, I have found

* [QTBUG-86962](https://bugreports.qt.io/issues/?jql=text%20~%20%22qopenglwidget%20qgraphicsview%22%20ORDER%20BY%20created%20DESC)
* [QTBUG-82181](https://bugreports.qt.io/browse/QTBUG-82181?jql=text%20~%20%22qopenglwidget%20qgraphicsview%22%20ORDER%20BY%20created%20DESC)
* [QTBUG-73040](https://bugreports.qt.io/browse/QTBUG-73040?jql=text%20~%20%22qopenglwidget%20qgraphicsview%22%20ORDER%20BY%20created%20DESC) (this one is interesting, talks about background style-sheets, can apparently avoid when running w/o stylesheets)

I didn't notice any workarounds in those issues unfortunately.

j9ac9k:
On PyQt6, after removing the `enableExperimental` check in `PlotCurveItem.paint`, `examples/MouseSelection.py` works for me (with adding `pg.setConfigOptions(useOpenGL=True)` at the top of the file).

How was PlotCurveItem broken for you?

pijyoi:
I get the following on Windows 10 when using QOpenGLWidget on either Qt5 and Qt6. So maybe it is OS dependent.
qt6_opengl_mouse_selection

I didn't import QtOpenGL in Qt.py for Qt6 because it is not used. So, in the previous code, Qt6 users would get a QtOpenGL import failure and thus have HAVE_OPENGL = False.

I have modified the code to allow Qt6 users to turn on enableExperimental. After all, it is an opt-in.

j9ac9k:
I do have access to a Windows machine so I'll see if I can replicate there.

pijyoi:
I just tried it on Linux and it seems to work.

pijyoi:
>Given this is the only place that enableExperimental is used, should we just get rid of that option (and if we get rid of that option, should we ditch every reference of QGLWidget instead of QOpenGLWidget?)

As a user, I could want to enable useOpenGL without enabling enableExperimental. I would like to ditch QGLWidget even if that breaks Windows. Users affected are those who had opted in to enableExperimental.

j9ac9k:
I get the same problem on windows with PyQt5 5.15 and PyQt6.

j9ac9k:
This PR LGTM, code is much cleaner, no more ambiguity between QGLWidget vs. QOpenGLWidget... Thanks so much for doing this @pijyoi !

pijyoi:
Maybe the following code might give a clue to someone knowledgeable with Windows OpenGL.
The data gets drawn if the x and y coords lie between [-1, 1]. It's not drawn within the axes either.

```python
import numpy as np
import pyqtgraph as pg
pg.setConfigOption('useOpenGL', True)
pg.setConfigOption('enableExperimental', True)

app = pg.mkQApp()
win = pg.plot()

xxx = np.linspace(-1, 1, 1000)
yyy = np.sin(2*np.pi*3*xxx)
curve = pg.PlotCurveItem(x=xxx, y=yyy, pen='g', clickable=True)
win.addItem(curve)
              
app.exec_()
```
Capture

j9ac9k:
@pijyoi do you happen to have a hidpi display or use fractional scaling (150% or 175%?) by chance?

pijyoi:
No, I have neither as a monitor.

Issue #1524, opened 2021-01-29
The setValue method of Parameter does not work for list type parameter


Fokatu:
Am I doing it wrong?

below is an example

```
from pyqtgraph.Qt import QtCore, QtGui

app = QtGui.QApplication([])
from pyqtgraph.parametertree import Parameter, ParameterTree

params = [
        {'name': 'Integer', 'type': 'int', 'value': 10},
        {'name': 'List', 'type': 'list', 'values': [1, 2, 3], 'value': 2},
        {'name': 'Named List', 'type': 'list', 'values': {"one": 1, "two": "twosies", "three": [3, 3, 3]}, 'value': 2},
]

p = Parameter.create(name='params', type='group', children=params)

p.child("Integer").setValue(5)
p.child("List").setValue(["a", "b", "c"])
p.child("Named List").setValue({"a": "a", "b": "b"})

t = ParameterTree()
t.setParameters(p, showTop=False)

win = QtGui.QWidget()
layout = QtGui.QGridLayout()
win.setLayout(layout)
layout.addWidget(QtGui.QLabel("These are two views of the same data. They should always display the same values."), 0,
                 0, 1, 2)
layout.addWidget(t, 1, 0, 1, 1)
win.show()
win.resize(800, 800)


if __name__ == '__main__':
    import sys
    if (sys.flags.interactive != 1) or not hasattr(QtCore, 'PYQT_VERSION'):
        QtGui.QApplication.instance().exec_()

```

ixjlyons:
`setValue` on a `ListParameter` selects one of the options in the original list/dict. You can use `setLimits` to change the allowed values, i.e. in the above example:

```python
p.child("List").setLimits(["a", "b", "c"])
p.child("List").setValue("b")
```

Fokatu:
@ixjlyons , thanks!
Another question. How can I add suffix for int parameter? When I try to add suffix, it does not show in the spin

ixjlyons:
It looks like suffix/units aren't fully supported. As a workaround for now, you can specify the format string:

```python
params = [
    {'name', 'Integer', 'type': 'int', 'value': 10, 'suffix': 'V', 'format': '{value:d}{suffixGap:s}{suffix:s}'},
    ...
```

I believe updating this line to something like the above would help:

[...]

Issue #1429, opened 2020-11-06
Regression: Spinbox siPrefix with dimensionless quantities impossible


bilderbuchi:
### Short description
`siPrefix` is very useful when working with very high or low numbers, and I appreciate this a lot, sometimes having every prefix range from `n` to `k` in one UI. Sometimes the quantities are dimensionless, sometimes the units are not known to the program or tracked otherwise. 

Since pyqtgraph 0.11, spinboxes need an SI `suffix` when using `siPrefix`. This has the apparent consequence that one cannot work with dimensionless numbers (or unitless numbers) anymore.
I could not find an issue explaining the reasoning behind this change, only commit 6b798ffed856d8b2f7d53544ca4c480b2f7911d5 in #410 by @campagnola.

I have not found a way to work around this issue. I've tried passing `''` or a space `' '` into `suffix`, but that did not help.

Can you please fix this regression and re-enable using prefix-only with the Spinbox?
Should it be important for users to specifically switch on this behaviour, maybe an approach is to have `None` as a default for  `suffix`, and supply `suffix=''` to enable working with dimensionless numbers, which would be fine by me.

### Code to reproduce
```python
import pyqtgraph
from pyqtgraph.Qt import QtGui

app = QtGui.QApplication([])

w = pyqtgraph.QtGui.QMainWindow()
g = pyqtgraph.QtGui.QWidget()
l = pyqtgraph.QtGui.QHBoxLayout()
label = pyqtgraph.QtGui.QLabel()
label.setText('Reynolds number')
s = pyqtgraph.SpinBox(siPrefix=True)
display = pyqtgraph.QtGui.QLabel()
display.setText('value')
s.valueChanged.connect(lambda v: display.setText(str(v)))

g.setLayout(l)
w.setCentralWidget(g)

l.addWidget(label)
l.addWidget(s)
l.addWidget(display)

w.show()
app.exec()
```

### Expected behavior
The old behaviour behaves as expected, enabling use of prefixes with dimensionless quantities:
![image](https://user-images.githubusercontent.com/327442/98345859-81433700-2015-11eb-9753-efa704f3a075.png)

In another application, I am heavily using the siPrefix to enable easy entry and display of high/low numbers (with unity only being implicit due to device interface), and it works very well with 0.10:
![image](https://user-images.githubusercontent.com/327442/98347653-f9aaf780-2017-11eb-8c67-9995826adb15.png)

### Real behavior
[...]

PR #1528, opened 2021-01-30
Support suffix for int parameters

ixjlyons:
Follow-up from #1524. Changes here up for discussion.

Technically this could be seen as a breaking change for `SpinBox`. `siPrefix` now has no effect for `SpinBox` in "int mode". Previously, `SpinBox` in int mode forces the underlying value to be an int, but it shows you a scaled representation in accordance with the `siPrefix`.

The alternative here would be to just let the scaled formatting happen and the parameter value is still an integer. I could see that being somewhat confusing when you ask for an int parameter though. Maybe that's an argument for specifying the format string on the parameter side instead.

### DIFF
--- a/examples/SpinBox.py
+++ b/examples/SpinBox.py
@@ -34,6 +34,8 @@ spins = [
     ("Float with custom formatting", 
      pg.SpinBox(value=23.07, format='${value:0.02f}',
                 regex='\$?(?P<number>(-?\d+(\.\d+)?)|(-?\.\d+))$')),
+    ("Int with suffix",
+     pg.SpinBox(value=999, step=1, int=True, suffix="V")),
     ("Int with custom formatting", 
      pg.SpinBox(value=4567, step=1, int=True, bounds=[0,None], format='0x{value:X}', 
                 regex='(0x)?(?P<number>[0-9a-fA-F]+)$',
--- a/examples/parametertree.py
+++ b/examples/parametertree.py
@@ -81,6 +81,7 @@ params = [
     {'name': 'Numerical Parameter Options', 'type': 'group', 'children': [
         {'name': 'Units + SI prefix', 'type': 'float', 'value': 1.2e-6, 'step': 1e-6, 'siPrefix': True, 'suffix': 'V'},
         {'name': 'Limits (min=7;max=15)', 'type': 'int', 'value': 11, 'limits': (7, 15), 'default': -6},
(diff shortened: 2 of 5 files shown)


j9ac9k:
There is another potentially related issue, #1429

ixjlyons:
I saw that also, though it'll require a deeper fix to work when you manually input a string like "1 M".

Fixed the test to reflect the new formatting. Note I had to construct a new `SpinBox` each time because `setOpts` doesn't clear previously set options - that's not necessarily a problem in general but it is for a parametrized test.

j9ac9k:
> I saw that also, though it'll require a deeper fix to work when you manually input a string like "1 M".
> 
> Fixed the test to reflect the new formatting. Note I had to construct a new `SpinBox` each time because `setOpts` doesn't clear previously set options - that's not necessarily a problem in general but it is for a parametrized test.

Don't mean to bike-shed here; so feel free to disregard, but should we turn a blank spinbox into a pytest fixture that's reset/restored on each test?

ixjlyons:
I'm not sure a fixture is necessary, but using parametrize would be cool.

j9ac9k:
parametrize didn't even cross my mind, yeah this is a textbook case for that!

bilderbuchi:
As the reporter of #1429, is there any useful input I can give here, to maybe make its resolution more probable?

j9ac9k:
@bilderbuchi if you can confirm this fix resolves your issue that would be good.  I'm mostly holding off to give some other folks a chance to chime in with any concerns.

bilderbuchi:
> @bilderbuchi if you can confirm this fix resolves your issue that would be good. I'm mostly holding off to give some other folks a chance to chime in with any concerns.

Sadly, no, the behaviour is still like 0.11, i.e."5 m" in a `pyqtgraph.SpinBox(siPrefix=True)` just results in a value of 5, not 0.005 as it should. 
Based on [this comment](https://github.com/pyqtgraph/pyqtgraph/pull/1528#issuecomment-770272755) I did not expect this to work already. I thought maybe some assistance/input could be useful to "pave the way"/prepare the codebase for resolving my issue.

Btw, there is already a test for this case, but it is [currently not executed](https://github.com/pyqtgraph/pyqtgraph/pull/1528/files#diff-dcb1d987372406837f63720ec9177e3c4d130be8e7418abbb8360b70b4bcd173R41) because it does not work.

PR #1533, opened 2021-02-01
Remove scale rotate translate Qt4 shims

pijyoi:
The scale, rotate, translate methods were already marked as obsolete as far back as Qt 4.8.


j9ac9k:
This LGTM, I'm a bit concerned about some of conflicts on the exiting PRs; but most of them are old enough that we will already havre to deal with merge conflicts so not the end of the world.

ixjlyons:
Isn't this somewhat likely to break some user code given these methods' usage in several examples? Maybe it should go through a deprecation cycle?

pijyoi:
If we need to restore the shims, I would suggest to make them available
only for Qt5.

PR #1534, opened 2021-02-01
GraphIcons: Extend and deprecate previous pixmaps

dgoeries:
I saw that deprecation warnings come with print here. Hence, I went this way as well. Sorry for all the commits, have to clean my rebase.


pyqtgraph/icons/__init__.py
```diff
@@ -12,14 +12,20 @@ class GraphIcon:
 
     def __init__(self, path):
         self._path = path
-        self._icon = None
         name = path.split('.')[0]
```
j9ac9k, line 15: I know this line isn't part of this commit, but this line I suspect should be changed; or at least put in the docstring what the appropriate type for `path` should be.  Perhaps we should use `os.path.basename()` instead?

pyqtgraph/icons/__init__.py
```diff
@@ -12,14 +12,20 @@ class GraphIcon:
...
         _ICON_REGISTRY[name] = self
+        self._icon = None
+
+    def _build_qicon(self):
+        icon = QtGui.QIcon(op.join(op.dirname(__file__), self._path))
```
j9ac9k, line 20: is it supposed to be `os.path` instead of `op`?

pyqtgraph/icons/__init__.py
dgoeries: I put a better explanation

dgoeries:
Updated, the scatter plot crash is unrelated

pyqtgraph/icons/__init__.py
j9ac9k: My apologies for not sharing specific concerns  The use-case I'm imagining is someone tries to use an image/icon that is not part of the library, and they need to provide a path that has some directories in it.  While that would work w/ this code, the name would include the whole path before the first period.  If we want just the filename we should likely use `name = os.path.basename(path)`


In the docstring you do specify that the images must be in the library; but is that a necessary restriction?

dgoeries: this should not be a user facing feature in that sense. That is why it is called `GraphIcon`

j9ac9k:
LGTM, thanks for the PR @dgoeries  merging.

PR #1540, opened 2021-02-03
Support siPrefix with no suffix in SpinBox

### LINKED ISSUE #1429: Regression: Spinbox siPrefix with dimensionless quantities impossible

ixjlyons:
Fixes #1429 

Having different behavior for `suffix=None` and `suffix=""` in `siParse` seems maybe a little awkward, but I think the overall behavior is pretty reasonable looking at the parametrized test cases. Input "1.2 m" with no other information is ambiguous, so there has to be some mechanism to force it to be interpreted as prefix milli rather than an arbitrary suffix.

@bilderbuchi if you could try this out and confirm all the spin boxes work as you'd expect, that'd be appreciated. If you have suggestions for additional test I/O, that would be nice too.

edit: I thought about modifying the tests for `SpinBox`, but I'd rather do that if/after #1528 is merged given the refactor there

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -139,7 +139,8 @@ def siParse(s, regex=FLOAT_REGEX, suffix=None):
         Suffix to check for in ``s``. The default (None) indicates there may or
         may not be a suffix contained in the string and it is returned if
         found. An empty string ``""`` is handled differently: if the string
-        contains a suffix, it is discarded.
+        contains a suffix, it is discarded. This enables interpreting
+        characters following the numerical value as an SI prefix.
     """
     s = asUnicode(s)
     s = s.strip()
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -119,10 +119,27 @@ def siFormat(x, precision=3, suffix='', space=True, error=None, minVal=1e-25, al
 
 def siParse(s, regex=FLOAT_REGEX, suffix=None):
     """Convert a value written in SI notation to a tuple (number, si_prefix, suffix).
-    
+
     Example::
-    
-        siParse('100 μV")  # returns ('100', 'μ', 'V')
+
+        siParse('100 µV")  # returns ('100', 'µ', 'V')
+
+    Note that in the above example, the µ symbol is the "micro sign" (UTF-8
+    0xC2B5), as opposed to the Greek letter mu (UTF-8 0xCEBC).
+
+    Parameters
+    ----------
+    s : str
+        The string to parse.
+    regex : re.Pattern, optional
+        Compiled regular expression object for parsing. The default is a
+        general-purpose regex for parsing floating point expressions,
+        potentially containing an SI prefix and a suffix.
+    suffix : str, optional
+        Suffix to check for in ``s``. The default (None) indicates there may or
+        may not be a suffix contained in the string and it is returned if
+        found. An empty string ``""`` is handled differently: if the string
+        contains a suffix, it is discarded.
     """
     s = asUnicode(s)
     s = s.strip()
@@ -130,15 +147,20 @@ def siParse(s, regex=FLOAT_REGEX, suffix=None):
         if s[-len(suffix):] != suffix:
             raise ValueError("String '%s' does not have the expected suffix '%s'" % (s, suffix))
         s = s[:-len(suffix)] + 'X'  # add a fake suffix so the regex still picks up the si prefix
-        
+
+    # special case: discard any extra characters if suffix is explicitly empty
+    if suffix == "":
+        s += 'X'
+
     m = regex.match(s)
     if m is None:
         raise ValueError('Cannot parse number "%s"' % s)
+
     try:
         sip = m.group('siPrefix')
     except IndexError:
         sip = ''
-    
+
     if suffix is None:
         try:
             suf = m.group('suffix')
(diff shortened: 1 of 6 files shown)


bilderbuchi:
Wow, this is looking great, thank you! Works as expected, as far as I can tell. 
It even handles weirder entries naturally, like `5e-3 M` becomes `5 k` (value 5000) 😁

bilderbuchi:
If you need a motivational example for this change, e.g. for the docs, I suggest using [dimensionless quantities](https://en.wikipedia.org/wiki/List_of_dimensionless_quantities), a famous one that is often very large (>1e6) is the [Reynolds number](https://en.wikipedia.org/wiki/Reynolds_number) from fluid mechanics.

j9ac9k:
Merged #1528 sorry it took me a bit to get that through.

ixjlyons:
Great. Rebased, updated the `SpinBox` tests, and added a couple notes to the docs.

pyqtgraph/functions.py
```diff
@@ -119,35 +119,58 @@ def siFormat(x, precision=3, suffix='', space=True, error=None, minVal=1e-25, al
...
-    
+
     Example::
-    
-        siParse('100 μV")  # returns ('100', 'μ', 'V')
+
+        siParse('100 µV")  # returns ('100', 'µ', 'V')
+
+    Note that in the above example, the µ symbol is the "micro sign" (UTF-8
+    0xC2B5), as opposed to the Greek letter mu (UTF-8 0xCEBC).
```
j9ac9k, line 128: nice note, I did not know this!

j9ac9k:
Thanks @ixjlyons this LGTM, merging!

pyqtgraph/functions.py
ixjlyons: Funny you mention that - I didn't either until figuring out why example that was already in the docstring wasn't working. Hex editor was vital on this one!

bilderbuchi: I can see why that can be useful for easier classification/consistency, but these essentially identical-looking (homoiconic?) UTF symbol are really annoying at times. See also different kinds of spaces, or also invisible spaces.

pyqtgraph/widgets/SpinBox.py
```diff
@@ -132,8 +132,10 @@ def setOpts(self, **opts):
         siPrefix       (bool) If True, then an SI prefix is automatically prepended
                        to the units and the value is scaled accordingly. For example,
                        if value=0.003 and suffix='V', then the SpinBox will display
-                       "300 mV" (but a call to SpinBox.value will still return 0.003). Default
-                       is False.
+                       "300 mV" (but a call to SpinBox.value will still return 0.003). In case
+                       the value represents a dimensionless quantity that might span many
+                       orders of magnitude, such as a Reynold's number, an SI
```
bilderbuchi, line 137: slight nitpick, the guy's name was Reynolds, so it's "a Reynolds number", no apostrophe.

bilderbuchi:
Great, thank you everybody! ❤️

pyqtgraph/widgets/SpinBox.py
j9ac9k: that is as slight of a nitpick as nitpicks get! 😆 I'd merge a PR making that change.

bilderbuchi: Hey, what can I say, I am/was a fluid dynamics guy :D

Issue #1535, opened 2021-02-01
On changing undeprecated behavior


irgolic:
Hi, I'm Rafael, one of the https://github.com/biolab/orange3 core dev team. We've depended on pyqtgraph since the Qt4 age for a number of our visualizations, including scatter plots and heatmaps.

We run an automated future check against the master branches of a couple of repositories, pyqtgraph included. Sometimes when a bug is fixed upstream, behavior changes; there existed code that uses the bug as a feature. These changes in behavior are desired, even though they may incur some upkeep downstream.

Some breaking changes occurred in the recent releases of pyqtgraph (https://github.com/biolab/orange3/pull/5046). We clamped it at first (https://github.com/biolab/orange3/pull/5082), but pyqtgraph plays a big role in Orange, so we maintain compatibility (https://github.com/biolab/orange3/pull/5207).

More recently, https://github.com/pyqtgraph/pyqtgraph/pull/1533 got merged – monkeypatched methods back from the Qt4 migration were removed. Thanks for bringing attention to these, so we're able to fully migrate our widgets to Qt5 (https://github.com/biolab/orange3/pull/5231). However, I'd agree with the conversation in the PR regarding a deprecation cycle preceding removal. From our point of view, part of the Qt API just disappeared.

I'm writing this issue to start a discussion about maintaining API and behavioral consistency. But I do realize that pyqtgraph is an `0.x` versioned project – generally in versioning, the major version number guarantees consistency.

Would you consider going `1.0.0`? Or idk, running a test for a downstream dependency? What do you guys think?

Love and thanks,
Rafael

j9ac9k:
Hi Rafael,

Thanks for reaching out.  My apologies for the head ache as I am the one that merged the PR.  I certainly didn't mean to cause downstream trouble.   While we are trying to move quick through a long backlog of issues and PRs that goes back for years, that's no justification for killing off portions of the public API that other libraries rely on; in the future we'll certainly be more mindful of going through a deprecation process.  Also, FWIW this is the first library I've maintained with a user size that's greater than 1 digit; so I certainly welcome input from people that use pyqtgraph.

Back to the specific issue here though, what would you like a deprecation phase to look like? 

[...]

PR #1541, opened 2021-02-04
use qWaitForWindowExposed instead of qWaitForWindowShown

pijyoi:
qWaitForWindowShown has been obsolete since Qt 5.6

In view of the comments in #1533  and #1535, the shim is left behind for Qt5 but removed for Qt6.
However, can this particular shim really be considered part of pyqtgraph's public API given that it is only used in pyqtgraph's unit tests?

### DIFF
--- a/pyqtgraph/Qt.py
+++ b/pyqtgraph/Qt.py
@@ -232,7 +232,6 @@ elif QT_LIB == PYQT6:
         QtOpenGLWidgets = FailedImport(err)
     try:
         from PyQt6 import QtTest
-        QtTest.QTest.qWaitForWindowShown = QtTest.QTest.qWaitForWindowExposed
     except ImportError as err:
         QtTest = FailedImport(err)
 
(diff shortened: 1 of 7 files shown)


j9ac9k:
I think we're good killing this off altogether, appreciate the concern given the feedback; doing a quick search on github for `pyqtgraph.QtTest` and `pg.QtTest` yields no results.

I'm not strongly opinionated on this, but if we did decide to deprecate it, we should try and emit a deprecation warning when someone would access that module.

pijyoi:
The problem with monkey patching the module directly:
```
>>> import pyqtgraph
>>> from PySide2 import QtTest
>>> QtTest.QTest.qWaitForWindowShown
<built-in method qWaitForWindowExposed of Shiboken.ObjectType object at 0x00000205FF74A4A8>
```
This is how #1533 affected biolab/orange3.

j9ac9k:
If we can't put a deprecation notice on this..I'm good with killing this off.

j9ac9k:
LGTM, thanks @pijyoi

PR #1546, opened 2021-02-06
Guard against key events when there's no time axis in ImageView

### LINKED ISSUE #1545: Arrow keys in ImageView Without Time Axis
### Short description
When you type an arrow key in to a ImageView without a time axis, the application crashes with a unhandled exception.

### Code to reproduce
```
from PySide2.QtWidgets import QMainWindow, QAction, qApp, QApplication
from PySide2.QtCore import Qt

import numpy as np
import pyqtgraph as pg

class BugWindow(QMainWindow):
    def __init__(self):
        QMainWindow.__init__(self)
        self.resize(800, 800)
        self.imv = pg.ImageView()
        self.setCentralWidget(self.imv)



app = QApplication([])
win = BugWindow()
win.show()

app.exec_()

```

### Expected behavior
Nothing, 

### Real behavior
An unhandled exception occurs.

```
  File "D:\Repositories\Art\pyqtgraph\pyqtgraph\imageview\ImageView.py", line 452, in keyPressEvent
    self.evalKeyState()
  File "D:\Repositories\Art\pyqtgraph\pyqtgraph\imageview\ImageView.py", line 476, in evalKeyState
    self.jumpFrames(1)
  File "D:\Repositories\Art\pyqtgraph\pyqtgraph\imageview\ImageView.py", line 516, in jumpFrames
    if self.axes['t'] is not None:
KeyError: 't'

```

### Tested environment(s)

 * PyQtGraph version: 0.11.1.dev0/commit: cd914495ece
 * Qt Python binding: 'PySide2 5.13.2 Qt 5.12.9'
 * Python version: 3.9.1
 * NumPy version: '1.19.5'
 * Operating system:  Windows 10
 * Installation method: conda and git

### Additional context
I have a patch at https://github.com/pknodle/pyqtgraph/tree/image-display-bugfix
I'm not sure if this fix affects any other part of the system.

ixjlyons:
Fixes #1545. `ImageView` should be robust to GUI interaction when there's no data and/or no time axis.

This is kind of the minimal change to fix the issue. If `ImageView` is ever extended to take other keyboard shortcuts (either internally or in a user's subclass), there are still potentially codepaths that would throw an exception, such as `setCurrentIndex`. It may be worth peppering in a few more of these checks.

### DIFF
--- a/pyqtgraph/imageview/ImageView.py
+++ b/pyqtgraph/imageview/ImageView.py
@@ -429,7 +429,10 @@ class ImageView(QtGui.QWidget):
         self.setParent(None)
         
     def keyPressEvent(self, ev):
-        #print ev.key()
+        if not self.hasTimeAxis():
+            super().keyPressEvent(ev)
+            return
+
         if ev.key() == QtCore.Qt.Key_Space:
             if self.playRate == 0:
                 self.play()
@@ -454,6 +457,10 @@ class ImageView(QtGui.QWidget):
             super().keyPressEvent(ev)
 
     def keyReleaseEvent(self, ev):
+        if not self.hasTimeAxis():
+            super().keyReleaseEvent(ev)
+            return
+
         if ev.key() in [QtCore.Qt.Key_Space, QtCore.Qt.Key_Home, QtCore.Qt.Key_End]:
             ev.accept()
         elif ev.key() in self.noRepeatKeys:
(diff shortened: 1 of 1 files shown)


j9ac9k:
Isn't that `super()` python3 syntax wonderful?

LGTM, thanks @ixjlyons merging

PR #1547, opened 2021-02-06
remove ProgressDialog from CI test

pijyoi:
It is not suitable as it creates a modal dialog. On MacOS, this triggers the CI 30 sec timeout.

### DIFF
--- a/examples/utils.py
+++ b/examples/utils.py
@@ -108,7 +108,6 @@ others = dict([
     ('DateAxisItem_QtDesigner', 'DateAxisItem_QtDesigner.py'),
     ('GraphicsScene', 'GraphicsScene.py'),
     ('MouseSelection', 'MouseSelection.py'),
-    ('ProgressDialog', 'ProgressDialog.py'),
 ])
 
 
@@ -117,4 +116,9 @@ trivial = dict([
     ('SimplePlot', 'SimplePlot.py'),    # Plotting.py
     ('LogPlotTest', 'LogPlotTest.py'),  # Plotting.py
     ('ViewLimits', 'ViewLimits.py'),    # ViewBoxFeatures.py
+])
+
+# examples that are not suitable for CI testing
+skiptest = dict([
+    ('ProgressDialog', 'ProgressDialog.py'),    # modal dialog
 ])
--- a/examples/test_examples.py
+++ b/examples/test_examples.py
@@ -70,10 +70,6 @@ conditionalExamples = {
         False,
         reason="Test is being problematic on CI machines"
     ),
-    "ProgressDialog.py": exceptionCondition(
-        not(platform.system() == "Linux"),
-        reason="QXcbConnection: XCB error"
-    ),
     'GLVolumeItem.py': exceptionCondition(
         not(platform.system() == "Darwin" and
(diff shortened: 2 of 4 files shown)


j9ac9k:
Ha, see you also got the warnings. Was going to take care of those later today.  Any suggestion on how to handle that QImage instantiation warning involving imgData on PyQt6?

examples/utils.py
```diff
@@ -118,3 +117,8 @@
     ('LogPlotTest', 'LogPlotTest.py'),  # Plotting.py
     ('ViewLimits', 'ViewLimits.py'),    # ViewBoxFeatures.py
 ])
+
+# examples that are not suitable for CI testing
+skiptest = dict([
+    ('ProgressDialog', 'ProgressDialog.py'),    # modal dialog
```
j9ac9k, line 123: Why don't we skip when `os.getenv("CI")`, since that variable is `True` on CI platforms, that way we can still run this test locally.

pijyoi:
>Any suggestion on how to handle that QImage instantiation warning involving imgData on PyQt6?

PyQt6 QImage throws an exception if you pass it an integer as input. So the warning is weird. Changelog for PyQt6 suggests that QImage constructor will be more compatible with PyQt5 in future. So maybe this will have to wait for a newer PyQt6. The question then is whether we are going to remain compatible with the current PyQt6 6.0.

examples/utils.py
pijyoi: I am not so sure about this. ProgressDialog.py is different from the rest in that it doesn't run the event loop via app.exec_().
Meanwhile, test_examples.py assumes that an event loop needs to run and does it by calling processEvents.

j9ac9k:
> > Any suggestion on how to handle that QImage instantiation warning involving imgData on PyQt6?
> 
> PyQt6 QImage throws an exception if you pass it an integer as input. So the warning is weird. Changelog for PyQt6 suggests that QImage constructor will be more compatible with PyQt5 in future. So maybe this will have to wait for a newer PyQt6. The question then is whether we are going to remain compatible with the current PyQt6 6.0.

K, yeah let's not do anything about that warning; ...I would say might be beneficial to try and catch/filter that warning, but I'm wondering if that's more trouble than it's worth, either way, that's outside the scope of this PR.

examples/utils.py
j9ac9k: Ahhh!  That is a very relevant distinction; thanks for catching it.

j9ac9k:
PR LGTM, thanks @pijyoi merging.

pijyoi:
> > > Any suggestion on how to handle that QImage instantiation warning involving imgData on PyQt6?
> > 
> > 
> > PyQt6 QImage throws an exception if you pass it an integer as input. So the warning is weird. Changelog for PyQt6 suggests that QImage constructor will be more compatible with PyQt5 in future. So maybe this will have to wait for a newer PyQt6. The question then is whether we are going to remain compatible with the current PyQt6 6.0.
> 
> K, yeah let's not do anything about that warning; ...I would say might be beneficial to try and catch/filter that warning, but I'm wondering if that's more trouble than it's worth, either way, that's outside the scope of this PR.

#1550 is a consequence of the change mentioned above.

PR #1548, opened 2021-02-06
Use relative imports in newly added CuPy code

NilsNemitz:
I have renamed the topmost pyqtgraph folder in my local development environment. That way, it is more obvious to myself where I am importing the development code (`import dev_pyqtgraph.pyqtgraph as pg`) and where I am using the production library (`import pyqtgraph`).

Independent of whether this is a good idea or not, this used to work with no problems. It was broken recently by the addition of cupy code, which refers to the top folder / module name explicitly in two places.

I propose changing the code to use the same "relative" import used everywhere else.


j9ac9k:
Bleh, I should have caught this when reviewing that PR.

j9ac9k:
LGTM! Thanks for the fix, merging.

Issue #1537, opened 2021-02-02
Pyqtgraph still uses Qt's inaccuarate transform inversions in some places


markotoplak:
As last evident in #1493, pyqtgraph cares greatly about inverse transform accuracy. But they are still places within it where Qt's transforms are used, like in `ViewBox.updateScaleBox`: there, pyqtgraph calls `mapRectFromParent`, which goes directly to Qt. In effect, `updateScaleBox` does not work properly if axes are zoomed a lot (to 1e15 and more) or data is very spread. 

In Orange, we worked around that with biolab/orange3#3571, but our solution is neither pretty nor general. I am writing this as a note because I do not know pyqtgraph well enough to make a general pull request.

In pyqtgraph's code there are at least 6 more places where `mapRectFromParent` is used and they might be similarly problematic. Also, I would guess there are probably other Qt method calls in pyqtgraph that avoid pyqtgraph's accurate transforms.

j9ac9k:
Looking at that source code, `.inverted()` is used all over the place

Just from that file, it's used in other suspicious places...

* `QGraphicsItemGroup.addToGroup`
* `QGraphicsItemGroup.removeFromGroup`
* `QGraphicsItem.boundingRegion`
* `QGraphicsItem.mapRectFromParent`
* `QGraphicsItem.mapRectFromScene`
* `QGraphicsItem.mapFromParent`
* `QGraphicsItem.mapFromScene`

... There were some circumstances where `.inverted()` was just fine to use... some of the code paths did not involve `qFuzzyIsNull` but I would have to review the documentation for those again.

If I was to create a PR with a fix within the library, would you be able to test it out and see if it would work within Orange (which I know that would mean having to undo that PR)?

j9ac9k:
In #1557 I did my best attempt to implement `mapRectToParent` while avoiding the use of `QTransform.inverted()` and while our test suite passes, I don't have a good example of where the original implementation misbehaves.  @markotoplak if you happen to have either an example of when `mapRectToParent` causes an issue, or if you could test out my branch, that would be appreciated.

PR #1556, opened 2021-02-07
Reduce reallocation in dynamic range limiter

NilsNemitz:
This is a follow-up to PR #1140, which added a work-around to plots disappearing at very large zoom level, where the view is <1E-6 of the vertical span of data. The code introduced there is bypassed at any _normal_ zoom level and thus has no impact on performance for typical uses.

However, when the code _is_ triggered, it currently creates a full copy of the dataset every time the view is changed in any way.
This PR reduces the number of re-allocations by allowing a hysteresis in the clipping limits:
Data sets are now considered suitable to plot when they are clipped at planes that are located between 300 000 and 3 000 000 units above and below the current view, where one unit is the current view height. This means that the view can now be zoomed in or out by a factor of 3 before the previous clipping operation is no longer considered sufficient. Only a very large amount of vertical scroll requires a re-clipping operation, and horizontal scroll never does.

As an additional minor improvement, re-clipping operations that reduce the displayed range of data (zooming in) are now done in-place on the previously cached `yDisp` dataset, and full re-allocation only occurs when data needs to be restored from the original `yData` set (while zooming out).

I hope the efforts to eliminate inaccurate transforms and fuzzy zero checks will eventually fix the root cause of the disappearing plots, but so far, the bug still seems to be present. I have attached a test that generates data to trigger it: Unzip and run the attached [plot_test_5.zip](https://github.com/pyqtgraph/pyqtgraph/files/5938303/plot_test_5.zip)
Three curves should be visible. Zoom in vertically, and the red and magenta curves disappear. It would be great to hear back from someone who can test this for Qt6, since I still do not understand which part of the rendering pipeline seems to "optimize away" the plots.

Changing `disable`  to `False` in line 27 enables the dynamic range limiting code in the test, and the plots should no longer disappear.

Unfortunately, such data with small normal variations and occasional large glitches is my actual use case.


j9ac9k:
I'm tagging @markotoplak as he opened #1537 talking about this very issue.

j9ac9k:
Also can say that this issue is still persistent in PyQt6.

j9ac9k:
@NilsNemitz this LGTM, thanks so much for this PR!  Merging.

PR #1557, opened 2021-02-07
Avoid More Users of QTransform.inverted()

j9ac9k:
This PR aims to implement `GraphicsItem` mapping methods that avoid the use of `QTransform.inverted()` and use `pyqtgraph.functions.invertQTransform`.

I have implemented `mapRectFromParent` as was requested in #1537 _but_ I make no claim to how well it actually works or it solving any issues.  Using the test data provided in #1556 (thanks @NilsNemitz ) the plot still disappears on me.

I put the relevant Qt methods commented out for someone to verify that I'm reading the C code correctly (this is likely _not_ the case).

I'm going to go about implementing some of the other mapping methods.  Those of you that have an interest in this feature, I encourage you to test out this PR.

Tagging @markotoplak and @NilsNemitz

### DIFF
--- a/pyqtgraph/graphicsItems/GraphicsItem.py
+++ b/pyqtgraph/graphicsItems/GraphicsItem.py
@@ -8,7 +8,6 @@ import weakref
 import operator
 from ..util.lru_cache import LRUCache
 
-
 class GraphicsItem(object):
     """
     **Bases:** :class:`object`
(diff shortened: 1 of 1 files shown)


goodboy:
So am I correct in assuming that the concerns that were brought up in #1418 that resulted in #1493, the "precision error cases" that we were trying to avoid, are already possibly being triggered in the code base?

j9ac9k:
> So am I correct in assuming that the concerns that were brought up in #1418 that resulted in #1493, the "precision error cases" that we were trying to avoid, are already possibly being triggered in the code base?

Absolutely, I suspect this ultimately the cause that was the issue that #1140 aimed to address.

Not all calls to `QTransform.inverted()` results in precision errors (in-fact, most don't!); for some kind of transforms, `qFuzzyIsNull` is not utilized at all (see discussion in #1418: https://github.com/pyqtgraph/pyqtgraph/pull/1418#issuecomment-755074380).  It's quite possible that many other methods calls to the various Qt bindings do not involve using a codepath through `QTransform.inverted()` that results in these precision errors.

pijyoi (won't tag him to not spam his notifications) noticed that the performance hit in checking what kind of transformation (scale, translate, etc...) to determine if we would run into the rounding issues was actually larger than just implementing `QTransform.adjacent() * 1 / QTransform.determinant()`, so we elected to just call those blindly-in place (and as we were doing that, if we still had a non-invertable matrix, we compute the pseudo-inverse).

Determining which code-paths run through `QTransform.inverted()` would be a pretty tall order, but I suspect most mapping methods involving parent (`.mapToParent`, `.mapFromParent`, etc...) would call `QTransform.inverted()` in the compiled library.  Even bigger is determining which calls to those mapping functions result in precision errors.

So in the meantime we have to take it on a case-by-case method, and evaluate the feasibility of using polymorphism to supersede the method, or throwing up a big warning sign and trying to navigate around it based on observed issues.

j9ac9k:
Lastly, I want to say this PR sort of terrifies me, so I'm not going to even consider merging this until I am _quite_ convinced that not only does this PR fix the initial issue, but that there are no unintended side-effects (I suspect my method is slower, but I don't see it being slow to the point of being noticeable outside of a profiler)

j9ac9k:
Given that I have yet to run into a use-case where this PR fixed a known issue, combined with some high risk changes and likely performance hits, I'm going to close this PR out.

PR #1558, opened 2021-02-07
Fix Small Heights in ErrorBarItem

### LINKED ISSUE #1149: ErrorBarItem does not plot for values smaller than 1e-12
### Short description
When I try to plot with error bars having values smaller than 1e-12, they do not show up. No error or warning.

### Code to reproduce
```python
import pyqtgraph, numpy

ey = numpy.array([2e-12, 1.1e-12, 1e-12, 0.9e-12])
p = pyqtgraph.plot()
p.addItem(pyqtgraph.ErrorBarItem(x=[0,1,2,3], y=[0,0,0,0], top=ey))
```

### Expected behavior
We should see 4 bars of diminishing height.

### Real behavior
We see only the first two.

### Tested environment(s)

 * PyQtGraph version: 0.10.0
 * Python version: 3.7
 * NumPy version: 1.15.4
 * Operating system: Linux Mint 19
 * Installation method: conda

### Additional note
PlotCurveItems appear with crazy small values, well below 1e-12, but their error bars don't. This confused me for a long time!

Thanks in advance, and sorry if this issue is fixed in 0.11 already.

j9ac9k:
Fixes #1149

I'm implementing a fix that @ixjlyons observed seemed to work when displaying values smaller than 1e-12; he noticed that when using `functions.arrayToQPath` instead of `QPainterPath.moveTo` and `QPainterPath.lineTo` the problem seemed to go away.  

So somehow this gets us around https://bugreports.qt.io/browse/QTBUG-75630

Leaving in draft pull request form for right now as there are still some `QPainterPath.moveTo` / `QPainterPath.lineTo` calls in the paint method that I want to do away with.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -662,6 +662,35 @@ def affineSlice(data, shape, origin, vectors, axes, order=1, returnCoords=False,
         return output
 
 
+def interweaveArrays(*args):
+    """
+    Parameters
+    ----------
+
+    args : numpy.ndarray
+           series of 1D numpy arrays of the same length and dtype
+    
+    Returns
+    -------
+    numpy.ndarray
+        A numpy array with all the input numpy arrays interwoven
+
+    Examples
+    --------
+
+    >>> result = interweaveArrays(numpy.ndarray([0, 2, 4]), numpy.ndarray([1, 3, 5]))
+    >>> result
+    array([0, 1, 2, 3, 4, 5])
+    """
+
+    size = sum(x.size for x in args)
+    result = np.empty((size,), dtype=args[0].dtype)
+    n = len(args)
+    for index, array in enumerate(args):
+        result[index::n] = array
+    return result
+
+
 def interpolateArray(data, x, default=0.0, order=1):
     """
     N-dimensional interpolation similar to scipy.ndimage.map_coordinates.
--- a/pyqtgraph/graphicsItems/ErrorBarItem.py
+++ b/pyqtgraph/graphicsItems/ErrorBarItem.py
@@ -67,9 +67,6 @@ class ErrorBarItem(GraphicsObject):
         
         beam = self.opts['beam']
         
-        ys = np.column_stack((y, y)).flatten()
-        xs = np.column_stack((x, x)).flatten()
-        
         height, top, bottom = self.opts['height'], self.opts['top'], self.opts['bottom']
         if height is not None or top is not None or bottom is not None:
             ## draw vertical error bars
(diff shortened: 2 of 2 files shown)


j9ac9k:
@ixjlyons I think this guy is ready for review; given you did the initial testing in the issue, figure I'd nominate you to review 😆

j9ac9k:
blerg, accidentally pushed an unrelated commit, I'll try and remove...

j9ac9k:
my goodness... If I could actually use the right variable names, not have my code diffs between branches leak, that would be great.  @ixjlyons I think I'm ready for you.  One thing I would like input from is if there is a better way to split the numpy arrays together besides `splicedArray = np.column_stack((x1, x2)).flatten()`

pyqtgraph/graphicsItems/ErrorBarItem.py
```diff
@@ -82,23 +85,27 @@ def drawPath(self):
                     y2 = y
                 else:
                     y2 = y + top
+
+            y1_y2 = np.column_stack((y1, y2)).flatten()
+            y2s = np.column_stack((y2, y2)).flatten()
```
ixjlyons, line 90: nit: move these (`y1s`, `y2s`, `x1s`, `x2s`) to just where they're needed to avoid creating them if there aren't beams

ixjlyons (changes requested):
I've thought about this "interleave" operation too. I would guess the following is about as good as you can get, but I haven't checked:

```python
out = np.empty(x1.size + x2.size, dtype=x1.dtype)
out[::2] = x1
out[1::2] = x2
```

It might make sense to throw something like that in a function or static method given it's use in a few different spots (also if the above is faster it's no longer a one-liner unfortunately).

Other than that this looks great to me. Kinda nice just directly replacing the `moveTo`/`lineTo` calls so I don't really need to work through all the logic to review :D

j9ac9k:
Yeah renaming the path components to which component they were is voice too.  Good call on making a method elsewhere to handle the splicing.  I'll do some benchmarking

j9ac9k:
Updated the PR with a generic interweaving function; also now create the spliced arrays only when they're actually needed.

pyqtgraph/functions.py
```diff
@@ -662,6 +662,35 @@ def affineSlice(data, shape, origin, vectors, axes, order=1, returnCoords=False,
         return output
 
 
+def interweaveArrays(*args):
```
ixjlyons, line 665: Nice, I was considering suggesting it be in `functions` and generalized like this.

ixjlyons (approved):

pyqtgraph/functions.py
j9ac9k: There was no reason to have that kind of functionality specific to that ErrorBarItems...figured may as well put it in `functions.py`

PR #1559, opened 2021-02-08
restrict PyQt6 QEvent.type() monkey patching to 6.0.0

pijyoi:
The bug that this workaround was for has been fixed in PyQt6 6.0.1.

However our feature test for the bug was faulty and triggers positive
for PyQt6 6.0.1. In other words, we would be needlessly monkey patching
where it wasn't needed.

Moving forward, do we maintain workarounds for older patch level versions of PyQt6? We already have the current workarounds, so it's a matter of restricting it to particular versions. However, if new issues surface with 6.0.0, wouldn't we just tell users to use the newer version rather than implement another workaround?

### DIFF
--- a/pyqtgraph/Qt.py
+++ b/pyqtgraph/Qt.py
@@ -387,10 +387,7 @@ if QT_LIB == PYQT6:
 
     # PyQt6 6.0.0 has a bug where it can't handle certain Type values returned
     # by the Qt library.
-    try:
-        # 213 is a known failing value
-        QtCore.QEvent.Type(213)
-    except ValueError:
+    if QtCore.PYQT_VERSION == 0x60000:
         def new_method(self, old_method=QtCore.QEvent.type):
             try:
                 typ = old_method(self)
--- a/pyqtgraph/widgets/SpinBox.py
+++ b/pyqtgraph/widgets/SpinBox.py
@@ -113,12 +113,6 @@ class SpinBox(QtGui.QAbstractSpinBox):
         
         self.editingFinished.connect(self.editingFinishedEvent)
 
-    def event(self, ev):
-        ret = super().event(ev)
-        if ev.type() == QtCore.QEvent.KeyPress and ev.key() == QtCore.Qt.Key_Return:
-            ret = True  ## For some reason, spinbox pretends to ignore return key press
-        return ret
-        
     def setOpts(self, **opts):
         """Set options affecting the behavior of the SpinBox.
         


j9ac9k:
>Moving forward, do we maintain workarounds for older patch level versions of PyQt6? We already have the current workarounds, so it's a matter of restricting it to particular versions. However, if new issues surface with 6.0.0, wouldn't we just tell users to use the newer version rather than implement another workaround?

I've been doing some thinking about this, and I can convince myself of both positions; however I ultimately think we should _not_ try and preserve fixes for 6.0.0 for a few reasons.

1. We would be on the hook to implement fixes for 6.0.0 long after newer versions have been released.
2. Qt6 is brand new, I don't think it's realistic to support each layer of a bleeding edge ... I don't think we can create support for 6.0.0 after 6.0.1 was released the same way we we would treat 5.15.2 support after 5.15.3 was released.

Since we've opened a line of discussion with some downstream users, I'll ping them for their input:

Pinging @markotoplak and @outofculture for their thoughts.

pijyoi:
By the way, one of two (identified) places in pyqtgraph itself that were the reason for _this_ workaround is:
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/widgets/SpinBox.py#L116-L120
which itself is implementing a workaround for some old unspecified version of Qt! (Code was last modified in 2012)

(I did try removing the code before and it seems that current Qt versions don't need _that_ workaround)

j9ac9k:
If you can have a test that fails I'll ring it via tox and check most of the supported versions.

pijyoi:
I can only come up with a test that succeeds for all current bindings even after commenting out the event() override in
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/widgets/SpinBox.py#L116-L120

```python
import pyqtgraph as pg
from pyqtgraph.Qt import QtCore, QtTest

cnt = 0
def callback():
    global cnt
    cnt += 1

app = pg.mkQApp()
sb = pg.SpinBox()
sb.editingFinished.connect(callback)
QtTest.QTest.keyClick(sb, QtCore.Qt.Key.Key_Return)
sb.editingFinished.disconnect()
assert cnt == 1
```

j9ac9k:
> ```python
> import pyqtgraph as pg
> from pyqtgraph.Qt import QtCore, QtTest
> 
> cnt = 0
> def callback():
>     global cnt
>     cnt += 1
> 
> app = pg.mkQApp()
> sb = pg.SpinBox()
> sb.editingFinished.connect(callback)
> QtTest.QTest.keyClick(sb, QtCore.Qt.Key.Key_Return)
> sb.editingFinished.disconnect()
> assert cnt == 1
> ```

I'll start testing this on pyqt5/pyside2 5.12+ variants

j9ac9k:
Just ran against early versions of 5.12 variants... this works fine; I did a search on the qt bugtracker and saw some bug relating to key events, but not for the enter key, and it was fixed in 5.12 alpha. 

I'm good with getting rid of those event methods

outofculture:
> > Moving forward, do we maintain workarounds for older patch level versions of PyQt6?
> Since we've opened a line of discussion with some downstream users, I'll ping them for their input

Goodness, no, not on my account. Upgrading patch versions is painless. I will happily do that all day if it means any added simplicity or ease for pyqtgraph.

pijyoi:
Just to document down in this thread, the other (identified) place where this bug gets triggered in 6.0.0 is ```widgetEventFilter()``` of:
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/parametertree/parameterTypes.py#L181

j9ac9k:
@pijyoi I haven't heard from the `orange3` folks, but as they don't support Qt6 yet, I think them trying to support PyQt6 6.0.0 is unlikely if 6.0.1 is released...

You planning on other changes on this PR, or is it ready to merge?

pijyoi:
Ready to merge.

j9ac9k:
Thanks again @pijyoi !

PR #1560, opened 2021-02-08
Cache scatter-plot items by hashable properties

j9ac9k:
This fixes an issue that the great folks at ephyviewer encountered:  https://github.com/NeuralEnsemble/ephyviewer/issues/132

In an effort to speed up the `ScatterPlotItem`, we attempted to cache things like the Pen, Brush, and the Symbol.  Turns out, caching by `id(object)` wasn't the greatest, as those id numbers can be reused; and you get an interesting effect that was documented in the linked issue.

I want to thank @jpgill86 for helping me reproduce the issue and for @lidstrom83 for pointing me in the right direction for where to look.

### DIFF
--- a/pyqtgraph/graphicsItems/ScatterPlotItem.py
+++ b/pyqtgraph/graphicsItems/ScatterPlotItem.py
@@ -224,7 +224,7 @@ class SymbolAtlas(object):
                     squareness=1.0 if n == 0 else 2 * w * h / (w**2 + h**2))
 
     def _keys(self, styles):
-        return [(id(symbol), size, id(pen), id(brush)) for symbol, size, pen, brush in styles]
+        return [(symbol, size, (pen.style(), pen.capStyle(), pen.joinStyle()), (brush.color().rgba(), brush.style())) for symbol, size, pen, brush in styles]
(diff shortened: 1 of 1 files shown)


jpgill86:
Thanks @j9ac9k for this fix! I tested it out and it seems to do the trick!

j9ac9k:
As evidence by our CI this is not quite right, also there may be performance issues here at play.

Appreciate you testing this out!

j9ac9k:
This caching is problematic for this line alone :|

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ScatterPlotItem.py#L100

When adding a symbol, the caching mechanism uses the name (`str`, which is hashable), but when passing in a `QPainterPath`, that's a bit trickier ...I'm trying to come up with a string-representation of QPainterPath, or some way to hash it, but nothing in the docs is sticking out.

j9ac9k:
As I've heard from the person that reported this issue; and our test suite passes, I'm going to go ahead and merge this... Thank you so much @jpgill86 for testing out this PR on your application.

PR #1561, opened 2021-02-09
remove app.deleteLater() from tests

pijyoi:
Multiple tests are executed by pytest with the same Python process. Deleting QApplication is likely to cause a crash.

Refer to comments in #1396


j9ac9k:
Thanks @pijyoi !

PR #1563, opened 2021-02-09
Flowchart documentation fix

RubendeBruin:
the widget() method should be used to display the flowchart in a window rather than the ctrlWidget().


ixjlyons (approved):

j9ac9k:
Thanks for the PR @RubendeBruin !

PR #1564, opened 2021-02-09
Correct id-based keying of scatter plot pixmap cache

lidstrom83:
Note: naively using the id function results in non-unique keys.

Alternatively, we could serialize the Qt objects and use these in the key. This would provide protection from the user mutating these later, but at a cost to performance.

### DIFF
--- a/pyqtgraph/graphicsItems/ScatterPlotItem.py
+++ b/pyqtgraph/graphicsItems/ScatterPlotItem.py
@@ -231,7 +231,7 @@ class SymbolAtlas(object):
             try:
                 return obj._id
             except AttributeError:
-                obj._id = next(self._idGenerator)
+                obj._id = next(SymbolAtlas._idGenerator)
                 return obj._id
 
         return [
--- a/pyqtgraph/graphicsItems/ScatterPlotItem.py
+++ b/pyqtgraph/graphicsItems/ScatterPlotItem.py
@@ -5,6 +5,7 @@ try:
     from itertools import imap
 except ImportError:
     imap = map
+import itertools
 import numpy as np
 import weakref
 from ..Qt import QtGui, QtCore, QT_LIB
@@ -161,6 +162,8 @@ class SymbolAtlas(object):
         pm = atlas.pixmap
 
     """
+    _idGenerator = itertools.count()
+
     def __init__(self):
         self._data = np.zeros((0, 0, 4), dtype=np.ubyte)  # numpy array of atlas image
         self._coords = {}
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/ScatterPlotItem.py
```diff
@@ -224,13 +227,16 @@ def diagnostics(self):
                     squareness=1.0 if n == 0 else 2 * w * h / (w**2 + h**2))
 
     def _keys(self, styles):
+        def getId(obj):
+            try:
+                return obj._id
+            except AttributeError:
+                obj._id = next(self._idGenerator)
```
j9ac9k, line 234: should this be `next(SymbolAtlas._idGenerator)` as it is a class variable and not an instance variable?

lidstrom83: Yes, thanks

j9ac9k:
hey @lidstrom83 good to see you back here.  Thanks for this update.   @danielhrisca would you mind benchmarking this branch in your tool?  I'd like to see how it compares.

lidstrom83:
Hey @j9ac9k, do you have any preference for the performance vs. safety trade-off between the id vs. serialization cache key approaches?

j9ac9k:
> Hey @j9ac9k, do you have any preference for the performance vs. safety trade-off between the id vs. serialization cache key approaches?

I do not, primarily because I don't have a good way to profile so it's hard to evaluate and make a fair comparison.

In terms of specific to this PR, my standard is simple, does it improve on the current implementation?  yes (and it works); then merge.  I don't want to hold up changes on ScatterPlotItem based on what the ideal implementation would be.

danielhrisca:
pyqt|fps bench| asammdf plot| asammdf plot 5x zoom in
--|--|-- | --
master|195 fps | 16.2s | 37.34s
lindstrom83:scatterplot|250 fps| 8.8s | 19.06s

j9ac9k:
That's absolutely great!

j9ac9k:
@lidstrom83 you planning any other cahnges on this PR or is this guy ready to merge?

danielhrisca:
> That's absolutely great!
> 
> EDIT just to verify, this is an improvement over 0.11.0 as well?

pyqt|fps bench| asammdf plot| asammdf plot 5x zoom in
--|--|-- | --
master|195 fps | 16.2s | 37.34s
lindstrom83:scatterplot|250 fps| 8.8s | 19.06s
0.11.0| 235 fps| 5.11s |13.7s

danielhrisca:
`master` and this PR were only tested with `_USE_QRECT=False`

lidstrom83:
> @lidstrom83 you planning any other changes on this PR or is this guy ready to merge?

No, this PR is done now. I'll dig further to get to the bottom of the rest of @danielhrisca's slowdown.

j9ac9k:
Sounds good, Merging.  Thanks @lidstrom83   @danielhrisca thanks for the benchmark info too!

PR #1566, opened 2021-02-10
Add more options to ScatterPlotSpeedTest

lidstrom83:


j9ac9k:
Wow @lidstrom83 this is amazing....

Only thing I suggest is using `translate` functionality, our non-English speaking colleagues would appreciate that; luckily it's straight forward to add.

Just after the inports

```python
translate = QtCore.QCoreApplication.translate
```

Then for each parameter that is actually displayed

Instead of 

```python
dict(name='_USE_QRECT', title='Use QRect:    ', type='bool',  ...)
```

Do

```python
dict(name='_USE_QRECT', title=translate('ScatterPlot', 'Use QRect:    ') , type='bool', ...)
```

Basically, for any text that is user-facing like this, we want to wrap it with `translate($context, "text to show")`

j9ac9k:
This LGTM, thank you so much @lidstrom83  this is amazing.

PR #1567, opened 2021-02-10
Handle case of 0-d array in infinite line setPos

ksunden:
Fixes a small regression found in my app introduced by #1310.

Basically a 0-d array should be treated as a scalar, it works downstream of this check, but since the check added ndarray it got caught trying to make a list, which is not allowed for 0-d.

np.ndim works for any array like, which will be what is passed, since the isinstance check runs first. (and indeed even if it was run on a QPointF instance, it just returns 0).

this will also make it work with subclasses, which the direct type check will not


j9ac9k:
thanks @ksunden; this LGTM, merging.

Issue #1550, opened 2021-02-06
Image Comparisons Fail on PyQt6 6.0.1


j9ac9k:
PyQt6 6.0.1 was released a few hours ago, as our CI grabs the latest versions of PyQt6, it grabbed 6.0.1, and now we're having problems w/ our `assertImageApproved` checks.  I am able to replicate the failures locally.

Looking at the changelog for PyQt6, there is nothing here to suggest that there should be an issue here:

https://riverbankcomputing.com/news/PyQt_v6.0.1_Released

```
FAILED pyqtgraph/graphicsItems/tests/test_ImageItem.py::test_ImageItem - AssertionError
FAILED pyqtgraph/graphicsItems/tests/test_ImageItem.py::test_ImageItem_axisorder - AssertionError
FAILED pyqtgraph/graphicsItems/tests/test_ROI.py::test_getArrayRegion - AssertionError
FAILED pyqtgraph/graphicsItems/tests/test_NonUniformImage.py::test_NonUniformImage_lut - AssertionError
FAILED pyqtgraph/graphicsItems/tests/test_ROI.py::test_getArrayRegion_axisorder - AssertionError
FAILED pyqtgraph/graphicsItems/tests/test_NonUniformImage.py::test_NonUniformImage_colormap - AssertionError
FAILED pyqtgraph/graphicsItems/tests/test_ROI.py::test_PolyLineROI - AssertionError
```

EDIT: fair warning to anyone attempting to upgrade their python environment from PyQt6 6.0.0 to 6.0.1, doing `pip install --upgrade PyQt6` is insufficient, you would also need to install `PyQt6-Qt`

pijyoi:
will be fixed by #1553

ksunden:
Closing as 1553 is merged

PR #1568, opened 2021-02-11
Change image tests to not rely on non-const QImage

pijyoi:
During an attempt at refactoring in #1223, it was found that PyQt bindings would fail the CI image comparison tests.
The CI image comparison tests also failed when PyQt6 was updated to 6.0.1, as reported in #1550.

Note that the failures were not due to the images being different but due to ```image_testing.py::assertImageApproved``` relying on QImage being non-const. This PR removes this unnecessary assumption.

The changes to ```functions.py``` is an unrelated change to allow images larger than 2GB for PyQt5 5.15.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1358,21 +1358,23 @@ def imageToArray(img, copy=False, transpose=True):
     The array will have shape (width, height, (b,g,r,a)).
     """
     fmt = img.format()
-    ptr = img.bits()
-    if QT_LIB in ['PySide', 'PySide2', 'PySide6']:
-        arr = np.frombuffer(ptr, dtype=np.ubyte)
-    else:
+    img_ptr = img.bits()
+
+    if QT_LIB.startswith('PyQt'):
+        # sizeInBytes() was introduced in Qt 5.10
+        # however PyQt5 5.12 will fail with:
+        #   "TypeError: QImage.sizeInBytes() is a private method"
+        # note that sizeInBytes() works fine with:
+        #   PyQt5 5.15, PySide2 5.12, PySide2 5.15
         try:
-            # removed in Qt6
-            nbytes = img.byteCount()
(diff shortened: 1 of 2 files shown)


j9ac9k:
Thanks @pijyoi this LGTM.  I'll merge momentarily

Unrelated, we're very close to releasing 0.12.0; given you've been making so many contributions lately, would you like us to hold off for any other particular change?  The issue we're waiting to fix is 1492.

pijyoi:
>would you like us to hold off for any other particular change?

I think a remaining issue is whether we need to restore the scale, rotate, translate shims for Qt5 bindings.

j9ac9k:
Oh, right, ... yes, probably should do that, and slap a deprecation warning.  I can handle that.

pijyoi:
@j9ac9k before you release 0.12.0, do you think it would be a good idea to reduce the shim surface in PyQt6 to avoid users relying on pyqtgraph's shims?
Candidates are those shims used in one place only in the code base

j9ac9k:
> @j9ac9k before you release 0.12.0, do you think it would be a good idea to reduce the shim surface in PyQt6 to avoid users relying on pyqtgraph's shims?
> 
> Candidates are those shims used in one place only in the code base

I can be convinced both ways on this. On one hand pyqtgraph doesn't aim to be a Qt abstraction library like QtPy or AnyQt.  On the flip side, how much more effort is it for us to expose the work we've done to end users and improve their QoL, especially as a frequent debug step we ask people to create issues is to try a different set of bindings.

@outofculture @ixjlyons @lidstrom83  you all have opinions here?

@pijyoi if you think other parts of the library should be deprecated, let us know.  I just attached deprecation warnings to portions of the libraries that have had comments/print statements, but we are not limited to deprecating just those portions.

pijyoi:
created #1575 for the shims that could be removed

PR #1569, opened 2021-02-11
Fix ScatterPlotItem performance regression

lidstrom83:


j9ac9k:
Hate to go back on a feature, but that transparent background is a _huge_ hit to performance; the benefit doesn't come close to the justifying the hit to performance though.

Sorry @dgoeries ; if you can think of a way to implement the intended functionality w/o the performance hit (or a very minor performance hit), then please feel free to re-submit that feature, in the meantime, I think we'll have to have transparent backgrounds be enabled by end users if they want to do it themselves, but we can't have it as a default.

janezd:
A note on transparent background (fd0e2b65afefdda2ddd08f9ee51a3914c0616b38): it didn't hit just the performance. For some reason it caused a black background in legend (https://github.com/biolab/orange3/pull/5680#discussion_r753185929), though I assume our use of palette was probably correct. This is happening in pyqtgraph 0.11, while in 0.12 it's OK, so I traced the problem to this change.

PR #1570, opened 2021-02-11
Different Plot speed measure

dgoeries:
@j9ac9k i change a bit the measure and then it becomes obvious for me between 0.11.0 and masterrr


examples/PlotSpeedTest.py
```diff
@@ -5,53 +5,54 @@
...
+
     ptr += 1
-    now = time()
-    dt = now - lastTime
-    lastTime = now
-    if fps is None:
-        fps = 1.0/dt
-    else:
-        s = np.clip(dt*3., 0, 1)
-        fps = fps * (1-s) + (1.0/dt) * s
```
lidstrom83, line 44: Any reason for taking out the moving average? Suppressing older samples allows the frame rate estimate be more responsive when changing plot options.

examples/PlotSpeedTest.py
```diff
@@ -5,53 +5,54 @@
...
+    # Measure
+    t_start = perf_counter()
+    curve.setData(data[ptr % 10])
+    app.processEvents(QtCore.QEventLoop.AllEvents)
+
+    # Statistics
+    elapsed += perf_counter() - t_start
+    average = elapsed / ptr
+    fps = 1 / average
+    p.setTitle('%0.2f fps - %0.5fs avg' % (fps, average))
```
lidstrom83, line 46: Elapsed time between frames would look nicer displayed in milliseconds.

examples/PlotSpeedTest.py
```diff
@@ -5,53 +5,54 @@
 """
 
 ## Add path to library (just for examples; you do not need this)
```
lidstrom83, line 7: This comment is referring to `import initExample` and should be kept above that line.

examples/PlotSpeedTest.py
lidstrom83: Incidentally, I find a measurable difference when replacing `s = np.clip(dt*3., 0, 1)` with `s = min(1, max(0, dt * 3.))`

dgoeries: I will change back.

j9ac9k:
Got super-seeded by #1610 Thanks @dgoeries for starting work on this!

PR #1572, opened 2021-02-12
Arbitrary scale center ROI

outofculture:
This starts from #67, handles the merge conflicts, and adds an example.

### DIFF
--- a/examples/ROIExamples.py
+++ b/examples/ROIExamples.py
@@ -126,7 +126,9 @@ r3b.addRotateHandle([0, 1], [1, 0])
 
 ## handles rotating/scaling around center
 r3b.addScaleRotateHandle([0, 0.5], [0.5, 0.5])
-r3b.addScaleRotateHandle([1, 0.5], [0.5, 0.5])
+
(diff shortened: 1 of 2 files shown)


j9ac9k:
In CI, there is always that one test, one that one pipeline...

```
================================== FAILURES ===================================
_________________________ pyqtgraph/tests/test_qt.py __________________________
[gw1] win32 -- Python 3.7.9 c:\hostedtoolcache\windows\python\3.7.9\x64\python.exe
worker 'gw1' crashed while running 'pyqtgraph/tests/test_qt.py::test_loadUiType'
----------- generated xml file: D:\a\pyqtgraph\pyqtgraph\pytest.xml -----------
```

this test failure has nothing to do w/ this PR, feel free to disregard.

j9ac9k:
This LGTM; I don't use ROIs much at all, but maybe a good future addition would be to have some kind of indication of where the center of scaling/rotation is (like some kind of crosshair)?

I'll say that's outside of the scope of this PR, so I'm going to merge this.

Thanks @outofculture and @vallsv sorry it's taken us so long to get this change in.

PR #1574, opened 2021-02-13
Make hdf5 example more efficient

lidstrom83:
I'm mostly using this example for staging how to revisit downsampling in PlotDataItem. The general strategy we use here should apply there. This updates the example to
- cache downsampled data at scales of powers of two to minimize cache size and computation
- load data for the width of the view range plus padding to avoid excessive reloads
- eagerly downsample and caches data before it's needed

I'm submitting this in draft form to first solicit feedback on the approach taken.

### DIFF
--- a/examples/hdf5.py
+++ b/examples/hdf5.py
@@ -4,22 +4,15 @@ In this example we create a subclass of PlotCurveItem for displaying a very larg
 data set from an HDF5 file that does not fit in memory. 
 
 The basic approach is to override PlotCurveItem.viewRangeChanged such that it
-displays only the visible portion of the data after downsampling. The data is downsampled
-and the results are cached. Downsampling is done at scales of powers of two to reduce
-computation and cache size, and is displayed with padding at the boundaries so fewer plot
-updates are required. Downsampling and caching begins eagerly in a separate thread when
-the item is initialized so that the data is ready (or closer to ready) when
-the user needs it. This could alternatively be precomputed and stored in the HDF5 file.
-Cache sizes are limited, and a relatively small amount of cached data
-at the largest scales is required to fluidly navigate through large data sets.
+displays only the visible portion of downsampled data, plus some padding to reduce the frequency
+of plot updates. Further, a relatively small amount of downsampled data is precomputed and stored
+in the HDF5 file, which is sufficient for navigation of the entire data set without aliasing or lag.
 """
 
 import initExample  # Add path to library (just for examples; you do not need this)
 
 import os
 import sys
-import threading
-from functools import lru_cache
 
 import h5py
 import numpy as np
@@ -36,18 +29,18 @@ plt.setXRange(0, 500)
 
 class BigDataPlot(pg.PlotCurveItem):
     def __init__(self, *args, **kwds):
-        self.sampler = None
+        self._plotData = None
         pg.PlotCurveItem.__init__(self, *args, **kwds)
-        
-    def setDataSampler(self, sampler):
-        self.sampler = sampler
-        self.loadData()
+
+    def setPlotData(self, plotData):
+        self._plotData = plotData
+        self.replot()
 
     def viewRangeChanged(self):
-        self.loadData(lazy=True)
+        self.replot(lazy=True)
 
-    def loadData(self, lazy=False):
-        if self.sampler is None:
+    def replot(self, lazy=False):
+        if self._plotData is None:
             self.setData([])
             return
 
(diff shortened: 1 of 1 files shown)


examples/hdf5.py
```diff
@@ -4,99 +4,184 @@
...
+        if pad:
+            pad = (x2 - x1) * self.padding
+            x1, x2 = x1 - pad, x2 + pad
+        if self.x is None:
+            i1 = max(0, min(len(self), int(x1)))
+            i2 = max(0, min(len(self), int(x2)))
+        else:
+            i1, i2 = np.searchsorted(self.x, [x1, x2])
+        ds = max(1, int((i2 - i1) / self.sampleLimit))
+        n = int(math.log2(ds))
```
j9ac9k, line 131: `math.log2` ... so glad we migrated to python3 👍

examples/hdf5.py
```diff
@@ -4,99 +4,184 @@
...
+        ds = 2 ** n
+        i1 = int(i1 / ds - 1) * ds
+        i2 = int(i2 / ds + 1) * ds
+        i1 = max(0, min(len(self), int(i1)))
```
j9ac9k, line 135: no need to cast `i1` to `int()` it should be an int from the previous operations

examples/hdf5.py
```diff
@@ -4,99 +4,184 @@
...
+    def __init__(self, x, y, sampleLimit=2500, cacheLimit=10000000, padding=0.3, eager=True):
+        if x is not None and len(x) != len(y):
+            raise ValueError
+
+        self.x = x
+        self.y = y
+        self.sampleLimit = sampleLimit  # at most 4 times this number of samples will be plotted
+        self.cacheLimit = cacheLimit  # at most 8 times this number of samples will be cached
+        self.padding = padding
+        self._minCacheLevel = max(1, int(math.log2(max(1, int(len(self) / self.cacheLimit)))))
```
j9ac9k, line 75: when looking at `math.log2` of an int, I think you normally just want the `int.bit_length()`

lidstrom83:
Sorry for the sloppiness. The recent push should resolve these.

lidstrom83:
Hey @goodboy I recall you taking an interest in the downsampling API. Ideally what we use here is applicable to other parts of the library, so it'd be nice to get it right in this PR. Any suggestions?

examples/hdf5.py
```diff
@@ -4,99 +4,199 @@
...
+            level=n,
+            level_size=len(out),
+            cache_size=self.cacheSize,
+            data_size=len(self.y),
+            ratio=len(self.y) // self.cacheSize
+        ))
+        return out
+
+
+def downsample(data, start=0, stop=None, ds=4, chunksize=1000000):
```
goodboy, line 167: this looks good :thumbs

examples/hdf5.py
```diff
@@ -4,99 +4,199 @@
...
+        n = ds.bit_length()
+        ds = 2 ** n
+        i1 = (i1 // ds - 1) * ds
+        i2 = (i2 // ds + 1) * ds
+        i1 = max(0, min(len(self), i1))
+        i2 = max(0, min(len(self), i2))
+        return i1, i2, ds, n
+
+    @lru_cache(maxsize=None)
+    def _loadLevel(self, n):
```
goodboy, line 147: I don't really understand why this is done in a thread?

Is there something in here that releases the GIL?

examples/hdf5.py
```diff
@@ -4,99 +4,199 @@
...
+            ds=ds,
+            n=n,
+            size=i2-i1,
+            downsampled_size=len(y),
+            cached=n >= self._minCacheLevel
+        ))
+        x = i if self.x is None else self.x[i]
+        return x, y
+
+    def __len__(self):
```
goodboy, line 125: why is this a method.

examples/hdf5.py
```diff
@@ -4,99 +4,199 @@
...
+        self.cacheLimit = cacheLimit  # at most 8 times this number of samples will be cached
+        self.cacheSize = 0
+        self.padding = padding
+        self._minCacheLevel = max(2, (len(self) // self.cacheLimit).bit_length())  # n < 2 (ds < 4) has no effect
+        self._maxCacheLevel = (len(self) // self.sampleLimit).bit_length()
+        self._last = None
+        if eager and (self._maxCacheLevel >= self._minCacheLevel):
+            # spawn a thread to eagerly downsample data up to self._maxCacheLevel so
+            # it will be ready for when the user zooms out
+            self._t = threading.Thread(target=lambda: self._loadLevel(self._maxCacheLevel))
```
goodboy, line 81: why do you have this `lambda`?
It's a shared mem thread, just read the `._maxCacheLevel` from the target if you're trying to read it "lazily".

examples/hdf5.py
```diff
@@ -4,99 +4,199 @@
...
-        self.scale(scale, 1)  # scale to match downsampling
 
-        
+        x1, x2 = vb.viewRange()[0]
+        if not lazy or not self.sampler.isRangeValid(x1, x2):
+            x, y = self.sampler[x1:x2]
+            self.setData(x, y)  # update the plot
+
+
+class DataSampler:
```
goodboy, line 64: Ok so I think my main critique here is why does this need to be a class?
All of the "state" being passed at init here is basically used once as an input that you could provide to a lone function and/or bind to a closure that can be called multiple times once you've bound in your base settings.

I would prefer to see all of this done in a set of functions.
If you have a class that defines basically one function that you call (in this case it appears to be `.__getitem__()`), that's not a class.

This class seems to imply there's some kind of "delayed computation" that needs to take place (in a thread that you start right away??)

I don't get why there's all this setting and getting of variables on this object when you can just basically pass it all as input.  If you require delayed computation, use a closure.

examples/hdf5.py
```diff
@@ -4,99 +4,199 @@
...
+
+        self.x = x
+        self.y = y
+        self.sampleLimit = sampleLimit  # at most 4 times this number of samples will be plotted
+        self.cacheLimit = cacheLimit  # at most 8 times this number of samples will be cached
+        self.cacheSize = 0
+        self.padding = padding
+        self._minCacheLevel = max(2, (len(self) // self.cacheLimit).bit_length())  # n < 2 (ds < 4) has no effect
+        self._maxCacheLevel = (len(self) // self.sampleLimit).bit_length()
+        self._last = None
```
goodboy, line 77: Yeah like this "caching" is basically, "have i called `.isRangeValid()` yet?".

Well if you haven't called `.__getitem__()` yet then no 😉

examples/hdf5.py
```diff
@@ -4,99 +4,199 @@
...
-            
-        self.setData(visible) # update the plot
-        self.setPos(start, 0) # shift to match starting index
-        self.resetTransform()
-        self.scale(scale, 1)  # scale to match downsampling
 
-        
+        x1, x2 = vb.viewRange()[0]
+        if not lazy or not self.sampler.isRangeValid(x1, x2):
+            x, y = self.sampler[x1:x2]
```
goodboy, line 60: this is your "lone method call" afaict.
Imo `.sampler` does not need to be a class.

examples/hdf5.py
```diff
@@ -4,99 +4,199 @@
...
+
+
+class DataSampler:
+    def __init__(self, x, y, sampleLimit=2500, cacheLimit=10000000, padding=0.3, eager=True):
+        if x is not None and len(x) != len(y):
+            raise ValueError
+
+        self.x = x
+        self.y = y
+        self.sampleLimit = sampleLimit  # at most 4 times this number of samples will be plotted
```
goodboy, line 71: Do these settings need to be dynamically changed at runtime?

goodboy (commented):
@lidstrom83 thanks for the shout out 😂 

Hopefully I'm not overstepping here but personally I would do this a bit differently and make it a little more, "functional" style.

examples/hdf5.py
lidstrom83: So the user can explore the plot without it hanging while the cache is being filled.

examples/hdf5.py
lidstrom83: No good reason. It just evolved that way to be taking as an argument what eventually turned into an attribute.

examples/hdf5.py
goodboy: Ahh. Interesting. Still does it need to be a method?

It seems you could just pass in all the inputs here.

examples/hdf5.py
goodboy: Even still, it's not really lazy since you start the thread right away.

examples/hdf5.py
lidstrom83: No good reason for that one.

examples/hdf5.py
```diff
@@ -4,99 +4,199 @@
...
+            downsampled_size=len(y),
+            cached=n >= self._minCacheLevel
+        ))
+        x = i if self.x is None else self.x[i]
+        return x, y
+
+    def __len__(self):
+        return len(self.y)
+
+    def _range(self, x1, x2, pad=True):
```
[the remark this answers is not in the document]
lidstrom83: It's called in isRangeValid a couple times and in __getitem__

examples/hdf5.py
lidstrom83: It's also storing the cached data that gets accessed in subsequent calls to `__getitem__`. How would you suggest storing this instead?

goodboy: well if you want to cache calls to `__getitem__` why not just `lru_cache` it?

Which var is storing the cached data?
It seems the only thing actually being cached is outputs from calls to `._loadlevel()` but those aren't vars here they're implicitly stored inside `lru_cache`.

examples/hdf5.py
```diff
@@ -4,99 +4,199 @@
...
+
+class DataSampler:
+    def __init__(self, x, y, sampleLimit=2500, cacheLimit=10000000, padding=0.3, eager=True):
+        if x is not None and len(x) != len(y):
+            raise ValueError
+
+        self.x = x
+        self.y = y
+        self.sampleLimit = sampleLimit  # at most 4 times this number of samples will be plotted
+        self.cacheLimit = cacheLimit  # at most 8 times this number of samples will be cached
```
goodboy, line 72: another one. why is `.cacheLimit` a var here, it doesn't seem to be used elsewhere.

examples/hdf5.py
j9ac9k: We don't use it anywhere else but `QtConcurrent` and `QFuture` might be worthwhile here; although looks like the threading code needed here is pretty minimal.

examples/hdf5.py
lidstrom83: No, and if they are things would break. They should at least be private until that is fixed.

examples/hdf5.py
lidstrom83: There's `isRangeValid` and `__getitem__`.  In addition to the cached data, the class also stores the last view range so it can determine if it has been invalidated.

examples/hdf5.py
lidstrom83: No reason. Just a sloppy draft thing.

examples/hdf5.py
lidstrom83: > @lidstrom83 thanks for the shout out 
> 
> Hopefully I'm not overstepping here but personally I would do this a bit differently and make it a little more, "functional" style.

Not at all! I much prefer that as well, if possible.

lidstrom83:
> @lidstrom83 thanks for the shout out 
> 
> Hopefully I'm not overstepping here but personally I would do this a bit differently and make it a little more, "functional" style.

examples/hdf5.py
lidstrom83: > well if you want to cache calls to `__getitem__` why not just `lru_cache` it?

This would result in a cache entry for potentially every pair of x bounds. Caching by level limits this significantly.

examples/hdf5.py
lidstrom83: Oh, sorry, I didn't think long enough about your question before answering. `_loadLevel` needs to take the argument explicitly because it's recursively defined.

lidstrom83:
In the recent commit I pushed as much logic as I could into pure functions. AFAICT the remaining state is necessary, so the class remains, but in a much simpler form.

Also, the cached data is now handled explicitly so the user has more control. In the example, it's now pre-computed and stored in the HDF5 file. It could alternatively be computed each time the program is run, either in the main thread or in a separate one.

examples/hdf5.py
goodboy: I still don't grok.

- the bare `lambda` is odd / unrequired.
- you could still do this with a function and not a method 

Maybe you want a `functools.partial` or just use the `args` to [`threading.Thread`](https://docs.python.org/3/library/threading.html#threading.Thread)?

examples/hdf5.py
goodboy: So then cache all levels and keep a reference to the func in the parent thread.
Call the now lru cached func ref in the parent thread as needed.

examples/hdf5.py
goodboy: > In addition to the cached data, the class also stores the last view range so it can determine if it has been invalidated.

Yeah maybe it's this part I'm not understanding.

If you have a range that's been invalidated don't you have to call all the same methods again?

If so, why not just have the caller of this `cache_x_range()` func do the range checking and re-call the downsampling steps as needed from client code instead of assigning class vars that are only written once afaict and then having it all triggered by multiple `__getitem__()` calls?

examples/hdf5.py
goodboy: I think that this confirms my thoughts that this shouldn't be a class / object then.

lidstrom83: There is state that needs to live somewhere even if this particular variable doesn't. Please see the recent commit for the most concise representation of this state that I could come up with.

examples/hdf5.py
```diff
@@ -4,99 +4,177 @@
...
+
+def computeDownsampleCache(data, minLevelSize=2500, maxLevelSize=10000000):
+    minLevel = max(2, (len(data) // maxLevelSize).bit_length())
+    maxLevel = (len(data) // minLevelSize).bit_length()
+
+    out = {}
+    dat = None
+    for lv in range(minLevel, maxLevel + 1):
+        ds = 2 ** lv
+        if dat is None:
```
goodboy, line 172: Maybe I'm missing something here but couldn't you just call this once before the loop?

```python
levels = range(minLevel, maxLevel + 1)
lv = next(levels)
ds = 2 ** lv
out[ds] = _downsample(data=data, start=0, stop=len(data), ds=ds)

for lv in levels:
    out[2 ** lv] = _downsample(data=dat, start=0, stop=len(dat), ds=4)
       
return out
```

examples/hdf5.py
```diff
@@ -4,99 +4,177 @@
...
-            scale = 1
+
+        x1, x2 = vb.viewRange()[0]
+        data = self._plotData.sample(x1, x2, lazy=lazy)
+
+        if data is not None:
+            self.setData(*data)  # update the plot
+
+
+class PlotData:
```
goodboy, line 58: Ye nice work. 
This is looking a heck of a lot cleaner and I can more easily see how everything is composed now.

examples/hdf5.py
```diff
@@ -4,99 +4,177 @@
...
+        self._lastRange = None
+
+    def sample(self, x1, x2, lazy=True, padding=0.3, **kwargs):
+        i1, i2, ds = plotDataRange(self._x, x1, x2, padding=padding, **kwargs)
+
+        if lazy and self._lastRange is not None:
+            i11, i21, ds1 = self._lastRange
+            i12, i22, _ = plotDataRange(self._x, x1, x2, padding=0, **kwargs)
+            if (i11 <= i12 <= i22 <= i21) and (ds1 == ds):
+                return
```
goodboy, line 77: This bit is slightly weird: using `None` to indicate the delayed computation.

I wonder if you could do something a little nicer with this.
I need to reread everything first; might take me a bit 😂

examples/hdf5.py
```diff
@@ -4,99 +4,177 @@
...
-        # Decide by how much we should downsample 
-        ds = int((stop-start) / self.limit) + 1
-        
-        if ds == 1:
-            # Small enough to display with no intervention.
-            visible = self.hdf5[start:stop]
-            scale = 1
+
+        x1, x2 = vb.viewRange()[0]
+        data = self._plotData.sample(x1, x2, lazy=lazy)
```
goodboy, line 52: When is `lazy != False`? Just when you do the thread spawning?

examples/hdf5.py
```diff
@@ -4,99 +4,177 @@
...
+    def setPlotData(self, plotData):
+        self._plotData = plotData
+        self.replot()
+
     def viewRangeChanged(self):
-        self.updateHDF5Plot()
-        
-    def updateHDF5Plot(self):
-        if self.hdf5 is None:
+        self.replot(lazy=True)
```
goodboy, line 40: Ahh ok so here?

examples/hdf5.py
goodboy: So in the lazy case where nothing changes, there's no state set, you're just getting the last range and computing the new one yah?

I wonder does this need to be inside the class then?
I guess to encapsulate all the `i` value details.

I wonder could you not just check against the last `x1`, `x2` inputs instead of recomputing the matching `i` value outputs.
If the function is injective you shouldn't need to check the outputs for each new input, just compare the inputs.

examples/hdf5.py
```diff
@@ -4,99 +4,177 @@
...
+
+
+class PlotData:
+    def __init__(self, x, y, cache=None):
+        if x is None:
+            x = len(y)
+        elif len(x) != len(y):
+            raise ValueError
+
+        self._x = x
```
goodboy, line 65: doesn't seem like either of these are getting set anywhere besides here.
i'm guessing it's needed for compatibility inside `PlotCurveItem`?

goodboy (commented):
Another round 🏄🏼 

Looking a lot better, nice work 😎

Achilles1515:
This downsampling routine can throw an exception when operating on the full contents of a user-provided x-axis array.

Make the following change in the example code to reproduce the error:
```python
# Ignore the H5 file for the time being and just manually provide a data array 
y = np.linspace(0, 499, 12200000)    # 12200000 is "arbitrary" size where I noticed the issue
x = np.linspace(0, 499, len(y))
cache = None

curve = BigDataPlot()
curve.setPlotData(PlotData(x, y, cache=cache))
```

Leads to the following exception:
```
  File "C:\Program Files\Python39\lib\site-packages\pyqtgraph\graphicsItems\GraphicsObject.py", line 23, in itemChange
    self.parentChanged()
  File "C:\Program Files\Python39\lib\site-packages\pyqtgraph\graphicsItems\GraphicsItem.py", line 472, in parentChanged
    self._updateView()
  File "C:\Program Files\Python39\lib\site-packages\pyqtgraph\graphicsItems\GraphicsItem.py", line 528, in _updateView
    self.viewRangeChanged()
  File "<stdin>", line 39, in viewRangeChanged
  File "<stdin>", line 54, in replot
  File "C:\Program Files\Python39\lib\site-packages\pyqtgraph\graphicsItems\PlotCurveItem.py", line 345, in setData
    self.updateData(*args, **kargs)
  File "C:\Program Files\Python39\lib\site-packages\pyqtgraph\graphicsItems\PlotCurveItem.py", line 398, in updateData
    raise Exception("X and Y arrays must be the same shape--got %s and %s." % (self.xData.shape, self.yData.shape))
Exception: X and Y arrays must be the same shape--got (2979,) and (2980,).
```

PR #1575, opened 2021-02-14
Reduce number of shims for PyQt6

pijyoi:


j9ac9k:
@pijyoi I like these changes (the key one I'm on the fence on); Given that you've done the bulk of the Qt.py shimming work recently, what do you think is a reasonable for the library to provide (given that you submitted this PR, I'm going to guess you think we should reduce the content in `Qt.py` itself)?

pijyoi:
Well, the shims made for Qt4/Qt5 are a done deal and have become a de facto API for pyqtgraph. e.g.
(I must admit that at one time I myself was using pyqtgraph as the compatibility layer for Qt4/Qt5, at least for QtGui/QtWidgets namespace.)

#1533 breaking https://github.com/biolab/orange-bio got me to thinking about the monkey-patching that was done for Qt6. Monkey-patching subjects downstream users of the library to side effects which they may not have wanted. If, for instance, another Qt-using library were to do its own monkey-patching, there would be incompatibility. There would be no problem, of course, if pyqtgraph were only to be used by itself. 

The remainder of the PyQt6 shims not modified by this PR are a tradeoff against having to make numerous changes all over the library just to be compatible to PyQt6; and even then, there is no guarantee that all places needing modification would have been caught.

It's not a problem to provide conveniences to the user, but ideally, the user should have been able to opt out of it. i.e. the shims should be made in a separate namespace, and not by monkey-patching.

> (the key one I'm on the fence on);

I am fine with dropping this particular commit because we are, after all, not modifying the original names.
i.e. ```QtCore.Qt.Key.Key_A``` remains unmodified as an enum while monkey-patched ```QtCore.Qt.Key_A``` is an integer

j9ac9k:
Hi @pijyoi 

>  If, for instance, another Qt-using library were to do its own monkey-patching, there would be incompatibility.

Do you mean if an end user does something like

```python
# file A.py

QtGui.stuff()
```

and 

```python
# file B.py

from someOtherQtLib import QtGui

QtGui.stuff()
```

where there may be inconsistent behavior? ... If this is the use-case, I'm not sure that pyqtgraph can really be on the hook here.


Using one of your changes as an example

```python
action = self.drag.exec_() if hasattr(self.drag, 'exec_') else self.drag.exec()
```

This is forcing maintainers/contributors to have to keep in mind these gotchas as we work throughout the code-base, and I know my memory isn't that great... If the trade-off here is that we risk incompatibilities in the example I created; I'm good with causing that kind of pain on our downstream users (maybe we can make start making the bindings more ..."private").

j9ac9k:
Also @pijyoi your work was cited in QtPy https://github.com/spyder-ide/qtpy/pull/225#issuecomment-770239879 👍

pijyoi:
I am thinking along the following: file A that doesn't import pyqtgraph directly is still affected by file B's import of pyqtgraph.
For the case of biolab/orange3, the mere fact of having imported pyqtgraph somewhere gave QGraphicsItem the Qt4 rotate method.

```python
# file A
from PySide2.QtWidgets import QGraphicsItem
print(hasattr(QGraphicsItem, 'rotate'))    # False
import B
print(hasattr(QGraphicsItem, 'rotate'))    # True
```

```python
# file B
import pyqtgraph as pg
```


>>action = self.drag.exec_() if hasattr(self.drag, 'exec_') else self.drag.exec()

In fact, this was the original version during the initial porting of PyQt6. But then I thought that it would be easier to maintain if the various differences were largely localized to Qt.py. So my thoughts then were the same as yours now: make things easier for pyqtgraph developers.


By the way, it looks like QtPy is going the route of monkey patching PySide6 to give it some Qt4 API.
https://github.com/spyder-ide/qtpy/pull/225/files

j9ac9k:
I see the problem now, thanks for show-casing it.  I can be talked into both sides of this, but given the amount of work you've done on shimming, I'm inclined to implement whatever you think is best.  

I'd like to get input from @outofculture @markotoplak @lidstrom83 @danielhrisca  (or anyone else that may have opinions here).

> By the way, it looks like QtPy is going the route of monkey patching PySide6 to give it some Qt4 API.

I suppose they sort of have to to maintain backward compatibility.

outofculture:
Yeah, I agree with @pijyoi on the problems with monkey patches. Since we already maintain our own Qt-alike namespace, if we want to provide our users with conveniences, it feels nicer to just override things in our namespace, rather than at the source. `pg.Qt.QObject().customMethod` or whatever's appropriate is fine.

j9ac9k:
Alright, I think I'm fairly convinced this should be merged, I'm going to hold out for another ~12 hours and give some of the other people I tagged an opportunity to chime in tho.

j9ac9k:
Haven't heard any other concerns, so I'm going to merge this.  Thanks @pijyoi !

PR #1576, opened 2021-02-14
Deprecate unused/unnecessary modules

lidstrom83:
- ordereddict
- pgcollections
- lru_cache
- pil_fix

### DIFF
--- a/examples/ExampleApp.py
+++ b/examples/ExampleApp.py
@@ -6,7 +6,7 @@ import subprocess
 from argparse import Namespace
 import pyqtgraph as pg
 from pyqtgraph.Qt import QtGui, QtCore, QT_LIB
-from pyqtgraph.pgcollections import OrderedDict
+from collections import OrderedDict
 from .utils import examples
 
 path = os.path.abspath(os.path.dirname(__file__))
--- a/pyqtgraph/configfile.py
+++ b/pyqtgraph/configfile.py
@@ -11,7 +11,7 @@ as it can be converted to/from a string using repr and eval.
 
 import re, os, sys, datetime
 import numpy
-from .pgcollections import OrderedDict
+from collections import OrderedDict
 from . import units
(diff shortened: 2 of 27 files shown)


j9ac9k:
Looks like some parts of our test suite are calling the now deprecated methods:

```
=============================== warnings summary ===============================
pyqtgraph/dockarea/tests/test_dockarea.py:5
pyqtgraph/dockarea/tests/test_dockarea.py:5
  /home/runner/work/pyqtgraph/pyqtgraph/pyqtgraph/dockarea/tests/test_dockarea.py:5: DeprecationWarning: OrderedDict is in the standard library for supported versions of Python. Will be removed in 0.13
    from pyqtgraph.ordereddict import OrderedDict

pyqtgraph/util/tests/test_lru_cache.py:1
pyqtgraph/util/tests/test_lru_cache.py:1
  /home/runner/work/pyqtgraph/pyqtgraph/pyqtgraph/util/tests/test_lru_cache.py:1: DeprecationWarning: No longer used in pyqtgraph. Will be removed in 0.13
    from pyqtgraph.util.lru_cache import LRUCache
```

In the second failure, `test_lru_cache.py` we should probably just suppress the warning emitted there.  In `test_dockarea.py` we should probably change the import itself.

lidstrom83:
Are these warnings causing failures? I'm fine with avoiding or suppressing them. I left them in the tests because I didn't think they'd cause test failures and to serve as an extra reminder to remove these modules later.

lidstrom83:
Oh, not the test_dockarea.py one. That was an oversight.

j9ac9k:
Hi @lidstrom83 I meant to reply earlier, idea here was that I'm hoping to eliminate all warnings from the test suite (makes it a bit easier to notice if we hit a new deprecation warning from an upstream library).

This LGTM, merging.  Thanks for the PR!.

PR #1573, opened 2021-02-12
Add deprecation warnings, Use non-deprecated upstream methods

j9ac9k:
In many portions of the codebase, there are commented out sections about methods or classes, or key-word arguments being deprecated, but they do not emit actual deprecation warnings.

This PR makes it so the use of deprecated methods emit deprecation warnings.  Furthermore, the code-base, examples and test-suite have been altered so that the deprecated methods are not used (exception in the test suite, where some tests test those explicitly).

This PR also restore the public API bits whose removal caused #1535  (restored, but added a deprecation warning).

Furthermore, thanks to @swt2c we got a heads up on issues involving Python 3.10, this PR addresses those as well.

Fixes #1535 
Fixes #1571

PR #1577, opened 2021-02-14
Remove duplicate import and remove whitespace

j9ac9k:
I merged #1573 a little quick, @ksunden pointed out some issues after the merge, this PR implements those changes.

### DIFF
--- a/pyqtgraph/__init__.py
+++ b/pyqtgraph/__init__.py
@@ -395,7 +395,7 @@ def plot(*args, **kargs):
     Accepts a *title* argument to set the title of the window.
     All other arguments are used to plot data. (see :func:`PlotItem.plot() <pyqtgraph.PlotItem.plot>`)
     """
-    mkQApp()      
+    mkQApp()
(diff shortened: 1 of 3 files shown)


pijyoi:
I suspect using ```np.can_cast()``` here would almost always end up casting it to np.float64.

if ```levels is None```, then we end up with dtype int32 which will then be casted to np.float64
```python
In [15]: np.array([0, 255]).dtype
Out[15]: dtype('int32')
In [16]: np.can_cast(np.int32, np.float32)
Out[16]: False
```
else, the caller probably provided levels as python int or float, which when converted to ndarray would also end up casted to np.float64

```levels``` is eventually passed as the 2nd and 3rd arguments to ```rescaleData()``` which uses them in a way that doesn't care about their type.

I could be wrong, ```makeARGB()``` is complex!
But I think this PR that is about removing deprecations should try to avoid changing behavior at the same time, i.e. a safe move would be to unconditionally cast to np.float64, as originally pointed out by @ksunden

j9ac9k:
That's completely reasonable, I'll update accordingly.

ksunden (approved):

PR #1578, opened 2021-02-14
Example: Add AutoRange SpeedTest Example

dgoeries:
@j9ac9k 

Another example one could add to the package.


examples/PlotLayoutSpeedTest.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+"""
+Update a simple plot as rapidly as possible to measure speed.
+"""
+
+## Add path to library (just for examples; you do not need this)
```
lidstrom83, line 7: This comment should go above the `import initExample` line

pijyoi:
Compared to the other performance tests, this test limits the time measurement to only the ```curve.setData()``` line. While this is a valid way of measurement, it does mean that the reported ```fps = 1 / average``` does not give the same meaning of including all other overheads needed to redraw the screen.

dgoeries:
I believe i finally found it and rewrote the example to autorange to show the regression between 0.11.0 and master.

lidstrom83:
Can this measurement not be achieved already with PlotSpeedTest.py? For example, by running it and then clicking the auto range button I can detect the performance drop from #1435.

dgoeries:
thanks, yes @lidstrom83 that is correct. I don't intend to merge this one. If you can confirm the performance drop from #1435 i am happy and done.

PR #1582, opened 2021-02-16
Pin sphinx version to avoid buggy sphinx.ext.viewcode

ixjlyons:
I'm watching sphinx releases so I'll see when 3.5.1 comes out and update accordingly

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,5 +1,5 @@
 pyside2
 numpy
 pyopengl
-sphinx
-sphinx_rtd_theme
+sphinx==3.4.3
+sphinx_rtd_theme==0.5.1


ksunden:
we should just pin != 3.5.0... we know the fix will be released soon

j9ac9k:
> we should just pin != 3.5.0... we know the fix will be released soon

I thought about that, but then I realized we have no guarantee that once that issue is fixed, that we won't run into issues on 3.5.1 (or if there were other issues on 3.5.0 that affect us but there was no fix for in 3.5.1)

ixjlyons:
I originally did `sphinx<3.5.0` but thought about for another minute and figured it wouldn't hurt to just lock down the versions and occasionally check and upgrade.

Btw I had used the rtd branch here to test builds on readthedocs (it's enabled but hidden). There's not really a need to keep the branch hanging around but I should've mentioned rationale for pushing to upstream.

PR #1583, opened 2021-02-16
RawImageWidget.py : fix port to QOpenGLWidget

pijyoi:
There are some minor differences between QGLWidget and QOpenGLWidget that were not addressed during the port to QOpenGLWidget.

Qt5 docs specify differences between QGLWidget and QOpenGLWidget:
https://doc.qt.io/qt-5/qopenglwidget.html#relation-to-qglwidget
```...when invoking paintGL().  QOpenGLWidget sets up the viewport via glViewport(). It does not perform any clearing.```
```...it is important to call glClear() as early as possible in paintGL()...```

need someone to test that hi-dpi continues to work without any need to call glViewport()

```python
import scipy.misc
import pyqtgraph as pg
from pyqtgraph.widgets.RawImageWidget import RawImageWidget, RawImageGLWidget

data = scipy.misc.face()

app = pg.mkQApp()
win = pg.Qt.QtWidgets.QMainWindow()
wgt = RawImageGLWidget()
wgt.setImage(data)
win.setCentralWidget(wgt)
win.resize(800, 600)
win.show()
app.exec_()
```

### DIFF
--- a/pyqtgraph/widgets/RawImageWidget.py
+++ b/pyqtgraph/widgets/RawImageWidget.py
@@ -134,6 +134,8 @@ if HAVE_OPENGL:
             glDisable(GL_TEXTURE_2D)
 
         def paintGL(self):
+            glClear(GL_COLOR_BUFFER_BIT)
+
             if self.image is None:
                 if self.opts is None:
                     return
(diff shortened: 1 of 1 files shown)


j9ac9k:
I think this undoes a change I made to support hidpi/fractional scaling; I'll test this out in a bit.

pijyoi:
The C++ source code shows that ```glViewport()``` is called for the user, taking into account ```devicePixelRatioF()```.
https://github.com/qt/qtbase/blob/5.12/src/widgets/kernel/qopenglwidget.cpp#L869-L886

So in any case, the current Python code that uses ```devicePixelRatio()``` instead of ```devicePixelRatioF()``` is wrong for Qt5.

j9ac9k:
Just tested it out on windows and macOS; PR works great.

Also just wanted to say, I appreciate you linking/encouraging me to look at the qt source code, that's something I need to get in the habit of doing more routinely.

Anyway, as I said, this LGTM, merging.  Thanks for the PR @pijyoi

PR #1587, opened 2021-02-18
app.dark_mode => app.property('darkMode')

irgolic:
A refactor, just to make it less magical.


j9ac9k:
Thank you @irgolic 

This LGTM.

PR #1589, opened 2021-02-18
Use OS palette to set background and foreground colors in Example App

### LINKED ISSUE #1588: pyqtgraph does not use QPalette colors
### Short description
pyqtgraph sets the background to white and text color to black. In example implementations of a dark mode, these are manually switched.

Qt uses QPalette to specify colors https://doc.qt.io/qt-5/qpalette.html#ColorRole-enum and automatically updates them when the style changes (e.g., Mac switching between light and dark mode). 

Pyqtgraph should use QPalette colors by default, otherwise, dependencies have to manually propagate palette changes to pyqtgraph.

### Code

The global QApplication instance exposes a `paletteChanged` signal. So essentially, you could include this snippet of code in any file that you know will be imported after the QApplication is created.
```python
@QApplication.instance().paletteChanged
def _(palette):
    bg = palette.base().color().name()
    fg = palette.windowText().color().name()

    setConfigOption('background', bg)
    setConfigOption('foreground', fg)
```
From my understanding, setting a config option like this only sets the default; it doesn't override a developer specifying a color like in `pg.PlotWidget(background='w')`.

j9ac9k:
Fixes #1588

### DIFF
--- a/examples/ExampleApp.py
+++ b/examples/ExampleApp.py
@@ -260,7 +260,6 @@ class PythonHighlighter(QSyntaxHighlighter):
             return False
 
 
-
 class ExampleLoader(QtGui.QMainWindow):
     def __init__(self):
         QtGui.QMainWindow.__init__(self)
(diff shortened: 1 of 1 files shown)


j9ac9k:
Complete bogus implementation, ... need to have the tie-ins occur elsewhere...

j9ac9k:
Going to close this out, as @irgolic mentioned a better method of going about this in #1588

PR #1594, opened 2021-02-19
Fix Plotitem regression

outofculture:
b7b431de8d31c1d894ffc368ff5cd562fcc13678 introduced a bug by changing the for-loop structure.

While I'm in here, I can also fix all the obvious problems.


j9ac9k:
Given I merged that PR, should have caught this!  ...thanks for patching this up @outofculture , this LGTM, merging.

PR #1595, opened 2021-02-20
Workaround for cupy indexing bug

outofculture:
The [bug](https://github.com/cupy/cupy/issues/4693) was caught by running our tests against a cupy backend, so yay, us!


j9ac9k:
Thanks for tracking this down @outofculture !  This LGTM, merging.

j9ac9k:
Thanks tracking this down @NilsNoreyson !

@outofculture you good with reverting this?

j9ac9k:
I'm actually a little reluctant to outright do away w/ the old codepath, maybe we have an if statement depending on the version of cupy?  After we get to 2 years since cupy 10.0 was released (what NEP-29 suggests we deprecate numpy versions) we then get rid of the old method?

PR #1596, opened 2021-02-20
Add ColorBarItem for simplified image level adjustment

NilsNemitz:
It seems that there currently isn't a convenient way to just display a simple color bar to accompany a false color image.
HistogramLUTItem is very powerful, but it has a few downsides:
- Colors are only indirectly matched to numerical values
- The LUT adjustments are not needed/appropriate when using percepetually uniform maps
- The histogram (besides requiring additional calculation) strongly associates the adjuster with one specific image

This is an alternative implementation that is more compact, and hopefully simple to use:
`cmap = pg.colormap.get('fire', source='colorcet')`
`bar = pg.ColorBarItem(values = (0, 30_000), cmap=cmap)`
`bar.setImageItem( img, insert_in=plt )`

It works smoothly with the new functions to get colormaps and has a convenience function to insert itself into the internal layout of a PlotItem.
It uses a LinearRegionItem superimposed on the color bar for interaction and can be assigned any number of ImageItems to control. Signals are emitted during and after range adjustments. 

This PR comes with an example and a first attempt at usable docstrings.
I hope it is useful to somebody, comments and suggestions are welcome!

color bar tests


NilsNemitz:
The failed test seems to be:
```
[gw1] node down: Not properly terminated
[gw1] [ 73%] FAILED pyqtgraph/tests/test_functions.py::test_interpolateArray_order0 
replacing crashed worker gw1
```
Unrelated?

pijyoi:
You should add your example into examples/utils.py

NilsNemitz:
I added a line to examples/utils.py, although I am not 100% sure what that does. :)
I converted the example to the style used in other examples and checked that it now runs as part of the tests.

When looking at the other examples, I saw that colorMaps.py had a copy-and-pasted wrong description.
I took the liberty to fix that to now read:
```
This example demonstrates generating ColorMap objects from external data.
It displays the full list of color maps available as local files or by import 
from Matplotlib or ColorCET.
```
Not sure if that is worth opening a separate PR for.

j9ac9k:
> I added a line to examples/utils.py, although I am not 100% sure what that does. :)

https://github.com/pyqtgraph/pyqtgraph/blob/master/examples/ExampleApp.py#L10

Adding the file there makes it show up in the example application that you can pull up via `python -m pyqtgraph.examples`

j9ac9k:
Running this example, I just get grayscale colors...

image


Also I got a segmentation fault when closing it out... ....the first time, and I can't replicate that...

j9ac9k:
> Unrelated?

Yeah that test has been occasionally failing in CI.

NilsNemitz:
@j9ac9k , sorry about that, I guess I should be using the locally included color maps instead of importing from colorcet, which might not be installed.

I found a left-over call to Q from my original example code. I could imagine that Qt tried to open two windows for the program and tripped over that. This commit should hopefully fix that. 

I hope this version will be more reasonably behaved. The color bar itself has no low-level functionality... I would expect all my bugs there to crash on the Python layer.

j9ac9k:
Oh right, I didn't have colorcet installed in the virtual environments I tested

j9ac9k:
tested the example again, I'm 👍 will give some other folks an opportunity to chime in before merging.

pijyoi:
In view of #1605, do you think it would be better to revert your example more to the style of ef1655d524ab910476c0e0354adca10639437c32?

NilsNemitz:
No problem with changing the example... may take a few days, though.
@pijyoi , are you happy with the changes you made to JoystickButton.py for #1605 ? Then I can emulate that style. 
I don't have a full multi-platform development environment here, so I cannot easily test what exits cleanly and what may crash.
And I am not sure if any special trickery is required to integrate with the testing code :)

j9ac9k:
> No problem with changing the example... may take a few days, though.
> 
> @pijyoi , are you happy with the changes you made to JoystickButton.py for #1605 ? Then I can emulate that style. 
> 
> I don't have a full multi-platform development environment here, so I cannot easily test what exits cleanly and what may crash.
> 
> And I am not sure if any special trickery is required to integrate with the testing code :)

I'm happy to test this PR on macOS and windows, maybe I should take this as the hint to setup my Linux dev environment

pijyoi:
@NilsNemitz , your original example would be the ideal formulation. Joystick.py instantiates MainWindow at module level only to remain compatible with the existing framework.

NilsNemitz:
@pijyoi I have reverted the example to be a class again. But I think in the current form, it does not run as part of "test.py". At least I failed to see it run or be referenced in the test output. The previous, "flat" version definitely ran. Any advice?

Otherwise, this commit
- defines horizontal colorbars and shows one in the example
- adds optional labels for the colorbar
- promotes the axes to be on top of the colorbar image instead of getting obscured
- removes mismatched tickmarks on colorbar axes except for the main axis
- exposes (and sets defaults for) the hoverPen and hoverBrush properties of LinearRegionItem

I also added a few words to the docstring of LinearRegionItem:
"Options are 'vertical' or 'horizontal'
The default is 'vertical', indicating that the region _is bounded by vertical lines_."

j9ac9k:
Your example definitely runs, but a window doesn't pop up:

```
pyqtgraph-pyqt5_512-py37 ❯ pytest -k "ColorBar" -v
============================================================= test session starts ==============================================================
platform darwin -- Python 3.7.9, pytest-6.2.2, py-1.10.0, pluggy-0.13.1 -- /Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.7.9/envs/pyqtgraph-pyqt5_512-py37/bin/python3.7
cachedir: .pytest_cache
rootdir: /Users/ogi/Developer/pyqtgraph, configfile: pytest.ini
plugins: xdist-2.2.1, forked-1.3.0
collected 214 items / 213 deselected / 1 skipped

examples/test_examples.py::testExamples[ ColorBarItem.py - PyQt5 ] PASSED                                                                [100%]

================================================= 1 passed, 1 skipped, 213 deselected in 2.23s =================================================

~/Developer/pyqtgraph add_ColorBarItem*
pyqtgraph-pyqt5_512-py37 ❯
```

j9ac9k:
hmm... I added an `assert False` in the `MainWindow.__init__()` method and the example still "passed".

pijyoi:
Replace
```python
## Start Qt event loop
if __name__ == '__main__':
    main_window = MainWindow()
    sys.exit(app.exec_())
```

With something like
```python
mkQApp("ColorBarItem Example")
main_window = MainWindow()

## Start Qt event loop
if __name__ == '__main__':
    QtWidgets.QApplication.instance.exec_()
```

j9ac9k:
Confirmed that worked, thanks @pijyoi

examples/ColorBarItem.py
```diff
@@ -0,0 +1,100 @@
...
+        bar.getAxis('bottom').setHeight(21)
+        bar.getAxis('top').setHeight(31)
+        gr_wid.addItem(bar, 0,2, 2,1) # large bar spanning both rows
+
+mkQApp("ColorBarItem Example")
+main_window = MainWindow()
+
+## Start Qt event loop
+if __name__ == '__main__':
+    QtWidgets.QApplication.instance().exec_()
```
j9ac9k, line 100: Can you replace with `pg.mkQApp().exec_()`?

NilsNemitz:
Dear @j9ac9k , thanks for looking at this.
I have switched the example initialization to `mkQApp().exec_()` in the latest commit.
I have also fixed some incorrect comments, particularly regarding the generated look-up table. As of now, this is fixed to produce 256 steps, independent of image data format.

j9ac9k:
Thanks @NilsNemitz this LGTM.  This feature IMO was critically missed, so I'm really glad this is now implemented.

PR #1597, opened 2021-02-20
RawImageWidget: transpose did not handle luminance only images

pijyoi:
This also fixes a performance issue due to calling ```makeQImage()``` with default arguments copy=True, transpose=True.

### DIFF
--- a/pyqtgraph/widgets/RawImageWidget.py
+++ b/pyqtgraph/widgets/RawImageWidget.py
@@ -99,6 +99,8 @@ if HAVE_OPENGL:
             img must be ndarray of shape (x,y), (x,y,3), or (x,y,4).
             Extra arguments are sent to functions.makeARGB
             """
+            if getConfigOption('imageAxisOrder') == 'col-major':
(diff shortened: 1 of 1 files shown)


j9ac9k:
Sorry you had to redo my attempted fix 😆   Thanks for addressing this, like your solution much better.  

Unrelated, do you know why `copy=True` is a default argument for `makeQImage`?

pijyoi:
>Unrelated, do you know why copy=True is a default argument for makeQImage?

Probably in order to be more forgiving in the types of arrays that it accepts by default. e.g. non-contiguous arrays
For performance, we would want copy=False and transpose=False.

Doing a grep for makeQImage, there are now only 3 uses of makeQImage that rely on either the default argument of copy and/or transpose.
1) GLViewWidget.py::readQImage()
    - this one is a good candidate for copy=False
    - but there are no users in the library of this method
2) image_testing.py (2 uses)

It would be a good practice to explicitly specify the arguments even if the defaults are what you want to serve as documentation.

PR #1598, opened 2021-02-21
Remove references to self from lambdas

outofculture:
Lambdas used as signal responders which reference `self` prevent GC of said object. Instance methods do not suffer the same problem. I wasn't sure if the `QTimer` one needed to be fixed, but it doesn't hurt to be safe.

I couldn't solve `pyqtgraph.WidgetGroup.WidgetGroup.mkChangeCallback`, though. That case can't be rewritten as an instance method, and it needs to remain associated with `self`, so it can't be a static function. Maybe it would be sufficient to add the appropriate `disconnect` somehow? But what would that look like? If anyone has ideas, I'm willing to take another stab at this one.

### DIFF
--- a/pyqtgraph/graphicsItems/GradientEditorItem.py
+++ b/pyqtgraph/graphicsItems/GradientEditorItem.py
@@ -446,10 +446,10 @@ class GradientEditorItem(TickSliderItem):
         
         self.rgbAction = QtGui.QAction(translate("GradiantEditorItem", 'RGB'), self)
         self.rgbAction.setCheckable(True)
-        self.rgbAction.triggered.connect(self.setColorModeToRGB)
+        self.rgbAction.triggered.connect(self._setColorModeToRGB)
         self.hsvAction = QtGui.QAction(translate("GradiantEditorItem", 'HSV'), self)
         self.hsvAction.setCheckable(True)
-        self.hsvAction.triggered.connect(self.setColorModeToHSV)
+        self.hsvAction.triggered.connect(self._setColorModeToHSV)
             
         self.menu = QtGui.QMenu()
         
@@ -493,7 +493,7 @@ class GradientEditorItem(TickSliderItem):
         self.updateGradient()
         self.linkedGradients = {}
         
-        self.sigTicksChanged.connect(self.updateGradientIgnoreArgs)
+        self.sigTicksChanged.connect(self._updateGradientIgnoreArgs)
         self.sigTicksChangeFinished.connect(self.sigGradientChangeFinished.emit)
 
     def showTicks(self, show=True):
@@ -567,10 +567,10 @@ class GradientEditorItem(TickSliderItem):
         self.sigTicksChanged.emit(self)
         self.sigGradientChangeFinished.emit(self)
 
-    def setColorModeToRGB(self):
+    def _setColorModeToRGB(self):
         self.setColorMode("rgb")
 
-    def setColorModeToHSV(self):
+    def _setColorModeToHSV(self):
         self.setColorMode("hsv")
 
     def colorMap(self):
@@ -591,7 +591,7 @@ class GradientEditorItem(TickSliderItem):
         self.gradRect.setBrush(QtGui.QBrush(self.gradient))
         self.sigGradientChanged.emit(self)
 
-    def updateGradientIgnoreArgs(self, *args, **kwargs):
+    def _updateGradientIgnoreArgs(self, *args, **kwargs):
         self.updateGradient()
 
     def setLength(self, newLen):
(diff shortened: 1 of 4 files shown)


pijyoi:
Could most of these new methods be replaced by a use of functools.partial?

For Parameters, a change in prototype of emitStateChanged
```python
def emitStateChanged(self, param, data, changeDesc):
    ...

self.sigOptionsChanged.connect(partial(self.emitStateChanged, changeDesc='options'))
```

pijyoi:
For mkChangeCallback, could the single call to self.mkChangeCallback(w) be converted to partial(self.widgetChanged, w=w)?

outofculture:
Okay, yeah, partials would be more elegant and solve the `mkChangeCallback` problem. I don't know if partials suffer from the same problem as lambdas; they might preserve the reference to self somehow. I was just working off of recommendations on stack overflow. I need a test that I can prove it one way or the other. I'll try to build that, and then use it to prove out whether partials would work.

j9ac9k:
> Okay, yeah, partials would be more elegant and solve the `mkChangeCallback` problem. I don't know if partials suffer from the same problem as lambdas; they might preserve the reference to self somehow. I was just working off of recommendations on stack overflow. I need a test that I can prove it one way or the other. I'll try to build that, and then use it to prove out whether partials would work.

Feel free to use PR #646 with a PyQt5/6 as an example.

Change:

```python
        self.vb.sigLogChanged.connect(self._logChanged)
```

back to a lambda and `test_ref_cycles` will fail with this exact issue with C++ object already deleted...

outofculture:
@j9ac9k way to pick the most challenging to convert to a partial! 😁

Playing around with that, I was able to prove that partials still trigger the error. Thanks, everyone, for the ideas, but it looks like partials aren't going to be helpful here. It looks like PyQt* is doing extra work to make sure instance methods get disconnected appropriately. I could use `functools.partialmethod` to declare these one-liner methods, but we wouldn't be much better off for it, and it still doesn't help with `mkChangeCallback`.

pijyoi:
Could you share how your partial() line looks like?

1) partial(self.method, arg='xyz')
or
2) partial(self.method, arg=self)?

(2) should cause problems. I am wondering if (1) also causes problems.

outofculture:
The first. The following fails the test: `self.vb.sigLogChanged.connect(partial(self.setLogMode, x=1, y=1))` with the same failure as a lambda. I just hard-coded the args, but I made sure that hard-coding them in the `_logChanged` method didn't also produce the same error.

pijyoi:
https://stackoverflow.com/questions/47941743/lifetime-of-object-in-lambda-connected-to-pyqtsignal/47945948#47945948

This here says that instance methods are treated specially. Mentions that lambda, partial and others don't get this special treatment. ☹

j9ac9k:
@pijyoi thanks for sharing that link, the addition of "static methods" is something I would never have thought to be applicable.  I am now taking a hard look at some of my projects involving pyqtgraph...

j9ac9k:
2 Questions...

1. Do we want to make some of these methods "private" by setting the `_` before the method name?
2. Not related to this PR, but is there a reason we don't use the `@Slot` decorator?

Regarding the private methods, I'm just bringing this up; I'm not advocating for a change to be made here...

outofculture:
Ooo, now that you mention it, I generally like private-ing methods that aren't useful as a public interface, yeah. This is a big library, and not having to worry about maintaining every little utility function is of great value to the maintainers. I'm changing it!

No opinion on the `@Slot` decorator. I'll read up on it, though.

outofculture:
Just a note that I looked at this some more (with help from @j9ac9k) and I failed to be able to disconnect lambdas as part of object deletion. `self.destroyed.connect(self._onDestroy)` does not ever call `self._onDestroy()`, at least not in the #646 test I was using. I even tried putting the `_onDestroy` onto a different object, under the suspicion that PyQt was disconnecting all instance methods associated with the to-be-deleted object before emitting its `destroyed` signal.

Any other ideas for improving this PR?

dgoeries:
Using the @Slot decorator is beneficial.

**Connecting a signal to a decorated Python method also has the advantage of reducing the amount of memory used and is slightly faster.**

j9ac9k:
@dgoeries I've seen this talked about before, but I've never been able to test this.  Every time I try and get specific with my slots...

```python
@Slot(tuple)  # using tuple as an example, I haven't done exhaustive testing, but this holds for many cases...
def someSlot(self, arg1):
    sender = self.sender()
    isinstance(sender, QObject)  # this returns False because sender is None in this case
    ...
```

but when I do...

```python
@Slot(object)
def someSlot(self, arg1):
    sender = self.sender()
    isinstance(sender, QObject)  # this returns True
    ...
```

So if I'm just decorating my slots to pass `object`, is there really any performance boost?

This is outside the scope of this PR, but I am eager to know if this works and if we can measure the performance benefit.  

In the meantime, I think I'm going to merge this as is.  Thanks @outofculture !

PR #1599, opened 2021-02-22
fix Example App launch of VideoSpeedTest

pijyoi:
Fixes the following:
1) ExampleApp could not choose binding to launch VideoSpeedTest
2) setGraphicsSystem is obsolete in Qt5
3) VideoSpeedTest RawGLImageWidget frame rate capped by vsync (this should have been a side effect of converting from QGLWidget to QOpenGLWidget)
   * on Windows 10, instantiating RawImageGLWidget also caps the frame rates of the other two (ImageItem, RawImageWidget)
   * on Linux, doesn't affect the other two


j9ac9k:
Hi @pijyoi thanks for this PR!  Had no idea there was a vsync cap issue on the benchmark, that certainly was problematic.

I'm messing with this PR, and noticing some major performance differences, unlikely due to anything in this PR, but curious if you've seen something equivalent, especially for Qt6 based variants (this is with otherwise default settings).  All numerical values are in FPS, I put NA for the RawGLImageWidget for some of my test environments as they are not in Python 3.9, which you need python 3.9.1+ for pyopengl to work on macOS 11.

| Binding      | GraphicsView + ImageItem | RawImageWidet | RawGLImageWidget |
| :----------- | ------ | ----- | ----- |
| PyQt5 5.12   | 340	| 390 	| NA 	|
| PyQt5 5.15   | 360	| 380	| NA	|
| PyQt6        | 270	| 290 	| 240 	|
| PySide2 5.12 | 300	| 350	| NA	|
| PySide2 5.15 | 360	| 410	| NA 	|
| PySide6      | 240	| 250	| 200 	|

I see no reason not to merge this PR, but wasn't sure where else to post this benchmark discrepancy

j9ac9k:
Also on PySide6, I can trigger a segfault when I close the two windows that come up from the "command-line usage" example (no segfault on pyside2, pyqt6)

```
Using PySide6
Fatal Python error: Segmentation fault

Current thread 0x000000010c4e7e00 (most recent call first):
<no Python frame>
```

j9ac9k:
Also realized I get the segfault on master right now as well..

stonebig:
hi, also on latest master with PyQt-5.15 on windows, the "3D Graphics" examples fail to launch, is it normal ? 
image

j9ac9k:
The issues I brought up are outside the scope of this PR.  This PR LGTM, merging.

pijyoi:
Ubuntu 20.04 64-bits using system Python 3.8, all packages from pypi.org.
No PySide2 5.12 wheel available for Python > 3.7
At least on my test system, values seem to be close together enough across different bindings.
Binding | GraphicsView + ImageItem | RawImageWidet | RawGLImageWidget
-- | -- | -- | --
PyQt5 5.12 | 590 | 720 | 500
PyQt5 5.15 | 600 | 700 | 480
PyQt6 | 600 | 700 | 480
PySide2 5.12 | NA | NA | NA
PySide2 5.15 | 580 | 710 | 480
PySide6 | 580 | 690 | 470

>Also on PySide6, I can trigger a segfault when I close the two windows that come up from the "command-line usage" example (no segfault on pyside2, pyqt6)

I have seen segfaults when running the examples on different systems. Seems to depend on Python version, binding, operating system etc. When they do happen, it's reproducible, not random. I don't get segfaults running the examples on my Windows 10 + Python 3.7 and Linux + Python 3.8 systems. 

At one time, I was trying to make test_examples.py give each example a chance to end by itself but that ended up triggering many segfaults on the CI. So in the end, I thought that maybe using ```process.kill()``` to terminate each example was intentional in order to not have to deal with these segfaults.

I suspect one reason (but not the only) is that the examples are written in an interactive manner, instead of wrapping up things in a master parent widget like a QMainWindow.

pijyoi:
Looking at
https://github.com/winpython/winpython/blob/master/changelogs/WinPythoncod-64bit-3.9.1.0.md,
I see that pyopengl package is included, so that should be sufficient to
run the 3D examples.

So it's not normal.
Are you able to run "python -m pyqtgraph.opengl.glInfo"? This one doesn't
depend on pyopengl.

j9ac9k:
Woah, those numbers... I'm on macOS 11, with no discrete GPU; I do have some other systems I can test on, which I may do...

on macOS I've seen _huge_ performance boosts going from Qt 5.12 to Qt 5.15, but apparently that does not appear to be the case on Ubuntu.

We can continue this discussion elsewhere, I find it interesting, but I am not seeing anything that indicates an issue with the library.

Issue #1526, opened 2021-01-30
Enhancement: Add `keys()` method to Parameter class


ntjess:
According to Python specifications, objects can be `**unpacked` if they possess a `__getitem__` method and a `keys` method. Since `Parameter`s already have a `__getitem__`, the `keys` method will allow such unpacking. I find this quite useful when a single parameter holds all the arguments for a function call:

```python
def myFunc(a=5, b=10, c=15):
  print('called')

params = Parameter(<has children a, b, c>)
myFunc(**params)
```

The current workaround is a somewhat ugly

```python
inputs = {name: param[name] for name in param.names}
```

Simple `keys` implementation:

```python
def keys(self):
  return self.names
```

j9ac9k:
I would certainly approve of this change; if someone wants to put a PR and do a test for it, I'll gladly review/merge 👍

j9ac9k:
Fixed by #1613

PR #1613, opened 2021-02-28
Add keys() method to Parameter class

chrismullins:
This returns the names attribute, allowing **unpacking of Parameter subclasses. Addresses #1526 

Hi all, I saw the request for contributors and thought I'd take a crack at this one. What would be some other tests to include?

### DIFF
--- a/pyqtgraph/parametertree/Parameter.py
+++ b/pyqtgraph/parametertree/Parameter.py
@@ -677,6 +677,9 @@ class Parameter(QtCore.QObject):
             names = (names,)
         return self.param(*names).setValue(value)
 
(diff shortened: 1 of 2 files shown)


j9ac9k:
Hi @chrismullins  This LGTM!  I think for this PR, the tests here are sufficient.  

If you want to add something more to this, I would look through one of the examples in the example directory, and look at one of the ones that use parameter items, and see if there is a good place to highlight this feature.  If you don't see a good place to highlight this functionality, we can leave this as is.

j9ac9k:
Hi @chrismullins I'm going to merge this; thanks again for the PR!

Issue #1590, opened 2021-02-19
ImageItem lut + levels combine is currently a pessimization


pijyoi:
### Short description

Intent of https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ImageItem.py#L419-L440 is to combine levels and lut such that apply levels is a no-op in the subsequent call to ```makeARGB()```

### Code to reproduce

```python
import pyqtgraph as pg
import numpy as np
pg.setConfigOptions(imageAxisOrder='row-major')

app = pg.mkQApp()
imv = pg.ImageView()
imv.show()
data = np.random.randint(0, 256, size=(4096, 4096), dtype=np.uint8)
imv.setImage(data)
app.exec_()
```

### Expected behavior

apply levels should take very little time in ```makeARGB()``` profiling

### Real behavior

both apply levels and apply lut take significant time to execute
```
> Entering functions.makeARGB
  check inputs: 0.0219 ms
  apply levels: 374.5632 ms
  apply lut: 230.2356 ms
  allocate: 0.0088 ms
  reorder channels: 84.9686 ms
  alpha channel: 11.7705 ms
< Exiting functions.makeARGB, total time: 703.3772 ms
```

### Tested environment(s)

 * PyQtGraph version: 0.11.1.dev0
 * Qt Python binding: PySide2 5.15.2 Qt 5.15.2
 * Python version: 3.7.9
 * NumPy version: 1.19.4
 * Operating system: Windows 10 x64
 * Installation method: pip install -e .

### Additional context
As far can I understand, the intended behavior relied on code that was changed in #793.

For uint8 data, 
* a lut of size 256 is created
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ImageItem.py#L424
* levels is automatically calculated as [0, 255]
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L1121-L1122
* and value of ```scale``` _before_ #793 would be set to ```lut.shape[0] - 1``` == 255

The following lines rely on levels[1] == scale in order to bypass apply levels 
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L1182-L1183

So for instance, if #793 were reverted (just for the purpose of demonstration of this issue), we would get the following:
(i.e. apply levels is practically a no-op)
```
> Entering functions.makeARGB
  check inputs: 0.0324 ms
[...]

PR #1617, opened 2021-03-02
Optimize makeARGB for ubyte images

pijyoi:
This PR optimizes the channel reordering portion of makeARGB by using the conversion routines of QImage.

It only optimizes for row major ubyte images and defers to the original code for the rest. Improvement in fps is measurable with VideoSpeedTest 

Would such a PR that optimizes for specific data types be acceptable?
Tagging @outofculture

### DIFF
--- a/examples/VideoSpeedTest.py
+++ b/examples/VideoSpeedTest.py
@@ -165,16 +165,22 @@ def mkData():
                 mx = 1.0
             else:
                 raise ValueError(f"unable to handle dtype: {cacheKey[0]}")
-            
+
+            chan_shape = (width, height)
             if ui.rgbCheck.isChecked():
-                data = xp.random.normal(size=(frames,width,height,3), loc=loc, scale=scale)
-                data = pg.gaussianFilter(data, (0, 6, 6, 0))
+                frame_shape = chan_shape + (3,)
             else:
-                data = xp.random.normal(size=(frames,width,height), loc=loc, scale=scale)
-                data = pg.gaussianFilter(data, (0, 6, 6))
-            if cacheKey[0] != 'float':
-                data = xp.clip(data, 0, mx)
-            data = data.astype(dt)
+                frame_shape = chan_shape
+            data = xp.empty((frames,) + frame_shape, dtype=dt)
+            view = data.reshape((-1,) + chan_shape)
+            for idx in range(view.shape[0]):
+                subdata = xp.random.normal(loc=loc, scale=scale, size=chan_shape)
+                # note: gaussian filtering has been removed as it slows down array
+                #       creation greatly.
+                if cacheKey[0] != 'float':
+                    xp.clip(subdata, 0, mx, out=subdata)
+                view[idx] = subdata
+
             data[:, 10, 10:50] = mx
             data[:, 9:12, 48] = mx
             data[:, 8:13, 47] = mx
--- a/examples/VideoSpeedTest.py
+++ b/examples/VideoSpeedTest.py
@@ -157,7 +157,7 @@ def mkData():
                 dt = xp.uint16
                 loc = 4096
                 scale = 1024
-                mx = 2**16
+                mx = 2**16 - 1
             elif cacheKey[0] == 'float':
                 dt = xp.float32
                 loc = 1.0
--- a/examples/VideoSpeedTest.py
+++ b/examples/VideoSpeedTest.py
@@ -18,6 +18,8 @@ import pyqtgraph as pg
 import pyqtgraph.ptime as ptime
 from pyqtgraph.Qt import QtGui, QtCore, QT_LIB
 
+pg.setConfigOption('imageAxisOrder', 'row-major')
+
 import importlib
 ui_template = importlib.import_module(f'VideoTemplate_{QT_LIB.lower()}')
 
@@ -166,7 +168,7 @@ def mkData():
             else:
                 raise ValueError(f"unable to handle dtype: {cacheKey[0]}")
 
-            chan_shape = (width, height)
+            chan_shape = (height, width)
             if ui.rgbCheck.isChecked():
                 frame_shape = chan_shape + (3,)
             else:
@@ -181,9 +183,9 @@ def mkData():
                     xp.clip(subdata, 0, mx, out=subdata)
                 view[idx] = subdata
 
-            data[:, 10, 10:50] = mx
-            data[:, 9:12, 48] = mx
-            data[:, 8:13, 47] = mx
+            data[:, 10:50, 10] = mx
+            data[:, 48, 9:12] = mx
+            data[:, 47, 8:13] = mx
             cache = {cacheKey: data} # clear to save memory (but keep one to prevent unnecessary regeneration)
 
         data = cache[cacheKey]
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1247,7 +1247,11 @@ def makeARGB(data, lut=None, levels=None, scale=None, useRGBA=False, output=None
         order = [2,1,0,3] # for some reason, the colors line up as BGR in the final image.
         
     # copy data into image array
-    if data.ndim == 2:
+    fastpath = try_fastpath_argb(xp, data, imgData, useRGBA)
+
+    if fastpath:
+        pass
+    elif data.ndim == 2:
         # This is tempting:
         #   imgData[..., :3] = data[..., xp.newaxis]
         # ..but it turns out this is faster:
@@ -1263,11 +1267,12 @@ def makeARGB(data, lut=None, levels=None, scale=None, useRGBA=False, output=None
     profile('reorder channels')
     
     # add opaque alpha channel if needed
-    if data.ndim == 2 or data.shape[2] == 3:
-        alpha = False
-        imgData[..., 3] = 255
-    else:
+    if data.ndim == 3 and data.shape[2] == 4:
         alpha = True
+    else:
+        alpha = False
+        if not fastpath:    # fastpath has already filled it in
+            imgData[..., 3] = 255
 
     # apply nan mask through alpha channel
     if nanMask is not None:
@@ -1280,6 +1285,50 @@ def makeARGB(data, lut=None, levels=None, scale=None, useRGBA=False, output=None
 
     profile('alpha channel')
     return imgData, alpha
+
+
+def try_fastpath_argb(xp, ain, aout, useRGBA):
+    # we only optimize for certain cases
+    # return False if we did not handle it
+    can_handle = xp is np and ain.dtype == xp.ubyte and ain.flags['C_CONTIGUOUS']
+    if not can_handle:
+        return False
+
+    nrows, ncols = ain.shape[:2]
+    nchans = 1 if ain.ndim == 2 else ain.shape[2]
+
+    Format = QtGui.QImage.Format
+
+    if nchans == 1:
+        in_fmt = Format.Format_Grayscale8
+    elif nchans == 3:
+        in_fmt = Format.Format_RGB888
+    else:
+        in_fmt = Format.Format_RGBA8888
+
+    if useRGBA:
+        out_fmt = Format.Format_RGBA8888
+    else:
+        out_fmt = Format.Format_ARGB32
+
+    if in_fmt == out_fmt:
+        aout[:] = ain
+        return True
+
+    npixels_chunk = 512*1024
+    batch = int(npixels_chunk / ncols / nchans)
+    batch = max(1, batch)
+    row_beg = 0
+    while row_beg < nrows:
+        row_end = min(row_beg + batch, nrows)
+        ain_view = ain[row_beg:row_end, ...]
+        aout_view = aout[row_beg:row_end, ...]
+        qimg = QtGui.QImage(ain_view, ncols, ain_view.shape[0], ain.strides[0], in_fmt)
+        qimg = qimg.convertToFormat(out_fmt)
+        aout_view[:] = imageToArray(qimg, copy=False, transpose=False)
+        row_beg = row_end
+
+    return True
 
 
 def makeQImage(imgData, alpha=None, copy=True, transpose=True):


outofculture:
> Would such a PR that optimizes for specific data types be acceptable?

Yes! I feel like the long-term plan for makeARGB should be something like "determine the shape of the data+args, send off to a purpose-built function". The general nature of that function has made optimization too complicated for me to think about.

j9ac9k:
> > Would such a PR that optimizes for specific data types be acceptable?
> 
> Yes! I feel like the long-term plan for makeARGB should be something like "determine the shape of the data+args, send off to a purpose-built function". The general nature of that function has made optimization too complicated for me to think about.

Looking at the function, I see a few places for some optimizations (not counting numba, which is 100% on the table IMO); but I'm not feeling comfortable making a go at it until

1. We're confident that the current function returns the _correct_ results
2. We have more parametrized test cases

outofculture:
Huge improvement, by the way. Using 2048x2048, my system sees a 4.5x (py3.7+qt5) and 8x (py3.9+pyside2) improvement. ( Aside: I should really set up ASV to consolidate and standardize these measurements )

Adding a Scale (`--levels 0,255`) or LUT drops the improvement to 1.5x. It's still an improvement, so no complaints. I don't grok the nature of this PR sufficiently to guess why it loses so much with the LUT.

pijyoi:
Specifying levels (even if equal to the dtype min,max) will trigger #1590.

pijyoi:
Rebased to include #1630 and #1632 
#1630 makes it such that adding "--levels vmin,vmax" will not trigger #1590. If vmin,vmax are the dtype limits, then rescale will get bypassed entirely.
#1632 improves scaling performance for all dtypes.

j9ac9k:
@outofculture I would review this but given the complexity and how much this impact your work, figure I'll give you the honors haha.

@pijyoi thanks for keeping at this.

examples/VideoSpeedTest.py
```diff
@@ -157,27 +159,31 @@ def mkData():
...
-                data = xp.clip(data, 0, mx)
-            data = data.astype(dt)
-            data[:, 10, 10:50] = mx
-            data[:, 9:12, 48] = mx
-            data[:, 8:13, 47] = mx
+                frame_shape = chan_shape
+            data = xp.empty((frames,) + frame_shape, dtype=dt)
+            view = data.reshape((-1,) + chan_shape)
+            for idx in range(view.shape[0]):
+                subdata = xp.random.normal(loc=loc, scale=scale, size=chan_shape)
```
outofculture, line 179: This omits the gaussian blurring of the previous code. I'm not sure if that was in there for any important reason. I read through the history, and it has always been a part of this code, and never with explanation. It does visibly alter the image, which is maybe more indicative of e.g. microscope images. I didn't notice any fps differences either way, so it might be safest to leave it in place.

pijyoi: The gaussian blurring is the function that slows down array creation the most. This becomes observable for larger matrices like 2048x2048. Large matrices are needed to make the profiling timings more measurable. It also makes changing the size on-the-fly from the GUI to be very slow, which makes it harder for experimentation.

It doesn't affect the frame rate. But it makes experimenting easier.

pyqtgraph/functions.py
```diff
@@ -1255,6 +1260,54 @@ def makeARGB(data, lut=None, levels=None, scale=None, useRGBA=False, output=None
...
+    npixels_chunk = 512*1024
+    batch = int(npixels_chunk / ncols / nchans)
+    batch = max(1, batch)
+    row_beg = 0
+    while row_beg < nrows:
+        row_end = min(row_beg + batch, nrows)
+        ain_view = ain[row_beg:row_end, ...]
+        aout_view = aout[row_beg:row_end, ...]
+        qimg = QtGui.QImage(ain_view, ncols, ain_view.shape[0], ain.strides[0], in_fmt)
+        qimg = qimg.convertToFormat(out_fmt)
```
outofculture, line 1304: Nothing else in `makeARGB` does batching, though maybe more should. How did you decide on the batch size of 512KPixels? Does this perform well in Windows? ( I'm unable to access my Windows-based test systems today. ) Is this a special case that makes batching particularly valuable, or is it worth exploring this strategy elsewhere?

examples/VideoSpeedTest.py
outofculture: That's legit. Could you leave something in as a comment for now? I'll dig a bit further to see if there's ever a good cause for smoothing things first.

pyqtgraph/functions.py
pijyoi: The primary purpose _here_ of the batching was to restrict the amount of extra memory used by going through the fast path due to ```QImage.convertToFormat()``` returning a new image.
So for example, without batching, if the input data was 4096x4096 uint8 (16MB), the ARGB output of ```QImage.convertToFormat()``` would result in the (temporary) consumption of another 64MB. The idea was that the user should not have to pay for more memory usage.

In this case, being cache-friendly was a secondary consideration. However 512KBytes was chosen to be a value that hopefully fit within common processor L2 cache-sizes. It's not too small that the looping overhead becomes dominant. But I would agree with you if it seems like a magic value.

As I said, _here_, it's less for performance than for reducing memory usage. After the speed up here, there were much larger bottlenecks elsewhere that would dwarf any minor gains from further tweaking.

I was trying out this batching strategy within ```rescaleData()``` where it seemed that the multiple operations operating on the same data would benefit, until I found that np.clip() was the issue. I will revisit the batching there some other time. 

In my opinion, the benefit has to be large enough to warrant the addition code complexity / ugliness.

examples/VideoSpeedTest.py
pijyoi: I took a look at ```pg.gaussianFilter()```. I counted at least 4 temporary copies of the input data would be in-flight within that function. So for a 4096x4096 uint8 array, the initial random float64 data would be 128MB. While passing through ```pg.gaussianFilter()```, another 512MB would be allocated.
This would be rather surprising to the user who thought his array was only 16MB.

pijyoi:
I implemented ```rescaleData_blocked()``` and tested it on Windows (AMD octa-core) and Linux (Intel dual-core).
```VideoSpeedTest.py --size=1024x1024 --levels=1,254```

Variation 1: using ```umath.minimum()``` and ```umath.maximum()```
Variation 2: using ```umath.clip()```

a) ```rescaleData_blocked()``` using Variation 1 did not perform any better. (Windows nor Linux)
b) ```rescaleData_blocked()``` using Variation 2 did not perform any better on Windows but performed measurably better on Linux.
c) ```rescaleData()``` performed similarly whether using Variation 1 or Variation 2. (Windows and Linux)

The (b) observation might be due to the different CPUs used for Windows vs Linux.

```rescaleData_blocked()``` is not limited to ubyte images but it does assume that the data is row-major since it loops on the 1st dimension.

**Comment by j9ac9k** [comment]
> `rescaleData_blocked()` is not limited to ubyte images but it does assume that the data is row-major since it loops on the 1st dimension.

I've been thinking about this looping, would it be feasible to make use of `np.apply_along_axis(func1d, 1 if pg.getConfigOption('imageAxisOrder') == 'col-major' else 0)`

https://numpy.org/doc/stable/reference/generated/numpy.apply_along_axis.html#numpy.apply_along_axis

Sorry for the drive by comment, looking at the loops, and handling the transpose, I've always thought this might be a good candidate for `apply_along_axis` and potentially regain some vectorization.

**Comment by pijyoi** [comment]
Thanks @j9ac9k for the input! It reminded me of ```np.nditer```. Now the batching code doesn't look ugly any more and C and F order are handled without any effort. ```rescaleData_blocked``` could fully replace ```rescaleData```.

I noticed that the batching size used was 8192 elements, so seems like it is targeting L1 cache size instead.

**Comment by j9ac9k** [comment]
> Thanks @j9ac9k for the input! It reminded me of `np.nditer`. Now the batching code doesn't look ugly any more and C and F order are handled without any effort. `rescaleData_blocked` could fully replace `rescaleData`.
> 
> I noticed that the batching size used was 8192 elements, so seems like it is targeting L1 cache size instead.

oh wow, that is a lot cleaner, you're right.

File: pyqtgraph/functions.py (hunk 2 above)
Comment (outofculture) [reply pyqtgraph/functions.py:1340]: Thanks for explaining all that.

Batching might have a pretty substantial performance impact in this case. On a whim, I played with the batch size while processing a 3072x3072 test, and saw:

32-128k: 80fps
256-1024k: 90fps
2048k: 70fps
no batching: 50fps

So at least by that measure, you picked a pretty good magic number! I've only got 256KB L2 cache here, so I don't know how this performs under different conditions.

**Comment by outofculture** [comment]
I'm having trouble quantifying something that showed up while I was testing this (sorry I'm not a better QA person). When using PySide2 to process a 3072-square image, the _reported_ fps is high, but my _perceived_ fps is much, much lower, and inconsistent, at that. PyQt5 (90fps) has smoothly flickering static. Pyside2 (180fps) has a jerky, slow mess. This behavior goes away when I'm on master (20fps). This behavior goes away when I increase my image size to 4096-square (110fps) or decrease it to 512-square (1050fps). This behavior goes away when I use cupy (110fps). This is all conducted on linux, and if need be I can dump all relevant versions and hw stats here.

My one thought is that this is just an artifact of the framerate being higher than the actual refresh rate of my screen. That's not exactly corroborated by the other measurements, which are nearly all higher than 60fps. I still want to put some research into what happens when I vsync throttle this operation, but I wanted to alert you to this behavior, in the mean time.

We're also going to need to come up with a better way of measuring perceived fps, eventually.

**Comment by j9ac9k** [comment]
> I'm having trouble quantifying something that showed up while I was testing this (sorry I'm not a better QA person).

Is this repeatable?  I wouldn't mind seeing if I can give this a go on my local machines; I have a high refresh rate monitor for my laptop external display, so easy for me to see if things are different on my 60 fps display vs. my 144 hz display.

> We're also going to need to come up with a better way of measuring perceived fps, eventually.

Not sure if this would help; but I was experimenting w/ a rolling average for computing FPS using the `deque` module; I create a `deque`, set a max size (like 1,000), and keep appending to it.  Every ~50 frames (probably a better way of doing it) I calculate the average elapsed times, and update the FPS display with the new value.

You can see my implementation here:
https://github.com/pyqtgraph/pyqtgraph/pull/1610/files

**Comment by outofculture** [comment]
> Is this repeatable?

I don't have any development systems right now except for my linux laptop, so I don't know. I'm just running `python examples/VideoSpeedTest.py --size 3072x3072` for all these numbers. Play with sizes, and maybe aim for getting some of the same reported fps values I was seeing?

> I was experimenting w/ a rolling average for computing FPS using the `deque` module

That might help, but I'm worried we'll have to measure things from outside of python to get a definitive result.

**Comment by pijyoi** [comment]
I switched umath.clip() back to umath.{minimum(),maximum()} in ```rescaleData_blocked```. On another Windows machine, umath.clip() is slower (visibly lower frame rate value) despite comments on https://github.com/numpy/numpy/issues/14281 saying otherwise.

I think the part that adds a lot of variation to the timings from frame to frame is that ```rescaleData``` API returns a freshly allocated array.

**Comment by outofculture** [comment]
To investigate, I added an ugly throttle to `ImageItem.render`:

```python
        now = datetime.now()
        if self._lastRender is None or (now - self._lastRender).total_seconds() >= (1 / 60):
            self._lastRender = now
        else:
            sleep(1 / 120)
            return
```
With this code in place, perceived fps was smooth and fast.

I'm not entirely sure what to do with this information, though. We should probably try to throttle _some_ function or functions to happen not much more than vsync, regardless, as anything more than that is just wasting cpu time.

One other note, I also checked that using PySide2 on master with a 1024-square image does not produce the jittery slow speeds, despite running at about the same reported fps as the 3072-square gets on this branch. This would suggest that it's not PySide2 always misbehaving at certain framerates.

pijyoi:
A script just to time ```rescaleData``` vs ```rescaleData_blocked```
```python
import time
import numpy as np

from pyqtgraph import functions

shape = 3072, 3072
data = np.random.randint(0, 255, size=shape, dtype=np.uint8)
params = data, 1.1, 1, np.uint8
niters = 10

for scale in [1.1, 1.5]:
    params = data, scale, 1, np.uint8

    t0 = time.perf_counter()
    for _ in range(niters):
        output = functions.rescaleData(*params)
    t1 = time.perf_counter()
    print('A', scale, t1 - t0)

    t2 = time.perf_counter()
    for _ in range(niters):
        output = functions.rescaleData_blocked(*params)
    t3 = time.perf_counter()
    print('B', scale, t3 - t2)
```

I got:
```
A 1.1 0.7497713000000001
B 1.1 0.6815833000000002
A 1.5 1.0996170999999997
B 1.5 1.0091117
```
So ```rescaleData_blocked``` is ~10% faster than ```rescaleData```.
What was unexpected was that the clipping timing depends on the number of elements that actually got clipped.
(the larger scale value of 1.5 causes more elements to need clipping)

pijyoi:
After running the script on Ubuntu 20.04 and Windows 10, switching between using ```umath.clip``` vs ```umath.{minimum,maximum}```, these are the findings (both using numpy 1.20.1):

Using MINMAX on Windows or Linux yields the same observation that clipping timings depend on the number of elements that actually got clipped.

On Windows, using ```umath.clip``` yields _much worse_ performance than using MINMAX. Timings are similar to using np.clip.

On Linux, using ```umath.clip``` yields _much better_ performance than using MINMAX. In addition, clipping timings do _not_ depend on the number of elements that actually got clipped.

j9ac9k:
@pijyoi is the performance difference big enough that we should try to utilize the optimal method on each platform?  I can get some benchmarks on macOS.

pijyoi:
> @pijyoi is the performance difference big enough that we should try to utilize the optimal method on each platform? I can get some benchmarks on macOS.

The difference is significant. On Linux, for the scale 1.1, it was 2x. For the scale 1.5, it was 3x.

pijyoi:
I have added in the code to use ```umath.clip``` on Linux. @j9ac9k , if you can test for macOS, that would be great.

pijyoi:
I think I understand why clipping in NumPy is slow relative to the other operations within rescaleData().
We perform clipping on float data. Clipping then involves comparisons of floats. To be NaN correct, NumPy will precede every float comparison with a NaN check.
https://github.com/numpy/numpy/blob/main/numpy/core/src/umath/clip.c.src

On platform runtimes where NaN check is expensive, this will greatly slow down float comparison operations.
Note that NumPy int comparisons are much faster.

In normal C/C++ code, we usually "know" when the data doesn't have NaNs; however, there's no mechanism to let NumPy know that. E.g our source data was uint8 or uint16, thus there wouldn't  be NaNs.

j9ac9k:
> The difference is significant. On Linux, for the scale 1.1, it was 2x. For the scale 1.5, it was 3x.

In that case, we should probably add a `clip` method to `functions.py` that breaks up the code completion path based on platform.

> I have added in the code to use `umath.clip` on Linux. @j9ac9k , if you can test for macOS, that would be great.

I'll try and get to this later today.

j9ac9k:
@pijyoi  Running your benchmark code on macOS 11.2.3 with numpy 1.19.5 and with 1.20.1, 

with `shape = 3072, 3072` and `fast_umath_clip = sys.platform in ['linux']`



```
A 1.1 0.4938556839999999
B 1.1 0.446205808
A 1.5 0.46635472700000014
B 1.5 0.47485443699999985
```

On 1.20.1, A and B for 1.5 scale were effectively equivalent.


with `shape = 3072, 3072` and `fast_umath_clip = sys.platform in ['linux', 'darwin']`

```
A 1.1 0.506588347
B 1.1 0.5065373409999999
A 1.5 0.47844122499999986
B 1.5 0.8859278949999998
```

Let me know if I should be testing something a bit different.

pyqtgraph/functions.py
outofculture: Regarding memory concerns, I just noticed [these docs](https://doc.qt.io/archives/qtforpython-5.14/PySide2/QtGui/QImage.html#PySide2.QtGui.PySide2.QtGui.QImage.convertTo), which says it'll do the conversion in place. It works to replace the last line referenced above with:
```python
        qimg.convertTo(out_fmt)
```

Batching might still be desired here; my system still _reports_ a significantly higher fps when batching at 512k. Yet, I also just now observed that the PySide2 jitter and slowness goes away when batching is disabled. So maybe batching is bad somehow? Ugh, this is still confusing me.

pijyoi:
QImage.convertTo() is "in-place" shallowly only. A new buffer is allocated
and the QImage instance is internally pointed to it.

pijyoi:
>On 1.20.1, A and B for 1.5 scale were effectively equivalent.

So, when using umath.{minimum,maximum},
- Windows and Linux clip timings are affected by the number of elements that actually need clipping.
- macOS clip timings are unaffected.

>with shape = 3072, 3072 and fast_umath_clip = sys.platform in ['linux', 'darwin']
>A 1.1 0.506588347
>B 1.1 0.5065373409999999
>A 1.5 0.47844122499999986
>B 1.5 0.8859278949999998

Could I verify if the (approximately) same timings were obtained if you ran the benchmark a few times?

I have added a small-integer clipping path to rescaleData_blocked(). It runs faster for Windows. Could you please also try that on macOS?

j9ac9k:
> > On 1.20.1, A and B for 1.5 scale were effectively equivalent.
> 
> So, when using umath.{minimum,maximum},
> 
> * Windows and Linux clip timings are affected by the number of elements that actually need clipping.
> * macOS clip timings are unaffected.
> 
> > with shape = 3072, 3072 and fast_umath_clip = sys.platform in ['linux', 'darwin']
> > A 1.1 0.506588347
> > B 1.1 0.5065373409999999
> > A 1.5 0.47844122499999986
> > B 1.5 0.8859278949999998
> 
> Could I verify if the (approximately) same timings were obtained if you ran the benchmark a few times?

I pulled your branch, re-added `darwin` to the list of platforms to run the `fast_umath_clip` and got the following results:

```
~/Developer/pyqtgraph fastpath_argb*
pyqtgraph-pyqt5_515-py38 ❯ python perf.py
A 1.1 0.4658261450000001
B 1.1 0.467011818
A 1.5 0.43373908400000016
B 1.5 0.7629562230000002

~/Developer/pyqtgraph fastpath_argb*
pyqtgraph-pyqt5_515-py38 ❯ python perf.py
A 1.1 0.45256862400000003
B 1.1 0.42464828099999996
A 1.5 0.42674245099999997
B 1.5 0.711061419

~/Developer/pyqtgraph fastpath_argb*
pyqtgraph-pyqt5_515-py38 ❯ python perf.py
A 1.1 0.46632524099999995
B 1.1 0.428802864
A 1.5 0.430931647
B 1.5 0.7163260119999999

~/Developer/pyqtgraph fastpath_argb*
pyqtgraph-pyqt5_515-py38 ❯
```

removing `darwin` from the list 🔥 

```
~/Developer/pyqtgraph fastpath_argb*
pyqtgraph-pyqt5_515-py38 ❯ python perf.py
A 1.1 0.487717837
B 1.1 0.192059957
A 1.5 0.4247874890000001
B 1.5 0.2040620019999999

~/Developer/pyqtgraph fastpath_argb*
pyqtgraph-pyqt5_515-py38 ❯ python perf.py
A 1.1 0.449667276
B 1.1 0.191115272
A 1.5 0.42235494299999987
B 1.5 0.19903449500000003

~/Developer/pyqtgraph fastpath_argb*
pyqtgraph-pyqt5_515-py38 ❯ python perf.py
A 1.1 0.473250743
B 1.1 0.18906075700000002
A 1.5 0.419320012
B 1.5 0.19165177300000003
```

...that's quite the improvement.

pijyoi:
To summarize: (hope I didn't get it wrong!)

|             |Windows|Linux|macOS|
|--|----------|------|-------|
|minmax  | varying | varying | constant | 
|nditer + minmax | varying | varying |  constant |
|nditer + ncu.clip | varying SLOW! | constant FAST! | varying slow|

Varying : means timing increases with the number of elements that actually get clipped. (i.e. scale 1.5 slower than scale 1.1)
Constant : scale 1.5 same speed as scale 1.1

j9ac9k:
> varying slow

That is an accurate characterization (that table does not reflect the `fits_int32` code-path you wrote).

Let me know if there is other conditions I should test (including different numpy versions and so on).

pijyoi:
I compared "nditer + fits_int32" vs "nditer + ncu.clip" on Linux and the latter is slightly faster.

So I would say that "nditer + minmax" is a safe replacement for "minmax". If nothing else, it at least removes memory allocation of the float32 / float64 buffer.
i.e. the current codepath ordering of
1) fast_umath_clip
2) fits_int32
3) nditer + minmax
seems roughly optimal, or at least doesn't penalize any platform.

Given the above, I am not so sure whether fast_umath_clip codepath should be left in. (Likely to be numpy version and compiler dependent)

j9ac9k:
> Given the above, I am not so sure whether fast_umath_clip codepath should be left in. (Likely to be numpy version and compiler dependent)

Normally I'm all for taking the performance hit and having a simpler variant of the code, but this sounds like a real performance critical codepath; I think if we comment/reference the upstream issues we can leave it in; unless there is a common linux platform that shows a performance hit as well.

That said, you're the one that's put the work into this, so I'll support whatever you think is best.

pijyoi:
CI testing fails for all macOS when using
```python
np.core.umath.clip(y, clip[0], clip[1], out=y, casting='unsafe', dtype=np.int32)
```
instead of
```python
np.core.umath.clip(y.astype(np.int32), clip[0], clip[1], out=y)
```
The resultant array is all -1

pijyoi:
This is what I think is the numpy clipping code (cobbled together from a few files)
```c
#include <math.h>

#ifdef _MSC_VER
#define npy_isnan(x) isnan(x)
#else
#define npy_isnan(x) __builtin_isnan(x)
#endif

#define PyArray_MIN(a, b) ((a) < (b)) ? (a) : (b)
#define PyArray_MAX(a, b) ((a) > (b)) ? (a) : (b)

#define _NPY_float_MIN(a, b) (npy_isnan(a) ? (a) : PyArray_MIN(a, b))
#define _NPY_float_MAX(a, b) (npy_isnan(a) ? (a) : PyArray_MAX(a, b))

#define _NPY_CLIP(x, min, max) \
    _NPY_float_MIN(_NPY_float_MAX((x), (min)), (max))

float maximum(float x, float y) {
    return _NPY_float_MAX(x, y);
}

float clip(float num, float vmin, float vmax) {
    return _NPY_CLIP(num, vmin, vmax);
}
```

You can see the compiled output here:
GCC x86_64 https://godbolt.org/z/WWr8r1
CLANG x86_64 https://godbolt.org/z/Gn68s9
MSVC x86_64 https://godbolt.org/z/Erszs9

GCC generated the most compact code, by far. This probably explains why Linux "platform" has a fast umath.clip().

pijyoi:
To pre-empt some questions:
As all this code is new and not battle-tested in the wild, I am keeping both rescaleData() and rescaleData_blocked().

@j9ac9k , I saw the comment in #1642 that you were waiting to release on this PR. I do hope that this gets some real-world testing before that. Don't want to be the guy that breaks people's applications. :wink:

j9ac9k:
> @j9ac9k , I saw the comment in #1642 that you were waiting to release on this PR. I do hope that this gets some real-world testing before that. Don't want to be the guy that breaks people's applications. 😉

Having been a person that broke code, and rolled a version, let me assure you, it's not so bad 😅  If you would rather _not_ have this PR in the next version, I'll respect that, and won't wait on this guy being ready to merge before the next release.  Just know if it was broken, our outstanding PR count is small enough I would be comfortable releasing a 0.12.1 in short order if it fixed a critical regression.

We could re-enable the `useExperimentalFeatures` to toggle `rescaleData_blocked()` too if you'd want.

pijyoi:
I think the removed Qt4 and added Qt6 support should be enough changes for
0.12.0. I expect some issues to crop up from there.

Not in favour of useExperimentalFeatures. The combination of useOpenGL and
useExperimentalFeatures already doesn't work on Windows since migration to
QOpenGLWidget.

pijyoi:
I have removed ```rescaleData_blocked()``` from this PR as it needs more testing.

pyqtgraph/functions.py
```diff
@@ -1269,6 +1274,50 @@ def makeARGB(data, lut=None, levels=None, scale=None, useRGBA=False, output=None
...
+
+    if in_fmt == out_fmt:
+        aout[:] = ain
+        return True
+
+    npixels_chunk = 512*1024
+    batch = int(npixels_chunk / ncols / nchans)
+    batch = max(1, batch)
+    row_beg = 0
+    while row_beg < nrows:
```
j9ac9k, line 1309: can't help but look at this and think of your implementation of `np.nditer` simplifying this a fair amount.  Is there a use-case here for that?

pijyoi: nditer is meant for cases where each element in the uint8 array can be processed independently. This is true for nchans==1 and could perhaps be done for nchans==4 (RGBA) by viewing the array as uint32. I don't think it can be done for the case nchans==3 (RGB).

j9ac9k:
This LGTM, @outofculture you have anything to add?

outofculture:
I still can't explain the jittery, slow behavior I see only on this branch, but am I the only person seeing that? If I am the only system experiencing it, and it's only for images of a certain size, it starts to feel like it's just a weird problem on my system. My fear is that this problem will be more widespread for our users, or that it will pop up later and be hard to track back here. Long-term, we need to come up with benchmarks that measure the actual visible framerate, and benchmarks that can capture and quantify "jittery". I'll work on those, and hopefully at some later time we can measure this problem, bisect to point at its cause, and address it then.

Otherwise, this branch is excellent optimizations.

j9ac9k:
I have replicated the issue @outofculture is describing on macOS/pyside 5.15.2/python3.8 (will test more configurations shortly).

The best way I can describe it is there are effectively two images; and one of the two images is shown the bulk of the time, but occasionally the other image is shown momentarily as a flicker.

pijyoi:
I have added a commit to #1648 that draws the white arrows at different positions for different frames. This makes the effect easier to see.

From my testing on Windows, the culprit seems to be related to disabling vsync in #1599.
```python
if RawImageGLWidget is not None:
    # don't limit frame rate to vsync
    sfmt = QtGui.QSurfaceFormat()
    sfmt.setSwapInterval(0)
    QtGui.QSurfaceFormat.setDefaultFormat(sfmt)
```

j9ac9k:
I commented out the vsync code, and added the code from the arrows, but it largely looks the same.  Here is another video in case it helps.

https://user-images.githubusercontent.com/646398/112173348-b7431b00-8bb2-11eb-9289-01dbcb15abc9.mov


Not sure it helps, but on the current `master` branch, I took the change you made with the arrows; and I commented out this bit:

```python
# if RawImageGLWidget is not None:
#     # don't limit frame rate to vsync
#     sfmt = QtGui.QSurfaceFormat()
#     sfmt.setSwapInterval(0)
#     QtGui.QSurfaceFormat.setDefaultFormat(sfmt)
```

and reran the test and got this result

https://user-images.githubusercontent.com/646398/112174832-f6be3700-8bb3-11eb-9193-54318cf45908.mov

pijyoi:
Some remarks:

The disable vsync code only kicks in when OpenGL is enabled.

On the fastpath_argb branch, you are running with OpenGL disabled. (i.e. vsync is left enabled by default anyway) 
You are getting 70 fps. The frame rate is high enough that multiple arrows seem to appear on screen at the same time.

On the master branch, you are running with OpenGL enabled. (vsync is enabled because you commented out the code)
You are getting 13 fps. The frame rate is slow enough that each different frame is distinct. (Only one arrow appears at a time.)
What happens if you were to reduce the size of the frame here such that the frame rate increases? Does the effect appear on master branch then?

j9ac9k:
Hi @pijyoi sorry I was tinkering around with a bunch of variations; and unintentionally added far more variables into the equation here, last thing I meant to do was to add to the noise.

Just so I give helpful information, what configurations are you wanting me to test?  

Also testing configurations with OpenGL is a bit tricky on macOS Big Sur (what I'm running) it's doable but I have to be running effectively a brand new python version (3.9.1+).

If I understand your comment right, you'd like me to report the fps on the master branch, with openGL enabled, and the vsync code commented out, with smaller image sizes?

pijyoi:
On Windows, just having a RawGLImageWidget instantiated (i.e. be selectable in the VideoSpeedTest GUI but not actually selected) affects the frame rate. On Windows, this slows down the frame rate for the other 2 widgets. This is true whether vsync is disabled or not. With vsync on, this caps all 3 widgets to the monitor refresh rate.
This doesn't seem to happen on Linux. I don't know what to expect on macOS.

It would be good to establish if the effect is already present on master branch. 
I believe that fastpath_argb tkjust makes it more visible by:
1) not using the gaussian filter
2) increasing the frame rate 

So using a lower image size is just a means to increase the frame rate on master branch. To simplify things, you could just disable OpenGL. The fps value isn't that important, the question is whether the jitter (?) effect already exists on master branch.

>If I understand your comment right, you'd like me to report the fps on the master branch, with openGL enabled, and the vsync code commented out, with smaller image sizes?

OpenGL disabled. Vsync code has no effect. Smaller image size.

j9ac9k:
> OpenGL disabled. Vsync code has no effect. Smaller image size.

`master`

image


`fastpath_argb`

image

j9ac9k:
I was just looking at this a bit more; and maybe the gitter effect isn't as noticeable on the master branch due to the blurring (which I assume is the gaussian filter at work?)

outofculture:
I haven't been able to reproduce this effect on master. I wouldn't trust that finding completely, though;  I want a way to programmatically detect this behavior, so I can run bisects and try extensive combinations of image sizes and options. I haven't been able to come up with a way to do that, though.

>     1. not using the gaussian filter

Yeah, I've been commenting the gaussian filter out on master when I run tests there.

>     2. increasing the frame rate

I made attempts with smaller image sizes on master, such that the _reported_ framerates were the same (~150fps) as when I experienced this behavior on fastpath_argb. Really small images (256x256) on master produce as similar effect, but not nearly as consistent or dramatic. It's also reporting much higher fps (1240ish), so it's enough to feel like anything that test shows could be for something else entirely. 

Maybe it only happens on large-ish images if the fps is also high?

pijyoi:
I added https://github.com/pyqtgraph/pyqtgraph/pull/1648/commits/d7ca338ebf861a59abd0230f30c28c9bc05af65a to #1648 that draws a moving white line that should be easier to visualize.

j9ac9k:
@pijyoi I can see the white line flickering and moving up

`rescale_blocked` branch  with 1024x1024 image

https://user-images.githubusercontent.com/646398/112250368-7c72ce80-8c16-11eb-894e-48aa9718ac8b.mov


`rescale_blocked` branch with 3072x3072 image

https://user-images.githubusercontent.com/646398/112250449-a0361480-8c16-11eb-86db-ca3c10c8f425.mov

NilsNemitz:
I guess the line would be flickering because you have more pixels in the image than on the screen.
I think the question is if 
- the line moves smoothly
- the line ever jumps back and forth 
- the line crosses the picture in a time that matches resolution and FPS.

pijyoi:
>maybe make a wider white bar so that the (nearest neighbor?) sampling doesn't cause on/off flickering ?

Another way that doesn't require code change may be to use non-square images, say 10000x512. (I found out that the .ui file limits the maximum value to 10000)

outofculture:
Okay, I have finally reproduced this behavior on master! It took using non-square images, weirdly. 7501x128 on my system (on master, using pyside2) shows the jitter. 128x7501 _does not_. Gah. This is probably an important clue about the ultimate cause. There's also a 3x difference in reported fps between those two scenarios, so the jitter might still just be tied to fps.

Well, whatever, I'm willing to take this as sufficient evidence that the problem exists regardless of the changes in this branch. Sorry to have held up an otherwise excellent PR.

pijyoi:
Master branch's VideoSpeedTest leaves image axis order as default of column major. This results in cases where the processing source buffer is F-contiguous while the destination buffer is C-contiguous. This really exercises the memory system and causes seemingly hard to explain variations in timings.

j9ac9k:
If we now have confirmation that this PR doesn't cause new breakage, is this ready to merge?

pijyoi:
Yes, ready to merge.

outofculture (approved):

j9ac9k:
@pijyoi I know I'm sounding like a broken record, but thanks for working on this.  As you know, makeARGB is a bottleneck for a lot of folks; so now doubt you have made a positive change for the better here.

Merging changes!

PR #1618, opened 2021-03-03
PyQt6 install sys.excepthook

pijyoi:
https://github.com/pyqtgraph/pyqtgraph/pull/1605#issuecomment-788784635
https://github.com/pijyoi/pyqtgraph/runs/1853387168?check_suite_focus=true

The PyQt6 run failures in the link above are all:
```
AttributeError: 'NoneType' object has no attribute 'color'
Fatal Python error: Aborted
```

Disregarding why the slot is being executed in the first place, the errors show that PyQt6, like PyQt5, aborts the process upon an uncaught Exception inside a slot. This PR installs the PyQt5 sys.excepthook also for PyQt6.

The other commits take the opportunity to remove more Qt4 from Qt.py.

### DIFF
--- a/pyqtgraph/Qt.py
+++ b/pyqtgraph/Qt.py
@@ -11,7 +11,6 @@ This module exists to smooth out some of the differences between PySide and PyQt
 """
 
 import os, sys, re, time, subprocess, warnings
-import importlib
 import enum
 
 from .python2_3 import asUnicode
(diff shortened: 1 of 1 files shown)


pijyoi:
It would also be good to make a decision on whether installing sys.excepthook for the user is desirable.

j9ac9k:
Hi @pijyoi thanks for implementing a fix.  Given we've had the excepthook installed all this time, and as far as I'm aware, we've heard no complaints about it, I'm fine keeping it (and thus utilizing it for PyQt6); but I am afraid I don't know why we would not want to install the excepthook.  What's the downside?

pijyoi:
simple script to test this PR
```python
from pyqtgraph.Qt import QtCore, QtWidgets

def callback():
	raise Exception("problem")

qapp = QtWidgets.QApplication([])
QtCore.QTimer.singleShot(500, callback)
QtCore.QTimer.singleShot(1000, QtWidgets.QApplication.quit)
qapp.exec_()
print("normal loop exit")
```

pijyoi:
>but I am afraid I don't know why we would not want to install the excepthook. What's the downside?

It always about whether a library should be doing something to the global Python process.
Just to make sure that we are good about it.

j9ac9k:
Searching for excepthook, I see nothing in the issue tracker, I have no problem with this.

j9ac9k:
Merging, thanks @pijyoi 

If you would like maintainer status of this project, we'd love to have you.  Feel free to email me at the email address I have attached to my github account.

Thanks again for the PR!

PR #1619, opened 2021-03-04
Stop PlotDataItem from always sending full style information to PlotCurveItem / ScatterPlotItem

NilsNemitz:
I noticed that PlotDataItem seems to *always* send full styling information to PlotCurveItem and ScatterPlotItem.

That seems unnecessary, and this PR disables these updates when only the data is updated, i.e. the `PlotDataItem.setData()` does not contain any pen, brush etc. information. 

A more sophisticated solution would distinguish between style updates directed at the PlotCurveItem and the ScatterPlotItem. But this simple version should reduce overhead for the common use case of setting the styles once and then repeatedly sending numerical data only.


j9ac9k:
Thanks @NilsNemitz this LGTM, merging!

Issue #314, opened 2016-04-18
error reporting in multiprocessing


eendebakpt:
In  `remoteproxy.py` the function `getResult` evaluates the result of a remote call. If the call resulted in an error, this is printed to stdout. In one of my applications the remote proxy triggers an error, but this is not a problem since the application itself handles the error. How can I prevent pyqtgraph from printing the error.

Note: what happens is that my IDE (Spyder) tries to execute some remote `getattr` functions for attributes that do not exist such as `_repr_pdf_` or `_repr_html_`

PR #1625, opened 2021-03-07
Named colors for dynamically applied palettes

NilsNemitz:
pyqtgraph currently does not handle changes to the overall palette of colors really well (#1407 ).
#1588 points out that there is a useful paletteChange() signal that would allow an application to react to a switch between light and dark mode. However, any color changes beyond the simple background/foreground options would need to be applied manually.

This PR attempts to simplify this.
The example user-side code (in `examples/PaletteApplicationExample.py`) demonstrates a full palette replacement by a single line command like `pg.palette.get('monogreen').apply()`

This can be applied at any time, and automatically updates the displayed plots, even if no new data is sent.

20210308 color variations

The implementation exploits that basically all color requests go through functions `mkColor()` / `mkPen()` / `mkBrush()`. These can now return `NamedPen` or `NamedBrush` objects, subclassed from `QPen` and `QBrush`. These are assigned functional names such as 'gr_fg' for the graphical foreground color, or 'p0' to 'pB' for a set of twelve predefined plot colors. Such color names can be used directly in the `pen=` or `brush=` keyword arguments of the plotting functions.

When a new palette is assigned, `NamedPen` / `NamedBrush` objects automatically update the super-class `QPen` / `QBrush` object according to the new palette. As the named objects otherwise function in the same way, no large changes to the codebase are required. However, the subclasses of `GraphicsObject`, `GraphicsWidget` and `GraphicsView` now have a function `styleHasChanged()`, which is called on palette updates and can be used to perform additional tasks only when required. For example, `ScatterPlotItem`'s SymbolAtlas needs to be cleared to rebuild in the new color set.

This PR demonstrates changes for most functionality in PlotCurveItem and ScatterPlotItem. If this method is adopted to improve color handling, there should be no major problems with including the remaining items. Some more cleanup is required.

I do not see any noticable impact on performance, since only minimal overhead is incurred from working with subclasses of QPen / QBrush. No additional code is executed except during the actual palette change.

The included palettes need some improvement, and would ultimately be user-definable and probably loaded from disk on request.

Comments and ideas are very welcome!


pyqtgraph/functions.py
```diff
@@ -293,75 +343,115 @@ def mkColor(*args):
...
-    if 'color' in kwds:
-        color = kwds['color']
-    elif len(args) == 1:
-        arg = args[0]
-        if arg is None:
-            return QtGui.QBrush(QtCore.Qt.NoBrush)
-        elif isinstance(arg, QtGui.QBrush):
-            return QtGui.QBrush(arg)
+    while ( # unravel single element sublists
+        ( isinstance(args, tuple) or isinstance(args,list) )
```
j9ac9k, line 353: you can do `isinstance(args, (tuple, list))` here

pyqtgraph/functions.py
```diff
@@ -293,75 +343,115 @@ def mkColor(*args):
...
-    if color is None:
-        color = mkColor('l')
-    if hsv is not None:
-        color = hsvColor(*hsv)
-    else:
-        color = mkColor(color)
-        
-    pen = QtGui.QPen(QtGui.QBrush(color), width)
-    pen.setCosmetic(cosmetic)
+    assert qpen is not None
```
j9ac9k, line 448: probably don't want an `assert` in functional code; as they can be squashed/suppressed when you run python with the `-O` argument (which I would expect a fair chunk of our user-base to actually do).

Probably to something like:

```python
if qpen is None:
    raise RuntimeError("some message")
```

pyqtgraph/graphicsItems/GraphicsObject.py
```diff
@@ -2,8 +2,16 @@
...
 from .GraphicsItem import GraphicsItem
+from .. import functions as fn
 
 __all__ = ['GraphicsObject']
+DEBUG = False
+
+try: # prepare common definition for slot decorator across PyQt / Pyside:
+    QT_CORE_SLOT = QtCore.pyqtSlot
+except AttributeError:
+    QT_CORE_SLOT = QtCore.Slot
```
j9ac9k, line 13: This should probably be put in `Qt.py` and handled the same way we handle `QtCore.Signal`

pyqtgraph/namedColorManager.py
```diff
@@ -0,0 +1,86 @@
...
+    DEFAULT_COLORS[key] = DEFAULT_COLORS[col]
+
+for idx, col in enumerate( ( # twelve predefined plot colors
+    'l','y','r','m','b','c','g','d','d','d','d','d'
+) ): 
+    key = 'p{:X}'.format(idx)
+    DEFAULT_COLORS[key] = DEFAULT_COLORS[col]
+    
+# define and instantiate a SignalSource object to pass signals to all pyqtgraph elements
+class NamedColorManager(QtCore.QObject): # this needs to emit QEvents
```
j9ac9k, line 39: QEvents or Signals?

NilsNemitz:
Dear @j9ac9k , thank you for the comments. I have updated the code accordingly.

The NamedPen and NamedBrush objects now take a manual reference to the NamedColorManager on init, that should remove the main cyclic imports.
I have added a compatibility function to intercept foreground / background setting through the legacy `setConfigOption()` method, and apply them directly to the palette.

pyqtgraph/functions.py
NilsNemitz: 👍 done

pyqtgraph/functions.py
NilsNemitz: `qpen` being None is really not supposed to ever happen here; it means I missed a possible execution path. This is an accidentally surviving debug statement.
But I guess throwing a recognizable error would be more helpful. Replaced by
` if qpen is None:`
`  raise ValueError('Failed to construct QPen from arguments '+str(args)+','+str(kargs) )`

pyqtgraph/graphicsItems/GraphicsObject.py
NilsNemitz: 👍 done
I didn't see that before.
Now `QtCore.Slot` is a library wide alias for `QtCore.PyQtSlot` when running on PyQt; the same way as `QtCore.Signal` is.

pyqtgraph/namedColorManager.py
NilsNemitz: QSignal. Oops. Rewrote the comment a little.

examples/PaletteApplicationExample.py
```diff
@@ -0,0 +1,162 @@
...
+            self.fps = self.fps * (1-s) + (1.0/dt) * s
+        self.plt.setTitle('%0.2f fps' % self.fps)
+        QtWidgets.QApplication.processEvents()  ## force complete redraw for every plot
+
+mkQApp("Palette test application")
+main_window = MainWindow()
+
+## Start Qt event loop
+if __name__ == '__main__':
+    QtWidgets.QApplication.instance().exec_()
```
j9ac9k, line 162: I think we updated these since you created this PR, but put this line here instead: `pg.mkQApp().exec_()` (this handles the case there is no `QApplication` instance, which I know you declare earlier, but makes it easier)

pyqtgraph/__init__.py
```diff
@@ -62,21 +62,26 @@
...
-def setConfigOption(opt, value):
-    if opt not in CONFIG_OPTIONS:
-        raise KeyError('Unknown configuration option "%s"' % opt)
-    if opt == 'imageAxisOrder' and value not in ('row-major', 'col-major'):
-        raise ValueError('imageAxisOrder must be either "row-major" or "col-major"')
-    CONFIG_OPTIONS[opt] = value
-
-def setConfigOptions(**opts):
-    """Set global configuration options. 
+# def setConfigOption(opt, value):
```
j9ac9k, line 65: Why is all this commented out?

j9ac9k: scroll down a little.... n/m that's why  https://i.redd.it/72lxlzwn81k61.jpg

pyqtgraph/colormap.py
```diff
@@ -176,7 +178,40 @@ def _get_from_colorcet(name):
         color=color_list) #, names=color_names)
     _mapCache[name] = cm
     return cm
-
+    
+def make_monochrome(color='green'):
```
j9ac9k, line 182: convert to `camelCase` please

pyqtgraph/functions.py
```diff
@@ -293,75 +345,115 @@ def mkColor(*args):
...
-        elif isinstance(arg, QtGui.QBrush):
-            return QtGui.QBrush(arg)
+    while ( # unravel single element sublists
+        isinstance(args, (tuple, list) )
+        and len(args) == 1 
+    ): 
+        args = args[0]
+    # now args is either a non-list entity, or a multi-element tuple
+    # short-circuits:
+    if isinstance(args, NamedBrush):
```
j9ac9k, line 361: no code change requested here, just thinking of how nice the pattern matching functionality in python 3.10 will be able to serve us, ...too bad we won't get access to it for another ~4 years in this library.

pyqtgraph/namedBrush.py
```diff
@@ -0,0 +1,82 @@
...
+        if isinstance(other, self.__class__):
+            return self._identifier == other._identifier
+        else:
+            return False
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+        
+    def __hash__(self):
+        return id(self)
```
j9ac9k, line 36: We got burned _really_ bad by this and had to fix that here: https://github.com/pyqtgraph/pyqtgraph/pull/1564

At the end of the day, QBrushs/QPens are mutable, and implementing `__hash__` in such a way is our attempt to make them not mutable, which I think we're going to run into issues in general if we try and enforce that criteria.

pyqtgraph/namedBrush.py
```diff
@@ -0,0 +1,82 @@
...
+        try:
+            qcol = color_dict[name]
+        except ValueError as exc:
+            raise ValueError("Color '{:s}' is not in list of defined colors".format(str(name)) ) from exc
+        if alpha is not None:
+            qcol.setAlpha( alpha )
+        if DEBUG: print('  NamedBrush '+name+' updated to QColor ('+str(qcol.name())+', alpha='+str(alpha)+')')
+        super().setColor( qcol )
+        
+    def paletteChange(self, color_dict):
```
j9ac9k, line 74: named very similarly to the Signal `paletteChanged`, could we overwrite `setPalette` ... but it's not really a palette object coming in, maybe `updatePalette`?

pyqtgraph/namedPen.py
```diff
@@ -0,0 +1,99 @@
...
+        if isinstance(other, self.__class__):
+            return self._identifier == other._identifier
+        else:
+            return False
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __hash__(self):
+        return id(self)
```
j9ac9k, line 40: see earlier comment about using the `id()` for `__hash__`

j9ac9k:
Hi @NilsNemitz  Sorry it's taken me a while to follow up on this.  We've been busy 😈 

First, I want to reiterate, this looks amazing, and I believe every effort getting this into pyqtgraph is very worthwhile.

In terms of code suggestion changes, I put most of my comments in-line, but I also would suggest changing some of the names of the signals and slots to be more in line with the Qt API in general, which we try and follow:

In this case this would mean:

* `paletteHasChangedSignal` -> `paletteChanged`
* `styleHasChanged` -> I'm not sure on this one, usually "changed" in the Qt API refers to a signal, but in this case I would think something like `updateStyle` would be appropriate.   Looking into the method, they already have 

One of the things that is bugging me a bit is it seems like we (and this isn't a we as in you, a few of us have fallen into this trap) are trying to treat the `QPen`, `QBrush` and `QColor` objects as if they're immutable.  This is definitely _not_ the case; the relevant objects can have their properties (and even colors) change.  These items are not hashable, and any attempt to make them so will likely cause is trouble.  I wish I could give more guidance here.

You have looked into the color-space here more than most of us by this point @NilsNemitz; do you see a solution of updating the existing QPen/QBrush/QColor values with new values to reflect the new colors we want to display less feasible than creating new instances of those objects?

pyqtgraph/namedBrush.py
NilsNemitz: This is only implemented here to allow the NamedColorManager to keep a list of weak references to all defined NamedColor objects. When a NamedColor object is deleted, the reference should be purged from he list immediately. A new object created with the same id would not be part of the list (or set). 
But as discussed on Slack, this logic might be more straightforward to implement as a WeakValueDictionary, with id as the key. That should avoid providing a weak, limited-use hash function here.

pyqtgraph/namedBrush.py
NilsNemitz: I am undecided on this. This takes the place of what was originally *intended* to be a signal alerting all NamedColor objects of a palette change. Unfortunately QColor does not handle signals. I like that the naming still seems to indicate this function.

NilsNemitz:
Dear @j9ac9k, thank you very much for the review.
I am still working on this, and I'll do my best to include your suggestions in the next update!

Is there some Qt convention on names for functions that handle signals? `styleHasChanged()` is named that because it gest connected to the `paletteHasChangedSignal()` (but is slightly more general in function). As this would be added to many or most of the existing graphics items, I'd like to have a name that communicates this function as clearly as possible. Ideas?

updateStyle might work, but this is very close to the names of functions that are already in use, I think.

NilsNemitz:
"trying to treat the QPen, QBrush and QColor objects as if they're immutable. "

I was originally hoping to use that feature to implement adjustable colors. However, that boat has already sailed: 
Overall, PyQtGraph treats colors as fixed, seen most strongly in `functions.mkColor()`, which returns a *copy* of any QColor it is passed directly. 

This treatment is supported by the behavior of the Python Qt bindings (or at least by PyQt5, which I am using): In the Qt c code, a QColor object can be passed to create e.g. a QPen object. This QColor object is kept intact, and the same object is returned by the getter function. So in principle, changing the original QColor object *should* affect the derived QPen and all colors that reference this QColor or QPen. In PyQt, this connection does not exist, as the getter function returns a copy of the original object (which I *suspect* is created on initializing the QPen).

I don't think the user can have any expectation of assigning a QColor and then achieving a useful response to changing its properties. Adding a NamedColor would provide precisely this functionality in a way that could be supported throughout the library. 

The NamedColor / NamedPen / NamedBrush code is the thinnest wrapper I could come up with to extend the functionality of QColor / QPen / QBrush to locate and update them after their initial creation. If someone has a good idea how to make this even thinner, I'd be happy to pursue that.

If we can agree to ask the user to treat colors as immutable, then only one `NamedColor` object needs to exist for each (named) color and opacity combination, and locating the pre-existing objects for re-use would be done transparently by `functions.mkColor()`.

We could consider disabling the related functions of `NamedColor`, but in the spirit of "we are all adults here", I am not sure if that is necessary, and it would add a lot of boilerplate code to block all the many methods that QColor provides. If we do this, then it might also be possible to simply disconnect such a modified NamedColor from future updates, which would revert it to the original color.

j9ac9k:
> Dear @j9ac9k, thank you very much for the review.
> I am still working on this, and I'll do my best to include your suggestions in the next update!
> 
> Is there some Qt convention on names for functions that handle signals? `styleHasChanged()` is named that because it gest connected to the `paletteHasChangedSignal()` (but is slightly more general in function). As this would be added to many or most of the existing graphics items, I'd like to have a name that communicates this function as clearly as possible. Ideas?
> 
> updateStyle might work, but this is very close to the names of functions that are already in use, I think.

Yeah, this one is just via approximating; what I generally do is I look at the most equivalent/nearest Qt object, look at all the methods/signals and try to use a name that is in line with those.  Here is a list of all the methods for QGraphicsItem: https://doc.qt.io/qt-5/qgraphicsitem-members.html

Signals are a bit tougher to track down, but I just know that `somethingChanged` is a pretty standard naming convention for Signals in Qt in general.  PyQtGraph often follows the convention `sigSomethingChanged` as well.  Collectively we try and follow Qt conventions whenever possible, but as there is no spec, it requires speculation. 

> I was originally hoping to use that feature to implement adjustable colors. However, that boat has already sailed:
Overall, PyQtGraph treats colors as fixed, seen most strongly in functions.mkColor(), which returns a copy of any QColor it is passed directly.

Blerg, I got myself crossed up with your implementation of `__hash__`, while the library uses colors as fixed, we have the ability to change that behavior in time.   For the perspective of this PR, this may be a non-issue, as you do handle dynamic colors.  Regarding the rest of the library, we have the capability of change that behavior given time/effort.

> If we can agree to ask the user to treat colors as immutable,...

I think this is a pretty tall order; and is in conflict with the upstream framework, and may be in conflict with an application that pyqtgraph is being integrated into.  


Also, there is some president for making wrappers for Qt objects like this in PyQtGraph; `Point` comes to mind: https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/Point.py  Our normal convention for this wrappers is just to drop the `Q`; I'm not suggesting you go through the effort of changing `NamedColor` to `Color` yet, but after we settle on the API/philosophy, that is like a change that should be made before merging (unless we can think of a reason to keep the names as is).

j9ac9k:
Hi @NilsNemitz 

I gave this one a second look-over and I don't have any real concerns other than some naming mechanisms to get more in-line with the Qt API, which I can certainly help with.  I am good with merging this once those are straightened out.

This is a big feature; and I suspect there are likely unintentional side-effects along the way; but I'm willing to sort those out as we go.

We may need to add some test code for this (test changing the global config option, check the colors of existing plots, that  sort of thing).

NilsNemitz:
Dear @j9ac9k ,
it took me a while to revise this, but I hope this has now reached the point where it could use a fresh look, and maybe some attempts to break it.
I have changed the implementation so that it no longer works with subclasses of QPen and QBrush. Instead, any color object (QColor, QPen, QBrush) that is requested through `pg.functions` by its PyQtGraph name is added to a central color registry.

This means that the color objects are perfectly ordinary Qt objects. But when a new palette is applied, the registry will mutate all registered color objects to match the new color scheme. Registered objects are marked by a monkey-patched "registration" attribute. This would have been a nice application for a Qt property... but QColors etc. unfortunately are not QObjects.

If the user wants to keep a version that is not auto-mutated, they can simply make a copy. The copy does not include the patched attribute.

I have added a new demonstration (and palette design tool) to the examples, called `PaletteTester`.
This also shows the palettes that are now included, including dark mode / light mode palettes that should hopefully be pretty universal, while avoiding the EGA-style extreme default colors (cyan, magenta...) There are also some fun, maybe not so useful palettes, including a generator for any color of pseudo-monochrome.

NilsNemitz:
If there is no major flaw in functionality and interactions, I hope we can discuss naming changes more seriously now.

Things that still need to be done (after pinning down the main design)
- write documentation
- most palettes now use a ColorMap to define a set of plot colors, this should probably be integrated with the existing intColor definitions
- palettes now include a set of ten colors that define a monochrome ramp to handle warmer / cooler grays this should probably also be represented by a ColorMap for convenience.
- Some GraphicsItems need some minor modifications to properly handle color updates. Currently these changes are mainly implemented for PlotDataItem, TextItem and AxisItem. But other items should already display correctly when drawn after the palettes is assigned.

pyqtgraph/graphicsItems/TextItem.py
```diff
@@ -66,6 +68,7 @@ def setText(self, text, color=None):
         This method sets the plain text of the item; see also setHtml().
         """
         if color is not None:
+            self.color = color
```
j9ac9k, line 71: if `setColor` is going to be a method, then `color` should not be a property, but be a getter (to line up with the Qt API more).  Property can probably be `self._color`?

pyqtgraph/colorRegistry.py
```diff
@@ -0,0 +1,363 @@
...
+            for key in DEFAULT_COLORS:
+                if key not in colors:
+                    raise ValueError("Palette definition is missing '"+str(key)+"'")
+            if DEBUG: print('  ColorRegistry: Setting palette, all color definitions are present.')
+            self.color_dic.clear()
+            self.color_dic.update(colors)
+
+        # notifies named color objects of new assignments:
+        for key in self.registered_objects:
+            ref, desc = self.registered_objects[key]
```
j9ac9k, line 351: can probably just loop over the values directly here?

```python
for ref, desc in self.registered_objects.values():
    ...
```

NilsNemitz: Never thought about that before. Saves a line here and look scleaner. --> done

pyqtgraph/Qt.py
```diff
@@ -317,7 +317,7 @@ def isQObjectAlive(obj):
     loadUiType = uic.loadUiType
 
     QtCore.Signal = QtCore.pyqtSignal
-    
+    QtCore.Slot   = QtCore.pyqtSlot
```
j9ac9k, line 320: I'm not convinced we want/need the slot decorator; it's been discussed that there might be some performance benefits to using it; but I think that's somewhat arguable.  I would propose for the purposes of this PR to remove it.

pyqtgraph/colorRegistry.py
```diff
@@ -0,0 +1,354 @@
...
+            length = len(color)
+            if length == 0: 
+                return False
+            if length >= 1:
+                name  = color[0]
+                # if color id is list-like, the first element needs to be str or None
+                if not ( color is None or isinstance(name,str) ): return False
+            if length >= 2: alpha = color[1]
+        else:
+            return False
```
j9ac9k, line 114: do we want to return False, or should we raise some kind of exception?  I always find it odd when a function returns False, but not True (or True and not False).  An alternative would be to return None, an empty tuple (which doesn't feel right to me either).

pyqtgraph/colorRegistry.py
```diff
@@ -0,0 +1,354 @@
...
+        self.registered_objects[registration] = (weakref.ref(obj), desc)
+
+    def unregister(self, registration):
+        """
+        Removes obj (QColor, QPen or QBrush) from the registry, usually called by finalize on deletion
+        """
+        obj, desc = self.registered_objects[registration]
+        # print('unregistering', registration, '(',str(obj),'):',str(desc))
+        del self.registered_objects[registration]
+        del obj, desc
```
j9ac9k, line 317: Do we delete these here due to pyside bindings holding the extra reference preventing the garbage collector from grabbing them?

pyqtgraph/colormap.py
```diff
@@ -192,12 +194,46 @@ def getFromColorcet(name):
...
+    return cmap
+    
+def makeMonochrome(color='green'):
+    """
+    Returns a ColorMap object imitating a monochrome computer screen
+    ===============  =================================================================
+    **Arguments:**
+    color            Primary color description. Can be one of predefined identifiers
+                    'green', 'amber', 'blue', 'red', 'lavender', 'pink'
+                    or a tuple of relative R,G,B contributions in range 0.0 to 1.0
```
j9ac9k, line 210: don't know how picky sphinx is about these indentations, might need an extra space?

pyqtgraph/colormap.py
```diff
@@ -192,12 +194,46 @@ def getFromColorcet(name):
...
+    ===============  =================================================================
+    """
+    name = 'monochrome-'+str(color)
```
j9ac9k, line 213: may I suggest an f-string here?

```python
name = f"monochrome-{color}"
```

pyqtgraph/colormap.py
```diff
@@ -192,12 +194,46 @@ def getFromColorcet(name):
...
+    leakage = np.array([    0,     1,     7,    21,    45,    80,   127,   191])
+    delta_arr = active - leakage
+    predefined = {
+        'green': (0.00, 1.00, 0.33), 'amber'   : (1.00, 0.50, 0.00),
+        'blue' : (0.00, 0.50, 1.00), 'red'     :  (1.00, 0.10, 0.00),
+        'pink' : (1.00, 0.10, 0.50), 'lavender': (0.67, 0.33, 1.00)
+    }
+    if color in predefined: color = predefined[color]
+    if not isinstance(color, tuple):
+        definitions = ["'"+key+"'" for key in predefined]
```
j9ac9k, line 225: also would recommend an f-string here as well

```python
definitions = [f"'{}key'" for key in predefined]
```

pyqtgraph/colormap.py
```diff
@@ -192,12 +194,46 @@ def getFromColorcet(name):
...
+        raise ValueError("'color' needs to be an (R,G,B) tuple of floats or one of "+definitions.join(', ') )
+    r, g, b = color
+    color_list = []
+    for leak, delta in zip(leakage, delta_arr):
+        color_tuple = (
+            r * delta + leak,
+            g * delta + leak,
+            b * delta + leak )
+        color_list.append(color_tuple)
```
j9ac9k, line 234: can probably make this into a list comprehension

```python
color_list = [
    (
        r * delta + leak,
        g * delta + leak,
        b * delta + leak
    ) for leak, delta in zip(leakage, delta_arr)
]
```

pyqtgraph/functions.py
```diff
@@ -225,147 +241,218 @@ def mkColor(*args):
...
+    # now args is either a non-list entity, or a multi-element tuple
+    if args is None:
+        raise TypeError('Color specifier cannot be None.')
+
+    if isinstance(args, QtGui.QColor):
+        if hasattr(args,'registration'):
+            return args # pass through registered pen directly
+        return QtGui.QColor(args)  ## return a copy of this color
+
+    err = 'Could not create a color from {:s}(type: {:s})'.format(str(args), str(type(args)))
```
j9ac9k, line 261: let's just use a f-string here again:

```python
err = f"Could not create a color from {args}(type: {type(args)})"
```

pyqtgraph/functions.py
```diff
@@ -225,147 +241,218 @@ def mkColor(*args):
...
+    result = COLOR_REGISTRY.getRegisteredColor(args)
+    if result is not None: # make a NamedPen
```
j9ac9k, line 263: what's a named pen? ;)

pyqtgraph/palette.py
```diff
@@ -0,0 +1,414 @@
...
+    ===================  =============================================================
+    """
+    if isinstance(identifier, (QtGui.QColor, QtCore.Qt.GlobalColor)):
+        return QtGui.QColor(identifier)
+    alpha = None
+    if isinstance(identifier, str): # return known QColor
+        name = identifier
+        if color_dict is None or name not in color_dict:
+            if name[0] != '#':
+                raise ValueError('Undefined color name '+str(identifier))
```
j9ac9k, line 124: would again recommend a f-string here `f"Undefined color name {identifier}"`

pyqtgraph/palette.py
```diff
@@ -0,0 +1,414 @@
...
+        # self.palette  = { # populate dictionary of QColors with legacy defaults
+        #     'b': QtGui.QColor(  0,  0,255,255), 'g': QtGui.QColor(  0,255,  0,255), 
+        #     'r': QtGui.QColor(255,  0,  0,255), 'c': QtGui.QColor(  0,255,255,255),
+        #     'm': QtGui.QColor(255,  0,255,255), 'y': QtGui.QColor(255,255,  0,255),
+        #     'k': QtGui.QColor(  0,  0,  0,255), 'w': QtGui.QColor(255,255,255,255),
+        #     'd': QtGui.QColor(150,150,150,255), 'l': QtGui.QColor(200,200,200,255), 
+        #     's': QtGui.QColor(100,100,150,255)
+        # }
+        # self.addDefaultRamp()
+        self.dark = None # is initially set when assigning system palette
```
j9ac9k, line 210: instead of setting to None, we can set default to what the current setup is:

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/Qt.py#L417-L420

```python
app = QtWidgets.QApplication.instance()
self.dark = app.property('darkMode')
```

This way you shouldn't have to worry about the case of `None`

pyqtgraph/palette.py
```diff
@@ -0,0 +1,414 @@
...
+        Setting 'gr_bg' with a mean color value < 127 will set the palette's 'dark' property
+        """
+        for key in colors:
+            col = identifier_to_QColor( colors[key], color_dict=self.palette )
+            if key == 'gr_bg':
+                bg_tuple = col.getRgb()
+                self.dark = bool( sum( bg_tuple[:3] ) < 3 * 127 ) # dark mode?
+            self.palette[key] = col
+            
+    def addDefaultRamp(self):
```
j9ac9k, line 305: should there be something in this function, or should it be removed?

pyqtgraph/palette.py
```diff
@@ -0,0 +1,414 @@
...
+        brightness = [
+            (np.sum( ref_color * needed[key] ), key) 
+            for key in needed
+        ]
+        brightness = sorted(brightness, key=lambda brightness: brightness[0])
+        # print( brightness )
+        avail = ('m3','m4','m5','m6','m7','m8')
+        colors = {}
+        for idx, (value, key) in enumerate(brightness):
+            colors[key] = avail[idx]
```
j9ac9k, line 406: dict comprehension time!

```python
colors = {key: avail[idx] for idx, (_, key) in enumerate(brightness}
```

Actually, relooking at this, probably better off... (was guessing on variable names, please make better choices than I did 😆  )

```python
colors = {key: avai for _, key, avai in zip(brightness, avail)
```

j9ac9k:
Hi @NilsNemitz  I did another look through on the code; most of my changes now involve using list comprehensions instead, f-strings, things of that nature.  I didn't make comments, but I think at this point it's ok to start deleting commented out code.

One thing I did want to discuss was naming of signals/slots.  

Generally "somethingChanged" refers to a signal in Qt land; and you have "styleHasChanged" as a slot.  I would recommend having `stylePaletteChanged`, an dI would suggest naming the slot `setStyle` _but_ `QGraphicsWidget.setStyle` is already taken; we could do `viewStylePaletteChanged` and `setViewStyle`; or `graphicsPaletteChanged` and `setGraphicsPalette` perhaps.  I'm open to other suggestions.  Seems the bulk of the naming conflicts come from [QGraphicsWidget](https://doc.qt.io/qt-5/qgraphicswidget-members.html).

pyqtgraph/Qt.py
NilsNemitz: Fair enough. It seems that this works without the slot decorator. --> done.

pyqtgraph/colorRegistry.py
NilsNemitz: This is in `colorRegistry.confirmColorDescriptor()`, which is tasked with checking that a given set of arguments can function as a 'descriptor' as used specifically by the color registry.
There are formats that are valid parameters for `mkColor()`, but not for the registry: Like `(255,0,0)` for red. Right now `mkColor()` just passes that to `getRegisteredColor()` and thus to `confirmColorDescriptor()`. `False` is a perfectly valid reply: It just means that the color registry will not handle this, since it looks for descriptors like 'r' or ('r',100). It does (reluctantly) handle the hex codes now ('#FF0000'), but only because all the required parsing is already done, and it seems wasteful to bounce back to `mkColor()` just repeat the same code...

If we decide to raise an error (probably in `getRegisteredColor()` instead of here) for invalid inputs, then we need a try/except block in `mkColor()`  

Is that acceptable?

pyqtgraph/colorRegistry.py
NilsNemitz: No, this is just paranoia. `obj` and `desc` should get marked for garbage collection and purged from the registry by the `finalize()` code as soon as the `unregister()` method is done.

We can simply avoid creating the extra references here by commenting out the first line (#314), too.
Although I would like to keep the code as comments, since I am pretty sure we'll need it to hunt down some issue with the weakref management in the future.

The issue with PySide shows up in the testing code:
If I create and register a number of color in PyQt5, then `finalize()` calls completely clear the registry once I delete them.
In PySide, the registry still holds a single living weakref to the latest registered color. My expectation is that this will be successfully removed as soon as a new QColor is defined and/or registered. So, for now, the tests accept a <single> surviving reference. We'll need to keep an eye on this, but for now it seems to be reasonable.

pyqtgraph/colormap.py
NilsNemitz: --> updated to modernized NumPy style docstring format.

pyqtgraph/colormap.py
NilsNemitz: Got point, I should get fluent with those. It does seem to work, even where the input is a tuple of floats.
--> changed.

pyqtgraph/colormap.py
NilsNemitz: --> done

pyqtgraph/colormap.py
NilsNemitz: --> done

pyqtgraph/functions.py
NilsNemitz: aww, I like `format`   :(
--> done

pyqtgraph/functions.py
NilsNemitz: No idea! --> changed

pyqtgraph/palette.py
NilsNemitz: I will grudgingly admit that these seem quite useful. --> changed

pyqtgraph/palette.py
NilsNemitz: For the moment, each palette has a `dark` property, which currently simply represents that the background color is darker than (127,127,127). Since we are looking only at the plot palette here, this is not strongly linked to the system palette. PyQtGraph has been plotting dark plots in a light window for a long time.

As the comment says, this gets initialized immediately, so there is nothing to worry about. But not setting it makes the linter complain...

So I think this is okay here. We might want some method somewhere that sets the initial plot palette based on the system mode... But that would happen somewhere else. And for continuity, I think PyQtGraph should always start up with a dark plotting palette until instructed otherwise.

pyqtgraph/palette.py
NilsNemitz: For now, all custom gray colors are defined manually, but this function should replace that eventually.
Sorry, I forgot that this was left as a stub for now.

pyqtgraph/palette.py
NilsNemitz: I think I'll change this to
```python
        choice = ('m3','m4','m5','m6','m7','m8') # 6 candidates for 6 needed colors
        colors = {
            key: choice[idx] # assign in order of dark to bright
            for idx, (_, key) in enumerate(brightness)
        }
```
It's not quite as compact, but I personally find it easier to see the ordering when the index is explicitly present.

NilsNemitz:
Dear @j9ac9k , thanks for looking over this (too long) PR!

I tried to make some improvements based on your comments.
There are still items on the to-do list. The monochrome ramp generator is one, another would be to integrate the color map used to find plot colors with the `intColor()` call in `pg.functions`.

But this is more quality of life stuff and shouldn't change API or functionality much. 

I renamed the signal `graphStyleChanged` and the slot `updateGraphStyle`, does that work better?
I think "graph" is not taken yet, and it seems an appropriate naming for PyQtGraph.

pyqtgraph/Qt.py
```diff
@@ -316,6 +316,7 @@ def isQObjectAlive(obj):
     loadUiType = uic.loadUiType
 
     QtCore.Signal = QtCore.pyqtSignal
+    # QtCore.Slot   = QtCore.pyqtSlot # The current policy is to avoid slot decorators.
```
outofculture, line 319: This line can be deleted, yeah?

outofculture, line 319: This line can be deleted.

pyqtgraph/__init__.py
```diff
@@ -60,21 +60,26 @@
...
+#     """Set global configuration options. 
     
-    Each keyword argument sets one global option. 
-    """
-    for k,v in opts.items():
-        setConfigOption(k, v)
+#     Each keyword argument sets one global option. 
+#     """
+#     for k,v in opts.items():
+#         setConfigOption(k, v)
```
outofculture, line 82: All this commented code can just be deleted.

pyqtgraph/styleRegistry.py
```diff
@@ -0,0 +1,369 @@
...
+    ('gr_fg','d'),  # graphical foreground
+    ('gr_bg','k'),  # graphical background
+    ('gr_txt','d'), # graphical text color
+    ('gr_hlt','r')  # graphical hover color
+]:
+    DEFAULT_COLORS[key] = DEFAULT_COLORS[col]
+
+del key, col # let the linter know that we are done with these
+
+def _expand_rgba_hex_string(hex):
```
outofculture, line 54: This behavior (or very similar) is already implemented inside `pyqtgraph.functions.mkColor`, so we should try to refactor that out and share it.

Oscar438:
Not sure if this has been abondoned or not, but I'd love to see this!

PR #1630, opened 2021-03-13
combine levels and lut only if both are present

pijyoi:
This PR does a partial fix for #1590.

Previously, if ```levels``` was present but ```lut``` was not, ```levels``` would be converted into ```lut```.
i.e. levels and ~lut --> ~levels and lut 
However, since makeARGB _always_ applies levels, this becomes a pessimization.
i.e. both levels and lut are now applied.

This PR changes it to become: levels and lut --> ~levels and lut

With this fix, in the (perhaps common) use case where the user does not provide a lut, the pessimization will no longer get triggered.


j9ac9k:
Thanks @pijyoi for narrowing down the code-paths and identifying unnecessary computation

PR #1632, opened 2021-03-14
rescaleData : workaround np.clip regression since numpy 1.17

pijyoi:
NumPy has an open issue where ```np.clip()``` has a large performance regression since 1.17 : https://github.com/numpy/numpy/issues/14281
This affects pyqtgraph's ```rescaleData()``` and accounts for a large chunk of the time taken for ```makeARGB()```.

This PR uses the workaround described in numpy issue above.

Using the "Code to reproduce" example given in https://github.com/pyqtgraph/pyqtgraph/issues/1590, performance improvements of this PR can be observed.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1009,7 +1009,18 @@ def rescaleData(data, scale, offset, dtype=None, clip=None):
         dtype = data.dtype
     else:
         dtype = np.dtype(dtype)
-    
+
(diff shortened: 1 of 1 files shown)


j9ac9k:
Tagging @NilsNemitz as the dynamic rescaling using `np.clip` as well and is in a fairly performance sensitive codepath.

Thanks for tracking this down @pijyoi   This LGTM.

Issue #617, opened 2018-01-16
PyQtGraph-core is outdated


pavel-kirienko:
First things first, *PyQtGraph rocks*. I am using it in several projects, mostly for real-time data visualization purposes. Two of the projects embed PyQtGraph as a git submodule through https://github.com/pyqtgraph/pyqtgraph-core.

The problem with the core repo is that it hasn't been updated in ages (last commit almost 3 years ago, latest release 0.9.10). Is it not maintained anymore? Is it still a good idea to use it as a submodule? If not, what is the current best practice? I don't want to depend on the PIP package because it complicates dependency management and distribution.

Thanks!

j9ac9k:
Sort of resolved with #1636 

There is little benefit for attempting to keep pyqtgraph-core up to date when with a small change to projects, the main repository can be used (without installing it via pip).  The documentation change in #1636 details how to do that.

While I was initially thinking we would update pyqtgraph-core, the addition of some supplemental functionality (colormaps being the first to come to mind) make the question of if something should be included with pyqtgraph-core or not not as obvious (you can make a case for both ways).

I hope this change will work out for you @pavel-kirienko if you need some assistance, let us know.

PR #1636, opened 2021-03-19
remove pyqtgraph-core from documentation

### LINKED ISSUE #1624: Remove reference to pyqtgraph-core repository from the documentation.
The [Embedding PyQtGraph as a sub-package of a larger project](https://pyqtgraph.readthedocs.io/en/latest/how_to_use.html#embedding-pyqtgraph-as-a-sub-package-of-a-larger-project) section of the documentation says that you can use the [pyqtgraph-core](https://github.com/pyqtgraph/pyqtgraph-core) repository as a sub tree in your project. However the pyqtgraph-core repository hasn't been updated in six years, The last commit is from May 2015.

I guess that this functionality is not used, so I think it's best to remove it from the documentation.

NilsNemitz:
The version of *pyqtgraph-core*, intended as a local stand-alone installation, is wildly out of date.

There is no immediate intention to update the core branch in the near future. This PR aims to fix the documentation and save users from struggling with the outdated code.

The good news is that the current main branch seems to work perfectly fine as a stand-alone installation, requiring exactly the same changes to the application code as were needed for using *pyqtgraph-core*: Prefixing the import statement. This is true at least for my development system on Windows. I hope it is also working as designed on other platforms.

I changed the instructions to explain such an installation instead. I added a reference to `git clone --depth 1` functionality for those who want to save a byte or two. 

The `git read-tree` section goes well beyond my comfort-zone. I tested the sequence of commands in the previous documentation with names changed to pull `pyqtgraph/master` rather than `pyqtgraph/core`. After adding the `--allow-unrelated-histories` flag, this results in an initial folder structure that looks as intended.

This closes #1624 and #617.

### DIFF
--- a/doc/source/how_to_use.rst
+++ b/doc/source/how_to_use.rst
@@ -101,9 +101,9 @@ Embedding PyQtGraph as a sub-package of a larger project
 
 When writing applications or python packages that make use of pyqtgraph, it is most common to install pyqtgraph system-wide (or within a virtualenv) and simply call `import pyqtgraph` from within your application. The main benefit to this is that pyqtgraph is configured independently of your application and thus you (or your users) are free to install newer versions of pyqtgraph without changing anything in your application. This is standard practice when developing with python.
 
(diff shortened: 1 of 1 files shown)


j9ac9k:
I have a branch locally that I was modifying to shift files around to make `pyqtgraph-core` a bit more feasible to update, but I'm not really sure I see the benefit of it at this point; so I'm 👍🏻  for eliminating it, and adjusting the documentation the way you have done it @NilsNemitz 

This LGTM, merging.  Thanks for the PR @NilsNemitz

Issue #1634, opened 2021-03-18
Different yData values obtained 0.11.0 vs 0.11.1


nicoddemus:
Hi folks!

First of all thanks a lot for the great graph library!

I've noticed a difference in behavior of versions `0.11.0` and `0.11.1` when comparing `item.curve.yData` with the original values passed during `plotItem.plot`.

This example reproduces the issue:

```python
from PyQt5.QtWidgets import QApplication
from pyqtgraph import PlotWidget, __version__
import numpy as np

app = QApplication([])
plot_window = PlotWidget()

x = np.array([2250, 2500, 2750, 3000, 3250, 3500, 3750])
y = np.array([41539180., 32281334., 23945627., 16794226., 11033336., 6815729., 4245312.])

plot_window.plotItem.plot(x, y)

item = plot_window.plotItem.items[0]
print(f"pyqtgraph version: {__version__}")
print("x matches:", np.all(np.isclose(item.curve.xData, x)))
print("y matches:", np.all(np.isclose(item.curve.yData, y)))
print(item.curve.yData)

plot_window.show()
app.exec()
```

In `0.11.0` I get this output:

```
pyqtgraph version: 0.11.0
x matches: True
y matches: True
[41539180. 32281334. 23945627. 16794226. 11033336.  6815729.  4245312.]
```

And this graph:

![image](https://user-images.githubusercontent.com/1085180/111697783-befe6a80-8814-11eb-976f-9f60a8cd4e3c.png)

While in `0.11.1` I get this:

```
pyqtgraph version: 0.11.1
x matches: True
y matches: False
[2000001.5 2000001.5 2000001.5 2000001.5 2000001.5 2000001.5 2000001.5]
```

The graph image however is the same:

![image](https://user-images.githubusercontent.com/1085180/111697862-dccbcf80-8814-11eb-9289-ef5ddefab10d.png)

---

This is not a big deal, I'm just curious if this is a bug or some expected behavior.

I browsed the CHANGELOG, but couldn't find anything obvious (at least to me as an user).

Thanks a lot!

j9ac9k:
👋🏻  @nicoddemus I had no idea you were a pyqtgraph user (would love to hear about your use of the library sometime); big fan of your open source work, we're hoping to incorporate pytest-qt in our library before much longer, but back to your issue.  

Our changelog is a bit (understatement of the century) tough to read.  As we migrate to faster release cycles we hope this to be less of an issue.

[...]

PR #1637, opened 2021-03-19
PlotDataItem: Fix view range <-> dynamic range limit pathology and switch to np.umath.clip

NilsNemitz:
Issue #1634 brought to light an issue with the current code that does dynamic range limiting in `PlotDataItem` (see #1140):

If all data is far above (or below) the screen, then it will all be clipped to the same value.
If auto-range is active, it will then center the view on the constant-valued clipped data, and set the view height to the minimum allowed.

Data will then be re-clipped according to the updated view, but since the view height is very small, the data will be clipped aggressively again. For far-away data, all clipped values will again be identical.

This process then repeats, with the center of the view range shifting towards the true data by approximately 1E6 every loop.
If the input data is all >1E10, this process takes an amusingly long time.

The solution is simple:
Never clip data when it is all off to the top or bottom, which we can judge from pre-existing information.
Then auto-range will jump to the proper range in one step, as intended. 
This also happens to be an optimization, since it saves calculation when nothing is visible.

The PR also switches `np.clip()` for `np.umath.clip()` throughout PlotDataItem.
This works around a problem with code overhead in numpy's main clip function, as discussed in  #1632 .

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -671,8 +671,8 @@ class PlotDataItem(GraphicsObject):
                         # workaround for slowdown from numpy deprecation issues in 1.17 to 1.20+
                         # x0 = np.clip(int((range.left()-x[0])/dx) - 1*ds, 0, len(x)-1)
                         # x1 = np.clip(int((range.right()-x[0])/dx) + 2*ds, 0, len(x)-1)
-                        x0 = np.core.umath.clip(int((range.left()-x[0])/dx) - 1*ds, 0, len(x)-1)
-                        x1 = np.core.umath.clip(int((range.right()-x[0])/dx) + 2*ds, 0, len(x)-1)
+                        x0 = fn.clip_array(int((range.left()-x[0])/dx) - 1*ds, 0, len(x)-1)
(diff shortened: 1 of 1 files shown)


j9ac9k:
In the comments, I would actually just put the full link to the issue (this has been _super_ helpful when seein gworkarounds for old Qt bugs that I periodically see, and then check out to see if the issue was resolved).

Link in this case is: https://github.com/numpy/numpy/issues/14281

pijyoi:
I have created a ```pg.clip_array``` function in #1641.
@NilsNemitz , ```np.core.umath.clip``` does not perform well on macOS and Windows. If you want, you could merge ```pg.clip_array``` into your PR and I will drop #1641.

NilsNemitz:
@pijyoi , that sounds good, it makes sense to have the workaround defined in only one place.
Could you check that your function leaves the input array intact? 
I need the "copy and clip only the copy" functionality of the normal np.clip without `out=` keyword.

j9ac9k:
Now that #1641 is merged, feel free to make use of `pg.functions.clip_array`

NilsNemitz:
Thank you @j9ac9k for handling this so quickly!
@pijyoi, since #1641 is already in, this now uses the new `functions.clip_array()` code.
Any other suggestions?

j9ac9k:
This LGTM, thanks for submitting this @NilsNemitz Merging.

PR #1644, opened 2021-03-21
Add Deprecation Warning to Hex Strings That Do Not Start With "#" in mkColor

j9ac9k:
@NilsNemitz `https://i.imgur.com/uyKgrYk.gif`

`mkColor` supports a wide variety of inputs to generate `QColor` objects.  One of the currently supported features is providing a hexadecimal value for RGB/RGBA via hex-string, from the docs 

```
     "RGB"           hexadecimal strings; may begin with '#'
     "RGBA"          
     "RRGGBB"       
     "RRGGBBAA"     
```

As we try and support more color-names, it will be more difficult to distinguish between color names and hexadecimal color codes; so we are going to stop supporting that.

Given this is a public API change, we're going to go through a deprecation phase, and won't actually remove this functionality until the next minor or major release is rolled over.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -223,10 +223,10 @@ def mkColor(*args):
      float           greyscale, 0.0-1.0
      int             see :func:`intColor() <pyqtgraph.intColor>`
      (int, hues)     see :func:`intColor() <pyqtgraph.intColor>`
-     "RGB"           hexadecimal strings; deprecated if they don't start
-     "RGBA"          with "#"
-     "RRGGBB"       
-     "RRGGBBAA"     
+     "#RGB"          hexadecimal strings prefixed with '#'
+     "#RGBA"         previously allowed use without prefix is deprecated and 
+     "#RRGGBB"       will be removed in 0.13
+     "#RRGGBBAA"     
      QColor          QColor instance; makes a copy.
     ================ ================================================
     """
@@ -244,7 +244,7 @@ def mkColor(*args):
             else:
                 warnings.warn(
                     "Parsing of hex strings that do not start with '#' is"
-                    "deprecated support will be removed in 0.13",
+                    "deprecated and support will be removed in 0.13",
                     DeprecationWarning, stacklevel=2
                 )
             if len(c) == 3:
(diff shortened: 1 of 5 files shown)


pyqtgraph/functions.py
```diff
@@ -233,13 +234,19 @@ def mkColor(*args):
...
                 try:
                     return Colors[c]
                 except KeyError:
                     raise ValueError('No color named "%s"' % c)
+            if c[0] == '#':
+                c = c[1:]
+            else:
+                warnings.warn(
+                    "Parsing of hex strings that do not start with '#' is"
+                    "deprecated support will be removed in 0.13",
```
NilsNemitz, line 247: --> deprecated *and* support will be removed in 0.13  ??

pyqtgraph/functions.py
```diff
@@ -222,8 +223,8 @@ def mkColor(*args):
      float           greyscale, 0.0-1.0
      int             see :func:`intColor() <pyqtgraph.intColor>`
      (int, hues)     see :func:`intColor() <pyqtgraph.intColor>`
-     "RGB"           hexadecimal strings; may begin with '#'
-     "RGBA"          
+     "RGB"           hexadecimal strings; deprecated if they don't start
+     "RGBA"          with "#"
```
NilsNemitz, line 227: Can we change the documentation to
```
"#RGB"     hexadecimal strings prefixed with '#'
"#RGBA"    previously allowed use without prefix is deprecated and will be removed in 0.13
```
The current version seems like too much of a double negative.

NilsNemitz:
I changed the warning to an error and ran the tests:
In examples/customPlot.py : line 45: `color="333333"`  needs to be `color="#333333"`
In examples/parameterTree.py : line 72:  `'value': "FF0"` needs to be  `'value': "#FF0"` 

With that, I get no reported errors excep for `FAILED pyqtgraph/tests/test_exit_crash.py`, which I always get, unfortunately.

So with these two additional changes, we no longer use colors without prefixes throughout the examples and test suite. I tested that the color selection box in the parameterTree example... No problems, it seems only the example initialization used the ultra-short format.

pyqtgraph/functions.py
j9ac9k: Good suggestion, incorporated.

pyqtgraph/functions.py
j9ac9k: Good point, docs shouldn't show how to not do it haha.

PR #1645, opened 2021-03-21
Replace main stanza with PyQt6 compatable variant

### LINKED ISSUE #1609: examples main stanza incorrect for PyQt6
### Short description

The following standard stanza is not working as intended for PyQt6.
```python
if __name__ == '__main__':
    import sys
    if sys.flags.interactive != 1 or not hasattr(QtCore, 'PYQT_VERSION'):
        pg.QtGui.QApplication.exec_()
```
PyQt5 installs an input hook by default., https://doc.bccnsoft.com/docs/PyQt5/python_shell.html. 
It appears that PyQt6 does not do the same by default.

### Code to reproduce

```
set PYQTGRAPH_QT_LIB=PyQt6
py -i examples\SimplePlot.py
```

### Expected behavior

A plot window should appear. (Plot window appears for PyQt5, PySide2, PySide6)

### Real behavior

Plot window does not appear.
Python REPL appears.

### Tested environment(s)

 * PyQtGraph version: 0.11.1.dev0 
 * Qt Python binding: PyQt6 6.0.1 Qt 6.0.1
 * Python version: 3.7.9
 * NumPy version: 1.19.5 
 * Operating system: Windows 10 x64

### Additional context
```sys.flags.interactive == 1``` is only true if python is executed with the "-i" argument. So in fact, the standard stanza is targeting only the case where the user wants to execute the example in interactive mode _and_ is running PyQt5.

If such a use case is not common, the standard stanza could be simplified to:
```python
if __name__ == '__main__':
    pg.Qt.QtWidgets.QApplication.exec_()
```

Another way to look at it is that out of the current 4 supported bindings, PyQt5 is the only one that installs an input hook by default.

Incidentally, examples\CLIexample.py did not import QtCore, so executing ```py -i examples\CLIexample.py``` under PyQt5 throws an exception.

j9ac9k:
Fixes #1609 

Thanks @pijyoi for identifying this issue.


ixjlyons:
How about `pg.mkQApp().exec_()`?

j9ac9k:
> How about `pg.mkQApp().exec_()`?

I was thinking about that, and I can't think of any side effects, @pijyoi  what do you think?

ixjlyons:
I pretty much always go with the template:

```python
import pyqtgraph as pg

app = pg.mkQApp()
...
app.exec_()
```

It feels the most Pythonic to me and is just convenient IMO. For some demos where it's more of a Qt application making use of some pyqtgraph components, maybe there it makes sense to demonstrate the interoperability with the Qt API.

pijyoi:
> I was thinking about that, and I can't think of any side effects, @pijyoi what do you think?

```pg.mkQApp().exec_()``` is fine.

If I did suggest ```pg.Qt.QtWidgets.QApplication.exec_()```, it was a tiny attempt to nudge people to use Qt5 namespace ```QtWidgets.QApplication``` rather than ```QtGui.QApplication```.

j9ac9k:
> `pg.mkQApp().exec_()` is fine.

I'll update this PR with that then  👍🏻

Issue #1620, opened 2021-03-04
Use funding to improve pyqtgraph


outofculture:
I'm seeking funding from [CZI](https://chanzuckerberg.com/rfa/essential-open-source-software-for-science/) to improve pyqtgraph, and I'd like to hear from others what parts of pyqtgraph are most in need of improvement, or are most easily addressed with monetary assistance. So far, ideas we have:

- GPU-equipped test runners
- dedicated benchmarking hardware
- improve `makeARGB` performance
  - the benchmarking hardware will make this easier
  - hiring a dedicated developer for this task
- tedious improvements like:
  - typing
  - architecture improvements
  - paying down tech debt
- outreach and community maintenance
  - documentation
  - hiring a developer to do bug squashing

What other ideas do people have?

( I'm hoping to hit as many active developers and users with this discussion as possible; it's also happening in slack, and I'll post to the mailing list. I'll try to keep this issue relatively in sync. )

outofculture:
Tagging @pijyoi - can you think of ways to use money to speed up `makeARGB` in particular?

j9ac9k:
I really want to hear from our users on this, so I'm going to tag some folks I know use this library, to solicit input from them, as I think our users should have a large say ... going off of memory here, so I'm sure I'm going to miss some people, but folks that come immediately to mind include:

@irgolic @markotoplak @pijyoi @hhslepicka @danielhrisca @NilsNemitz @ntjess @mliberty1

outofculture:
Worth noting that we may not win funding with our proposal, but even if we don't, this discussion will be a nice synthesis of project goals.

pijyoi:
>can you think of ways to use money to speed up makeARGB in particular?

QImage supports other formats than Format_ARGB32, so for those cases, having makeARGB() convert the data to ARGB format is a waste.
e.g. 
1) grayscale images --> Format_Grayscale8 (since Qt 5.5) or Format_Indexed8
2) RGB images --> Format_RGB888

For uses of LUT that is for colorization, those could perhaps be converted to using Format_Indexed8

jpgill86:
Mentioning also @samuelgarcia 😄

danielhrisca:
I'm only using a small part of pyqtgraph, but for my use case I would certainly like to see improved performance of 2D plots.

j9ac9k:
> I'm only using a small part of pyqtgraph, but for my use case I would certainly like to see improved performance of 2D plots.

[...]

PR #1646, opened 2021-03-21
Remove unhelpful warnings

### LINKED ISSUE #1623: Remove warning: QApplication was created before pyqtgraph was imported
Due to the plugin nature of my project I have to import PyQtGraph after I have created my `QApplication` object. On MacOS this results in a warning by PyQtGraph. For example ...

```
from PyQt5 import QtWidgets
import pyqtgraph
```

prints the following to stdout:

```
Warning: QApplication was created before pyqtgraph was imported; there may be problems.
```

It does not state what these problems may be but I never encountered any issues. I vaguely recall reading that it had something to do with setting the graphics system to `raster`, but I can't find anything about it anymore. If this is the case I think the warning can be removed since `setGraphicsSystem` had been removed from Qt5, and Qt4 is no longer supported by PyQtGraph.

So, can the warning be removed? And if not, can you please make it a proper warning by using Pythons logging or warning mechanism?

 * PyQtGraph version: 0.11.1
 * Operating system: OSX 10.14.1

j9ac9k:
One of the comments from #1620 was regarding the wording in the docs relating to `useOpenGL`; while the wording did give a warning, it was not particularly helpful.  We either support that option, or we don't, so that was removed.

References to `weave` have been removed since that hasn't been in the library forever.

Fixes #1623 by removing another unhelpful warning as well.

### DIFF
--- a/doc/source/config_options.rst
+++ b/doc/source/config_options.rst
@@ -26,8 +26,6 @@ imageAxisOrder     str                 'col-major'        For 'row-major', image
                                                           change in the future.
 editorCommand      str or None         None               Command used to invoke code editor from ConsoleWidget.
 exitCleanup        bool                True               Attempt to work around some exit crash bugs in PyQt and PySide.
-useWeave           bool                False              Use weave to speed up some operations, if it is available.
-weaveDebug         bool                False              Print full error message if weave compile fails.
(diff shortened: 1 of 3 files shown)


NilsNemitz:
Looks like a pretty good idea. Reference to 'useWeave' and 'weaveDebug' indeed occur *only* in `__init__.py`, so it seems pretty safe to remove them :)

PR #1649, opened 2021-03-21
Fix functions.clip_array()

pijyoi:
#1641 introduced functions.clip_array().
This PR fixes some defects:
1) if both vmin and vmax were None, an uninitialized array would be returned
2) using combination of umath.minimum + umath.maximum benefits only win32 and penalizes other platforms

Development for this PR was done in #1648 where timings before and after this change are recorded.


j9ac9k:
Thanks for the follow up @pijyoi

PR #1650, opened 2021-03-22
Do not use clip_array on scalars (use clip_scalar instead)

NilsNemitz:
As discussed in #1641, the replacement for `np.clip()` in PlotDataitem now accidentally uses @pijyoi 's `functions.clip_array()` method to limit the range of scalar variables. Although the tests for clipToView pass, indicating that this actually seems to work, this is not intended... and not fast either.

This PR adds a `functions.clip_scalar(val, vmin, vmax)` function and uses that in the clipToView part of PlotDataItem.
The appeal is to have less ambiguity than using np.clip for everything, as well as generally better performance:

Direct python comparisons are about 100x faster than the current `np.clip()` on Windows:
```python
>>> timeit('clip_scalar( random.random(), -0.1, 0.1)', number=10000000, globals=globals())
2.789680599999997
>>> timeit('np.clip( random.random(), -0.1, 0.1)', number=10000000, globals=globals())
217.01810209999996 
```

If None or an array is passed to this function, it fails right away with a pretty clear error message, so I think handling that explicitly is not needed here?

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -998,6 +998,9 @@ def solveBilinearTransform(points1, points2):
     
     return matrix
 
+def clip_scalar(val, vmin, vmax):
+    """ convenience function to avoid using np.clip for scalar values """
+    return vmin if val < vmin else vmax if val > vmax else val
 
 def clip_array(arr, vmin, vmax, out=None):
     # replacement for np.clip due to regression in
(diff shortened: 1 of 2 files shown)


j9ac9k:
Hi @NilsNemitz 

Thanks for the PR.   This looks good to me, merging!

PR #1651, opened 2021-03-23
collections.abc not just collections

j9ac9k:
Just saw a portion of the library we were not using collections.abc in but using collections, so fixing that to not emit scary warnings and maintain python 3.9+ compatibility.


NilsNemitz (commented):
This looks good to me. I checked that the color map example is still able to load all Matplotlib maps without glitches or missing data.

I would propose two additional changes:
1) rename: 
`_get_from_matplotlib()` --> `getFromMatplotlib()`
`_get_from_colorcet()` --> `getFromColorcet()`
`_get_from_file()` --> `_getFromFile()`
This promotes the first two to first class user accessible functions as suggested by @Honkomonk in #1424

2) Move the Color Maps example to a different section in utils.py, so that it appears in the ExampleApp. Because choosing a colormap out of the large pool is a lot easier with a visual reference.

PR #1653, opened 2021-03-23
Fix to accidentally disabled style updates in PlotDataItem

### LINKED ISSUE #1652: commit 574c2d2 breaks a lot of PlotDataItem methods
Commit 574c2d2 / pull request #1619 broke a lot of my code. I have the user being allowed to update the line styles, pens, etc., and all that appeared to stop working.

After looking through it, it appears that this is because PlotDataItem.updateItems() was given the default argument update_style=False. This method is called in many places in PlotDataItem without the update_style argument. For example, setPen(), setSymbol(), etc. In all of these locations, when updateItems() is called, it should be called with update_style=True.

I'm afraid I'm not comfortable enough with the codebase to put in a pull request to fix this, as there are many places in PlotDataItem that updateItems() is called where I'm unsure what the update_style argument should be. It might also be worth seeing where this fell through the cracks in the testing framework.

A contrived minimal working example is below (it was taken from https://www.learnpyqt.com/tutorials/plotting-pyqtgraph/, with 1 line added). Running it under commit 5fa3901 (the prior commit) shows the line widths being updated a few times a second. Running it under commit 574c2d2 shows this not happening.

```
from PySide2 import QtWidgets, QtCore
from pyqtgraph import PlotWidget, plot
import pyqtgraph as pg
import sys  # We need sys so that we can pass argv to QApplication
import os
from random import randint

class MainWindow(QtWidgets.QMainWindow):

        super(MainWindow, self).__init__(*args, **kwargs)

        self.graphWidget = pg.PlotWidget()
        self.setCentralWidget(self.graphWidget)

        self.x = list(range(100))  # 100 time points
        self.y = [randint(0,100) for _ in range(100)]  # 100 data points

        self.graphWidget.setBackground('w')

        pen = pg.mkPen(color=(255, 0, 0))
        self.data_line =  self.graphWidget.plot(self.x, self.y, pen=pen)

        self.timer = QtCore.QTimer()
        self.timer.setInterval(50)
        self.timer.timeout.connect(self.update_plot_data)
        self.timer.start()

    def update_plot_data(self):

        self.x = self.x[1:]  # Remove the first y element.
        self.x.append(self.x[-1] + 1)  # Add a new value 1 higher than the last.

        self.y = self.y[1:]  # Remove the first 
        self.y.append( randint(0,100))  # Add a new random value.

        self.data_line.setData(self.x, self.y)  # Update the data.

        self.data_line.setPen(width=randint(1,10))  # Update the line widths (added to show issue)


w = MainWindow()
w.show()
sys.exit(app.exec_())
```

NilsNemitz:
#1652 points out that PR #1619 broke most of the style update methods in PlotDataItem, such as `setPen()`.

The error was that `def updateItems(self, update_style=False):` disabled the propagation of pen, brush etc. settings from PlotDataItem to PlotCurveItem and ScatterPlotItem *unless explicitly requested*. This request only occurred when the style properties were set through the `setData()` method. I have now gone through the individual methods and added an explicit setting of update_style=True / False to each update, so that changes to e.g. log mode do not update the pens/brushes, but calls to setPen, setShadowPen etc. do.

Since most of out example code sets the colors on initialization or through `setData()`, this problem was not obvious in the tests. I added a set of tests that check that the properties correctly arrive with the ScatterPlotItem and PlotCurveItem associated with the PlotDataItem.

Closes #1652, I hope.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -242,7 +242,7 @@ class PlotDataItem(GraphicsObject):
             return
         self.opts['fftMode'] = mode
         self.xDisp = self.yDisp = None
-        self.updateItems()
+        self.updateItems(update_style=False)
         self.informViewBoundsChanged()
 
     def setLogMode(self, xMode, yMode):
@@ -260,7 +260,7 @@ class PlotDataItem(GraphicsObject):
             return
         self.opts['logMode'] = [xMode, yMode]
         self.xDisp = self.yDisp = None
-        self.updateItems()
+        self.updateItems(update_style=False)
         self.informViewBoundsChanged()
 
 
@@ -269,7 +269,7 @@ class PlotDataItem(GraphicsObject):
             return
         self.opts['derivativeMode'] = mode
         self.xDisp = self.yDisp = None
-        self.updateItems()
+        self.updateItems(update_style=False)
         self.informViewBoundsChanged()
 
     def setPhasemapMode(self, mode):
@@ -277,7 +277,7 @@ class PlotDataItem(GraphicsObject):
             return
         self.opts['phasemapMode'] = mode
         self.xDisp = self.yDisp = None
-        self.updateItems()
+        self.updateItems(update_style=False)
         self.informViewBoundsChanged()
 
     def setPointMode(self, mode):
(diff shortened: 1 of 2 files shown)


j9ac9k:
hi @NilsNemitz thanks for patching this up so quickly.

Only gripe I have (and I should have caught this on the original PR) is to use `camelCase` so we should modify `update_style` to `updateStyle`

j9ac9k:
We have confirmation from the author that this PR Fixes #1652

PR #1655, opened 2021-03-25
Update version to 0.12.0

ksunden:


j9ac9k:
merging before CI finished; we're living on the edge here!

Issue #1657, opened 2021-03-26
One more test failure with Python 3.10.0a6


swt2c:
Hi, I'm seeing one more test failure with Python 3.10.0a6 (not sure why I didn't see this one before?):

```
=================================== FAILURES ===================================
_____________________ testExamples[ ImageView.py - PyQt5 ] _____________________


Python 3.10.0a6 (default, Mar 22 2021, 00:00:00) [GCC 11.0.0 20210210 (Red Hat 11.0.0-0)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> >>> ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
  File "/builddir/build/BUILD/pyqtgraph-0.12.0/examples/ImageView.py", line 64, in <module>
    imv.roiClicked()
  File "/builddir/build/BUILD/pyqtgraph-0.12.0/pyqtgraph/imageview/ImageView.py", line 566, in roiClicked
    self.ui.splitter.setSizes([self.height()*0.6, self.height()*0.4])
TypeError: index 0 has type 'float' but 'int' is expected
>>> 

Failed ImageView Example Test Located in ImageView.py 
----------------------------- Captured stdout call -----------------------------
ImageView 

Python 3.10.0a6 (default, Mar 22 2021, 00:00:00) [GCC 11.0.0 20210210 (Red Hat 11.0.0-0)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> >>> ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
  File "/builddir/build/BUILD/pyqtgraph-0.12.0/examples/ImageView.py", line 64, in <module>
    imv.roiClicked()
  File "/builddir/build/BUILD/pyqtgraph-0.12.0/pyqtgraph/imageview/ImageView.py", line 566, in roiClicked
    self.ui.splitter.setSizes([self.height()*0.6, self.height()*0.4])
TypeError: index 0 has type 'float' but 'int' is expected
>>> 

=========================== short test summary info ============================
FAILED examples/test_examples.py::testExamples[ ImageView.py - PyQt5 ]
====== 1 failed, 209 passed, 5 skipped, 1 deselected in 145.50s (0:02:25) ======
```

j9ac9k:
[...]

PR #1662, opened 2021-03-26
added cast to int so that QSplitter.setSizes works on Python 3.10

NilsNemitz:
While I am not set up to test on 3.10, the error message in #1657 seems quite specific:
`TypeError: index 0 has type 'float' but 'int' is expected`

This adds a cast to int to the parameters in the `QSplitter.setSizes()` call in question.
That might be enough to fix this issue and it doesn't break any other tests.

### DIFF
--- a/pyqtgraph/canvas/Canvas.py
+++ b/pyqtgraph/canvas/Canvas.py
@@ -113,7 +113,7 @@ class Canvas(QtGui.QWidget):
         
         if not self.sizeApplied:
             self.sizeApplied = True
-            s = min(self.width(), max(100, min(200, self.width()*0.25)))
+            s = int( min(self.width(), max(100, min(200, self.width()//4))) )
             s2 = self.width()-s
             self.ui.splitter.setSizes([s2, s])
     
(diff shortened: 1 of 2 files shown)


j9ac9k:
Hi @NilsNemitz 

thanks for hoping on this.  doing a squick search in the library, there is potentially a few other places this could happen.

Potentially here: 

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/canvas/Canvas.py#L116-L118

and potentially here: 

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/WidgetGroup.py#L22-L24

the second I'm not so sure about.

Would you mind updating the first location as well, and take a glance at the second location and see if you think that's impacted as well?

NilsNemitz:
Dear @j9ac9k , I have changed to calculation to integer (and added a cast as a safety net) in the first case.
In the second, the only use of restoreSplitter I can find takes data from splitterState (right above in the code). And it goes through string encoded QByteArray, which should be int enough. 
Although restoreSplitter also accepts a list, I think we don't use that internally. Since this does not seem like a commonly used function, we might be able to leave the responsibility to the user to make sure calls are valid.

Let me know if I missed anything!

j9ac9k:
Thanks @NilsNemitz that's the conclusion I came to as well.  This LGTM, merging.

PR #1666, opened 2021-03-26
revert style change reduction in PlotDataItem

NilsNemitz:
I had a look at #1652, the failure of list-wise style definitions through PlotDataItem.

The problem is that
- PlotDataItem does multiple calls to set the data of the underlying ScatterPlotItem, of which only the first one applies the style information.
- It seems that ScatterPlotItem does not retain the per-point styling information on a call to , even if the data is not changed.

The only immediate solution I have is to revert to always transmitting the full style information.
This PR does that, and in the example given in  #1652, I do see the colors as intended then.

A proper fix will need closer inspection into why there are redundant `setData()` calls to ScatterPlotItem, which seems less than ideal.


j9ac9k:
Thanks @NilsNemitz for following up on this.  Always hate rolling back a good change, probably just a reflection of needing a more robust test suite.

Issue #1689, opened 2021-04-03
Parameters no longer have 'default' buttons when no default is specified


ntjess:
See https://github.com/pyqtgraph/pyqtgraph/commit/b41c4a71e54006e51808bc9fdba8ccf342526955#r46088064 for context.

Parameters used to create a 'default' arrow to reset a value, but now this is not the case

j9ac9k:
Fixed with the merging of #1690

PR #1690, opened 2021-04-03
revert to old behavior: default created from value if not present

ntjess:
Addresses #1689

### DIFF
--- a/pyqtgraph/parametertree/Parameter.py
+++ b/pyqtgraph/parametertree/Parameter.py
@@ -201,6 +201,7 @@ class Parameter(QtCore.QObject):
 
         if 'default' not in self.opts:
             self.opts['default'] = None
+            self.setDefault(self.opts['value'])
     
         ## Connect all state changed signals to the general sigStateChanged
         self.sigValueChanged.connect(self._emitValueChanged)
--- a/pyqtgraph/parametertree/tests/test_Parameter.py
+++ b/pyqtgraph/parametertree/tests/test_Parameter.py
@@ -9,32 +9,22 @@ def test_parameter_hasdefault():
     # default unspecified
     p = Parameter(**opts)
     assert p.hasDefault()
+    assert p.defaultValue() == opts["value"]
 
-    p.setDefault(1)
+    p.setDefault(2)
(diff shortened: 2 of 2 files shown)


ixjlyons:
This seems fine to me. This all came from a desire to be able to pass `default=None` in order to make `None` the default value (#847), but I never really understood that use case very well. It was also a breaking change, particularly removing the "reset to default" buttons for most of the `parameterTree` example.

I did have [an idea](https://github.com/pyqtgraph/pyqtgraph/commit/b41c4a71e54006e51808bc9fdba8ccf342526955#r46090789) to address this specifically for the reset button rather than changing the `hasDefault` logic, but I think this is probably more in line with how `Parameter` was originally designed, where the default is the initial value unless you specify differently.

I'd recommend removing `test_parameter_hasdefault_none` and modifying `test_parameter_hasdefault` a bit:

```python
def test_parameter_hasdefault():
    opts = {'name': 'param', 'type': int, 'value': 1}

    # default unspecified
    p = Parameter(**opts)
    assert p.hasDefault()
    assert p.defaultValue() == opts["value"]

    p.setDefault(2)
    assert p.hasDefault()
    assert p.defaultValue() == 2

    # default specified
    p = Parameter(default=0, **opts)
    assert p.hasDefault()
    assert p.defaultValue() == 0

    # default specified as None
    p = Parameter(default=None, **opts)
    assert not p.hasDefault()
```

j9ac9k:
ooooof just saw you're working on the master branch; ok had to break out some stack overflow to deal with that haha... I'll try and incorporate the suggested changes from @ixjlyons ...

and doesn't look like you enabled maintainers to modify the branch so I can't make those changes.  No problem, we'll get this merged later.

ntjess:
@j9ac9k sorry about that, didn't realize I had to enable this. Do you know where the setting is?

j9ac9k:
Hmm.... wonder if it has to do with the branch being named `master`

j9ac9k:
ok, I think I know what's wrong; nothing on your end, I recently enabled 2FA and I think I was using non-token style passwords for my command line interaction.

Issue #1676, opened 2021-03-30
PR #1556 makes new data not appear


EfremBraun:
PR #1556 introduced a flaw where when new data is plotted, the data might not show up until the user pans the scene (if that data wouldn't have shown up prior to the fix given in PR #1140 ). This is unfortunately my use case, as I allow my user to show/hide plot elements.

I'm not sure exactly how to fix this, but it looks like PR #1556 made it so that the  fix done in PR #1140 is only performed upon panning.

A minimal working example is below. It's copy/pasted from plot_test_5.zip given in PR #1556 with a button added to change the data that is being shown. If you run this example using any version of PyQtGraph past commit 33365fb (PR #1556), upon hitting the button you'll see see the magenta curve disappear. If you run it with a version of PyQtGraph prior to commit 33365fb but after commit 65e90fa (PR #1440), the magenta curve will not disappear.

```
from PySide2 import QtWidgets
import sys
import numpy as np
import pyqtgraph as pg
# import dev_pyqtgraph.pyqtgraph as pg

class MainWindow(QtWidgets.QMainWindow):
    def __init__(self, *args, **kwargs):
        super(MainWindow, self).__init__(*args, **kwargs)
        self.graphWidget = pg.PlotWidget()
        self.buttonWidget = QtWidgets.QPushButton()

        self.buttonWidget.clicked.connect(self.button_clicked)

        layout = QtWidgets.QVBoxLayout()
        layout.addWidget(self.graphWidget)
        layout.addWidget(self.buttonWidget)
        widget = QtWidgets.QWidget()
        widget.setLayout(layout)
        self.setCentralWidget(widget)

        rng = np.random.default_rng()
        length = 10_000
        self.y_data_a = rng.normal( scale=0.1, size=length )
        self.y_data_b = rng.normal( scale=0.1, size=length )
        self.y_data_c = rng.normal( scale=0.1, size=length )
        self.y_data_d = rng.normal( scale=0.1, size=length )
        idx = rng.integers( length, size=3)
        err = np.abs( rng.normal( scale=2e10, size=3 ) )
        self.y_data_b[ idx ] += err
        self.y_data_c[ idx ] -= 0.000001*err
        self.y_data_d[ idx ] -= err
        self.x_data = [idx for idx, val in enumerate(self.y_data_a)]
        self.graphWidget.setYRange( -0.1, 0.1)
[...]

PR #1691, opened 2021-04-04
Make sure dynamic range limiter runs even when there isn't previous data

NilsNemitz:
#1676 located a bug with the PlotDataItem dynamic range limiter. 

This is caused by the limiter not running on freshly set data, which may then end up being not visible due to the Qt bug the limiter is intended to work around.

The reason for the limiter not running was the addition of code to check the re-usability of the previous clipped data in #1556: The test conditions were set up so that no code path (at all) would get executed when the cache of previous data was not initialized.

This PR changes the code so that the code to create a new clipped data set runs in this case.
In my testing, this makes the example included with report #1676 function as expected.


j9ac9k:
Thanks @NilsNemitz for this PR and your quick work on patching up issues involving dynamic rescaling as they're identified.

PR #1668, opened 2021-03-27
make ImageItem combine levels + lut not be a pessimization

pijyoi:
This PR continues #1630 in trying to fix the pessimization identified in #1590

Cases tested with PYQTGRAPHPROFILE=functions.makeARGB:
1) VideoSpeedTest.py --size=3072x3072 --levels=0,255
   * both levels and lut bypassed (fastest path)
2) VideoSpeedTest.py --size=3072x3072 --levels=50,200
   * levels applied, lut bypassed
3) VideoSpeedTest.py --size=3072x3072 --lut
   * levels bypassed, lut applied (was previously both applied)
4) VideoSpeedTest.py --size=3072x3072 --levels=50,200 --lut
   * levels bypassed, lut applied (was previously both applied)
5) VideoSpeedTest.py --dtype=uint16 --size=3072x3072 --levels=0,65535
   * levels applied, lut bypassed
6) VideoSpeedTest.py --dtype=uint16 --size=3072x3072 --levels=100,10000
   * levels applied, lut bypassed
7) VideoSpeedTest.py --dtype=uint16 --size=3072x3072 --lut
   * levels bypassed, lut applied (was previously both applied)
8) VideoSpeedTest.py --dtype=uint16 --size=3072x3072 --levels=100,10000 --lut
   * levels bypassed, lut applied (was previously both applied)
 
#1630 fixed cases 1,2,5,6
This PR fixes cases 3,4,7,8

It would good to take a look at #792 and #793 to get context on what this PR is working around.
What this PR does is to explicitly supply a ```scale``` argument with a specific value to ```makeARGB()``` such that ```makeARGB()``` does not use its own default. This causes rescaling to be bypassed, which was the intention of combining levels + lut in the first place.

PR #1693, opened 2021-04-05
Bypass makeARGB() for uint8, uint16 and float images

pijyoi:
This PR supercedes #1668.
8-bit grayscale images and 256-entry lut colormap images can skip makeARGB() entirely by using ```Format_Grayscale8``` and ```Format_Indexed8``` respectively instead of ```Format_ARGB32```.
For such a use-case, levels + lut combination to lut-only becomes an optimization and would be the fastest codepath available.

Sample program to compare performance against master.
Try changing the colormap and changing the levels.
```python
import pyqtgraph as pg
import numpy as np
app = pg.mkQApp()
imv = pg.ImageView()
imv.show()
size = (8192, 8192)
data = np.random.randint(256, size=size, dtype=np.uint8)
imv.setImage(data)
app.exec_()
```


pyqtgraph/functions.py
```diff
@@ -1333,6 +1333,47 @@ def try_fastpath_argb(xp, ain, aout, useRGBA):
     return True
 
 
+def _ndarray_to_qimage(arr, fmt):
```
j9ac9k, line 1336: we already have an `arrayToQPath` should we name this with a similar convention?  (I'm just asking, "no" is a perfectly good answer too :) )

j9ac9k:
I don't have a technical comment here, just wanted to say I appreciate these diffs @pijyoi as with your changes, especially with the comments, I feel like I'm starting to get a better handle on the image components of this library, which is an area I really don't use much in my own projects.

One thing that I had always wondered was how come we couldn't pass the data a bit more directly to the display, and you seem to have recognized that potential optimization

```python
            if is_passthru:
                 # both levels and lut are None
                 # these images are suitable for display directly
                 if image.ndim == 2:
                     fmt = QtGui.QImage.Format.Format_Grayscale8
                 elif image.shape[2] == 3:
                     fmt = QtGui.QImage.Format.Format_RGB888
                 elif image.shape[2] == 4:
                     fmt = QtGui.QImage.Format.Format_RGBA8888
```

As with most things image related, I'm going to make myself available for testing, but defer to @outofculture for providing much technical feedback.

Thanks again for the PR to further improve image performance within the library.

pyqtgraph/functions.py
pijyoi: One issue is whether functions.py is considered public API. ```ndarrayToQImage()``` "looks" too public.
I understand that camelCase is the convention in this library for public facing methods. How about for free standing functions?

j9ac9k: That's fair.  We don't have clear guidelines on naming conventions, and I think that's a fair assessment to say that cameCase convention is for more public facing methods. 

Unrelated to this PR but we should likely solicit input to form a consistent style guide for the library.  I'm good with leaving this as is 👍🏻

pijyoi:
The following table documents some useful combinations and the intended outcomes:

1) if levels are 2d, it will be extra slow path
2) if levels are 1d and the user supplied an alpha channel, the levels will also get applied onto the alpha channel. this is probably erroneous usage.
3) multichannel images are not supposed to have user-supplied lut.

| data   | channels| lvl_1d | lut | fmt        | remarks |
| --     |  --   |  --    | --  | --         | --      |
| uint8  | 1     |  N     |  N  | Grayscale8 |         |
| uint8  | 3     |  N     |  N  | RGB888     |         |
| uint8  | 4     |  N     |  N  | RGB8888    |         |
| uint8  | 1     |  Y     |     | Indexed8   |         |
| uint8  | 1     |        |  Y  | Indexed8   |         |
| uint8  | 3     |  Y     |  N  | ARGB32     | apply efflut only |
| uint16 | 1,3,4 |  N     |  N  | ARGB32     | apply levels only, probably not useful |
| uint16 | 1     |  N     |  Y  | ARGB32     | apply efflut only, probably not useful |
| uint16 | 1     |  Y     |     | ARGB32     | apply efflut only |
| uint16 | 3     |  Y     |  N  | ARGB32     | apply efflut only |
| float32, float64  | 1,3   |        |     | ARGB32     | levels and lut applied as necessary, nan locations made transparent |

pijyoi:
It turns out that ```lut.take(indices)``` is slower than ```lut[indices]```.
As usual, Linux runs faster than Windows on the same machine.

Windows
```python
In [83]: indices = np.random.randint(65536, size=(6144, 10240), dtype=np.uint16)
In [84]: lut = np.random.randint(256, size=65536, dtype=np.uint8)
In [85]: %timeit lut.take(indices)
285 ms ± 13.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
In [86]: %timeit lut[indices]
172 ms ± 209 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

WSL2
```python
In [7]: indices = np.random.randint(65536, size=(6144, 10240), dtype=np.uint16)
In [8]: lut = np.random.randint(256, size=65536, dtype=np.uint8)
In [9]: %timeit lut.take(indices)
248 ms ± 476 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
In [10]: %timeit lut[indices]
101 ms ± 241 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

outofculture:
I've previously found that much of the Windows/Linux difference is from anything that might call malloc (or whatever the modern equivalent is). What do your timings look like if you pre-allocate an `out=out`? Although, I don't actually know how to call `lut[indices]` with an output; `lut.__getitem__(indices)` won't accept `out=out`...

pijyoi:
Well, we could just test it out for ```np.take()``` with and w/o pre-allocation.

Windows
```python
In [101]: out = np.ones(indices.shape, dtype=lut.dtype)
In [103]: %timeit lut.take(indices)
292 ms ± 14.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
In [104]: %timeit lut.take(indices, out=out)
300 ms ± 9.92 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

WSL2
```python
In [11]: out = np.ones(indices.shape, dtype=lut.dtype)
In [12]: %timeit lut.take(indices)
246 ms ± 390 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
In [13]: %timeit lut.take(indices, out=out)
247 ms ± 933 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

outofculture:
Huh! I guess different functions are different? I really wouldn't have predicted a slowdown when pre-allocating, though. When I added the _processingBuffer to ImageItem, it had significant impact in Windows, but there's no arguing with observation.

pijyoi:
You can try this on Windows:
Open the task manager and look at the memory usage of Python process.
While running ```%timeit lut.take(indices, out=out)```, note that memory usage increases during the run.
While running ```%timeit np.add(indices, 0, out=indices)```, note that memory usage does not increase during the run.

This leads to the conclusion that the ```out``` parameter in ```np.take()``` is implemented as:
```python
def take_external(a, indices, out=None):
    res = take_internal(a, indices)    # always let numpy allocate
    if out is None:
        out = res
    else:
        out[:] = res
    return out
```
i.e. the ```out``` parameter is purely to have a uniform interface as the rest of the numpy library.

pijyoi:
Oops, ```lut[indices]``` is faster than ```lut.take(indices, axis=0)``` only for the case where lut is 1d. So massaging the rgb or rgba lut into 32-bits to make it into a 1d lut is still necessary.

pijyoi:
The last commit distinguishes between ```levels_lut``` and ```colors_lut```.
```levels_lut``` uses memory lookup to perform rescaling due to levels, effectively trading computation for memory accesses.
```colors_lut``` is the user-supplied lut. (Although named as "colors", it could also be grayscale)

If "colors_lut" has <= 256 entries, we would prefer to make use of Format_Indexed8, as it results in a smaller image size.
e.g. for the case where input data is uint16 with a color_lut of <= 256 entries, levels_lut is used to "rescale" the input data to uint8 and color_lut is set to be the ColorTable of the Indexed8 QImage. 

The code could be simplified if we delegate the case where color_lut > 256 entries to makeARGB(). This would make sense if such a use case is uncommon.

**Comment by pijyoi** [comment]
It turns out that color_lut with 512 entries is fairly common in pyqtgraph.
HistogramLUTItem, which ImageView uses, gives you a 512-entry color_lut for non-uint8 images.
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/HistogramLUTItem.py#L188-L201

**Comment by outofculture** [comment]
The only benchmarking we have right now is centered around makeARGB itself. Seeing as this PR is about bypassing makeARGB, those benchmarks will not measure this code. I don't want to accept this without an easy way to track and compare its performance. We'll need to either change the benchmark script to go one level higher and test ImageItem, or we'll need to move this code down into makeARGB somehow. @pijyoi do you have a sense of how hard the second option would be? My read is that it's pretty difficult. If you want to take a stab at either option, that would be great. Otherwise, I can hopefully get to it in a week or two.

**Comment by pijyoi** [comment]
The first option seems straightforward, replace calls to makeARGB() with ImageItem::setImage(autoLevels=False) followed by ImageItem::render(). Only thing is that I haven't used asv before.

Regarding the second option of integrating into makeARGB():
```
$ git grep -e "makeRGBA(" -e "makeARGB(" | grep -v -e tests -e functions -e benchmarks
examples/GLImageItem.py:tex1 = pg.makeRGBA(data[shape[0]//2], levels=levels)[0]       # yz plane
examples/GLImageItem.py:tex2 = pg.makeRGBA(data[:,shape[1]//2], levels=levels)[0]     # xz plane
examples/GLImageItem.py:tex3 = pg.makeRGBA(data[:,:,shape[2]//2], levels=levels)[0]   # xy plane
pyqtgraph/graphicsItems/ImageItem.py:        fn.makeARGB(image, lut=lut, levels=levels, output=self._processingBuffer)
pyqtgraph/widgets/RawImageWidget.py:            argb, alpha = fn.makeARGB(self.opts[0], *self.opts[1], **self.opts[2])
pyqtgraph/widgets/RawImageWidget.py:                self.image, alpha = fn.makeARGB(img, *args, **kwds)
```
There is only one real user of makeARGB() --> ImageItem.py
Yet, for display purposes, ImageItem doens't really need the conversion to the 4-channel ARGB format stage.
So, in my opinion, it is makeARGB() that needs to be broken up, and not to have more things integrated in.

**Comment by j9ac9k** [comment]
I'll just weigh in and say that I support doing the first option.  We aren't _really_ looking to maximize makeARGB performance, we care about how fast images update; which the overwhelming majority of that time is spent in makeARGB... if there are cases where profiling makeARGB no longer makes sense, and we can profile the next level up in ImageItem, that's probably what we should be doing.

**Comment by pijyoi** [comment]
A sample asv run for size 4096
```
[  0.00%] ·· Benchmarking existing-pyc__work_venv_allqt_scripts_python.exe
[  3.33%] ··· renderImageItem.Time4096Suite.time_10x_renderImageItem_float_levels_nolut                                877±0ms
[  6.67%] ··· renderImageItem.Time4096Suite.time_10x_renderImageItem_float_levels_uint16lut                            1.73±0s
[ 10.00%] ··· renderImageItem.Time4096Suite.time_10x_renderImageItem_float_levels_uint8lut                             1.66±0s
[ 13.33%] ··· renderImageItem.Time4096Suite.time_10x_renderImageItem_uint16_levels_nolut                               457±0ms
[ 16.67%] ··· renderImageItem.Time4096Suite.time_10x_renderImageItem_uint16_levels_uint16lut                           521±0ms
[ 23.33%] ··· renderImageItem.Time4096Suite.time_10x_renderImageItem_uint16_nolevels_nolut                             407±0μs
[ 26.67%] ··· renderImageItem.Time4096Suite.time_10x_renderImageItem_uint16_nolevels_uint16lut                         508±0ms
[ 30.00%] ··· renderImageItem.Time4096Suite.time_10x_renderImageItem_uint16_nolevels_uint8lut                          460±0ms
[ 33.33%] ··· renderImageItem.Time4096Suite.time_10x_renderImageItem_uint8_levels_nolut                               1.55±0ms
[ 36.67%] ··· renderImageItem.Time4096Suite.time_10x_renderImageItem_uint8_levels_uint16lut                           1.86±0ms
[ 40.00%] ··· renderImageItem.Time4096Suite.time_10x_renderImageItem_uint8_levels_uint8lut                            1.78±0ms
[ 43.33%] ··· renderImageItem.Time4096Suite.time_10x_renderImageItem_uint8_nolevels_nolut                              348±0μs
[ 46.67%] ··· renderImageItem.Time4096Suite.time_10x_renderImageItem_uint8_nolevels_uint16lut                         1.71±0ms
```

**Comment by pijyoi** [comment]
Regarding uint16 luts, it's less confusing to think of them as luts with between 257 to 65536 entries.
1) Thus using uint8 image data with a "uint16 lut" is really a usage error.
2) "uint16 luts" commonly contain much less than 65536 entries. HistogramLUTItem returns luts with either 256 or 512 entries.

**Comment by pijyoi** [comment]
With the last few commits, only uint16 (mono) images with colors_lut > 256 entries will use the levels + colors_lut combining "optimization".

**Comment by pijyoi** [comment]
>Would it be of interest/beneficial to get benchmark info on one of the Apple ARM M1 processors? I do have access to a machine now that I could run the benchmark suite. Qt will be releasing a macOS Universal Binary (native ARM support) version for Qt 6.2, but I believe numpy has native support right now.

Not so particularly useful. The render timings table was more a sanity check. That said, rendering timings is a pure computation measurement and doesn't take into account the painting step. The latter, I believe, would be quite dependent on the platform.

>While out of scope for this PR, but loosely related, we should add a blurb in the documentation for performance considerations when calling ImageItem.setImage While not all use-cases allow for flexibility here, many do, and it would be good to provide some guidance...

Off-hand (will update if I think of more):
1) Use row-major.
   * instantiate as ImageItem(axisOrder='row-major') or 
   * pg.setConfigOption('imageAxisOrder', 'row-major')
2) Image.setImage(data, autoLevels=False)
   * autoLevels is an "opt-out" parameter. It defaults to True if levels is not also set
3) Use C-contiguous image data
4) Use new version of numpy (1.20 has SIMD improvements, noticeable on Linux platforms)
5) enable use of numba with pg.setConfigOption('useNumba', True)
   * won't be useful for uint8 image data
   * not useful if you are only displaying 1 image, since JIT overhead is quite large. (and we didn't enable JIT caching on disk)
6) If using floating-point image data, prefer float32 to float64.
   * Many examples use np.random.{random, normal} which return float64
7) If using color lookup tables, use <= 256 entries
   * the colormaps included with pyqtgraph have 256 entries
   * ImageView widget will however interpolate it to 512 entries for non-uint8 image
8) Avoid using different levels per channel for RGB images. It is an unoptimized codepath

**Comment by outofculture** [comment]
Okay, I'm back. Is this ready @pijyoi?

**Comment by pijyoi** [comment]
Yes it's ready.

File: benchmarks/renderImageItem.py [hunk 2]
```diff
@@ -0,0 +1,122 @@
+import numpy as np
```
outofculture, line 1: This benchmark supercedes the makeARGB one, so we should just delete that file as part of this.

benchmarks/renderImageItem.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def make_test(dtype, kind, use_levels, lut_name, func_name):
+    def time_test(self):
+        data = getattr(self, dtype + "_data")
+        levels = data["levels"] if use_levels else None
+        lut = getattr(self, lut_name + "_lut", None) if lut_name is not None else None
+        if kind == "numba":
+            pg.setConfigOption("useNumba", True)
+        for _ in range(1):
```
outofculture, line 91: Priming the GPU is a much better choice; thanks for that improvement! We can get rid of all the cruft the 10x version needed.

benchmarks/renderImageItem.py
```diff
@@ -0,0 +1,122 @@
...
+        self.float_data, self.uint16_data, self.uint8_data, self.uint16_lut, self.uint8_lut = self._create_data(
+            size, np
+        )
+        if numba is not None:
+            # ensure JIT compilation
+            pg.setConfigOption("useNumba", True)
+            prime_numba()
+            pg.setConfigOption("useNumba", False)
+        if cp:
+            renderQimage(cp.asarray(self.uint16_data["data"]))  # prime the gpu
```
outofculture, line 55: typo. should be `renderQImage`

benchmarks/renderImageItem.py
pijyoi: Done. I have also changed floating point to use float32.

benchmarks/renderImageItem.py
pijyoi: Done.

benchmarks/renderImageItem.py
pijyoi: Fixed. I did run VideoSpeedTest.py on an old machine with cupy but not the asv benchmark.

outofculture:
So, the cupy-enabled benchmarks are showing a ~2-4x slowdown on this branch, depending on argument combinations. I'm having a lot of trouble _caring_, though, seeing as how the numba times often crush the old cupy times... But no, the main branch's cupy deserves to be protected. @pijyoi it sounds like working on cupy isn't convenient for you? I have time to take this on, if you want to just hand it off to me.

pijyoi:
>it sounds like working on cupy isn't convenient for you? I have time to take this on, if you want to just hand it off to me.

Yes, please do.
Not only am I not familiar with cupy, the machine on which I do have an Nvidia card is on PCIe 1.0, which is surely not representative of current machines.

pijyoi:
I suppose not converting the lut to a cupy ndarray is an oversight, as done in makeARGB():
```python
    if lut is not None and not isinstance(lut, xp.ndarray):
        lut = xp.array(lut)
```
Both benchmarks/renderImageItem.py and examples/VideoSpeedTest.py create luts in cuda memory, that's why they work.
ImageView.py would probably fail.

outofculture:
This got merged with #1786

PR #1694, opened 2021-04-05
DateAxisItem: Account for daylight saving time

dgoeries:
This should fix the initial setting of the _utcOffset_ to respect the daylight saving time. According to the official docs it is recommended to use `tm_zone results from localtime() to obtain timezone information.`

### DIFF
--- a/pyqtgraph/graphicsItems/DateAxisItem.py
+++ b/pyqtgraph/graphicsItems/DateAxisItem.py
@@ -222,7 +222,9 @@ class DateAxisItem(AxisItem):
 
         super(DateAxisItem, self).__init__(orientation, **kwargs)
         # Set the zoom level to use depending on the time density on the axis
-        self.utcOffset = utcOffset or get_utc_offset()
+        if utcOffset is None:
+            utcOffset = get_utc_offset()
+        self.utcOffset = utcOffset
         
         self.zoomLevels = OrderedDict([
             (np.inf,      YEAR_MONTH_ZOOM_LEVEL),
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/DateAxisItem.py
```diff
@@ -185,6 +185,17 @@ def tickValues(self, minVal, maxVal, minSpc):
              autoSkip=[1, 5, 10, 25])
 ], "99:99:99")
 
+
+def get_utc_offset():
```
j9ac9k, line 189: Can you rename to camelCase `getOffsetFromUtc` which should be more in-line w/ the Qt API: http://doc.qt.io/qt-5/qtimezone.html#offsetFromUtc

j9ac9k:
Hey @dgoeries thanks for doing this!  I was just thinking about this might have been getting neglected.

I haven't looked much at this functionality, but would migrating the timezone/date management from the python standard library to the Qt API be beneficial, or have you not looked at it sufficiently yet?

dgoeries:
I have not looked at it sufficiently, but this one would do great for a hotfix.

pyqtgraph/graphicsItems/DateAxisItem.py
dgoeries: thanks, done

j9ac9k:
That part of the Qt functionality was not in Qt4, which we supported when DateAxisItem was introduced.  But that's another topic for another time!

j9ac9k:
LGTM, thanks @dgoeries !

Issue #1253, opened 2020-06-13
Optionally use numba for CPU intensive operations


j9ac9k:
Migrating content from the slack channel to the issue tracker.

There are some operations in pyqtgraph that can be very expensive with large numpy arrays, such as accurate downsampling.  During the course of discussion, @campagnola suggested we implement methods that if they detect numba is installed, use it accordingly.

Example

```python
import timeit
import numpy as np
try:
    from numba import jit
    have_numba = True
except ImportError:
    def jit(**kwargs):
        return lambda fn: fn
    have_numba = False


def pkDownsample(x, y, ds):
    if have_numba:
        return pkDownsample_jit(x, y, ds)
    else:
        return pkDownsample_numpy(x, y, ds)


def pkDownsample_numpy(x, y, ds):
    n = len(x) // ds
    x1 = np.empty((n,2))
    x1[:] = x[:n*ds:ds,np.newaxis]
    x2 = x1.reshape(n*2)
    y1 = np.empty((n,2))
    y2 = y[:n*ds].reshape((n, ds))
    y1[:,0] = y2.max(axis=1)
    y1[:,1] = y2.min(axis=1)
    y2 = y1.reshape(n*2)
    return x2, y2


@jit(nopython=True)
def pkDownsample_jit(xSrc, ySrc, ds):
    n = len(xSrc) // ds
    xDst = np.empty((n*2), xSrc.dtype)
    yDst = np.empty((n*2), ySrc.dtype)
    for i in range(0, len(xDst) // 2):
        dstPtr = i * 2
        srcPtr = i * ds
        xDst[dstPtr] = xSrc[srcPtr]
        xDst[dstPtr+1] = xSrc[srcPtr]
        mn = ySrc[srcPtr]
        mx = ySrc[srcPtr]
        for j in range(1, ds):
            y = ySrc[srcPtr + j]
            mn = y if y < mn else mn
            mx = y if y > mx else mx
        yDst[dstPtr] = mx
        yDst[dstPtr+1] = mn
    return xDst, yDst



x = np.linspace(0, 1, 1000)
y = np.random.normal(size=1000)
ds = 7

x1, y1 = pkDownsample_numpy(x, y, ds)
x2, y2 = pkDownsample_numba(x, y, ds)
assert x1.shape == x2.shape
assert y1.shape == y2.shape
assert np.all(x1 == x2)
assert np.all(y1 == y2)


[...]

PR #1695, opened 2021-04-06
add usage of numba (for rescale)

pijyoi:
This PR adds optional usage of numba to pyqtgraph as permitted by #1253.
As a working example, a numba implementation for rescaleData() is included.
A fairly new version >= 0.53 of numba is required for the dynamic guvectorize functionality. As it happens, version 0.53 is the first version with a Python 3.9 wheel.

Benchmarking code with timings obtained on the same machine
```python
import time
import argparse
import numpy as np
import pyqtgraph as pg
from pyqtgraph.functions import rescaleData

parser = argparse.ArgumentParser()
parser.add_argument('--numba', action='store_true')
args = parser.parse_args()

if args.numba:
    try:
        pg.setConfigOptions(useNumba=True)
        import pyqtgraph.functions_numba
    except KeyError:
        pass

shape = 3072, 3072
data_uint8 = np.random.randint(256, size=shape, dtype=np.uint8)
data_float32 = np.random.random(shape).astype(np.float32)
niters = 20

if args.numba:
    # force the JIT
    rescaleData(data_uint8, 1.0, 0.0, dtype=np.uint8)
    rescaleData(data_float32, 1.0, 0.0, dtype=np.uint8)

for data in [data_uint8, data_float32]:
    for alpha in [0.1, 0.25]:
        if data.dtype.kind == 'u':
            minval, maxval = alpha * 255, (1 - alpha) * 255
        else:
            minval, maxval = alpha, 1 - alpha
        scale = 255 / (maxval - minval)
        offset = minval
        params = data, scale, offset, np.uint8

        t0 = time.perf_counter()
        for _ in range(niters):
            output = rescaleData(*params)
        t1 = time.perf_counter()
        print(data.dtype, alpha, '{:.6f}'.format((t1 - t0) / niters))

if args.numba:
    print(pyqtgraph.functions_numba.rescale_clip.types)
    print(pyqtgraph.functions_numba.rescale_noclip.types)
```

|    | Windows numpy 1.20.2 | Windows numba | WSL2 numpy 1.20.2 | WSL2 numba |
|--| -- | -- | -- |--|
|uint8 0.1   | 0.029979 | 0.007479 | 0.010434 | 0.006391 |
|uint8 0.25  | 0.030011 | 0.007832 | 0.010233 | 0.006176 |
|float32 0.1 | 0.044040 | 0.008983 | 0.011722 | 0.006383 |
|float32 0.25| 0.073928 | 0.008264 | 0.011681 | 0.006399 |

Windows platform benefits the most from the use of numba here.
Linux platform already has fairly good performance with numpy 1.20

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,7 +70,7 @@ jobs:
     - name: Install Dependencies
       run: |
         pip install --upgrade pip
-        pip install ${{ matrix.qt-version }} numpy${{ matrix.numpy-version }} scipy pyopengl h5py matplotlib
+        pip install ${{ matrix.qt-version }} numpy${{ matrix.numpy-version }} scipy pyopengl h5py matplotlib numba
         pip install .
         pip install pytest pytest-cov pytest-xdist coverage
     - name: "Install Linux VirtualDisplay"
--- a/pyqtgraph/__init__.py
+++ b/pyqtgraph/__init__.py
@@ -56,6 +56,7 @@ CONFIG_OPTIONS = {
                                  # The default is 'col-major' for backward compatibility, but this may
                                  # change in the future.
     'useCupy': False,  # When True, attempt to use cupy ( currently only with ImageItem and related functions )
+    'useNumba': False, # When True, use numba
 } 
 
 
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1124,7 +1124,8 @@ def rescaleData(data, scale, offset, dtype=None, clip=None):
         return data_out.astype(out_dtype, copy=False)
 
     numba_fn = getNumbaFunctions()
-    if numba_fn and out_dtype in (np.uint8, np.uint16):
+    if numba_fn and clip is not None:
+        # if we got here by makeARGB(), clip will not be None at this point
         return numba_fn.rescaleData(data, scale, offset, out_dtype, clip)
 
     return _rescaleData_nditer(data, scale, offset, work_dtype, out_dtype, clip)
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -17,6 +17,7 @@ import math
 
 import numpy as np
 from .util.cupy_helper import getCupy
+from .util.numba_helper import getNumbaFunctions
 
 from . import debug, reload
 from .Qt import QtGui, QtCore, QT_LIB, QtVersion
@@ -1121,8 +1122,12 @@ def rescaleData(data, scale, offset, dtype=None, clip=None):
 
         # don't copy if no change in dtype
         return data_out.astype(out_dtype, copy=False)
-    else:
-        return _rescaleData_nditer(data, scale, offset, work_dtype, out_dtype, clip)
+
+    numba_fn = getNumbaFunctions()
+    if numba_fn:
+        return numba_fn.rescaleData(data, scale, offset, out_dtype, clip)
+
+    return _rescaleData_nditer(data, scale, offset, work_dtype, out_dtype, clip)
 
 
 def applyLookupTable(data, lut):
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1124,7 +1124,7 @@ def rescaleData(data, scale, offset, dtype=None, clip=None):
         return data_out.astype(out_dtype, copy=False)
 
     numba_fn = getNumbaFunctions()
-    if numba_fn:
+    if numba_fn and out_dtype in (np.uint8, np.uint16):
         return numba_fn.rescaleData(data, scale, offset, out_dtype, clip)
 
     return _rescaleData_nditer(data, scale, offset, work_dtype, out_dtype, clip)
--- a/pyqtgraph/functions_numba.py
+++ b/pyqtgraph/functions_numba.py
@@ -1,0 +1,23 @@
+import numpy as np
+import numba
+
+@numba.guvectorize(
+    '(n),(),(),(),()->(n)', nopython=True)
+def rescale_clip(xx, scale, offset, vmin, vmax, yy):
+    for i in range(xx.size):
+        val = (xx[i] - offset) * scale
+        yy[i] = min(max(val, vmin), vmax)
+
+@numba.guvectorize(
+    '(n),(),()->(n)', nopython=True)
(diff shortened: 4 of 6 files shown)


j9ac9k:
Thanks @pijyoi for taking this on!  Those improvements 😍 


On macOS

| dtype        |  macOS numpy 1.20.1 | macos numba | speedup |
| :----------- | :-----------------  | :---------- | ------: |
| uint8 0.1    | 0.015515 			 | 0.012122    | 1.3x    |
| uint8 0.25   | 0.012500			 | 0.011960	   | 1.1x    |
| float32 0.1  | 0.039571			 | 0.012279    | 3.2x    |
| float32 0.25 | 0.055800			 | 0.012409    | 4.5x    |

I don't think using the minimum version of number as being the one to support python 3.9 is particularly a good criteria.  For example, PySide2 5.12.6 (which pyqtgraph supports) only works with up to Python 3.7.  That said I have no problem with setting the minimum version to the current version if there is some functionality need.

pijyoi:
The following script demonstrates a limitation of numba even with ahead-of-time compilation: output datatypes do not participate in signature selection. Caught by #1697
expected output ```[255, 256, 257]```, got ```[255, 0, 1]```
```python
import numpy as np
import numba

@numba.guvectorize(
    ['int32[:], uint8[:]',
     'int32[:], uint16[:]'],
    '(n)->(n)', nopython=True)
def simple(xx, yy):
    for i in range(xx.size):
        yy[i] = xx[i]

print(simple.types)

data = np.array([255, 256, 257], dtype=np.int32)
out = np.empty_like(data, dtype=np.uint16)
print(simple(data, out=out))
```

pijyoi:
>I don't think using the minimum version of numba as being the one to support python 3.9 is particularly a good criteria. For example, PySide2 5.12.6 (which pyqtgraph supports) only works with up to Python 3.7. That said I have no problem with setting the minimum version to the current version if there is some functionality need.

The dynamic guvectorize functionality available only since numba 0.53 is no longer in use, so the version check could be removed.

j9ac9k:
> > I don't think using the minimum version of numba as being the one to support python 3.9 is particularly a good criteria. For example, PySide2 5.12.6 (which pyqtgraph supports) only works with up to Python 3.7. That said I have no problem with setting the minimum version to the current version if there is some functionality need.
> 
> The dynamic guvectorize functionality available only since numba 0.53 is no longer in use, so the version check could be removed.

Ahh, I had misunderstood.   Too bad the guvectorize functionality didn't quite meet our needs.  I know very little about numba, I'm happy to do some more research here too to try and help out; admittedly the functions I'm looking at using numba on are the downsampling methods in PlotDataItem which are much simpler.

pijyoi:
>admittedly the functions I'm looking at using numba on are the downsampling methods in PlotDataItem which are much simpler.

Would the current implemented structure meet the needs of the example given in #1253?

j9ac9k:
> > admittedly the functions I'm looking at using numba on are the downsampling methods in PlotDataItem which are much simpler.
> 
> Would the current implemented structure meet the needs of the example given in #1253?

I don't see why the current implementation wouldn't work.  My original thought was to create something like an `@if_jit` decorator, where it would compile the function if `useNumba` was set to `True`,  otherwise do nothing

```python
@if_jit
def some_function():
    ...
```

I am not particularly experienced with numba, so I'm not in a position to really dictate what the implementation should look like, and for that, I would likely follow the lead of others more knowledgable than me (which is you know, just about everyone at this point).

pijyoi:
So I will wait for #1697 to be merged before rebasing and adding numba to it. Do I also add numba wheel as a dependency to the CI?

outofculture:
Sorry for the delay; #1697 is merged.

I believe you should try to add numba to the CI dependencies, yes, and we should make sure the relevant tests run with and without `useNumba`.

outofculture:
Also, I love where this is going, but I have a question about long-term direction. Numba is going to be most effective if we can get all the `makeARGB` operations down to a single loop. [I started](https://github.com/outofculture/pyqtgraph/blob/21e7f394d6a7eacce55a2e20ed689d8dc5613917/pyqtgraph/functions.py#L1000) trying to write that, but my attempt was ugly and a combinatorial mess, and I'm hoping a better approach exists. Do you think a piecemeal application of numba will be able to match that single-loop approach? Do you know of a better way to use numba to avoid the mess I attempted?

pijyoi:
>Do you think a piecemeal application of numba will be able to match that single-loop approach? Do you know of a better way to use numba to avoid the mess I attempted?

I really have little experience with numba and I don't have a feel on what sort of code it excels at compiling to significantly faster code. From the timings recorded in the 1st comment, using numpy 1.20 on Linux is competitive with numba. Windows and macOS start from a lower bar.

Numba will always be an optional dependency. If you end up having to write a special de-vectorized version of the code for numba to accept it, then it needs to give significant speedups for it to be worthwhile. Unfortunately, there is no way to tell until you have written that special version and benchmarked it.

Recently, I tried using numba on some other (unrelated to pyqtgraph) code. For one function, it gave maybe 25% speedup. For another function, it became slower. For both functions, I had to massage the code to make it pass though ```nopython=True``` mode. So anecdotally, I would be cautious about using numba.

By the way, numba will generate a RuntimeWarning if it sees NaNs in the data, even though the result is correct. Any way to mask it in the CI?

j9ac9k:
> By the way, numba will generate a RuntimeWarning if it sees NaNs in the data, even though the result is correct. Any way to mask it in the CI?

Yes, `pytest.ini` is where we store the warnings we filter out.  Alternatively if we don't want this warning emitted in the codebase, I believe there is a context manager that can be used to swallow warnings.  If memory serves we do this filtering in PlotDataItem.py (I'm on mobile I'll try and look up for an example later).

pijyoi:
While consulting the docs: https://docs.pytest.org/en/documentation-restructure/how-to/warnings.html
```
DeprecationWarning and PendingDeprecationWarning are hidden by the standard library by default so you have to explicitly configure them to be displayed in your pytest.ini:
```
Maybe pyqtgraph should enable the above two. It also looks like pytest.ini could do with some cleanup.

j9ac9k:
O, I hadn't noticed that before; you're right we should absolutely enable those.  I'll get to work cleaning up `pytest.ini`

PR #1697, opened 2021-04-06
Protect makeARGB with tests and benchmarks

outofculture:
In order to make optimizing makeARGB safer, this adds more thorough tests and some better ASV config. ASV probably won't work out of the gate on most systems and lacks instructions, so that's mostly for maintainer use for now.

Included is an automated test of CuPy, though of course our CI doesn't have CUDA-enabled GPUs to run them yet, so it just skips most places.

@pijyoi your numba branch could merge this branch to bring in test_makeARGB.py, to which you could add the `setConfigOption("useNumba", True)` code necessary to test your new code. The test fails for me when I run pytest on that. Feel free to ping me if you can't reproduce.


j9ac9k:
Hi @outofculture 

Thanks for making this PR; having an easier way to benchmark would go a very long ways!

We chatted about this in a different context, but would you be ok with doing away with the conda-forge/conda requirement and using virtualenv (I couldn't get venv to work but I didn't try that much).  This would make it easier for other to test as virtuanenv is a much easier dependency to have on hand than miniconda.

outofculture:
> doing away with the conda

Oh, right; I had forgotten. @j9ac9k Done!
```
pip install virtualenv
asv setup
asv run
```
Worked for me with the latest here.

pijyoi:
@outofculture , I can reproduce the numba failure and have made a interim workaround pending a better solution.

Now, I am just a beginner in numba but the issue seems to be that guvectorize does not create new specialised versions if the difference in function signature is only in the output dtype. Thus a compiled signature with lut_dtype uint8 was being used by subsequent tests using lut_dtype uint16. Kind of like C++ not being able to overload on return type.

outofculture:
Nice; I almost convinced myself "we probably don't need both 8- and 16-bit LUTs". If anyone thinks of other combinations or edge cases or data or whatnot to test, I can add those and regenerate the reference outputs ( I used mid-2019 master; maybe I should comment that? ).

pijyoi:
Oh another thing, output of makeARGB is endian dependent. So the tests would need some shim before comparison.

pijyoi:
>I almost convinced myself "we probably don't need both 8- and 16-bit LUTs"

This is non-obvious. In practice, the user would only use LUTs with a small number of entries, typically 256. However, within ImageItem.py, for uint16 images, the levels + lut combine logic will blow up the supplied small LUT to an effective LUT with 65536 entries, suitable for direct indexing.

Right now there are 4 variants of numba rescale : {uint8, uint16) x (clip, noclip), which is really too many.
makeARGB() does not actually invoke the noclip variants, so I am thinking of dropping those.
i.e. we should not optimize for codepaths that aren't actually invoked by pyqtgraph library, unless doing so is free.

j9ac9k:
@outofculture on my mac this blows up due to being unable to find `'cupy-cuda110'` (understandably cupy isn't available on macOS).  is there a way to handle optional dependencies in asv?

asv.conf.json
```diff
@@ -80,7 +82,8 @@
     //     cases: 'linux2', 'win32', 'cygwin', 'darwin'.
     //
     "exclude": [
```
j9ac9k, line 84: can we add here `{"sys_platform": "darwin", "cupy-cuda110": ""}` I was able to setup the environment that way on Darwin

asv.conf.json
```diff
@@ -40,7 +40,7 @@
 
     // The Pythons you'd like to test against.  If not provided, defaults
     // to the current version of Python used to run `asv`.
-    "pythons": ["2.7", "3.8"],
+    "pythons": ["3.8", "3.9"],
```
j9ac9k, line 43: Do we really want to test 2 versions of python here?   Perhaps should not provision and just run against the version `asv` is installed against.

asv.conf.json
```diff
@@ -53,9 +53,11 @@
     // followed by the pip installed packages).
     //
     "matrix": {
+//        "numpy": ["1.17", "1.18", "1.19", ""],
```
j9ac9k, line 56: I think testing against more numpy versions (latest we support vs. oldest we support) might be more interesting than python versions; but as it stands, the test suite takes long enough to run that probably don't need to worry about adding more test configurations.

benchmarks/makeARGB.py
```diff
@@ -1,72 +1,130 @@
...
     time_test.__name__ = func_name
     return time_test
 
 
-for dt in ['float', 'uint16', 'uint8']:
-    for levels in [True, False]:
-        for ln in [None, 'uint8', 'uint16']:
-            name = f'time_makeARGB_{dt}_{"" if levels else "no"}levels_{ln or "no"}lut'
-            setattr(TimeSuite, name, make_test(dt, levels, ln, name))
+for cupy in [True, False]:
```
j9ac9k, line 85: Can we make `cupy` here only have a `True` option  if cupy is installed?

asv.conf.json
outofculture: Yes; excludes would work if we want to pin each platform to the best available cuda version, and expect our runners to always have installed that version. We could be more flexible and install whatever version matches the output of `nvcc --version`. I'm thinking I should make a setup.py action that performs that test and re-writes the asv config file.

In any case, I should also write some documentation about how to run the benchmarks ( and why ).

asv.conf.json
outofculture: Yeah, agreed. If I can pare down the combinations, numpy would be the next variable I'd want to add.

benchmarks/makeARGB.py
outofculture: I hadn't imagined we'd have separate GPU and CPU benchmark systems, and I definitely want to see both measurements.

j9ac9k: oh I agree, we should have one benchmark system, for both.  What happens here with `True` for `cupy` is if you don't have `cupy` installed/available, half the tests just fail.    The idea here is that those tests don't run at all if cupy is not installed.

outofculture:
Okay, we need to reduce the number of distinct trials in the benchmarking. I think we want to be able to see:

- a range of image sizes
- cupy ( if available )
  - we may want to look at how different cuda versions perform, too?
- combinations of lut/levels/dtype/numpy that are representative of primary user experiences
- combinations of lut/levels/dtype/numpy that perform different from most other combinations
  - or really, a representative combo for each class of performance characteristics, as judged by the similarity of the curve across image size ( like was used in the VIS2021 paper )

We could just guess, but I'd prefer collecting some data first, to be able to discern those combo categories. We're going to want to collect data from a few different systems, to make sure OS and hardware are accounted for in our categorization. So in support of that, I want to merge the asv stuff with the full combinatoric explosion, get a few samples and later cut things out.

asv.conf.json
j9ac9k: so on docker/linux this isn't so much an issue, we can have the runner have any version of CUDA we want, up to the version of CUDA available on the host.  On Windows, it's a different story.

benchmarks/makeARGB.py
outofculture: Oh! Yes, I read your comment backwards ( or maybe sideways ). Thanks for elucidating; I agree now.

outofculture:
> Oh another thing, output of makeARGB is endian dependent. So the tests would need some shim before comparison.

I don't have access to any big-endian systems. If someone who does could make a recording ( by running `pyqtgraph.tests.test_makeARGB.save_reference` ), I would be happy to incorporate that.

pijyoi:
> I don't have access to any big-endian systems. If someone who does could make a recording ( by running `pyqtgraph.tests.test_makeARGB.save_reference` ), I would be happy to incorporate that.

You could reference the shim from 02a48afc60d99b242287b63b1fd7d451b7857bf5

outofculture:
Gah, sorry. I couldn't help but fix up the setup helpers a little. Only a little, though, I promise.

tools/setupHelpers.py
```diff
@@ -476,9 +476,57 @@ def getVersionStrings(pkg):
...
+
+    def run(self) -> None:
+        config = DEFAULT_ASV
+        try:
+            cuda_check = subprocess.check_output(["nvcc", "--version"])
+            match = re.search(r"release (\d\d.\d)", cuda_check.decode("utf-8"))
+            ver = match.groups()[0]  # e.g. 11.0
+            ver_str = ver[0] + ver[1] + ver[3]  # e.g. 110
+            config["matrix"][f"cupy-cuda{ver_str}"] = ""
+        except (FileNotFoundError, CalledProcessError):
```
j9ac9k, line 524: instead of try/except/pass why not run

```python
from contextlib import suppress

with suppress(FileNotFoundError, CalledProcessError):
   # what you have in the try block
```

tools/setupHelpers.py
```diff
@@ -476,9 +476,57 @@ def getVersionStrings(pkg):
...
+        pass
+
+    def finalize_options(self) -> None:
+        pass
+
+    def run(self) -> None:
+        config = DEFAULT_ASV
+        try:
+            cuda_check = subprocess.check_output(["nvcc", "--version"])
+            match = re.search(r"release (\d\d.\d)", cuda_check.decode("utf-8"))
```
j9ac9k, line 520: probably should change to `\d{1,2}` instead of `\d\d`, this will make it handle use case of older cuda versions such as 9.0 (or even older, on the off-chance we want to test those).

tools/setupHelpers.py
```diff
@@ -476,9 +476,57 @@ def getVersionStrings(pkg):
...
+            ver = match.groups()[0]  # e.g. 11.0
+            ver_str = ver[0] + ver[1] + ver[3]  # e.g. 110
```
j9ac9k, line 522: I apologize as you're walking into some issues I have with manipulating strings by their index position, which I personally try to avoid (not sure why) so feel free to disregard.

```python
ver_str = ver.replace(".", "", 1)  # the 1 at the end is for the number of times, alternatively, don't bound it, and just have `ver.replace(".", "")`
```

tools/setupHelpers.py
```diff
@@ -476,9 +476,57 @@ def getVersionStrings(pkg):
...
+            config["matrix"][f"cupy-cuda{ver_str}"] = ""
+        except (FileNotFoundError, CalledProcessError):
+            pass
+
+        with open("asv.conf.json", "w") as conf_file:
+            conf_file.write(json.dumps(config, indent=2))
```
j9ac9k, line 528: I love the idea of having the config as a dictionary that is dumped to a json BTW

tools/setupHelpers.py
outofculture: Ooo! I'd never seen that before. Love it!

j9ac9k: isn't going to python3 only grand?

tools/setupHelpers.py
```diff
@@ -476,9 +476,55 @@ def getVersionStrings(pkg):
...
+        pass
+
+    def run(self) -> None:
+        config = DEFAULT_ASV
+        with suppress(FileNotFoundError, subprocess.CalledProcessError):
+            cuda_check = subprocess.check_output(["nvcc", "--version"])
+            match = re.search(r"release (\d{1,2}\.\d)", cuda_check.decode("utf-8"))
+            ver = match.groups()[0]  # e.g. 11.0
+            ver_str = ver.replace(".", "")  # e.g. 110
+            config["matrix"][f"cupy-cuda{ver_str}"] = ""
```
j9ac9k, line 523: much better way to handle if cupy is available or not 👍🏻

PR #1700, opened 2021-04-08
Fix roi get array region

outofculture:
Taking over for #618 and using the fixed array region method in a few of the subclasses.


j9ac9k:
Thank you @malikolivier for the original PR, sorry we didn't get to it for so long.  

Thanks @outofculture for patching this up, this LGTM, merging!

PR #1701, opened 2021-04-09
avoid numpy scalar overflow

pijyoi:
the "offset" argument passed into rescaleData() is typically an element
of an ndarray, i.e. it comes from the lower bound value of the "levels"
ndarray. as such, arithmetic operations on it can overflow.

this triggers a runtime warning in the test suite.

the workaround is to convert it to a Python (integer) scalar.


j9ac9k:
LGTM, thanks @pijyoi !

PR #1706, opened 2021-04-10
promote use of '#' in mkPen

ntjess:


j9ac9k:
For the sake of documentation, will put why we're making this change.  Due to future attempts with having named colors, we've deprecated the use of hex-strings as color codes _unless_ it starts with a `#`.  While this change has been made everywhere in the library, this one example in the docstring got missed. 

Thanks @ntjess for the PR.

Issue #1692, opened 2021-04-05
PyQt6 6.1 flag -> enum possible breaking change


pijyoi:
PyQt6 6.1 will rename flag name (e.g. QtCore.Qt.Orientation**s**) to enum name (e.g. QtCore.Qt.Orientation).

See https://www.riverbankcomputing.com/pipermail/pyqt/2021-March/043713.html
and development snapshot Changelog at https://riverbankcomputing.com/software/pyqt/download

This would be a breaking change for the generated ui code, e.g.:
```bash
$ git grep Orientations
examples/exampleLoaderTemplate_pyqt6.py:        self.splitter.setOrientation(QtCore.Qt.Orientations.Horizontal)
pyqtgraph/canvas/CanvasTemplate_pyqt6.py:        self.splitter.setOrientation(QtCore.Qt.Orientations.Horizontal)
pyqtgraph/canvas/CanvasTemplate_pyqt6.py:        self.vsplitter.setOrientation(QtCore.Qt.Orientations.Vertical)
pyqtgraph/console/template_pyqt6.py:        self.splitter.setOrientation(QtCore.Qt.Orientations.Vertical)
pyqtgraph/graphicsItems/PlotItem/plotConfigTemplate_pyqt6.py:        self.gridAlphaSlider.setOrientation(QtCore.Qt.Orientations.Horizontal)
pyqtgraph/graphicsItems/PlotItem/plotConfigTemplate_pyqt6.py:        self.alphaSlider.setOrientation(QtCore.Qt.Orientations.Horizontal)
pyqtgraph/imageview/ImageViewTemplate_pyqt6.py:        self.splitter.setOrientation(QtCore.Qt.Orientations.Vertical)
```

What should be the way forward for pyqtgraph if the old name (Orientations) were removed in PyQt6 6.1?
1) drop support for PyQt6 < 6.1 from a particular commit?

j9ac9k:
Hi @pijyoi 

Can't tell if your issue got cut off, or you're only trying to present one option.

Those template files really drive me up the wall, wonder how feasible it would be to outright remove them, and any place that uses them in the library just loads the `.ui` file itself.

Also, given that Qt 6.1 is going to have so many more fixes/and more functionality (eventually Qt Multimedia is being rolled back in) I'm not opposed to dropping Qt 6.0 support; but to do so because of some template files seems a bit silly.

j9ac9k:
[...]

PR #1707, opened 2021-04-11
add PyQt6 6.1 forwards compatibility

pijyoi:
This addresses #1692 to provide forwards compatibility to PyQt6 6.1.

It adds a temporary shim for the renamed enums such that when PyQt6 6.1 is released, it will run on the ui files generated using PyQt6 6.0. Only forwards-compatibility is provided. Backwards-compatibility is deliberately not provided.

The decision by pyqtgraph library to drop PyQt6 6.0 support would coincide with the regeneration of the ui files using PyQt6 6.1 (and also removal of the temporary shim.)

### DIFF
--- a/pyqtgraph/Qt.py
+++ b/pyqtgraph/Qt.py
@@ -376,6 +376,15 @@ if QT_LIB == PYQT6:
         QtCore.QEvent.type = new_method
         del new_method
 
+    # PyQt6 6.1 renames some enums and flags to be in line with the other bindings.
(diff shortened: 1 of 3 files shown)


j9ac9k:
Thanks for this @pijyoi

Looks like the final release is scheduled for [April 27th](https://wiki.qt.io/Qt_6.1_Release) as of right now. 

Hopefully they stop with these changes so we can minimize these changes in the future :|

PR #1709, opened 2021-04-11
GridItem: Add custom line width based on pen option.

nodedge:
Hello, 

Thank you for this awesome library.

We wish to contribute to it via this small PR.

The point is to customize the width of the GridItem lines.

Happy review!

### DIFF
--- a/examples/GraphicsScene.py
+++ b/examples/GraphicsScene.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 ## Add path to library (just for examples; you do not need this)
+from PySide2.QtGui import QPen, QColor
+
 import initExample
 
 from pyqtgraph.Qt import QtCore, QtGui
@@ -15,27 +17,27 @@ class Obj(QtGui.QGraphicsObject):
     def __init__(self):
         QtGui.QGraphicsObject.__init__(self)
         GraphicsScene.registerObject(self)
-        
+
     def paint(self, p, *args):
         p.setPen(pg.mkPen(200,200,200))
         p.drawRect(self.boundingRect())
-        
+
     def boundingRect(self):
         return QtCore.QRectF(0, 0, 20, 20)
-        
+
     def mouseClickEvent(self, ev):
         if ev.double():
             print("double click")
         else:
             print("click")
         ev.accept()
-        
+
     #def mouseDragEvent(self, ev):
         #print "drag"
         #ev.accept()
         #self.setPos(self.pos() + ev.pos()-ev.lastPos())
-        
-        
+
+
 
 vb = pg.ViewBox()
 win.setCentralItem(vb)
@@ -56,6 +58,9 @@ prox.setPos(100,0)
 vb.addItem(prox)
 
 g = pg.GridItem()
+p = QPen(QColor(0, 255, 255))
+p.setWidthF(1)
+g.opts["pen"] = p
 vb.addItem(g)
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


j9ac9k:
Hi @nodedge,

Thanks for the PR.  I'm in support of the feature, but in its current implementation it breaks default behavior. 

For example, running `python example/GraphicsScene.py` on this PR.

image


Same thing but on `master` right now:

image

This can likely be worked around by checking if the pen has a default width, in which case set the pen width to `self.pixelWidth()` but otherwise set to the width you're working with.

Or alternatively a "better" implementation IMO would be to multiply the pen width by `self.pixelWidth()`; as it makes sense to have the pen width of the GridItem be in screen pixels, and have nothing to do with the coordinate system it's in.

@NilsNemitz @ixjlyons do either of you have input.

ixjlyons:
Probably the simplest way to handle this is to just extract the width in `setPen` and use that in `generatePicture`. Something like:

```diff
diff --git a/pyqtgraph/graphicsItems/GridItem.py b/pyqtgraph/graphicsItems/GridItem.py
index efb1afdb..ff9f8b6f 100644
--- a/pyqtgraph/graphicsItems/GridItem.py
+++ b/pyqtgraph/graphicsItems/GridItem.py
@@ -35,6 +35,7 @@ class GridItem(UIGraphicsItem):
             self.opts['pen'] = fn.mkPen(*args, **kwargs)
 
         self.picture = None
+        self._penWidth = self.opts['pen'].width()
         self.update()
 
 
@@ -157,7 +158,6 @@ class GridItem(UIGraphicsItem):
 
                 linePen = self.opts['pen']
                 lineColor = self.opts['pen'].color()
-                lineWidth = self.opts['pen'].width()
                 lineColor.setAlpha(c)
                 linePen.setColor(lineColor)
 
@@ -171,9 +171,9 @@ class GridItem(UIGraphicsItem):
                 for x in range(0, int(nl[ax])):
                     linePen.setCosmetic(False)
                     if ax == 0:
-                        linePen.setWidthF(lineWidth)
+                        linePen.setWidthF(self._penWidth*self.pixelWidth())
                     else:
-                        linePen.setWidthF(lineWidth)
+                        linePen.setWidthF(self._penWidth*self.pixelHeight())
                     p.setPen(linePen)
                     p1 = np.array([0.,0.])
                     p2 = np.array([0.,0.])
```

NilsNemitz:
@nodedge , @j9ac9k 
Without having the time to check in detail:
It might also be useful to make sure that the `QPen.setCosmetic()` gets set?
see: https://doc.qt.io/qt-5/qpen.html#isCosmetic
I assume that the line width is supposed to be constant, and not a function of the zoom-level?

j9ac9k:
Ahh yes, GridItem pens should most certainly be cosmetic.

ixjlyons:
It's [explicitly set to False](https://github.com/pyqtgraph/pyqtgraph/blob/6f74d56266dc452d2b30085008ab274eff8bca33/pyqtgraph/graphicsItems/GridItem.py#L171) - not sure why.

nodedge:
Thank you for your answers.
@ixjlyons I have just tried your solution.

The example "GraphicsScene" is not broken anymore. 
I added a custom pen in this same example, but the gridline width does not seem to be modified.
Instead, there is more space between the lines.
If we maximize the window, the grid spacing seems to be reset.

j9ac9k:
I think I got to the bottom of this issue.  The problem is this line I believe:

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/GridItem.py#L152

Since QPen's are mutable, we change the width of the pen's further down here:

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/GridItem.py#L166-L169

That change in width sort of throws out the initial setting we made.

We can work around this by modifying 

```python
                linePen = fn.mkPen(self.opts['pen'])
```

This returns a copy of the pen, where the width that was initially set to is preserved.

@nodedge would you be ok with me pushing my changes to your branch?

ixjlyons:
I think this can be closed now, see #2304 

Thanks for your efforts @nodedge

PR #1710, opened 2021-04-11
run pyqtgraph tests before examples

pijyoi:
Recently, the CI has been failing frequently, crashing on Linux / Python 3.8 / PySide2 at ```test_ref_cycles.py::test_ImageView()```. This appears to be a case of being at the wrong place at the wrong time, i.e. automatic garbage collection got triggered around that point.

By arranging to have the pyqtgraph tests run before the examples, this particular crash seems to have gone away.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -106,7 +106,7 @@ jobs:
     - name: Run Tests
       run: |
         mkdir $SCREENSHOT_DIR
-        pytest . -v \
+        pytest pyqtgraph examples -v \
           -n auto \
           --junitxml pytest.xml \
           --cov pyqtgraph --cov-report=xml --cov-report=html


j9ac9k:
Thanks for this @pijyoi.  Just curious how did you determine the importance of the ordering between examples and the regular test suite?

pijyoi:
I tried out this comment about how to trigger garbage collection before each test (https://github.com/pytest-dev/pytest/discussions/8153#discussioncomment-214812) within ```test_ref_cycles.py```
That moved the crash to the first test ```test_PlotWidget()``` and ```test_ImageView()``` passed instead, so that seemed to point to the crash possibly being caused by automatic garbage collection getting triggered by the rapid creation and destruction of objects in ```test_ref_cycles.py```. 

To be clear, I don't think automatic garbage collection was causing a crash with pyqtgraph. I think it was causing the test machinery, which itself also uses Python, to crash.

Reordering ```pyqtgraph``` before ```examples``` was the next random thing I tried.

j9ac9k:
Those green checkmarks 😍 

This LGTM, thanks @pijyoi for looking into this!

Merging.

PR #663, opened 2018-04-13
feature GLGradientLegendItem

feketeimre:
Adds a hud like colorbar to the 3D view.

PR #1711, opened 2021-04-12
Add GLGradientLegendItem (Reimplement #663)

j9ac9k:
This PR takes #663; where I merged master, and then fixed a handful of issues.  Unfortunately, the example is still not running as intended, maybe someone with more knowledge of the 3D/OpenGL functionality of pyqtgraph can give some input.

### DIFF
--- a/examples/GLGradientLegendItem.py
+++ b/examples/GLGradientLegendItem.py
@@ -1,0 +1,43 @@
+import initExample
+
+import pyqtgraph as pg
+import pyqtgraph.opengl as gl
+import numpy
+
+app = pg.mkQApp()
+w = gl.GLViewWidget()
+w.show()
+w.setWindowTitle("pyqtgraph example: GLGradientLegendItem")
+w.setCameraPosition(distance=60)
+
+gx = gl.GLGridItem()
+gx.rotate(90, 0, 1, 0)
+w.addItem(gx)
+
+md = gl.MeshData.cylinder(rows=10, cols=20, radius=[5.0, 5], length=20.0)
+md._vertexes[:, 2] = md._vertexes[:, 2] - 10
+
+# set color based on z coordinates
+color_map = pg.colormap.get("CET-L10")
+
+h = md.vertexes()[:, 2]
+# remember these
+h_max, h_min = h.max(), h.min()
+h = (h - h_min) / (h_max - h_min)
+colors = color_map.map(h, mode="float")
+md.setFaceColors(colors)
+m = gl.GLMeshItem(meshdata=md, smooth=True)
+w.addItem(m)
+
+legendLabels = numpy.linspace(h_max, h_min, 5)
+legendPos = numpy.linspace(1, 0, 5)
+legend = dict(zip(map(str, legendLabels), legendPos))
+
+gll = gl.GLGradientLegendItem(
+    pos=(10, 10), size=(50, 300), gradient=color_map, labels=legend
+)
+w.addItem(gll)
+
+## Start Qt event loop unless running in interactive mode.
+if __name__ == "__main__":
+    pg.exec()
--- a/pyqtgraph/opengl/items/GLGradientLegendItem.py
+++ b/pyqtgraph/opengl/items/GLGradientLegendItem.py
@@ -1,0 +1,107 @@
+from ... Qt import QtCore, QtGui
+from ... import functions as fn
+from OpenGL.GL import *
+from ..GLGraphicsItem import GLGraphicsItem
+
+__all__ = ['GLGradientLegendItem']
+
+class GLGradientLegendItem(GLGraphicsItem):
+    """
+    Displays legend colorbar on the screen.
+    """
+
+    def __init__(self, **kwds):
+        """
+        Arguments:
+            pos: position of the colorbar on the screen, from the top left corner, in pixels
+            size: size of the colorbar without the text, in pixels
+            gradient: a pg.ColorMap used to color the colorbar
+            labels: a dict of "text":value to display next to the colorbar.
+                The value corresponds to a position in the gradient from 0 to 1.
+            fontColor: sets the color of the texts
+            #Todo:
+                size as percentage
+                legend title
+        """
+        GLGraphicsItem.__init__(self)
+        glopts = kwds.pop("glOptions", "additive")
+        self.setGLOptions(glopts)
+        self.pos = (10, 10)
+        self.size = (10, 100)
+        self.fontColor = (1.0, 1.0, 1.0, 1.0)
+        self.stops = None
+        self.colors = None
+        self.gradient = None
+        self.setData(**kwds)
+
+    def setData(self, **kwds):
+        args = ["size", "pos", "gradient", "labels", "fontColor"]
+        for k in kwds.keys():
+            if k not in args:
+                raise Exception(
+                    "Invalid keyword argument: %s (allowed arguments are %s)"
+                    % (k, str(args))
+                )
+
+        self.antialias = False
+
+        for arg in args:
+            if arg in kwds:
+                setattr(self, arg, kwds[arg])
+
+        ##todo: add title
+        ##todo: take more gradient types
+        if self.gradient is not None and hasattr(self.gradient, "getStops"):
+            self.stops, self.colors = self.gradient.getStops("float")
+            self.qgradient = self.gradient.getGradient()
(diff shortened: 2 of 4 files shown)


pyqtgraph/opengl/items/GLGradientLegendItem.py
```diff
@@ -0,0 +1,93 @@
...
+
+        #draw labels
+        #scaling and translate doent work on rendertext
+        glColor4f(*self.fontColor)
+        for labelText,labelPosition in self.labels.items():
+            ## todo: draw ticks, position text properly
+            x = 1.1*self.size[0]+self.pos[0]
+            y = self.size[1]- labelPosition *self.size[1]+self.pos[1]+8
+            ##todo: fonts
+            # self.view().renderText(x,y,labelText)
```
j9ac9k, line 86: Had to comment out this method, as there is no `renderText` method in in GLViewWidget.  Best I can figure, this method was never a part of GLViewWidget.

examples/GLGradientLegendItem.py
```diff
@@ -0,0 +1,44 @@
...
+
+# set color based on z coordinates
+cmap = pg.ColorMap((0,.25,.5,.75,1),((0,0,255),(0,255,255),(0,255,0),(255,255,0),(255,0,0)))
+h = md.vertexes()[:,2]
+#remember these
+hmax, hmin = h.max(), h.min()
+h = h-hmin
+h = h/h.max()
+colors = cmap.map(h)/255.0
+colors = numpy.c_[colors,numpy.ones(len(colors))]
```
ixjlyons, line 29: Looks like this is the problem. `ColorMap` already gives back RGBA so the extra ones aren't needed.

examples/GLGradientLegendItem.py
```diff
@@ -0,0 +1,44 @@
...
+md._vertexes[:,2] = md._vertexes[:,2]-10
+
+# set color based on z coordinates
+cmap = pg.ColorMap((0,.25,.5,.75,1),((0,0,255),(0,255,255),(0,255,0),(255,255,0),(255,0,0)))
+h = md.vertexes()[:,2]
+#remember these
+hmax, hmin = h.max(), h.min()
+h = h-hmin
+h = h/h.max()
+colors = cmap.map(h)/255.0
```
ixjlyons, line 28: Can use `map(h, mode="float")` and get rid of the division by 255 for clarity. I dunno, maybe that's not really that much clearer.

examples/GLGradientLegendItem.py
```diff
@@ -0,0 +1,44 @@
...
+md = gl.MeshData.cylinder(rows=10, cols=20, radius=[5., 5], length=20.)
+md._vertexes[:,2] = md._vertexes[:,2]-10
+
+# set color based on z coordinates
+cmap = pg.ColorMap((0,.25,.5,.75,1),((0,0,255),(0,255,255),(0,255,0),(255,255,0),(255,0,0)))
+h = md.vertexes()[:,2]
+#remember these
+hmax, hmin = h.max(), h.min()
+h = h-hmin
+h = h/h.max()
```
ixjlyons, line 27: Putting this all one one line `h = (h - hmin) / (hmax - hmin)` might make it more obvious it's just normalizing the height.

pyqtgraph/opengl/items/GLGradientLegendItem.py
ixjlyons: Looks like it would've worked before moving to `QOpenGLWidget`: https://doc.qt.io/qt-5/qglwidget.html#renderText

Some ideas for replacing it: https://stackoverflow.com/questions/28216001/how-to-render-text-with-qopenglwidget

j9ac9k: Didn't even dawn on me to consider looking at QGLWidget; thanks for the links.

j9ac9k:
A-ha, thanks @ixjlyons that was indeed it (the issue w/ the colors)

image

Still need to sort out the text situation, and do a bit more cleanup, and of course, we can't show-case the jet-colormap.

examples/GLGradientLegendItem.py
j9ac9k: Nice find, here I was pulling my hair out looking at everywhere except the example.

j9ac9k:
@ixjlyons I implemented the renderText method as best I could based on the SO post you linked (nice find by the way!).  Unfortunately I'm not actually seeing any text; but it runs without error.  Might taking another look?

pijyoi:
Reading https://doc.qt.io/qt-5/qopenglwidget.html#painting-techniques, it would seem to say that if you want to mix QPainter and OpenGL, then OpenGL calls need to be wrapped between ```beginNativePainting()``` and ```endNativePainting()```. Quite a big change needed (and not even certain that would eventually work.)

pijyoi:
I patched up GLGradientLegendItem.py so that the text gets displayed.
Previous changes of this PR to GLViewWidget.py can be dropped. (Lots of conflicts anyway)
resized


GLGradientLegendItem.py
```python
from ... Qt import QtGui
from OpenGL.GL import *
from ..GLGraphicsItem import GLGraphicsItem


class GLGradientLegendItem(GLGraphicsItem):
    """
    Displays legend colorbar on the screen.
    """

    def __init__(self, **kwds):
        """
        Arguments:
            pos: position of the colorbar on the screen, from the top left corner, in pixels
            size: size of the colorbar without the text, in pixels
            gradient: a pg.ColorMap used to color the colorbar
            labels: a dict of "text":value to display next to the colorbar.
                The value corresponds to a position in the gradient from 0 to 1.
            fontColor: sets the color of the texts
            #Todo:
                size as percentage
                legend title
        """
        GLGraphicsItem.__init__(self)
        glopts = kwds.pop("glOptions", "additive")
        self.setGLOptions(glopts)
        self.pos = (10, 10)
        self.size = (10, 100)
        self.fontColor = (1.0, 1.0, 1.0, 1.0)
        self.stops = None
        self.colors = None
        self.gradient = None
        self.setData(**kwds)

    def setData(self, **kwds):
        args = ["size", "pos", "gradient", "labels", "fontColor"]
        for k in kwds.keys():
            if k not in args:
                raise Exception(
                    "Invalid keyword argument: %s (allowed arguments are %s)"
                    % (k, str(args))
                )

        self.antialias = False

        for arg in args:
            if arg in kwds:
                setattr(self, arg, kwds[arg])

        ##todo: add title
        ##todo: take more gradient types
        if self.gradient is not None and hasattr(self.gradient, "getStops"):
            self.stops, self.colors = self.gradient.getStops("float")
        self.update()

    def paint(self):
        if self.pos is None or self.stops is None:
            return
        self.setupGLState()
        glMatrixMode(GL_PROJECTION)
        glPushMatrix()
        glLoadIdentity()
        glOrtho(0.0, self.view().width(), self.view().height(), 0.0, -1.0, 10.0)
        glMatrixMode(GL_MODELVIEW)
        glLoadIdentity()
        glDisable(GL_CULL_FACE)

        glClear(GL_DEPTH_BUFFER_BIT)

        # draw the colorbar
        glTranslate(self.pos[0], self.pos[1], 0)
        glScale(self.size[0], self.size[1], 0)
        glBegin(GL_QUAD_STRIP)
        for p, c in zip(self.stops, self.colors):
            glColor4f(*c)
            glVertex2d(0, 1 - p)
            glColor4f(*c)
            glVertex2d(1, 1 - p)
        glEnd()

        # draw labels
        # scaling and translate doesnt work on rendertext
        fontColor = QtGui.QColor(*[x * 255 for x in self.fontColor])

        painter = QtGui.QPainter(self.view())
        painter.setPen(fn.mkPen(fontColor))
        for labelText, labelPosition in self.labels.items():
            ## todo: draw ticks, position text properly
            x = 1.1 * self.size[0] + self.pos[0]
            y = self.size[1] - labelPosition * self.size[1] + self.pos[1] + 8
            ##todo: fonts
            painter.drawText(x, y, labelText)
        painter.end()

        glMatrixMode(GL_PROJECTION)
        glPopMatrix()
        glMatrixMode(GL_MODELVIEW)
```

j9ac9k:
Rebased incorporating the changes to `test_examples.py` made in #1892

j9ac9k:
Thank you @feketeimre for the original submission, thank you @pijyoi for re-implementing it!  Merging.

PR #1715, opened 2021-04-14
small TargetItem fixes

outofculture:
Playing around with TargetItem more, I found a couple of errors:

```
    |    File "/home/martin/src/acq4/acq4/acq4/devices/Pipette/pipette.py", line 500, in __init__
    |      self.target.setLabelAngle(dev.yawAngle())
    |    File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/TargetItem.py", line 357, in setLabelAngle
    |      if self.label() is not None and angle != self.label().angle():
    |  TypeError: 'int' object is not callable
    |==============================<<
```

Which is deprecated, so no biggie, but we may as well fix it.

This next one I couldn't make an isolated test of, but the fix was just reorganizing the `__init__` method:

```
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/GraphicsObject.py", line 23, in itemChange
    self.parentChanged()
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/GraphicsItem.py", line 472, in parentChanged
    self._updateView()
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/GraphicsItem.py", line 529, in _updateView
    self.viewTransformChanged()
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/graphicsItems/TargetItem.py", line 452, in viewTransformChanged
    self.offset.x() * viewPixelSize[0], self.offset.y() * viewPixelSize[1]
AttributeError: 'TargetLabel' object has no attribute 'offset'
```

I tweaked the example, too, just to exercise the class a little.

**Edited** to not have quite so much irrelevant stack.

### DIFF
--- a/examples/InfiniteLine.py
+++ b/examples/InfiniteLine.py
@@ -32,13 +32,7 @@ p1.addItem(inf1)
 p1.addItem(inf2)
 p1.addItem(inf3)
 
-targetItem1 = pg.TargetItem(
-    label=True,
-    symbol="crosshair",
-    labelOpts={
-        "angle": 0
-    }
-)
+targetItem1 = pg.TargetItem()
 
 targetItem2 = pg.TargetItem(
     pos=(30, 5),
@@ -47,10 +41,10 @@ targetItem2 = pg.TargetItem(
     symbol="star",
     pen="#F4511E",
     labelOpts={
-        "angle": 45,
         "offset": QtCore.QPoint(15, 15)
     }
 )
+targetItem2.label().setAngle(45)
 
 targetItem3 = pg.TargetItem(
     pos=(10, 10),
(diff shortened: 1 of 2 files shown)


j9ac9k:
blerg `TextItem.angle` is not a method 🤦🏻  

Thanks for cleaning this up on my behalf @outofculture

outofculture:
No worries; I thought I had tested it beforehand, but somehow I missed these.

j9ac9k:
this points out some API inconsistencies we have w/ Qt (similar issues we have with PlotCurveItem).

Unfortunately to make the API change here will be quite breaking; ... as I don't know how we can gradually change something from a property to method, and have an overlapping period where both methods work.... but this issue has nothing to do w/ this PR, and likely something that will be a painful break.

Issue #1718, opened 2021-04-16
ColorMap.getLookupTable() returns int32 instead of byte


pijyoi:
### Short description

ColorMap.getLookupTable() returns dtype int32 although it is supposed to return dtype uint8.
Data values returned in fact do fit within uint8.

### Code to reproduce

```python
import pyqtgraph as pg

cmap = pg.colormap.get('CET-L8')
lut = cmap.getLookupTable(nPts=256, mode='float')
print(lut.dtype, lut.shape, lut.min(), lut.max())    # float64 (256, 3) 0.00392156862745098 0.9921568627450981
lut = cmap.getLookupTable(nPts=256, mode='byte')
print(lut.dtype, lut.shape, lut.min(), lut.max())    # int32 (256, 3) 1 253
```

### Expected behavior

A uint8 lut should be returned.

### Real behavior

A uint32 lut got returned.

### Tested environment(s)

 * PyQtGraph version: 0.12.1
 * Qt Python binding: 
 * Python version: PySide6 6.0.2 Qt 6.0.2
 * NumPy version: 1.20.2
 * Operating system: Windows 10
 * Installation method: pip install -e .

### Additional context

This causes a pessimization during rendering of an ImageItem. After application of the colormap, we end up with a data array that is 4 times larger than it should have been.

pijyoi:
Fixed by #1722

PR #1722, opened 2021-04-17
ColorMap: keep RGBA values as uint8, clean up docstrings

NilsNemitz:
As @pijyoi points out in #1718, ColorMap does not actually provide byte values when in byte mode.

This originates in the very first conversion of color data, which passes it through `mkColor()` and then gets the RGBA values as long form integers by `QColor.getRgb()`.

This PR casts the data to np.unit8 at this stage, and this change seems to correctly propagate all the way to `getLookupTable()`
An alternative approach would be to store the data as floats from `QColor.getRgbF()` and convert to uint8 only when requested.
That would -in principle- allow working with higher bit-depth color data without changing the current behavior. Is that preferable?

I have tried to clean up the docstrings a little. They were promising HSV interpolations that do not actually exist.
The `mode` parameter now raises a deprecation warning.
I have also removed the explicit `mode='rgb'` setting from the nonuniform image example.

### DIFF
--- a/pyqtgraph/colormap.py
+++ b/pyqtgraph/colormap.py
@@ -222,19 +222,20 @@ class ColorMap(object):
 
     def __init__(self, pos, color, mapping=CLIP, mode=None): #, names=None):
         """
-        ===============     =================================================================
+        ===============     =======================================================================
         **Arguments:**
         pos                 Array of positions where each color is defined
         color               Array of colors.
                             Values are interpreted via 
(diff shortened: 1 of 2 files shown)


j9ac9k:
This looks good to me as is.  I'll give some time for some other folks to chime in.

j9ac9k:
LGTM, thanks for this PR @NilsNemitz merging!

PR #1724, opened 2021-04-18
Use math module methods for scalars

j9ac9k:
Using numpy methods that are intended for vectorized operations is substantially slower than using the math module, so when feasible the math module methods should be used.

When benchmarking the results, I saw ~10% performance improvement in various methods in the optics demo.  In terms of the regular library, the benefit I saw in the paint speed of `InfiniteLine.py` was quite minimal unfortunately.  

Given that I had already done this work, figured I would submit it for others to take comment on and eventually merge if deemed worthwhile.


examples/GLLinePlotItem.py
```diff
@@ -30,14 +30,14 @@
 w.addItem(gz)
 
 def fn(x, y):
```
pijyoi, line 32: This function is not used.

examples/GLLinePlotItem.py
```diff
@@ -30,14 +30,14 @@
...
-    return np.cos((x**2 + y**2)**0.5)
+    return np.cos(np.hypot(x, y))
 
 n = 51
 y = np.linspace(-10,10,n)
 x = np.linspace(-10,10,100)
 for i in range(n):
     yi = np.array([y[i]]*100)
```
pijyoi, line 39: The following syntax looks weird: A single element of an ndarray is extracted, replicated as a Python list, then converted back to an ndarray.
```python
yi = np.array([y[i]]*100)
```

Maybe numpy broadcasting could be used:
```python
yi = y[i]
d = np.hypot(x, yi)
z = 10 * np.cos(d) / (d+1)
pts = np.column_stack([x, np.full_like(x, yi), z])
```

On the other hand, with the use of hypot, it's even less obvious what function is being plotted.

examples/GLLinePlotItem.py
j9ac9k: Realized this just after I committed the code.

examples/GLLinePlotItem.py
j9ac9k: Your suggestion performed better, so I'll incorporate it, given this is an example and its functional, I see no reason not to make a change.

j9ac9k:
Thanks @pijyoi for taking a look.  You're far more experienced with numpy than I am.  If you notice any other changes that should be made to this library along the aims this PR means to address (use `math` module for singular values, use stable methods `hypot` instead of explicit calculations), let me know.

j9ac9k:
I think this PR is ready, but if anyone can think of other functions I should check for here, I'll gladly explore those changes too.

The only other math module optimization that I could think of, is adopting `pg.Point` to use complex-numbers, where the imaginary component is on the y-axis.  I think that's a big enough of a departure that exploration of that should be part of a separate discussion.

j9ac9k:
I changed use-cases to `math.isfinite` and `math.isnan` where appropriate.  I think the PR is ready for review in case anyone wants to look things over.

j9ac9k:
codeql found 4 new errors and 2 new warnings, so I guess I'll patch those up too...

j9ac9k:
CI, go home, you're drunk....

j9ac9k:
Converting to draft to prevent merging, ...getting some weird errors w/ mkColor now

j9ac9k:
Fixed issue with `mkColor`, turns out that `np.isfinite` works differently (better?) on scalars than `math.isfinite` which is a bit problematic ... 

I'm concerned with some of the CI behavior; will table this for the night and try again tomorrow.

j9ac9k:
I think this guy is ready for merging.  @pijyoi hate to ping you as you're not a maintainer, but given your expertise with numpy; I'm curious if you think there are any other opportunities I have missed, or if one of the changes doesn't work the way I think it does.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -660,7 +661,7 @@ def getData(self):
                         eps = np.finfo(y.dtype).eps
                     else:
                         eps = 1
-                    y = np.sign(y) * np.log10(np.abs(y)+eps)
+                    y = np.copysign(np.log10(np.abs(y)+eps), y)
```
NilsNemitz, line 664: nice

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -372,7 +372,7 @@ def showGrid(self, x=None, y=None, alpha=None):
         if y is not None:
             self.ctrl.yGridCheck.setChecked(y)
         if alpha is not None:
-            v = np.clip(alpha, 0, 1)*self.ctrl.gridAlphaSlider.maximum()
+            v = fn.clip_scalar(alpha, 0, 1)*self.ctrl.gridAlphaSlider.maximum()
```
NilsNemitz, line 375: change limits to 0. and 1. ?

pyqtgraph/graphicsItems/ROI.py
```diff
@@ -1247,8 +1248,8 @@ def getAffineSliceParams(self, data, img, axes=(0,1), fromBoundingRect=False):
         vx = img.mapToData(self.mapToItem(img, QtCore.QPointF(1, 0))) - origin
         vy = img.mapToData(self.mapToItem(img, QtCore.QPointF(0, 1))) - origin
         
-        lvx = np.sqrt(vx.x()**2 + vx.y()**2)
-        lvy = np.sqrt(vy.x()**2 + vy.y()**2)
+        lvx = hypot(vx.x(), vx.y())
+        lvy = hypot(vy.x(), vy.y())
```
NilsNemitz, line 1252: add some comment to explain hypot?

pyqtgraph/graphicsItems/ROI.py
```diff
@@ -1332,8 +1333,8 @@ class Handle(UIGraphicsItem):
     properties of the ROI they are attached to.
     """
     types = {   ## defines number of sides, start angle for each handle type
-        't': (4, np.pi/4),
-        'f': (4, np.pi/4), 
+        't': (4, pi/4),
+        'f': (4, pi/4),
```
NilsNemitz, line 1337: I prefer math.pi to demonstrate accurate definition

pyqtgraph/graphicsItems/ROI.py
```diff
@@ -1663,17 +1664,14 @@ def __init__(self, pos1, pos2, width, **args):
         d = pos2-pos1
         l = d.length()
         ang = Point(1, 0).angle(d)
-        ra = ang * np.pi / 180.
+        ra = radians(ang if ang is not None else 0.)
```
NilsNemitz, line 1667: In the future, maybe `Point` could use a way to ask for for the angle in radians.
Right now, Point calculates in radian, converts for return, only for the result to be converted back.

pyqtgraph/graphicsItems/ROI.py
```diff
@@ -1914,7 +1912,7 @@ def shape(self):
             center = br.center()
             r1 = br.width() / 2.
             r2 = br.height() / 2.
-            theta = np.linspace(0, 2*np.pi, 24)
+            theta = np.linspace(0, 2 * pi, 24)
```
NilsNemitz, line 1915: math.pi ?

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -243,7 +242,7 @@ def generateShape(self):
         tr = QtGui.QTransform()
         tr.translate(devPos.x(), devPos.y())
         va = atan2(v.y(), v.x())
-        tr.rotate(va * 180.0 / pi)
+        tr.rotate(degrees(va))
```
NilsNemitz, line 245: There is a `rotateRadians()` method for QTransform, would that be better to use here?

pyqtgraph/graphicsItems/TextItem.py
```diff
@@ -208,7 +208,7 @@ def updateTransform(self, force=False):
         angle = -self.angle
         if self.rotateAxis is not None:
             d = pt.map(self.rotateAxis) - pt.map(Point(0, 0))
-            a = np.arctan2(d.y(), d.x()) * 180 / np.pi
+            a = degrees(atan2(d.y(), d.x()))
             angle += a
         t.rotate(angle)
```
NilsNemitz, line 213: `rotateRadians()` instead?
Needs conversion of a, so it doesn't simplify things a lot though.

pyqtgraph/opengl/GLViewWidget.py
```diff
@@ -344,7 +345,7 @@ def orbit(self, azim, elev):
             self.opts['rotation'] = q
         else: # default euler rotation method
             self.opts['azimuth'] += azim
-            self.opts['elevation'] = np.clip(self.opts['elevation'] + elev, -90, 90)
+            self.opts['elevation'] = fn.clip_scalar(self.opts['elevation'] + elev, -90, 90)
```
NilsNemitz, line 348: change to -90. and 90 to ensure float character ?

pyqtgraph/opengl/items/GLScatterPlotItem.py
```diff
@@ -3,7 +3,8 @@
 from OpenGL.arrays import vbo
 from .. GLGraphicsItem import GLGraphicsItem
 from .. import shaders
-from ... import QtGui
+from ...functions import clip_array
```
NilsNemitz, line 6: import as fn instead, just for consistent style?

pyqtgraph/opengl/items/GLScatterPlotItem.py
```diff
@@ -62,12 +63,11 @@ def initializeGL(self):
         ## Generate texture for rendering points
         w = 64
         def fn(x,y):
-            r = ((x-(w-1)/2.)**2 + (y-(w-1)/2.)**2) ** 0.5
-            return 255 * (w/2. - np.clip(r, w/2.-1.0, w/2.))
+            r = np.hypot((x-(w-1)/2.), (y-(w-1)/2.))
+            return 255 * (w/2. - clip_array(r, w/2.-1.0, w/2.))
```
NilsNemitz, line 67: I think this can be w/2 in Python3

pyqtgraph/widgets/ColorMapWidget.py
```diff
@@ -161,7 +161,7 @@ def map(self, data, mode='byte'):
             elif op == 'Set':
                 colors[mask] = colors2[mask]            
                 
-        colors = np.clip(colors, 0, 1)
+        colors = fn.clip_array(colors, 0, 1)
```
NilsNemitz, line 164: change to 0. and 1. ?

pyqtgraph/widgets/JoystickButton.py
```diff
@@ -41,7 +42,7 @@ def getState(self):
         
     def setState(self, *xy):
         xy = list(xy)
-        d = (xy[0]**2 + xy[1]**2)**0.5
+        d = hypot(xy[0], xy[1])
```
NilsNemitz, line 45: add # length ?

NilsNemitz (approved):
I have no idea how to check all these changes across so many files for weird edge cases in the behavior of math vs. numpy functions.
But almost all of it makes the code more pleasant to read in addition to being (at least technically) faster.
There are maybe some residual style consideration, please check the comments.

My strongest suggestion is maybe to stick to spelling out
`math.pi` throughout the code.
If I see just `pi`, I never know if I can rely on this to be accurate, or if there is a definition of 
`pi=3.14` hiding somewhere.

examples/GLScatterPlotItem.py
```diff
@@ -84,10 +85,10 @@ def update():
...
     s = -np.cos(d2*2+phase)
     color = np.empty((len(d2),4), dtype=np.float32)
-    color[:,3] = np.clip(s * 0.1, 0, 1)
-    color[:,0] = np.clip(s * 3.0, 0, 1)
-    color[:,1] = np.clip(s * 1.0, 0, 1)
-    color[:,2] = np.clip(s ** 3, 0, 1)
+    color[:,3] = fn.clip_array(s * 0.1, 0, 1)
+    color[:,0] = fn.clip_array(s * 3.0, 0, 1)
+    color[:,1] = fn.clip_array(s * 1.0, 0, 1)
+    color[:,2] = fn.clip_array(s ** 3, 0, 1)
```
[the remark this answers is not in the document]
j9ac9k: oh, good call! yeah GL Colors need to be floats

examples/GLVolumeItem.py
```diff
@@ -23,27 +24,21 @@
...
     return ps
-    
-    #return ((1./81.) * (1./np.pi)**0.5 * (1./a0)**(3/2) * (r/a0)**2 * (r/a0) * np.exp(-r/(3*a0)) * np.sin(th) * np.cos(th) * np.exp(2 * 1j * phi))**2 
 
 
 data = np.fromfunction(psi, (100,100,200))
-positive = np.log(np.clip(data, 0, data.max())**2)
-negative = np.log(np.clip(-data, 0, -data.min())**2)
+positive = np.log(clip_array(data, 0, data.max())**2)
+negative = np.log(clip_array(-data, 0, -data.min())**2)
```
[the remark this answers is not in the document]
j9ac9k: blerg you're right, that's normally the issue I catch/identify, in this case I think it needs to be clipped to `-eps`; ...i'll have to see what `eps ** 2` is ...

pyqtgraph/functions.py
```diff
@@ -268,17 +263,19 @@ def mkColor(*args):
                 g = int(c[2:4], 16)
                 b = int(c[4:6], 16)
                 a = int(c[6:8], 16)
+            else:
+                raise ValueError(f"Unknown how to convert {c} to color")
```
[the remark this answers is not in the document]
j9ac9k: Without this, the static code checker was having a cow, as there was a code-path through this function that did not ensure that `r` `g` `b` and `a` were set.

pyqtgraph/functions.py
```diff
@@ -288,16 +285,14 @@ def mkColor(*args):
...
-        (r, g, b, a) = args
+        r, g, b, a = args
     else:
         raise TypeError(err)
-    
-    args = [r,g,b,a]
-    args = [0 if np.isnan(a) or np.isinf(a) else a for a in args]
-    args = list(map(int, args))
+
+    args = np.nan_to_num((r, g, b, a), nan=0, posinf=255, neginf=0).astype(int).tolist()
```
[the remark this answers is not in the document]
j9ac9k: so `args = [int(a) if math.isfinite(a) else 0 for a in args]` doesn't work due to a case with PySide2 5.12 having integer overflow when passing in one of the odd numpy values (like `np.inf` or something like that).  What _could_ work is `[int(a) if np.isfinite(a) else 0 for a in args]` but what I didn't like about that was I was using a numpy function that is meant to take in array-like fields, but on a scalar.  This PR aims to break that pattern throughout the entire library.

the `np.nan_to_num` seemed to be a good compromise.

Although since you brought this up,  I tested the unpacking, which is why I converted back to a list, but turns out I actually don't need to do that, that works on a numpy array too

```python
>>> def test(a, b, c, *args):
>>>     print(a, b, c, args)
>>> array = np.arange(10)
>>> test(*array)
0 1 2 (3, 4, 5, 6, 7, 8, 9)
```

This means this can be simplified a bit by ditching the `.tolist()`

examples/GLLinePlotItem.py
```diff
@@ -29,19 +29,16 @@
...
 y = np.linspace(-10,10,n)
 x = np.linspace(-10,10,100)
 for i in range(n):
-    yi = np.array([y[i]]*100)
-    d = (x**2 + yi**2)**0.5
+    yi = y[i]
+    d = np.hypot(x, yi)
     z = 10 * np.cos(d) / (d+1)
-    pts = np.vstack([x,yi,z]).transpose()
+    pts = np.column_stack([x, np.full_like(x, yi), z])
```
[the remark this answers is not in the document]
pijyoi: The logic is this, if x has shape (N,), it is treated as (N,1). Other shapes are left as is. i.e. (1,N) would not be treated as a column.
If you look at the example at https://github.com/pyqtgraph/pyqtgraph/issues/1732#issue-863957932, it does work in an intuitive manner because both X[i,:] and X[:,i] end up with a shape (N,)

pyqtgraph/functions.py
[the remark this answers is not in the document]
pijyoi: Some time ago, I found that unpacking an ndarray is slower than calling tolist() and then unpacking that. Probably ndarray unpacking is implemented as a separate function call per element.

Regarding np.isfinite() vs math.isfinite(), the difference is that np.isfinite() operates on floats so will first convert the input to a float. This matters for big integers. So:
```python
In [28]: type(2**64)
Out[28]: int

In [29]: math.isfinite(2**64)
Out[29]: True

In [30]: np.isfinite(2**64)
TypeError                                 Traceback (most recent call last)
<ipython-input-30-ffa69f841cb0> in <module>
----> 1 np.isfinite(2**64)

TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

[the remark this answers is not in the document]
j9ac9k: Thanks for that explanation! That's really good to know.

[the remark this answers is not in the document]
NilsNemitz: Okay, one more try:
``` python
args = [
    fn.clip_scalar( int(a), 0, 255) 
    if math.isfinite(a) else 0 
    for a in args]
```
Makes it clear that we are doing input conditioning for mkColor.
And I don't think that it does a siginificantly larger number of checks and calculations than the other approaches.

pyqtgraph/functions.py
```diff
@@ -60,19 +60,15 @@ def siScale(x, minVal=1e-25, allowUnicode=True):
     
     if isinstance(x, decimal.Decimal):
         x = float(x)
-        
     try:
-        if np.isnan(x) or np.isinf(x):
+        if not math.isfinite(x):
```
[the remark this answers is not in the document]
j9ac9k: Yeah, this method does not appear to be available in python2, which is likely why it wasn't used in this library.

pyqtgraph/debug.py
```diff
@@ -202,7 +202,7 @@ def findRefPath(startObj, endObj, maxLen=8, restart=True, seen={}, path=None, ig
             #print prefix+"  FRAME"
             continue
         try:
-            if any([r is x for x in  path]):
+            if any(r is x for x in  path):
```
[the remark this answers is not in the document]
j9ac9k: This is one of my personal code-smells, i think some linters flag this too.

pyqtgraph/functions.py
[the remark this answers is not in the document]
pijyoi: I think this would fail the CI. There is a test that tests for failure with the value 1e129. It ends up entering this branch:
```python
        elif np.issubdtype(type(args[0]), np.floating):
            r = g = b = int(args[0] * 255)
            a = 255
```
So, r, g, b are now big Python ints which would get clipped to [0, 255] and get accepted instead.
By the way, it's this transformation to big Python int that's causing all this trouble.

pyqtgraph/functions.py
```diff
@@ -60,19 +60,15 @@ def siScale(x, minVal=1e-25, allowUnicode=True):
...
             return(1, '')
     except:
-        print(x, type(x))
-        raise
+        raise TypeError("Invalid type for x, should be numeric")
```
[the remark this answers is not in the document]
j9ac9k: Good point! Fixed

pyqtgraph/functions.py
```diff
@@ -432,16 +427,16 @@ def makeArrowPath(headLen=20, headWidth=None, tipAngle=20, tailLen=20, tailWidth
...
     path.moveTo(0,0)
     path.lineTo(headLen, -headWidth)
     if tailLen is None:
-        innerY = headLen - headWidth * np.tan(baseAngle*np.pi/180.)
+        innerY = headLen - headWidth * math.tan(math.radians(baseAngle))
         path.lineTo(innerY, 0)
     else:
         tailWidth *= 0.5
-        innerY = headLen - (headWidth-tailWidth) * np.tan(baseAngle*np.pi/180.)
+        innerY = headLen - (headWidth-tailWidth) * math.tan(math.radians(baseAngle))
```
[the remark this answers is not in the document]
j9ac9k: you and me both haha.  Good thing it's in version control, and `git bisect` is a thing.

pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -149,7 +151,7 @@ def dataBounds(self, ax, frac=1.0, orthoRange=None):
                 # All-NaN data is acceptable; Explicit numpy warning is not needed.
                 warnings.simplefilter("ignore")
                 b = (np.nanmin(d), np.nanmax(d))
-            if any(np.isinf(b)):
+            if math.isinf(b[0]) or  math.isinf(b[1]):
```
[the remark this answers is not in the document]
j9ac9k: there was indeed!

j9ac9k:
you would think by now I would know to run the test suite before pushing my code....

pyqtgraph/Point.py
```diff
@@ -124,11 +112,11 @@ def angle(self, a):
...
             return None
         ## Probably this should be done with arctan2 instead..
-        ang = np.arccos(clip(self.dot(a) / (n1 * n2), -1.0, 1.0)) ### in radians
+        rads = acos(fn.clip_scalar(self.dot(a) / (n1 * n2), -1.0, 1.0)) ### in radians
         c = self.cross(a)
         if c > 0:
-            ang *= -1.
-        return ang * 180. / np.pi
+            rads *= -1.
+        return degrees(rads)
```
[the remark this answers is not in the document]
j9ac9k: Can we though?  That `-1` is applied only when `Point.cross(other)` returns a positive value.  I could potentially do something with `math.copysign`

pyqtgraph/graphicsItems/TextItem.py
j9ac9k: I'm ok with this use-case; the `self.rotateAxis is not None` is a less likely case than not I figure, so I an see the case where that's the code-path that should take the (slight) hit in performance.

pyqtgraph/Point.py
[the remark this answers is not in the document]
NilsNemitz: Doh, I just missed that.
``` python
if c > 0: return degrees( -rad )
return degrees( rad )
```
is something I might write in my own code, though.

pyqtgraph/graphicsItems/TextItem.py
NilsNemitz: Ah, it seems that it is self.angle that would need conversion here, and that seems to be defined in degrees. So sticking with plain `.rotate()` is probably best here.

pyqtgraph/Point.py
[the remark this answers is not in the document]
j9ac9k: Fixed this by using much cleaner atan2 method.

pyqtgraph/functions.py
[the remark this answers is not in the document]
j9ac9k: I'm going to merge this PR as is, as it works w/o generating errors on any of the supported combinations; of course, if someone identifies a faster / more robust method to unpack safely, a PR would be appreciated 👍🏻

PR #1725, opened 2021-04-18
clean-up of PlotDataItem downsample methods

### LINKED ISSUE #1704: PlotDataItem division by 0 when 'clipToView' option enabled
In pyqtgraph/graphicsItems/PlotDataItem.py, lines 683 thru 688:

```
dx = float(x[-1]-x[0]) / (len(x)-1)
# workaround for slowdown from numpy deprecation issues in 1.17 to 1.20+
# x0 = np.clip(int((range.left()-x[0])/dx) - 1*ds, 0, len(x)-1)
# x1 = np.clip(int((range.right()-x[0])/dx) + 2*ds, 0, len(x)-1)
x0 = fn.clip_scalar(int((range.left()-x[0])/dx) - 1*ds, 0, len(x)-1)
x1 = fn.clip_scalar(int((range.right()-x[0])/dx) + 2*ds, 0, len(x)-1)
```

x0 and x1 are computed without previously checking whether dx is 0, possibly raising a divide by 0 exception.

NilsNemitz:
I had a look at #1704, where PlotDataItem errors out when clipToView is enabled with a dataset of zero width.

The error occurs in an optimization shortcut that tries to guess the clipping points for equally spaced data.
However, I don't think that shortcut is a worthwhile optimization:

It tends to mis-estimate the clipping point dramatically for non-uniform data such as used in the text.
There is a check to detect this mis-estimation and fall back to a full binary search, but 
- this only triggers if clipping was too aggressive, there is no detection (at all) for excess unclipped points.
- this check does not account for the additional points that need to be preserved when downsampling is active.

Finally, the optimization seems simply unnecessary:
On my system, the np.searchsorted call used in the precise detection only takes approximately 5 microseconds to search through 10 million datapoints. I would consider that negligible, especially since, unless spacing is perfectly uniform, the shortcut method will leave a significant number of unclipped points, which will slow down drawing (250 ns per line segment?).

I have removed the shortcut method, which also removes the code-path where the division by zero occurred.

So this should fix #1704 .


NilsNemitz:
I then tried to add a test for a zero-width data set to avoid future regression. However, the tests for the clipping functions of PlotDataItem turned out to be non-functional:
They compared the actual clipped data to the expectation for `PlotDataItem.viewRect()` (implemented by GraphicsItem).
This caches a heavy transform inversion, and seems to invalidate this cache only on an update of the transform, probably during redraw. This never happened in the test, and all tests effectively tested the same condition.

This also means that the clipping and dynamic range limiting operations were consistently executed with wrong assumptions about what's actually on screen, at least with the sequence of execution on Windows / PyQt5.

I have changed the code to determine the view range from PlotDataItem.getViewBox().viewRect(), which is already updated when  the view range change is signalled. The code only falls back to PlotDataItem.viewRect() when the viewbox cannot provide a viewRect(). However, I have not managed to trigger execution for this code path at all, so I am not sure what problem having the secondary viewRect() method in GraphicsItem addressed.

There is code that seems to fall back to a scene transform when the viewbox is not aware of a transform.
If somebody has additional information on this, please let me know :) 

This seemed like the least disruptive method to fix the delayed update.
The tests are now able to update the view range correctly and test that no more than one point is drawn outside the screen on each side (so that the connecting lines are accurate to the edge of the screen).
They also test for a crash if all data is at identical x coordinate.

NilsNemitz:
I cleaned up some of the logic in the getData() method to avoid replicating the `getViewPort().viewRect()` to `viewRect()` fall-back.
In doing so, I found that the 'mean' downsample method had a strange method of picking the x-coordinate:
The mean y-value was plotted over the left-most point in the group, which optically skews the curve.
In the interest of speed, I have replaced that by offsetting the sampled x-values by ds//2, half the downsampling step. That should plot the mean value over the approximately middle point of the interval, which should look... less bad.

NilsNemitz:
In order to reduce unneeded recalculations, GraphicsItem now includes information on what axes where updated in the `sigRangeChanged` signal (which already included a ViewBox reference and the actual viewRect). Now recalculations should trigger only if the relevant axis is changed: If clipToView is disabled, the horizontal axis changes do not require handling, although the dynamic range limiting still needs to handle vertical axis changes.

Unfortunately, this not provide a real speed-up while dynamic range limiting is enabled: Without this, recalculations could largely be avoided when clipToView and automatic downsampling are disabled.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -166,7 +166,8 @@ def __init__(self, *args, **kargs):
         self.scatter.sigClicked.connect(self.scatterClicked)
         self.scatter.sigHovered.connect(self.scatterHovered)
 
-        self._viewRangeWasChanged = False
+        self._xViewRangeWasChanged = False
+        self._yViewRangeWasChanged = False
```
NilsNemitz, line 170: Replaced generic self._viewRangeWasChanged flag with 
axis specific self._xViewRangeWasChanged and self._yViewRangeWasChanged

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -634,154 +635,159 @@ def getData(self):
...
-                    y2 = y[:n*ds].reshape((n, ds))
-                    y1[:,0] = y2.max(axis=1)
-                    y1[:,1] = y2.min(axis=1)
-                    y = y1.reshape(n*2)
+        if( self.xDisp is not None
+            and not (self._xViewRangeWasChanged and self.opts['clipToView'])
+            and not (self._xViewRangeWasChanged and self.opts['autoDownsample'])
+            and not (self._yViewRangeWasChanged and self.opts['dynamicRangeLimit'] is not None)
+        ):
+            return self.xDisp, self.yDisp
```
NilsNemitz, line 643: Early return here avoids having all the following code in an indented block

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -634,154 +635,159 @@ def getData(self):
...
+        x = self.xData
+        y = self.yData
+        view = self.getViewBox()
+        if view is None:
+            v_rng = None
+        else:
+            v_rng = self.getViewBox().viewRect() # this is always up-to-date
+        if v_rng is None:
+            v_rng = self.viewRect()
+            # assert v_rng is None # halt if this actually did anything!
```
NilsNemitz, line 654: use getViewBox().viewRect() unless it is not available.
Fall back to current behaviour of self.viewRect() (which does an inversion of the view transform to find the same numbers in all of my tests...) only if this is not available.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -634,154 +635,159 @@ def getData(self):
...
+            if view is None or view.autoRangeEnabled()[0]:
+                pass # no ViewBox to clip to, or view will autoscale to data range.
+            else:
+                # clip-to-view always presumes that x-values are in increasing order
+                if v_rng is not None and len(x) > 1:
+                    # print('search:', v_rng.left(),'-',v_rng.right() )
+                    # find first in-view value (left edge) and first out-of-view value (right edge)
+                    # since we want the curve to go to the edge of the screen, we need to preserve
+                    # one down-sampled point on the left and one of the right, so we extend the interval
+                    x0 = np.searchsorted(x, v_rng.left()) - ds
```
NilsNemitz, line 706: go straight to the accurate clipping points provided by np.searchsorted

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -634,154 +635,159 @@ def getData(self):
...
+                    x0 = fn.clip_scalar(x0, 0, len(x)) # workaround
+                    # x0 = np.clip(x0, 0, len(x))
+
+                    x1 = np.searchsorted(x, v_rng.right()) + ds
```
NilsNemitz, line 710: This was previously adding 2*ds, but I think one down-sampling block is sufficient here.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -634,154 +635,159 @@ def getData(self):
...
+        if ds > 1:
+            if self.opts['downsampleMethod'] == 'subsample':
+                x = x[::ds]
+                y = y[::ds]
+            elif self.opts['downsampleMethod'] == 'mean':
+                n = len(x) // ds
+                # x = x[:n*ds:ds]
+                stx = ds//2 # start of x-values; try to select a somewhat centered point
+                x = x[stx:stx+n*ds:ds] 
+                y = y[:n*ds].reshape(n,ds).mean(axis=1)
```
NilsNemitz, line 725: the starting offset of ds//2  (e.g. 2 for ds=5) picks a somewhat centered x value to plot the mean y value over.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -882,16 +888,33 @@ def scatterClicked(self, plt, points, ev):
...
+    #   """ view transform (and thus range) has changed, replot if needed """
+    # viewTransformChanged is only called when the cached viewRect of GraphicsItem
+    # has already been invalidated. However, responding here will make PlotDataItem
+    # update curve and scatter later than intended.
+    #   super().viewTransformChanged() # this invalidates the viewRect() cache!
+        
+    def viewRangeChanged(self, vb=None, ranges=None, changed=None):
+        """ view range has changed; re-plot if needed """
+        update_needed = False
+        if changed is None or changed[0]:
```
NilsNemitz, line 901: changes to horizontal axis require a recalculation if auto-downsampling or clipping are active.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -882,16 +888,33 @@ def scatterClicked(self, plt, points, ev):
...
+        if changed is None or changed[0]: 
+            # if ranges is not None:
+            #     print('hor:', ranges[0])
+            self._xViewRangeWasChanged = True
+            if( self.opts['clipToView']
+                or self.opts['autoDownsample']
+            ):
+                self.xDisp = self.yDisp = None
+                update_needed = True
+        if changed is None or changed[1]:
```
NilsNemitz, line 910: changes to vertical axis require a recalculation if dynamic range limiter is active

pyqtgraph/graphicsItems/ViewBox/ViewBox.py
```diff
@@ -1543,11 +1543,12 @@ def updateViewRange(self, forceX=False, forceY=False):
                     link.linkedViewChanged(self, ax)
 
             # emit range change signals
+            # print('announcing view range changes:',self.state['viewRange'] )
             if changed[0]:
                 self.sigXRangeChanged.emit(self, tuple(self.state['viewRange'][0]))
             if changed[1]:
                 self.sigYRangeChanged.emit(self, tuple(self.state['viewRange'][1]))
-            self.sigRangeChanged.emit(self, self.state['viewRange'])
+            self.sigRangeChanged.emit(self, self.state['viewRange'], changed)
```
NilsNemitz, line 1551: ViewBox has already tracked which axis was changed in this update, and now includes that in the signal.

pyqtgraph/graphicsItems/tests/test_PlotDataItem.py
```diff
@@ -114,14 +114,36 @@ def test_clipping():
...
     c.setClipToView(True)
-    w.setXRange(200, 600)
-    for x_min in range(100, 2**10 - 100, 100):
-        w.setXRange(x_min, x_min + 100)
-
+    for x_min in range(-200, 2**10 - 100, 100):
+        x_max = x_min + 100
+        w.setXRange(x_min, x_max, padding=0)
         xDisp, _ = c.getData()
-        vr = c.viewRect()
```
NilsNemitz, line 122: This never updates without redraw, and is delayed even if processEvents calls are added.

pyqtgraph/graphicsItems/PlotDataItem.py
j9ac9k: Does this codepath need to be concerned with the situation when `self.getViewBox()` might return None?

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -634,154 +635,159 @@ def getData(self):
...
+                    # print('search:', v_rng.left(),'-',v_rng.right() )
+                    # find first in-view value (left edge) and first out-of-view value (right edge)
+                    # since we want the curve to go to the edge of the screen, we need to preserve
+                    # one down-sampled point on the left and one of the right, so we extend the interval
+                    x0 = np.searchsorted(x, v_rng.left()) - ds
+                    x0 = fn.clip_scalar(x0, 0, len(x)) # workaround
+                    # x0 = np.clip(x0, 0, len(x))
+
+                    x1 = np.searchsorted(x, v_rng.right()) + ds
+                    x1 = fn.clip_scalar(x1, 0, len(x))
```
j9ac9k, line 711: probably doesn't matter, but should we clip to `clip_scaler(x1, x0, len(x))` ?

pyqtgraph/graphicsItems/ViewBox/ViewBox.py
```diff
@@ -87,7 +87,7 @@ class ViewBox(GraphicsWidget):
     sigYRangeChanged = QtCore.Signal(object, object)
     sigXRangeChanged = QtCore.Signal(object, object)
     sigRangeChangedManually = QtCore.Signal(object)
-    sigRangeChanged = QtCore.Signal(object, object)
+    sigRangeChanged = QtCore.Signal(object, object, object)
```
j9ac9k, line 90: this change may break a lot of code out there... might be worth seeing if we can do something with `sigYRangeChanged` and `sigXRangeChanged` independently.  Not saying making this change is out of the question; just pointing out that we should at least examine if we can do without it.

pyqtgraph/graphicsItems/PlotDataItem.py
NilsNemitz: No, the previous code is just `range = self.viewRect()` .  A `None` result is handled by the individual clipping/limiting routines, mostly by simply not doing anything in that case. (I changed the name to v_rng to avoid shadowing the range command.)

pyqtgraph/graphicsItems/PlotDataItem.py
NilsNemitz: Good point, I changed that now.

pyqtgraph/graphicsItems/ViewBox/ViewBox.py
NilsNemitz: I agree that this is problematic, but it won't break as much code as you think right now:
https://doc.qt.io/qtforpython/overviews/signalsandslots.html
"The signals and slots mechanism is type safe: The signature of a signal must match the signature of the receiving slot. (In fact a slot may have a shorter signature than the signal it receives because it can ignore extra arguments.)"

I came across that behavior because we are doing that already:
The previous definition of `viewRangeChanged` (which this signal get connected to) was
`def viewRangeChanged(self)`
So this has been silently discarding the (two) objects already.
The third, 'changed' object should likewise be discarded by other code, although I am unsure of the behavior if the other code uses an explicit `@Slot()` decorator.

Working with the sigXRangeChanged and sigYRangeChanged signals individually is difficult, because ideally, a range change might be simultaneously announced for both x and y axis. That should be handled by single data update.
If we react to the individual signals, we will receive two notifications, with no way to know that the second one is on the way. In that case I see no good way to avoid a second recalculation.

((Indeed, I might suggest deprecating the two specific signals. They are convenient if you fully understand what gets called when, but the large number of similar signals makes it difficult to figure out that order.))

pyqtgraph/graphicsItems/PlotDataItem.py
j9ac9k: this is sort of nit-picky, would you mind moving the `and`'s from the start of the new-line to the end of the previous line?  I find that a bit easier to read myself.

```python
        if( self.xDisp is not None and
            not (self._xViewRangeWasChanged and self.opts['clipToView']) and
            not (self._xViewRangeWasChanged and self.opts['autoDownsample']) and
            not (self._yViewRangeWasChanged and self.opts['dynamicRangeLimit'] is not None)
        ):
```

j9ac9k:
This PR is showing merge conflicts; so can't quite merge yet.  

May I suggest reconsidering renaming `v_rng` back to `view_range`?  The more code-bases I manage, the more I am developing a distaste for abbreviations; I'm willing to be overruled if someone has strong opinions on this.

The other thing, which I'm not convinced is a good idea or not, is to set `_yViewRangeWasChanged` and `_xViewRangeWasChanged` as dynamic properties for the PlotDataItem.  Main advantage here is if the default state is `False`, we wouldn't have to initialize a value, and PlotDataItem would emit a `QDynamicPropertyChangeEvent`, which we could catch and trigger relevant calculations that way.

Oh, and lastly, probably should delete some of the newly commented out lines.

Thanks so much for this @NilsNemitz I know this one was a doozy!

pyqtgraph/graphicsItems/PlotDataItem.py
NilsNemitz: I mildly prefer my logic to be visible at the start of the line, but no strong opinion on that.
--> changed

NilsNemitz:
Dear @j9ac9k ,
not sure what caused the merge conflicts, but git somehow decided that we need all the data correction/downsampling methods twice. I hope I have managed to delete the correct subset now.

The dynamic object properties are `None` when not initialized (at least they are not an "invalid QVariant" as in C!), so I do initialize these on startup.

I have converted all three "change needs to be handled" flags to properties, including "styleWasChanged". I wasn't able to come up with a good way to take advantage of the change event, but in principle I agree that this might be quite useful under some circumstances.

So it might be a good idea to try that as the standard way of handling such change flags. It does stand out in the code.

I changed `v_rng` to `view_range`. I agree that that is nicer to look at.
I changed `v_rng` from `range` originally, because my linter doesn't like `range` as a variable name.

Some commented out code is deleted now; some other is kept because it helps me remember how stuff used to work when it turns out that the latest revision breaks somebody's stuff. If it mysteriously doesn't do that, then feel free to remind me to have a clean-up pass later!

j9ac9k:
Thanks @NilsNemitz  I know this one was a doozy!  Merging.

Issue #1543, opened 2021-02-04
`Extent` parameter for ImageItem?


ntjess:
Maybe I missed it in the documentation, but is there a way to add an `extent` to an `ImageItem` (a la `pyplot.imshow(...extent=[xmin, xmax, ymin, ymax]?)` I am plotting spatial data and it would be great if my axes could line up with the true spatial values. Currently, I can add an offset to my image but I cannot show fractional axis steps per pixel.

Use case: Each pixel is 0.05mm from the next. I would like the x- and y-axes to represent mm, but I am not sure how to do this with the existing `ImageItem` support.

Thanks!

ixjlyons:
[setRect](https://pyqtgraph.readthedocs.io/en/latest/graphicsItems/imageitem.html#pyqtgraph.ImageItem.setRect) should do what you want. It could be nice to add some flexibility in the API so you don't have to construct a `QRect(F)` yourself, and maybe add it to the options available via [setImage](https://pyqtgraph.readthedocs.io/en/latest/graphicsItems/imageitem.html#pyqtgraph.ImageItem.setImage).

```python
import numpy as np
import pyqtgraph as pg
from pyqtgraph.Qt import QtCore

m_per_pixel = 0.05e-3
img_size = 600

app = pg.mkQApp()

win = pg.plot()
win.setAspectLocked(True)

data = np.random.normal(size=(img_size, img_size))
img = pg.ImageItem(data)
img.setRect(QtCore.QRectF(0, 0, img_size*m_per_pixel, img_size*m_per_pixel))

win.setLabel("bottom", text="x", units="m")
win.setLabel("left", text="y", units="m")

win.addItem(img)
app.exec_()
```

ntjess:
worked like a charm, thanks!

j9ac9k:
@ixjlyons what do you think the call to `setImage` should look like?

```python
img.setImage(data, extent=[1.5, 1.5])  # for x and y, in some fractional value?
```

ntjess:
FWIW, copied from `matplotlib.imshow`:

```
extentfloats (left, right, bottom, top), optional
The bounding box in data coordinates that the image will fill. The image is stretched individually along x and y to fill the box.

The default extent is determined by the following conditions. Pixels have unit size in data coordinates. Their centers are on integer coordinates, and their center coordinates range from 0 to columns-1 horizontally and from 0 to rows-1 vertically.

[...]

Issue #1705, opened 2021-04-10
Is it possible to plot a correlation matrix with pyqtgraph?


Bio-Otto:
I am trying to plot a correlation matrix.  For this purpose, I want to use the infrastructure of pyqtgraph. But I could not see any examples. Can you provide guidance on this issue? Thanks.

j9ac9k:
@Bio-Otto 

We don't have any convenience methods to plot those from the get-go.  

If you can generate your own correlation matrix as a numpy object, you can certainly plot one using an [ImageItem](https://pyqtgraph.readthedocs.io/en/latest/graphicsItems/imageitem.html)

If you can provide some sample data here, I can try and do a simple example.

Bio-Otto:
I put a sample script below and also the output figure. 
``` python
import pandas as pd
import seaborn as sn
import matplotlib.pyplot as plt

data = {'A': [45,37,42,35,39],
        'B': [38,31,26,28,33],
        'C': [10,15,17,21,12]
        }

df = pd.DataFrame(data,columns=['A','B','C'])

corrMatrix = df.corr()
sn.heatmap(corrMatrix, annot=True)
plt.show()
```
![resim](https://user-images.githubusercontent.com/57147948/114318584-54d89d00-9b16-11eb-85b4-5f6432860174.png)

j9ac9k:
So, first off, I'll _freely_ admit that our API/functionality here isn't as clean as seaborn, and I think this would be a great thing to add; and I'll gladly work w/ anyone that wants to submit a PR extending similar functionality here)

Here is the code I quickly wrote up:

```python
import pyqtgraph as pg
import numpy as np
# import pandas as pd

pg.setConfigOptions(imageAxisOrder="row-major")

app = pg.mkQApp()
data = {
    'A': [45,37,42,35,39],
    'B': [38,31,26,28,33],
    'C': [10,15,17,21,12]
}

columns = ["A", "B", "C"]

# df = pd.DataFrame(data, columns=columns)
# corrMatrix = df.corr().to_numpy()
corrMatrix = np.array([
    [ 1.        ,  0.5184571 , -0.70188642],
    [ 0.5184571 ,  1.        , -0.86094096],
    [-0.70188642, -0.86094096,  1.        ]
])

colorMap = pg.colormap.getFromMatplotlib("magma")
lut = colorMap.getLookupTable()

correlogram = pg.ViewBox()
[...]

Issue #1608, opened 2021-02-24
setImage(..) autoLevels not considering all data when calculating min,max


leon-vv:
### Short description

Since `setImage` skips pixels for large images, the max and min values are off. Therefore the `levels` determined from the data is incorrect, causing an apparent empty image.

Please see `setImage` in `ImageItem.py` and the screenshot below.

![image](https://user-images.githubusercontent.com/4085287/109031439-05a0ef00-76c5-11eb-8326-3f5844b7ed31.png)

### Expected behavior

The levels should have been determined correctly.

### Real behavior

Since only a subset of the image is considered, the min and max values are off resulting in a wrong look up table. Note that in my field (physics) single pixel 'hits' are very common, and this behavior results in empty images _sometimes_, depending on whether the pixel happens to be part of the image considered for the minimum and maximum values. I spent multiple hours debugging this..

In any case, thanks for all the work you all have put into this project.

PR #1726, opened 2021-04-18
Add axis convenience methods and matrix display example

NilsNemitz:
In #1705, we show that PyQtGraph can plot a decent looking matrix representation.

However, it is quite obvious that the axis configuration includes a lot of boilerplate code.
This PR introduces some convenience functions:
`PlotItem.showAxes( (bool, bool, bool, bool), showValues=(bool, bool, bool, bool), size=(width,height) )`
takes simple tuples of Booleans to configure which axes are shown, and for which of them tick labels are displayed.
The order follow the Qt convention of left, top, right, bottom.
A size can also be given to assign extra space only for the labeled axes, and without the inconvenience of handling that horizontal axes now need `setHeight()` while vertical axes need `setWidth()`

`ViewBox.setDefaultPadding( padding )`
takes a user-defined value for the default padding fraction (which is so far fixed at 0.02).
In this way setting `ViewBox.setDefaultPadding(0.)` is sufficient to autorange the viewbox to the exact image bounds.
Non-zero values effectively rescale the padding algorithm to match the new factor.

PlotItem wraps this ViewBox method for convenience.

The default Z position of AxisItem has been changed to be 0.1, in order to always draw the axes on top of the image... With the current default of -1000, the ImageItem obscures them completely. I am not aware of the reason for the previous default setting, though.

These convenience methods greatly simplify the code to plot the correlation matrix, and the following has been added as an example:
``` python
# -*- coding: utf-8 -*-
"""
This example demonstrates ViewBox and AxisItem configuration to plot a correlation matrix.
"""
import initExample

import numpy as np
import pyqtgraph as pg
import numpy as np

corrMatrix = np.array([
    [ 1.        ,  0.5184571 , -0.70188642],
    [ 0.5184571 ,  1.        , -0.86094096],
    [-0.70188642, -0.86094096,  1.        ]
])
columns = ["A", "B", "C"]

app = pg.mkQApp()
win = pg.PlotWidget()

# create correlation matrix image with correct orientation:
correlogram = pg.ImageItem(image=corrMatrix, axisOrder='row-major')

plotItem = win.getPlotItem()        # get PlotItem of the main PlotWidget
plotItem.getViewBox().invertY(True) # orient y axis to run top-to-bottom
plotItem.setDefaultPadding(0.0)     # plot without padding data range
plotItem.addItem(correlogram)       # display correlogram
# show full frame, label tick marks at top and left sides, with some extra space for labels
plotItem.showAxes( True, showValues=(True, True, False, False), size=20 )

# define major tick marks and labels:
ticks = [ (idx+0.5, label) for idx, label in enumerate( columns ) ]
for side in ('left','top','right','bottom'):
    plotItem.getAxis(side).setTicks( (ticks, []) ) # add list of major ticks and no minor ticks
plotItem.getAxis('bottom').setHeight(10) # include some additional space at bottom of figure

colorMap = pg.colormap.get("cet-d1")     # choose a perceptually uniform, diverging color map
bar = pg.ColorBarItem( values=(-1,1), cmap=colorMap) # generate an adjustabled color bar, initially spanning -1 to 1
bar.setImageItem(correlogram, insert_in=plotItem)    # link color bar and color map to correlogram, and show it in plotItem

win.show()
app.exec_()
```


pyqtgraph/graphicsItems/ImageItem.py
```diff
@@ -202,11 +202,45 @@ def setOpts(self, update=True, **kargs):
...
+        the origin at (0.5, 0.5), the center of the first pixel.
+        
+        Parameters
+        ----------
+        point: QtCore.Pointf or tuple (x,y) of float
+            Location of axis origin in terms of image pixels
+        scale: float or tuple (scale_x, scale_y) of float
+            Pixel size in plot coordinates, non-square sizes can be given as a tuple.
+        """
+        if isinstance(point, QtCore.QPointF):
```
j9ac9k, line 228: We can probably use `QPoint` here as well;

```python
if isinstance(point, (QtCore.QPointF, QtCore.QPoint)):
```

NilsNemitz:
Dear @j9ac9k , I have resolved the merge conflict that resulted from an update to the clipping code fighting a change to the automatic padding logic.

I have changed the matrix display example to be a full QMainWindow, since the tests on my system seemed to have trouble getting the program to close again.

To further simplify the plotting example, I have added a `setOrigin()` method to `ImageItem`. This moves the displayed image around in pretty much the same way that `setRect()` already does, but makes this a little more intuitive to use:
To make the plot axes draw through the center of the corner pixel / matrix element, the command is simply
`img.setOrigin( (0.5, 0.5) )` now. @pijyoi , do you think that makes sense as a shortcut ?

j9ac9k:
The static code checker raised a warning with this PR:

https://github.com/pyqtgraph/pyqtgraph/security/code-scanning/1302?query=ref%3Arefs%2Fpull%2F1726%2Fmerge

At a quick glance I'm not sure of an easy way to address this.

pijyoi:
>To make the plot axes draw through the center of the corner pixel / matrix element, the command is simply
img.setOrigin( (0.5, 0.5) ) now. @pijyoi , do you think that makes sense as a shortcut ?

My common use case is nicely described in the following post:
https://stackoverflow.com/questions/64896878/how-to-plot-2d-data-image-with-x-and-y-coordinates

It only requires calling setRect(), no translation nor scaling.

e.g. for your use case, it would be:
```python
import numpy as np
import pyqtgraph as pg

data = np.random.randint(256, size=(4,5), dtype=np.uint8)

app = pg.mkQApp()

win = pg.PlotWidget()
win.show()
plt = win.getPlotItem()
imgitem = pg.ImageItem(data, axisOrder='row-major', autoLevels=False)
# in the general case, w, h, dx, dy are arbitrary floats
h, w = data.shape
dx = dy = 1
imgitem.setRect(QtCore.QRectF(-dx/2, -dy/2, w, h))
plt.addItem(imgitem)

app.exec_()
```

NilsNemitz:
Thank you for the feedback @pijyoi !

For the use case in the stackoverflow post, `setRect()` makes perfect sense.

For the matrix example, not so much:
In your example, you need to grab `h, w = data.shape` just to define the `QRectF`.
Under the hood, `setRect()` then basically just throws them away again:
```python
 def setRect(self, rect):
        tr = QtGui.QTransform()
        tr.translate(rect.left(), rect.top())
        tr.scale(rect.width() / self.width(), rect.height() / self.height())
        self.setTransform(tr)
```
So there is unneeded code on the user side just to communicate that the scaling remains at the default of 1.

Additionally, `setRect( QtCore.QRectF() )` cannot be used before data is set: I am pretty sure that `self.width` and `self.height` are None before that.

The shift in origin by `setOrigin(0.5, 0.5)` can be set on initialization and will continue to apply to all matrices of different size that are plotted later.

So I think having an alternative (and somewhat orthogonal) method to position ImageItems on the grid would be a good addition to the API.

If it is better to define this in terms of an image translation or the axis origin is another question. I kind of like the axis origin here, because that seems to better match the situation where thinking in terms of a shift, rather than a target box is helpful: In the matrix example #1705, the thought is "how do I point the 0, 1, 2,... ticks of the axes at the *center* of the boxes indicating the matrix elements?"

pyqtgraph/graphicsItems/ImageItem.py
NilsNemitz: Done.

j9ac9k:
> I don't think I touched anything related to the code checker complaint.

yeah I didn't think you had, code scanner likely picked it up due to the diff... I'm good w/ patching that up as part of a separate PR.

pijyoi:
>So I think having an alternative (and somewhat orthogonal) method to position ImageItems on the grid would be a good addition to the API.

Maybe should revisit the discussion in #1543

NilsNemitz:
Dear @pijyoi , thank you for the reminder about #1543!

I have updated the PR to reflect my proposal in this matter:
- We promote the already existing `setPos()` and `setScale()` methods from QGraphicsItem so that they appear in the documentation. (`setScale()` needs a minor extension to handle separate x and y scales.)
- We drop my proposed `setOrigin()`, which basically replicates these functions.
- We extend `setRect()` to handle x,y,w,h parameters directly or as a tuple (in addition to the current QRectF)
- We add a keyword `rect` to `setOpts()`, which calls `setRect()`. I think that fits nicely in the established naming scheme and adds this as an option to both the initalization call and to `setImage()`.
- We add an explanatory sentence to the ImageItem class docstring to explain "map this image here" and persistent offset + scale.

I think that would address #1543: It basically replicates the matplotlib `extent` keyword, named `rect` here for continuity with `setRect()`. It also updates the ImageItem documentation (somewhat) so that the already available methods are easier to find.

A minor technical detail:
The `QGraphicsItem.pos()` property seems to be stored and handled separately from the transform. ImageItem seems to apply *both* the offset set by `setPos()` and the offset set by `setRect()`, which results in unexpected behavior. I think the two methods cooperate better if `setRect()` is changed to also handle its translations through `setPos()` calls.

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -1169,6 +1172,58 @@ def showAxis(self, axis, show=True):
...
+            showValues = (False, False, False, False)
+        elif showValues is None:
+            showValues = (None, None, None, None)
+        if size is None:
+            size = (None, None)
+        elif not isinstance(size, collections.abc.Sized):
+            size = (size, size)
+        for idx, key in enumerate(('left','top','right','bottom')):
+            if selection[idx] is True : self.showAxis(key)
+            if selection[idx] is False: self.hideAxis(key)
```
j9ac9k, line 1213: can we redo these lines by having

```python
self.showAxis(key) if selection[idx] else self.hideAxis(key)
```

NilsNemitz: I think we can't because this is really a three-way choice:
`True` means turn on
`False` means turn off
`None` (or anything else, really) means leave as is.

j9ac9k: ahh, you're right, that cannot be replicated with the ternary operator.

Probably should do `if/elif/else` then

```python
for idx, key in enumerate(('left', 'top', 'right', 'bottom')):
    if selection[idx] is True:
        self.showAxis(key)
    elif selection[idx] is False:
        self.hideAxis(key)
    else:
        continue # don't do anything otherwise
    ...
```

j9ac9k: actually, we can redo this entire loop now that I look at it...

```python
for idx, key in enumerate(('left','top','right','bottom')):
    ax = self.getAxis(key)
    if selection[idx] is True: 
        self.showAxis(key)
        ax.setStyle(showValues=True )
            if key in ('left','right') and size[0] is not None:
                ax.setWidth( size[0] )
            elif key in ('top', 'bottom') and size[1] is not None:
                ax.setHeight( size[1] )
    elif selection[idx] is False:
        self.hideAxis(key)
        ax.setStyle(showValues=False )
        if key in ('left','right') and size[0] is not None:
            ax.setWidth( None ) # revert to auto-width
        elif key in ('top', 'bottom') and size[1] is not None:
            ax.setHeight( None ) # revert to auto-height
    else:
        continue  # if selection[idx] is not True or False do nothing...
```

This gets us out of doing the same if-check `showValues[idx] is ...` multiple times.

NilsNemitz: I may be missing something here, but some of the repeated code handles the separation of displaying just the axis and the text labels associated with the axis.
We could maybe shorten the code a little bit by only handling the labels when the axis is actually visible. But since the axis is never really deleted, it still has a well-defined state even if it is not shown. So I think there is less potential for unexpected behavior if we always configure the labels as requests, visible or not.

NilsNemitz:
Dear @j9ac9k , I have added an explicit `if`, `elif`, `else` construction to make the "on/off/no change" distinction obvious.
I am not seeing a great potential for streamlining the axes visible / labels visible checks, and I think they are most obvious as written now... but I'll be happy to look at it again.

Dear @pijyoi, this now avoids defining new methods, and instead promotes the existing `setPos()` and `setScale()` methods (by documentation and by allowing separate x/y scales) to complement `setRect()`. Your opinion on this would be valued!

pijyoi:
If we take a look at the implementation:
https://code.woboq.org/qt5/qtbase/src/widgets/graphicsview/qgraphicsitem_p.h.html#QGraphicsItemPrivate::TransformData
We can see that there are multiple (stacking) transformations that can be applied:
1) QTransform instance from setTransform()
2) List of transforms from setTransformations(). (We are not concerned with this)
3) rotation from setRotation()
4) scale_xy from setScale()

Note that setPos() is not in the above. i.e. it exists separately.

Given the above, it's probably not a good idea to have setScale() use (1) if it has 2 arguments and (4) if it has only 1 argument. Personally, I am not a fan of shadowing parent methods with different function call signatures.

I think setRect() should continue to use (1) to implement translate() and scale(). This also has the benefit of not changing  behaviour for existing uses of setRect().
setPos() can be used for simpler cases to shift the axis by 0.5 unit like for your matrix plotting example. Likewise for setScale() with just one argument.

I don't think it's likely that the user would mix uses of setRect() with setPos() and setScale(). I expect that the same ImageItem would be used to display arrays containing the same kind of data. In fact, with the existing implementation before this PR, the user can already mix uses of setRect() with setPos() and setScale(), but we don't hear any issues about that.

j9ac9k:
Hey everyone,

I'm just catching up here, and trying to understand the nature of the ~disagreement~ opposing viewpoints.  Please correct me if I'm mistaken here.

If I'm understanding correctly, @NilsNemitz wants to introduce some other methods to shift the position/scale of ImageItems outside of [.setRect()](https://pyqtgraph.readthedocs.io/en/latest/graphicsItems/imageitem.html#pyqtgraph.ImageItem.setRect) in part due to not needing all the information needed to construct a `QRectF` object (such as width and height).  For example of moving or scaling an ImageItem would not necessarily need a QRectF to occur.  This potentially solves the use-case of trying to manipulate the ImageItem before a window is displayed (or in other words, before an ImageItem has a height or width).

On the flip side, @pijyoi you're indicating that having those other methods doesn't really buy us anything, and being able to manipulate an ImageItem before it's displayed isn't exactly a feature that people have been requesting; so we would be taking on maintaining a more complicated API while adding minimal additional functionality to users.

Please correct me if I'm wrong here.

pijyoi:
What I am uncomfortable with is the modification of setRect() from using QTransform::translate() to setPos(). They are different things as far as Qt internals are concerned.

Regarding the inability of setRect() to set the coords prior to setting the image data, in a few of my applications, the coords are not known prior and are a property tied to each of the image data. So at least for these use-cases, it is natural for me to always have to call setImage() followed by setRect() upon each image data update.

For applications with fixed pixel sizes and offsets, I can see where having to repeatedly call setRect() would be considered an annoyance.

What I was suggesting:
1) Leave setRect() as is. Don't modify behavior for existing users.
2) Promote setPos() and setScale(single_argument) for simpler use-cases

j9ac9k:
> What I was suggesting:
> 1. Leave setRect() as is. Don't modify behavior for existing users.
> 2. Promote setPos() and setScale(single_argument) for simpler use-cases

This sounds perfectly reasonable to me.  I'm a bit reluctant to touch `setRect` given it does the job it aims to do adequately well; and even if we were to add a function signature that accepted a wider variety of inputs, that wouldn't really match with `setRect` being the name of it.

NilsNemitz:
Dear @j9ac9k , thanks for chiming in on this.

I agree with (1.), let's keep setRect() as it is.

I have two issues that make me think (2.) isn't the ideal solution:
- independent *x* and *y* scaling seems like good functionality to have. Example: Scientific camera systems can typically bin images independently for *x* and *y* directions of the sensor. Setting the *x* scale to the *x* binning factor is then an easy way to restore the display to proper coordinates. If we use `setScale()`, we'll never be able to add this functionality to the easy method, since Qt has chosen to code this as `x.scale(scale, scale);` internally, even though the raw scale method seems to scale *x* and *y* separately.
- the `setRect()` and `setPos()` (+setScale) methods don't play well with each other: In the list of Qt transformations that @pijyoi listed earlier, `setRect()` acts on the very first ("setTransform") step. The adjustments for `setPos()` and `setScale()` are applied separately. That means that if `setRect()` has ever been used on an ImageItem, `setPos()` and `setScale()` will work with a pre-scaled, pre-translated object and no longer give the expected results. In return, any setting that has been made by `setPos()` cannot be undone by `setRect()`. That is probably not a problem now, since using these methods requires familiarity with the PyQtGraph source and the underlying Qt objects. But if we promote a mix of `setRect()` and `setPos()` / `setScale()` in the documentation, I expect confusion.

So I have come to think that a new method (with a new name that doesn't conflict with existing methods) might be the best solution. QGraphicsItem already has methods for [everything and the kitchen sink](https://doc.qt.io/archives/qt-4.8/qgraphicsitem-members.html), so it's not like we are ruining a beautiful minimalist API here. 

This new method should work at the same transformation step as `setRect()`: 
- Then only the latest setting will take effect and the potential for confusion is greatly reduced. 
- This also opens up independent *x* and *y* scaling.

There is a certain challenge in finding a good name for such a method, and suggestions would be welcome.

My ideas right now would be either
`setImageTransform( x, y, scale=1.0 )`  to indicate that this works on the first transform stage.
or
`setOrigin( -x, -y, scale=1.0 )`, which indicates that this is related to placement without shadowing `setPos`. It also fits well with the original trigger question of "how do I draw axes though the center of the 0,0 pixel?" 
The scaling functionality would seem tacked on, though. And separating out scaling to a separate method is inconvenient on the transform stage, where we would want to clear the current transform and assign a new one in one step, like `setRect()` already does.

j9ac9k:
> independent x and y scaling seems like good functionality to have.

I agree, this is functionality we should probably expose.  The `QGraphicsScale` class does have `setXScale`, `setYScale`  and `setZScale` methods in case we were looking for validation there.

> the `setRect()` and `setPos()` (+setScale) methods don't play well with each other

O; that's a problem.

> I am not a fan of shadowing parent methods with different function call signatures.

@pijyoi yeah I'm not a fan of this too; in a pinch I'll consider it, but I don't think this feature requests meets that threshold.

My first thought here is that the first line of `setRect()` is `self.resetTransform()`; can we create our own method that would effectively "reset" any transformations that would be called via `setPos` `setXScale`, `setYScale` (or whatever we name them to be).

For example

```python

class ImageItem:

    ...
     self.transforms = []  # contains a list of transforms we've applied via _not_ setRect
    
    def setPos(self, (x, y)):
        setPosTransform = ....
        self.transforms.append(setPosTransform)

    def resetTransform(self):
        for transform in self.transforms:
            # undo each transform one at a time
        self.transforms.clear()
        return super().resetTransform()
```

NilsNemitz:
Thank you for the clear statements of opinion.

My updated interpretation is this
The QTransforms already used by setRect 
- have the lowest potential for confusing interactions
- have the best functionality, including separate x/y scaling.

So I think it is preferable to work with these, rather than the limited and conflicting `setPos` and `setScale`. But instead of adding another frontend for the QTransforms to our code, we can just promote using the QTransforms directly. The obstacles to the user doing that are
- not knowing where to look
- uncertainty if this will cause a conflict with PyQtGraph's internal functions.

We can address these points in the documentation.

This PR now
- reverts setRect to work as before (QTransforms for everything), except for the added (x,y,w,h) parameter input 
- no longer promotes `setPos` and `setScale` methods
- extends the documentation of ImageItem to include an example for directly setting up transforms
- generally revises the documentation for ImageItem
    - to better match the new NumPy inspired style
    - to hint at the transform-based nature of setRect
    - to establish the new `rect` keyword argument
    - to include a summary of the performance hints @pijyoi compiled for PR #1693
    - to include some useful undocumented methods, like `setBorder`
    - to hide the `dataTransform()` and `inverseDataTransform()` methods which do not appear very helpful for the user (?)
   
When looking at the `dataTransform()` methods, it seems that they rebuild the transform every time they themselves or the corresponding (and undocumented) `mapToData()` or `mapFromData()` methods are called. I changed the code to only do that recalculation when it is needed, that is, when axisOrder changes.

@pijyoi , @j9ac9k , let me know what you think :)

pyqtgraph/graphicsItems/ImageItem.py
```diff
@@ -197,17 +246,42 @@ def setOpts(self, update=True, **kargs):
...
+        
+        Sets translation and scaling of this ImageItem to display the current image within the rectangle given
+        as ``QtCore.QRect`` or ``QtCore.QRectF`` `rect`, or described by parameters `x, y, w, h`, defining starting 
+        position, width and height.
+
+        This method cannot be used before an image is assigned.
+        See the :ref:`examples <ImageItem_examples>` for how to manually set transformations.
+        """
+        rect = None
+        if len(args) == 4:
```
j9ac9k, line 266: probably should do `if len(args) in (2, 4)` as there are some valid QRectF constructors involving combinations of QPointF and QSizeF.... seems like a lot of work to check user-input in this case tho...

j9ac9k:
```python
rect = None
if isinstance(args, (QRect, QRectF):
    rect = args
elif hasattr(args, '__len__'):
    arg = args[0]
    rect = QRectF(*arg)
else:
    rect = QRectF(*args)
```

This should keep working w/ all compatible QRectF initializers.

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -1171,6 +1173,64 @@ def showAxis(self, axis, show=True):
...
+            showValues = (True, False, False, True)
+        elif showValues is False: # shortcut: disable all labels
+            showValues = (False, False, False, False)
+        elif showValues is None: # leave labelling untouched
+            showValues = (None, None, None, None)
+        if size is None:
+            size = (None, None)
+        elif not isinstance(size, collections.abc.Sized):
+            size = (size, size)
+        for idx, key in enumerate(('left','top','right','bottom')):
```
j9ac9k, line 1212: this loop bugs me a bit, because there is a fair amount of almost-duplicated code; ...earlier when I commented on this, I missed the case where you could have `None`, I think I re-wrote this to handle that use-case accordingly.

```python
axes = ('left', 'top', 'right', 'bottom')
for sub_selection, showValue, axis in zip(selection, showValues, axes):
    if sub_selection is True:
        self.showAxis(axis)
    elif sub_selection is False:
        self.hideAxis(axis)
    else:
        pass
    if showValue is None:
        continue
    ax = self.getAxis(axis)
    if showValue:
        ax.setStyle(showValues=showValue)
        if axis in ('left', 'right') and size[0] is not None:
            ax.setWidth(size[0] if showValue else None)
        elif axis in ('top', 'bottom') and size[1] is not None:
            ax.setHeight(size[1] if showValue else None)
```

j9ac9k:
I think advertising `QTransform` a bit more in the documentation is a great compromise.  `QTransform` always felt a little low-level, but it's really useful, and I don't think pyqtgraph should try and obscure away its usage.  I've been thinking about doing that a bit more in general, but wasn't sure where a good place to do it would be; but `ImageItem` feels like a fantastic place.

I made a few comments on the code changes; I need to read over the doc changes a bit more....which I may not get to for a bit.

pijyoi:
I think promoting setTransform(QTransform) is a good idea. Thanks for documenting the performance hints!

pyqtgraph/graphicsItems/ImageItem.py
NilsNemitz: The idea is not primarily to construct a QRectF, but rather to set the display rectangle directly from float values. But the code seemed easiest to read for building up the QRectF.

How's this for a minimalistic approach?
```python
if isinstance(args, (QtCore.QRectF, QtCore.QRect)):
            rect = args # use QRectF or QRect directly
        else:
            if hasattr(args[0],'__len__'):
                args = args[0] # promote tuple or list of values
            rect = QtCore.QRectF( *args ) # QRectF(x,y,w,h), but also accepts other initializers
```
That either produces a suitable rectangle, or results in a quite readable stack trace. This one happens for giving two float values:
```
File "C:\e\prog\pg_dev\pyqtgraph\pyqtgraph\graphicsItems\ImageItem.py", line 270, in setRect
    rect = QtCore.QRectF( *args ) # QRectF(x,y,w,h), but also accepts other initializers
  QRectF(): too many arguments
  QRectF(Union[QPointF, QPoint], QSizeF): argument 1 has unexpected type 'int'
  QRectF(Union[QPointF, QPoint], Union[QPointF, QPoint]): argument 1 has unexpected type 'int'
  QRectF(float, float, float, float): not enough arguments
  QRectF(QRect): argument 1 has unexpected type 'int'
  QRectF(QRectF): argument 1 has unexpected type 'int'
```

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
NilsNemitz: Here's what I get if I try to sort out the duplicate checks, but try to keep the bypass for None values explicit. I also realized that setting size to None should revert back to auto-size, because otherwise there is no way to do that. 

``` python
all_axes = ('left','top','right','bottom')
        for show_axis, show_value, axis_key in zip(selection, showValues, all_axes):
            if show_axis is None:
                pass # leave axis display as it is.
            else:
                if show_axis: self.showAxis(axis_key)
                else        : self.hideAxis(axis_key)
                
            if show_value is None:
                pass # leave value display as it is.
            else:
                ax = self.getAxis(axis_key)
                ax.setStyle(showValues=show_value)
                if size is not False: # size adjustment is requested
                    if axis_key in ('left','right'):
                        if show_value: ax.setWidth(size[0])
                        else         : ax.setWidth( None )
                    elif axis_key in ('top', 'bottom'):
                        if show_value: ax.setHeight(size[1])
                        else         : ax.setHeight( None )
```

pyqtgraph/graphicsItems/ImageItem.py
```diff
@@ -262,17 +262,12 @@ def setRect(self, *args):
...
-            rect = QtCore.QRectF( *args ) # QRectF(x,y,w,h)
-        elif len(args) == 1:
-            arg = args[0]
-            if isinstance(arg, (QtCore.QRectF, QtCore.QRect)):
-                rect = arg # use QRectF directly
-            elif hasattr(arg,'__len__') and len(arg) == 4:
-                rect = QtCore.QRectF( *arg ) # unpack (x,y,w,h) for QRectF
-        if rect is None:
-            raise ValueError("setRect argument must be a QRectF, QRect, or x,y,w,h")
+        if isinstance(args, (QtCore.QRectF, QtCore.QRect)):
```
pijyoi, line 265: ```args``` is a tuple. So it should be:
```python
if len(args) == 1 and isinstance(args[0], (QtCore.QRectF, QtCore.QRect)):
    rect = args[0]
else:
    # other conditions
```

pyqtgraph/graphicsItems/ImageItem.py
```diff
@@ -262,8 +262,10 @@ def setRect(self, *args):
         This method cannot be used before an image is assigned.
         See the :ref:`examples <ImageItem_examples>` for how to manually set transformations.
         """
-        if isinstance(args, (QtCore.QRectF, QtCore.QRect)):
-            rect = args # use QRectF or QRect directly
+        if len(args) == 0:
```
pijyoi, line 265: This should be "elif"

NilsNemitz: Thank you!
The intention was indeed to reset the transform and be done. I have added a return statement for the len == 0 case. 

Sorry about the messed up tuple parsing. My mental stack overflowed on the logic to parse a tuple-with-a-tuple-in-it into a tuple-without-a-tuple-in-it. The messed up code passed tests locally, though, so I added some basic testing for the values / tuple / QRect / QRectF input cases.

doc/source/images/gen_example_imageitem_transform.py
```diff
@@ -0,0 +1,45 @@
...
+        exporter = exp.ImageExporter(self.scene())
+        exporter.parameters()['width'] = 420
+        exporter.export('example_imageitem_transform.png')
+
+mkQApp("ImageItem transform example")
+main_window = MainWindow()
+
+## Start Qt event loop
+if __name__ == '__main__':
+    mkQApp().exec_()
```
pijyoi, line 45: with #1774 merged, this is now "pg.exec()"

NilsNemitz:
Dear @pijyoi ,
In case you have any time, this now segfaults in the numba makeARGB test, specifically on Ubuntu/pyside/Python 3.7:
`pyqtgraph/tests/test_makeARGB.py::test_numba_makeARGB_against_generated_references 
/home/runner/work/_temp/73ebc8a5-f1e6-4ceb-991d-d80e6875c15a.sh: line 4:  3011 Segmentation fault      (core dumped) pytest pyqtgraph examples -v --junitxml pytest.xml`

I probably broke something, but I currently don't know where to look. If you have any hints, please let me know!

The single merge conflict with master was a change to 
`def setAutoDownsample(self, active=True):`(renaming the parameter `ads` to `active` for the sake of documentation.)
There I had to manually accept the new code for `def _ensure_proper_substrate(data, substrate):` , but the code should be identical to master, and does not show any changes marked in "Files changed".

Function-wise, this PR changes the methods `dataTransform()` and `inverseDataTransform()` as well as `mapToData()` and `mapFromData()` to work with cached values, but I don't see how the Numba makeARGB functions would interact with that; or why it would work on other platforms if I broke that.

The rest of the changes to `ImageItem` here are to the documentation. 

Sorry for the trouble!

Otherwise I hope that I have now implemented all the changes you suggested. :)

pijyoi:
When only a single combination of (platform, binding, python) fails, it's usually some CI flakiness.
For instance, all passed on your own fork https://github.com/NilsNemitz/pyqtgraph/actions/runs/869055759
All passed on my own fork too.

j9ac9k:
There was something up with github actions, as the CodeQL status checker timed out; .... so I'm going to chalk this up to CI glitch.  I'm rerunning CI again and we'll see what happens.

NilsNemitz:
Thank you for the confidence, it does seem like all the regular tests ran without crashing now.

NilsNemitz:
Looking over the issue log, it seems that this would close 
#1543 through the addition of the `setRect` keyword argument
#1608 through documenting that setData does not sample all pixels.

j9ac9k:
Huh, a merge conflict, wonder what caused that....

pyqtgraph/graphicsItems/ImageItem.py
```diff
@@ -342,45 +419,71 @@ def setImage(self, image=None, autoLevels=None, **kargs):
...
-            tr.rotate(-90)
-        return tr
+        # transforms are updated in setOpts call.
+        return self._dataTransform
+        # tr = QtGui.QTransform()
+        # if self.axisOrder == 'row-major':
+        #     # transpose
+        #     tr.scale(1, -1)
+        #     tr.rotate(-90)
+        # return tr
```
j9ac9k, line 451: Should probably get rid of this commented out code.

j9ac9k:
Hi @NilsNemitz 

I think this is ready for merging once the merge conflict is resolved; and some of the commented out code is removed.  

Sorry it's taken me so long to get to this!

NilsNemitz:
Dear @j9ac9k , I hope I got all the commented out fragments.
I had a quick look over the generated (revised) docs, and that also seems to work :) 

Let me know if you have anything else to fix!

tests/graphicsItems/test_ImageItem.py
```diff
@@ -185,6 +185,45 @@ def test_ImageItem_axisorder():
...
+            'tr22': ( tr1.m22(), tr2.m22() ),
+            'tr23': ( tr1.m23(), tr2.m23() ),
+            'tr31': ( tr1.m31(), tr2.m31() ),
+            'tr32': ( tr1.m32(), tr2.m32() ),
+            'tr33': ( tr1.m33(), tr2.m33() )
+        }
+        for key, values in dic.items():
+            val1, val2 = values
+            if val1 != val2:
+                print(key,'mismatch:', val1,'!=',val2 )
```
j9ac9k, line 205: This print-statement should probably be moved to the assert statement that calls this function.  I'm trying to think of a good way to embed this; only other alternative I can propose is to have this function handle the assert call as well; and have this print-statement baked in there.

j9ac9k: If you _really_ want to be slick about it, you can tell pytest how to display failed asserts for these types (totally not a requirement for this PR, just sharing)

https://docs.pytest.org/en/6.2.x/assert.html#defining-your-own-explanation-for-failed-assertions

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -1171,7 +1174,63 @@ def showAxis(self, axis, show=True):
...
+        elif selection is False: # shortcut: disable all axes
+            selection = (False, False, False, False)
+        if showValues is True: # shortcut: defaults arrangement with labels at left and bottom
+            showValues = (True, False, False, True)
+        elif showValues is False: # shortcut: disable all labels
+            showValues = (False, False, False, False)
+        elif showValues is None: # leave labelling untouched
+            showValues = (None, None, None, None)
+        if size is not False and not isinstance(size, collections.abc.Sized):
+            size = (size, size) #
```
j9ac9k, line 1211: a `#` managed to escape the mass deletion of comments!

tests/graphicsItems/test_ImageItem.py
NilsNemitz: At the moment, we don't seem to test transform matrices a lot, so it is probably easier to keep everything together in the test function.
I changed the helper function to be `assert_equal_transforms()`.
It now checks all the elements and passes the assertion only when they are all the same. If they are not, it raises "AssertionError: Matrix element mismatch" with additional lines listing the mismatched matrix elements. These additional lines do not appear in the summary, but are captured in the detailed output:
```
E       AssertionError: Matrix element mismatch
E         tr11: 6.0 != 2.0
E         tr22: 16.0 != 4.0
E         tr31: -6.0 != -2.0
E         tr32: -16.0 != -4.0
E
E       assert False
```
That's the most helpful idea I currently have for providing some diagnostic output.

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
NilsNemitz: It appears that this dangerous specimen has devoured the entire team of commentary it had been assigned to protect. Horrible thing, that. I have replaced the `#` with a less dangerous one, and recruited a new volunteer comment to explain that this is intended to `# make sure that size is either False or a full set of (width, height)`

NilsNemitz:
Dear @j9ac9k , I tried to clean up the matrix test assertion somewhat, let me know if you have ideas how to make this more elegant. I looked for the code checker output, and did not see any complaints there.

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
j9ac9k: 😆

j9ac9k:
Hey look, the segfault I was diagnosing in my PR.  Anyway not going to hold off on this PR because of it, as it is clearly unrelated.

This LGTM, thanks for updating.  Merging.

PR #1730, opened 2021-04-21
fix ExampleApp crash on PowerShell

pijyoi:
When the ExampleApp is launched from Windows PowerShell, selecting a binding other than the default and then launching an example may cause a crash of the ExampleApp itself. This may have to do with spawnle being documented as not thread-safe on Windows. https://docs.python.org/3/library/os.html#os.spawnle

This PR fixes it by replacing use of os.spawn* with subprocess.Popen, as suggested by the Python docs.
It also simplifies the code.


j9ac9k:
Thanks @pijyoi this looks good to me. Merging!

PR #1735, opened 2021-04-22
unify _dpiRatio() and devicePixelRatio()

### LINKED ISSUE #1732: GLViewWidget::readQImage() partial image on hidpi
### Short description

On a system with device pixel ratio of 2.0, GLViewWidget::readQImage() does not save the full screen image. Instead, about a quarter is saved.

### Code to reproduce

```python
import pyqtgraph.opengl as gl
import pyqtgraph as pg
import numpy as np

app = pg.mkQApp()
w = gl.GLViewWidget()
w.opts['distance'] = 30
w.show()

n = 101
s = np.linspace(-10,10,n)
X, Y = np.meshgrid(s, s)
Z = 10 * np.sinc(np.hypot(X, Y))
for i in range(100):
    color = pg.glColor((i,n*1.3))
    row = np.column_stack((X[i,:], Y[i,:], Z[i,:]))
    col = np.column_stack((X[:,i], Y[:,i], Z[:,i]))
    plt = gl.GLLinePlotItem(pos=row, color=color)
    w.addItem(plt)
    plt = gl.GLLinePlotItem(pos=col, color=color)
    w.addItem(plt)

app.exec_()
qimg = w.readQImage()
qimg.save('qimage.png')

```

### Expected behavior

expected

### Real behavior

qimage

### Tested environment(s)

 * PyQtGraph version: 0.12.1 
 * Qt Python binding: PySide and PyQt 
 * Python version: 3.8.9
 * NumPy version: 1.20.2 

### Additional context
Verified that on a non-hidpi system, the full screen image is saved.

I don't quite understand the following code found in GLViewWidget.py
```python
    def width(self):
        if self._dpiRatio.is_integer():
            return super().width()
        else:
            return int(super().width() * self._dpiRatio)
```
If dpr is 2.0, it takes the first branch.
If dpr is 1.5, it would take the second branch.
i.e. if I were to multiply width() by dpr to get the full image width, wouldn't that cause double multiplication when dpr is 1.5?

pijyoi:
trying to fix #1732

### DIFF
--- a/pyqtgraph/opengl/GLViewWidget.py
+++ b/pyqtgraph/opengl/GLViewWidget.py
@@ -59,10 +59,6 @@ class GLViewWidget(QtWidgets.QOpenGLWidget):
         self.makeCurrent()
 
 
-    @property
-    def _dpiRatio(self):
-        return self.devicePixelRatioF() or 1
-
     def _updateScreen(self, screen):
         self._updatePixelRatio()
         if screen is not None:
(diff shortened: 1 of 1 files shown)


j9ac9k:
I have to say, I love the diff 🤣  I'll start testing on my platforms.

j9ac9k:
Thanks for submitting the fix here @pijyoi Merging!

Issue #1731, opened 2021-04-21
Example Remote plotting: low resolution remote plot when using screen scaling


pellico:
### Short description
When I execute Remote Plotting example, the remote plot show low quality graph see below picture
### Code to reproduce
Configure screen scaling to 150% or 200%
Launch pyqtexample
```
python3 -m pyqtgraph.examples
```
Execute _Remote Plotting_ example

#### Bad picture

<img width="605" alt="bad_picture" src="https://user-images.githubusercontent.com/16132532/115558827-6b6bba80-a2b3-11eb-8481-368e05c6f3b5.PNG">

#### Better picture

If I set the environmental variable QT_AUTO_SCREEN_SCALE_FACTOR=0, picture is better but it seems that in RemoteGraphicsView a different font is used.
Compare X axis number in the below picture

<img width="402" alt="better_picture" src="https://user-images.githubusercontent.com/16132532/115560125-9571ac80-a2b4-11eb-802f-bc30a1d816a0.PNG">

### Expected behavior
Same font and graph resolution of remote plot as in local plot.

### Real behavior
Different font and low resolution image in remote plotting

### Tested environment(s)

 * PyQtGraph version: 0.12.1
 * Qt Python binding: PySide2 5.15.2 Qt 5.15.2
 * Python version: Python 3.8.7 (32bit for Windows10) (64bit in Ubuntu 20.04)
 * NumPy version: 1.20.2
 * Operating system:  Windows10,Ubuntu 20.04
 * Installation method: pip (Windows10), pipenv (Ubuntu 20.04)

### Additional context
I am using a 4K monitor (3840 x 2160 display resolution)

pijyoi:
@pellico , could you test out the master branch? A fix has been merged to address the low resolution.

pellico:
Yes it works! I have tried in Windows 10. Fonts are a little bit different see below. But it is definitely working. Thank you very much.
This is the remote 
![image](https://user-images.githubusercontent.com/16132532/115838989-7f7dfc00-a41a-11eb-9f08-8b60dc2bea98.png)
This is local:
![image](https://user-images.githubusercontent.com/16132532/115839088-97558000-a41a-11eb-94af-3aa96f497027.png)

pijyoi:
@pellico , the remote and local fonts should look the similar now in the current master branch.

pellico:
Yes I have verified on window10. Fonts are exactly the same.
Thank you

PR #1736, opened 2021-04-22
be hidpi-aware for remote rendering

pijyoi:
Get and set device pixel ratio for remote rendering.
Should address the low resolution issue (but not the font issue) mentioned in #1731.

Using examples\RemoteGraphicsView.py:
before:
before
after:
after


j9ac9k:
Thanks @pijyoi , LGTM!

PR #1738, opened 2021-04-23
use Format_RGB32 when using QImage as paint device

pijyoi:
This allows sub-pixel font antialiasing to take place.
This addresses the font issue mentioned in #1731.

Qt docs https://doc.qt.io/qt-5/qimage.html#Format-enum say:
"Rendering is best optimized to the Format_RGB32 and Format_ARGB32_Premultiplied formats"

Before (top is the remote render, bottom is the local render):
before
After (top is the remote render, bottom is the local render): 
after

One thing that I noticed is that the remote render is brighter than the local render. This includes the plot, not just the text. And this is the case even before this PR.

### DIFF
--- a/pyqtgraph/widgets/RemoteGraphicsView.py
+++ b/pyqtgraph/widgets/RemoteGraphicsView.py
@@ -80,17 +80,14 @@ class RemoteGraphicsView(QtGui.QWidget):
                 self.shm = mmap.mmap(-1, size, self.shmtag) ## can't use tmpfile on windows because the file can only be opened once.
             else:
                 self.shm = mmap.mmap(self.shmFile.fileno(), size, mmap.MAP_SHARED, mmap.PROT_READ)
-        self.shm.seek(0)
-        data = self.shm.read(w*h*4)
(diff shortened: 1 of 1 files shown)


NilsNemitz:
@pijyoi , that's neat :)
Regarding the brightness issue, this vaguely sounds like some gamma correction is applied twice.
I have no good idea how to track that down, but you could maybe check if the `QImage`s you are rendering to and displaying have a color space assigned to them? There is a `QImage.colorSpace()` method to read that back.

pijyoi:
Looks like a freshly created QImage doesn't have any color space.
```python
In [10]: qimg = QtGui.QImage(100, 100, QtGui.QImage.Format_RGB32)

In [11]: qimg.colorSpace().isValid()
Out[11]: False
```

pijyoi:
I have verified using a color picker that the colors between the remote and local plots are actually the same values. You have to compare the rendered images at the same spots.

j9ac9k:
Thanks @pijyoi for this fix!  Truth be told I've never actually used RemoteView outside of the example.  The diff looks nice too.

Merging!

pijyoi:
In fact, there are other places in the library that also need this change. ImageExporter.py for example.

Issue #1728, opened 2021-04-20
PySide6: export dialog does not show param values


pijyoi:
### Short description

With PySide6, the export dialog does not show the params' values. The other 3 bindings work fine.

### Code to reproduce

launch examples/SimplePlot.py and activate the export dialog.

### Expected behavior

![good](https://user-images.githubusercontent.com/2657027/115426314-f658a080-a232-11eb-81ea-fd0041da0d30.png)

### Real behavior

![bad](https://user-images.githubusercontent.com/2657027/115426339-fce71800-a232-11eb-920e-bbdbc09873da.png)

### Tested environment(s)

 * PyQtGraph version: 0.12.1.
 * Qt Python binding: PySide6 6.0.2 Qt 6.0.2
 * Python version: 3.8.9
 * NumPy version: 1.20.2 
 * Operating system: Windows 10 (Also Linux with similar major.minor versions)
 * Installation method: pip install -e . 

### Additional context

As far as I can tell, this issue has been present since the addition of PySide6 support (i.e. 78bc0fd3ca206a60e0ee2a1095379bee869485e3).
examples/parametertree.py seems to work fine.

j9ac9k:
now that ... is odd.  Wonder if that's a pyside6 bug.

pijyoi:
If I manually edit exportDialogTemplate_pyside6.py from
```python
        self.paramTree = ParameterTree(Form)
        __qtreewidgetitem1 = QTreeWidgetItem()
        __qtreewidgetitem1.setText(0, u"1");
        self.paramTree.setHeaderItem(__qtreewidgetitem1)
        self.paramTree.setObjectName(u"paramTree")
        self.paramTree.header().setVisible(False)
```
to
```python
        self.paramTree = ParameterTree(Form)
        self.paramTree.setObjectName(u"paramTree")
        self.paramTree.headerItem().setText(0, u"1")
        self.paramTree.header().setVisible(False)
```
then the parameter values show up.

Conversely, if I were to modify the PySide2 generated file to look like the PySide6 version, then the parameter values won't show up. So the issue is that pyside6-uic.exe (compared to the other 3 bindings) generates code that trigger some bug / mis-feature in Qt.

Not sure how to proceed from here.

j9ac9k:
Probably should commit the manual changes to the template; but of course that's not a long-term fix.

[...]

PR #1739, opened 2021-04-23
Export dialog: ParameterTree needs 2 columns

pijyoi:
To address #1728, where pyside{2,6}-uic.exe generate code that cause Parameter values to not show up.
The hand-edited version drops some redundant statements and reduces the verbosity of the original generated code.


j9ac9k:
This PR should likely remove the binding specific template files as well.  

Another reason to ditch the binding specific templates is that pyinstaller (or any freezing application) has a tendency not to be able to recognize to import them from this statement:

```python
import importlib
ui_template = importlib.import_module(
    f'.exportDialogTemplate_{QT_LIB.lower()}', package=__package__)
```

and subsequently the frozen application doesn't work until the user manually adds it to the "hidden imports" section. 

Since I already have countless virtual environments on my system for various python/qt-binding combinations, I'll check this out.  My only question is should we add a test to see if _any_ parameter value shows up (this way `tox` could potentially identify problematic python/qt combinations).

pijyoi:
I found an alternative, if a second column is added to the header of the ParameterTree, then the param values will get shown. The issue could thus be phrased as : "If you use a QWidgetTreeItem for the header, then it needs to have the same number of columns as the non-header items". This is a less invasive change than the previous commit. Whether this commit or the previous commit is preferable is arguable.

>Another reason to ditch the binding specific templates is that pyinstaller (or any freezing application) has a tendency not to be able to recognize to import them from this statement:

I see, so that's the unintended consequence of trying to simplify the code from the original if, else branches.

>My only question is should we add a test to see if any parameter value shows up (this way tox could potentially identify problematic python/qt combinations)

I don't know how to do that programmatically. Testing whether some widget got painted, that is.

j9ac9k:
> I see, so that's the unintended consequence of trying to simplify the code from the original if, else branches.

For the record, while this is a nice to have, as a pyinstaller user, I don't feel the onus is on the pyqtgraph library to work w/ pyinstaller out of the box.  Pyinstaller gives enough debug info, and enough documentation that users should be able to sort it out.  In short, if things were equal, I would say let's go the way pyinstaller is less likely to break; but I don't think we should go too far out of our way.

> I don't know how to do that programmatically. Testing whether some widget got painted, that is.

Ahh, I was wondering if we could do something like `QWidgetTreeItem.isVisible()` or something like that... either way probably not the end of the world to leave as-is.

pijyoi:
I think I found the right property to set in the ui file: columnCount. This should be the most direct fix.

The decision to hand-write the forms can be deferred.

j9ac9k:
Confirmed this works, thanks @pijyoi for not only the fix, but identifying the fix with the most minimal change.  Merging.

PR #1740, opened 2021-04-24
Improve Performance of pg.Point methods

j9ac9k:
While working on #1724 I noticed the `pg.Point` methods weren't making good use of the QPointF superclass methods at its disposal.  I decided I would modify that within Point itself, and modify calls to point object to use the modified methods.  I also deprecated some methods in favor of equivalent methods implemented with QPointF.

### DIFF
--- a/examples/GLIsosurface.py
+++ b/examples/GLIsosurface.py
@@ -7,7 +7,7 @@ This example uses the isosurface function to convert a scalar field
 ## Add path to library (just for examples; you do not need this)
 import initExample
 
-from pyqtgraph.Qt import QtCore, QtGui
+import numpy as np
 import pyqtgraph as pg
 import pyqtgraph.opengl as gl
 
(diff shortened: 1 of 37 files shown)


pijyoi:
Manhattan length is defined as abs(x) + abs(y), not the same as hypot(x, y) = sqrt(x*x + y*y)

pijyoi:
The definition of angle() for Point is non-conventional. We would expect ```p1.angle(p2)``` to mean the angle from p1 to p2.
```python
In [2]: p1 = pg.Point(1, 0)
In [3]: p2 = pg.Point(0, 1)
In [4]: p1.angle(p2)
Out[4]: -90.0
In [6]: v1 = pg.Vector(1, 0)
In [7]: v2 = pg.Vector(0, 1)
In [8]: v1.angle(v2)
Out[8]: 90.0
```

j9ac9k:
> Manhattan length is defined as abs(x) + abs(y), not the same as hypot(x, y) = sqrt(x_x + y_y)

Well, that's embarrassing 😆

j9ac9k:
umm... I screwed something up git wise, and clearly didn't think this out right to begin with, going to close this PR work on what the implementation should look like and resubmit.

PR #1741, opened 2021-04-24
Various performance improvements to pg.Point

j9ac9k:
Trying again, this PR fixes the breakage I made in 

This change makes use of QPointF methods which perform faster than the python
equivalent methods.  Furthermore, some tests are added, and I fix the breakage I made with `Point.angle()`

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/Point.py#L103

### DIFF
--- a/pyqtgraph/Point.py
+++ b/pyqtgraph/Point.py
@@ -17,7 +17,7 @@ class Point(QtCore.QPointF):
             if isinstance(args[0], QtCore.QSizeF):
                 QtCore.QPointF.__init__(self, float(args[0].width()), float(args[0].height()))
                 return
-            elif isinstance(args[0], float) or isinstance(args[0], int):
+            elif isinstance(args[0], (int, float)):
                 QtCore.QPointF.__init__(self, float(args[0]), float(args[0]))
                 return
             elif hasattr(args[0], '__getitem__'):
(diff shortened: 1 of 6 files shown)


pijyoi:
Rightly or wrongly, Point.angle() implemented in #1724 gives the same output as before #1724.
i.e. pg.Point(1, 0).angle(pg.Point(0, 1)) = -90
What I was pointing out was that pg.Point.angle() gives the reverse answer to pg.Vector.angle().
pg.Vector(1, 0).angle(pg.Vector(0, 1)) = 90

Although pg.Vector.angle() gives the convention that I expect, I don't know if the definition of Point.angle() can be changed here without breakage.
There's maybe one usage of Point.angle() within the library. It might be better to deprecate Point.angle() altogether.

j9ac9k:
> Rightly or wrongly, Point.angle() implemented in #1724 gives the same output as before #1724.
i.e. pg.Point(1, 0).angle(pg.Point(0, 1)) = -90
What I was pointing out was that pg.Point.angle() gives the reverse answer to pg.Vector.angle().
pg.Vector(1, 0).angle(pg.Vector(0, 1)) = 90

Ahh ok so I didn't break master 😅 

> Although pg.Vector.angle() gives the convention that I expect, I don't know if the definition of Point.angle() can be changed here without breakage.  

The breakage I think can be avoided by changing `p1.angle(p2)` to `p2.angle(p1)`

> There's maybe one usage of Point.angle() within the library. It might be better to deprecate Point.angle() altogether.

I hadn't considered that, but that's not out of the question.  Presently `Point.angle()` is used in `SRTTransform.py` and in a few places in `ROI.py`

pyqtgraph/Point.py
```diff
@@ -87,47 +93,67 @@ def __pow__(self, a):
...
-        return self[0]*a[1] - self[1]*a[0]
+        if not isinstance(a, QtCore.QPointF):
+            a = Point(a)
+        return self.x() * a.y() - self.y() * a.x()
         
     def proj(self, b):
         """Return the projection of this vector onto the vector b"""
         b1 = b / b.length()
-        return self.dot(b1) * b1
+        return self.dot(b) * b1
```
pijyoi, line 144: This changes the meaning. Try for instance ```pg.Point(3, 4).proj(pg.Point(0, 100))```

j9ac9k: that, was an unintentional change.

j9ac9k:
@pijyoi I modified all the calls  in the library of `p1.angle(p2)` to be `p2.angle(p1)`, so now `Point.angle()` is in line with `Vector.angle()` with regard to expected results.  I'm not against deprecating the method, but I didn't feel like that was a decision we had to make now.

j9ac9k:
well crud, was going to merge this but the static checker came back with a potentially uninitialized error for `lp1` and `lp0` in `ROI.py`

j9ac9k:
I think I'm going to merge this with the errors, as I don't know enough about ROI.py to see how this should be altered such that variables aren't potentially unbound.

Issue #1737, opened 2021-04-22
How to speed up PlotDataItem.setData()? And/or only redraw NEW points when real time plotting streaming data?


mnavaroli:
Situation:
I am using pyqtgraph (embedded in a PySide2 application) to try to plot an ECG heart signal in real time. Something akin to [the monitor in this video](https://www.youtube.com/watch?v=PPXcsZ6hh6g). I want the streaming output to look very smooth like in the video (updating the plot at ~30 Hz), but the methods I am using right now seem to be too slow to do so. I am able to get the desired output for one signal of ~600 data points, but trying to plot 3 or 4 signals each with ~600 data points is too slow (~10 Hz). 

The data:
- 6 voltage timestreams from ECG leads sampled at ~200 Hz, data loaded into a buffer that is accessed by my Python program 30 times per second (or attempts to, at least). 
- X/Y axes for plot are fixed (X goes from 0-3 seconds), so that the data streams from left to right.
- Given how fast heart signals occur, down sampling the data is tough as I want to be able to capture very high frequency effects in the data timestream.

Current methodology:
- Calling PlotDataItem.setData() 30 times per second on the last ~3 seconds of data from the buffer. My understanding is that PlotDataItem.setData() will clear the plot and redraw the entire line, but between calls I only have ~8 new data points.
- Profiling my code, I am mostly dominated by the calls to PlotDataItem.setData() .

Trying to figure out:
- Is there a way to use pyqtgraph to only draw the new ~8 data points instead of redrawing all 600 points?
- If not, are there any other packages or clever ploys to draw faster? 
- If not, is there a better language to try to be plotting real-time 2D curves?

It seems clearly doable as there are plenty of devices like in the video I linked that have the capability to plot at ~60 Hz, but I am unsure what kind of processing power they use and what languages/packages they use.

NilsNemitz:
Hi @mnavaroli,
That sounds like an exciting project!
If you can condense your code into a minimal working example, we'll be *better* able to help you.

Right now, I only have a few starting points:
- look at the speed demos in the examples:
  run `- python -m pyqtgraph.examples`
  scroll down to benchmarks and select "Line Plot Update", look at how the code does it there.
  Basically, with the data in a suitable format, we can plot a line segment in 250 ns or so. >600 data points should be possible.
- Make sure your data is in a numpy array. PyQtGraph has to convert everything else.
[...]

Issue #1712, opened 2021-04-12
Color based on value


DrPaprikaa:
Hi,
I would like to draw in real time a line plot (think reading the data from a sensor), with the points colored based on their values.
Two cases I can think of :
- The point is red or green, if < or > 0
- More broadly, a color scale : all my points are in [-1, 1], create a color map of n colors (n bins) and display the colored points.

How could I do that ?
Thanks

j9ac9k:
Before your call to `PlotDataItem.setData(newData)` you need to figure out the color you want to use.

```python
newColor = yourFunctionMappingDataToColor(newData)
PlotDataItem.setData(newData, pen=newColor)
```

j9ac9k:
Just realized you're trying to color the points, not the lines... that's a bit different.  

@lidstrom83 may know of an easier way of doing this, but the way I would do it is first, create a lookup table

```pyqtgraph
nPts = 512
colormap = pg.colormap.get(colormap_name)
valueRange = np.linspace(-1, 1, num=nPts)
colors = colormap.getLookupTable(-1.0, 1.0, nPts=nPts)
brushes = colors[np.searchsorted(valueRange, y)]
 
plotItem.setData(x, y, symbolBrush=brushes)
```

```

DrPaprikaa:
Thanks ! I can't seem to get it working tho, what shall `colormap_name` be ?
I also get the error `module 'pyqtgraph.colormap' has no attribute 'get'`

j9ac9k:
oh, if you're not looking for a continuous colormap, but want to highlight a range that's a bit different.

> Thanks ! I can't seem to get it working tho, what shall `colormap_name` be ?

This would be whatever colormap you wanted, you can see the available ones by running the Color Map example.
![image](https://user-images.githubusercontent.com/646398/114886880-cc2f5a80-9dbc-11eb-9ead-e671dea96d86.png)


I also get the error module 'pyqtgraph.colormap' has no attribute 'get'.

That's odd, here is the method, what version of pyqtgraph are you using?  The colormap various `get` methods are fairly new, but are in the last released version.

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/colormap.py#L48

> To be clear, I would like to color one point at a time, the last point that I received, depending on it's value.

Oh, if you're not looking for as continuous colormap but something more discrete, I think you can do that by the following (I haven't actually tested)

```python
nPts = 5
[...]

PR #1742, opened 2021-04-24
Make plotting with gradients more fun

NilsNemitz:
Inspired by #1712, I looked at what options we have for plotting with colors according to values.

It turns out that QPen lets us do that without problem, but currently it is necessary to manually define a gradient. As a user, it is also not obvious that the coordinates used in the definition of the gradient directly correspond to the plotted values.

This PR adds convenience methods to get a `QBrush` or `QPen` from a ColorMap object.
It extends the functionality of the already existing `getGradient()` method to handle repeating and mirrored mapping modes.

It also adds a `reverse()` method to ColorMap, which reverses an existing map, which is often convenient for imported maps.

An example is included based on the discussion of #1737, demonstrating that smoothly updating plots are possible even with fancy coloring options.


examples/ColorGradientPlots.py
```diff
@@ -0,0 +1,139 @@
...
+
+        self.traces = (
+            {'crv': curve0, 'buf': np.zeros( length ), 'ptr':0, 'ds': DataSource( signal_period=0.55 ) },
+            {'crv': curve1, 'buf': np.zeros( length ), 'ptr':0, 'ds': DataSource( signal_period=0.61, negative_period=0.55 ) },
+            {'crv': curve2, 'buf': np.zeros( length ), 'ptr':0, 'ds': DataSource( signal_period=0.65 ) },
+            {'crv': curve3, 'buf': np.zeros( length ), 'ptr':0, 'ds': DataSource( signal_period=0.52 ) },
+        )
+        self.timer = QtCore.QTimer(timerType=QtCore.Qt.PreciseTimer)
+        self.timer.timeout.connect(self.update)
+        timestamp = time.time()
```
j9ac9k, line 98: Probably should use `time.perf_counter()` for float type, or `time.perf_counter_ns()`  for int type (in ns).  Also we have [ptime.py](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/ptime.py) but I'm not sure it's actually relevant any more or not...

j9ac9k:
I just ran the example, this is slick!

The one suggestion I would have is to add a title to each plot with a short description of how it is determined what color to apply.

NilsNemitz:
Dear @j9ac9k , I have added text labels to each plot to detail what they show. I have also switched to the performance counter, although that does not matter much as we are already averaging over a few frames to get a good number.

I also got Sphinx to work on my computer and did a full pass over the color map documentation.
Let me know what you think!

doc/source/colormap.rst
```diff
@@ -1,8 +1,92 @@
...
+Sources for color maps
+----------------------
+
+Color maps can be user defined by assigning a number of *stops* over the range of 0 to 1. A color is given for each stop, and the in-between values are generated by interpolation.
+
+When map colors directly represent values, an improperly designed map can obscure detail over certain ranges of values, while creating false detail in others. PyQtGraph includes the perceptually uniform color maps provided by the `colorcet project` <https://colorcet.holoviz.org/>. Color maps can also be imported from the `colorcet` library or from `Matplotlib`, if either of these is installed.
+
+To see all available color maps, please run the ColorMap demonstration available in the suite of :ref:`examples`.
+
+Examples
```
ixjlyons, line 23: I'm a bit on the fence on this. I'm absolutely in favor of having small usage examples with visuals in the docs, but I think we should be careful about how to do it in a way that's scalable and won't lead to docs that are out of sync with the library. I've been working on an extension to [sphinx-gallery](https://sphinx-gallery.github.io/stable/index.html) which could be used to generate a gallery of the full example scripts. I've also dreamt of a sphinx extension to generate renderings of code blocks in the docs. I suppose we can always leave this in and replace it later if something like that ever gets going. Currently waiting on readthedocs to support installation of apt packages to install xvfb for rendering.

One thing we could do for now is use your scripts for generating these images and just list that code here via [`literalinclude`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) and appropriate line numbers. It might be worth converting the code to a more "script-like" structure in that case.

pyqtgraph/colormap.py
```diff
@@ -47,16 +55,25 @@ def listMaps(source=None):
...
+    
+    Parameters
+    ----------
+    name: str
+        Name of color map. In addition to the included maps, this can also 
+        be a path to a file in the local folder. See the files in the
+        `pyqtgraph/colors/maps/` folder for examples of the format.
+    source: str, optional
+        If omitted, a locally stored map is returned. Otherwise
+        - 'matplotlib' imports a map defined by Matplotlib.
```
ixjlyons, line 71: If there's no blank line between bullet points, they don't get rendered correctly. I'd just remove the bullet points and write it as a single sentence.

pyqtgraph/colormap.py
```diff
@@ -47,16 +55,25 @@ def listMaps(source=None):
...
+
+    Returns a ColorMap object from a local definition or imported from another library.
+    The generated ColorMap objects are cached for fast repeated access.
+    
+    Parameters
+    ----------
+    name: str
+        Name of color map. In addition to the included maps, this can also 
+        be a path to a file in the local folder. See the files in the
+        `pyqtgraph/colors/maps/` folder for examples of the format.
```
ixjlyons, line 68: Double backticks for RST, annoying I know

ixjlyons (commented):
Very nice demo. I've left some minor feedback mostly regarding RST formatting.

pyqtgraph/colormap.py
```diff
@@ -180,22 +197,23 @@ def getFromColorcet(name):
...
 
-    The colors for intermediate values are determined by interpolating between 
+    The colors for intermediate values are determined by interpolating between
     the two nearest colors in RGB color space.
+    
+    A ColorMap object provides access to the interpolated colors by indexing with a float value:
+    `cm[0.5]` returns a QColor corresponding to the center of ColorMap `cm`.
 
-    To provide user-defined color mappings, see :class:`GradientWidget <pyqtgraph.GradientWidget>`.
+    To provide interactively user-defined color mappings, see :class:`GradientWidget <pyqtgraph.GradientWidget>`.
```
j9ac9k, line 215: a nice shorthand here would be 

```
:class:`~pyqtgraph.GradientWidget`
```

pyqtgraph/colormap.py
```diff
@@ -330,25 +398,107 @@ def mapToFloat(self, data):
...
+        """
+        Returns a `QtGui.QLinearGradient` object corresponding to the ColorMap.
+        The span and orientiation is given by two points in plot coordinates.
+        
+        When no parameters are given for `p1` and `p2`, the gradient is mapped to the 
+        `y` coordinates 0 to 1, unless the color map is defined for a more limited range.
+        
+        Parameters
+        ----------
+        p1: QtCore.QpointF, default (0,0)
```
j9ac9k, line 414: super nit-picky, I know but should be `QtCore.QPointF`

pyqtgraph/colormap.py
NilsNemitz: I added the missing blank lines, since I found this hard to read when it mixes the 'matplotlib' and 'colorcet' explanations in the same line.

pyqtgraph/colormap.py
NilsNemitz: ⇒ changed.

pyqtgraph/colormap.py
NilsNemitz: ⇒ changed.

doc/source/colormap.rst
```diff
@@ -1,8 +1,92 @@
...
+that are colored according to their values along the horizontal or vertical axis.
+
+
+
+Sources for color maps
+----------------------
+
+Color maps can be user defined by assigning a number of *stops* over the range of 0 to 1. A color is given for each stop, and the in-between values are generated by interpolation.
+
+When map colors directly represent values, an improperly designed map can obscure detail over certain ranges of values, while creating false detail in others. PyQtGraph includes the perceptually uniform color maps provided by the `colorcet project` <https://colorcet.holoviz.org/>. Color maps can also be imported from the `colorcet` library or from `Matplotlib`, if either of these is installed.
```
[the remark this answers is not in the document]
NilsNemitz: ⇒ changed.
Colorcet doesn't seem to quite agree on whether the name is capitalized or not :)

doc/source/colormap.rst
```diff
@@ -1,8 +1,92 @@
-ColorMap
-========
+Color maps
+==========
+
+A color map defines a relationship between scalar data values and a range of colors. 
+Color maps are commonly used to generate false color images, color scatter-plot points, 
+and illustrate the height of surface plots.
+
+PyQtGraph's ColorMap object can also be used a convenient source of colors from a
```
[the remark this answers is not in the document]
NilsNemitz: ⇒ changed.
I think a good approach is to have the first reference to `ColorMap` appear as the class reference, and the just use it as plain text through the rest of the page. Do we have other guidelines for that yet?

doc/source/colormap.rst
```diff
@@ -1,8 +1,92 @@
-ColorMap
-========
+Color maps
```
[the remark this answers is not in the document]
NilsNemitz: ⇒ changed.

pyqtgraph/colormap.py
NilsNemitz: ⇒ changed.

NilsNemitz:
Dear @ixjlyons and @j9ac9k , thank you for the crash course in rst formatting. I am still making things up as I go along. I hope things work reasonably well now.

I have managed to directly pull the example code from the source files that generate the images. There is a convenient `dedent` parameter that takes out the extra spaces not needed in the examples.

I hope we'll get working support for the sphinx gallery at some point. When your library does graphics, it seems helpful to have images in the documentation!

Until it is a little clearer how this will work in practice, I'd suggest that we try to collect all the code that generates example images. This is in the images folder now, but that's only because I wasn't sure what the best location would be.

For the current sphinx format, a width of 420 pixels seems to work reasonably well for a half-width image, and that keeps the files small enough (17kB and 36 kB here) to just throw them into git for now; at least as long as we don't update them all the time :)

pyqtgraph/colormap.py
```diff
@@ -115,7 +135,7 @@ def _getFromFile(name):
     return cm
 
 def getFromMatplotlib(name):
-    """ import colormap from matplotlib definition """
+    """ Generates a ColorMap object from a Matplotlib definition. Same as ``colormap.get(name, source='matplotlib')``. """
```
j9ac9k, line 138: should this perhaps be 

```
:method:~`pyqtgraph.colormap.get`
```

that way we get links within the docs?

j9ac9k: actually, scratch that, you're demonstrating an equivalent function, not looking to reference one...

pyqtgraph/colormap.py
```diff
@@ -221,29 +239,28 @@ class ColorMap(object):
...
-                            :func:`mkColor() <pyqtgraph.mkColor>`.
-        mapping             Mapping mode (ColorMap.CLIP, REPEAT, MIRROR or DIVERGING)
-                            controlling mapping of relative index to color. String representations
-                            'clip', 'repeat', 'mirror' or 'diverging' are also accepted.
-                            CLIP maps colors to [0.0;1.0] and is the default.
-                            REPEAT maps colors to repeating intervals [0.0;1.0];[1.0-2.0],...
-                            MIRROR maps colors to [0.0;-1.0] and [0.0;+1.0] identically
-                            DIVERGING maps colors to [-1.0;+1.0]
-        ===============     =======================================================================
+        __init__(pos, color, mapping=ColorMap.CLIP)
```
j9ac9k, line 242: just out of curiosity, why is this line needed?  I see the output renders fine; but I'm just curious.

pyqtgraph/colormap.py
```diff
@@ -221,29 +239,28 @@ class ColorMap(object):
...
+        
+        Parameters
+        ----------
+        pos: array_like of float in range 0 to 1
+            Assigned positions of specified colors
+        color: array_like of colors
+            List of colors, interpreted via 
+            :func:`mkColor() <pyqtgraph.mkColor>`.
+        mapping: str or int, optional
+            Controls
```
j9ac9k, line 252: get rid of newline?

pyqtgraph/colormap.py
NilsNemitz: It makes it so that `ColorMap.CLIP` is spelled out just like that. The automatic signature shows the numerical value, and `mapping=2` is not so helpful.

pyqtgraph/colormap.py
NilsNemitz: Probably. 
Sphinx has some kind of obscure method to determine line breaks in the parameter definitions, though.

With the bonus newline, this renders as 
mapping _(str or int, optional)_ - Controls [text continues]

Without the bonus newline, I _think_ this rendered as
mapping _(str or int, optional)_ -
Controls [text continues]

I find the hanging dashes ugly...
But I agree that it is probably not worth fighting sphinx over.

⇒ will remove

examples/ColorGradientPlots.py
NilsNemitz: (already updated)

j9ac9k:
Should have merged this days ago, sorry for the wait @NilsNemitz This LGTM.  Thanks for jumping on this and taking the opportunity from the now-closed issue.

PR #1746, opened 2021-04-28
fix: cprint of remote output needs decode()

pijyoi:
To trigger the bug that this PR fixes, under Windows OS, run examples/RemoteGraphicsView.py after editing ```debug=False``` to ```debug=True```.
Debug output from the remote end will raise an exception and hang the remote end.

Local end debug output is colorized in white.
Remote end debug output is colorized in green.

This bug affects Windows OS only as the FileForwarder class is only used on Windows.

### DIFF
--- a/pyqtgraph/multiprocess/processes.py
+++ b/pyqtgraph/multiprocess/processes.py
@@ -489,12 +489,12 @@ class FileForwarder(threading.Thread):
             while not self.finish.is_set():
                 line = self.input.readline()
                 with self.lock:
-                    cprint.cout(self.color, line, -1)
+                    cprint.cout(self.color, line.decode('utf8'), -1)
         elif self.output == 'stderr' and self.color is not False:
             while not self.finish.is_set():
                 line = self.input.readline()
                 with self.lock:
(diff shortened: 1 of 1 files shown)


NilsNemitz:
@pijyoi , if FileForwarder is only used on Windows, then the fix seems good.
https://docs.python.org/3/library/sys.html :
The character encoding is platform-dependent. Non-Windows platforms use the locale encoding (see locale.getpreferredencoding()).
**On Windows, UTF-8 is used for the console device. Non-character devices such as disk files and pipes use the system locale encoding (i.e. the ANSI codepage).**

Maybe add a comment that this is windows-only, and thus encoding for the `sys` streams is known to be UTF-8  ?

pijyoi:
In fact 482dd2ee335d6dd3146417f7d520d7a0b413340f (which is a Windows-only commit) had already modified the 3rd branch to use decode('utf8')

j9ac9k:
> It seems that this 'utf8' encoding came about in Python 3.6

With every day, comes another reason I'm so happy we adopted NEP-29.

LGTM, merging.

PR #1747, opened 2021-04-28
Update symbols documentation and inverted up and down arrows of scatterplotitem

fbordignon:


j9ac9k:
Hi @fbordignon thanks for this PR.  I have been trying to think of a way to draw the symbols in the documentation, but while worrying about how to do that, the very least we can do is have the docs up to date.

fbordignon:
We can use unicode symbols such as ◇ □, but it risks issues with encoding of the source code.

j9ac9k:
> We can use unicode symbols such as ◇ □, but it risks issues with encoding of the source code.

I think the thing to do would be to make use of https://sphinx-gallery.github.io/stable/index.html once readthedocs will let us install xvfb as part of the build process.

j9ac9k:
Hmm... this is what this looks like in the docs:

image

I think you need a blank link before the first and last line of the list for the list to render correctly.

NilsNemitz:
Hi @fbordignon ,
I am a total dummy with the rst / sphinx formatting, but I think the problem is that sphinx does not like the list of symbols to be inside the table of arguments.

How about moving the table of supported symbols to the docstring of `setSymbol()`, and just referencing it in setData:
"For a list of supported symbols, see  :func:\`~ScatterPlotItem.setSymbol\`.  "
In `setSymbol`, the list can just be included in the main text of the docstring, and doesn't need to go into a table of parameters.

fbordignon:
> Hi @fbordignon ,
> I am a total dummy with the rst / sphinx formatting, but I think the problem is that sphinx does not like the list of symbols to be inside the table of arguments.
> 
> How about moving the table of supported symbols to the docstring of `setSymbol()`, and just referencing it in setData:
> "For a list of supported symbols, see :func:`~ScatterPlotItem.setSymbol`. "
> In `setSymbol`, the list can just be included in the main text of the docstring, and doesn't need to go into a table of parameters.


Good solution. Please note that originally this PR was only a documentation issue, but I've found that the arrows up and down were inverted and provided a fix for that. Please take a look at the second commit [7977c9a](https://github.com/pyqtgraph/pyqtgraph/pull/1747/commits/7977c9a3eb57558c17dc60a9140f730066a3177b)

j9ac9k:
This LGTM.  Thanks for fixing the up/down arrow bit.  @NilsNemitz any other comments, I think this is ready to merge.

NilsNemitz:
I think this is a nice improvement to the docs. A full documentation pass for ScatterPlotItem would probably be a good idea, but that shouldn't hold up putting in the missing symbols first.
Once we have a few more examples of what we want the updated https://pyqtgraph.readthedocs.io/en/latest/ pages to look like, maybe @fbordignon will come back with some additional helpful revisions? They'd certainly be welcome.

So I'd vote merge :)

fbordignon:
> I think this is a nice improvement to the docs. A full documentation pass for ScatterPlotItem would probably be a good idea, but that shouldn't hold up putting in the missing symbols first.
> Once we have a few more examples of what we want the updated https://pyqtgraph.readthedocs.io/en/latest/ pages to look like, maybe @fbordignon will come back with some additional helpful revisions? They'd certainly be welcome.
> 
> So I'd vote merge :)

Added the suggestion to my personal medium priority backlog =D

PR #1748, opened 2021-04-28
fix bug when plotting boolean arrays

outofculture:
Just converts them in `DataPlotItem.getData`. Adds a test.


j9ac9k:
Hi @outofculture this LGTM, I'm curious what the use-case here was.

outofculture:
Thanks for looking it over!

Use case: plotting a timeline of whether a particular setting is on or off (used in a comparison of other stuff happening during that time).

PR #1749, opened 2021-04-29
useCupy option should be settable after ImageItem init

outofculture:
This adds a test and fixes a bug in how we check for the useCupy option.

### DIFF
--- a/pyqtgraph/tests/test_makeARGB.py
+++ b/pyqtgraph/tests/test_makeARGB.py
@@ -6,6 +6,10 @@ from typing import Dict, Any, Union, Type
 
 from pyqtgraph import getCupy, getConfigOption, setConfigOption
 from pyqtgraph.functions import makeARGB as real_makeARGB
+try:
+    import cupy
+except ImportError:
+    cupy = None
 
 IN_2D_INT8 = np.array([[173, 48, 122, 41], [210, 192, 0, 5], [104, 56, 102, 115], [78, 19, 255, 6]], dtype=np.uint8)
 IN_RGB_INT8 = np.array(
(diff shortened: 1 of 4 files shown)


outofculture:
Huh...
```
 >>> Traceback (most recent call last):
  File "/Users/runner/work/pyqtgraph/pyqtgraph/examples/customGraphicsItem.py", line 42, in boundingRect
NameError: name 'QtCore' is not defined
```

j9ac9k:
> Huh...
> 
> ```
>  >>> Traceback (most recent call last):
>   File "/Users/runner/work/pyqtgraph/pyqtgraph/examples/customGraphicsItem.py", line 42, in boundingRect
>     return QtCore.QRectF(self.picture.boundingRect())
> NameError: name 'QtCore' is not defined
> ```

This is a bug with PyQt5 that @pijyoi identified, basically it appears in our examples, sometimes the Qt event loop keeps going, even after `sys.exit()` has returned (how does that happen?!)

j9ac9k:
Problem I have here is can we figure out how to do these tests w/o emitting warnings?

```
=============================== warnings summary ===============================
pyqtgraph/util/cupy_helper.py:13: 2 warnings
pyqtgraph/graphicsItems/tests/test_ImageItem.py: 33 warnings
pyqtgraph/graphicsItems/tests/test_ROI.py: 54 warnings
pyqtgraph/tests/test_functions.py: 1 warning
pyqtgraph/tests/test_makeARGB.py: 2 warnings
pyqtgraph/imageview/tests/test_imageview.py: 2 warnings
pyqtgraph/tests/test_ref_cycles.py: 1 warning
pyqtgraph/widgets/tests/test_histogramlutwidget.py: 2 warnings
  /home/runner/work/pyqtgraph/pyqtgraph/pyqtgraph/util/cupy_helper.py:13: UserWarning: cupy library could not be loaded, but 'useCupy' is set.
    warn("cupy library could not be loaded, but 'useCupy' is set.")
```

(wait why is `test_ref_cycles` emitting a warning?)

pyqtgraph/tests/test_makeARGB.py
```diff
@@ -4290,9 +4290,11 @@ def assert_correct(data, key, levels, lut, scale, use_rgba):
     _do_something_for_every_combo(assert_correct)
 
 
+setConfigOption("useCupy", True)
```
pijyoi, line 4293: The same Python process is used to run all the tests, including those tests outside of this file. So any config option that gets modified needs to be restored back to its old value after.

j9ac9k: I'm thinking we should make a `restore_useCuPy` fixture that will automatically revert `useCuPy` after a test that uses it is finished.  I know we could manually add a few lines to the test, that isn't terribly complicated, but it's just going to be one more thing you have to remember should you add/review a test; a fixture might be easier to understand what it's doing.

outofculture:
Thanks for catching that config option leak. This new version doesn't warn, either!

outofculture:
@j9ac9k I could turn the test skip and config reset into a single, self-contained decorator, yeah. It's only being done in two places right now, so maybe wait until the 3rd? If you want it now, though, I'm happy to make it.

j9ac9k:
> @j9ac9k I could turn the test skip and config reset into a single, self-contained decorator, yeah. It's only being done in two places right now, so maybe wait until the 3rd? If you want it now, though, I'm happy to make it.
 
I'm not partial to that, was just floating this idea as a possibility (as I recently added a `conftest.py` file).

j9ac9k:
Thanks @outofculture there is something so rewarding about having no warnings on the results of the test suite.  This LGTM, merging.

PR #1754, opened 2021-05-01
Remove pytest extensions from ci

j9ac9k:
This PR hopes to address some of the intermittent failures that we have been seeing in CI.  As the CI machines seem to not have significant resources, we should likely not run pytest-xdist.  Also, as we are not keeping track of coverage, there is no reason to collect that information at it stands.


j9ac9k:
Now if I could only figure out how to handle that exception on the RemoteGraphicsView example that raises the `pyqtgraph.multiprocess.remoteproxy.ClosedError` occasionally on macOS.

j9ac9k:
I've re-run a number of times, and only the Remote Graphics View example has been erroring on me; so I'm going to chalk that up as a success and worry about that other test later.

PR #1757, opened 2021-05-03
Support horizontal HistogramLUT{Item,Widget}

### LINKED ISSUE #99: HistogramLUTItem should support multiple orientations
Right now, `HistogramLUTItem` really only makes sense to put on the right edge of a layout. It would be really nice if `HistogramLUTItem.__init__()` took an orientation parameter (i.e., `'left'`, `'right'`, `'top'`, `'bottom'`; default to `'right'`) that made it have a layout appropriate for docking to that side of an image.

My use case is that my images are short and wide; I'd really like to put the histogram beneath them to
more efficiently use screen space.

ixjlyons:
A few things going on here (sorry):

- Support for horizontal orientation (fixes #99)
- Documentation improvements (and conversion to numpy-style docstrings)
- A few minor bug fixes and cosmetic-ish code changes

This felt very ugly to implement, so suggestions for cleaning it up would be appreciated.

I also thought perhaps we could just assume the orientation should be vertical if gradient position is left or right and horizontal otherwise. However, from a user perspective, I'd probably rather specify orientation and let the default gradient position take effect rather than use the gradient position to control orientation.

### DIFF
--- a/pyqtgraph/graphicsItems/HistogramLUTItem.py
+++ b/pyqtgraph/graphicsItems/HistogramLUTItem.py
@@ -27,7 +27,7 @@ __all__ = ['HistogramLUTItem']
 class HistogramLUTItem(GraphicsWidget):
     """
     :class:`~pyqtgraph.GraphicsWidget` with controls for adjusting the display of an
-    image.
+    :class:`~pyqtgraph.ImageItem`.
 
     Includes:
 
@@ -68,10 +68,8 @@ class HistogramLUTItem(GraphicsWidget):
     --------
     :class:`~pyqtgraph.ImageItem`
         HistogramLUTItem is most useful when paired with an ImageItem.
-
     :class:`~pyqtgraph.ImageView`
         Widget containing a paired ImageItem and HistogramLUTItem.
-
     :class:`~pyqtgraph.HistogramLUTWidget`
         QWidget containing a HistogramLUTItem for widget-based layouts.
     """
@@ -112,18 +110,19 @@ class HistogramLUTItem(GraphicsWidget):
         self.gradient = GradientEditorItem(orientation=self.gradientPosition)
         self.gradient.loadPreset('grey')
 
-        ori = 'horizontal' if self.orientation == 'vertical' else 'vertical'
+        # LinearRegionItem orientation refers to the bounding lines
+        regionOrientation = 'horizontal' if self.orientation == 'vertical' else 'vertical'
         self.regions = [
             # single region for mono levelMode
-            LinearRegionItem([0, 1], ori, swapMode='block'),
+            LinearRegionItem([0], regionOrientation, swapMode='block'),
             # r/g/b/a regions for rgba levelMode
-            LinearRegionItem([0, 1], ori, swapMode='block', pen='r',
+            LinearRegionItem([0, 1], regionOrientation, swapMode='block', pen='r',
                              brush=fn.mkBrush((255, 50, 50, 50)), span=(0., 1/3.)),
-            LinearRegionItem([0, 1], ori, swapMode='block', pen='g',
+            LinearRegionItem([0, 1], regionOrientation, swapMode='block', pen='g',
                              brush=fn.mkBrush((50, 255, 50, 50)), span=(1/3., 2/3.)),
-            LinearRegionItem([0, 1], ori, swapMode='block', pen='b',
+            LinearRegionItem([0, 1], regionOrientation, swapMode='block', pen='b',
                              brush=fn.mkBrush((50, 50, 255, 80)), span=(2/3., 1.)),
-            LinearRegionItem([0, 1], ori, swapMode='block', pen='w',
+            LinearRegionItem([0, 1], regionOrientation, swapMode='block', pen='w',
                              brush=fn.mkBrush((255, 255, 255, 50)), span=(2/3., 1.))
         ]
         self.region = self.regions[0]  # for backward compatibility.
@@ -135,7 +134,7 @@ class HistogramLUTItem(GraphicsWidget):
             region.sigRegionChanged.connect(self.regionChanging)
             region.sigRegionChangeFinished.connect(self.regionChanged)
 
-        # grad position to axis orientation
+        # gradient position to axis orientation
         ax = {'left': 'right', 'right': 'left',
               'top': 'bottom', 'bottom': 'top'}[self.gradientPosition]
         self.axis = AxisItem(ax, linkView=self.vb, maxTickLength=-10, parent=self)
@@ -400,7 +399,7 @@ class HistogramLUTItem(GraphicsWidget):
 
         Options are 'mono' or 'rgba'.
         """
-        if mode not in ('mono', 'rgba'):
+        if mode not in {'mono', 'rgba'}:
             raise ValueError(f"Level mode must be one of {{'mono', 'rgba'}}, got {mode}")
 
         if mode == self.levelMode:
--- a/pyqtgraph/graphicsItems/HistogramLUTItem.py
+++ b/pyqtgraph/graphicsItems/HistogramLUTItem.py
@@ -82,37 +82,55 @@ class HistogramLUTItem(GraphicsWidget):
     sigLookupTableChanged = QtCore.Signal(object)
     sigLevelsChanged = QtCore.Signal(object)
     sigLevelChangeFinished = QtCore.Signal(object)
-    
-    def __init__(self, image=None, fillHistogram=True, rgbHistogram=False, levelMode='mono', gradientPosition='right'):
+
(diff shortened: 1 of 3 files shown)


pyqtgraph/graphicsItems/HistogramLUTItem.py
```diff
@@ -25,177 +26,268 @@
...
-    
-    def __init__(self, image=None, fillHistogram=True, rgbHistogram=False, levelMode='mono', gradientPosition='right'):
+
+    def __init__(self, image=None, fillHistogram=True, levelMode='mono',
+                 gradientPosition='right', orientation='vertical'):
         GraphicsWidget.__init__(self)
         self.lut = None
         self.imageItem = lambda: None  # fake a dead weakref
         self.levelMode = levelMode
-        self.rgbHistogram = rgbHistogram
```
ixjlyons, line 63: `git blame` indicates this was added but never implemented

pyqtgraph/graphicsItems/HistogramLUTItem.py
```diff
@@ -25,177 +26,268 @@
...
+
         pen = self.region.lines[0].pen
-        rgn = self.getLevels()
-        p1 = self.vb.mapFromViewToItem(self, Point(self.vb.viewRect().center().x(), rgn[0]))
-        p2 = self.vb.mapFromViewToItem(self, Point(self.vb.viewRect().center().x(), rgn[1]))
+
+        mn, mx = self.getLevels()
+        vbc = self.vb.viewRect().center()
         gradRect = self.gradient.mapRectToParent(self.gradient.gradRect.rect())
+        if self.orientation == 'vertical':
```
ixjlyons, line 215: Not sure if moving this stuff all up here really improves readability compared to a block of if statements in the draw loop. This is a particular spot where it felt like there's got to be a better way

pyqtgraph/graphicsItems/HistogramLUTItem.py
```diff
@@ -25,177 +26,268 @@
...
+            self.vb.setMouseEnabled(x=False, y=True)
+        else:
+            self.vb.setMaximumHeight(152)
+            self.vb.setMinimumHeight(45)
+            self.vb.setMouseEnabled(x=True, y=False)
+
+        self.gradient = GradientEditorItem(orientation=self.gradientPosition)
         self.gradient.loadPreset('grey')
+
+        ori = 'horizontal' if self.orientation == 'vertical' else 'vertical'
```
j9ac9k, line 115: is this right?  shouldn't it be

```python
ori = 'horizontal if self.orientation == 'horizontal' else 'vertical'
```

pyqtgraph/graphicsItems/HistogramLUTItem.py
```diff
@@ -343,15 +444,15 @@ def _showRegions(self):
             self.regions[0].setVisible(True)
             self.gradient.show()
         else:
-            raise ValueError("Unknown level mode %r" %  self.levelMode) 
-    
+            raise ValueError("Unknown level mode {}".format(self.levelMode))
```
j9ac9k, line 447: this is an improvement, but since we're on python 3.7 as a minimum, you could use a f-string if you wanted to

pyqtgraph/graphicsItems/HistogramLUTItem.py
ixjlyons: Yeah I considered adding a comment here because it's confusing. For `LinearRegionItem`, orientation refers to the orientation of the *lines*.

pyqtgraph/graphicsItems/HistogramLUTItem.py
ixjlyons: Heh, forgot I even did this. f-strings are good for this kind of thing so I'll go ahead and update this and a couple others.

pyqtgraph/graphicsItems/HistogramLUTItem.py
j9ac9k: Oh right, may I suggest renaming the variable to `regionOrientation` or something along those lines.

ixjlyons:
Somehow forgot to add the new keyword arg to the docstring. Added.

pyqtgraph/graphicsItems/HistogramLUTItem.py
j9ac9k: that would be a cool feature tho!

pyqtgraph/graphicsItems/HistogramLUTItem.py
```diff
@@ -277,56 +371,66 @@ def getLevels(self):
...
-            assert None not in (min, max)
             self.region.setRegion((min, max))
         else:
             if rgba is None:
-                raise TypeError("Must specify rgba argument when levelMode != 'mono'.")
-            for i, levels in enumerate(rgba):
-                self.regions[i+1].setRegion(levels)
-        
+                rgba = 4*[(min, max)]
+            for levels, region in zip(rgba, self.regions[1:]):
```
j9ac9k, line 397: just wanted to say I love this change 👍🏻

pyqtgraph/graphicsItems/HistogramLUTItem.py
ixjlyons: I think `levelMode='rgba'` does what this was probably intended to mean. Check `examples/HistogramLUT.py` which has a toggle to switch between mono/rgba.

ixjlyons:
> Somehow forgot to add the new keyword arg to the docstring. Added.

I had also forgotten to add 'top', and 'bottom' to the `gradientPosition` doc. Added.

pyqtgraph/graphicsItems/HistogramLUTItem.py
j9ac9k: oh cool, it is implemented there, that looks really slick

j9ac9k:
Thanks @ixjlyons for this PR, this looks good, the improvement in documentation and code-readability is greatly appreciated, merging!

pijyoi:
What is the expected behavior when an RGBA image is used?
In the example below, the histograms seem to be "crushed" by the alpha channel.
```python
import numpy as np
import scipy.misc

import pyqtgraph as pg

app = pg.mkQApp()
win = QtWidgets.QMainWindow()
win.resize(880, 600)
win.show()

cw = QtWidgets.QWidget()
cw.show()
win.setCentralWidget(cw)

layout = QtWidgets.QGridLayout()
cw.setLayout(layout)
layout.setSpacing(0)

view = pg.GraphicsView()
vb = pg.ViewBox()
vb.setAspectLocked()
vb.invertY(True)
view.setCentralItem(vb)
layout.addWidget(view, 0, 1, 3, 1)

hist = pg.HistogramLUTWidget(gradientPosition="left")
layout.addWidget(hist, 0, 2)

monoRadio = QtWidgets.QRadioButton('mono')
rgbaRadio = QtWidgets.QRadioButton('rgba')
layout.addWidget(monoRadio, 1, 2)
layout.addWidget(rgbaRadio, 2, 2)
monoRadio.setChecked(True)

def setLevelMode():
    mode = 'mono' if monoRadio.isChecked() else 'rgba'
    hist.setLevelMode(mode)

monoRadio.toggled.connect(setLevelMode)

face = scipy.misc.face()
data = np.full_like(face, 255, shape=face.shape[:2] + (4,))
data[..., :3] = face

img = pg.ImageItem(data, axisOrder='row-major')
vb.addItem(img)
vb.autoRange()

hist.setImageItem(img)

app.exec_()
```

ixjlyons:
I noticed that as well. The histogram comes from `ImageItem.getHistogram` and, well, it's true that if your alpha channel is all 255s, that bin is going to tower over everything whether the histogram is computed on the flattened image or per channel. Suggestions welcome for how to handle it. I could imagine an option to ignore the alpha channel or allowing more flexibility on the "count" axis of the histogram. Maybe this case is common enough to handle automatically with a special case, i.e. detect when a histogram bin value is >= the image width*height. That wouldn't really be robust for 'mono' mode with an RGB image though. For 'rgba' mode, normalizing the histograms could help if each curve gets its own density axis. Maybe there's something obvious I'm not thinking of.

pijyoi:
Well, the question is whether it is meaningful to manipulate the "levels" of the alpha channel, The (common?) use-case for alpha channel seems to be for the whole image to take only 2 distinct values: 0 and some other value (typically 255 or 1.0). One could say that it is wrong usage to supply an RGBA image to HistogramLUT.

ixjlyons:
I agree it's probably not too common to want to control or even see the alpha level, but I'm not sure I'd consider it incorrect usage to pair an `ImageItem` with an alpha channel together with a `HistogramLUTItem`. I'm thinking we could add an "rgb" `levelMode`.

PR #1766, opened 2021-05-07
VideoSpeedTest.py : Add numba checkbox

pijyoi:
Allows user to interactively enable/disable numba to see its effect on the frame rate


j9ac9k:
Thanks @pijyoi this LGTM!  Worked nicely on macOS. Merging!

PR #1768, opened 2021-05-07
Format_RGB32 will always have alpha equal 255

pijyoi:
this also removes the assumption that we are on little-endian.

```python
In [18]: qimg = QtGui.QImage(1, 1, QtGui.QImage.Format.Format_RGB32)

In [19]: qimg.fill(QtCore.Qt.GlobalColor.transparent)

In [20]: np.frombuffer(qimg.bits(), dtype=np.uint8)
Out[20]: array([  0,   0,   0, 255], dtype=uint8)
```

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1503,7 +1503,6 @@ def imageToArray(img, copy=False, transpose=True):
     the QImage is collected before the array, there may be trouble).
     The array will have shape (width, height, (b,g,r,a)).
     """
-    fmt = img.format()
     img_ptr = img.bits()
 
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks @pijyoi this LGTM.  Sorry this sat in the queue for so long.  Merging.

pijyoi:
Not sure how, but this PR's changes seem to have become undone.
It wasn't so much to remove the redundant operation as to remove the little-endian assumption.
Could perhaps just migrate the 2 remaining uses of ```imageToArray()``` to ```qimage_to_ndarray()```

Issue #1770, opened 2021-05-12
PySide6 6.1 Test Suite Errors


j9ac9k:
Just saw PySide6 6.1 was released, so I installed it and ran against the test suite, we had a few errors and warnings which weren't there before:

```
================================================ warnings summary ================================================
pyqtgraph/graphicsItems/tests/test_InfiniteLine.py: 1 warning
pyqtgraph/graphicsItems/tests/test_ROI.py: 14 warnings
  /Users/ogi/Developer/pyqtgraph/pyqtgraph/widgets/GraphicsView.py:363: DeprecationWarning: Function: 'localPos() const' is marked as deprecated, please check the documentation for more information.
    lpos = ev.localPos()

-- Docs: https://docs.pytest.org/en/stable/warnings.html
============================================ short test summary info =============================================
FAILED pyqtgraph/tests/test_Vector.py::test_Vector_init - TypeError: 'PySide6.QtGui.QVector3D.__init__' called ...
FAILED pyqtgraph/tests/test_srttransform3d.py::testMatrix - TypeError: 'PySide6.QtGui.QVector3D.__init__' calle...
FAILED examples/test_examples.py::testExamples[ RemoteGraphicsView.py - PySide6 ]
```

Also looks like the `remoteproxy.py` raised an error

```
[gw7] darwin -- Python 3.9.1 /Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.9.1/envs/pyqtgraph-pyside6-py39/bin/python3.9
0

Python 3.9.1 (default, Dec 20 2020, 08:49:26)
[Clang 12.0.0 (clang-1200.0.32.28)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> >>> ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... /Users/ogi/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py:506: RemoteExceptionWarning: ===== Remote process raised exception on request: =====
  warnings.warn("===== Remote process raised exception on request: =====", RemoteExceptionWarning)
/Users/ogi/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py:507: RemoteExceptionWarning: Traceback (most recent call last):
  File "/Users/ogi/Developer/pyqtgraph/pyqtgraph/multiprocess/remoteproxy.py", line 222, in handleRequest
    result = getattr(opts['obj'], opts['attr'])
[...]

PR #1771, opened 2021-05-13
some fixes PySide6 6.1.0

pijyoi:
fix the easy parts of #1770 for PySide6 6.1.0.
i.e. RemoteGraphicsView.py still fails

### DIFF
--- a/pyqtgraph/dockarea/Dock.py
+++ b/pyqtgraph/dockarea/Dock.py
@@ -198,7 +198,7 @@ class Dock(QtGui.QWidget, DockDrop):
         self.drag.setMimeData(mime)
         self.widgetArea.setStyleSheet(self.dragStyle)
         self.update()
-        action = self.drag.exec_() if hasattr(self.drag, 'exec_') else self.drag.exec()
+        action = self.drag.exec() if hasattr(self.drag, 'exec') else self.drag.exec_()
         self.updateStyle()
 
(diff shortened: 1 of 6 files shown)


pyqtgraph/dockarea/Dock.py
```diff
@@ -198,7 +198,7 @@ def startDrag(self):
         self.drag.setMimeData(mime)
         self.widgetArea.setStyleSheet(self.dragStyle)
         self.update()
-        action = self.drag.exec_() if hasattr(self.drag, 'exec_') else self.drag.exec()
+        action = self.drag.exec() if hasattr(self.drag, 'exec') else self.drag.exec_()
```
ksunden, line 201: this just simplifies to `self.drag.exec` now

pijyoi: Qt.py provides Qt5 shim for QApplication.exec() but not for QDialog.exec()

ksunden: I mean... the true and false cases are exactly the same in this diff

ksunden: nevermind, I misread it

j9ac9k:
thanks @pijyoi !  I'll try and take a closer look at the `RemoteProxy.py` error and see if I can get to the bottom of that one...

PR #1772, opened 2021-05-13
don't print out expected Exception

pijyoi:
this fixes the RemoteGraphicsView portion of #1770.

I _think_ this arises from https://bugreports.qt.io/projects/PYSIDE/issues/PYSIDE-1523?filter=allissues.
In a PySide6 6.1.0 installation, within the file Lib/site-packages/PySide6/glue/qtcore.cpp, there is code that probes for some attribute "im_func". In such a case, failure is considered normal. However, remoteproxy.py reports all exceptions, which causes CI to fail.

The quick and dirty workaround done here is to avoid printing out this particular exception.

The real issue is that not all exceptions are errors.


j9ac9k:
Thanks for digging into this @pijyoi, I'm good with this "fix", Merging.

PR #1773, opened 2021-05-13
Add a Code of Conduct

outofculture:
This should help our community!

@j9ac9k and @ntjess you two should especially look at this, seeing as you're both included as Responsible Community Leaders 😁

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -51,9 +51,7 @@ Examples of unacceptable behavior include:
 ## Where does this code of conduct apply
 
 This Code of Conduct applies within all online community spaces, official organized meetups, and when an individual is
-officially representing the community in public spaces. Examples of representing our community include using an official
-e-mail address, posting via an official social media account, or acting as an appointed representative at an online or
-offline event.
+officially representing the community in public spaces, online or offline.
 
 ## What to do in case of violations
 
@@ -92,7 +90,9 @@ or harmful.
 
 Community leaders have the right and responsibility to remove, edit, or reject comments, commits, code, wiki edits,
 issues, and other contributions that are not aligned to this Code of Conduct, and will communicate reasons for
-moderation decisions when appropriate.
+moderation decisions when appropriate. This removal should be as transparent as possible, with information available in
+the same scope of context as the original moderated content, linking to an incident report, the acting moderators,
+and/or this document as appropriate.
 
 All community leaders are obligated to respect the privacy and security of the reporter of any incident.
 
@@ -122,13 +122,14 @@ case. Resolutions might include:
 A response must be sent within one week to the person who filed the report with either a resolution or an explanation of
 why the situation is not yet resolved.
 
-A Code of Conduct transparency report will then be published with anonymized information about any violations that might
-have occurred. This report should be handled with care not to divulge personally identifying information about victims,
-reporters, and violators, and should serve as a means to ensure that members will be comfortable reporting violations
-and that our community will be kept accountable for supporting and encouraging safe spaces.
+A Code of Conduct transparency report will then be published
+to [the PyQtGraph users' list](https://groups.google.com/g/pyqtgraph) with anonymized information about any violations
+that might have occurred. This report should be handled with care not to divulge personally identifying information
+about victims, reporters, and violators, and should serve as a means to ensure that members will be comfortable
+reporting violations and that our community will be kept accountable for supporting and encouraging safe spaces.
 
 ## Attribution
 
 This Code of Conduct is adapted from
-the [Contributor Covenant, version 2.0][https://www.contributor-covenant.org/version/2/0/code_of_conduct.html] and
+the [Contributor Covenant, version 2.0](https://www.contributor-covenant.org/version/2/0/code_of_conduct.html) and
 the [PuPPy Code of Conduct](https://www.pspython.com/pages/code-of-conduct/).
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,134 @@
+# Code of Conduct
+
+Diversity is one of our huge strengths, but it can also lead to communication issues. To support a welcoming environment
+for all, regardless of individual differences, we have a few ground rules that we ask people to adhere to when they
+participate in this community. These rules apply equally to founders, organizers, contributors, users and affiliates —
+in short, to all participants.
+
+This isn’t an exhaustive list of things that you must do, or can’t do. Rather, take it in the spirit in which it’s
+intended. It’s a guide to make it easier to enrich all of us and the technical communities in which we participate, and
+which we represent.
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our community a harassment-free experience for
+everyone, regardless of background or identity. This includes, but is not limited to, members of any race, ethnicity,
+culture, national origin, color, immigration status, social and economic class, educational level, sex, sexual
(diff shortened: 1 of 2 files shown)


j9ac9k:
Thanks @outofculture for working on this, I'll get to reading right now!

ntjess:
It looks good to me! A few minor clarifications:

> Community leaders have the right and responsibility to remove, edit, or reject comments, commits, code, wiki edits,
> issues, and other contributions that are not aligned to this Code of Conduct, and will communicate reasons for
> moderation decisions when appropriate.

In the extreme (and unlikely) case of moderators who become content gatekeepers, may I propose that the "communicate reasons for moderation when appropriate" include some clause that this communication will be visible at least in the same scope as the original offense (i.e. if a commit is blacklisted, the communication regarding this decision should be made public in that same scope as so as to clearly indicate which moderators were responsible for the decision and the justification to whoever looks back). Not sure if this is required, but again this language would prevent content moderators from making decisions behind closed doors and silently 'gatekeeping' allowed content.

> A Code of Conduct transparency report will then be published with anonymized information about any violations that might have occurred.
I suppose this brings up the same question in my mind -- is it worth specifying where the report will be published if required?

j9ac9k:
> Examples of representing our community include using an official e-mail address

Would suggest rewording to

> representing our community include using official e-mail correspondence 

As we don't have our own email domain...

Also it appears one of the links at the bottom is not formatted correctly.

j9ac9k:
I think this is good @outofculture and I'm good w/ merging/implementing.

janto:
Can we please not have of a code of conduct like this? From my limited involvement with this project, I've found the community pleasant enough :smiley:  It does not need to codify required behaviour.

There are problems with these CoC's, a few are
* Depriving the accused of facing their accuser. Assumptions of guilt. Judgement resting on the accuser feeling offended (or feeling offended for some hypothetical other)
* Handing over moral judgement to a small self-selected (often ideologically singular) committee. Ironically these committees are likely to lack in political diversity. 
* The problem of assuming kindness as the highest virtue. As opposed to, say, truthfulness and sincerity that I would much rather prefer from a scientific / data oriented community.
* Various chilling effects
* It excludes disagreeable geniuses :smiling_imp:

I am well aware that these CoCs have snuck into many other OSS projects. Hopefully these mistakes can be rectified.

Disagreement welcome :stuck_out_tongue_winking_eye:

j9ac9k:
Hi @janto 

We will consider modifications for a CoC, but having been previously a member of online communities that become toxic, due to a lack of respect by some individuals; I will do everything in my power to ensure that doesn't happen here.  Having a CoC, while not a guarantee of everything going right, sets both standards and expectations for members.

But to your points, I'm going to shuffle the order a little bit to address them.

> Depriving the accused of facing their accuser. Assumptions of guilt. Judgement resting on the accuser feeling offended (or feeling offended for some hypothetical other)

I'm not sure how you got this, judgment is made by CoC members after information is gathered.  There is most certainly a higher standard than an accuser feeling offended.  The first bullet point on potential outcomes is "Nothing (if we determine no violation occurred)."  Also the CoC committee members do not determine guilt.  We seek to evaluate violations of the CoC.

> Handing over moral judgement to a small self-selected (often ideologically singular) committee. Ironically these committees are likely to lack in political diversity.

Well, we are a group of volunteers, I don't see an alternative to self-selection.  I'm certainly not going to request people that don't want to partake in these kinds of evaluations take on that responsibility. In terms of being ideologically singular; maybe that's the case, maybe not, I really don't know.  All I know is the 3 of us agreed on the contents of the CoC.  We all have different backgrounds and experiences.

> Various chilling effects

We are certainly open to hearing of cases if someone thinks they were mistreated.  I do have some some CoC related failures in mind that have occurred in other online communities.  I have seen some cases of CoC committee actions have gone wrong.  In all the cases that come to mind, there have been very public apologies made.  I have also seen CoC committee actions gone right before, and it has made communities strong/better in the process.  That's what we aim to do here.

> The problem of assuming kindness as the highest virtue. As opposed to, say, truthfulness and sincerity that I would much rather prefer from a scientific / data oriented community.
> It excludes disagreeable geniuses 😈

There is nothing wrong with being disagreement.  This is often how the sausage is made; there is no one maintainer/contributor that has a complete handle of this library or the dependencies it uses.  Many changes, especially new features, are developed through disagreement, and eventually a consensus if all goes right.  If individuals  have no interest in being kind and respectful to others in the process, regardless of the knowledge and experience they may bring, I (and I am speaking only for myself here) would rather not have their involvement in this project. That's how you end up with a situation like RMS and FSF. 

If there is a specific change you would like to see made, we are certainly open to suggestions, but these concerns do not rise to the level of outright discarding a CoC IMO.

outofculture:
I don't have much else to add, but a few questions.

>     * Handing over moral judgement to a small self-selected (often ideologically singular) committee. Ironically these committees are likely to lack in political diversity.

I share this concern. I don't actually know much about the ideological positions of either of the other two members of the committee, but it's at least true that we all thought a CoC was a good idea 😄

Relatedly, this CoC doesn't address the problem of "what to do if the CoC committee violates the CoC". It does specify that everyone is bound by its rules, but the committee shouldn't be in charge of ruling on their own behavior. I wish we were under the NUMFocus umbrella, and we could rely on them to be our arbiter.

In any case, it bears mentioning that the committee _already_ has this power, without a CoC in place: we are all already maintainers. Having a CoC means we now have guidance on when and how we should exercise that power. Compared to the status quo of "maintainers may act arbitrarily", a CoC serves as a declaration that we expect them (us) to behave better than that.

>     * Various chilling effects

What specific _positive_ behaviors would this CoC discourage? Do you worry it would cut out all attempts at humor, for fear of stepping on toes? ( Or worse, that everyone would exclusively use puns as the only safe form of humor?! If that should ever come to pass, rest assured, the world would burn. ) False accusation abuse also jumps to mind, but being an almost-entirely online community, the evidence is never likely to be in doubt, only the interpretation. I could see us adding some language warning about this possible abuse, but as a failure mode, it will have to defeat the committee's good intentions toward the entire community, and they'll hopefully spot this.

janto:
Hey @outofculture, @j9ac9k 

The CoC makes it more difficult to objectively address politically sensitive policies.

> a CoC serves as a declaration that we expect them (us) to behave better than that.

a CoC declares a certain *type* of expected behavior and implies what "better" means.

What problem do you foresee with replacing "kindness" with "truthful and sincere"? I think it makes more sense for a data community.

ntjess:
@janto thanks for weighing in.

I would argue that increased care in addressing sensitive policies can be a _good_ thing when thinking twice about what is posted helps avoid damaging remarks (as often simple rephrasing does). I would re-emphasize the purpose of this CoC in PyQtGraph speaks more about attacking _people_ than policies. That being said, if you have examples of how topic or policy discussion is _prevented_ as a result of a CoC we would be happy to answer more specific concerns.

> What problem do you foresee with replacing "kindness" with "truthful and sincere"?

There can be big differences between sincerity and kindness when remarks are not made in good faith. For instance, "Only an idiot would write this code" or "you must be blind, because there's no other reason you could've made that mistake" are both remarks that can be considered "sincere", but that we wouldn't want to encourage in PyQtGraph spaces.

> I think it makes more sense for a data community.

Including "kindness" means discussions hopefully focus more on the _technical_ aspects of what was wrong rather than the _person_ doing the coding, or at least not include remarks about them or their abilities. Using kindness as a baseline rather than sincerity encourages those in a discussion to look for common ground and understanding first. IMO this is more helpful than hindering to a data community.

As before if you can refer to specific instances in which "kindness" was so subjective that it led to suppression of technical expression we can certainly take those cases into account if needed.

PR #1774, opened 2021-05-13
implement pg.exec()

pijyoi:
Related to #1770. PySide6 6.1 deprecates exec_() thus running the examples on PySide6 6.1 have a deprecation warning.
This PR implements pg.exec() and changes all examples to make use of it.

The exec() and exec_() shims added in #1771 are reverted. The intent then was to change the examples from ```pg.mkQApp().exec_()``` to ```pg.mkQApp().exec()```. They had not actually been used yet.

### DIFF
--- a/examples/Arrow.py
+++ b/examples/Arrow.py
@@ -51,4 +51,4 @@ anim = a.makeAnimation(loop=-1)
 anim.start()
 
 if __name__ == '__main__':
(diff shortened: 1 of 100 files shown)


j9ac9k:
@pijyoi thank you for this PR; and keeping the number of warnings to a nice round number of 0.

Issue #1760, opened 2021-05-04
PyOpenGL rendering broken after upgrade from 0.11.1 to 0.12.1


noxobscura:
### Short description
The Data lines (paths) are not correctly displayed using pyopengl after updating from 0.11.1 to 0.12.1.

### Code to reproduce

```python
import numpy
import cProfile
import logging

import pyqtgraph as pg
from PyQt5 import QtCore,uic
from PyQt5.QtGui import *
from PyQt5.QtCore import QRect, QTimer

try:
    import OpenGL
    pg.setConfigOption('useOpenGL', True)
    pg.setConfigOption('enableExperimental', True)
except Exception as e:
    print(f"Enabling OpenGL failed with {e}. Will result in slow rendering. Try installing PyOpenGL.")

    

def program(columns=8, samples=10000, channels=64):
    app = QApplication([])
    win = pg.GraphicsLayoutWidget()
    pg.setConfigOptions(imageAxisOrder='row-major')
    win.resize(1280,768)
    win.ci.layout.setSpacing(0)
    win.ci.layout.setContentsMargins(0,0,0,0)
    win.show()

    data            = numpy.zeros((samples, channels+1))
    plots           = [win.addPlot(row=i/columns+1,col=i%columns) for i in range(channels)]
    curves          = list()

    x = numpy.linspace(0, 1, samples, endpoint=True)
    f = 2 # Frequency in Hz
    A = 1 # Amplitude in Unit
    y = A * numpy.sin(2*numpy.pi*f*x).reshape((samples,1)) # Signal

    data[:,0]   = x
    data[:,1:]  = numpy.repeat(y, channels, axis=1)
    
    for chn_no,p in enumerate(plots, 1):
        c       = pg.PlotCurveItem(pen=(chn_no,channels * 1.3))
        p.addItem(c)
        curves.append((c, chn_no))
          
    def update():
        nonlocal data

        data[:,1:] = numpy.roll(data[:,1:], 100, axis=0)
            
        for curve,data_index in curves:
            curve.setData(data[:,0],data[:,data_index])

    timer = QTimer()
    timer.timeout.connect(update)
    timer.start(30)
    return app.exec_()   


if __name__ == "__main__":
    logging.basicConfig(level=logging.INFO)
    #cProfile.run("program()", sort="cumtime")
    program()
```

### Expected behavior
Before updating to 0.12.1 it looks like this:
[...]

PR #1776, opened 2021-05-15
Add "GLTextItem"

shikishima-TasakiLab:
I created an item that can display text in a GLViewWidget using GLUT.

```python
gvw = GLViewWidget()

item = GLTextItem(pos=(x, y, z), color=(r, g, b, a), text='string to display', font=GLUT_BITMAP_HELVETICA_18)
gvw.addItem(item)

item.setData(pos=(x, y, z), color=(r, g, b, a), text='string to display', font=GLUT_BITMAP_HELVETICA_18)
```


pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -0,0 +1,77 @@
...
+                    elif isinstance(value, tuple) or isinstance(value, list):
+                        if len(value) != 3:
+                            raise Exception('"len(color)" must be 4.')
+                elif arg == 'font':
+                    if value not in [GLUT_BITMAP_8_BY_13, GLUT_BITMAP_9_BY_15, GLUT_BITMAP_TIMES_ROMAN_10, GLUT_BITMAP_TIMES_ROMAN_24, GLUT_BITMAP_HELVETICA_10, GLUT_BITMAP_HELVETICA_12, GLUT_BITMAP_HELVETICA_18]:
+                        raise Exception('"font" must be "GLUT_BITMAP_8_BY_13", "GLUT_BITMAP_9_BY_15", "GLUT_BITMAP_TIMES_ROMAN_10", "GLUT_BITMAP_TIMES_ROMAN_24", "GLUT_BITMAP_HELVETICA_10", "GLUT_BITMAP_HELVETICA_12", or "GLUT_BITMAP_HELVETICA_18".')
+                setattr(self, arg, value)
+        self.update()
+    
+    def initializeGL(self):
```
j9ac9k, line 66: I this needed if it just calls the super-class?

j9ac9k:
Hi @shikishima-TasakiLab 

Thanks for the PR!  This functionality is what is needed to get #1711 ready for merging I believe.

I'll hop on this branch and mess with it in the next day or so.  Would you be willing to add an example to the example directory?  

Thanks again!

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: This is not needed. It's not needed. I removed it in [0e8c31a](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/0e8c31a93f81a2c03411b9178c8adc57f3fa7438).

shikishima-TasakiLab:
@j9ac9k 
I added an example in [7c2c13c](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/7c2c13ca7eccdf583ef06042e99e17c198a84bb1).

GLTextItem

If you get the following error, you may need to install "freeglut3".
```
OpenGL.error.NullFunctionError: Attempt to call an undefined function glutInit, check for bool(glutInit) before calling
```

pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -0,0 +1,74 @@
...
+            if k not in args:
+                raise Exception('Invalid keyword argument: %s (allowed arguments are %s)' % (k, str(args)))
+        for arg in args:
+            if arg in kwds:
+                value = kwds[arg]
+                if arg == 'pos':
+                    if isinstance(value, np.ndarray):
+                        if value.shape != (3,):
+                            raise Exception('"pos.shape" must be (3,).')
+                    elif isinstance(value, tuple) or isinstance(value, list):
```
j9ac9k, line 50: we can combine these in one check

```python
elif isinstance(value, (tuple, list)):
```

pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -0,0 +1,74 @@
...
+                            raise Exception('"len(pos)" must be 3.')
+                elif arg == 'color':
+                    if isinstance(value, np.ndarray):
+                        if value.shape != (4,):
+                            raise Exception('"color.shape" must be (4,).')
+                    elif isinstance(value, tuple) or isinstance(value, list):
+                        if len(value) != 4:
+                            raise Exception('"len(color)" must be 4.')
+                elif arg == 'font':
+                    if value not in [GLUT_BITMAP_8_BY_13, GLUT_BITMAP_9_BY_15, GLUT_BITMAP_TIMES_ROMAN_10, GLUT_BITMAP_TIMES_ROMAN_24, GLUT_BITMAP_HELVETICA_10, GLUT_BITMAP_HELVETICA_12, GLUT_BITMAP_HELVETICA_18]:
```
j9ac9k, line 61: would you mind breaking up this list so that it's not super long?

```python
if value not in [
    GLUT_BITMAP_8_BY_13,
    GLUT_BITMAP_9_BY_15,
    GLUT_BITMAP_TIMES_ROMAN_10,
    GLUT_BITMAP_TIMES_ROMAN_24,
    GLUT_BITMAP_HELVETICA_10,
    GLUT_BITMAP_HELVETICA_12,
    GLUT_BITMAP_HELVETICA_18
]:
```

examples/GLTextItem.py
```diff
@@ -0,0 +1,24 @@
...
+
+txtitem1 = gl.GLTextItem(pos=(0.0, 0.0, 0.0), text='text1')
+gvw.addItem(txtitem1)
+
+txtitem2 = gl.GLTextItem()
+txtitem2.setData(pos=(1.0, -1.0, 2.0), color=(0.5, 1.0, 0.5, 1.0), text='text2')
+gvw.addItem(txtitem2)
+
+if __name__ == '__main__':
+  mkQApp().exec_()
```
j9ac9k, line 24: we _just_ changed this (like I merged the PR with this diff a few minutes ago).  

Can you change this to 

```python
pg.exec()
```

This gives Qt 6.1 compatibility without emitting deprecation warnings (looks like you'll have to `import pyqtgraph as pg` at the top too

examples/GLTextItem.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+"""
+Simple examples demonstrating the use of GLTextItem.
+
+"""
+
```
j9ac9k, line 6: please add this line as the first import statement

```python
import initExample ## Add path to library (just for examples; you do not need this)
```

We do this to support examples running when pyqtgraph hasn't actually been installed via `pip`

pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -0,0 +1,74 @@
...
+from OpenGL.GLUT import *
+import numpy as np
+from pyqtgraph.opengl.GLGraphicsItem import GLGraphicsItem
+
+class GLTextItem(GLGraphicsItem):
+    """Draws text in 3D."""
+
+    pos:Union[np.ndarray, List[float]] = np.array([0.0, 0.0, 0.0])
+    color:np.ndarray = np.array([1.0, 1.0, 1.0, 1.0])
+    text:str = ''
```
j9ac9k, line 12: `pos`, `color` and `text` should almost certainly not be class variables, but object properties, so should probably move them into the `__init__` method.

If you were using lists instead of numpy arrays, it may be possible to move the position of all of these TextItems, by mutating the list (vs. just the instance you want to change).

pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -0,0 +1,74 @@
...
+        text                  String to display.
+        font                  GLUT font. (Default: GLUT_BITMAP_HELVETICA_18)
+        ====================  ==================================================
+        """
+        args = ['pos', 'color', 'text', 'font']
+        for k in kwds.keys():
+            if k not in args:
+                raise Exception('Invalid keyword argument: %s (allowed arguments are %s)' % (k, str(args)))
+        for arg in args:
+            if arg in kwds:
```
j9ac9k, line 44: probably don't need this if-check given you're checking for this in the for loop above.

pyqtgraph/opengl/items/GLTextItem.py
j9ac9k: EDIT, actually you can likely combine this with the if check earlier; since `np.ndarray` supports `len()` 

```python
if arg == 'pos':
    if len(value) != 3:
        raise Exception(....)
```

j9ac9k:
type-annotations; suppose it was a matter of time before the first PR came w/ those.  

@ixjlyons @outofculture how do you feel about them, should we accept PRs that have them already (assuming they're right of course), ask to remove them, thoughts?  I'm of the mind-set we should likely remove them as we don't have a system for verifying them.  I do want to have these at some point, but I'm not sure doing the type-annotations inline is going to be feasible (try the type annotations for `functions.mkColor`, I'll wait...).  We can probably do type-annotations via stub files (`.pyi`) but I'm not sure we're ready to have that discussion yet.

@shikishima-TasakiLab thanks for the update and the example, that's perfect.  I did just merge a PR that modified how we `exec` the QApplication event loop to avoid deprecation warnings with PySide6 6.1, I made a comment there, would suggest making that change.

I also made some small changes regarding trying to get the line-width down and combining some `isinstance` checks.

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: > we can combine these in one check
> 
> ```python
> elif isinstance(value, (tuple, list)):
> ```

Thanks. I have increased my knowledge.
I fixed it in [1ba5203](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/1ba52032d743a7ee03941576a40f9893fe13fa04).



> EDIT, actually you can likely combine this with the if check earlier; since `np.ndarray` supports `len()`
> 
> ```python
> if arg == 'pos':
>     if len(value) != 3:
>         raise Exception(....)
> ```

It is true that `np.ndarray` supports `len()`. However, it will pass the check in the following case.
```python
pos = np.array([
    [0.1, 1.2, 5.3, 4.6],
    [4.5, 6.8, 4.7, 8.6],
    [5.6, 8.7, 1.9, 2.0]
])

print(len(pos))
```
```
3
```

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: I fixed it in [f0f9e01](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/f0f9e01bc78fe4bd9f001e1290c0a7ab78829c4e).

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: I fixed in [e97bd80](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/e97bd800ddf16d23729192ce7b6afc9f6549a504).

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: If this if-check is not done, a KeyError will probably be raised if all the arguments are not specified.
```python
value = kwds[arg]
```

examples/GLTextItem.py
shikishima-TasakiLab: I fixed in [
b98244e](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/b98244e7d34ce368d6d49af76545daf2d4f0ade1).

examples/GLTextItem.py
shikishima-TasakiLab: I fixed in [
b98244e](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/b98244e7d34ce368d6d49af76545daf2d4f0ade1).

pyqtgraph/opengl/items/GLTextItem.py
j9ac9k: Ahh that makes sense 👍🏻

examples/GLTextItem.py
j9ac9k: 👍🏻

pyqtgraph/opengl/items/GLTextItem.py
j9ac9k: 👍🏻

pyqtgraph/opengl/items/GLTextItem.py
j9ac9k: oh, right, I misread!

j9ac9k:
Hi @shikishima-TasakiLab it's looking much nicer! 

1. A few other comments, I would suggest doing away with the type-hints.  We don't use these elsewhere in the library, we have no mechanism to test their correctness.  We'll likely have support for them in the future, but we're not ready to start thinking about how to roll those out yet.
2. I would modify the color stuff a bit, I'm going to put comments in-line for suggested changes.  

I think that will cover all the issues I have, some of the other maintainers may request some other changes.  Thanks so much for this feature!

pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -0,0 +1,90 @@
...
+    """Draws text in 3D."""
+
+    def __init__(self, **kwds):
+        """All keyword arguments are passed to setData()"""
+        GLGraphicsItem.__init__(self)
+        glopts = kwds.pop('glOptions', 'additive')
+        self.setGLOptions(glopts)
+
+        self.pos:Union[np.ndarray, List[float], Tuple[float, ...]] = np.array([0.0, 0.0, 0.0])
+        self.color:Union[np.ndarray, List[float], Tuple[float, ...]] = np.array([1.0, 1.0, 1.0, 1.0])
```
j9ac9k, line 17: We have some mechanisms already in-place to support robust color generation:

```python

...
    def __init__(self, **kwargs):
        ....
        self.color = (1.0, 1.0, 1.0, 1.0)  # use a tuple
        ....

    def setData(self, **kwds):
        ....
        if arg == "color":
            self.color = glColor(value)
        ....
```

The reason we want to use `glColor` here is that we have a _very_ wide-ranging supported color input validation in `pyqtgraph.functions.mkColor` and have a layer that wraps that, `pyqtgraph.functions.glColor` around that.

pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -0,0 +1,90 @@
...
+                            '"GLUT_BITMAP_HELVETICA_18".')
+                setattr(self, arg, value)
+        self.update()
+    
+    def paint(self):
+        if len(self.text) < 1:
+            return
+        self.setupGLState()
+
+        glColor4d(float(self.color[0]), float(self.color[1]), float(self.color[2]), float(self.color[3]))
```
j9ac9k, line 87: with the above change, we should be able to replace this with

```python
glColor4d(*self.color)
```

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: > ```python
> if arg == "color":
>     self.color = glColor(value)
> ```
It's not appropriate because `glColor()` returns None.

In the `__init__` method, `color` is now initialized as a tuple.
The `setData` method has been left unchanged. ([97e39cd](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/97e39cde831c710998d9cf8b02064144a1986cf9))

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: I fixed in [97e39cd](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/97e39cde831c710998d9cf8b02064144a1986cf9).

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: I fixed in [2f26a72](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/2f26a720647e082cb4638cd267ab55f639165b70).

ixjlyons:
Looks reasonable to me. My only suggestion is to add something else to the example. A `GLAxisItem` would be fine, just to make it more clear how the text works in the 3D space. Or you could add a `GLTextItem` or two to an existing example instead of making a new one.

shikishima-TasakiLab:
@ixjlyons 

> Looks reasonable to me. My only suggestion is to add something else to the example. A `GLAxisItem` would be fine, just to make it more clear how the text works in the 3D space. Or you could add a `GLTextItem` or two to an existing example instead of making a new one.

I added GLGridItem and GLAxisItem to the example.

GLTextItem_example2
GLTextItem_example3

pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -0,0 +1,92 @@
...
+                value = kwds[arg]
+                if arg == 'pos':
+                    if isinstance(value, np.ndarray):
+                        if value.shape != (3,):
+                            raise Exception('"pos.shape" must be (3,).')
+                    elif isinstance(value, (tuple, list)):
+                        if len(value) != 3:
+                            raise Exception('"len(pos)" must be 3.')
+                elif arg == 'color':
+                    if isinstance(value, np.ndarray):
```
j9ac9k, line 56: We can probably get rid of all these checks, and let any exception that come up during `fn.mkColor` handle issues here.  For example, this would fail on the case of `color  = (127, 127, 127)` but `fn.glColor` would return a correct result for that use-case.

j9ac9k:
Hi @shikishima-TasakiLab Sorry for the delay, a few of us maintainers had some non-code related issues that occupied our time the last few days.

Back to this PR.  I had one more issue, I think we can get rid of all the checks for attempting to check that the color input was correct.  `fn.glColor` (and by extension `fn.mkColor`) should handle that for us; the restrictions here are too strict; and will reject what should be valid inputs (instead of loosening the checks a bit, I recommend doing away w/ them entirely).

After that, I think this PR is ready for merging unless anyone else has any objections.

pijyoi:
To get it to run on Windows platform, I had to build freeglut and then rename and copy the resulting DLL into a sub directory of the pyopengl installation.

j9ac9k:
huh; I'll give this PR look on my windows box, thanks for checking @pijyoi wonder why that is...

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: I fixed in [389bce9](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/389bce973f6dfaa44707ba402a6f63ac2b5b62b0).

shikishima-TasakiLab:
@j9ac9k 
> Hi @shikishima-TasakiLab Sorry for the delay, a few of us maintainers had some non-code related issues that occupied our time the last few days.
> 
> Back to this PR. I had one more issue, I think we can get rid of all the checks for attempting to check that the color input was correct. `fn.glColor` (and by extension `fn.mkColor`) should handle that for us; the restrictions here are too strict; and will reject what should be valid inputs (instead of loosening the checks a bit, I recommend doing away w/ them entirely).
> 
> After that, I think this PR is ready for merging unless anyone else has any objections.

Right, all the necessary checks are done in `fn.glColor`. I removed the if-check for the "color" argument.

j9ac9k:
I'm going to try and replicate the issue that @pijyoi came across; and if I have the same issue, I'll see if I can fix that up.

j9ac9k:
this indeed does not work on my windows/python3.9/pyside6 setup

```
(.venv-pyside6-py39) PS C:\Users\ogi\source\repos\j9ac9k\pyqtgraph> python .\examples\GLTextItem.py
Using PySide6
  File "C:\Users\ogi\source\repos\j9ac9k\pyqtgraph\examples\GLTextItem.py", line 27, in <module>
    txtitem1 = gl.GLTextItem(pos=(0.0, 0.0, 0.0), text='text1')
  File "C:\Users\ogi\source\repos\j9ac9k\pyqtgraph\pyqtgraph\opengl\items\GLTextItem.py", line 23, in __init__
    glutInit()
  File "C:\Users\ogi\source\repos\j9ac9k\pyqtgraph\.venv-pyside6-py39\lib\site-packages\OpenGL\GLUT\special.py", line 333, in glutInit
    _base_glutInit( ctypes.byref(count), holder )
  File "C:\Users\ogi\source\repos\j9ac9k\pyqtgraph\.venv-pyside6-py39\lib\site-packages\OpenGL\platform\baseplatform.py", line 423, in __call__
    raise error.NullFunctionError(
OpenGL.error.NullFunctionError: Attempt to call an undefined function glutInit, check for bool(glutInit) before calling
```

shikishima-TasakiLab:
@j9ac9k 
> this indeed does not work on my windows/python3.9/pyside6 setup
> 
> ```
> (.venv-pyside6-py39) PS C:\Users\ogi\source\repos\j9ac9k\pyqtgraph> python .\examples\GLTextItem.py
> Using PySide6
>   File "C:\Users\ogi\source\repos\j9ac9k\pyqtgraph\examples\GLTextItem.py", line 27, in <module>
>     txtitem1 = gl.GLTextItem(pos=(0.0, 0.0, 0.0), text='text1')
>   File "C:\Users\ogi\source\repos\j9ac9k\pyqtgraph\pyqtgraph\opengl\items\GLTextItem.py", line 23, in __init__
>     glutInit()
>   File "C:\Users\ogi\source\repos\j9ac9k\pyqtgraph\.venv-pyside6-py39\lib\site-packages\OpenGL\GLUT\special.py", line 333, in glutInit
>     _base_glutInit( ctypes.byref(count), holder )
>   File "C:\Users\ogi\source\repos\j9ac9k\pyqtgraph\.venv-pyside6-py39\lib\site-packages\OpenGL\platform\baseplatform.py", line 423, in __call__
>     raise error.NullFunctionError(
> OpenGL.error.NullFunctionError: Attempt to call an undefined function glutInit, check for bool(glutInit) before calling
> ```

This is exactly the error caused by not having GLUT installed.

> @j9ac9k
> I added an example in [7c2c13c](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/7c2c13ca7eccdf583ef06042e99e17c198a84bb1).
> 
> GLTextItem
> 
> If you get the following error, you may need to install "freeglut3".
> 
> ```
> OpenGL.error.NullFunctionError: Attempt to call an undefined function glutInit, check for bool(glutInit) before calling
> ```

j9ac9k:
🤦🏻‍♂️  I should read more closely... an extra dependency is a bit of a tall order; I'm going to need other maintainers to chime in @outofculture @ixjlyons @NilsNemitz ?

shikishima-TasakiLab:
I found a way to draw text without using GLUT!
I'm working on implementing it now.

NilsNemitz:
@shikishima-TasakiLab , thanks for all the work! 

Having text available in the 3d plots seems like a must-have for any serious application of these plots.

Fundamentally, FreeGLUT as a dependency is very lightweight, I think it takes only a few hundred kB. 
As far as I can tell, it does not bring in any further dependencies, and it is MIT licensed.

The potential roadblock is that I could not find a pip package. (There is an Anaconda package.)
We might not want to start manually handling binary dependencies over this...

So having a way to draw text without GLUT would be the nicest solution!

shikishima-TasakiLab:
Referring to the following website, I implemented it using QPainter, and was able to draw text without using GLUT.

[c++ - How to render text with QOpenGLWidget - Stack Overflow](https://stackoverflow.com/questions/28216001/how-to-render-text-with-qopenglwidget/33674071)

I applied this change in [6e5f09b](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/6e5f09b3712038cd9aaa16bf84a958c684f0589d).

pijyoi:
There appears to be some rendering issues on Windows, perhaps related to #1760. Attached are screenshots from Windows and Linux, both taken right after program launch.

Linux
gltextitem_linux

Windows. "text1" is missing and "text2" is at a different position.
gltextitem_windows

shikishima-TasakiLab:
> There appears to be some rendering issues on Windows when using QPainter, perhaps related to #1760. Attached are screenshots from Windows and Linux, both taken right after program launch.

I have also tested it in a Windows environment. However, the error you encountered did not occur.

GLTextIItem_PySide6

- Environment
  - OS : Windows 10 Education 20H2 19042.985
  - Python : 3.7.9
  - PySide6 : 6.1.0
  - numpy : 1.20.3
  - PyOpenGL : 3.1.5

j9ac9k:
Sometimes I wonder if I should take a long break from this library and study openGL a bit; it's such a mystery to me 😆 

After the kid are in bed tonight, I'll try and test this out on the various platforms.  I now have access to windows/macOS/linux development environments so I'll put this PR through the paces there and see if I can isolate some conditions for which it does not work.

pijyoi:
I too, was using PySide6 6.1.0.

I found that the expected output appears if I disable hidpi scaling through the following environment variable:
```
QT_ENABLE_HIGHDPI_SCALING = 0
```
So it's likely to be a hidpi issue and not a Windows-only issue.
The device pixel ratio may need to be taken into account when computing the screen coordinates.

shikishima-TasakiLab:
> I too, was using PySide6 6.1.0.
> 
> I found that the expected output appears if I disable hidpi scaling through the following environment variable:
> 
> ```
> QT_ENABLE_HIGHDPI_SCALING = 0
> ```
> 
> So it's likely to be a hidpi issue and not a Windows-only issue.
> The device pixel ratio may need to be taken into account when computing the screen coordinates.

I found that the error occurs when the display scale is not 100%.

pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -78,6 +78,8 @@ def paint(self):
 
         text_pos = self.__project(self.pos, modelview, projection, viewport)
         text_pos[1] = viewport[3] - text_pos[1]
+
+        text_pos /= self.view().devicePixelRatio()
```
shikishima-TasakiLab, line 82: I solved the error by dividing the coordinates by the pixel ratio of the device !

pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -0,0 +1,107 @@
...
+        for k in kwds.keys():
+            if k not in args:
+                raise Exception('Invalid keyword argument: %s (allowed arguments are %s)' % (k, str(args)))
+        for arg in args:
+            if arg in kwds:
+                value = kwds[arg]
+                if arg == 'pos':
+                    if isinstance(value, np.ndarray):
+                        if value.shape != (3,):
+                            raise Exception('"pos.shape" must be (3,).')
```
j9ac9k, line 46: Can you change these from bare exceptions to `ValueError` exceptions?  (same further down...

pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -0,0 +1,107 @@
...
+                    elif isinstance(value, (tuple, list)):
+                        if len(value) != 3:
+                            raise Exception('"len(pos)" must be 3.')
+                elif arg == 'color':
```
j9ac9k, line 50: I think you can get rid of all your logic here, and just say:

```python
from pyqtgraph import functions as fn
...


value = fn.mkColor(value)
```

We have a lot of logic here to support a _wide_ variety of inputs, and we raise the appropriate exceptions; end result should be what you want.

pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -0,0 +1,107 @@
...
+                        value_len = len(value)
+                        if value_len == 3:
+                            value = QtGui.QColor(value[0], value[1], value[2])
+                        elif value_len == 4:
+                            value = QtGui.QColor(value[0], value[1], value[2], value[3])
+                        else:
+                            raise Exception('"len(color)" must be 3 or 4.')
+                elif arg == 'font':
+                    if isinstance(value, QtGui.QFont) is False:
+                        raise Exception('"font" must be QFont.')
```
j9ac9k, line 66: Please raise `TypeError` here instead 👍🏻

pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -0,0 +1,107 @@
...
+            return
+        
+        proj_vec[0:3] /= proj_vec[3]
+
+        return np.array([
+            viewport[0] + (1.0 + proj_vec[0]) * viewport[2] / 2.0,
+            viewport[1] + (1.0 + proj_vec[1]) * viewport[3] / 2.0,
+            (1.0 + proj_vec[2]) / 2.0
+        ])
+
```
j9ac9k, line 107: Looks like github is complaining about no newline at the end of the file, can you verify that?

j9ac9k:
Hi @shikishima-TasakiLab 

Sorry I let this hang for a while.  I had my hands full of two other PRs (one of which is absolutely massive).  I should have reviewed and given feedback earlier, my appologies.

I tried this out, it works great on macOS.  I'll give it another run on my windows laptop with a hidpi monitor in the next day or so.

I added some comments to help clean things up a bit...mostly nit-picky stuff.

Also, can you add your example to this file here, that way it is available in the example app, and it is run as part of our test suite.

https://github.com/pyqtgraph/pyqtgraph/blob/master/examples/utils.py#L63-L71

Thanks again for this PR!

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: I fixed in [8946076](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/8946076752840caf734c783e1d05c330702620c6).

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: I fixed in [8946076](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/8946076752840caf734c783e1d05c330702620c6).

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: I fixed in [8946076](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/8946076752840caf734c783e1d05c330702620c6).

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: There were four ' ' in the end of the file, so I removed them. ([8946076](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/8946076752840caf734c783e1d05c330702620c6))

pyqtgraph/opengl/items/GLTextItem.py
```diff
@@ -0,0 +1,95 @@
+from OpenGL.GL import *
+import numpy as np
+from pyqtgraph.Qt import QtCore, QtGui
+from pyqtgraph.opengl.GLGraphicsItem import GLGraphicsItem
+import pyqtgraph.functions as fn
+
```
j9ac9k, line 6: To silence the static code analyzer, can you insert in here:

```python
__all__ = ['GLTextItem']
```

j9ac9k:
Hi @shikishima-TasakiLab Thanks for updating the PR, I'll break out my hidpi machine and do some testing there.  I just pointed out one issue the static code-checker flagged.  

One the `__all__ =` bit is added, and  assuming this looks fine on the HiDPI machine with different display scaling, I think this is ready to merge.  

Thanks again for creating this and for figuring out how to do this without an extra dependency!

pyqtgraph/opengl/items/GLTextItem.py
shikishima-TasakiLab: I fixed in [e2aa1ef](https://github.com/pyqtgraph/pyqtgraph/pull/1776/commits/e2aa1ef733db223d0b28d68fa5f3568283250021).

j9ac9k:
This LGTM, as soon as the CI finishes, I will merge.

Thanks for sticking with this @shikishima-TasakiLab !

Issue #1756, opened 2021-05-02
parameterTree parameter with "dec" option on crashes around zero


llombard:
### Short description
when stepping up or down around zero in a dec spinbox, the step should be some default "minStep", otherwise a fraction of zero is still zero. However, it crashes.

### Code to reproduce
start pyqtgraph "ParameterTree" example
change "DEC stepping" to 0 and step up or down

### Expected behavior
the value should step up or down by a minStep value

### Real behavior
crashes

```
Traceback (most recent call last):
  File "C:\apps\Anaconda3\lib\site-packages\pyqtgraph\widgets\SpinBox.py", line 403, in stepBy
    step = self.opts['minStep']
KeyError: 'minStep'
```

### Tested environment(s)

 * PyQtGraph version: 0.11.0
 * Qt Python binding: 'PyQt5 5.9.2 Qt 5.9.7'
 * Python version: Python 3.8.5 (default, Sep  3 2020, 21:29:08) [MSC v.1916 64 bit (AMD64)]
 * NumPy version: 1.19.2
 * Operating system: windows
 * Installation method: conda

ixjlyons:
Confirmed on master. There's a note about requiring 'minStep' for dec mode in the `SpinBox` docs, but I think it should set a reasonable default as a fallback. Plus `WidgetParameterItem` doesn't forward that option to `SpinBox` for float type parameters.

So, I'd say `SpinBox` needs to set a default 'minStep' if it's not provided (I guess 'step' would be a reasonable default) and `WidgetParameterItem` needs to handle `minStep` for float the float case.

ixjlyons:
Fixed by #1779

PR #1779, opened 2021-05-16
Use fallback minStep in dec mode SpinBox (Issue 1756)

manschloemark:
Hello, I made some changes addressing issue #1756

If a SpinBox is in 'dec' mode but there is no 'minStep' value it will set 'minStep' to the same value as 'step'. This fixes the crash mentioned in the issue.

I also made a change to examples/parametertree.py that sets 'minStep' to 1.0e-12 in the DEC stepping spinbox. I did this because while the change mentioned above prevented a crash, it set the minimum step to 1.0, which meant the spinbox could not step from 1 Hz to 0.9 mHz, it would go to 0 Hz instead. Setting 'minStep' to a very small value keeps the behavior similar to how it was before the fix.


j9ac9k:
Hi @manschloemark thanks for the PR!

This LGTM, but the `SpinBox` stuff is more of @ixjlyons 's wheelhouse, so I'll let him address/merge this PR.

Thanks again!

ixjlyons:
Looks good to me, thanks!

PR #1786, opened 2021-05-19
Bypass makeARGB

outofculture:
This continues #1693 but with corrected cupy behavior.

@pijyoi yeah, the lookup table substrate was the problem.

With that, this branch now provides across-the-board improvements to cupy-enabled benchmarks!


pijyoi:
>Huh. I'm just now noticing that I've been completely neglecting RawImageWidget, which uses makeARGB.

Well, RawImageWidget and RawImageGLWidget are not made any slower by this PR...

outofculture:
Agreed!

I was playing with refactors, but for now, this behaves well and looks good. Let's merge it!

pijyoi:
Found an issue when using ImageView. To trigger it, run the below script and change the colormap by right-clicking on the HistogramLUTItem. It seems to have to do with HistogramLUTItem returning a Callable for self.lut.
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ImageItem.py#L413
```python
import pyqtgraph as pg
import cupy as xp

pg.setConfigOption('useCupy', True)

shape = (1024, 1024)
data_uint8 = xp.random.randint(256, size=shape, dtype=xp.uint8)
data_uint16 = xp.random.randint(4096, size=shape, dtype=xp.uint16)
data_float32 = xp.random.normal(size=shape)

pg.mkQApp()

imv = pg.ImageView()
imv.show()
imv.setImage(data_float32)

pg.exec()
```

Issue #1784, opened 2021-05-19
fix for PYSIDE-1563 could break makeQImage()


pijyoi:
http://bugreports.qt.io/browse/PYSIDE-1563
https://codereview.qt-project.org/c/pyside/pyside-setup/+/348876
https://github.com/qtproject/pyside-pyside-setup/commit/cc011c8980cc4ce02e3d9cceb74ee028c204214e

Certain parts of pyqtgraph use QImage to wrap existing memory buffers and expect that future changes to those buffers will affect the QImage.
1) https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ImageItem.py#L222
2) https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/widgets/RemoteGraphicsView.py#L248

The fix referred to above seems to make a private copy of the input buffer. Thus future changes to the input buffer would not be seen by the QImage.

Furthermore, for usages where no further change to the memory buffer is required, the fix also means that ```QImage(buffer)``` is no longer zero-copy.
e.g. https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/widgets/RemoteGraphicsView.py#L83

j9ac9k:
@pijyoi I can bring this issue to the attention of the pyside2 devs, but I don't think I can speak to the specifics (and appropriate fixes) as well as you could.  Would you be ok commenting on their bug tracker, or pinging one of the devs in their [gitter chat](https://gitter.im/PySide/pyside2)?  

@outofculture is this something you might be up for doing either?

I've found the pyside2 devs to be very responsive and receptive of input when you can point to specific issues.

pijyoi:
A useful post on how QImage is intended to be used: https://www.qt.io/blog/2017/01/20/inside-qimage-touch-qt-quick-opengl. 

The 2 cases that I listed above (ImageItem.py and RemoteGraphicsView.py) fall under the blog's "Case 3: Wrapping and modifying external image data".

[1] https://github.com/qtproject/pyside-pyside-setup/commit/cc011c8980cc4ce02e3d9cceb74ee028c204214e

We could convert ImageItem.py to "Case 2: Wrapping external, read-only image data" and RemoteGraphicsView.py to "Case 1: Image data owned by the QImage" and this would avoid breakage when [1] arrives. (This makes RemoteGraphicsView.py make 1 extra copy even in the absence of [1])

[...]

PR #1795, opened 2021-05-22
don't rely on QImage-wrapped buffer being writable

pijyoi:
This works around #1784 for ImageItem.py.
It converts a potential PySide6 6.1.1 breakage to a performance loss (that would be limited to PySide6).

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1423,10 +1423,6 @@ def ndarray_to_qimage(arr, fmt):
     h, w = arr.shape[:2]
     bytesPerLine = arr.strides[0]
     qimg = QtGui.QImage(img_ptr, w, h, bytesPerLine, fmt)
-
-    # Note that the bindings that support ndarray directly already hold a reference
-    # to it. The manual reference below is only needed for those bindings that take
-    # in a raw pointer.
     qimg.data = arr
     return qimg
 
--- a/pyqtgraph/graphicsItems/ImageItem.py
+++ b/pyqtgraph/graphicsItems/ImageItem.py
@@ -235,7 +235,7 @@ class ImageItem(GraphicsObject):
             self._processingBuffer = self._xp.empty(shape[:2] + (4,), dtype=self._xp.ubyte)
         else:
             self._processingBuffer = self._displayBuffer
-        self.qimage = fn.makeQImage(self._displayBuffer, transpose=False, copy=False)
+        self.qimage = None
 
     def setImage(self, image=None, autoLevels=None, **kargs):
         """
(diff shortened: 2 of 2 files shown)


j9ac9k:
This might not be necessary: https://codereview.qt-project.org/c/pyside/pyside-setup/+/350435

pijyoi:
The implementation in the link adds a new bool copy argument that defaults to ```True```, i.e. the new breaking behavior is the default. This means that ```ndarray_to_qimage()``` would need to be modified to feature test for availability of this new argument, and if available, call it with ```False```. Until ```ndarray_to_qimage()``` is modified, PySide6 6.1.1 would be broken for pyqtgraph.

In comparison, this PR does a safe pre-emptive change. It makes PySide6 6.1.1 be _only_ a performance loss until ```ndarray_to_qimage()``` gets modified. Even though QImage() is now called for each image update, that's fine because the call to QImage to wrap an existing buffer is fast.

j9ac9k:
Ahh, I didn't read your post/evaluate the code clearly enough; this code change should not result in a performance loss across the board (while you were quite explicit that that was not the case 🤦🏻‍♂️ ).

Either way, this looks like it should be merged, but as with most image item related matters, I'll defer to @outofculture 👍🏻

j9ac9k:
Looks like they're backtracking 🎊 

https://codereview.qt-project.org/c/pyside/pyside-setup/+/350854

pijyoi:
It looks like the comment (made by me) about how QImage holds a reference to the ndarray was only true for PyQt5. 
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L1427-L1429

j9ac9k:
> It looks like the comment (made by me) about how QImage holds a reference to the ndarray was only true for PyQt5.

Do you want to update the comment to reflect that?

Also since these kind of implementations seem to change on us, wonder if it would be worthwhile to add some check per binding kind that checks to see if a copy is made/not-made so that we can catch changes in the upstream implementations earlier on; but that's clearly outside the scope of this PR.

pijyoi:
>>It looks like the comment (made by me) about how QImage holds a reference to the ndarray was only true for PyQt5.

>Do you want to update the comment to reflect that?

Looking back at #724, the statement would have been true for PySide2 < 5.12. I had assumed wrongly that when the leak was fixed in PySide2 >= 5.12, the fixed behavior was to release the reference upon destruction of the QImage; instead no reference is held at all.

It would be best to remove the comment altogether, rather than update it.

j9ac9k:
> ... PySide2 < 5.12` 

So many issues with those bindings that I kept running into while trying to maintain the CI suite.

Would you be up for removing it?  I don't mind doing it, I just like to get a thumbs-up before I go making changes on other peoples branches.

pijyoi:
> Would you be up for removing it? I don't mind doing it, I just like to get a thumbs-up before I go making changes on other peoples branches.

Please go ahead.

PR #1797, opened 2021-05-23
Color map linearization

NilsNemitz:
While working on #1625 , I repeatedly ran into the problem of defining colors that are perceptually well separated.

This became more manageable by sampling colors from the included perceptually uniform color maps. However, the included maps are not suitable for the purpose of e.g. defining warm and cool gray scales.

This PR adds 
- a color ramp generator that can create such ramps with approximately equal perceptual distance between equally spaced points. This generator takes a hue and saturation to generate not only slightly tinted gray scales, but also monochrome color scales that imitate old computer screens.
- a `linearize()` method for `ColorMap` that adjusts the positions of the provided color stops according to the perceptual distance of the assigned colors, approximately linearizing the overall map in terms of perceived rate of change
- a `colorDistance()` function that calculates the perceived distances between a set of QColors according to CIE76. This seems adequate for the current application, but we might later add the more accurate metrics based on the CIELab color space.
- a `CIELabColor()` function that generates a QColor from a set of CIE L* a* b* values
- a `colorCIELab()` function that calculates CIE L* a* b* values for a QColor.

It also:
- fixes a bug that prevented `ColorMap.__init__()` from accepting a list of QColors.
- moves the color-related examples into a `Colors` folder in the Example app.
- cleans up the code of the previous color map example a little
- adds a new example for the linearization and color ramp generation
- adds some tests for conversion to and reconversion from CIELab; unfortunately I currently have no reference to confirm exact L* a* b* values. 

This is basically part of #1625, but it seemed reasonable to factor this out as a separate PR. It shouldn't break any existing applications.

linearized color maps


pyqtgraph/functions.py
```diff
@@ -372,6 +372,161 @@ def hsvColor(hue, sat=1.0, val=1.0, alpha=1.0):
...
+    
+    Parameters
+    ----------
+    L: float
+        Lightness value ranging from 0 to 100
+    a, b: float
+        (green/red) and (blue/yellow) coordinates.
+    alpha: float, optional
+        Opacity, ranging from 0 to 1
+    """
```
j9ac9k, line 402: would you mind adding a [Notes](https://numpydoc.readthedocs.io/en/latest/format.html#notes) section to this docstring referencing where this "standard" or calculation comes from?

pyqtgraph/tests/test_functions.py
```diff
@@ -245,3 +245,18 @@ def test_siParse(s, suffix, expected):
...
+        qcol.setRgbF( *np.random.random((3)) )
+        color_list.append(qcol)
+    
+    for qcol1 in color_list:
+        vec_Lab  = pg.functions.colorCIELab( qcol1 )
+        qcol2 = pg.functions.CIELabColor(*vec_Lab)
+        for val1, val2 in zip( qcol1.getRgb(), qcol2.getRgb() ):
+            if abs(val1-val2) > 1:
+                print(f'Excess CIELab reconversion error ({qcol1.name() } > {vec_Lab } > {qcol2.name()})')
+                assert False
```
j9ac9k, line 262: file needs to end with a newline

pyqtgraph/tests/test_functions.py
```diff
@@ -245,3 +245,18 @@ def test_siParse(s, suffix, expected):
...
+        qcol = pg.Qt.QtGui.QColor()
+        qcol.setRgbF( *np.random.random((3)) )
+        color_list.append(qcol)
+    
+    for qcol1 in color_list:
+        vec_Lab  = pg.functions.colorCIELab( qcol1 )
+        qcol2 = pg.functions.CIELabColor(*vec_Lab)
+        for val1, val2 in zip( qcol1.getRgb(), qcol2.getRgb() ):
+            if abs(val1-val2) > 1:
+                print(f'Excess CIELab reconversion error ({qcol1.name() } > {vec_Lab } > {qcol2.name()})')
```
j9ac9k, line 261: move this print-statement into the assert error message

```python
assert False, f"Excess CIELab ....."
```

This way the test suite won't capture that output, but it will be displayed on the test summary.

pyqtgraph/functions.py
```diff
@@ -372,6 +372,161 @@ def hsvColor(hue, sat=1.0, val=1.0, alpha=1.0):
...
+    for idx, val in enumerate(vec_XYZ):
+        if val > 0.20689: # 6.9456e-7:
+            vec_XYZ[idx] = vec_XYZ[idx]**3
+        else:
+            vec_XYZ[idx] = (vec_XYZ[idx] - 16/116) / 7.787
+    vec_XYZ = VECTOR_XYZn * vec_XYZ # apply white reference
+    # print(f'XYZ: {vec_XYZ}')
+
+    # convert XYZ to linear RGB
+    vec_RGB =  MATRIX_RGB_FROM_XYZ @ vec_XYZ
```
j9ac9k, line 417: MATRIX MULTIPLICATION OPERATOR HAS ENTERED THE LIBRARY!!!! 

👍🏻

j9ac9k:
Hi @NilsNemitz 

Thanks for this PR.  I just found some nit-picky stuff I brought up.  Most of my issues is with referencing where these calculations come from in the doc-string....

Anyway once the merge-conflict is resolved, and the extra commented out code is removed, I think this will be ready for merging.

pyqtgraph/functions.py
NilsNemitz: It's most excellent that the matrix multiplication syntax in Python (since 3.5) is right up there with the competing numerical languages 😄

pyqtgraph/tests/test_functions.py
NilsNemitz: Today I learn... Thanks, added.

pyqtgraph/tests/test_functions.py
NilsNemitz: Fixed itself while tidying up merge conflicts.

pyqtgraph/functions.py
NilsNemitz: Done. I also realized that the docstring needed manual import into the functions section of the docs.

NilsNemitz:
Dear @j9ac9k , Thanks for looking this over!

I have updated the PR to include your suggestions.

examples/utils.py
```diff
@@ -30,6 +28,11 @@
         ('Verlet chain', 'verlet_chain_demo.py'),
         ('Koch Fractal', 'fractal.py'),
     ])),
+    ('Colors', OrderedDict([
```
j9ac9k, line 31: Just wanted to comment to 👍🏻 the new section.

pyqtgraph/colormap.py
```diff
@@ -195,9 +198,84 @@ def getFromColorcet(name):
...
+            raise ValueError(f"Invalid color descriptor '{color}'")
+    l_vals = np.linspace(l_min, l_max, num=10)
+    color_list = []
+    for l_val in l_vals:
+        qcol = QtGui.QColor()
+        qcol.setHslF( h_val, s_val, l_val )
+        color_list.append( qcol )
+    return ColorMap( None, color_list, name=name, linearize=True )
+
+def testBarData(length=768, width=32):
```
j9ac9k, line 267: I had to double check, I thought that pytest might mistake this function as a "test" function, but the naming convention doesn't _quite_ match; function names need to be `test_*` (having an underscore).  We can leave this as-is, but if you can think of another function name that might be better; I would encourage a replacement.

pyqtgraph/colormap.py
```diff
@@ -255,24 +332,35 @@ def __init__(self, pos, color, mapping=CLIP, mode=None): #, names=None):
...
+            self.pos = np.linspace(0.0, 1.0, num=len(color))
+        else:
+            self.pos = np.array(pos)
+            order = np.argsort(self.pos)
+            self.pos = self.pos[order]
+        
+        self.color = np.zeros( (len(color), 4) ) # stores float rgba values
+        for cnt, idx in enumerate(order):
+            self.color[cnt] = mkColor(color[idx]).getRgbF()
+        # alternative code may be more efficinet, but fails to handle lists of QColor.
```
j9ac9k, line 352: totally good with this kind of code being left in-place but commented out 👍🏻

pyqtgraph/functions.py
```diff
@@ -372,6 +372,198 @@ def hsvColor(hue, sat=1.0, val=1.0, alpha=1.0):
...
+    Parameters
+    ----------
+        L: float
+            Lightness value ranging from 0 to 100
+        a, b: float
+            (green/red) and (blue/yellow) coordinates, typically -127 to +127.
+        alpha: float, optional
+            Opacity, ranging from 0 to 1
+
+    Notes
```
j9ac9k, line 403: 🙌🏻  thank you so much for including this!

j9ac9k:
Hi @NilsNemitz I just went over this again; love the changed.  

The static code checker flagged a few issues (unused variable, import not used, that sort of thing).  I think once those are addressed, we can merge this guy.

Also thank you so much for the detailed `Notes` section, I was curious where all that math came from.

pyqtgraph/colormap.py
NilsNemitz: The previous code was quite elegant, but it silently assumed a 2-dimensional list of values.
It seems like it should be possible to adapt it to handle a 1-dimensional list of ready-made colors, too, but I just couldn't figure it out. Maybe someone else wants to have a go; so I would propose to leave this in for reference.

pyqtgraph/colormap.py
NilsNemitz: --> renamed to `colormap.modulatedBarData()`

examples/utils.py
NilsNemitz: Great, I was getting worried that all the color things are overwhelming the other examples.

NilsNemitz:
Dear @j9ac9k , thanks for the comments! I have looked over the output of the code checker and I hope I have made it happy now.
`testBarData` is `modulatedBarData` now to distinguish it from test code.

j9ac9k:
hmm...there is a test failure on macOS; I'll try and look into it:

```
Failed Color Maps Example Test Located in colorMaps.py
----------------------------- Captured stdout call -----------------------------
Color Maps

WARNING:matplotlib.font_manager:Matplotlib is building the font cache; this may take a moment.
INFO:matplotlib.font_manager:Failed to extract font properties from /System/Library/Fonts/Apple Color Emoji.ttc: In FT2Font: Could not set the fontsize (error code 0x17)
INFO:matplotlib.font_manager:Failed to extract font properties from /System/Library/Fonts/LastResort.otf: tuple indices must be integers or slices, not str
INFO:matplotlib.font_manager:Failed to extract font properties from /System/Library/Fonts/Supplemental/NISC18030.ttf: In FT2Font: Could not set the fontsize (error code 0x17)
INFO:matplotlib.font_manager:generated new fontManager
```

examples/test_examples.py
```diff
@@ -59,6 +59,14 @@ def buildFileList(examples, files=None):
 
 darwin_opengl_reason = ("pyopenGL cannot find openGL library on big sur: "
                         "https://github.com/python/cpython/pull/21241")
+       
+darwin_pyside611_mpl_broken = (
+    platform.system() == "Darwin" and
+    Qt.VERSION_INFO.startswith("PySide6 6.1.1 ")
+)
+darwin_pyside611_mpl_reason = (
+    "Matplotlib causes segfaults with PySide 6.1.1 on OSX even without drawing operations."
```
pijyoi, line 68: Note that it is not a segfault, just a informational log containing the string "error"

pijyoi:
>This appears to be a PySide6 6.1.1/matplotlib specific issue, and not an issue specific to this PR.
I'm going to advise we skip this example for this test config, the example does work, we just get that warning/error from matplotlib

Maybe #1821 could address this

NilsNemitz:
And #1821 does indeed seem to fix this issue.
Awesome detective work @pijyoi !

pyqtgraph/functions.py
```diff
@@ -372,13 +372,172 @@ def hsvColor(hue, sat=1.0, val=1.0, alpha=1.0):
...
+def colorCIELab(qcol):
+    """
+    Describes a QColor by an array of CIE L*a*b* values.
+    Also see :func:`CIELabColor() <pyqtgraph.CIELabColor>` .
+
+    Parameters
+    ----------
+    qcol: QColor
+        QColor to be converted
+    alpha: Boolean, optional
```
j9ac9k, line 471: this function does not appear to take the optional alpha parameter.

NilsNemitz: It turned out that it isn't actually very helpful to have a `(L a b alpha)` tuple, and where needed, alpha can be taken from the `qcol` input directly.
Good that you spotted this, I have removed that bit of the docstring.

j9ac9k:
Hi @NilsNemitz 

I'm sorry to keep bringing up small things, I promise I want to merge this 😅 

On looking things over, I see we have function name `colorCIELab` and we have `CIELabColor`  I get one function generates a QColor object from the `L a b` values, and the other describes a QColor in `L a b` values.

I think the functions should be renamed to make it a bit more obvious what is the case.  Scanning through the Qt API has not been terribly helpful since all QColor from/to are methods to the QColor class (and there are few plain functions in Qt, it's almost all methods, which have different namespace).  

I'll think a bit more on what other appropriate names would be for those functions; but I think `fromCIELabToQColor` and `fromQColorToCIELab` ... but I'm not convinced those are much better either.

j9ac9k:
after a chat w/ @NilsNemitz I'm good w/ the function naming convention.  It was brought to my attention that all our color conversion functions use the name naming scheme, it would be silly to change this.

PR #1792, opened 2021-05-21
Fix the wrong conversion of QVector3D to numpy array in GLViewWidget.py

3DAlgoLab:
Fix the wrong conversion of QVector3D to numpy array.

PR #1799, opened 2021-05-23
Force cameraPosition() to return Vector class. 

3DAlgoLab:
In quaternion mode, GLViewWidget has problem with QVector3D conversion in ```def pixelSize```, #1792 

I think Vector class minus(-) operator has to return Vector class not Qvector3D, but as it has no minus operator, it returns QVector3D. That's the origin of the problem. 

But as a quick way to fix it, force to convert it to Vector in ```cameraPosition() ``` suggested by [pijyoi](https://github.com/pijyoi).


j9ac9k:
Hi @3DAlgoLab sorry it took me a while to follow up, some of the other PRs I've been working on have been taking all my time.  This LGTM, merging.

Issue #1783, opened 2021-05-18
GLMeshItem memory leak in drawing of faces


wubella:
### Short description
When rendering faces, there appears to be a memory leak relating to line 199 and 200 of GLMeshItem.py

### Code to reproduce
This can be reproduced using the GLMeshItem example script. Using top to check memory usage, the amount used by the process slowly increases as updates are called in the scene (triggered by rotating the view, etc.). If there are no changes the memory usage is constant, but otherwise will grow without bound (albeit slowly in the example). Larger meshes with cause faster accumulation.

### Expected behavior
Near-constant memory usage when rendering the faces in the scene.

### Real behavior
Increasing memory usage when rendering faces.

### Tested environment(s)

 * PyQtGraph version: tested on both 0.11.0 and 0.12.1
 * Qt Python binding:  5.15.4
 * Python version: 3.8
 * NumPy version: 1.17.3
 * Operating system: Ubuntu 20.04
 * Installation method: pip

### Additional context
changing astype on faces from np.uint to np.uint32 or np.uintc on line 199 of pyqtgraph/opengl/items/GLMeshItem.py resolves the issue for me, but I don't know if that is the correct solution.

pijyoi:
I wasn't able to observe any long running increase in memory usage neither in Windows 10 64-bits nor in Ubuntu 20.04 64-bits.

Regardless, np.uint is 32-bits on Windows 64-bits (LLP64) but is 64-bits on Linux 64-bits (LP64). Which means that on Linux platforms, PyOpenGL would need to further convert it to 32-bits in order to make it usable as GL_UNSIGNED_INT (32-bits).

wubella:
To see the increase, repainting the scene is required. For GLMeshItem example script, I constantly orbited the camera to force repaints, but the long running increase is quite slow for the example. For my application with large mesh and constant updates, it's much more noticeable. This was on Ubuntu 20.04 64-bits.

Regardless, it sounds like pyqtgraph is expecting glDrawElements to do the appropriate conversion on their end, and this is an issue with PyOpenGl, not pyqtgraph. Is that accurate?

pijyoi:
>To see the increase, repainting the scene is required. For GLMeshItem example script, I constantly orbited the camera to force repaints, but the long running increase is quite slow for the example.

[...]

PR #1800, opened 2021-05-24
Fixing GLMeshItem memory leak in face drawing on 64-bit Linux (#1783)

wubella:
PyOpenGl requires 32-bit uint as corresponding argument to GL_UNSIGNED_INT. On 64-bit Linux, np.uint translates to a 64-bit uint, so changing the astype to explicitly be np.uint32 as required.


j9ac9k:
Thanks for this PR @wubella !

Verified the openGL requirement via google search here: https://www.khronos.org/opengl/wiki/OpenGL_Type

LGTM, Merging.

PR #1802, opened 2021-05-24
update cx_freeze example and add a workaround for templates

fbordignon:
Updated the example to a working state with cx_Freeze 6.2

Added a workaround to copy the unreferenced template files to the packaged app.


j9ac9k:
Thanks for the PR @fbordignon !

I was considering removing the freeze app content because I was not sure it was actually being used.  If you have suggestions for changes that can be made in the library that could make the integration into freeze apps easier, I'd certainly love to hear them.

PR #1804, opened 2021-05-26
fix click bug on removable ROIs; test

outofculture:
Looks like `in` stopped working at some point, but this `int` version should be safer and work everywhere.

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -801,7 +801,7 @@ class ROI(GraphicsObject):
         if ev.button() == QtCore.Qt.RightButton and self.contextMenuEnabled():
             self.raiseContextMenu(ev)
             ev.accept()
-        elif ev.button() in self.acceptedMouseButtons():
+        elif int(ev.button()) & int(self.acceptedMouseButtons()) > 0:
             ev.accept()
             self.sigClicked.emit(self, ev)
         else:
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -801,7 +801,7 @@ class ROI(GraphicsObject):
         if ev.button() == QtCore.Qt.RightButton and self.contextMenuEnabled():
             self.raiseContextMenu(ev)
             ev.accept()
-        elif int(ev.button()) & int(self.acceptedMouseButtons()) > 0:
+        elif ev.button() & self.acceptedMouseButtons() > 0:
             ev.accept()
             self.sigClicked.emit(self, ev)
         else:
--- a/pyqtgraph/graphicsItems/tests/test_ROI.py
+++ b/pyqtgraph/graphicsItems/tests/test_ROI.py
@@ -153,6 +153,24 @@ def check_getArrayRegion(roi, name, testResize=True, transpose=False):
 
     win.hide()
 
+
+def test_mouseClickEvent():
+    plt = pg.GraphicsView()
+    plt.show()
+    resizeWindow(plt, 200, 200)
+    vb = pg.ViewBox()
+    plt.scene().addItem(vb)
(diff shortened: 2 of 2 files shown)


j9ac9k:
Thanks for the fix @outofculture that's too bad the `in` operator doesn't work, it feels so right for this use case.

Elsewhere in ROI.py we use `ev.button() & self.acceptedMouseButtons()` the `int` version I'm pretty sure was broken on PyQt6 6.0.0, but we don't support that version due to their non-integer alternative to enums.

Should we cahnge the other usage of `self.acceptedMouseButtons()` or should we do away with the integer casting (or just use two different methods)?

outofculture:
Oh, I hadn't seen it used elsewhere. `int` casting isn't necessary, so I'm fine skipping that. Well, fine except that now my branch name isn't clever, but 🤷🏽

j9ac9k:
> Well, fine except that now my branch name isn't clever, but 🤷🏽

we can't get everything we want in life....

pijyoi:
```python
       elif ev.button() & self.acceptedMouseButtons() > 0:
```
0fa4557ad6237f9d86b671921f23a2a92449d719 is the commit that added PyQt6 compatibility wrt to enums and flags.
No testing against ```> 0```
I recall that you can't generally test against an integer value, It might end up being False just because you are comparing two different types.
It would be a good idea to physically test out the change with a PyQt6 venv.

pijyoi:
Ran the ImageView example in PyQt 6.1 and did a mouse click within the ROI
```python
[19:54:39]  Error sending click event:

    |==============================>>
    |    File "C:\work\github\pyqtgraph\examples\ImageView.py", line 67, in <module>
    |      pg.exec()
    |    File "C:\work\github\pyqtgraph\pyqtgraph\Qt.py", line 447, in exec_
    |      return app.exec() if hasattr(app, 'exec') else app.exec_()
    |    File "C:\work\github\pyqtgraph\pyqtgraph\widgets\GraphicsView.py", line 355, in mouseReleaseEvent
    |      super().mouseReleaseEvent(ev)
    |    File "C:\work\github\pyqtgraph\pyqtgraph\GraphicsScene\GraphicsScene.py", line 240, in mouseReleaseEvent
    |      if self.sendClickEvent(cev[0]):
    |    File "C:\work\github\pyqtgraph\pyqtgraph\GraphicsScene\GraphicsScene.py", line 389, in sendClickEvent
    |      debug.printExc("Error sending click event:")
    |    --- exception caught here ---
    |    File "C:\work\github\pyqtgraph\pyqtgraph\GraphicsScene\GraphicsScene.py", line 387, in sendClickEvent
    |      item.mouseClickEvent(ev)
    |    File "C:\work\github\pyqtgraph\pyqtgraph\graphicsItems\ROI.py", line 804, in mouseClickEvent
    |      elif ev.button() & self.acceptedMouseButtons() > 0:
    |  TypeError: '>' not supported between instances of 'MouseButtons' and 'int'
    |==============================<<
```

j9ac9k:
@pijyoi I'm going to fix this in #1807 thanks for catching it.

j9ac9k:
I swear, qt bindings are never happy (PyQt5 5.12 + Python 3.8)

```
================================================================================= warnings summary ==================================================================================
tests/graphicsItems/test_ROI.py::test_PolyLineROI
  /Users/ogi/Developer/pyqtgraph/pyqtgraph/graphicsItems/ROI.py:804: DeprecationWarning: an integer is required (got type MouseButtons).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
    elif ev.button() & self.acceptedMouseButtons():

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

I think I may just suppress this warning.

j9ac9k:
oh, it did catch it, but it was suppressed; ... that's a problem; you can see the exception when you run `pytest -s`

PR #1806, opened 2021-05-27
Revert Point.angle behavior change

campagnola:
Commit a534132c6 introduced a behavior change in `Point.angle()`. This causes problems downstream that are difficult to debug since the change occurs silently.

### DIFF
--- a/examples/fractal.py
+++ b/examples/fractal.py
@@ -56,7 +56,7 @@ def update():
         p2 = pts[i+1]
         v2 = p2 - p1
         t = p1 - pts[0]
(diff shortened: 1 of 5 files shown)


j9ac9k:
I knew taking the time for making half-decent commits was going to pay off someday!

Sorry for the head-ache, this LGTM, I don't agree with the naming; and maybe we'll deprecate someday and replace with `angleFrom` and `angleTo`, but that's a discussion for another time.

j9ac9k (commented):
LGTM

PR #1807, opened 2021-05-28
Merge test data + make tests directory which contain tests

### LINKED ISSUE #1025: Tests do not fail on warnings
### Short description

The pyqtgraph library should not emit warnings when used as intended. The current tests do not reflect this.

### Additional context

For issue #835, a test was suggested to check if no warning is printed. Since this specific NumPy warning was tested to be fixed with #1024, tests are mainly interesting for every other part of the code and for future updates of third party libraries. Therefore, tests in general should fail on warnings.

### LINKED ISSUE #1064: PyQt5 test images missing
Since commit c58a177561829920fadfdcc1fbee2160f44126b3 the image comparison is massively relaxed for PyQt5 because the test images do not work for PyQt5. That commit says "we will nee[d] new test images to cover". This seems not to have happened so far, and in my tests, the deviation between test and new images are in the order of hundreds of pixels. So this is an issue that is left to be done.

j9ac9k:
This PR aims to get pyqtgraph in-line with standard python package directory structures, by creating a `tests` directory on the root level, and putting all tests in there, including the images from [test-data](https://github.com/pyqtgraph/test-data) repository.  The best example of this is in the "Good Integration Practices" on the [pytest documentation page](https://docs.pytest.org/en/6.2.x/goodpractices.html#tests-outside-application-code).

Instead of just copying the images, I generated new images with Qt6, which the test suite passed with out of the box for the most part, even after resetting the pxCount to 0 for Qt5 and Qt6.  On windows, some of the ROI tests had a few pixels off, but setting pxCount to 2 there seemed to address the issue (which is much smaller than the 10 it was before for Qt5).

With `setup.py` we can ensure that test data is _not_ bundled with the pip package, and we can get more reliable use of `conftest.py` files (which we make minimal use of presently).

While doing the huge shuffle of the test suite, a number of issues presented themselves (including a call to `QApplication.instance().deleteLater()`).  As I addressed them; I realized that the test suite no longer emits warnings; so I addressed #1025.  Turns out pytest runs into an unraisable exception with how we call our examples, which this was addressed.

I took the opportunity to fix some of the issues the static code checker had flagged (non-secure temporary files), and I modified `debug.printExc` so it doesn't just print an exception (while suppressing it), but it actually raises a Runtime Warning.  This is to prevent the case of #1804 where the exception was swallowed by this mechanism, despite there being a test to catch that very thing.

~Unfortunately, on my local machine, I'm triggering segfaults with `test_ImageItem.py`, `test_NonUniformImage.py` and `test_InfiniteLine.py`.  Until these segfaults are resolved, I'm leaving this PR in draft form.~

Fixes #1025 
Fixes #1064


j9ac9k:
For the segfaults, initially I was suspicious of `qWaitForWindowExposed` vs. `qWaitForWindowActive` but after changing those, while the segfaults appeared to happen less regularly; they still do occur.

Furthermore, they do not happen when running any single test by itself.

```
tests/graphicsItems/test_ArrowItem.py .                                                                                                                                                                  [ 74%]
tests/graphicsItems/test_AxisItem.py ......                                                                                                                                                              [ 78%]
tests/graphicsItems/test_ErrorBarItem.py .                                                                                                                                                               [ 79%]
tests/graphicsItems/test_GraphicsItem.py ..                                                                                                                                                              [ 80%]
tests/graphicsItems/test_ImageItem.py sszsh: segmentation fault  pytest -x -k "not examples"

~/Developer/pyqtgraph merge-test-data* 10s
pyqtgraph-pyside2_515-py38 ❯ pytest -k "test_ImageItem"
============================================================================================= test session starts ==============================================================================================
platform darwin -- Python 3.8.6, pytest-6.2.1, py-1.10.0, pluggy-0.13.1
rootdir: /Users/ogi/Developer/pyqtgraph, configfile: pytest.ini
plugins: cov-2.11.1, xdist-2.2.0, forked-1.3.0
collected 247 items / 239 deselected / 8 selected

tests/graphicsItems/test_ImageItem.py ss...                                                                                                                                                              [ 62%]
tests/graphicsItems/test_ImageItemFormat.py ...                                                                                                                                                          [100%]

================================================================================= 6 passed, 2 skipped, 239 deselected in 2.88s =================================================================================

~/Developer/pyqtgraph merge-test-data*
pyqtgraph-pyside2_515-py38 ❯
```

j9ac9k:
found a bunch more tests, migrated them over as well, ...segfaults occur in 

* `test_ImageItem.py`
* `test_ViewBox.py`
* `test_InfiniteLine.py`
* `test_NonUniformImage.py`

this command runs without any errors! 😆 

```
pytest tests -k "not test_ImageItem and not test_InfiniteLine and not test_NonUniformImage and not test_ViewBox"
```

j9ac9k:
I can no longer reproduce a segfault locally, but two segfaults in the CI is enough to make me suspiscious.

j9ac9k:
Tests are migrated over, static code quality checker is fine, "silenced" exceptions have been addressed.   I think the next step is to regenerate images with Qt5 (or Qt6), and tighten the tolerance for assertImageApproved for Qt5 and Qt6.

NilsNemitz:
DEar @j9ac9k , thanks for the massive overhaul.
With the revised tests, you might want to check if 
#1025 and #1064 can be closed!

j9ac9k:
Good news regarding the pixel tolerance, it does not appear to be needed when the images are generated using Qt6 bindings.  I've only done the PlotCurveItem images, I'm going to do the other images shortly (I'm setup to do one at a time, so it's going to take a bit!).

j9ac9k:
ok, I made CI angry; ...will investigate more later.  With the new images, `tox` passed on macOS across the board with the new images

j9ac9k:
passes on macOS, segafaults on linux, crashes on windows, "works on my machine!"

j9ac9k:
image

So this is the hold up on windows, there appears to be 1 blue and 1 yellow pixel that is the difference between what was shown, and what was expected.  Will tinker with scaling settings....

pijyoi:
If you want to refactor into the function getImageFromWidget(), then  imageToArray() needs to be called with copy=True, or manually make a copy of the resultant ndarray.

j9ac9k:
> If you want to refactor into the function getImageFromWidget(), then imageToArray() needs to be called with copy=True, or manually make a copy of the resultant ndarray.

I'll keep that in mind, but I reverted that change (due to the segfaults on ubuntu).... I may look to refactor it out, as I think it can provide useful functionality on its own; but right now I'm debating how to pass the assertImageAppoved tests, without having to provide a fudge-factor for Windows systems, which at a glance looks like what I will have to do.

j9ac9k:
Oh the `test_getArrayRegion` tests, only half of the tests within there fail, specifically the

```python
        (pg.ROI([1, 1], [27, 28], pen='y'), 'baseroi'),
        (pg.RectROI([1, 1], [27, 28], pen='y'), 'rectroi'),
```

The other ones pass

```python
        (pg.EllipseROI([1, 1], [27, 28], pen='y'), 'ellipseroi'),
        (pr, 'polylineroi'),
```

j9ac9k:
Just noticed the following comment in the code:

```python
# on windows, one edge of one ROI handle is shifted slightly; letting this slide with pxCount=10
```

This was applied to one of the tests failing; but not to all the ones failing...

j9ac9k:
I'm doing some last minute cleanup on this PR, I'm going to let it sit for a day, and merge, so we can get back to #1796 and ensure the image tests pass that way.

j9ac9k:
and now I screwed up my git history here 🤦🏻

j9ac9k:
I figured out how to run the examples without generating warnings (therefore I can enable error on warning for the test as suite); unfortunately on tox, on (some) qt6 environments I'm getting connection refused errors for the remote view example.  The don't get the errors when I running that one example by itself.

Am still investigating...

j9ac9k:
I have no more planned changes, this guy is ready for review.

pijyoi:
While fixing endian-issues for s390 platform, it was found that the saved images had their R and B channels swapped:
https://github.com/pyqtgraph/pyqtgraph/issues/1667#issuecomment-810055769

The following patch tries to fix that: https://github.com/pijyoi/pyqtgraph/commit/7f602bfc4e5fe54ce7bac2350af7069c5cd4c4a9
It would however require regeneration of all the images.

This also fixes the endian issue properly. (The ImageTester uses ImageItem which requires that the data be RGBA)

j9ac9k:
> While fixing endian-issues for s390 platform, it was found that the saved images had their R and B channels swapped:
> 
> https://github.com/pyqtgraph/pyqtgraph/issues/1667#issuecomment-810055769
> 
> 
> 
> The following patch tries to fix that: https://github.com/pijyoi/pyqtgraph/commit/7f602bfc4e5fe54ce7bac2350af7069c5cd4c4a9
> 
> It would however require regeneration of all the images.
> 
> 
> 
> This also fixes the endian issue properly. (The ImageTester uses ImageItem which requires that the data be RGBA)

I'll apply the patch and regenerate the images.  Thank you so much!!

pijyoi:
There's a mistake in my makePng() translation
```python
    # qim = fn.makeQImage(img.transpose(1, 0, 2), alpha=False)
    # transposing the image is probably wrong, but we just do a direct
    # translation of the code above
    img = img.swapaxes(0, 1).copy()
    qim = fn.ndarray_to_qimage(img, QtGui.QImage.Format.Format_RGBX8888)
```

It should simply be
```python
    qim = fn.ndarray_to_qimage(img, QtGui.QImage.Format.Format_RGBX8888)
```

The default arguments to makeQImage are ```copy=True``` and ```transpose=True```.
So the ```img.transpose(1, 0, 2)``` had the effect of passing in ```transpose=False```

j9ac9k:
I haven't grabbed the patch yet; probably won't actually get to it for a few hours.

j9ac9k:
I'll start on this now; will update here when I finish (likelihood of distraction is high 😆 )

j9ac9k:
@pijyoi the static code analyzer doesn't like the change to `functions.py`

```python
    if depth in (8, 24, 32):
        dtype = np.uint8
        nchan = depth // 8
    elif depth in (16, 64):
        dtype = np.uint16
        nchan = depth // 16
    shape = qimg.height(), qimg.width()
    if nchan != 1:
        shape = shape + (nchan,)
    data = np.frombuffer(img_ptr, dtype=dtype).reshape(shape)
    return data
```

Specifically, it is complaining about `nchan` and `dtype` being potentially uninitialized.  Should I provide an else with a runtime error or something along those lines?

pijyoi:
Sure, please fix it with a ValueError, unsupported image type.

pijyoi:
```nonuniform_image/colormap-3x3.png``` looks very different. Probably due to some older PR breaking things w/o the test suite catching it.

master
colormap-3x3

this PR
colormap-3x3

j9ac9k:
Nice catch ! Will try looking into it in a few hours.

j9ac9k:
@pijyoi you are spot on:

image

I'll see if I can reuse the image that's still on `test-data`, if not, can easily regenerate.

j9ac9k:
Updated the PR , unless anyone spots anything funky (including my screwing up of my git history, which I've only done a dozen or so times on this branch), I think this should be ready to merge (of course pending the result of the static code checker).

PR #1809, opened 2021-05-30
Sanitize input for plotItem.showGrid (was documentation fix for plotItem.showGrid alpha setting)

### LINKED ISSUE #1484: Int conversion issues when showing plot grid
In '0.11.1.dev0' (and 0.11.1), the following code will issue an error:

```python
import pyqtgraph as pg
from warnings import simplefilter
simplefilter('error', DeprecationWarning)

pw = pg.PlotWindow()
pi = pw.plotItem
pi.showGrid(True, True, alpha=0.5)
#   File "C:\Users\njessurun\AppData\Local\Continuum\anaconda3\envs\py38\lib\site-# packages\pyqtgraph\graphicsItems\PlotItem\PlotItem.py", line 382, in showGrid
#     self.ctrl.gridAlphaSlider.setValue(v)
# TypeError: setValue(self, int): argument 1 has unexpected type 'numpy.float64'
```

Qt library: 'PyQt5 5.14.2 Qt 5.14.2'

NilsNemitz:
Looking through some issues, I came across #1484:

setting `plotItem.showGrid(True, True, alpha=0.5)` raises `TypeError: setValue(self, int): argument 1 has unexpected type 'numpy.float64'`

This is a documentation error:
The expected alpha value does not range from 0.0 to 1.0 (as in the current docstring), but instead ranges from 0-255.

This PR updates the docstring to show that and adds a cast to int to avoid raising an error.
The `alpha=0.5` value in the original report will no longer result in an error and draw an invisible, fully transparent grid. 
Setting `alpha=127` creates the desired half-transparent grid.

So this should close #1484.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -362,7 +362,7 @@ class PlotItem(GraphicsWidget):
         **Arguments:**
         x               (bool) Whether to show the X grid
         y               (bool) Whether to show the Y grid
-        alpha           (0-255) Opacity of the grid
+        alpha           (0.0-1.0) Opacity of the grid
         ==============  =====================================
         """
         if x is None and y is None and alpha is None:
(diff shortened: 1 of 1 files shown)


ixjlyons:
I don't think this is a documentation error actually. The slider goes from 0 to 255:

https://github.com/pyqtgraph/pyqtgraph/blob/e29f86578f120fda2281cfddb7429d03470fe29b/pyqtgraph/graphicsItems/PlotItem/plotConfigTemplate.ui#L298-L300

So using `alpha` as a [0, 1] scaler for the max of the slider is doing what's intended, it's just that `QSlider.setValue` takes an int type. `setValue(int(v))` should suffice.

NilsNemitz:
True. 

@ixjlyons , this update restores the 0-1 range and just adds the cast to int.

ixjlyons (approved):
Looks good to me. The comment there is helpful too :+1:

j9ac9k:
LGTM, merging.  Thanks @NilsNemitz !

PR #1812, opened 2021-05-31
Feature: Facilitates Parameter-Function Interaction

ntjess:
I've been excited for quite a while now about a PR for Parameters which infer their children from function definitions, since ~95% of my usage of parameters is to interact with a specific function. This allows for said interaction with multiple customization points:

* The way the function is run can be modified (i.e. by button, changed parameter, and/or changing parameter)
* Title formatting
* Ignorable / overridable modifications to default values
* 'Deferred' evaluation for inputs from another scope (i.e. an outer variable)

Please refer to the test `parametertree/tests/test_Parameter::test_interact` and example `examples/InteractiveParameter` for some usage.

See an 'in-the-wild` use in the modified `ScatterPlotSpeedTest` example.

I'm not happy about some of the API (i.e. clobbering the `Parameter` namespace), but hopefully this starts the conversation.


j9ac9k:
Hi @ntjess thanks for the PR.

I want to look this over but I have my hands full with 1807 and 1796 for the next little bit.  I'll tag @ixjlyons or @lidstrom83  to take a look if they're available; as both have more experience with parameter trees than I do.

If this PR still hasn't been reviewed in a few days I'll try and take a closer look.

ntjess:
Just to drop another note on why this beats the status quo IMO: there is the significant benefits of
* Using externally defined functions -- they don't have to access a `param` structure in the function body
* Much more reusability for functions defined this way -- they can be called programmatically with different arguments just like any other function, compared to having to modify the `param` structure and call the function with no arguments

README.md
```diff
@@ -103,6 +103,6 @@ Here is a partial listing of some of the applications that make use of PyQtGraph
 * [rapidtide](https://rapidtide.readthedocs.io/en/latest/)
 * [argos](https://github.com/titusjan/argos)
 * [PySpectra](http://hasyweb.desy.de/services/computing/Spock/node138.html)
-* [Semi-Supervised Semantic Annotator](https://gitlab.com/ficsresearch/s3ah)
+* [Semi-Supervised Semantic Annotator](https://gitlab.com/ficsresearch/s3a)
```
j9ac9k, line 106: ooooooooof

my bad....

j9ac9k:
Hi @ntjess unfortunately with the merge of #1807 I moved the test files around, so you have a merge conflict.  You'll need to merge master into your branch, and handle the merge conflict before this can be merged (going to try and take a closer look at this today).

j9ac9k:
Hi @ntjess We don't use a lot of decorators in pyqtgraph, but I think the use-case here is absolutely perfect. 

I would request you expand your variable names a bit more and instead of using shortened/abbreviated names of variables, to write them out (format instead of fmt, decorator instead of deco, that sort of thing).  This not only makes it easier to read, but it's easier for people new to the library to follow along.  Obviously there are some cases where pyqtgraph was already using shorthand for variable names, and modifying those use-cases is outside the scope of this PR.

ntjess:
@j9ac9k sounds good.

* What do you think is the best localization strategy for items like the title formatter, run behavior, etc.? I could make them function parameters instead, but in my use case they rarely change, which motivated me to externalize it into a user-settable default. In your opinion, does this clobber the general `Parameter` namespace with variables?
* I would also want to have context managers for the above settings, which would avoid the `try-finally` paradigm I used in the test cases -- something like `with Parameter.setInteractiveOpts(runFormat=..., titleFormat=...):` but again this is one more item in the Parameter namespace
* I have a `runFunc` parameter as described in the documentation, but I'm thinking it could be avoided if I just forced users in this scenario to pass a wrapped version of their function in instead, i.e.
  ```python
  from functools import wraps
  wrapped = wraps(docFunc)(runFunc)
  interact(wrapped)
  ```
I initially liked the extra parameter since the intentionality is clearer to those who don't make library or wrapped functions that often, but I think `wraps` isn't too bad. What are your thoughts?

ntjess:
Another option is moving all this functionality to a subclass of `GroupParameter` named `InteractiveParameter`, then there is reason to put lots of stuff into the namespace...

ntjess:
Fits best in `GroupParameter`, since every interactive parameter is contained in one anyway

ntjess:
@ixjlyons would love your feedback on this. I made a tentative addition that searches function documentation for user-defined options. This means `interact` will pull options from functions with documentation like:

```python
def a(x=5):
  """
  any text here

  Parameters
  -----------
  x: int
    Some description
    [x.options]
    step=5
    limits=[0, 10]
  """
```

Additionally, if `docstring_parser` is in the python path the `[x.options]` line is unnecessary (docstring_parser knows a wide variety of annotation style guides and can find where each parameter is defined, so it doesn't need to be tagged with a header). So, it would accept:

```python
def a(x=5):
  """
  any text here
  Parameters
  -----------
  # Or rst style
  x: int
    Some description
    # Spacing optional
    step=5
    limits=[0, 10]
  """
```

I wasn't sure if people would like that change (I'm all for it, but the PR is democratic :stuck_out_tongue: ), so I stuck it on another branch that can be merged: https://github.com/ntjess/pyqtgraph/tree/interactive-params-with-docparser.

At present, values can be set to anything that passes `ast.literal_eval`, which covers almost all initialization use cases I've needed experimentally.

You can check out example docstring usage in examples/paramtetertree/test_docparser.py

ntjess:
In the interest of lowering the barrier to understanding, I typed a much more detailed summary of these features here: https://pyqtgraph--1812.org.readthedocs.build/en/1812/parametertree/interactiveparameters.html

ntjess:
`interact` logic moved back to its own file thanks to param reorganization. Better options context, management of hooked up signals via `InteractiveFunction`. Relevant information should be updated as well in the ParameterTree RST docs

pyqtgraph/parametertree/interactive.py
```diff
@@ -0,0 +1,463 @@
...
+    name = chDict['name']
+    # Make sure args without defaults have overrides
+    if chDict['value'] is RunOpts.PARAM_UNSET and name not in toExec.deferred and name not in toExec.extra:
+        raise ValueError(f'Cannot interact with "{toExec} since it has required parameter "{name}"'
+                         f' with no default or deferred value provided.')
+    child = parent.addChild(chDict, existOk=existOk)
+    # I tried connecting directly to the runnables in `toExec`, but they result in early garbage collection. This
+    # doesn't happen with local functions
+    def runner(param, val):
+        toExec.runFromChangedOrChanging(param, val)
```
ntjess, line 445: @j9ac9k This prevents garbage collection of interactive functions created at line 393. Is there a better way of making sure the `runFrom...` function of `toExec` is called without an eternal dangling reference here?

j9ac9k: for something like that preventing garbage collection, the only other thing I can think of is having some kind of dictionary being kept in the global namespace that holds onto them.  I think this is a fair use of `global` which I know most people shy away from, but for something like this, I think it's fair.

ntjess: Would it be fair to assume InteractiveFunction should share the same life cycle as the initial function? If so, I could do this instead:

```python
func.interactiveRef = toExec
```

I think this would only fail for lambdas defined in the call to `interact`. In any case, I like it a lot better than an inaccessible reference inside a function call...

j9ac9k: from what I've gathered form this, I would assume this is a fair assumption.

ntjess:
Directly superseded by #2318

PR #1813, opened 2021-06-01
propose fix LineROI coordinate initialization

SergiiChernov:
In my case (pyqtgraph 0.11.0 from a few months ago) LineROI does some irrelevant coordinate conversion upon initialization. Please ignore if this was fixed already.

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -1659,18 +1659,32 @@ class LineROI(ROI):
     
     """
     def __init__(self, pos1, pos2, width, **args):
-        pos1 = Point(pos1)
-        pos2 = Point(pos2)
-        d = pos2-pos1
-        l = d.length()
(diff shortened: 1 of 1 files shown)


ixjlyons:
The geometry in `LineROI` does appear to be off actually, but it looks to me like it always has been. Are you describing the behavior between #1741 and #1806 getting merged? I don't think a release was made in between.

The proposed fix works, but `pos1.x()` and `pos2.x()` could be equal, so the way `Point.angle` works is preferred numerically. I'm not sure what you mean by irrelevant coordinate transformation -- what's here is roughly equivalent except the sign corrections.

With #1806 in, all we need  is to reverse the angle call and switch the sign on `c.x()`:

```diff
ra = d.angle(Point(1, 0), units="radians")
c = Point(width/2. * sin(ra), -width/2. * cos(ra))
```

j9ac9k:
Hi @SergiiChernov 

This PR got superseded by #1876   Thanks for bringing the issue to our attention and proposing a fix!

PR #1816, opened 2021-06-02
Fix #1814

### LINKED ISSUE #1814: Parameter value equality comparisons fails when values are numpy array-like
In a few places (namely `setLimits` and `valueIsDefault`), direct value comparisons with `==` fail on array-like comparisons. This can be fixed simply by using `pg.eq` instead of `==`. Parameter operations aren't speed-optimized, so there is limited benefit to avoiding rich comparison.

ntjess:
I think the two fixes are the only places where array values could be present. Most of the other compares are on values known to be strings, etc.

Fixes #1814


j9ac9k:
Hi @ntjess thanks for the PR.  I updated your post so github knows to close #1814 when this gets merged.  This LGTM, merging.

PR #1818, opened 2021-06-06
Convert Qt Enums to Qt6 Namespace

j9ac9k:
While doing benchmarking for #1796 I noticed on PyQt6 bindings, a _lot_ of time was being spent shuffling the Enum namespace during the initialization while loading Qt.py.  The benchmark I used was to update the line plot 500 times.  43% of the runtime was spent in `Qt.py`, of which promote_enums consumed 38% of the overall runtime.  Attached is the output of of the call-graph.

It was brought to my attention by @The-Compiler that the Qt6 Enum namespace has been present in Qt5 from at least Qt 5.12; and after doing some testing, sure enough it was; so decided a migration was in order.

The fantastic [PyQtEnumConverter](https://gitlab.com/JakobDev/PyQtEnumConverter) library handled the bulk of the work.  It missed a lot of cases of `self.EnumValue`, but luckily the test suite was more than happy to error out along the way.   As I patched known issues in the test suite, I used my editors "find all" functionality to find other usages of those particular enums to try and convert cases that the test suite had no coverage of.  I think I got them all at this point.

Lastly, I took the opportunity to remove Qt 6.0 support from the README and some of the specific shims we had in place for it.  I have a hard time imagining what a use-case for someone needing Qt 6.0 support when 6.1 is released.

pyqt6-plotlinespeedtest

Sorry for the huge diff, I don't mean to give everyone head-aches with merge-conflicts.  

tagging @pijyoi due to their work in making the enum namespace equivalent; figure you may have some input

Looking through the diff before I hit submit, I see one of my changes in MultiPlotSpeedTest.py got caught up.  I don't mind restoring it if it would make people feel better.


pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -99,14 +99,14 @@ def setCompositionMode(self, mode):
...
                                                       is opaque. Otherwise, it uses the alpha channel to blend
                                                       the image with the background.
-        QtGui.QPainter.CompositionMode_Overlay        The image color is mixed with the background color to
+        QtGui.QPainter.CompositionMode.CompositionMode_Overlay        The image color is mixed with the background color to
                                                       reflect the lightness or darkness of the background.
-        QtGui.QPainter.CompositionMode_Plus           Both the alpha and color of the image and background pixels
+        QtGui.QPainter.CompositionMode.CompositionMode_Plus           Both the alpha and color of the image and background pixels
                                                       are added together.
-        QtGui.QPainter.CompositionMode_Multiply       The output is the image color multiplied by the background.
+        QtGui.QPainter.CompositionMode.CompositionMode_Multiply       The output is the image color multiplied by the background.
```
NilsNemitz, line 109: These seem to break the docstring alignment.

pijyoi:
PyQtEnumConverter should not be run on the generated *Template_{binding}.py files.

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -247,7 +247,7 @@ def generateShape(self):
         return dti.map(tr.map(self._path))
 
     def mouseDragEvent(self, ev):
-        if not self.movable or int(ev.button() & QtCore.Qt.LeftButton) == 0:
+        if not self.movable or int(ev.button() & QtCore.Qt.MouseButton.LeftButton) == 0:
```
pijyoi, line 250: Casting to ```int``` likely doesn't work in PyQt6.

j9ac9k:
> PyQtEnumConverter should not be run on the generated *Template_{binding}.py files.

yuuuup... just realized that 🤦🏻

pyqtgraph/graphicsItems/PlotCurveItem.py
j9ac9k: ooooooooof... yea I'm going to have to take a closer look at those :|

pijyoi:
You might as well remove the following chunk from Qt.py
```python
if QT_LIB == PYQT6:
    # shim the old names for QPointF mouse coords
    QtGui.QSinglePointEvent.localPos = lambda o : o.position()
    QtGui.QSinglePointEvent.windowPos = lambda o : o.scenePosition()
    QtGui.QSinglePointEvent.screenPos = lambda o : o.globalPosition()

    QtWidgets.QApplication.exec_ = QtWidgets.QApplication.exec
```
Since PySide6 6.1.0, use of localPos, windowPos, screenPos, exec_ generates a deprecation warning. The codebase has since been changed to use the following style: ```ev.position() if hasattr(ev, 'position') else ev.localPos()```

pijyoi:
>I have a hard time imagining what a use-case for someone needing Qt 6.0 support when 6.1 is released.

Qt6 <= 6.0.3 had a soft memory leak for every queued signal emitted. This makes it unsuitable for long running applications. Since PyQt6 didn't release a 6.0.4, it means that to get the fix for this bug, you need PyQt6 >= 6.1.0. That should be good enough a reason to move to 6.1.

This memory leak is actually quickly visible in the various SpeedTest examples where a QTimer is being fired as fast as possible.

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -459,7 +459,7 @@ def mouseClickEvent(self, ev):
 
     def mouseDragEvent(self, ev):
         targetItem = self.parentItem()
-        if not targetItem.movable or int(ev.button() & QtCore.Qt.LeftButton) == 0:
+        if not targetItem.movable or int(ev.button() & QtCore.Qt.MouseButton.LeftButton) == 0:
```
pijyoi, line 462: Another cast to int that doesn't work in PyQt6

j9ac9k:
The intermittent segfault I'm seeing on pyside2/python3.7 has me concerned.  

I think at this point everything else should be good.

pijyoi:
DateAxisItem_QtDesigner.ui and designerExample.ui are loaded during runtime, so don't generate files for them.

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: Nice catch!

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: Nice catch, fixed.

j9ac9k:
I'm going to call it a night; will try some more trial and error regarding the segfault tomorrow.

Thanks for looking this over @pijyoi @NilsNemitz

pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -1105,33 +1105,33 @@ def generateDrawSpecs(self, p):
...
                 elif self.orientation == 'top':
-                    alignFlags = QtCore.Qt.AlignHCenter|QtCore.Qt.AlignBottom
+                    alignFlags = QtCore.Qt.AlignmentFlag.AlignHCenter|QtCore.Qt.AlignmentFlag.AlignBottom
                     rect = QtCore.QRectF(x-width/2., tickStop-offset-height, width, height)
                 elif self.orientation == 'bottom':
-                    alignFlags = QtCore.Qt.AlignHCenter|QtCore.Qt.AlignTop
+                    alignFlags = QtCore.Qt.AlignmentFlag.AlignHCenter|QtCore.Qt.AlignmentFlag.AlignTop
                     rect = QtCore.QRectF(x-width/2., tickStop+offset, width, height)
 
                 if QT_LIB == 'PyQt6':
```
pijyoi, line 1128: This if-else chunk can now be simplified to just ```textFlags = alignFlags | QtCore.Qt.TextFlag.TextDontClip```

pyqtgraph/parametertree/ParameterItem.py
```diff
@@ -166,9 +166,9 @@ def titleChanged(self):
         if QT_LIB == 'PyQt6':
```
pijyoi, line 166: This if-else chunk should now be able to be simplified to ```textFlags = QtCore.Qt.TextFlag.TextSingleLine```

pyqtgraph/widgets/GradientWidget.py
```diff
@@ -38,25 +38,25 @@ def __init__(self, parent=None, orientation='bottom',  *args, **kargs):
         self.item.sigGradientChangeFinished.connect(self.sigGradientChangeFinished)
         self.setCentralItem(self.item)
         self.setOrientation(orientation)
-        self.setCacheMode(self.CacheNone)
-        self.setRenderHints(QtGui.QPainter.Antialiasing | QtGui.QPainter.TextAntialiasing)
+        self.setCacheMode(self.CacheModeFlag.CacheNone)
+        self.setRenderHints(QtGui.QPainter.RenderHint.Antialiasing | QtGui.QPainter.RenderHint.TextAntialiasing)
 
         if QT_LIB == 'PyQt6':
```
pijyoi, line 44: This if-else chunk can now be simplified to
```frame_style = QtWidgets.QFrame.Shape.NoFrame | QtWidgets.QFrame.Shadow.Plain```

**Comment by j9ac9k** [comment]
my git game is weak; I keep running into this silly conflict, eventually I'll figure out how to resolve it...

**Comment by j9ac9k** [comment]
Doing a search for `PyQt6` there may be some other places to clean up the code, like RemoteGraphicsView.py, perhaps ROI.py,  and maybe `ndarray_to_qimage` in `functions.py`

Ok, calling it a night this time for sure.

File: examples/VideoTemplate_pyside2.py [hunk 7]
```diff
@@ -78,7 +78,7 @@ def setupUi(self, MainWindow):
         self.gridLayout_4.setObjectName(u"gridLayout_4")
         self.rawImg = RawImageWidget(self.page_2)
         self.rawImg.setObjectName(u"rawImg")
-        sizePolicy = QSizePolicy(QSizePolicy.Expanding, QSizePolicy.Preferred)
```
pijyoi, line 81: VideoTemplate_pyside2.py needs to be reverted.

examples/exampleLoaderTemplate_pyside2.py
```diff
@@ -17,7 +17,7 @@ def setupUi(self, Form):
         self.gridLayout_2 = QtWidgets.QGridLayout(Form)
         self.gridLayout_2.setObjectName("gridLayout_2")
         self.splitter = QtWidgets.QSplitter(Form)
```
pijyoi, line 19: exampleLoaderTemplate_pyside2.py needs to be reverted.

j9ac9k:
Still seeing the intermittent (but common enough) segfault on PySide2 5.12 and Python 3.7 on macOS during the test for scatter plot item.  I've committed reverts to some changes for debugging purposes.

j9ac9k:
The segfault is avoided by skipping `test_ROI`.  I have some other matters to attend to now, but will revisit this later.

I can see this being some kind of resource constraint in the CI system, which is likely why I cannot replicate locally.

j9ac9k:
Well, an identical segfault occurred in an unrelated PR and given this PR hasn't merged, I'm now confident in saying this segfault is not the result of this PR.

https://github.com/pyqtgraph/pyqtgraph/pull/1726/checks?check_run_id=2764478711

pijyoi:
>Doing a search for PyQt6 there may be some other places to clean up the code, like RemoteGraphicsView.py, perhaps ROI.py, and maybe ndarray_to_qimage in functions.py

Yes. ROI.py has special handling for KeyboardModifiers for PyQt6 6.0.x

RemoteGraphicsView.py and ndarray_to_qimage have special handling for QImage for PyQt6 6.0.0

j9ac9k:
Sorry for the huge diff, let's hope the merge conflicts aren't too bad!

pijyoi:
@j9ac9k , I think there are 3 instances of  ```ItemIsFocusable```  in https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/GraphicsScene/GraphicsScene.py that may need to be changed to ```GraphicsItemFlag.ItemIsFocusable```

j9ac9k:
> @j9ac9k , I think there are 3 instances of `ItemIsFocusable` in https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/GraphicsScene/GraphicsScene.py that may need to be changed to `GraphicsItemFlag.ItemIsFocusable`

you're right... have some work to do tonight will try and get to this in the next day or so.

PR #1819, opened 2021-06-06
add function to disable autoscaling

bsobhani:
HistogramLUTItem has a function to enable autorange but no simple function to disable. Adding the simple function to disable.


j9ac9k:
Hi @bsobhani 

Thanks for the PR, this LGTM.  Merging!

PR #1824, opened 2021-06-09
remove little-endian assumption for image export

pijyoi:
Remove little-endian assumption for image saving in:
1) ImageExporter.py
2) GLViewWidget.py

Note that for ImageExporter.py, we leave the image format as ARGB32 instead of switching to RGBA8888 in order to not have any unknown/unintended changes in rendering. (It is known, for instance, from #1738, that using RGB32 enables subpixel text font rendering)

### DIFF
--- a/pyqtgraph/exporters/Exporter.py
+++ b/pyqtgraph/exporters/Exporter.py
@@ -128,8 +128,8 @@ class Exporter(object):
         while len(childs) > 0:
             ch = childs.pop(0)
             tree = self.getPaintItems(ch)
-            if (ch.flags() & ch.ItemStacksBehindParent) or \
-               (ch.zValue() < 0 and (ch.flags() & ch.ItemNegativeZStacksBehindParent)):
+            if (ch.flags() & ch.GraphicsItemFlag.ItemStacksBehindParent) or \
+               (ch.zValue() < 0 and (ch.flags() & ch.GraphicsItemFlag.ItemNegativeZStacksBehindParent)):
                 preItems.extend(tree)
             else:
                 postItems.extend(tree)
--- a/pyqtgraph/exporters/ImageExporter.py
+++ b/pyqtgraph/exporters/ImageExporter.py
@@ -1,8 +1,9 @@
 from .Exporter import Exporter
 from ..parametertree import Parameter
-from ..Qt import QtGui, QtCore, QtSvg, QT_LIB
+from ..Qt import QtCore, QtGui, QtWidgets
 from .. import functions as fn
 import numpy as np
+import sys
 
 translate = QtCore.QCoreApplication.translate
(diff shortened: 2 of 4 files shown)


pyqtgraph/exporters/ImageExporter.py
```diff
@@ -73,15 +74,8 @@ def export(self, fileName=None, toBytes=False, copy=False):
...
-        bg = np.empty((h, w, 4), dtype=np.ubyte)
-        color = self.params['background']
-        bg[:,:,0] = color.blue()
-        bg[:,:,1] = color.green()
-        bg[:,:,2] = color.red()
-        bg[:,:,3] = color.alpha()
-
-        self.png = fn.makeQImage(bg, alpha=True, copy=False, transpose=False)
-        self.bg = bg
+        self.png = QtGui.QImage(w, h, QtGui.QImage.Format.Format_ARGB32)
```
j9ac9k, line 77: thanks for looking out for merge conflicts 😆  with my PR

j9ac9k:
I'll give @outofculture some time to review/approve, this LGTM.  I must say I love the diff, so much easier to read.

Also thanks for the point about the rendering; suppose that's a different PR to see if we can make that be an option on the export, as I imagine as having sub-pixel font rendering would be beneficial.

outofculture:
Nice work. Can we add a test to `tests/exporters/test_image.py`? Something like:

```python
def test_ImageExporter_toBytes():
    p = pg.plot()
    exp = ImageExporter(p.getPlotItem())
    img = exp.export(toBytes=True)
    data = qimage_to_ndarray(img)
    if sys.byteorder == 'little':
        black = (0, 0, 0, 255)
    else:
        black = (255, 0, 0, 0)
    assert np.count_nonzero(np.all(data != black, axis=-1)) == 0, "Exported image should be entirely black."
```

I'm not sure about that last assertion, but I can't think of anything stronger.

outofculture (approved):

pyqtgraph/exporters/Exporter.py
```diff
@@ -128,8 +128,8 @@ def getPaintItems(self, root=None):
         while len(childs) > 0:
             ch = childs.pop(0)
             tree = self.getPaintItems(ch)
-            if (ch.flags() & ch.ItemStacksBehindParent) or \
-               (ch.zValue() < 0 and (ch.flags() & ch.ItemNegativeZStacksBehindParent)):
+            if (ch.flags() & ch.GraphicsItemFlag.ItemStacksBehindParent) or \
+               (ch.zValue() < 0 and (ch.flags() & ch.GraphicsItemFlag.ItemNegativeZStacksBehindParent)):
```
j9ac9k, line 132: Nice catch!  I thought I got them all...

PR #1827, opened 2021-06-09
Fix BusyCursor to use internal stack provided by setOverrideCursor/restoreOverrideCursor

outofculture:
`restoreOverrideCursor` previously cleared all cursors, so the old code had a bug where nested busy cursors would never clear themselves.

I don't actually know for sure that all our supported versions of Qt have the new behavior, so I'm going to keep this as draft until the CI tells me it's safe to use this new code.

### DIFF
--- a/pyqtgraph/widgets/BusyCursor.py
+++ b/pyqtgraph/widgets/BusyCursor.py
@@ -8,12 +8,19 @@ __all__ = ["BusyCursor"]
 
 @contextmanager
 def BusyCursor():
+    """
+    Display a busy mouse cursor during long operations.
+    Usage::
+        with BusyCursor():
+            doLongOperation()
+    May be nested. If called from a non-gui thread, then the cursor will not be affected.
+    """
     app = QtCore.QCoreApplication.instance()
-    active = (app is not None) and (QtCore.QThread.currentThread() == app.thread())
+    in_gui_thread = (app is not None) and (QtCore.QThread.currentThread() == app.thread())
     try:
-        if active:
+        if in_gui_thread:
             QtGui.QApplication.setOverrideCursor(QtGui.QCursor(QtCore.Qt.CursorShape.WaitCursor))
         yield
     finally:
-        if active:
+        if in_gui_thread:
             QtGui.QApplication.restoreOverrideCursor()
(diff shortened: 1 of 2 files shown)


j9ac9k:
I had no idea this was even a thing in the library...

pyqtgraph/widgets/BusyCursor.py
```diff
@@ -1,35 +1,28 @@
...
-            self._active = True
-        else:
-            self._active = False
-
-    def __exit__(self, *args):
-        if self._active:
-            BusyCursor.active.pop(-1)
-            if len(BusyCursor.active) == 0:
-                QtGui.QApplication.restoreOverrideCursor()
+    app = QtCore.QCoreApplication.instance()
```
j9ac9k, line 20: you can have `pg.mkQApp()` here, it will create an app instance if one doesn't exist and return that, or return the current instance.

pyqtgraph/widgets/BusyCursor.py
```diff
@@ -1,35 +1,28 @@
-from ..Qt import QtGui, QtCore, QT_LIB
+# -*- coding: utf-8 -*-
+from contextlib import contextmanager
```
j9ac9k, line 2: another nice thing with going to python3 only!

pyqtgraph/widgets/BusyCursor.py
j9ac9k: unless you don't want this to start a QApplication instance; in which case, never mind :)

outofculture: I don't think this would be an appropriate `mkQApp` situation. I don't really know the full implications of making an app here, but in either case, not making one is at least the previous behavior, so that's some sense of safety in the decision.

pyqtgraph/widgets/BusyCursor.py
```diff
@@ -1,35 +1,28 @@
...
-__all__ = ['BusyCursor']
+from ..Qt import QtGui, QtCore
 
-class BusyCursor(object):
-    """Class for displaying a busy mouse cursor during long operations.
+__all__ = ["BusyCursor"]
+
+
+@contextmanager
+def BusyCursor():
```
j9ac9k, line 10: are the `()` necessary here?

outofculture: Yeah, since it's a function declaration, not a class any more. No parens here gives a SyntaxError.

j9ac9k: omg...i was _convinced_ this was a class ....

j9ac9k: 🤦🏻‍♂️

PR #1828, opened 2021-06-10
Add a few ImageView improvements

outofculture:
Taking up #396 (thanks, @radjkarl!), this PR does a little more cleanup. Some of the features were incomplete or had bugs. I fixed some of these, omitted others. Here's what happened to everything:

`setHistogramLabel` looks good and is connected to the example now.

`discreteTimeLine`, now an init argument, controls snap-to-frame in the timeline. This can now handle non-integer xvals. The example now uses this option. A test now checks this.

`nframes` handles zero frames now. A test now checks this.

`play`, `keyPressEvent` and `timeIndex` methods are working smoothly. Spacebar now toggles play-through, instead of only starting it. Example now plays.

`opts` needed additional documentation and scaffolding, and I didn't want to put in the work for what didn't feel particularly impactful.

`setHistogramPrintView` needed some improvement, but I didn't understand the expected behavior well enough to take it over, so I just removed it. If this is still valuable to someone, a version with a more thoroughly described spec would be an acceptable addition.


ixjlyons:
@outofculture is this ready for a review or were there still some things you wanted to work on?

outofculture:
@ixjlyons argh, no, it's not ready. I was testing things and ran into some problems that didn't have a clear solution, and I haven't come back to it, and now I'll have to start from scratch to figure out what those problems were. I might pare this down again in the process, but I'll put this back on the top of my pyqtgraph stack.

outofculture:
Okay, I finally got back to this and cleaned it up. The test is saner now, I cleaned up a few more docstrings, and I made sure pause/unpause would restore the previous play speed. This is ready for review.

pyqtgraph/imageview/ImageView.py
```diff
@@ -728,22 +792,28 @@ def normalize(self, image):
...
-        if not self.ignorePlaying:
+        if not self.ignoreTimeLine:
             self.play(0)
 
         (ind, time) = self.timeIndex(self.timeLine)
         if ind != self.currentIndex:
             self.currentIndex = ind
             self.updateImage()
+        if self.discreteTimeLine:
+            with fn.SignalBlock(self.timeLine.sigPositionChanged, self.timeLineChanged):
```
j9ac9k, line 803: nice use of `SignalBlock` probably should use it more in this library!

pyqtgraph/imageview/ImageView.py
```diff
@@ -761,25 +831,31 @@ def updateImage(self, autoHistogramRange=True):
...
+        -------
+        int
+            The index of the frame closest to the timeline slider.
+        float
+            The time value of the slider.
+        """
         if not self.hasTimeAxis():
-            return (0,0)
-        
+            return 0, 0
```
j9ac9k, line 845: nit-pick, but for type-correctness, probably should return `0, 0.0`

tests/imageview/test_imageview.py
```diff
@@ -12,6 +13,31 @@ def test_nan_image():
...
+    assert iv.nframes() == count
+    speed = count / 2
+    iv.play(speed)
+    assert iv.playRate == speed
+    iv.timeLine.setPos(0.51)
+    assert iv.playRate == 0
+    ind, val = iv.timeIndex(iv.timeLine)
+    assert ind == count // 2
+    assert val == 0.5
+    iv.togglePause()
```
j9ac9k, line 33: perhaps we should add a comment here indicating that this call to `togglePause` _should_ get playback going?

j9ac9k:
Hi @outofculture I did a look-through, looks good to me.  I just got nit-picky about the float vs. int return type and requested a comment in the test function, I think this is ready for merging otherwise!   Thanks for your work on this!

Issue #1781, opened 2021-05-18
PySide6 Rolling Out QPainter/GraphicsView Methods With NumPy Integration - Accepting Feature Requests


j9ac9k:
Tagging @pijyoi @lidstrom83 @campagnola @outofculture @NilsNemitz due to wanting input from you all on this.  Also going to tag some Orange devs @irgolic @markotoplak since you all have a vested interest here as well.

The PySide team at the Qt company have been slowly rolling out updates with optimized handling of numpy arrays into various methods.  With the [release of PySide6 6.1](https://www.qt.io/blog/qt-for-python-6.1), they added 3 new methods (with experimental support requiring you rebuild PySide from source)

* `QXYSeries.appendNp(PyArrayObject*, PyArrayObject*)`
* `QXYSeries.replaceNp(PyArrayObject*, PyArrayObject*)`
* `QPainter.drawPointsNp(PyArrayObject*, PyArrayObject*)`

For PySide2 5.15.2, the developers there added some highly optimized version of `QPainter.drawPolyline` that takes a numpy array in at the request of a maintainer of PythonQwt.

In conversation with the PySide developers on their [gitter channel](https://gitter.im/PySide/pyside2) they have expressed a willingness to add support to more methods that integrate with NumPy arrays, and I'm opening this issue to get a dialog started on what some good candidates for for methods that pyqtgraph could benefit from.

@lidstrom83 pointed out a candidate, for scatter plots would be `QPainter.drawPixmapNp(ndarray, QPixmap)`, where the ndarray has shape (4, n) (or (n, 4)); and each n entry would encode x, y, w, h coordiates for the QPixmap.

I know a lot of work has been done to create our own [arrayToQPath](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L1666-L1783) and [makeQImage](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L1393-L1497); but are there other QPainter methods (or GraphicsView methods in general) that we rely on where we would get a substantial performance boost if Qt accepted a numpy array? 

One of the first things that comes to mind is `PColorMeshItem`, given our [nested for loops](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/PColorMeshItem.py#L201-L216).

Doing some searching in the library, there may be some optimizations to be had in `AxisItem.py` as well, but I'm not seeing any major bottlenecks here.

[...]

Issue #1492, opened 2021-01-07
performance regression for ScatterPlot


danielhrisca:
### Short description
As suggested in #1359 I open this issue.
In [this custom plot widget](https://github.com/danielhrisca/asammdf/blob/master/asammdf/gui/widgets/plot.py#L1490) each curve is added to a dedicated ViewBox, and all the ViewBoxes are X-linked. I usually check the performance of the ScatterPlot (with dots) with ~1000-2000 curves.

In the profile data shared in #1359 I've used ~1700 curves.

There is a big drop in performance when comparing 0.11.1 with 0.11.0

### Code to reproduce

```python
import pyqtgraph as pg
import numpy as np
```

### Tested environment(s)

 * PyQtGraph version: 0.11.0 and 0.11.1
 * Qt Python binding: 5.15.2
 * Python version: 3.7.7 x64
 * NumPy version: 1.19.4
 * Operating system: Windows 10 x64
 * Installation method: pip

j9ac9k:
Hi @danielhrisca Thanks for reporting.  We made significant changes in #1420 how `ScatterPlotItem`, and @lidstrom83 was able to benchmark some significant performance improvements; but we are definitely curious to know if we slowed down performance.

Can you provide some dummy data or a script that we could use with your custom plot widget that we could replicate the drop in performance?

danielhrisca:
You will need to install the development branch of my projects first (it remove a [this monkey patched](https://github.com/danielhrisca/asammdf/blob/master/asammdf/gui/__init__.py#L6)  ``updateSpots`` method)

```
pip install git+https://github.com/danielhrisca/asammdf/@development
```

and then you can use this script

```python
from asammdf.gui.plot import plot
from asammdf import Signal
import numpy as np

v = np.arange(10000)
signals = [
    Signal(v, v, name=f'S_{i}')
    for i in range(400)
]

plot(signals)
```

j9ac9k:
Assigned "high-priority" label as this is impacted from a recent change; so I'd like to try and zero-in on where things went wrong exactly, as #1420 was supported to have significant benefits.  

@danielhrisca which Qt bindings are you using?  @lidstrom83 noticed that there was some very different performance behavior for PyQt5/Pyside2 and attempted to compensate for that by the use of `_USE_QRECT`

danielhrisca:
I'm using pyqt5.15 and pyqt5.13

j9ac9k:
[...]

PR #1829, opened 2021-06-11
try using QPainter.drawPixmapFragments

pijyoi:
This PR implements the method shown in https://github.com/pyqtgraph/pyqtgraph/issues/1781#issuecomment-858306326 to collapse multiple calls to drawPixmap() to a single call to drawPixmapFragments().

It's only meant to be an optimization for PySide bindings. On PySide bindings, it shows only a very small speedup on ScatterPlotSpeedTest.py. This is only meant to be a proof of concept to see if it's worthwhile to take things further (by someone else with more knowledge of how scatter plots are used).

As an aside, during testing with ScatterPlotSpeedTest.py, it appeared to me that using _USE_QRECT was slower on _all_ current bindings. i.e. the premise that it was faster on PyQt bindings didn't seem to be true. Perhaps someone more familiar with scatter plots could check this out.

### DIFF
--- a/pyqtgraph/graphicsItems/ScatterPlotItem.py
+++ b/pyqtgraph/graphicsItems/ScatterPlotItem.py
@@ -1134,9 +1134,7 @@ class ScatterPlotItem(GraphicsObject):
 
                 if _USE_PXFRAGS:
                     # x, y is the center of the target rect
-                    # drawPixmapFragments takes floating-point coords,
-                    # so casting to int here is for rounding towards zero
-                    xy = pts[:, viewMask].T.astype(int)
+                    xy = pts[:, viewMask].T
                     sr = self.data['sourceRect'][viewMask]
 
                     if not hasattr(self, 'pixmap_fragments'):
--- a/pyqtgraph/graphicsItems/ScatterPlotItem.py
+++ b/pyqtgraph/graphicsItems/ScatterPlotItem.py
@@ -29,22 +29,6 @@ elif QT_LIB in ['PyQt5', 'PyQt6']:
 __all__ = ['ScatterPlotItem', 'SpotItem']
 
 
-# When pxMode=True for ScatterPlotItem, QPainter.drawPixmap is used for drawing, which
-# has multiple type signatures. One takes int coordinates of source and target
-# rectangles, and another takes QRectF objects. The latter approach has the overhead of
-# updating these objects, which can be almost as much as drawing.
-# For PyQt5, drawPixmap is significantly faster with QRectF coordinates for some
-# reason, offsetting this overhead. For PySide2 this is not the case, and the QRectF
-# maintenance overhead is an unnecessary burden. If this performance issue is solved
-# by PyQt5, the QRectF coordinate approach can be removed by simply deleting all of the
-# "if _USE_QRECT" code blocks in ScatterPlotItem. Ideally, drawPixmap would accept the
-# numpy arrays of coordinates directly, which would improve performance significantly,
-# as the separate calls to this method are the current bottleneck.
-# See: https://bugreports.qt.io/browse/PYSIDE-163
-
-_USE_QRECT = False
-_USE_PXFRAGS = True
-
 ## Build all symbol paths
 name_list = ['o', 's', 't', 't1', 't2', 't3', 'd', '+', 'x', 'p', 'h', 'star',
              'arrow_up', 'arrow_right', 'arrow_down', 'arrow_left', 'crosshair']
@@ -434,18 +418,11 @@ class ScatterPlotItem(GraphicsObject):
             ])
         ]
 
-        if _USE_QRECT:
-            dtype.extend([
-                ('sourceQRect', object),
-                ('targetQRect', object),
-                ('targetQRectValid', bool)
-            ])
-            self._sourceQRect = {}
-
         self.data = np.empty(0, dtype=dtype)
         self.bounds = [None, None]  ## caches data bounds
         self._maxSpotWidth = 0      ## maximum size of the scale-variant portion of all spots
         self._maxSpotPxWidth = 0    ## maximum size of the scale-invariant portion of all spots
+        self._pixmapFragments = PixmapFragments()
         self.opts = {
             'pxMode': True,
             'useCache': True,  ## If useCache is False, symbols are re-drawn on every paint.
@@ -582,10 +559,6 @@ class ScatterPlotItem(GraphicsObject):
         newData = self.data[len(oldData):]
         newData['size'] = -1  ## indicates to use default size
         newData['visible'] = True
-
-        if _USE_QRECT:
-            newData['targetQRect'] = [QtCore.QRectF() for _ in range(numPts)]
-            newData['targetQRectValid'] = False
 
         if 'spots' in kargs:
             spots = kargs['spots']
@@ -846,18 +819,6 @@ class ScatterPlotItem(GraphicsObject):
                     list(zip(*self._style(['symbol', 'size', 'pen', 'brush'], data=dataSet, idx=mask)))
                 ]
                 dataSet['sourceRect'][mask] = coords
-                if _USE_QRECT:
-                    rects = []
-                    for c in coords:
-                        try:
-                            rect = self._sourceQRect[c]
-                        except KeyError:
-                            rect = QtCore.QRectF(*c)
-                            self._sourceQRect[c] = rect
-                        rects.append(rect)
-
-                    dataSet['sourceQRect'][mask] = rects
-                    dataSet['targetQRectValid'][mask] = False
 
             self._maybeRebuildAtlas()
         else:
@@ -875,8 +836,6 @@ class ScatterPlotItem(GraphicsObject):
                 list(zip(*self._style(['symbol', 'size', 'pen', 'brush'])))
             )
             self.data['sourceRect'] = 0
-            if _USE_QRECT:
-                self._sourceQRect.clear()
             self.updateSpots()
 
     def _style(self, opts, data=None, idx=None, scale=None):
@@ -1059,8 +1018,6 @@ class ScatterPlotItem(GraphicsObject):
         self.prepareGeometryChange()
         GraphicsObject.viewTransformChanged(self)
         self.bounds = [None, None]
-        if _USE_QRECT:
-            self.data['targetQRectValid'] = False
 
     def setExportMode(self, *args, **kwds):
         GraphicsObject.setExportMode(self, *args, **kwds)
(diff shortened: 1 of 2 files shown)


j9ac9k:
> As an aside, during testing with ScatterPlotSpeedTest.py, it appeared to me that using _USE_QRECT was slower on all current bindings. i.e. the premise that it was faster on PyQt bindings didn't seem to be true. Perhaps someone more familiar with scatter plots could check this out.

I have pyenv setup w/ a bazillion virtual environments for pyqtgraph set for macOS and Linux.  I can iterate through all the combinations and see if I can identify where it makes sense to use that option.

j9ac9k:
Tagging @ibrewster and @danielhrisca due their invested interest in scatter plot performance

j9ac9k:
I ran the profiler on the ScatterPlotSpeedTest on here for a bit and let it chew on it; and it does not appear the painting is currently the bottleneck.  Almost ~40% of the time is occupied in `SymbolAtlas.__getitem__`

`ScatterPlotItem.paint` was only occupying 6% of the overall time (!) means that this improvement while making a < 10% improvement in FPS, is likely a huge improvement in the speed of the paint methods; and probably worthwhile of implementing, but there are probably bigger fish to fry in the ScatterPlotItem optimizations (potentially implementing [QPixmapCache](http://doc.qt.io/qt-6/qpixmapcache.html) in-place of the SymbolAtlas.

In case anyone is curious:

ScatterPlotSpeedTestGraph

pijyoi:
>It's only meant to be an optimization for PySide bindings.

With the last commit, it is now an optimization for both PySide and PyQt bindings.

j9ac9k:
I just noticed that this diff results in a larger difference in the pan/zoom use-case in the speed-test ( >10%) for PySide6.  That difference alone makes me think this change is definitely worth integrating into the library.

pijyoi:
>That difference alone makes me think this change is definitely worth integrating into the library.

drawPixmapFragments > non-_USE_QRECT > _USE_QRECT

The drawPixmapFragments codepath could replace both the old _USE_QRECT and non-_USE_QRECT codepaths since it is now faster for all bindings.

On the one hand, deleting the _USE_QRECT codepath would simplify the code a lot. On the other hand, I am loath to delete out such a large chunk of code without more inputs from the original author(s) (@lidstrom83)

Perhaps at least the non-_USE_QRECT codepath could be left behind for now to allow users to test it out interactively with ScatterPlotSpeedTest.py?

j9ac9k:
> Perhaps at least the non-_USE_QRECT codepath could be left behind for now to allow users to test it out interactively with ScatterPlotSpeedTest.py?

lidstrom83 has had some matters offline arise so I haven't been able to touch base much recently; I've reached out... I think we should sit on this for at least a few days in hopes of getting their input.  I'd also like to get input from the two people I tagged as historically they have had fairly demanding scatter plot item performance needs.

pijyoi:
Linking to #1492 where various benchmarks seem to have been done

lidstrom83:
I couldn't figure out how to make drawPixmapFragments work before - I'm happy to see that you did!

Upgrading numpy from 1.19.5 to 1.20.3 makes USE_QRECT slower on the speed test with PyQt5/6 in the New Item and Reuse Item modes. Not for the Pan/Zoom and Hover modes, however. I find USE_PXFRAGS to be faster across the board, except in PyQt5/6 in Pan/Zoom and Hover modes where it's slightly slower. I think it's due to PyQt5/6 requiring us to do this:

`
                        frags_lst = list(map(sip.wrapinstance,                            itertools.count(frags.ctypes.data, frags.strides[0]),                            itertools.repeat(QtGui.QPainter.PixmapFragment, frags.shape[0])))`


I was reluctant to add the USE_QRECT code mess, but the performance gains seemed to justify it. If it's now just a small gain in a couple use cases it may not be worth keeping it. I'm fine with removing the other two code paths. Especially so if we can speed up the above code somehow.

pijyoi:
>I couldn't figure out how to make drawPixmapFragments work before

In terms of bindings exposed, the PySide drawPixmapFragments api is an anomaly: pointer to an contiguous array of structs. The other drawXXXs apis aren't like this. So I do wonder whether it was intentional for drawPixmapFragments.

>>I think it's due to PyQt5/6 requiring us to do this:

>frags_lst = list(map(sip.wrapinstance, itertools.count(frags.ctypes.data, frags.strides[0]), itertools.repeat(QtGui.QPainter.PixmapFragment, frags.shape[0])))

In fact, the instance pointers in the Python list are then dereferenced by PyQt and copied into a freshly allocated contiguous array. PyQt then calls the underlying C++ api, which is the one directly exposed by PySide.

>Especially so if we can speed up the above code somehow.

I added caching of the instance pointers. Does that help?

j9ac9k:
@lidstrom83 good to hear from you!  Thanks for chiming in, I was able to replicate the performance benefits you detail with `_USE_QRECT` on PyQt5/6 with zoom/pan/hover as you detail.

> I added caching of the instance pointers. Does that help?

@pijyoi this does indeed help, `_USE_PXFRAGS` outperforms `_USE_QRECT` by ~10% on hover/mouse with the default settings in the benchmarks, and the difference in performance grows as the number of points increases.  Nice work!

j9ac9k:
> In terms of bindings exposed, the PySide drawPixmapFragments api is an anomaly: pointer to an contiguous array of structs.

@pijyoi just out of curiosity, how do you determine this?

pijyoi:
>just out of curiosity, how do you determine this (pointer to a contiguous array of structs)?

Take the following C++ apis. All of them take a pointer to a contiguous array of structs.
drawLines and drawRects have additional overloads that take a QVector.
```c++
drawPixmapFragments(const PixmapFragment *fragments, int fragmentCount);
drawLines(const QLineF *lines, int lineCount);
drawRects(const QRectF *rectangles, int rectCount);
drawPoints(const QPointF *points, int pointCount);
```

If you look at the python bindings, ```drawLines``` and ```drawRects``` do not expose their pointer to structs api, whereas ```drawPixmapFragments``` exposes it. (```drawPixmapFragments``` does not have any other QVector overload api to begin with)
```python
In [2]: QtGui.QPainter.drawPixmapFragments?
Signature:
QtGui.QPainter.drawPixmapFragments(
    self,
    fragments: PySide6.QtGui.QPainter.PixmapFragment,
    fragmentCount: int,
    pixmap: PySide6.QtGui.QPixmap,
    hints: PySide6.QtGui.QPainter.PixmapFragmentHints = Default(QPainter.PixmapFragmentHints),
) -> None

In [3]: QtGui.QPainter.drawLines?
Docstring:
drawLines(self, lines: typing.Sequence[PySide6.QtCore.QLineF]) -> None
drawLines(self, lines: typing.Sequence[PySide6.QtCore.QLine]) -> None
drawLines(self, pointPairs: typing.Sequence[PySide6.QtCore.QPointF]) -> None
drawLines(self, pointPairs: typing.Sequence[PySide6.QtCore.QPoint]) -> None
```

At first sight, the PySide binding to ```drawPixmapFragments``` seems strange as it looks as though we could only ever pass in one ```PixmapFragment```. Determining that it would take a wrapinstance(raw pointer to first element) was by trial and error, which is what https://github.com/pyqtgraph/pyqtgraph/issues/1781#issuecomment-858306326 was showing.

So, in fact, rather than add numpy specific apis like ```drawPointsNp```, it would be better to just expose the pointer to structs c++ api, which would be numpy-independent. After all, QRectF, QLineF, QPointF are just a struct of 4,4,2 float64s respectively in memory, Though, of course, it is non-Pythonic and non-type-safe.

j9ac9k:
> So, in fact, rather than add numpy specific apis like drawPointsNp, it would be better to just expose the pointer to structs c++ api, which would be numpy-independent.

This might be an easier ask than specialized numpy methods; which would be a _really_ tall ask to try and get both bindings to try and implement.

lidstrom83:
> I added caching of the instance pointers. Does that help?

Yep. It's faster across the board now.

pijyoi:
There's a doubt that I have regarding the old-codepaths.

The USE_QRECT codepath uses ```drawPixmap(const QRectF &target, const QPixmap &pixmap, const QRectF &source)``` whereas
the non-USE_QRECT codepath uses ```drawPixmap(int x, int y, const QPixmap &pixmap, int sx, int sy, int sw, int sh)```

Since the source pixmap coords are all integral, use of ```source``` vs ```sx```, ```sy```, ```sw```, ```sh``` is equivalent.
However, for the target coords, the non-USE_QRECT codepath truncates the coords to integers whereas the USE_QRECT codepath uses the untruncated floating-point coords.

I am going to guess that the preferred behavior was to use untruncated floating point target coords. That it was an API limitation for the non-USE_QRECT codepath to be using truncated coords.

danielhrisca:
> There's a doubt that I have regarding the old-codepaths.
> 
> The USE_QRECT codepath uses `drawPixmap(const QRectF &target, const QPixmap &pixmap, const QRectF &source)` whereas
> the non-USE_QRECT codepath uses `drawPixmap(int x, int y, const QPixmap &pixmap, int sx, int sy, int sw, int sh)`
> 
> Since the source pixmap coords are all integral, use of `source` vs `sx`, `sy`, `sw`, `sh` is equivalent.
> However, for the target coords, the non-USE_QRECT codepath truncates the coords to integers whereas the USE_QRECT codepath uses the untruncated floating-point coords.
> 
> I am going to guess that the preferred behavior was to use untruncated floating point target coords. That it was an API limitation for the non-USE_QRECT codepath to be using truncated coords.

I always got better performance with `_USE_QRECT=False`

pyqtgraph/graphicsItems/ScatterPlotItem.py
```diff
@@ -164,6 +171,29 @@ def _mkBrush(*args, **kwargs):
...
+                self.ptrs = list(map(sip.wrapinstance,
+                    itertools.count(self._array.ctypes.data, self._array.strides[0]),
+                    itertools.repeat(QtGui.QPainter.PixmapFragment, self._array.shape[0])))
+            else:
+                self.ptrs = wrapInstance(self._array.ctypes.data, QtGui.QPainter.PixmapFragment)
+        return self._array[:size]
+
+    def draw(self, painter, size, pixmap):
+        if QT_LIB.startswith('PyQt'):
+            painter.drawPixmapFragments(self.ptrs[:size], pixmap)
```
lidstrom83, line 192: This throws an error for zero length data: self.ptrs is None in that case.

lidstrom83: > If you look at the python bindings, `drawLines`, `drawRects`, `drawPoints` do not expose their pointer to structs api, whereas `drawPixmapFragments` exposes it.

> So, in fact, rather than add numpy specific apis like `drawPointsNp`, it would be better to just expose the pointer to structs c++ api, which would be numpy-independent. After all, QRectF, QLineF, QPointF are just a struct of 4,4,2 float64s respectively in memory, Though, of course, it is non-Pythonic and non-type-safe.

@j9ac9k have you been in touch with the PySide developers about this?

lidstrom83:
> If you look at the python bindings, drawLines, drawRects, drawPoints do not expose their pointer to structs api, whereas drawPixmapFragments exposes it.

> So, in fact, rather than add numpy specific apis like drawPointsNp, it would be better to just expose the pointer to structs c++ api, which would be numpy-independent. After all, QRectF, QLineF, QPointF are just a struct of 4,4,2 float64s respectively in memory, Though, of course, it is non-Pythonic and non-type-safe.

@j9ac9k have you been in touch with the PySide developers about this?

pyqtgraph/graphicsItems/ScatterPlotItem.py
j9ac9k: I opened an issue regarding `drawPointsNp` only got a brief reply but suggested they were leaning against implementing it as they didn't think it was general enough.

I did ask on their gitter channel if it would be feasible to expose the pointer for the underlying struct of data the other draw methods take.  Hard for me to gauge work effort involved there.

j9ac9k:
> > If you look at the python bindings, drawLines, drawRects, drawPoints do not expose their pointer to structs api, whereas drawPixmapFragments exposes it.
> 
> > So, in fact, rather than add numpy specific apis like drawPointsNp, it would be better to just expose the pointer to structs c++ api, which would be numpy-independent. After all, QRectF, QLineF, QPointF are just a struct of 4,4,2 float64s respectively in memory, Though, of course, it is non-Pythonic and non-type-safe.
> 
> @j9ac9k have you been in touch with the PySide developers about this?

I opened an issue regarding `drawPointsNp` only got a brief reply but suggested they were leaning against implementing it as they didn't think it was general enough.  It was a brief message, but I had the impression that they were open to hearing a stronger pitch, but as pijyoi realized we can do it the way he did in this PR; I didn't feel the need to push them on this front 

I did ask on their gitter channel if it would be feasible to expose the pointer for the underlying struct of data the other draw methods take.  Hard for me to gauge work effort involved there.  I haven't received a reply yet.

Regarding the `draw*Np` methods; I'm not that thrilled about it as it would only work with one set of bindings; and I'm afraid it would complicate the code for trying to implement equivalent functionality in PyQt5/6.

pyqtgraph/graphicsItems/ScatterPlotItem.py
pijyoi: fixed

pijyoi:
>Hard for me to gauge work effort involved there.

Potentially the effort involves removing the following lines from 
```...\Lib\site-packages\PySide6\typesystems\typesystem_gui_common.xml```
and rebuilding PySide

```xml
<modify-function signature="drawLines(const QLineF*,int)" remove="all"/>
<modify-function signature="drawLines(const QPointF*,int)" remove="all"/>
<modify-function signature="drawRects(const QRectF*,int)" remove="all"/>
<modify-function signature="drawPoints(const QPointF*,int)" remove="all"/>
```

j9ac9k:
> >Hard for me to gauge work effort involved there.
> 
> 
> 
> Potentially the effort involves removing the following lines from 
> 
> ```...\Lib\site-packages\PySide6\typesystems\typesystem_gui_common.xml```
> 
> and rebuilding PySide
> 
> 
> 
> ```xml
> 
> <modify-function signature="drawLines(const QLineF*,int)" remove="all"/>
> 
> <modify-function signature="drawLines(const QPointF*,int)" remove="all"/>
> 
> <modify-function signature="drawRects(const QRectF*,int)" remove="all"/>
> 
> <modify-function signature="drawPoints(const QPointF*,int)" remove="all"/>
> 
> ```

I'll try to recompile pyside myself and see what happens.

I think this PR is ready for merging once the use_qrect and non use_pxfrags and non_qrect code paths are removed.

j9ac9k:
Thanks @pijyoi 

There are some unused imports now (`repeat` and `chain`) at the top, would you mind removing that?  Also could you add a comment about why we preserve a reference for PyQt5/6 bindings (getting the dereference comment you made here into the code-base as a comment), I can see that being worth noticing the next time someone goes through this code.

pijyoi:
Hmm, HistogramLUTWidget still segfaulting on PySide6 Linux, despite #1835 being merged. I thought the CI auto-merges to the tip of master before running?

j9ac9k:
> Hmm, HistogramLUTWidget still segfaulting on PySide6 Linux, despite #1835 being merged. I thought the CI auto-merges to the tip of master before running?

It does (as evidence by one of the older draft PRs that needed updating after the enum namespace change when promote_enums was removed.  I suspect what's more likely another source of segfaults within the test suite.

j9ac9k:
I just manually triggered a re-run; we'll see if it's a repeatable fault.

j9ac9k:
Saw the fault there, but given I've seen the HistogramLUTwidget segfault before I'm confident it has little to do with this PR; so I'll merge this PR.

Thanks @pijyoi for making this happen, don't know what's better, the diff or the performance boost, both are pretty awesome :D

PR #1833, opened 2021-06-12
fix ScatterPlot render issues on PyQt6 6.1

### LINKED ISSUE #1832: examples/ScatterPlot.py : different output for PyQt6
### Short description

When running examples/ScatterPlot.py, PyQt6 binding creates an output that looks different from the other bindings.

### Code to reproduce

```examples/ScatterPlot.py```

### Expected behavior

Ouput from all bindings should look similar.
Output from PyQt5, PySide2, PySide6
scatter_pyqt5

### Real behavior

Output from PyQt6 looks different than other bindings
scatter_pyqt6

### Tested environment(s)

 * PyQtGraph version: master 2021/06/12 
 * Qt Python binding: PyQt5 5.15, PySide2 5.15, PySide6 6.1.1, PyQt6 6.1.0 
 * Python version: 3.8.10
 * NumPy version: 1.20.3 
 * Operating system: Windows 10 20H2

### Additional context

This seems to be a regression with PyQt6 6.1 with respect to PyQt6 6.0.
Checking out a commit of pyqtgraph that works with both PyQt6 6.0 and PyQt6 6.1 shows that 6.0 gives the expected output while 6.1 gives the differing output.

### PR DESCRIPTION
**Pull request by pijyoi**
This PR fixes #1832.

The problem may have to do with the signature of ``QImage.fill()``` being changed slightly.

PyQt5 5.15, PyQt6 6.0
```python
In [4]: QtGui.QImage.fill?
Docstring:
fill(self, Qt.GlobalColor)
fill(self, Union[QColor, Qt.GlobalColor, QGradient])
fill(self, int)
Type:      builtin_function_or_method
```

PyQt6 6.1
```python
In [2]: QtGui.QImage.fill?
Docstring:
fill(self, Qt.GlobalColor)
fill(self, Union[QColor, Qt.GlobalColor, int, QGradient])
fill(self, int)
Type:      builtin_function_or_method
```

Doing a search for other potential problems in the library yields:
```
$ git grep "fill(0"
pyqtgraph/graphicsItems/ROI.py:        im.fill(0x0)
pyqtgraph/widgets/RemoteGraphicsView.py:            self.img.fill(0xffffffff)
```
I don't understand ROI enough to say whether it needs the fix.

ScatterPlot needs the fix because it needs the undrawn background to be transparent.
RemoteGraphicsView doesn't seem to be affected, likely because it draws on the whole QImage.

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -23,6 +23,7 @@ from .GraphicsObject import GraphicsObject
 from .UIGraphicsItem import UIGraphicsItem
 from .. import getConfigOption
 import warnings
+import sys
 
 translate = QtCore.QCoreApplication.translate
 
@@ -1287,7 +1288,7 @@ class ROI(GraphicsObject):
             return np.empty((width, height), dtype=float)
         
         im = QtGui.QImage(width, height, QtGui.QImage.Format.Format_ARGB32)
-        im.fill(0x0)
+        im.fill(QtCore.Qt.GlobalColor.transparent)
         p = QtGui.QPainter(im)
         p.setPen(fn.mkPen(None))
         p.setBrush(fn.mkBrush('w'))
@@ -1297,8 +1298,9 @@ class ROI(GraphicsObject):
         p.translate(-bounds.topLeft())
         p.drawPath(shape)
         p.end()
-        mask = fn.imageToArray(im, transpose=True)[:,:,0].astype(float) / 255.
-        return mask
+        cidx = 0 if sys.byteorder == 'little' else 3
+        mask = fn.qimage_to_ndarray(im)[...,cidx].T
+        return mask.astype(float) / 255
         
     def getGlobalTransform(self, relativeTo=None):
         """Return global transformation (rotation angle+translation) required to move 
--- a/pyqtgraph/graphicsItems/ScatterPlotItem.py
+++ b/pyqtgraph/graphicsItems/ScatterPlotItem.py
@@ -120,7 +120,7 @@ def renderSymbol(symbol, size, pen, brush, device=None):
     penPxWidth = max(math.ceil(pen.widthF()), 1)
     if device is None:
         device = QtGui.QImage(int(size+penPxWidth), int(size+penPxWidth), QtGui.QImage.Format.Format_ARGB32)
-        device.fill(0)
+        device.fill(QtCore.Qt.GlobalColor.transparent)
     p = QtGui.QPainter(device)
     try:
         p.setRenderHint(p.RenderHint.Antialiasing)


pijyoi:
So a MWE for reporting the issue upstream would be:

PyQt5, PyQt6 6.0
```python
In [3]: qimg = QtGui.QImage(1, 1, QtGui.QImage.Format.Format_ARGB32)
In [4]: qimg.fill(0)
In [5]: hex(qimg.pixel(0, 0))
Out[5]: '0x0'
```

PyQt 6.1
```python
In [16]: qimg = QtGui.QImage(1, 1, QtGui.QImage.Format.Format_ARGB32)
In [17]: qimg.fill(0)
In [18]: hex(qimg.pixel(0, 0))
Out[18]: '0xff000000'
```

pijyoi:
>I don't understand ROI enough to say whether it needs the fix.

After examining it closer, ROI.py needs the ```QtCore.Qt.GlobalColor.transparent``` fix for big-endian systems.
On little-endian systems, the original code grabbed channel 0 (blue), ignoring the alpha channel. Thus it is not affected by the PyQt6 6.1 alpha value issue.
On big-endian systems, channel 0 is the alpha channel and thus is affected by the PyQt6 6.1 alpha value issue.

The issue can be simulated on little-endian systems by changing the code to select channel 3, the alpha channel.
On the other bindings, running examples\ROIExamples.py will show a triangular ROI.
On PyQt6 6.1, it will show a rectangular ROI.

The big-endian simulated outputs are as follows:

PySide6 6.1
roi_good

PyQt6 6.1
roi_bad

pijyoi:
By the way, 2fb04b754c0b11f70ae82b187bb50ba567986883 seems to be triggering more segfaults.
I tried running this PR with that commit reverted and the CI passes:
https://github.com/pijyoi/pyqtgraph/actions/runs/931219179

j9ac9k:
> By the way, [2fb04b7](https://github.com/pyqtgraph/pyqtgraph/commit/2fb04b754c0b11f70ae82b187bb50ba567986883) seems to be triggering more segfaults.
> I tried running this PR with that commit reverted and the CI passes:
> https://github.com/pijyoi/pyqtgraph/actions/runs/931219179

Thanks for identifying this.  I don't see anything in that test that will cause a segfault, but doesn't mean it is not triggering an issue that is caused elsewhere.  I'll try and investigate this a bit more.

j9ac9k:
> So a MWE for reporting the issue upstream would be:
> 
> PyQt5, PyQt6 6.0
> 
> ```python
> In [3]: qimg = QtGui.QImage(1, 1, QtGui.QImage.Format.Format_ARGB32)
> In [4]: qimg.fill(0)
> In [5]: hex(qimg.pixel(0, 0))
> Out[5]: '0x0'
> ```
> 
> PyQt 6.1
> 
> ```python
> In [16]: qimg = QtGui.QImage(1, 1, QtGui.QImage.Format.Format_ARGB32)
> In [17]: qimg.fill(0)
> In [18]: hex(qimg.pixel(0, 0))
> Out[18]: '0xff000000'
> ```

Thanks for making this!

j9ac9k:
This change looks good to me; hate that we have to have more byte order specific code; but what can you do.  Merging.

pijyoi:
>hate that we have to have more byte order specific code; but what can you do.

In theory, the ```QtCore.Qt.GlobalColor.transparent``` fix would make the endian-specific code unnecessary. But then little and big endian would be using different channels for creating the mask. Since the mask shape is painted in white, it would work However, I thought that it would be less confusing to the reader to make both little and big endian use the same channel.

j9ac9k:
Just got confirmation from Phil at Riverbank that this will be fixed in 6.1.1; not that it matters from our end anymore.

PR #1834, opened 2021-06-12
Add option to limit LinearRegionitem bounds to a secondary item

ixjlyons:
Picks up from #740. The actual clipping wasn't really implemented there, so I took a shot at it.

I ended up using the item's `ViewBox` to compute its bounds via `childrenBounds` because that should be fairly robust for a variety of different item types. Notably, using some form of mapping the item's `boundingRect` doesn't work for `PlotDataItem` because it doesn't implement `boundingRect`. It also fails silently, so it'd be tricky for a user to figure out you need to use `PlotDataItem.curve` or similar instead.

Another benefit is `childrenBounds` gives bounds that are more appropriate for data - if you use the region for computation, hitting the limit would give values you expect.

The downside is `childrenBounds` may be a little expensive to put in `LinearRegionItem.boundingRect`. I haven't figured out if there's a better way to be notified of an item's range changing or if there's a cheaper-to-compute object to cache for invalidating the bounds.

WIP for now to:

- [x] Further evaluate if there's a better mechanism for updating the bounds (or if there's a better place to do the update than `boundingRect`)
- [x] Handle the case that a user calls `setBounds` directly - I think this should disable the `clipItem`
- [x] Tests for more edge cases


ixjlyons:
Script I'm using for testing:

```python
import numpy as np
import pyqtgraph as pg

app = pg.mkQApp()

# un-bounded LRI
p1 = win.addPlot(y=np.random.normal(size=100),
                 title="unbounded")
lr1 = pg.LinearRegionItem((10, 30))
p1.addItem(lr1)

# explicit bounds
p2 = win.addPlot(y=np.sin(2*np.pi*np.arange(100)/100),
                 title="explicit bounds [20, 80]")
lr2 = pg.LinearRegionItem((10, 30), bounds=[20, 80])
p2.addItem(lr2)

win.nextRow()

# bound to a PlotDataItem
x = np.linspace(-1, 1, 10)
p3 = win.addPlot(title="bound to PlotDataItem, limited span")
pdi3 = p3.plot(x=x, y=np.sin(x)/x)
lr3 = pg.LinearRegionItem((-0.5, 0.5), clipItem=pdi3, span=(0.1, 0.9))
p3.addItem(lr3)

# horizontal orientation, bound to ImageItem
p4 = win.addPlot(title="bound to ImageItem, horizontal")
img = pg.ImageItem(image=np.random.randn(20, 20))
p4.addItem(img)
p4.setAspectLocked(True)
lr4 = pg.LinearRegionItem((5, 10), orientation="horizontal", clipItem=img)
p4.addItem(lr4)

# lr4.setBounds((0, 40))  # overriding clipItem with explicit setBounds

win.nextRow()

# "remote" PlotDataItem to bound to
p5 = win.addPlot(title="bound to a this plot data")
pdi5 = p5.plot(y=np.random.randn(100), pen=None, symbol="o")

# bound to a PlotDataItem in a different ViewBox
p6 = win.addPlot(title="&lt;-- bound to that plot data")
x = np.linspace(-200, 200, 100)
p6.plot(x=x, y=np.random.randn(x.size))
lr6 = pg.LinearRegionItem((20, 80), clipItem=pdi5)
p6.addItem(lr6)

pg.exec()
```

j9ac9k:
@ixjlyons you may want to merge master into this branch, I'm getting some enum related errors from PyQt6 bindings; luckily no merge conflicts when I did a merge locally :D

j9ac9k:
FWIW I don't think the performance hit is that bad for using childrenBounds... I enabled the profiler, and never got to a runtime for that method of 1 ms (most of the time was under 0.2 ms).  It didn't run all that many times either.

ixjlyons:
Alright I think this is ready for review. Added some tests that I think exercise the feature pretty well. If we come up with a better approach later it should be possible to replace it without breaking anything.

I do kind of want to issue a warning if you pass a `PlotItem` as `clipItem`, but I'm not seeing a straightforward way to do that without replicating the warning in `__init__` and `setClipItem`.

Also, I've wondered if `clipItem` is the right name.

j9ac9k:
@ixjlyons This LGTM, merging.  Thanks for sticking with this for so long!  @achennu thanks for proposing the feature to begin with, sorry it's taken us so long to incorporate it!

PR #1835, opened 2021-06-14
skip pytest monkeypatch

pijyoi:
Disabling the tests using pytest monkeypatch seems to stop the segfaults.

PR #1837, opened 2021-06-16
Revert "skip pytest monkeypatch"

j9ac9k:
Reverts pyqtgraph/pyqtgraph#1835


j9ac9k:
Turns out we're still seeing segfaults we thouht this PR would fix, so we're reverting this change..

PR #665, opened 2018-04-13
feature More parameter item types

feketeimre:
Pen: Pops up a dialogue that allows the user to customize a pen. Setting pen value is not working yet.
Progress bar: For indication things.
Slider: Easier way to set values that dont require precision.
Fonts: Picking font types. Next thing could be a Font dialog.
Calendar: For picking dates or intervals
Open/save file/files/directory: Pops up an open/save file/directory dialog to select a file/directory. Filter string and caption can be defined too.

A PenSelectorDialog widget was created for the pen parameter item too.

Also added these parameter items to the example.

PR #1844, opened 2021-06-18
feature More parameter item types

ntjess:
It's highly likely that I failed or missed something in this effort. The pull request is an attempted direct copy from @feketeimre's older one (#665), but that passes all current `master` tests. If they can confirm whether anything is missing, this should be much easier to merge into the current branch and we can close the very outdated `feature3` pull request

__TODO:__
- [x] Input sanitation on line-edit file parameter (or set to readonly)
- [x] Convert calendar to string on saveState (maybe add a `format` option); QDate has convenient `toString` method
- [x] Convert font to string on saveState
- [x] Better value saving on pen (no obvious way to include the entire pen state in the `value` field, so I'm currently playing with always saving a black pen and setting options like `color`, `width`, etc. to compensate. Has the downside of functioning differently from every other parameter whose value is self-contained...
- [x] A few Parameter bugfixes along the way (Slider optsChanged, WidgetParameterItem's signal hookup on value change 
__________
Pen: Pops up a dialouge that allows the user to customize a pen. Setting pen value is not working yet.
Progress bar: For indication things.
Slider: Easier way to set values that dont require precision.
Fonts: Picking font types. Next thing could be a Font dialog.
Calendar: For picking dates or intervals
Open/save file/files/directory: Pops up an open/save file/directory dialog to select a file/directory. Filter string and caption can be defined too.

A PenSelectorDialog widget was created for the pen parameter item too.

Also added these parameter items to the example.


pijyoi:
Some bindings issues:
1) PySide{2, 6} does not have QVariant
2) PyQt6 requires long spelling for enums (see #1818)

feketeimre (approved):

ntjess:
@pijyoi wow, turns out combo boxes implicitly convert enums to ints, fun...

It should be fixed now, thanks.

pyqtgraph/widgets/PenSelectorDialog.py
```diff
@@ -0,0 +1,99 @@
...
+    def __init__(self, initialPen = None, parent = None):
+        QtGui.QDialog.__init__(self, parent)
+        self.ui = PenSelectorDialogbox.Ui_Dialog()
+        self.ui.setupUi(self)
+        # Combo boxes don't like enum values, so they get converted to int. But, PySide/Qt6 doesn't accept int
+        # pen constructor values, so map the int to enum values. PySide/Qt6 Enums let you iterate over them, but
+        # PyQt5 doesn't, so construct manually for compatibility
+        ps = QtCore.Qt.PenStyle
+        lineStyles = [ps.SolidLine,ps.DashLine,ps.DotLine,ps.DashDotLine,ps.DashDotDotLine]#,ps.CustomDashLine]
+        self.lineStyles = {int(style): style for style in lineStyles}
```
pijyoi, line 17: PyQt6 doesn't allow cast of enums to int

pyqtgraph/parametertree/parameterTypes.py
```diff
@@ -741,3 +743,240 @@ class TextParameter(Parameter):
...
+        tree.resizeColumnToContents(0)
+        tree.setItemWidget(self, 0, self.layoutWidget)
+
+    def buttonClicked(self):
+        #open up the pen selector dialog
+        self.oldPen = fn.mkPen(self.pen)
+        self.pdialog = PenSelectorDialog(fn.mkPen(self.pen),QtGui.QApplication.activeWindow())
+        self.pdialog.penChanged.connect(self.penChanged)
+        self.pdialog.finished.connect(self.penChangeFinished)
+        self.pdialog.exec_()
```
pijyoi, line 940: PyQt6 has renamed exec_() to exec()
PySide6 6.1 deprecates exec_() in favor of exec()

pyqtgraph/parametertree/parameterTypes.py
```diff
@@ -741,3 +743,240 @@ class TextParameter(Parameter):
...
+        if dialogType in ["getExistingDirectory","openDirectory"]:
+            fn = dialogFunction(None, dialogCaption, self.lastDirectory)
+            fn = os.path.realpath(asUnicode(fn))
+            if fn=="": return 0
+            self.lastDirectory = os.path.dirname(fn)
+            self.label.setText(str(fn))
+            tooltip = str(os.path.realpath(fn))
+        elif dialogType in ["getOpenFileNames","openFiles"]:
+            fn = dialogFunction(None, dialogCaption, self.lastDirectory, filterString)
+            fn = [os.path.realpath(asUnicode(f)) for f in fn if os.path.isfile(f) and f != ""]
```
pijyoi, line 805: this throws an Exception if the user clicks on Cancel in the dialog

pyqtgraph/widgets/PenSelectorDialogbox.py
```diff
@@ -0,0 +1,114 @@
+# -*- coding: utf-8 -*-
+
+# Form implementation generated from reading ui file 'PenSelectorDialogbox.ui'
+#
+# Created by: PyQt4 UI code generator 4.11.4
```
pijyoi, line 4: I know this was generated using PyQt4, but if it had been generated with Qt5 or Qt6, many of the widgets come from QtWidgets, not QtGui.

pyqtgraph/widgets/PenSelectorDialogbox.py
```diff
@@ -0,0 +1,114 @@
...
+#
+# WARNING! All changes made in this file will be lost!
+
+from ..Qt import QtCore, QtGui
+
+try:
+    _fromUtf8 = QtCore.QString.fromUtf8
```
pijyoi: None of the current bindings has QtCore.QString

pyqtgraph/widgets/PenSelectorDialogbox.py
ntjess: Many of the widgets in pyqtgraph come from `QtGui`... Is this just old code? I thought it was for a reason, but I'm happy to transition to `QtWidgets` instead

pijyoi:
In Qt4, there was only QtCore and QtGui. In Qt5, QtGui was split into QtGui
and QtWidgets.

Pyqtgraph when adding Qt5 support chose to emulate the Qt4 interface by
monkey-patching all of QtWidgets into QtGui.

pyqtgraph/parametertree/parameterTypes.py
ntjess: I had more difficulty porting this over without problems, so I replaced it with a file parameter I've been using for about a 8 months without many issues. It supports almost all these features, but with (IMO) a bit smoother of a option interface.

pijyoi: ok, but you need to change the enums to the long spelling and also the exec_().
You should try running on PyQt6 6.1.1.

ntjess: Jeez, installed pyside6 and I see what you mean. The enums should be fixed there as well.

ntjess:
Ok to merge? Suggestions? Comments? I want to do some housekeeping on parameter types but also want to wait until this is resolved.

ixjlyons:
I haven't looked over the code in detail yet, but a few high-level notes from playing around with the example:

- The pen parameter should probably have the parameter name in col 0 and the button in col 1, and maybe a default button as well - it seems like it could be implemented as a `WidgetParameterItem`? I suppose the intent may have been that it's clear enough you're looking at a button with a "pen symbol" but in a GUI you might want multiple pen controls for different things.
- The pen doesn't show display with its default value at first - i.e. if you open the pen dialog and cancel, it will update to red (the default).
- The file picker would probably work best as the traditional line edit + "browse" button combo. Horizontal space in a parameter tree can get cramped, but having the file path in the push button isn't very effective.
- We'll need a doc string for each new item that I suppose should just highlight any non-standard options, then add the new parameters/items to `doc/source/parametertree/parameterTypes.rst`. At a glance, the file picker looks like the only one with new options.

ntjess:
@ixjlyons Thanks for the feedback. Do you recommend the current value of the file picker not show up anywhere visually? I'm ok with it and it's an easy change, I just wanted to be sure. Some alternatives are to just show the filename instead of full path or use ellipses (change elide mode on the label)

* Are you OK with the file API -- `existing`, `asFolder`, `fileFilter`, `selectMultiple`, `startDir`? I.e., should any names be changed for clarity?
  * Currently `fileFilter` requires the same format as Qt (e.g. `text files (*.txt);; ...`), should a different format be allowed / provisioned for easier use?
* Is `DontUseNativeDialog` common enough to expose at the parameter level? I know for me it caused some valuable differences in earlier Qt versions (<5.12), and people who want to add widgets to the filepicker cannot use the native choice.

j9ac9k:
> @ixjlyons Thanks for the feedback. Do you recommend the current value of the file picker not show up anywhere visually? I'm ok with it and it's an easy change, I just wanted to be sure. Some alternatives are to just show the filename instead of full path or use ellipses (change elide mode on the label)
> 
> * Are you OK with the file API -- `existing`, `asFolder`, `fileFilter`, `selectMultiple`, `startDir`? I.e., should any names be changed for clarity?

This is the kind of API nitpick I can get behind! 

* `asFolder`  I would advice using "Directory" instead of "Folder".  

Is there a reason we shouldn't match up the existing Qt API here and just use `directory` (instead of `startDir`), `filter` (instead of `fileFilter`) and so on?  There is likely a good idea why not, it's not the sort of thing I know much about unfortunately.

>   * Currently `fileFilter` requires the same format as Qt (e.g. `text files (*.txt);; ...`), should a different format be allowed / provisioned for easier use?

Unless there is demand for another format, I would stick w/ that for two reasons.

1. it should be familiar to people w/ Qt experience already
2. If someone learns about this format from pyqtgraph, they can apply it elsewhere in the Qt framework
3. Minimizes our support footprint

If there is a feature request to add another syntax, we can evaluate that on its own merit, but even if we come up with a potentially simpler format; there is likely no guarantee it would be simpler for enough use-cases to make it worthwhile.

> * Is `DontUseNativeDialog` common enough to expose at the parameter level? I know for me it caused some valuable differences in earlier Qt versions (<5.12), and people who want to add widgets to the filepicker cannot use the native choice.

`DontUseNativeDialog` should likely go away, I think it was introduced in this library as a work-around for some issue on macOS eons ago.

ntjess:
@j9ac9k

> Is there a reason we shouldn't match up the existing Qt API here and just use directory (instead of startDir), 

`directory` may refer to either starting directory or whether a directory should be selected, so I tried to disambiguate the two.

> filter (instead of fileFilter)

I tend not to use builtin functions as variables, which is why I stayed away from `filter` (but it makes sense here to match the Qt API).

> and so on? 

The other accepted arguments are boolean versions of enum flags which are easier to store in the `Parameter` state and slightly simpler (my initial reasoning). I could also accept `fileMode`/`options`/etc. argument with stringified versions of desired flags, i.e.

`Parameter.create(type='file', fileMode='ExistingFiles', options=['ShowDirsOnly', 'DontResolveSymlinks'], viewMode='Detail')`

Additionally, I could wait until #1120 is merged and use true bitmasks for these. FWIW, my initial reason for the boolean interface was simplicity and it might be intimidating to expose every aspect of a file dialog even if ~80% of it isn't used in common practice.

j9ac9k:
> @j9ac9k
> 
> > Is there a reason we shouldn't match up the existing Qt API here and just use directory (instead of startDir),
> 
> `directory` may refer to either starting directory or whether a directory should be selected, so I tried to disambiguate the two.

Ahh good point!
 
> > filter (instead of fileFilter)
> 
> I tend not to use builtin functions as variables, which is why I stayed away from `filter` (but it makes sense here to match the Qt API).

Oh, missed the part this was a variable name, yeah built-ins as variable names is a no-no, could do `filter_` (the same way qt bindings have navigated around `exec`), but I'm not strongly opinionated on this and am good w/ what you think is best.


> The other accepted arguments are boolean versions of enum flags which are easier to store in the `Parameter` state and slightly simpler (my initial reasoning). I could also accept `fileMode`/`options`/etc. argument with stringified versions of desired flags, i.e.
> 
> `Parameter.create(type='file', fileMode='ExistingFiles', options=['ShowDirsOnly', 'DontResolveSymlinks'], viewMode='Detail')`

Either option (list of bools, list of stringified enums) sounds good to me; probably should support default behavior, and put a link in the documentation to the Qt docs for more info.  

> * I could also override `saveState` to convert enums into string values before output. This means arguments like `options` can directly accept `Option.DontResolveSymlinks|Option.ReadOnly`. In general, though, I like the principle 'explicit is better than implicit' where the saved state is the same one input by the user when possible
> 
> FWIW, my initial reason for the boolean interface was simplicity and it might be intimidating to expose every aspect of a file dialog even if ~80% of it isn't used in common practice.

You've thought about this more than I have at this point, I'm good w/ it either way 👍🏻

ntjess:
@ixjlyons 
> The file picker would probably work best as the traditional line edit + "browse" button combo. Horizontal space in a parameter tree can get cramped, but having the file path in the push button isn't very effective.

Fair point, horizontal space is a premium in parameter trees. How about this? If you like it, I'll push the fix.

https://user-images.githubusercontent.com/23620506/123718154-6e070500-d84c-11eb-8c12-d93447db405e.mp4

ixjlyons:
That seems reasonable, definitely an improvement.

It still seems like it'd be useful to have the path in a line edit so it can be manually changed or copied/pasted. Maybe a button with just `...` to save space or maybe it could use `asSubItem` to span both columns?

I'm totally fine saving that as an enhancement idea for later.

ntjess:
Take 2


https://user-images.githubusercontent.com/23620506/123736678-918f7700-d86f-11eb-98a9-bbdb461a2a3e.mp4

I also took the liberty of making a slightly more interactive pen parameter :slightly_smiling_face: 

https://user-images.githubusercontent.com/23620506/123738633-2a73c180-d873-11eb-8bbd-6530558f5616.mp4

ntjess:
I'm realizing right after pushing that a line edit will not interact nicely with `ExistingFiles`, unless I use ast.literal_eval and go down that rabbit hole... Perhaps make it `readonly` if multiple selection is enabled?

There's also no input sanitation currently, i.e. if you edit directly you can violate all constraints set on the parameter. I don't quite know what the ideal solution is other than to leave it for now. Maybe make it readonly by default, and the adventurous user can enable it if they so choose?

ntjess:
With that, I think I've addressed all comments from the previous round. @ixjlyons and @j9ac9k thanks for your prior input, are you up for another go?

File picker is readonly by default and no sanitation or `completer` is provided (I'm considering it out of scope for now, someone can submit a PR if they wish). Those wanting to copy-paste can specify `readonly=False` in the opts at their own risk.

ntjess:
With Scipy sprints and the new release coming up I know this will get delayed for a bit, any recommendations on when I can bump again? I'm looking to do a `parameterTypes` restructuring PR to avoid every new type modifying the same file. However, it would be easiest to wait until after this is addressed.

Thanks :)

j9ac9k:
Hi @ntjess I think the sprints, at least on Sunday, would be a great time to review this PR actually.  @ixjlyons had an interest in revisiting the PR backlog, this issue, along w/ Nils's color palette one are some big PRs that would benefit from getting more eyes on IMO.  Given many of us will be there, I think we can get some real-time feedback and hopefully merge 🤞🏻

pyqtgraph/parametertree/parameterTypes.py
```diff
@@ -833,7 +833,16 @@ def popupFilePicker(parent=None, winTitle='', nameFilter='', directory=None, sel
...
     if relativeTo is not None:
-        fList = [os.path.relpath(file, relativeTo) for file in fList]
+        # Forward slashes are returned by qt, so preserve this convention regardless
+        # of the os. Since os.path will always prefer os-flavored separators, use pathlib to consistently
+        # return posix flavor
+        newFileList = []
+        for file in fList:
+            # pathlib and darwin have case sensitivity issues using relative_to, so use os.path for this
+            relative = os.path.relpath(file, relativeTo)
+            relativePosix = Path(relative).as_posix()
```
j9ac9k, line 845: for this, I think we can use `os.fsdecode(filename)`

j9ac9k: Can probably do the whole thing in a list comprehension

```python
newFileList = [os.fsdecode(os.path.relpath(file, relativeTo)) for file in fList]
```

examples/parametertree.py
```diff
@@ -145,11 +154,11 @@ def valueChanging(param, value):
         ch2.sigValueChanging.connect(valueChanging)
         
 
-
 def save():
     global state
     state = p.saveState()
-    
+QtCore.QTimer.singleShot(0, save)
```
j9ac9k, line 160: We had really weird issues involving `QTimer.singleShot()` (see #1605 )

Would suggest creating a timer, and explicitly calling `setSingleShot`  You can see an example here: https://github.com/pyqtgraph/pyqtgraph/pull/1605/commits/85e894dd73d21b479bcc0d98431e17ed88d583d3

pyqtgraph/parametertree/parameterTypes.py
```diff
@@ -741,3 +748,635 @@ class TextParameter(Parameter):
...
+        return self._value
+
+    def _retrieveFileSelection_gui(self):
+        curVal = self.param.value()
+        if isinstance(curVal, list) and len(curVal):
+            # All files should be from the same directory, in principle
+            # Since no mechanism exists for preselecting multiple, the most sensible
+            # thing is to select nothing in the preview dialog
+            curVal = Path(curVal[0])
+            if curVal.is_file():
```
j9ac9k, line 900: Given this (and the `.as_posix()` call) are the only places we're using pathlib module, maybe we should consider removing it and sticking with `os.path` instead?

```python
os.path.isfile(curVal[0])
```

Can use that here

pyqtgraph/parametertree/parameterTypes.py
```diff
@@ -741,3 +748,635 @@ class TextParameter(Parameter):
...
+                curVal = curVal.parent
+        useDir = curVal or str(os.getcwd())
```
j9ac9k, line 902: we can get rid of the `str()` casting, as `os.getcwd()` returns a string already

pyqtgraph/parametertree/parameterTypes.py
```diff
@@ -741,3 +748,635 @@ class TextParameter(Parameter):
...
+        opts = self.param.opts
+        startDir = Path(useDir).absolute()
```
j9ac9k, line 904: can replace this with

```python
startDir  = os.path.absolute(useDir)
```

pyqtgraph/parametertree/parameterTypes.py
```diff
@@ -741,3 +748,635 @@ class TextParameter(Parameter):
...
+        if startDir.is_file():
+            opts['selectFile'] = startDir.name
+            startDir = startDir.parent
+        if startDir.exists():
+            opts['directory'] = str(startDir)
```
j9ac9k, line 909: wrapping `str(Path)` can cause huge problems, if you decide to stick with pathlib, use Path.as_posix(), or for a path-or-string compatible method use `os.fsdecode(startDir)`

examples/parametertree.py
ntjess: Would you object to a `safeSingleShot` function in `functions.py` that spawns a `QTimer` this way, if the paradigm needs to be reused everywhere?

pyqtgraph/parametertree/parameterTypes.py
ntjess: Helpful feedback, thanks. I wish these tidbits were easier to find while perusing the internet (first example coming to mind is this should be in the pathlib docs!)

ntjess: The official examples even state if you need a string version, just use `str()`

pyqtgraph/parametertree/parameterTypes.py
ntjess: does `fsdecode` turn backslash into forward slash? Don't see it in the docs

ntjess: @pijyoi @ixjlyons Looking for some sort of consensus on this. Either unix-style paths can be preserved (at the expense of converting from `str` -> `Path` -> `str` (`Path(relative).as_posix()`), or the return value changes to os-flavored paths if `relativeTo` is supplied. This is because `os.path.normpath` converts to os-flavored. But, `pathlib` wouldn't need to be used.

I'm in favor of risking type conversion to provide consistent path behavior, but that's mainly because I haven't experienced any negative effects from `pathlib` /  `Path` usage yet. @j9ac9k mentioned some inconsistencies with how `pathlib` treats casing, but I think that might be for operations that *create* paths -- in contrast, this usage would just be directly representing a path. For instance, on Windows calling `str(Path(filename))` still preserves casing. I could be wrong, though. Not sure what the best tests would be.

examples/parametertree.py
pijyoi: Why is the delayed call to ```save()``` needed? You have the following right before ```pg.exec()```
```python
## test save/restore
state = p.saveState()
p.restoreState(state)
```

pijyoi:
While integrating PenParameter into #1910, I found that it returns a pen with isCosmetic()==False. Not sure how that is happening since the default is pg.mkPen().isCosmetic()==True.

examples/parametertree.py
ntjess: IIRC, there was an issue where the restored state didn't match the actual state before saving, so I delayed the initial save for things like the gradient editor's size to finalize. But testing it now, the `singleShot` seems unnecessary.

ntjess:
@pijyoi since `mkPen` is used everywhere except the initial pen creation, I will update that to be consistent everywhere. It does raise the question -- should `cosmetic` be a modifiable pen property? Aside from `brush` (which would be fairly complex to expose as a modifiable parameter), I don't think there are any other exposable properties (unless I made some mechanism for custom dash offsets, but I would call that out of scope as well)

j9ac9k:
I don't think we need a parameter for custom dash offsets, I would call that fairly out of scope....

In terms of `cosmetic` being a settable parameter, I can see a good case one way or the other, but I do think the default should be `True` as that's what we're dependent on in this library for most usages of QPen.

ntjess:
`cosmetic` and redundant `singleShot` should be fixed now, and I took the opportunity to also _compare_ the saved and restored states instead of just performing the operations.

j9ac9k:
next up, see if I can get to the bottom of this CI failure :| I'll try and poke around at that

ntjess:
> next up, see if I can get to the bottom of this CI failure :| I'll try and poke around at that

😆 It would be nice not to get a 'failure' email every time I push

ntjess:
~80% of the comments seem to be about the file picker, so I'm including a script here I used for testing to make it easier for everyone else, too. (Ironically, Github doesn't support uploading a `.py` file, lol)

```python
from pyqtgraph.parametertree.parameterTypes import QtEnumParameter
from pyqtgraph import parametertree as pt
import pyqtgraph as pg

app = pg.mkQApp()
dlg = QtWidgets.QFileDialog
optionsEnum = QtEnumParameter(dlg.Option)

children = [
  dict(name='options', type='group', children=[
    dict(name=kk, type='bool', value=False) for kk in optionsEnum.enumMap
  ]),
  dict(name='relativeTo', type='str'),
  dict(name='directory', type='str'),
  dict(name='windowTitle', type='str'),
  dict(name='nameFilter', type='str'),
  QtEnumParameter(dlg.AcceptMode),
  QtEnumParameter(dlg.FileMode),
  QtEnumParameter(dlg.ViewMode),
  QtEnumParameter(dlg.DialogLabel),

]
fileParam = pt.Parameter.create(name='file dialog', type='file')
param = pt.Parameter.create(name='file tester', type='group', children=[
  fileParam, *children
])

def onOptsChanged(param, _changes):
  useList = []
  for child in param.children():
    if child.value():
      useList.append(child.name())
  fileParam.setOpts(options=useList)

def onRestChanged(param, value):
  if isinstance(param, QtEnumParameter):
    # Need string rep
    for kk, vv in param.enumMap.items():
      if vv == value:
        value = kk
        break
  if param.type() == 'str' and not param.value():
    # Edge case for disabling string options
    value = None
  fileParam.setOpts(**{param.name(): value})

param.child('options').sigTreeStateChanged.connect(onOptsChanged)
for ch in param.children()[2:]:
  ch.sigValueChanged.connect(onRestChanged)
fileParam.sigValueChanged.connect(lambda p, val: print(val))
tree = pt.ParameterTree()
tree.setParameters(param)
tree.show()
pg.exec()
```

j9ac9k:
This PR in its current form has some merge conflicts; also I'm really concerned about the CI failure, I'll reach out over slack to see if we can diagnose that a bit; suppose first step is to see if we can replicate and go from there.

ntjess:
Rebased onto master, merge conflicts should be gone but the CI failure is still occurring. Is that linked to any of my changes? I can't reproduce locally on my windows machine so it's difficult for me to analyze

j9ac9k:
with what appears to be a segfault, causes like this can be really tough to isolate; even worse when it only happens in CI but not locally... best you can hope for is it fails in CI consistently 😆

j9ac9k:
what a rollercoaster this one was....

A lot of thanks for this PR.

Thank you @feketeimre for the original submission
Thank you @ntjess for updating/modernizing + bearing through the CI issue
Thank you to the reviewers of this PR from the SciPy sprints.

Squashing and merging this one!

PR #1845, opened 2021-06-19
Separate out mouse events stolen by AxisItem

### LINKED ISSUE #1843: Regression: modified zoom behavior with showGrid
### Short description

Enabling the grid on a `PlotItem`causes zoom on the `ViewBox` to affect only the y axis when it should affect both. Zooming by scolling on the x axis still works. A more severe side-effect can be seen with `setAspectLocked`.

Bisect points to #1726 - I haven't looked further into it yet, but this should probably hold up 0.12.2 in my view.

### Code to reproduce

```python
import pyqtgraph as pg

app = pg.mkQApp()

p = pg.plot([1, 2, 3], [4, 5, 6])
p.showGrid(x=True, y=True)
#p.setAspectLocked(True)

app.exec()
```

### Tested environment(s)

 * PyQtGraph version: 30bc138
 * Python version: 3.9.5
 * Operating system: Linux

NilsNemitz:
While it graphically makes sense to draw the AxisItem on top of what is plotted, particularly for images, this has unintended side-effects:
If AxisItem is placed above the ViewBox, which places itself at z order -100, then it will grab mouse events inside its bounding box. And this bounding box overlaps the plot. It overlaps *all* of the plot when the ticks are extended to draw a grid.

This PR adds a check to the handling of mouse events in AxisItem:
If the (wheel event) or (button down) occurred within the bounds of the linked ViewBox, then the ViewBox gets to handle the event as if it had been assigned to it directly. 

A good test for this is to run the "GraphicsItems / Scatter Plot" example and add a grid by including
```w3.showGrid(x=True, y=True)``` at the end of example (3):
Mouse wheel zoom and drag events should still behave differently when performed on the axes vs. plot, even with the grid drawn on top. Mouse clicks also still get properly directed to the individual points of the ScatterPlotItem.

While looking at this, I came across the line:
```python 
p.translate(0.5,0)  ## resolves some damn pixel ambiguity 
```
in the `drawPicture()` method (L.1154) of AxisItem.

Does anybody have some context for that? Are we drawing all of our axis ticks offset half a pixel to the right?
Is this compensated for in generating the tick specs? I did not see it.
And why does it affect only the x direction?

I am tempted to remove the translation...

Closes #1843.

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -1188,11 +1188,12 @@ class AxisItem(GraphicsWidget):
         lv = self.linkedView()
         if lv is None:
             return
+        # Did the event occur inside the linked ViewBox (and not over the axis iteself)?
         if lv.sceneBoundingRect().contains(event.scenePos()):
-            print('in')
+            # pass event to linked ViewBox without marking it as single axis zoom
             lv.wheelEvent(event)
         else:
-            print('out')
+            # pass event to linked viewbox with appropriate single axis zoom parameter
             if self.orientation in ['left', 'right']:
                 lv.wheelEvent(event, axis=1)
             else:
@@ -1203,8 +1204,11 @@ class AxisItem(GraphicsWidget):
         lv = self.linkedView()
         if lv is None:
             return
+        # Did the mouse down event occur inside the linked ViewBox (and not the axis)?
         if lv.sceneBoundingRect().contains(event.buttonDownScenePos()):
+            # pass event to linked ViewBox without marking it as single axis pan
             return lv.mouseDragEvent(event)
+        # otherwise pass event to linked viewbox with appropriate single axis parameter
         if self.orientation in ['left', 'right']:
             return lv.mouseDragEvent(event, axis=1)
         else:
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -1188,9 +1188,11 @@ class AxisItem(GraphicsWidget):
         lv = self.linkedView()
         if lv is None:
             return
-        if lv.geometry().contains(event.scenePos()):
+        if lv.sceneBoundingRect().contains(event.scenePos()):
+            print('in')
             lv.wheelEvent(event)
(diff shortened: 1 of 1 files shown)


j9ac9k:
I don't have context for that translation, but there have been a ton of new Qt versions since that was added so I can imagine this is no longer an issue.  I would encourage removing it and going through all the examples.

NilsNemitz:
I had a look and did not see anything. Another pair of eyes would be welcome, though.
Here is what we are looking for:
20210620 offset -10
This is the Crosshair example with the translation command in question changed
 to `p.translate(-10,0)`
This clearly shifts all labels and ticks visibly to the left, for both x and y axis.
Note that the crosshair is at 1183, but points at the 1200 tick.

If `p.translate(0.5, 0)` is needed for alignment, then removing it should leave us with tick marks similarly offset to the left, but only by 0.5 pixels. I'd expect that the "pixel ambiguity" refers to this rounding wrong under some circumstances, so that ticks show up too far left (by one pixel).  I did not see any evidence for that throughout the examples.

In the crosshair example, both settings (0.5 and no translation) appear identical:
0.5 pixel horizontal shift:
20210620 offset 0 5_zoom

no shift:
20210620 offset 0_zoom

j9ac9k:
@NilsNemitz looking at that screenshot (not the zoomed in one) the left axis has some misaligned digits, is this some artifact from the screenshot or does it actually look like that?

NilsNemitz:
Dear @j9ac9k , the normally scaled screenshot replaces the +0.5 translation with a -10.0 translation.
Basically, it shows the changes expected for removing the statement, but magnified by a factor of 21.

We see that this moves all tick marks and their labels to the left, while leaving the axis lines intact.
And it appears that Qt has some interesting breakage when the labels try to translate offscreen! I am pretty sure that the screenshot represents what was actually shown on screen.

I eliminated the translation and looked for similar issues throughout the examples, but I did not manage to find any.

j9ac9k:
> I eliminated the translation and looked for similar issues throughout the examples, but I did not manage to find any.

Good enough for me.

j9ac9k:
LGTM, thanks for renaming `ev` to `event`.  Merging!

Issue #1850, opened 2021-06-21
Regression: empty color bar


pijyoi:
### Short description

Empty color bar gets plotted instead of Viridis

### Code to reproduce

Example from #1849

### Expected behavior

Expected behavior from #1849 (Viridis color bar)

### Real behavior
![bad_color](https://user-images.githubusercontent.com/2657027/122764172-c597f480-d2d1-11eb-88a4-0407982b1833.png)

### Tested environment(s)

 * PyQtGraph version: master ba517aba0e3dbddb925fcb47f9a0359432fd4a10
 * Qt Python binding: PyQt5 5.15.4 
 * Python version: 3.8.10
 * NumPy version: 1.20.3 
 * Operating system: Windows 10 20H2
 * Installation method: pip install -e .

### Additional context
bisect points to 7d41e8a8789a7037c061f08c6e6e633ba94c4ee3

NilsNemitz:
Dear @pijyoi , dear @paulmueller ,

I am looking at that now.
My current understanding is that the plot shown above is in fact the "expected", documented behavior.
Even if it certainly isn't the desired one.

`stops, colors = cmap.getStops('float')`
tries to get a float description of the colors, the first one is
[0.26666667 0.00392157 0.32941176 1.        ]

A little later
`pg.QtGui.QColor(*[c for c in color]` feeds these into the QColor constructor, which expects integer values 0-255.

I think the plot above is a viridis gradient at opacity 1 out of 255.

The previous code was this:
```python
def getStops(self, mode):
        ## Get fully-expanded set of RGBA stops in either float or byte mode.
        if mode not in self.stopsCache:
            color = self.color
            if mode == self.BYTE and color.dtype.kind == 'f':
                color = (color*255).astype(np.ubyte)
            elif mode == self.FLOAT and color.dtype.kind != 'f':
                color = color.astype(float) / 255.
            self.stopsCache[mode] = (self.pos, color)
        return self.stopsCache[mode]
``` 
And since that did not recognize the 'float' identifier, it silently returned data in the default byte format.

A hotfix to the example code would be to change line 14 to
`stops, colors = cmap.getStops(pg.ColorMap.BYTE)`
which seems to produce the desired color gradient on my system.

I'll look into making a PR that supports string input and probably defaults to the more typical byte format return.

pijyoi:
Closing as issue was due to the example code relying on undefined behaviour that happened to work in older versions of the library.

Issue #1849, opened 2021-06-21
SVG export: linearGradient not exported correctly


paulmueller:
I am using a custom colorbar (derived from https://gist.github.com/maedoc/b61090021d2a5161c5b9) to show a simple colorbar next to my plots. When exporting the colorbar as SVG (PNG works), the gradient is not shown correctly. The [`linearGradient`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/linearGradient) tag in the resulting SVG file contains the right `steps`, but there is some issue with the way it is scaled, probably due to a bad usage of  [`gradientUnits="userSpaceOnUse"`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/gradientUnits).

### Short description

On SVG export, the linearGradient is not set correctly. This looks like a bad combination of [`gradientUnits="userSpaceOnUse"`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/gradientUnits) and SVG coordinates.

### Code to reproduce

This is a minimal working example. Run the script and export the entire scene as SVG.
```python
import numpy as np
import pyqtgraph as pg
from pyqtgraph.graphicsItems.GradientEditorItem import Gradients



class ColorGradient(pg.GraphicsObject):

    def __init__(self, cmap, width, height):
        pg.GraphicsObject.__init__(self)

        # handle args
        w, h = width, height
        stops, colors = cmap.getStops('float')
        smn, spp = stops.min(), stops.ptp()
        stops = (stops - stops.min())/stops.ptp()

        # setup picture
        self.pic = pg.QtGui.QPicture()
        p = pg.QtGui.QPainter(self.pic)

        # draw bar with gradient following colormap
        p.setPen(pg.mkPen('k'))
        grad = pg.QtGui.QLinearGradient(w/2.0, 0.0, w/2.0, h*1.0)
        for stop, color in zip(stops, colors):
            grad.setColorAt(1.0 - stop, pg.QtGui.QColor(*[c for c in color]))
        p.setBrush(pg.QtGui.QBrush(grad))
        p.drawRect(pg.QtCore.QRectF(0, 0, w, h))

        # done
        p.end()

        
    def paint(self, p, *args):
        # paint underlying mask
        p.setPen(pg.QtGui.QColor(255, 255, 255, 0))
        p.setBrush(pg.QtGui.QColor(255, 255, 255, 200))
        
        # paint colorbar
        p.drawPicture(0, 0, self.pic)
[...]

PR #1853, opened 2021-06-21
make getColors and getStops behavior more consistent; add tests

NilsNemitz:
This PR seeks to address the not very well-defined behavior of the `getColors` and `getStops` methods of ColorMap, which lead to problems such as exhibited in #1850.

The issue there seems to be that an undefined mode parameter to the `getStops()` method used to return the internal color values "as stored", which in this case used to be in byte format.

Returning the data in an undefined internal format seems like a bad idea:
E.g. restoring the originally intended hsv interpolations or other color space conversions benefits from a change to a higher precision internal float representation. The performance cost of converting to the user-desired format is also small, since `getStops()` already implements a caching method.

This PR tries to update the `getColors()` and `getStops()` methods as follows:
- default return format is byte, which would have been the previous default in most cases I can see, and works best with QColor initialization (see the example code in #1849).
- `getColors()` now internally calls `getStops()` to reduce code duplication and to make use of the existing cache.
- both methods can now also return colors as a list of QColors, which was already implemented, but not documented. I believe it may have worked only when the internal representation was byte.
- both methods now support string specifiers 'byte', 'float' and 'qcolor' in addition to the integer enumeration values. Unknown method strings will raise an error. 

The PR also adds some basic tests for these format conversions; more tests for the code in colormap.py seems needed, though.

### DIFF
--- a/pyqtgraph/colormap.py
+++ b/pyqtgraph/colormap.py
@@ -5,6 +5,8 @@ from os import path, listdir
 from collections.abc import Callable, Sequence
 import warnings
 
+__all__ = ['ColorMap']
+
 _mapCache = {}
 
 def listMaps(source=None):
@@ -598,32 +600,44 @@ class ColorMap(object):
         pen.setCosmetic(True)
         return pen
 
-    def getColors(self, mode=None):
-        """Returns a list of all color stops, converted to the specified mode.
-        If `mode` is None, no conversion is performed.
+    def getColors(self, mode=BYTE):
+        """
+        Returns a tuple (stops, colors) containing a list of all stops (ranging 0.0 to 1.0)
+        and a list of the associated colors.
+        
+        The parameter `mode` can be one of
+        - `ColorMap.BYTE` or 'byte' to return colors as RGBA tuples in byte format (0-255)
+        - `ColorMap.FLOAT` or 'float' to return colors as RGBA tuples in float format (0.0 to 1.0)
+        - `ColorMap.QCOLOR' or 'qcolor' to return a list of QColors
+        The default is byte format.
+        """
+        stops, color = self.getStops(mode=mode)
+        return color
+
+    def getStops(self, mode=BYTE):
+        """
+        Returns a tuple (stops, colors) containing a list of all stops (ranging 0.0 to 1.0)
+        and a list of the associated colors.
+        
+        The parameter `mode` can be one of
+        - `ColorMap.BYTE` or 'byte' to return colors as RGBA tuples in byte format (0-255)
+        - `ColorMap.FLOAT` or 'float' to return colors as RGBA tuples in float format (0.0 to 1.0)
+        - `ColorMap.QCOLOR' or 'qcolor' to return a list of QColors
+        The default is byte format.
         """
         if isinstance(mode, str):
             mode = self.enumMap[mode.lower()]
 
-        color = self.color
-        if mode in [self.BYTE, self.QCOLOR] and color.dtype.kind == 'f':
-            color = (color * 255).astype(np.ubyte)
-        elif mode == self.FLOAT and color.dtype.kind != 'f':
-            color = color.astype(float) / 255.
-
-        if mode == self.QCOLOR:
-            color = [QtGui.QColor(*x) for x in color]
-
-        return color
-
-    def getStops(self, mode):
-        ## Get fully-expanded set of RGBA stops in either float or byte mode.
         if mode not in self.stopsCache:
             color = self.color
             if mode == self.BYTE and color.dtype.kind == 'f':
                 color = (color*255).astype(np.ubyte)
             elif mode == self.FLOAT and color.dtype.kind != 'f':
                 color = color.astype(float) / 255.
+            elif mode == self.QCOLOR:
+                if color.dtype.kind == 'f':
+                    color = (color*255).astype(np.ubyte)
+                color = [QtGui.QColor(*x) for x in color]
             self.stopsCache[mode] = (self.pos, color)
         return self.stopsCache[mode]
 
--- a/pyqtgraph/colormap.py
+++ b/pyqtgraph/colormap.py
@@ -602,13 +602,13 @@ class ColorMap(object):
 
     def getColors(self, mode=BYTE):
         """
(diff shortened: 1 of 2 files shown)


j9ac9k:
This change LGTM, does this fix #1850 ?

pijyoi:
#1850 was due to a bug in the user code in that they were asking for 'float' but using them as byte.

j9ac9k:
@pijyoi unrelated to this issue, but pyqtgraph will be participating in the sprints at the SciPy conference on July 17th (and 18th?, waiting on confirmation from them).  Most of the maintainers will be there, if you're available, we'd love you to join in.

pijyoi:
>pyqtgraph will be participating in the sprints at the SciPy conference on July 17th (and 18th?) ... if you're available, we'd love you to join in.  

Sounds interesting. Should be able to make it.

j9ac9k:
> Sounds interesting. Should be able to make it.

That's great.  If there is an issue/area you think should get some attention; or you'd like to collaborate with some people on, let me know and I'll be sure bring attention to it somehow.

j9ac9k:
Thanks @NilsNemitz Merging!

pijyoi:
@NilsNemitz , while trying to learn how to use ```getGradient()```, I noticed that ```getGradient()``` calls ```getStops(BYTE)``` instead of ```getStops(QCOLOR)```. The result is that QColor stops don't get cached in ```self.stopsCache```.

NilsNemitz:
Hi @pijyoi ,

Hmm. I guess getGradient should be calling `getStops(QCOLOR)`, since it seems that this what it really wants to have:
```python
pos, color = self.getStops(mode=self.BYTE)
color = [QtGui.QColor(*x) for x in color]
```
I'll immediately agree that that should be improved. Changing this to 
```python
pos, color = self.getStops(mode=self.QCOLOR)
```
might be sufficient? That would avoid unneeded conversions and generate the cache entry.

As implemented now, getStops never needs to deliver QColors to create a gradient. Then it seems consistent with the current strategy that these are not in the cache? It seems that calls to the mapping functions will also not put the QColors in the cache. It makes a bit more sense there, since the mapping relies on interpolating RGB values, and -I think- wouldn't benefit from starting from QColors. 

There's an unrelated point here: It might be better to have QColor interpolations work with the float data. The current code will always quantize to bytes, even if the original data is stored as higher resolution floats. 

Let me know if I missed your main point regarding the caching :)

pijyoi:
>I'll immediately agree that that should be improved. Changing this to
pos, color = self.getStops(mode=self.QCOLOR)
might be sufficient? That would avoid unneeded conversions and generate the cache entry.

Yes, this is what I meant.
Although in actual usage, the user should probably cache themselves the resulting ```QLinearGradient```. I did that in https://github.com/pyqtgraph/pyqtgraph/pull/1711#issuecomment-887146820. I want the colormap in ```QLinearGradient``` form in order to pass it directly to ```QPainter.fillRect()```

On the other hand, it's kind of hard for the user to know whether a particular method is considered cheap or expensive (and hence should perform their own caching)

>There's an unrelated point here: It might be better to have QColor interpolations work with the float data. The current code will always quantize to bytes, even if the original data is stored as higher resolution floats.

I see the code that you are referring to. But you are the one who wrote it all, so you would know best.

Regarding the line ```color = [QtGui.QColor(*x) for x in color]```, if the object being iterated over is a 2D-ndarray, I have found that it is faster to do ```color = [QtGui.QColor(*x.tolist()) for x in color]```
and if the object is small-ish, even ```color = [QtGui.QColor(*x) for x in color.tolist()]```

```python
In [11]: lut = pg.colormap.get('cividis').getLookupTable()

In [12]: %timeit [QtGui.QColor(*x) for x in lut]
1.9 ms ± 23.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [13]: %timeit [QtGui.QColor(*x.tolist()) for x in lut]
1.1 ms ± 120 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [14]: %timeit [QtGui.QColor(*x) for x in lut.tolist()]
933 µs ± 34.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
i.e. most of the benefits are obtained from the 2nd form, without having the doubled memory usage of the 3rd form.

NilsNemitz:
Dear @pijyoi ,
Hehe... All I have done is stick another level of decorations on this and then tried to clean up the code just enough so that is doesn't collapse into a pile of toothpicks.

I looked at working with float colors again: I still think it would be the most proper way to handle this. But as far as I can tell, there is no single-line constructor for QColors that takes float rgb values. And I don't know enough about python's inner workings to figure out an efficient way of looping over
`qcol = QColor; qcol.setRgbF(*values)` in the code above.

So I'd leave it as it is; at least for now.

I put the other suggestions into PR #1927. Further comments are very welcome.
I think that it is probably sufficient to cache the generated gradient on the user side, unless you think this will find performance critical use in multiple places. To just draw with a gradient pen, I'd expect that the most natural use would be to put the gradient pen in a variable, and then pass that to the plot function. In that case, reuse is straightforward and hopefully obvious.

pijyoi:
>there is no single-line constructor for QColors that takes float rgb values

There is a static factory function though ```QColor.fromRgbF()```. Probably there were too many constructor overloads.
```python
  elif mode == self.QCOLOR:
      if color.dtype.kind == 'f':
          color = [QtGui.QColor.fromRgbF(*x.tolist()) for x in color]
      else:
          color = [QtGui.QColor(*x.tolist()) for x in color]
```

Internally, PySide6 seems to always call ```fromRgbF()```
```python
In [48]: QtGui.QColor(51, 85, 170)
Out[48]: PySide6.QtGui.QColor.fromRgbF(0.200000, 0.333333, 0.666667, 1.000000)
In [49]: QtGui.QColor.fromRgb(51, 85, 170)
Out[49]: PySide6.QtGui.QColor.fromRgbF(0.200000, 0.333333, 0.666667, 1.000000)
In [50]: QtGui.QColor.fromRgbF(0.2, 1/3, 2/3)
Out[50]: PySide6.QtGui.QColor.fromRgbF(0.200000, 0.333333, 0.666667, 1.000000)
```

PR #1861, opened 2021-06-23
progress dialog fix

outofculture:
Looks like this broke in d455da9aeca73d25ae75c4fc1db4fa691db9b30d so I added a test this time.


j9ac9k:
Blerg, ... keep thinking I got them all.

Thank you for adding a test!

j9ac9k:
Blerg, ... keep thinking I got them all.

Thank you for adding a test!

I think its supposed to be in the middle:

```python
        self.setWindowModality(QtCore.Qt.WindowModality.WindowModal)
```

PR #1862, opened 2021-06-24
Deprecate MetaArray Entirely (for 0.14 release)

j9ac9k:
We've had some internal discussions about this, but basically MetaArray is not used elsewhere within the library, it's not tested, and I suspect [xarray](https://xarray.pydata.org/en/stable/) could likely fulfill the roll better.

Doing a github advanced search for MetaArray, I'm showing some other repositories that have import statements for it, but didn't see much actual usage.

If you are dependent on MetaArray, and this deprecation will affect you, please open a discussion and get a conversation started, we want to hear from you.

### DIFF
--- a/pyqtgraph/metaarray/MetaArray.py
+++ b/pyqtgraph/metaarray/MetaArray.py
@@ -132,6 +132,10 @@ class MetaArray(object):
   
     def __init__(self, data=None, info=None, dtype=None, file=None, copy=False, **kwargs):
         object.__init__(self)
+        warnings.warn(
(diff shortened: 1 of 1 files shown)


outofculture:
Notably, as one of the users of this class, I will be extracting MetaArray to its own library (while also considering moving to xarray).

j9ac9k:
> Notably, as one of the users of this class, I will be extracting MetaArray to its own library

Let me know if you'd like help with this (not the xarray evaluation, but the package redistribution).

I also don't foresee 0.14 coming any time soon, I suspect we're looking at a year-is before it's actually removed from the library.

Issue #1591, opened 2021-02-19
GradientLegend


Breviel:
### Short description
GradientLegend generates the legend upside down

### Code to reproduce

```python
from PyQt5 import QtCore, QtGui
from PyQt5.QtWidgets import QWidget, QApplication
import pyqtgraph as pg
import sys

class AnalysePlot(QWidget):

    def __init__(self, parent=None):
        QWidget.__init__(self, parent)

        self.setWindowTitle('Model parameter endring')
        self.setAutoFillBackground(True)
        self.setStyleSheet("color: yellow;"
                           "background-color: rgba(0, 0, 0, 255);")

        self.view = pg.GraphicsView(self)
        self.layout = pg.GraphicsLayout(border=(100, 100, 100))
        self.view.setCentralItem(self.layout)
        self.view.show()
        self.view.resize(640, 480)
        self.layout.setSpacing(5)

        self.p1 = self.layout.addPlot(title="Parametrene")
        self.p1.setFixedWidth(620)
        self.p1.setFixedHeight(460)
        self.p1.showGrid(1, 1, 0.3)

        self.gradLeg = pg.GradientLegend(size=(25, 100), offset=(-15, 15))

        self.updateGraph()

    def updateGraph(self):
        self.p1.clear()

        startParams = [-721, 617, 0, -666, 0, 446, 2586, 0, 4, 0, 0, -210, 2722, 165, 12, 0, 588, 0, 140, -689]
        params = [-719, 616, -11, -672, -17, 447, 2605, 2, 8, -1, 0.0, -209, 2722.0, 165.0, 12.0, 0.0, 588.0, 0.0,\
                          140.0, -689.0]
        arrayDif = []

        grad = QtGui.QLinearGradient(QtCore.QPointF(0, -20), QtCore.QPointF(20, 20))
        grad.setColorAt(0.0, pg.mkColor('r'))
        grad.setColorAt(0.125, pg.mkColor('#FF9933'))
        grad.setColorAt(0.25, pg.mkColor('y'))
        grad.setColorAt(0.375, pg.mkColor('g'))
        grad.setColorAt(0.5, pg.mkColor('b'))
        grad.setColorAt(0.625, pg.mkColor('g'))
        grad.setColorAt(0.75, pg.mkColor('y'))
        grad.setColorAt(0.875, pg.mkColor('#FF9933'))
        grad.setColorAt(1.0, pg.mkColor('r'))
[...]

PR #1864, opened 2021-06-24
Draw GradientLegend in ViewBox coordinates

### LINKED ISSUE #1847: GradientLegend broken in y-inverted viewbox
### Short description
If GradientLegend is placed in a ViewBox with inverted Y axis, the text is flipped vertically. If the y-axis is not inverted, then the text is upright but the rest of the item is upside-down. This changed was introduced in [5fa3901c0](https://github.com/pyqtgraph/pyqtgraph/commit/5fa3901c00f4ffb9cc9951d7cac4af02fb56a497). Prior to that, a similar issue was present, but only for viewboxes that were _not_ inverted (apparently this had only ever been used in y-inverted views, which is common when working with images).

### Code to reproduce
```
import pyqtgraph as pg
p1 = pg.plot(title='inverted')
l1 = pg.GradientLegend([25, 300], [-20, -30])
l1.pen = pg.mkPen(color='w', width=1, cosmetic=True)
p1.addItem(l1)
p1.invertY()
p2 = pg.plot(title='normal')
l2 = pg.GradientLegend([25, 300], [-20, -30])
l2.pen = pg.mkPen(color='w', width=1, cosmetic=True)
p2.addItem(l2)
```

### Expected behavior
item placement and visual appearance should be the same regardless of whether `p.invertY()` is called. The legend item should appear in the upper-right, and the max value (red) should be at the top.

NilsNemitz:
This is the second attempt to fix GradientLegend to draw in the correct orientation independent of axis settings.

The current implementation only looks at the pixel size to determine the drawing size. As a result, it loses information on the orientation of the coordinate system. Inverting the y-axis then causes the vertical orientation to flip. See #1591 and #1847.

Inverting the x axis likewise results in a horizontal flip, but nobody has tried that yet.

This PR tries to address the issue by skipping the scaling to scene coordinates, and instead drawing directly in viewport coordinates. This seems to fix the issue on my system, and appears to even work with windows screen scaling set to 125%.
It also mostly seems to export correctly. (With the exception of all SVG fonts appearing as MSShellDiag on my system...)

The getViewBox() code of UIGraphicsItem implies that no view box might be available under some circumstances.
In that case, the code falls back to a dummy calculation. If someone has a test case for this, please let me know.

I have also taken the liberty to move the GradientLegend default z position to 100, so that it draws over a default plot, rather than under it.

Here is some test code that draws a GradientLegend on a plot with different axis orientations:
```python
import numpy as np
from pyqtgraph.Qt import QtCore, QtGui, QtWidgets, mkQApp
import pyqtgraph as pg

class MainWindow(pg.GraphicsLayoutWidget):
    """ example application main window """
    def __init__(self):
        super().__init__()
        self.resize(900,600)
        layout = self # we are using a GraphicsLayoutWidget as main window for convenience

        xdata = np.array( [0.0, 0.2, 1.0] )
        ydata = np.array( [0.0, 0.8, 1.0] )
        for r_idx in range(2):
            for c_idx in range(2):
                p = layout.addPlot(row=r_idx, col=c_idx)
                if c_idx > 0: p.invertX()
                if r_idx > 0: p.invertY()
                leg = pg.GradientLegend([25, 150], [15, 10])

                leg.setColorMap( pg.colormap.get('viridis') )
                leg.setLabels({'val0': 0., 'val1/2': 0.5, 'val1': 1.})
                leg.brush = pg.mkBrush(color='#33001180')
                leg.pen = pg.mkPen(color='#CCCCFFAA', width=1, cosmetic=True)
                leg.textPen = pg.mkPen(color='#CCCCFF', width=1, cosmetic=True)

                p.addItem( leg )
                p.addItem( pg.PlotCurveItem( x=xdata, y=ydata, pen='y' ) )
        self.show()

mkQApp("gradientlegend test")
main_window = MainWindow()

## Start Qt event loop
if __name__ == '__main__':
    mkQApp().exec()
```

Old output on the left <------> New output on the right :

GradientLegend orientation

Closes #1847


NilsNemitz:
Since elements are now drawn in fixed sizes, the code could be further improved to avoid recreating the drawn elements for each paint, but the current approach seemed least likely to break anything.

pyqtgraph/graphicsItems/GradientLegend.py
```diff
@@ -45,14 +46,17 @@ def setLabels(self, l):
...
-        p.scale(unit[0], -unit[1])
 
+        p.save() # save painter state before we change transformation
+        view = self.getViewBox()
+        if view is not None:
+            trans = view.sceneTransform()
+            p.setTransform( trans ) # draw in ViewBox pixel coordinates
+            rect = view.rect()
+        else:
+            rect = QtCore.QRectF(0,0, 100, 100) # use dummy values when no associated ViewBox exists.
```
j9ac9k, line 57: probably should get rid of this line?

NilsNemitz: Dear @j9ac9k , that's a good point.
When I added that, I thought there would still be some productive operations happening later in the code, but after cleaning up the transformations, that isn't really the case.
So the solution is probably to simply return from paint call early when there is no view box. With the dummy values plugged in, nothing useful would end up drawing anyway.

I changed to code to that approach now.

campagnola:
Thanks Nils! For what it's worth, I wouldn't have expected anyone to know the prior expected behavior of this code; there were no tests or examples and a non-obvious use case.

When I run the example code in #1847, I see the legend showing up at bottom-right in the y-inverted view (I think it should be top-right), and in the Y-not-inverted view, it doesn't appear at all. Ideally it'd look the same in either case. It's not obvious how to get there, though, so if I get some time, I'll have a look at this later in the week.

NilsNemitz:
Dear @campagnola , thank you for testing!
I am at a little bit of a loss here, since on my system I seem to get what I was expecting to get:
gradient legend test
I also tried running the test with screen scaling set to something other than 100%, but I still see the legend in both plots.

If you run my example above, does that also misplace the legend for some subset of orientations?


Regarding the positioning:
In the way that I mentally parsed the offset logic, the offset is the legend corner position in screen coordinates. Since that typically counts right and down, I would think that (10, 10) would place the top left corner of the legend ten pixels right and ten pixels down from the top left corner of the plot.
Negative numbers then start on the opposite side of the plot and count (in the opposite direction) to the opposite edge of the legend. So I would argue that your example of  [-20, -30] _should_ place the bottom right corner of the legend -20, -30 pixels from the bottom right corner of the plot... Which is what it seems to be doing.

Now, I _will_ admit that the color bar is currently upside down, with the minimum at the top. But I do know how to fix that. :)

NilsNemitz:
In the hope that we'll figure out how to make this work on all systems, this commit fixes the orientation of the colorbar.

It also adds the option of populating the bar gradient from a ColorMap object and moves background and text colors to variables rather than hard-coding them.

pyqtgraph/graphicsItems/GradientLegend.py
```diff
@@ -15,17 +15,25 @@ def __init__(self, size, offset):
         self.offset = offset
         UIGraphicsItem.__init__(self)
         self.setAcceptedMouseButtons(QtCore.Qt.MouseButton.NoButton)
-        self.brush = QtGui.QBrush(QtGui.QColor(200,0,0))
+        self.brush = QtGui.QBrush(QtGui.QColor(255,255,255,100)) # background color
```
NilsNemitz, line 18: self.brush was previously never used, This now holds the brush that draws the legend background.

pyqtgraph/colormap.py
```diff
@@ -104,7 +104,6 @@ def _getFromFile(name):
         else:
             csv_mode = False
         for line in fh:
-            name = None
```
NilsNemitz, line 107: This is accidentally left over from the original idea of naming individual colors. Removing this line ensures that the returned color map is named correctly and not None).

j9ac9k:
I'm running the example code #1847 on this PR on macOS on this PR and get the following result:

```python
import pyqtgraph as pg

app = pg.mkQApp()
p1 = pg.plot(title='inverted')
l1 = pg.GradientLegend([25, 300], [-20, -30])
l1.pen = pg.mkPen(color='w', width=1, cosmetic=True)
p1.addItem(l1)
p1.invertY()
p2 = pg.plot(title='normal')
l2 = pg.GradientLegend([25, 300], [-20, -30])
l2.pen = pg.mkPen(color='w', width=1, cosmetic=True)
p2.addItem(l2)

pg.exec()
```

image

the image looks the same if I comment out the `p1.invertY()` bit.   Regarding the positioning, I think this PR follows the positioning of `LegendItem`, if you edit `examples/Legend.py` to make the offset `[-20, -30]` it is put in the bottom right corner.

j9ac9k:
Oddly enough; I think the MWE showed a different unrelated bug, had to move the `p1.invertY()` to be earlier in the example to get an actual inverted axis.

```python
import pyqtgraph as pg

app = pg.mkQApp()
p1 = pg.plot(title='inverted')
p1.invertY()
l1 = pg.GradientLegend([25, 300], [-20, -30])
l1.pen = pg.mkPen(color='w', width=1, cosmetic=True)
p1.addItem(l1)
p2 = pg.plot(title='normal')
l2 = pg.GradientLegend([25, 300], [-20, -30])
l2.pen = pg.mkPen(color='w', width=1, cosmetic=True)
p2.addItem(l2)

pg.exec()
```

image

campagnola:
Ok, I figured out why the second legend wasn't showing up for me: when I pasted the test code into my terminal, I forgot to add an extra return to execute the last line :D 

This looks good to me. I agree with your interpretation on how the position argument should work. It does change the behavior, but it's a pretty benign change that will be easy for any users (if there are any besides me) to fix.

j9ac9k:
> Ok, I figured out why the second legend wasn't showing up for me: when I pasted the test code into my terminal, I forgot to add an extra return to execute the last line :D

that would do it!  Thanks for following up @campagnola and thanks for the PR @NilsNemitz this LGTM, merging.

PR #1869, opened 2021-06-26
Disable unused ColorBarItem ticks

### LINKED ISSUE #1866: ColorBarItem does not draw ticks correctly when exporting scaled image
### Short description

I've spotted two issues with the new ColorBarItem that involves the ticks of the AxisItems when the plot is exported at a resolution different from the screen resolution:

1. The ticks of the not-used axes are drawn as 1-px wide dots
2. The ticks of the used axis (the color bar ticks) are not aligned with the colorbar

This issue is present for both PNG and SVG exporters (did not test the others), so it's probably not exporter-related.

### Code to reproduce

Run the following code, then export the entire plot as PNG and make sure to choose a resolution that is twice the original resolution (e.g. width=1200).

```python
import numpy as np
import pyqtgraph as pg

pg.setConfigOption("background", "w")
pg.setConfigOption("foreground", "k")


if __name__ == '__main__':
    app = pg.mkQApp()

    pw = pg.plot()

    cm = pg.ColorMap(*zip(*Gradients["viridis"]["ticks"]))
    
    # use ColorBarItem
    bar = pg.ColorBarItem(cmap=cm, values=(0,20),interactive=False)
    pw.plotItem.layout.setColumnFixedWidth(4, 5) # enforce some space to axis on the left
    pw.plotItem.layout.addItem(bar,2,5)

    app.exec()
```

### Expected behavior

There should only be ticks on the right side and the ticks should not extend outside of the colorbar.

### Real behavior

1. The top, left, and bottom axes show axes ticks. (red arrows)
2. The right axis extends out of the axis (blue arrows)
(zoomed in and annotated with gimp)
image

## Ideas
The first issue (red arrows) can be resolved by disabling ticks altogether for the left, top, and bottom axes by adding
`axis.setTicks([])` in this `else` case: https://github.com/pyqtgraph/pyqtgraph/blob/3fbb4f6b026b838293c9cbfc026a03dc47d62257/pyqtgraph/graphicsItems/ColorBarItem.py#L95-L96

### Tested environment(s)

 * PyQtGraph version: 0.12.1
 * Qt Python binding: 'PyQt5 5.15.1 Qt 5.15.1'
 * Python version: 3.8.5 (default, May 27 2021, 13:30:53)
 * NumPy version: '1.18.5'
 * Installation method: pip### Additional context

### Additional context
https://github.com/ZELLMECHANIK-DRESDEN/ShapeOut2/issues/87

NilsNemitz:
This addresses the issue described in #1866 by applying the solution conveniently included in the report.
Thank you @paulmueller !

Since ColorBarItem uses an unlinked AxisItem to display the scale information, only the ticks on the main axis move with adjustment. The unsynchronized ticks on the opposite axis (used to close the frame) were hidden by setting their size to zero.
 However, these length-zero ticks become visible at large screen or export resolutions, possibly as a result of Qt's handling of line end caps?

The proper solution is to clear out the list of ticks, so that no drawing occurs in the first place. This is also much more elegant, and I am sure it is technically faster. 

The only other change here is to move the ColorBarItem AxisItems from z=1 to z=0.5 which is the new default for AxisItems generated by PlotItem.

Closes #1866

### DIFF
--- a/pyqtgraph/graphicsItems/ColorBarItem.py
+++ b/pyqtgraph/graphicsItems/ColorBarItem.py
@@ -85,7 +85,7 @@ class ColorBarItem(PlotItem):
         for key in ['left','right','top','bottom']:
             self.showAxis(key)
             axis = self.getAxis(key)
-            axis.setZValue(1)
+            axis.setZValue(0.5)
             # select main axis:
             if self.horizontal and key == 'bottom':
                 self.axis = axis
@@ -93,7 +93,8 @@ class ColorBarItem(PlotItem):
                 self.axis = axis
                 self.axis.setWidth(45)
             else: # show other axes to create frame
-                axis.setStyle( showValues=False, tickLength=0 )
+                axis.setTicks( [] )
+                axis.setStyle( showValues=False )
         self.axis.setStyle( showValues=True )
         self.axis.unlinkFromView()
         self.axis.setRange( self.values[0], self.values[1] )


j9ac9k:
LGTM, thanks for the PR @NilsNemitz and thanks for the report and proposed solution @paulmueller

Issue #1787, opened 2021-05-19
GLScatterPlotItem: Segmentation fault: 11 on MacOS while addItem


johanneshauck:
Hi everyone,

I had some problems using a QApp on MacOS, which was developed on windows. The problem was a segmentation fault after adding a GLScatterPlotItem to the GLViewWidget.

The following hardware was used:
- Python 3.8.3 (tags/v3.8.3:6f8c832, May 13 2020, 22:37:02) [MSC v.1924 64 bit (AMD64)] on win32
- Python 3.8.10 (v3.8.10:3d8993a744, May  3 2021, 08:55:58) [Clang 6.0 (clang-600.0.57)] on MacOS 10.14.6

I added two test files to demonstrate problem and solution in a minimal working example from the pyqtgraph.examples.

testGLScatterPlotItemWin.py runs on windows, but causes a segmentation fault on macOS. Adding the line w.show() in testGLScatterPlotItemMacOS.py fixed the problem and runs on MacOS.

[testGLScatterPlotItem.zip](https://github.com/pyqtgraph/pyqtgraph/files/6509880/testGLScatterPlotItem.zip)

j9ac9k:
Hi @johanneshauck,

Thanks for the example code.

I do get the segfault on macOS when running `testGLScatterPlotItemWin.py` on macOS.  I do not get a segfault running `testGLSCatterPlotItemMacOS.py` on macOS (although that file appears to be using many undefined variables).

I'm not sure what pyqtgraph here could do for the end user here; with the openGL stuff, there are countless platform specific issues you can run to.  We do our best to have example code that works on all platforms, and that's why the examples in the example directory are part of our test suite.

Since python cannot catch segfaults, what do you think pyqtgraph can/should do here instead?

pijyoi:
Actually, this issue is related to the change from QGLWidget to QOpenGLWidget and is not limited to macOS. Since that change, it is a requirement to call ```show()``` on GLViewWidget before adding any items to it.

Now, the catch is that this failure will only occur for those items which make OpenGL calls in their initializeGL method. There are only two of them:
1) GLImageItem
2) GLScatterPlotItem

That's why the first item added ```GLGridItem``` doesn't cause issues.

[...]

PR #1871, opened 2021-06-29
Allow adding items to GLViewWidget before show

pijyoi:
Since the move from QGLWidget to QOpenGLWidget, it has become necessary for GLViewWidget to be shown before it gets a valid OpenGL Context. This results in some scripts previously working with pyqtgraph <= 0.11.1 to fail with pyqtgraph >= 0.12.0, with little clue as to the reason, e.g. #1787

~~This PR prints a more helpful warning when it detects such a situation, and hopefully the user will be able to remedy it themselves.~~

Technically, only GLImageItem and GLScatterPlotItem were affected because they made OpenGL calls in their pseudo-InitializeGL() method. ~~This PR also modifies GLImageItem to defer its initialization to a later stage as its initialization is trivial.~~

Annotated script documenting the required change. This example also demonstrates how to put a GLViewWidget inside a QMainWindow.
```python
import pyqtgraph as pg
import pyqtgraph.opengl as gl
import numpy as np
import scipy.misc

app = pg.mkQApp()
win = QtWidgets.QMainWindow()
win.resize(640, 480)
# with QGLWidget, QMainWindow::show() can be called here
# win.show()

glv = gl.GLViewWidget()
glv.opts['elevation'] = 90
glv.opts['azimuth'] = 0    

win.setCentralWidget(glv)
# with QOpenGLWidget, QMainWindow::show() must be called _only_ after
#   QMainWindow::setCentralWidget()
#   GLImageItem and GLScatterPlotItem can only be added after GLViewWidget
#       has been shown
win.show()

image = scipy.misc.face()
tex, _ = pg.makeRGBA(image)
glii = gl.GLImageItem(tex)
s = 1/128
glii.scale(s, s, 1)
w, h = tex.shape[:2]    # GLImageItem uses a transposed image
w *= s
h *= s
glii.translate(-w/2, -h/2, 0)
glv.addItem(glii)

spots = np.array([[-w/2, -h/2, 0, 1, 0, 0, 0.5],
                  [-w/2,  h/2, 0, 0, 1, 0, 0.5],
                  [ w/2,  h/2, 0, 0, 0, 1, 0.5],
                  [ w/2, -h/2, 0, 0, 1, 1, 0.5]])
glspi = gl.GLScatterPlotItem(pos=spots[:,:3], size=50, color=spots[:,3:])
glv.addItem(glspi)

app.exec_()
```

### DIFF
--- a/pyqtgraph/opengl/GLViewWidget.py
+++ b/pyqtgraph/opengl/GLViewWidget.py
@@ -434,10 +434,11 @@ class GLViewWidget(QtWidgets.QOpenGLWidget):
         return xDist / self.width()
         
     def mousePressEvent(self, ev):
-        self.mousePos = ev.localPos()
+        lpos = ev.position() if hasattr(ev, 'position') else ev.localPos()
+        self.mousePos = lpos
         
     def mouseMoveEvent(self, ev):
-        lpos = ev.localPos()
+        lpos = ev.position() if hasattr(ev, 'position') else ev.localPos()
         diff = lpos - self.mousePos
         self.mousePos = lpos
         
--- a/pyqtgraph/opengl/GLViewWidget.py
+++ b/pyqtgraph/opengl/GLViewWidget.py
@@ -102,8 +102,7 @@ class GLViewWidget(QtWidgets.QOpenGLWidget):
         if self.isValid():
             self.makeCurrent()
             try:
-                item.initializeGL()
-                item.setInitialized()
+                item.initialize()
             except:
                 self.checkOpenGLVersion('Error while adding item %s to GLViewWidget.' % str(item))
                 
@@ -129,10 +128,12 @@ class GLViewWidget(QtWidgets.QOpenGLWidget):
         self.update()        
         
     def initializeGL(self):
+        """
+        Initialize items that were not initialized during addItem().
+        """
         for item in self.items:
-            if not item.initialized():
-                item.initializeGL()
-                item.setInitialized()
+            if not item.isInitialized():
+                item.initialize()
         
     def setBackgroundColor(self, *args, **kwds):
         """
--- a/pyqtgraph/opengl/GLViewWidget.py
+++ b/pyqtgraph/opengl/GLViewWidget.py
@@ -8,7 +8,6 @@ import warnings
 from math import cos, sin, tan, radians
 ##Vector = QtGui.QVector3D
 
-ShareWidget = None
 
 class GLViewWidget(QtWidgets.QOpenGLWidget):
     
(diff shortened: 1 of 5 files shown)


pijyoi:
For another user example where the script worked with pyqtgraph <= 0.11.1 but failed with pyqtgraph >= 0.12.0
https://groups.google.com/g/pyqtgraph/c/cTm6QDLTowI/m/xNKkMZJaBwAJ

pijyoi:
Instead of warning the user about where show() should be called, change the code such that show() can be called at any point.

For the example given above, we can now call show() right before exec().
```python
import pyqtgraph as pg
import pyqtgraph.opengl as gl
import numpy as np
import scipy.misc

app = pg.mkQApp()
win = QtWidgets.QMainWindow()
win.resize(640, 480)

glv = gl.GLViewWidget()
glv.opts['elevation'] = 90
glv.opts['azimuth'] = 0    

win.setCentralWidget(glv)

image = scipy.misc.face()
tex, _ = pg.makeRGBA(image)
glii = gl.GLImageItem(tex)
s = 1/128
glii.scale(s, s, 1)
w, h = tex.shape[:2]    # GLImageItem uses a transposed image
w *= s
h *= s
glii.translate(-w/2, -h/2, 0)
glv.addItem(glii)

spots = np.array([[-w/2, -h/2, 0, 1, 0, 0, 0.5],
                  [-w/2,  h/2, 0, 0, 1, 0, 0.5],
                  [ w/2,  h/2, 0, 0, 0, 1, 0.5],
                  [ w/2, -h/2, 0, 0, 1, 1, 0.5]])
glspi = gl.GLScatterPlotItem(pos=spots[:,:3], size=50, color=spots[:,3:])
glv.addItem(glspi)

win.show()
app.exec_()
```

pijyoi:
Above script running on Linux with this PR.
Screenshot at 2021-07-01 13-07-24


Above script running on Linux with master. (On Windows, PyOpenGL throws exceptions and no window is shown at all. On Linux, no exceptions are thrown and the following faulty image is shown) 
Screenshot at 2021-07-01 13-07-56

pyqtgraph/opengl/GLViewWidget.py
```diff
@@ -99,10 +98,12 @@ def reset(self):
 
     def addItem(self, item):
         self.items.append(item)
-        if hasattr(item, 'initializeGL'):
+
+        if self.isValid():
             self.makeCurrent()
             try:
                 item.initializeGL()
+                item.setInitialized()
```
ixjlyons, line 106: Are both required? I see none of the subclasses call the parent `initializeGL` since it doesn't do anything. Maybe `setInitialized` could call `initializeGL`? So `GLGraphicsItem.initializeGL` is the user-facing method (consistent with Qt) and the base class handles the state variable for internal usage. In that case `initialize` and `isInitialized` might be clearer.

ixjlyons (commented):
Minor optional suggestion but this looks good to me. I saw the progress of this PR and it's great you identified a way to reclaim backward compatibility.

pyqtgraph/opengl/GLViewWidget.py
pijyoi: That's a great suggestion, looks less ugly now.

ixjlyons (approved):

j9ac9k:
Thanks @pijyoi for this PR!  Merging!

PR #1873, opened 2021-07-01
Set focus to the example filter + Ctrl+F listener for focus

ntjess:


examples/ExampleApp.py
```diff
@@ -318,6 +318,7 @@ def __init__(self):
 
         textFil = self.ui.exampleFilter
         self.curListener = None
+        self.ui.exampleFilter.setFocus()
```
ntjess, line 321: The entire goal of this PR is to be able to start typing to filter as soon as you open the example app (without having to click-focus the search bar)

ixjlyons (approved):

j9ac9k:
Thanks @ntjess this LGTM!  Merging!

PR #1875, opened 2021-07-02
bug: PyQt6 does not have localPos()

pijyoi:
this bug must have existed since the removal of the mouse shims from Qt.py in #1818.

### DIFF
--- a/pyqtgraph/opengl/GLViewWidget.py
+++ b/pyqtgraph/opengl/GLViewWidget.py
@@ -430,10 +430,11 @@ class GLViewWidget(QtWidgets.QOpenGLWidget):
         return xDist / self.width()
         
     def mousePressEvent(self, ev):
-        self.mousePos = ev.localPos()
+        lpos = ev.position() if hasattr(ev, 'position') else ev.localPos()
+        self.mousePos = lpos
         
     def mouseMoveEvent(self, ev):
-        lpos = ev.localPos()
+        lpos = ev.position() if hasattr(ev, 'position') else ev.localPos()
         diff = lpos - self.mousePos
         self.mousePos = lpos
         


ntjess:
Any thoughts on adding a `position` shim in `Qt` to avoid the `if` statement entirely? Then we can use `position` everywhere, considering `pos`/`localPos` are deprecated? I think Ogi mentioned there was a reason this wasn't implemented, though

```python
if QT_LIB in [PYQT5, PYSIDE2]:
  # These bindings don't support 6-style mouse event position()
  def position(self):
    return self.pos() # or localPos()?
  QtGui.QMouseEvent.position = position
```

pijyoi:
We are trying hard to avoid adding new monkey-patches to the Qt modules. It changes the modules globally and is not appropriate for a library.

By the way, it can be easy to get confused with the various mouse event APIs:
1) QMouseEvent.pos() -> QPoint
2) QMouseEvent.localPos() -> QPointF
3) QMouseEvent.position() -> QPointF (new for Qt6)
4) QGraphicsSceneMouseEvent.pos() -> QPointF
5) QWheelEvent.position() -> QPointF
6) QDragMoveEvent.pos() -> QPoint
7) QDragMoveEvent.posF() -> QPointF 
8) QDragMoveEvent.position() -> QPointF (specific to PyQt6 even though Qt6 C++ doesn't define it)

In general, where there is a float version, we should use that, so (1) and (6) should no longer be used.
(8) is an incompatibility between PyQt6 vs the rest of the bindings.

j9ac9k:
What a fantastic summary of the mouse APIs that's for that @pijyoi; it's almost as if we need a custom flake8 plugin to check for these things...

Regarding monkey patching, is that still an issue with #1846 having been merged?

pijyoi:
There are some Qt4 shims awaiting removal in 0.13.0
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/Qt/__init__.py#L244-L295

There's a QtTest.QTest.qwait shim for PySide{2,6}, but that's probably not so bad since QtTest is not used by user code.
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/Qt/__init__.py#L320-L327

ntjess:
@pijyoi that makes a lot of sense; I was thinking forward (rather than backward) shims would be acceptable, but I definitely see why there are enough reasons to reduce the clobbering.

Issue #1815, opened 2021-06-01
Suggestion: Raise TypeError instead of Exception in pg.eq


ntjess:
The error raised in https://github.com/pyqtgraph/pyqtgraph/blob/e29f86578f120fda2281cfddb7429d03470fe29b/pyqtgraph/functions.py#L544 could be more clearly specified as a `TypeError` than a general Exception to make it easier to catch in outer scopes.

This could be a breaking change, so it is understandable if it is rejected.

ixjlyons:
I don't think this is technically a breaking change. If someone currently catches it with `except Exception:` or a bare `except:`, that'd still catch the `TypeError`. I'm in favor of being more specific.

j9ac9k:
I _hate_ bare exceptions, so anything is better than that IMO.  I approve of this change too.

ntjess:
Closed by #1885

PR #1885, opened 2021-07-08
Raise TypeError instead of Exception

jennifer-manriquez:
#### What does this PR do?
This is to raise a TypeError instead of general Exception in the eq function, which is in charge of checking a is b equality,  to make it easier to catch in outer scopes. 

#### What testing has been done on this PR?
`pytest tests`

#### What are the relevant issues?
 Raise TypeError instead of Exception in pg.eq [#1815](hhttps://github.com/pyqtgraph/pyqtgraph/issues/1815). 
Design aspects discussed in issue. 

#### Do the docs need to be updated?
No

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -722,7 +722,7 @@ def eq(a, b):
         else:
             return e.all()
     else:
-        raise Exception("== operator returned type %s" % str(type(e)))
+        raise TypeError("== operator returned type %s" % str(type(e)))
 
 
 def affineSliceCoords(shape, origin, vectors, axes):


j9ac9k:
I'm good with merging this, I'm just wondering if the exception string should be converted to a f-string (in an attempt to get away from the `%` syntax of string generation).

@ntjess as this impacts parameter trees, do we need to have other changes go along w/ this?

j9ac9k:
BTW thanks for the PR @jennifer-manriquez !

jennifer-manriquez:
Happy to update the syntax, also, I was wondering which were the cases in which this exception could be raised, maybe I could add it to the eq function tests if I find one, before merging.

ntjess:
Maybe this should be treated separately from the `eq` modification for custom types (discussed here: https://pyqtgraph.slack.com/archives/CD9ABDT3R/p1622648129121900, not sure how to permalink outside slack before the history disappears), but if that _does_ get implemented the logic will have to change somewhat significantly anyway.

That's a bit more to sign up for than the issue describes, though, so I think this should be good for the time being.

j9ac9k:
@ntjess in terms of saving the post before slack history engulfs it, probably easiest to migrate the text into a new issue or discussion.  I agree we should implement what we discussed there, looks like we have some work to do, I agree with thinking that's really a separate issue tho.

@jennifer-manriquez I would say let's leave the style for the string creation as is; I wouldn't be surprised if we go through the library changing these up later (using [pyupgrade](https://github.com/asottile/pyupgrade) perhaps?) 

Anyway, this LGTM, merging!  Thanks for the PR @jennifer-manriquez

ntjess:
@j9ac9k Ok to close #1815 or do the magic words just need to be added to this merged PR?

j9ac9k:
oh right, no point in adding the magic words now, the ship has sailed, but go ahead and manually close it out 👍🏻

PR #1870, opened 2021-06-28
Update README

edumur:
Add matplotlib as optional library.
Replace hdf5 by h5py.
Add links towards the github page or website of the different linraries.

PR #1891, opened 2021-07-11
Clean up redundancy in README

ixjlyons:
There's some redundant info that could've been removed as part of #1870

Check https://github.com/ixjlyons/pyqtgraph/blob/readme-redundant-optional-deps/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,22 +35,13 @@ Currently this means:
 
 * Python 3.7+
 * Qt 5.12-5.15, 6.1
-* Required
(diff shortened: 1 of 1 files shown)


j9ac9k:
I'm on mobile so I don't have the best view to see the readme, but we will be supporting numpy 1.18+ pretty soon (2 weeks).

j9ac9k:
On second thought we can update the readme with the new numpy revision after we pass the date; and update the CI while we're at it.

Issue #662, opened 2018-04-13
How to do overpainting with GLViewWidget?


traveller59:
## Environment
* pyqtgraph 0.10.0
* Ubuntu 16.04
## Problems
I have add following code to paintGL:
```Python
        painter = QPainter(self)
        painter.setRenderHint(QPainter.Antialiasing)
        painter.drawText(10, 10, 'test')
        painter.end()
```
These code works well with official pyqt hellogl example, but don't work with pyqtgraph.

NilsNemitz:
Dear @traveller59 , I hope you have figured this all out by now.

In principle, the QPainter should be able to draw over the contents of GLViewWidget.
But it is not simple to switch the painter between the OpenGL "native mode" and the normal Qt behavior. I found this as a possible starting point:
https://stackoverflow.com/questions/39171120/qopenglwidget-qpainterdrawrect-only-draws-four-arrows?noredirect=1&lq=1

Within PyQtGraph, only graphicsItems/PlotCurveItem.py seems to implements this kind of mode switching, so you might want to look at that, too.

On the other hand, the issues might be as simple as needing to set a visible pen, or that the view may be scaled such that your coordinates 10,10 are be off-screen.

Please let us know if you have a problem specifically with the PyQtGraph use of OpenGl!

j9ac9k:
I think we can close this, citing #1892 as a solution here.  Thanks @pijyoi for creating the example!

PR #1892, opened 2021-07-12
implement GLPainterItem

pijyoi:
While playing around with GLTextItem, I found that it would not scale well when drawing many text items. In the following script, for example, it is very sluggish to rotate the scene.
```python
import pyqtgraph as pg
import pyqtgraph.opengl as gl

pg.mkQApp("GLTextItem Example")

gvw = gl.GLViewWidget()
gvw.opts['elevation'] = 90
gvw.opts['azimuth'] = 0
gvw.opts['distance'] = 50
gvw.show()

size = 50
griditem = gl.GLGridItem()
griditem.setSize(size, size)
griditem.setSpacing(1, 1)
gvw.addItem(griditem)

axisitem = gl.GLAxisItem()
gvw.addItem(axisitem)

for xi in range(size+1):
    for yi in range(size+1):
        pos = (xi - size/2, yi - size/2, 0)
        gvw.addItem(gl.GLTextItem(pos=pos, text='.'))
pg.exec()
```

I also came across #662, which seems to be asking for a generalization of what GLTextItem provides.

As a demonstration / proof-of-concept, this PR implements GLPainterItem, allowing the user to issue multiple arbitrary painter method calls. The (rough) equivalent example to the above script is provided and it rotates smoothly. I think the main overhead in the GLTextItem example above must have been in constructing and destructing the QPainter object.

### DIFF
--- a/examples/GLBarGraphItem.py
+++ b/examples/GLBarGraphItem.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-Demonstrate use of GLLinePlotItem to draw cross-sections of a surface.
+This example demonstrates the use of GLBarGraphItem.
 
 """
 ## Add path to library (just for examples; you do not need this)
--- a/examples/GLPainterItem.py
+++ b/examples/GLPainterItem.py
@@ -48,10 +48,9 @@ class GLPainterItem(pg.opengl.GLGraphicsItem.GLGraphicsItem):
         painter.drawText(rect, af.AlignBottom | af.AlignLeft, 'BL')
         painter.drawText(rect, af.AlignBottom | af.AlignRight, 'BR')
 
-        opts = self.view().getCameraPosition()
-        opts['fov'] = self.view().fov()
+        opts = self.view().cameraParams()
         lines = []
-        center = opts['pos']
+        center = opts['center']
         lines.append(f"center : ({center.x():.1f}, {center.y():.1f}, {center.z():.1f})")
         for key in ['distance', 'fov', 'elevation', 'azimuth']:
             lines.append(f"{key} : {opts[key]:.1f}")
--- a/examples/GLPainterItem.py
+++ b/examples/GLPainterItem.py
@@ -1,0 +1,94 @@
+"""
+Demonstrate using QPainter on a subclass of GLGraphicsItem.
+"""
+## Add path to library (just for examples; you do not need this)
+import initExample
+
+import pyqtgraph as pg
+import pyqtgraph.opengl
+from pyqtgraph.Qt import QtCore, QtGui
+import OpenGL.GL as GL
+
+SIZE = 32
+
+class GLPainterItem(pg.opengl.GLGraphicsItem.GLGraphicsItem):
+    def __init__(self, **kwds):
+        super().__init__()
+        glopts = kwds.pop('glOptions', 'additive')
+        self.setGLOptions(glopts)
+
+    def compute_projection(self):
+        modelview = GL.glGetDoublev(GL.GL_MODELVIEW_MATRIX)
+        projection = GL.glGetDoublev(GL.GL_PROJECTION_MATRIX)
+        mvp = projection.T @ modelview.T
+        mvp = QtGui.QMatrix4x4(mvp.ravel().tolist())
+
+        # note that QRectF.bottom() != QRect.bottom()
+        rect = QtCore.QRectF(self.view().rect())
+        ndc_to_viewport = QtGui.QMatrix4x4()
+        ndc_to_viewport.viewport(rect.left(), rect.bottom(), rect.width(), -rect.height())
+
+        return ndc_to_viewport * mvp
+
+    def paint(self):
+        self.setupGLState()
+
+        painter = QtGui.QPainter(self.view())
+        self.draw(painter)
+        painter.end()
+
+    def draw(self, painter):
+        painter.setPen(QtCore.Qt.GlobalColor.white)
+        painter.setRenderHints(QtGui.QPainter.RenderHint.Antialiasing | QtGui.QPainter.RenderHint.TextAntialiasing)
+
+        rect = self.view().rect()
+        af = QtCore.Qt.AlignmentFlag
+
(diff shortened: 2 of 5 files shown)


j9ac9k:
Hi @pijyoi thanks for providing the code to demonstrate the sluggish behavior, I'll run that through a profiler in the next day or so, but I suspect you're right with respect to the cause of the slow-down.

I don't have any direct feedback for the code itself; I'm just glad some of the 3D plots are getting some attention

I guess I do have one question here, and this isn't just specific to this PR, but some of the 3D capability in general; and that has to do with types. 

The pyopengl calls return numpy arrays.  We deal w/ Qt objects routinely because...well, it's self explanatory.  Should we try and minimize our usage of native python data structures (lists, tuples) and so on?  

What made me think of it here `GLPainterItem.project` which returns a list, that is promptly unpacked into a QPointF constructor.  Should we be trying to stick with using QPointF, QVector{2, 3, 4}D, and use their associated methods to handle the various transformations more?  I know we ran into trouble w/ `QTransform.inverted()`.... but anyway I'm curious on your thoughts on this.

pijyoi:
>I don't have any direct feedback for the code itself; I'm just glad some of the 3D plots are getting some attention

Rather than add yet another class to pyqtgraph that may not fit anyone's particular usage, I have changed this PR to be an example of how to subclass GLGraphicsItem. GLPainterItem was after all just a thin wrapper around GLGraphicsItem. This also prevents poorly-defined methods from becoming a support burden for pyqtgraph.

While doing so, I noticed that opengl/GLTextItem.py was not doing relative imports.

>What made me think of it here GLPainterItem.project which returns a list, that is promptly unpacked into a QPointF constructor. Should we be trying to stick with using QPointF, QVector{2, 3, 4}D, and use their associated methods to handle the various transformations more? I know we ran into trouble w/ QTransform.inverted().... but anyway I'm curious on your thoughts on this.

I am not against that, but the issue is that with such small objects, the overhead of calling into a DLL is non-negligible.
In C++ code, the accessor and the simpler methods of QPointF, QVector{2,3,4}D should be all inline. When using a binding, these inline methods become full-fledged functions and have high overhead. So to use modification of a point as an example, represented as either a python list, numpy array or QPointF

Python list or numpy array (high overhead for numpy array)
```python
pt[0] += 1
```

QPointF (both high overhead and unwieldy)
```python
pt.setX(pt.x() + 1)
```

j9ac9k:
> I am not against that, but the issue is that with such small objects, the overhead of calling into a DLL is non-negligible.
In C++ code, the accessor and the simpler methods of QPointF, QVector{2,3,4}D should be all inline. When using a binding, these inline methods become full-fledged functions and have high overhead. 

You are clearly more experienced and knowledgable than I am in this area; I have no feel for what the overhead is like.

I just keep thinking on the speed difference with `QTransform.inverted()` vs. `np.linalg.inv`, and that difference has got me thinking about trying to stick to Qt bindings instead of numpy equivalent methods when a method exists that is relevant for us here (I recognize part of that was the numpy array construction and re-construction of the QTransform with the result).

It's not that I doubt you, but I am curious to learn more about the trade off and at what point should we want to use native python objects  vs. Qt objects, would this be represented in memory consumption?

pijyoi:
>It's not that I doubt you, but I am curious to learn more about the trade off and at what point should we want to use native python objects vs. Qt objects, would this be represented in memory consumption?

I would say speed only matters if you are intending for the function to handle a large number of the objects. So to use the ```project()``` method as an example, if it takes in only one element, then it doesn't matter if it's a QVector3D, numpy array, tuple, list. If you want it to handle multiple elements at a time, then it should probably be a numpy array. You wouldn't be able to vectorize a list of QVector3D.

As a matter of personal opinion, I don't find it useful for a Python method that takes 1 element to be "overloaded" to be able to handle either a QVector3D, a numpy array, a list, a tuple or *(x, y, z).

Sometimes, it's also a matter of not knowing that there is an equivalent method in Qt. Again using the ```project()``` method, C++ Qt seemed to have just the thing needed: https://doc.qt.io/qt-5/qvector3d.html#project. Alas, it is not present in the Python binding. Instead, the ```QMatrix4x4::viewport()``` method turned out to be what was needed. It might have been obvious to a someone familiar with Qt OpenGL. It wasn't to me.

pijyoi:
If GLViewWidget.opts is treated as private, then there is no way to retrieve the camera angles.

j9ac9k:
@pijyoi since we have a `setCameraPosition`, we should probably implement a `cameraPosition` method to GLViewWidget that returns the camera arguments that would go into `setCameraPosition`.  

It's hard to justify a dictionary is intended to be read-only via the public API.

pijyoi:
```cameraPosition()``` is already defined to return the camera position in xyz.
I created a ```getCameraPosition()``` method that would return the settable arguments from ```setCameraPosition()``` and also accessor methods for ```fov``` since fov is not a camera position.
 
Or maybe we could create instead methods ```cameraParams()``` and ```setFov()```. ```fov``` could be returned in ```cameraParams()```.

j9ac9k:
> `cameraPosition()` is already defined to return the camera position in xyz.

right, should have looked there first before I started typing!

> I created a `getCameraPosition()` method that would return the settable arguments from `setCameraPosition()` and also accessor methods for `fov` since fov is not a camera position.
> 
> Or maybe we could create instead methods `cameraParams()` and `setFov()`. `fov` could be returned in `cameraParams()`.

I think the `setCameraParams`/`cameraParams` is a good solution here!

j9ac9k:
@pijyoi this LGTM, and I recognize this PR represents a proof of concept/example, but since this works so nicely, should we try and roll some of this functionality into GLTextItem?

pijyoi:
I forgot to add the test skip for Big Sur, is it supposed to pass the CI?

j9ac9k:
> I forgot to add the test skip for Big Sur, is it supposed to pass the CI?

The pyopengl related failures only happen on macOS 11, our CI pipelines use macOS 10.15 so I would not expect the example to fail there.

j9ac9k:
Love the other changes you made to the example tests.

PR #1893, opened 2021-07-12
Code completion in pyqtgraph's console

ntjess:
Uses either Jedi or poor man's Jedi (if you don't have it in your environment) to provide code completion on Ctrl+Space

Non-jedi version just suggests words from past history, jedi just spawns an interpreter and calls `.complete()`

- [ ] Non-jedi completer:
    - [ ] `CLAUSE` definitions should probably be checked to ensure they're exhaustive (https://github.com/pyqtgraph/pyqtgraph/pull/1893/files#diff-3d7000126a6585e9ef1feafd8b50057555f610a325295b295fb5a0452e1ffbbbR15)
    - [ ] I have no idea how to make a robust completion model, some review of my implementation would be good
    - [ ] Should ignore history that results in an exception
    - [x] Double+ dot-accessors (`pg.plot(x, y).getViewBox()`)
    - [x] Assignment operators (`x = np.ones`)
- [ ] Jedi completer:
    - [ ] all `import` statements aren't prefixed properly
    - [ ] API compatibility issues for older Jedi versions
- [ ] Implementation of `PopupLineEditor` was pulled from one of my other projects so it still needs converting to pyqtgraph format:
    - [x] Lambdas connected to slots -> local or new functions
    - [ ] Function documentation
    - [x] Type hint removal
    - [ ] Auditing next/prev focusing logic
        - [ ] Tab key should complete selection and close popup instead of cycling through options
        - [ ] Platform-agnostic hotkey instead of "Ctrl" (Does `Qt.Key.Key_Control` map to "Option" on Mac?)
    - [ ] Resolve issues on Qt bindings that open the popup in the screen corner rather than on top of the line edit
- [ ] Compile regexes (this is an easy one)

 - If someone has a more intuitive (least "astonishing") alternative for refreshing the completion model compared to waiting for Ctrl+Space, I'm all ears.
 - If someone has a 'dumb' replacement for missing `jedi` that's better than just looking for words in the history, I'm all ears.
     - I could probably try `dir` or `vars` on the current object etc., but that sounded like a lot of work and at a certain point I don't want to re-implement `jedi` just for this console

https://user-images.githubusercontent.com/23620506/125359019-b25bd000-e337-11eb-8d27-0ac809ea3325.mp4

### DIFF
--- a/pyqtgraph/console/CmdInput.py
+++ b/pyqtgraph/console/CmdInput.py
@@ -1,16 +1,21 @@
-from ..Qt import QtCore, QtGui
+from ..Qt import QtCore, QtWidgets
+from ..widgets.PopupLineEditor import PopupLineEditor
 from ..python2_3 import asUnicode
 
-class CmdInput(QtGui.QLineEdit):
+class CmdInput(PopupLineEditor):
     
     sigExecuteCmd = QtCore.Signal(object)
-    
-    def __init__(self, parent):
-        QtGui.QLineEdit.__init__(self, parent)
+    sigCompleteRequested = QtCore.Signal(object) # Current text
+
+    def __init__(self, parent, *args, **kwargs):
+        kwargs.setdefault('validatePrefix', False)
+        kwargs.setdefault('clearOnComplete', False)
+        super().__init__(parent, *args, **kwargs)
         self.history = [""]
         self.ptr = 0
     
     def keyPressEvent(self, ev):
+        ctrlPressed = (ev.modifiers() & QtCore.Qt.KeyboardModifier.ControlModifier)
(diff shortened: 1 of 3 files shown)


j9ac9k:
2 space indentations?! what is this an indent for ants?!

j9ac9k:
Joking aside, this is a great feature for the console, thanks for the PR @ntjess

ntjess:
> 2 space indentations?! what is this an indent for ants?!

:laughing: I can always appreciate a Zoolander reference. I almost get distracted by cavernous 4-space indents... Eh, to each their own :stuck_out_tongue:

PR #1895, opened 2021-07-15
Update contributing guide ahead of scipy sprints

j9ac9k:
Our contributing guide is horrifically out of date, this updates it along with giving some code for how to create a virtual environment and install dependencies.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,77 +4,86 @@ Contributions to pyqtgraph are welcome! Be kind and respectful! See [our Code of
 
 Please use the following guidelines when preparing changes:
 
-## Submitting Code Changes
+## Development Environment Creation
 
-* The preferred method for submitting changes is by github pull request against the "master" branch.
-* Pull requests should include only a focused and related set of changes. Mixed features and unrelated changes may be rejected.
-* For major changes, it is recommended to discuss your plans on the mailing list or in a github issue before putting in too much effort.
-* The following deprecations are being considered by the maintainers
-  * `pyqtgraph.opengl` may be deprecated and replaced with `VisPy` functionality
-  * After v0.11, pyqtgraph will adopt [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) which will effectively mean that python2 support will be deprecated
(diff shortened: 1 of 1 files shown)


ixjlyons (approved):
Looks good to me, this should be really helpful for the sprints.

The comment about WSL was a bit surprising though -- I use it (when I have to) and it works. It does require some additional setup but I'm not sure it makes sense to discourage it. I guess if it's intended to avoid helping people out with getting it set up, that's fine.

j9ac9k:
> The comment about WSL was a bit surprising though -- I use it (when I have to) and it works. It does require some additional setup but I'm not sure it makes sense to discourage it. I guess if it's intended to avoid helping people out with getting it set up, that's fine.

This came from some digital walkthrough I tried to get someone else up and running with pyqtgraph development on windows.  This individual had done web dev primarily beforehand and defaulted to WSL, which I don't think was rendering a window.  I can remove the bit about it, or if there is a link that can explain a bit more about how to get windows to display via WSL I'll gladly link that in.  I do have access to a windows machine, I should probably spend a little time trying to craft a procedure to get that working there.

ixjlyons:
I don't know of a best practices type article to refer to. It's basically just a clunky X11 forwarding setup with a server running on the windows side. There are probably several options for it but I ended up landing on https://mobaxterm.mobatek.net/

Thinking about it again, just leave it in - it's a headache figuring out how to get other people's bashrc set up right so saying it's not supported is fine by me :)

j9ac9k:
> it's a headache figuring out how to get other people's bashrc set up right so saying it's not supported is fine by me :)

That's the spirit! 😆

pijyoi:
Currently, to run pytest, the user needs to also install h5py and scipy. 
Matplotlib is optional due to the use of ```pytest.importorskip("matplotlib")```. It would be nice if h5py could also be made optional in the same way.
I think scipy was a hard requirement.

j9ac9k:
> I think scipy was a hard requirement.

Good point on `h5py`, I think that's a "good first issue" for the sprints, I'll create the issue.  SciPy, while not needed to install pyqtgraph is a requirement for the image components, which a bulk of the testing encompasses, and feels silly to run the test sutie without SciPy, so I'm good leaving it as a hard requirement for development/test-suite.

PR #1897, opened 2021-07-15
PlotCurveItem: setup modelview and projection

pijyoi:
This fixes an "experimental" feature that stopped working on Windows platform since the addition of support for Qt6 (and drop of support for Qt4).
i.e. change from QGLWidget to QOpenGLWidget

Minimal testable example at https://github.com/pyqtgraph/pyqtgraph/pull/1525#issuecomment-770304898

Would fix #1760

PR #1902, opened 2021-07-17
simplify modelview projection computation

pijyoi:
simplify and shorten expressions used in #1897
1) no need to get rect(), which is actually defined as
	QRect(0, 0, width(), height())
2) use col-maj data() instead of row-maj copyDataTo()
   - glLoadMatrixf() takes col-maj


j9ac9k:
Nice, that is definitely way simpler!  

If you're interested in joining the sprints this weekend, this is where we are: https://discord.gg/5VypRxrQ

A few people are online there now; I'll be joining in a few hours.

pijyoi:
>If you're interested in joining the sprints this weekend, this is where we are: https://discord.gg/5VypRxrQ

I am getting a "Messages Failed To Load" in the #pyqtgraph channel

PR #1903, opened 2021-07-17
Cleanup GLViewWidget

pijyoi:
tries to cleanup various parts of GLViewWidget

UPDATE:
non-cosmetic changes:
* remove redefinitions of width(), height(), devicePixelRatio()
* trivially implement readQImage() with grabFrameBuffer()
* fail upfront if not at least OpenGL (non-ES) 2.0
* fix setCameraPosition euler mode
* fix renderToArray() on hidpi systems
* change renderToArray() to output non-transposed images (this could be considered an API change)

### DIFF
--- a/examples/GLBarGraphItem.py
+++ b/examples/GLBarGraphItem.py
@@ -13,9 +13,9 @@ import numpy as np
 
 app = pg.mkQApp("GLBarGraphItem Example")
 w = gl.GLViewWidget()
-w.opts['distance'] = 40
 w.show()
 w.setWindowTitle('pyqtgraph example: GLBarGraphItem')
+w.setCameraPosition(distance=40)
 
 gx = gl.GLGridItem()
 gx.rotate(90, 0, 1, 0)
--- a/examples/GLImageItem.py
+++ b/examples/GLImageItem.py
@@ -15,9 +15,9 @@ import numpy as np
 
 app = pg.mkQApp("GLImageItem Example")
 w = gl.GLViewWidget()
-w.opts['distance'] = 200
 w.show()
 w.setWindowTitle('pyqtgraph example: GLImageItem')
+w.setCameraPosition(distance=200)
 
 ## create volume data set to slice three images from
 shape = (100,100,70)
--- a/examples/GLLinePlotItem.py
+++ b/examples/GLLinePlotItem.py
@@ -13,9 +13,9 @@ import numpy as np
 
 app = pg.mkQApp("GLLinePlotItem Example")
 w = gl.GLViewWidget()
-w.opts['distance'] = 40
 w.show()
 w.setWindowTitle('pyqtgraph example: GLLinePlotItem')
+w.setCameraPosition(distance=40)
 
 gx = gl.GLGridItem()
 gx.rotate(90, 0, 1, 0)
--- a/examples/GLScatterPlotItem.py
+++ b/examples/GLScatterPlotItem.py
@@ -15,9 +15,9 @@ import numpy as np
 
 app = pg.mkQApp("GLScatterPlotItem Example")
 w = gl.GLViewWidget()
-w.opts['distance'] = 20
 w.show()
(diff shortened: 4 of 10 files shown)


j9ac9k:
The diff so far looks great @pijyoi 

If we're looking to clean this up, I would suggest in the `reset` method to change from

```python
self.setBackgroundcolor('k')
```

to 

```python
self.setBackgroundColor(pg.getconfigOption('background')
```

Please keep in mind I know next to nothing about openGL, but should we be using `glLoadMatrixd` instead of `glLoadMatrixf`, since most numpy arrays are of type double?

Lastly, perhaps try and get rid of the bare exceptions in `drawItemTree` and `keyReleaseEvent` (not sure what kind of exception would be raised from drawItemTree though).

pijyoi:
>but should we be using glLoadMatrixd instead of glLoadMatrixf, since most numpy arrays are of type double?

The matrices are generated from QMatrix4x4 which uses float. Both ```data()``` and ```copyToData()``` return a Python tuple of the float values converted to double. This tuple is then passed directly (without conversion to numpy) to glLoadMatrixf. 

Now, under the hood, PyOpenGL would then need to convert this Python tuple to a numpy array.

Since the values originated as float, I thought glLoadMatrixf would be the appropriate call.

j9ac9k:
> Since the values originated as float ...

Missed that bit.  Thanks for the explanation.

pijyoi:
script to exercise renderToArray()
```python
import pyqtgraph.opengl as gl
import pyqtgraph as pg
import numpy as np
import OpenGL.GL as GL

app = pg.mkQApp()
glv = gl.GLViewWidget()
glv.setCameraPosition(distance=30)
glv.show()

n = 101
s = np.linspace(-10,10,n)
X, Y = np.meshgrid(s, s)
Z = 10 * np.sinc(np.hypot(X, Y))
for i in range(100):
    color = pg.glColor((i,n*1.3))
    row = np.column_stack((X[i,:], Y[i,:], Z[i,:]))
    col = np.column_stack((X[:,i], Y[:,i], Z[:,i]))
    plt = gl.GLLinePlotItem(pos=row, color=color)
    glv.addItem(plt)
    plt = gl.GLLinePlotItem(pos=col, color=color)
    glv.addItem(plt)

arr = glv.renderToArray((640, 480), format=GL.GL_RGBA)
qimg = pg.functions.ndarray_to_qimage(arr, QtGui.QImage.Format.Format_RGBA8888)
qimg.save('qimage.png')
```

pijyoi:
>Lastly, perhaps try and get rid of the bare exceptions in drawItemTree and keyReleaseEvent

Probably not for drawItemTree. Assuming the user is allowed to create their own subclasses of GLGraphicsItem, then any exception could be possible.

j9ac9k:
@pijyoi this LGTM, I'm curious about one change though, what does this line come from in RawImageWidget and GLImageItem ?

```
            glDisable(GL_TEXTURE_2D)
```

pijyoi:
I suspect the code for GLImageItem was copied from GLVolumeItem which does use GL_TEXTURE_3D.
And then RawGLImageWidget was probably copied from GLImageItem.

```glEnable(GL_TEXTURE_2D)``` should probably be paired with a ```glDisable(GL_TEXTURE_2D)```, otherwise it doesn't make sense to disable something that wasn't even enabled to begin with.

j9ac9k:
> I suspect the code for GLImageItem was copied from GLVolumeItem which does use GL_TEXTURE_3D.
> And then RawGLImageWidget was probably copied from GLImageItem.
> 
> `glEnable(GL_TEXTURE_2D)` should probably be paired with a `glDisable(GL_TEXTURE_2D)`, otherwise it doesn't make sense to disable something that wasn't even enabled to begin with.

Looking at the entirety of the files that had that modification, I see what you're talking about.  

This LGTM 👍🏻

PR #1904, opened 2021-07-17
tests/exporters/test_hdf5.py: skip if no h5py

### LINKED ISSUE #1898: Make h5py an optional dependency in the test suite (just like matplotlib)
Currently if you attempt to run the test suite, and do not have h5py installed, one of the tests will fail due to it not being installed.  This library should be made an optional dependency, much the same way that matplotlib has been as pointed out by here: https://github.com/pyqtgraph/pyqtgraph/pull/1895#issuecomment-881146056

Artturin:
Closes #1898


j9ac9k:
Thanks for the PR @Artturin this LGTM, Merging!

Issue #1906, opened 2021-07-18
GLVolumeItem example relies on undefined casting behavior


pijyoi:
The following lines in the GLVolumeItem example relies on various float to ubyte casting side effects. In particular, casting of +Inf to ubyte has different behavior on arm64 platforms, leading to a "wrong" image that looks similar to the one in #1775
```python
# Original Code
d2[..., 0] = positive * (255./positive.max())
d2[..., 1] = negative * (255./negative.max())
```

Possible replacement that also avoids some other side effects,
However, the reliance on all these side effects makes the original code rather suspect.
```python
# Reformulated Code
# Both positive.max() and negative.max() are negative-valued.
# Thus the next 2 lines are _not_ bounded to [0, 255]
positive = positive * (255./positive.max())
negative = negative * (255./negative.max())
# When casting to ubyte, the original code relied on +Inf to be
# converted to 0. On arm64, it gets converted to 255.
# Thus the next 2 lines change +Inf explicitly to 0 instead.
positive[np.isinf(positive)] = 0
negative[np.isinf(negative)] = 0
# When casting to ubyte, the original code relied on the conversion
# to do modulo 256. The next 2 lines do it explicitly instead as
# documentation.
d2[..., 0] = positive.astype(int) % 256
d2[..., 1] = negative.astype(int) % 256
```

j9ac9k:
Given this is an example, I think it's safe to say that the intent was strictly a demo purposes to generate some nifty geometry in not too many lines of code.  If we're getting undefined casting behavior (especially if it's different across architectures), we should probably make the suggested changes here to get consistency everywhere.

tl;dr I'm not too worried about code intent, but I think consistency across platforms/architecture is more important.

PR #1907, opened 2021-07-18
fix GLVolumeItem example for arm64

pijyoi:
implement changes suggested in #1906


j9ac9k:
thanks for testing on ARM64 and spotting this difference!  LGTM, merging!

PR #1908, opened 2021-07-18
more readable names for color map references in ColorBarItem API

NilsNemitz:
I recently realized that `ColorBarItem` uses the internal shorthand `cmap` in two of its user-facing methods. I would propose renaming this to `colorMap` to be more consistent and obvious.

Since ColorBarItem is still a rather fresh addition to the library, hopefully that won't cause too much trouble.
We can keep the current methods working, in which case this is most importantly a change to the documentation.

This would affect the method `setCmap()` and the parameter `cmap` in the init code.

### DIFF
--- a/pyqtgraph/graphicsItems/ColorBarItem.py
+++ b/pyqtgraph/graphicsItems/ColorBarItem.py
@@ -67,7 +67,7 @@ class ColorBarItem(PlotItem):
             colorMap = cmap
         self.img_list  = [] # list of controlled ImageItems
         self.values    = values
-        self.cmap      = colorMap
+        self._colorMap = colorMap
         self.rounding  = rounding
         self.horizontal = bool( orientation in ('h', 'horizontal') )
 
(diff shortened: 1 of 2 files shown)


j9ac9k:
👍🏻 on the date based deprecation....

I think 1 year warning is good here from the merge should be good (so removed in first release after July, 2022?

j9ac9k:
I think this is ready to merge once a date is put in the deprecation warning.

pyqtgraph/graphicsItems/ColorBarItem.py
```diff
@@ -35,33 +36,40 @@ class ColorBarItem(PlotItem):
...
+            warnings.warn(
+                "The parameter 'cmap' has been renamed to 'colorMap' for clarity. "
+                "The old name will no longer be available in any version of PyQtGraph released after July 2022.",
+                DeprecationWarning, stacklevel=2
+            )
+            colorMap = cmap
         self.img_list  = [] # list of controlled ImageItems
         self.values    = values
-        self.cmap      = cmap
+        self.cmap      = colorMap
```
outofculture, line 70: Though to be in proper Qt style, `self.colorMap` should be a _method_ that returns the color map. So we actually want to store the value in `self._colorMap`, have the deprecated property for `cmap`, and an accessor method `self.colorMap()`.

```python
def colorMap(self):
    return self._colorMap

@property
def cmap(self):
    warnings.warn('deprecated!', DeprecationWarning, stacklevel=2)
    return self._colorMap
```

NilsNemitz: Dear @outofculture , I didn't even consider that somebody might want to get the color map *back* from the ColorBarItem, but you are right, we should have that. While retrieving this as `some_color_map_object.cmap` was never documented, this is indeed a good way to remove that with a proper deprecation period.

Added, and I am going to look this up here the next time I need it 👍

j9ac9k:
LGTM, thanks for the submission @NilsNemitz !

And with this merge, we will have our lowest outstanding PR count since the end of 2014.

PR #1909, opened 2021-07-18
None-check AxisItem.label before access

NilsNemitz:
I was clicking things while the tests were running, and I managed to cause the error messages included below. These were printed by the test suite after all tests had been declared "pass".

They both seem to originate from `AxisItem`'s `resizeEvent()` getting called after `self.label` has already been set to `None` by the `close()` method. I am not aware of another reason that would change `self.label` to be None.

This mini PR adds a None check, so that tear down can still proceed cleanly even if resize events are generated. 

```
========================== 265 passed, 7 skipped in 280.11s (0:04:40) ===========================      
  File "C:\e\prog\pg_dev\pyqtgraph\pyqtgraph\graphicsItems\AxisItem.py", line 236, in resizeEvent      
    br = self.label.boundingRect()
AttributeError: 'NoneType' object has no attribute 'boundingRect'
  File "C:\e\prog\pg_dev\pyqtgraph\pyqtgraph\graphicsItems\AxisItem.py", line 236, in resizeEvent      
    br = self.label.boundingRect()
AttributeError: 'NoneType' object has no attribute 'boundingRect'
```

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -233,6 +233,10 @@ class AxisItem(GraphicsWidget):
 
         ## Set the position of the label
         nudge = 5
+        if self.label is None: # self.label is set to None on close, but resize events can still occur.
+            self.picture = None
+            return
+            
         br = self.label.boundingRect()
         p = QtCore.QPointF(0, 0)
         if self.orientation == 'left':


pijyoi:
See #1835. The issue seemed to be due to using pytest monkey patch.

NilsNemitz:
Is that because the pytest patching keeps the objects alive longer than intended? 
Is the timing otherwise deterministic enough to know that resizeEvent will no longer be called when the label is None?

My impression was that I may have clicked on one of the figures just as it was being closed: I was trying to do something else while the tests were running and constantly creating and closing figures under my mouse. I don't really remember seeing this error before.

If this is a simple race condition, I am still thinking we might add the check. Or is there a more elegant way to fix this?

pijyoi:
I can get it to always show up by running only the two tests using the monkeypatch. Conversely, if I run all the tests, it rarely shows up.

>Is that because the pytest patching keeps the objects alive longer than intended?

I have no idea, but it seems to me that the monkey patching is introducing some unintended side effects.

```
(pyqt6) PS C:\work\github\pyqtgraph> pytest -v .\tests\graphicsItems\test_AxisItem.py -k "stopAxisAtTick or tickFont"
================================================= test session starts =================================================
platform win32 -- Python 3.8.10, pytest-6.2.4, py-1.10.0, pluggy-0.13.1 -- c:\work\venv\pyqt6\scripts\python.exe
cachedir: .pytest_cache
rootdir: C:\work\github\pyqtgraph, configfile: pytest.ini
collected 6 items / 4 deselected / 2 selected

tests/graphicsItems/test_AxisItem.py::test_AxisItem_stopAxisAtTick PASSED                                        [ 50%]
tests/graphicsItems/test_AxisItem.py::test_AxisItem_tickFont PASSED                                              [100%]

=========================================== 2 passed, 4 deselected in 0.41s ===========================================
  File "C:\work\github\pyqtgraph\pyqtgraph\graphicsItems\AxisItem.py", line 236, in resizeEvent
    br = self.label.boundingRect()
AttributeError: 'NoneType' object has no attribute 'boundingRect'
  File "C:\work\github\pyqtgraph\pyqtgraph\graphicsItems\AxisItem.py", line 236, in resizeEvent
    br = self.label.boundingRect()
AttributeError: 'NoneType' object has no attribute 'boundingRect'
  File "C:\work\github\pyqtgraph\pyqtgraph\graphicsItems\AxisItem.py", line 236, in resizeEvent
    br = self.label.boundingRect()
AttributeError: 'NoneType' object has no attribute 'boundingRect'
```

NilsNemitz:
You are right, I also get these messages reliably if I just run this subset of tests.

This bit
```python
def close(self):
        self.scene().removeItem(self.label)
        self.label = None
        self.scene().removeItem(self)
```
looks like a race condition to me, though, when considering that any resize events that still get triggered will error out after self.label has been set to None. Commenting out the `self.label = None` line stops the error from appearing, so what we are seeing is still consistent with an only slightly unexpected order of operations.

I am not sure how Qt handles events in this case. I assume the removeItem call should unsubscribe the AxisItem from receiving any? But is this filtered again at the end of the event queue? Or is it within spec that anything that is already queued up will still get dispatched to AxisItem?

j9ac9k:
http://doc.qt.io/qt-6/qgraphicsscene.html#removeItem

the bit about:

> The ownership of item is passed on to the caller

We can probably change this order from this:

```python
def close(self):
        self.scene().removeItem(self.label)
        self.label = None
        self.scene().removeItem(self)
```

to this

```python
def close(self):
        self.scene().removeItem(self.label)
        self.scene().removeItem(self)
        self.label = None
```

FWIW on macOS I cannot replicate this issue

j9ac9k:
yeah this PR LGTM, I don't know of a better way of handling this despite being able to reproduce.  

Thanks for the PR @NilsNemitz

PR #1910, opened 2021-07-18
PlotSpeedTest: add param tree control panel

pijyoi:
inspired by ScatterPlotSpeedTest.

Allows easier testing of the speed benefits of useOpenGL and/or enableExperimental.

On a particular Linux system, useOpenGL=True and enableExperimental=False slowed to a crawl.

### DIFF
--- a/examples/PlotSpeedTest.py
+++ b/examples/PlotSpeedTest.py
@@ -8,45 +8,190 @@ Update a simple plot as rapidly as possible to measure speed.
 import initExample
 
 from collections import deque
-from pyqtgraph.Qt import QtGui, QtCore
+from pyqtgraph.Qt import QtCore, QtGui, QtWidgets, QT_LIB
 import numpy as np
 import pyqtgraph as pg
 from time import perf_counter
+import pyqtgraph.parametertree as ptree
+import pyqtgraph.functions as fn
+import itertools
+import argparse
+
+if QT_LIB == 'PySide2':
+    wrapinstance = pg.Qt.shiboken2.wrapInstance
+elif QT_LIB == 'PySide6':
+    wrapinstance = pg.Qt.shiboken6.wrapInstance
+elif QT_LIB in ['PyQt5', 'PyQt6']:
+    wrapinstance = pg.Qt.sip.wrapinstance
+
+# defaults here result in the same configuration as the original PlotSpeedTest
+parser = argparse.ArgumentParser()
+parser.add_argument('--noise', dest='noise', action='store_true')
+parser.add_argument('--no-noise', dest='noise', action='store_false')
+parser.set_defaults(noise=True)
+parser.add_argument('--nsamples', default=5000, type=int)
+parser.add_argument('--frames', default=50, type=int)
+parser.add_argument('--fsample', default=1000, type=float)
+parser.add_argument('--frequency', default=0, type=float)
+parser.add_argument('--amplitude', default=5, type=float)
+args = parser.parse_args()
+
+# don't limit frame rate to vsync
+sfmt = QtGui.QSurfaceFormat()
+sfmt.setSwapInterval(0)
+QtGui.QSurfaceFormat.setDefaultFormat(sfmt)
+
+class LineInstances:
+    def __init__(self):
+        self.alloc(0)
+
+    def alloc(self, size):
+        self.arr = np.empty((size, 4), dtype=np.float64)
+        self.ptrs = list(map(wrapinstance,
+            itertools.count(self.arr.ctypes.data, self.arr.strides[0]),
+            itertools.repeat(QtCore.QLineF, self.arr.shape[0])))
+
+    def array(self, size):
+        if size > self.arr.shape[0]:
+            self.alloc(size + 16)
+        return self.arr[:size]
+
+    def instances(self, size):
+        return self.ptrs[:size]
+
+class MonkeyCurveItem(pg.PlotCurveItem):
+    def __init__(self, *args, **kwds):
+        super().__init__(*args, **kwds)
+        self.monkey_mode = ''
+        self._lineInstances = LineInstances()
+
+    def setMethod(self, param, value):
+        self.monkey_mode = value
+
+    def paint(self, painter, opt, widget):
+        if self.monkey_mode not in ['drawPolyline', 'drawLines']:
+            return super().paint(painter, opt, widget)
+
+        painter.setRenderHint(painter.RenderHint.Antialiasing, self.opts['antialias'])
+        painter.setPen(pg.mkPen(self.opts['pen']))
+
+        if self.monkey_mode == 'drawPolyline':
+            painter.drawPolyline(fn.arrayToQPolygonF(self.xData, self.yData))
+        elif self.monkey_mode == 'drawLines':
+            lines = self._lineInstances
+            npts = len(self.xData)
+            even_slice = slice(0, 0+(npts-0)//2*2)
+            odd_slice = slice(1, 1+(npts-1)//2*2)
+            for sl in [even_slice, odd_slice]:
+                npairs = (sl.stop - sl.start) // 2
+                memory = lines.array(npairs).reshape((-1, 2))
+                memory[:, 0] = self.xData[sl]
+                memory[:, 1] = self.yData[sl]
+                painter.drawLines(lines.instances(npairs))
 
 app = pg.mkQApp("Plot Speed Test")
 
-p = pg.plot()
-p.setWindowTitle('pyqtgraph example: PlotSpeedTest')
-p.setRange(QtCore.QRectF(0, -10, 5000, 20))
-p.setLabel('bottom', 'Index', units='B')
-curve = p.plot()
+default_pen = pg.mkPen()
 
-data = np.random.normal(size=(50, 5000))
-ptr = 0
+children = [
+    dict(name='sigopts', title='Signal Options', type='group', children=[
+        dict(name='noise', type='bool', value=args.noise),
+        dict(name='nsamples', type='int', limits=[0, None], value=args.nsamples),
+        dict(name='frames', type='int', limits=[1, None], value=args.frames),
+        dict(name='fsample', title='sample rate', type='float', value=args.fsample, units='Hz'),
+        dict(name='frequency', type='float', value=args.frequency, units='Hz'),
+        dict(name='amplitude', type='float', value=args.amplitude),
+    ]),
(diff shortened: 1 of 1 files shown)


j9ac9k:
on macOS with AntiAlias disabled and penWidth of 1


| `useOpenGL` | `enableExperimental` | FPS |
| ------------- | --------------------| ----- |
| False | False |  280      |
| False | True |    280     |
| True | False |  1.4       |
| True | True |  290       |

j9ac9k:
Something isn't being reset, when I first run I'm getting values in excess of 350ish fps, but when I start changing options, and then revert back, I'm getting 280ish fps

Regarding options, would be great if we could add some more checks such as 

*`skipFiniteCheck` 
* `connect='pairs'`
* [QPen join style](https://doc.qt.io/qt-5/qpen.html#setJoinStyle) (this one I have a hope that will impact the FPS hit we take from going to a thick line, but I suppose I can test this out manually before being added as a parameter tree).

j9ac9k:
Actually on the Pen one, hold off on that, hopefully we'll be merging #1844 soon which will have a PenSelectorDialog to the ParameterTree

pijyoi:
I have rebased to incorporate the merge of @dgoeries commit. Probably would need to change the setTitle update rate to be time-based rather than once every 50 updates. This is because the 50 value was based on the old fixed samples size of 5000 and we now allow the user to change the number of samples.

pijyoi:
Added an option to choose between plotting random and sine tone. Plotting "regular" signals is significantly faster than a random signal (due to less pixels being rendered, I guess). In particular, pen width > 1 is usable.

NilsNemitz:
That's very interesting. I have stumbled on an explanation before that considered the number of self-overlapping pixels to be particularly problematic. Not sure if that is true or not, but the reasoning was that the overlapping part of the same line needed special treatment in case the line was rendered with partial opacity:
A simple sequence of individual lines (which supposedly plots faster) would stack to give reduced transparency.
For a combined multi-segment line, the user expects all parts to merge into a single structure before translucency is equally applied to all of it.
Why that wouldn't fall back to a simpler rendered for fully opaque lines? Or draw fully opaque into a buffer and then blit that with translucency?  I have no idea :) 
But it seems somewhat consistent with the massive slowdown of zig-zagging random lines.

pijyoi:
>A simple sequence of individual lines (which supposedly plots faster) would stack to give reduced transparency.
For a combined multi-segment line, the user expects all parts to merge into a single structure before translucency is equally applied to all of it.

I added in an option to let the user try out plotting using ```drawPolyline``` (which PythonQwt uses) and ```drawLines```.
The drawLines implementation works only for PySide{2,6}

Now, while ```drawLines``` is much slower than ```drawPath``` with configuration (random, width=1), its performance with (random, width=2) remains consistent, whereas  ```drawPath``` would have absolutely choked.

j9ac9k:
> Now, while `drawLines` is much slower than `drawPath` with configuration (random, width=1), its performance with (random, width=2) remains consistent, whereas `drawPath` would have absolutely choked.

...woah

So while we've been thinking the limitation might have to do with the raster renderer, this might be a limitation of the raster render _and_ the use of `QPainterPath`.

Which reminds me, I still need to submit a feature request drawLines take a pointer to a numpy array on the pyside channels...

j9ac9k:
Nice use of #1844 for the Pen parameter tree :D 

When selecting some options, the plot doesn't render at all: 

image

Looks like toggling `useOpenGL` brings it back.

pijyoi:
Were you using pen width > 1?

The following says that macOS only supports opengl line width 1.
https://community.khronos.org/t/gllinewidth-gives-gl-invalid-value-for-widths-1/69023

j9ac9k:
I did indeed have a pen width of 2.  Didn't realize that wasn't supported! .... on rerunning, now I can't replicate....wonder if I entered a bogus value and didn't notice.

Thanks for the script; here's the output on my machine:

```
INFO:OpenGL.acceleratesupport:No OpenGL_accelerate module loaded: No module named 'OpenGL_accelerate'
VENDOR: Intel Inc.
RENDERER: Intel(R) Iris(TM) Plus Graphics OpenGL Engine
VERSION: 2.1 INTEL-16.4.5
MAX_TEXTURE_SIZE: 16384
MAX_3D_TEXTURE_SIZE: 2048
ALIASED_LINE_WIDTH_RANGE: [1 7]
SMOOTH_LINE_WIDTH_RANGE: [0 8]
SMOOTH_LINE_WIDTH_GRANULARITY: 0.125
```

For the fun of it, I set the pen width to 50, and it worked...

image

...why can't I replicate...

j9ac9k:
Also some property does not appear to be resetting.

I can start up the test, default settings, get ~450 fps.
Toggle useOpenGL/enableExperimental and get ~300 fps
Un-toggle useOpenGL/enableExperimental and get ~200 fps  <- on this config I'm effectively my first config.

pijyoi:
It seems that something sticks in Qt library the moment you have instantiated an QOpenGLWidget.
I verified that un-toggling useOpenGL does switch the viewport widget back to QWidget.

I noticed something similar in VideoSpeedTest.py that conditionally instantiates a RawImageGLWidget depending on whether PyOpenGL is installed. I recall that I could get higher frame rates by uninstalling PyOpenGL. Also different behavior on Windows and Linux.

j9ac9k:
> It seems that something sticks in Qt library the moment you have instantiated an QOpenGLWidget.
> I verified that un-toggling useOpenGL does switch the viewport widget back to QWidget.
> 
> I noticed something similar in VideoSpeedTest.py that conditionally instantiates a RawImageGLWidget depending on whether PyOpenGL is installed. I recall that I could get higher frame rates by uninstalling PyOpenGL. Also different behavior on Windows and Linux.

That's sneaky!

In the update function; I added a print statement `print(type(pw.viewport()))` but it's reporting the correct type.  If you know of a way to programmatically detect when it's still using the openGL capabilities that would be handy.

In the Qt docs, there is quite the blurb about cleaning/tear-down, but I'm not sure how relevant that is here.

pijyoi:
As a comparison, I tried porting PlotSpeedTest.py to use QtCharts:
https://gist.github.com/pijyoi/c8a2eb963a36a5d5bb14cc3b75a8bab0

Observations:
1) The OpenGL stickiness effect is also present
2) Doesn't choke with pen-width > 1

j9ac9k:
Nice work on the QtChart bit; ....wonder how it gets around that issue... It uses QGraphicsView and QGraphicsScene..

Looks like the deafult line thickness is 2 even!

https://github.com/qt/qtcharts/blob/7029c2d66beaacff1a0bc0336906ec52308de703/src/charts/linechart/qlineseries.cpp#L173-L183


Oh, I see why ... they use `painter.drawLine` when the a non-solid line is specified.  In the chart example, you can make it choke by specifying a non-solid pen parameter.

https://github.com/qt/qtcharts/blob/b69841fb16dd107454f7eead9e6e7b3db9ed4a2e/src/charts/linechart/linechartitem.cpp#L442-L448

pijyoi:
So it's consistent with what can be demonstrated with PlotSpeedTest.py

1) Plotting Method "pyqtgraph" uses drawPath which will choke. This is the alternate drawing method of QtCharts.
2) Plotting Method "drawLines" draws individual line segments and doesn't choke. This is the default drawing method of QtCharts.

j9ac9k:
> So it's consistent with what can be demonstrated with PlotSpeedTest.py
> 
> 1. Plotting Method "pyqtgraph" uses drawPath which will choke. This is the alternate drawing method of QtCharts.
> 2. Plotting Method "drawLines" draws individual line segments and doesn't choke. This is the default drawing method of QtCharts.

Yeah, this just shows the need we have for using `QPainter.drawLines` with a pointer to a numpy array (which I know, even if they add that feature, it won't propagate to versions of Qt we support :( ).  I can submit the feature request, but if you'd like to do it, by all means, truthfully you'll probably do a better job explaining the specifics of the issue than I would.

pijyoi:
I don't think there's a need to ask for a ```QPainter.drawLines``` numpy overload just yet. The workaround using ```wrapInstance``` is not too slow.

Btw, it's ```drawLines(QLineF, int count)``` that would be the preferred overload. The numpy overloads that are currently available are only usable with a self-compiled PySide6.

The usage would be something like:
```python
arr = np.empty((count, 4), dtype=np.float64)   # A QLineF is 4 floats
ptr = wrapInstance(arr.ctypes.data, QLineF)
painter.drawLines(ptr, count)
```
similar to what is done for ```drawPixmapFragments```

j9ac9k:
> The numpy overloads that are currently available are only usable with a self-compiled PySide6.

I should probably run this test myself; but was there a noticeable improvement in this use case?

pijyoi:
I am not confident nor keen to build such a big package as Qt.

j9ac9k:
> I am not confident nor keen to build such a big package as Qt.

I misunderstood, I thought this needed a custom PySide build, not a custom Qt build.

pijyoi:
>I misunderstood, I thought this needed a custom PySide build, not a custom Qt build.

I am not sure, I just assumed that Qt (even if unmodified) would need to be built together. Assuming that PySide can be built alone, one would still need the Qt headers, Qt link libraries and dlls. And the C++ compiler would need to be the same.

j9ac9k:
You can definitely build pyside independently of Qt.  I did that a while
back for something; but there are instructions here
https://doc.qt.io/qtforpython/gettingstarted.html#guides-per-platform

pijyoi:
I find that if I remove the vsync-unlock code, then toggling useOpenGL False->True->False results in an fps of 60Hz.

j9ac9k:
@pijyoi if there isn't a good way to "un-stuck" the openGL toggle, should we just add a little text blurb about it, and perhaps add a CLI argument to run with or without it (like VideoSpeedTest)?  ...This is a great PR, I'd like not not hold it up due to a limitation of the framework.

pijyoi:
I added an option --allow-opengl-toggle. User thus has to explicitly opt-in to use this feature.

Despite the stickiness, I did find it useful to be able to toggle between OpenGL <-> non-OpenGL. It was useful for discovering which combinations were absolutely un-usable.

j9ac9k:
> I added an option --allow-opengl-toggle. User thus has to explicitly opt-in to use this feature.
> 
> Despite the stickiness, I did find it useful to be able to toggle between OpenGL <-> non-OpenGL. It was useful for discovering which combinations were absolutely un-usable.

That's perfect, thanks for the PR @pijyoi !

PR #1911, opened 2021-07-18
Extend ColorMap with HSL cycles and subset generation

NilsNemitz:
This PR provides two color map tools that will be helpful in getting #1625 to work:
1) A generator function `makeHslCycle()` that can trace various circles and spirals through HSL color space. This basically generalizes the output of `functions.intColor()`. The idea is that the `intColor` code can at some point be replaced by sampling from such a color map, which will allow the generated colors to match different plotting palettes.
2) A ColorMap method `getSubset()` which returns a new ColorMap object that contains only the subset of "stops" in the specified range. This also allows the creation of reversed and rotated variants of the existing maps. The application is that the included color maps are very good sources for sampling plot colors, but often the cyclical maps don't start at the right point, or only some part of the full map matches the desired style. 

Other changes:
- The monochrome generator now defaults to a neutral grayscale, which is probably a better generic default than the "green monitor" gradient.
- The file extension for color maps in HTML hex color format was changed from .txt to .hex. This avoids the problem that the code tries to load the license files in the same folder, which isn't ideal. We previously did not ship any color maps in that format, so that change is unlikely to cause any problems. The code also fails with a little more helpful error message when it does try to load an invalid file. 

I still need to write some tests for the color map manipulation but I wanted to post this already:
I have failed to come up with an elegant method for extracting the ColorMap subsets, so let me know if you have a good idea for improving this. The sampling needs to work in both forward and backwards direction, and it needs to be able to wrap (1.0 to 0.0) and (0.0 to 1.0). I am sure there is a nice, simple way to do so...

### DIFF
--- a/pyqtgraph/colormap.py
+++ b/pyqtgraph/colormap.py
@@ -1,6 +1,6 @@
 import numpy as np
 from .Qt import QtGui, QtCore
-from .functions import mkColor, eq, colorDistance
+from .functions import mkColor, eq, colorDistance, clip_scalar, clip_array
 from os import path, listdir
 from collections.abc import Callable, Sequence
 import warnings
@@ -33,7 +33,7 @@ def listMaps(source=None):
         files = listdir( pathname )
         list_of_maps = []
         for filename in files:
-            if filename[-4:] == '.csv':
+            if filename[-4:] == '.csv' or filename[-4:] == '.hex':
                 list_of_maps.append(filename[:-4])
         return list_of_maps
     elif source.lower() == 'matplotlib':
@@ -95,11 +95,11 @@ def _getFromFile(name):
         filename = path.join(dirname, 'colors/maps/'+filename)
     if not path.isfile( filename ): # try suffixes if file is not found:
         if   path.isfile( filename+'.csv' ): filename += '.csv'
-        elif path.isfile( filename+'.txt' ): filename += '.txt'
+        elif path.isfile( filename+'.hex' ): filename += '.hex'
     with open(filename,'r') as fh:
         idx = 0
         color_list = []
-        if filename[-4:].lower() != '.txt':
+        if filename[-4:].lower() != '.hex':
             csv_mode = True
         else:
             csv_mode = False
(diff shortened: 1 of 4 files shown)


NilsNemitz:
I have also added some hand-picked plot colors in the form of color map data now. I think these do reasonably well at
- being bright and colorful
- not hanging around in the ultra intense magenta and cyan edges of the color space too much
- offering nine colors that I think are distinguishable even when they are not in the same corner of the screen.
This comes in a "bright" version for dark backgrounds and a slightly darker version that works a little better on a bright background.

Here is some example code to play around with the new functions:
```python
import pyqtgraph as pg
import numpy as np

pg.mkQApp()

cmA = pg.colormap.get('PAL-relaxed')
cmB = pg.colormap.get('PAL-relaxed_bright')

cmC = pg.colormap.makeHslCycle( hue=(0.0,4.0), saturation=(1.0,0.2), lightness=(0.8,0.2), steps=72 )
cmD = pg.colormap.makeHslCycle( hue=(0.0,-0.5), saturation=(1.0,0.2), lightness=(0.2,0.8), steps=72 )
cmE = pg.colormap.makeHslCycle( hue=(0.0, 0.5), saturation=(1.0,0.2), lightness=(0.2,0.8), steps=72 )
cmF = pg.colormap.makeHslCycle()
# cmD = pg.colormap.get('viridis')

cm2 = cmF.getSubset( 2.333, -0.667 )
cm3 = cmF.getSubset( -1.333, +0.667 )
cm4 = cmF.getSubset( 0.667, -0.667 )
cm5 = cmF.getSubset( 0.667, +0.667 )
cm6 = cmF.getSubset( 1.000, -1.000 )

layout = pg.GraphicsLayoutWidget()
layout.addItem( pg.ColorBarItem(cmap=cmA, interactive=False, orientation='horizontal') )
layout.nextRow()
layout.addItem( pg.ColorBarItem(cmap=cmB, interactive=False, orientation='horizontal') )
layout.nextRow()
layout.addItem( pg.ColorBarItem(cmap=cmC, interactive=False, orientation='horizontal') )
layout.nextRow()
layout.addItem( pg.ColorBarItem(cmap=cmD, interactive=False, orientation='horizontal') )
layout.nextRow()
layout.addItem( pg.ColorBarItem(cmap=cmE, interactive=False, orientation='horizontal') )
layout.nextRow()
layout.addItem( pg.ColorBarItem(cmap=cmF, interactive=False, orientation='horizontal') )
layout.nextRow()
layout.addItem( pg.ColorBarItem(cmap=cm2, interactive=False, orientation='horizontal') )
layout.nextRow()
layout.addItem( pg.ColorBarItem(cmap=cm3, interactive=False, orientation='horizontal') )
layout.nextRow()
layout.addItem( pg.ColorBarItem(cmap=cm4, interactive=False, orientation='horizontal') )
layout.nextRow()
layout.addItem( pg.ColorBarItem(cmap=cm5, interactive=False, orientation='horizontal') )
layout.nextRow()
layout.addItem( pg.ColorBarItem(cmap=cm6, interactive=False, orientation='horizontal') )

layout.resize(800,600)
layout.show()

pg.exec()
```
First two outputs are the new palettes.
The next four are various HSL loops and spirals, the last one should reproduce the default `intColor()` set.
After that are four subsets of this, with two different starting points, and going backwards from there. Two of the subsets wrap around over the 1.0 to 0.0 boundary of the cyclical map.
The last color map is a full reverse of the source map.
new palette,HSL loops, subsets

j9ac9k:
I grab `.hex` files I think we need to modify `setup.py` to tell setuptools to bundle those files as well.

pyqtgraph/colormap.py
```diff
@@ -191,20 +195,66 @@ def getFromColorcet(name):
...
+    """
+    if isinstance( hue, (tuple, list) ):
+        hueA, hueB = hue
+    else:
+        hueA = hue
+        hueB = hueA + 1.0
+    if isinstance( saturation, (tuple, list) ):
+        satA, satB = saturation
+    else:
+        satA = satB = saturation
```
j9ac9k, line 241: if-else clause that does the same thing in both cases?

PR #1912, opened 2021-07-18
push bullet lists over 2 spaces to get them to show up as such in the…

outofculture:
… docs

### DIFF
--- a/pyqtgraph/parametertree/SystemSolver.py
+++ b/pyqtgraph/parametertree/SystemSolver.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from collections import OrderedDict
 import numpy as np
 import copy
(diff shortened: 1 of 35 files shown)


pyqtgraph/parametertree/SystemSolver.py
```diff
@@ -52,16 +53,16 @@ class SystemSolver(object):
     1) The *defaultState* class attribute: This is a dict containing a 
        description of the variables in the system--their default values,
        data types, and the ways they can be constrained. The format is::
-       
+       ```
```
ixjlyons, line 56: `SystemSolver` isn't currently included in the docs but the two colons at the end of the line above put this indented block in a literal environment.

pyqtgraph/parametertree/SystemSolver.py
```diff
@@ -52,16 +53,16 @@ class SystemSolver(object):
...
-       
-       * *value* is the default value. May be None if it has not been specified
-         yet.
-       * *type* may be float, int, bool, np.ndarray, ...
-       * *constraint* may be None, single value, or (min, max)
-            * None indicates that the value is not constrained--it may be 
-              automatically generated if the value is requested.
-       * *allowed_constraints* is a string composed of (n)one, (f)ixed, and (r)ange. 
+       ```
+         * *value* is the default value. May be None if it has not been specified
```
ixjlyons, line 59: Incidentally, indenting this bulleted list makes it part of the literal environment above. I'm not sure how to delineate literal blocks with trailing indented content. I tried `.. code-block::` instead of the shorthand `::` above and added an extra line between the `{name: ...` line and the list but it still made the list part of the code block.

outofculture: I added a `Where:` on a line by itself in between the two, which should fix this.

ixjlyons: Perfect

PR #1914, opened 2021-07-18
Remove the use of pyqtgraph.ptime

j9ac9k:
With us supporting python3.7+, we have no more need for the ptime module
and can instead safely use perf_counter for everything.

### DIFF
--- a/examples/ImageItem.py
+++ b/examples/ImageItem.py
@@ -34,14 +34,14 @@ data = np.random.normal(size=(15, 600, 600), loc=1024, scale=64).astype(np.uint1
 i = 0
 
 updateTime = perf_counter()
-fps = 0
+elapsed = 0
 
 timer = QtCore.QTimer()
 timer.setSingleShot(True)
 # not using QTimer.singleShot() because of persistence on PyQt. see PR #1605
 
 def updateData():
-    global img, data, i, updateTime, fps
+    global img, data, i, updateTime, elapsed
 
     ## Display the data
     img.setImage(data[i])
@@ -49,11 +49,11 @@ def updateData():
 
     timer.start(1)
     now = perf_counter()
-    fps2 = 1.0 / (now-updateTime)
+    elapsed_now = now - updateTime
     updateTime = now
-    fps = fps * 0.9 + fps2 * 0.1
-    
-    #print "%0.1f fps" % fps
+    elapsed = elapsed * 0.9 + elapsed_now * 0.1
+
+    # print(f"{1 / elapsed:.1f} fps")
     
 timer.timeout.connect(updateData)
 updateData()
--- a/examples/ImageItem.py
+++ b/examples/ImageItem.py
@@ -9,7 +9,7 @@ import initExample
 from pyqtgraph.Qt import QtCore, QtGui
 import numpy as np
 import pyqtgraph as pg
-import pyqtgraph.ptime as ptime
+from time import perf_counter
 
 app = pg.mkQApp("ImageItem Example")
 
@@ -33,7 +33,7 @@ view.setRange(QtCore.QRectF(0, 0, 600, 600))
 data = np.random.normal(size=(15, 600, 600), loc=1024, scale=64).astype(np.uint16)
 i = 0
 
-updateTime = ptime.time()
+updateTime = perf_counter()
 fps = 0
 
 timer = QtCore.QTimer()
@@ -48,7 +48,7 @@ def updateData():
     i = (i+1) % data.shape[0]
 
     timer.start(1)
-    now = ptime.time()
+    now = perf_counter()
     fps2 = 1.0 / (now-updateTime)
     updateTime = now
     fps = fps * 0.9 + fps2 * 0.1
--- a/examples/PlotSpeedTest.py
+++ b/examples/PlotSpeedTest.py
@@ -3,13 +3,15 @@
 """
 Update a simple plot as rapidly as possible to measure speed.
 """
-import initExample
+
 ## Add path to library (just for examples; you do not need this)
-from time import perf_counter
+import initExample
+
 from collections import deque
 from pyqtgraph.Qt import QtGui, QtCore
 import numpy as np
 import pyqtgraph as pg
+from time import perf_counter
 
 app = pg.mkQApp("Plot Speed Test")
 
--- a/examples/infiniteline_performance.py
+++ b/examples/infiniteline_performance.py
@@ -4,7 +4,8 @@ import initExample ## Add path to library (just for examples; you do not need th
 from pyqtgraph.Qt import QtGui, QtCore
 import numpy as np
 import pyqtgraph as pg
-from pyqtgraph.ptime import time
+from time import perf_counter
+
 app = pg.mkQApp("Infinite Line Performance")
 
 p = pg.plot()
@@ -20,7 +21,7 @@ for i in range(100):
 
 data = np.random.normal(size=(50, 5000))
 ptr = 0
-lastTime = time()
+lastTime = perf_counter()
 fps = None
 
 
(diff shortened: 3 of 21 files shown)


j9ac9k:
EDIT also fixes the nit-pick issue identified by @ixjlyons in the previous PR I merged but didn't address 😬

examples/PlotSpeedTest.py
```diff
@@ -30,10 +31,10 @@ def update():
 
     ptr += 1
     # Measure
-    t_start = perf_counter()
+    t_start = pg.time()
     curve.setData(data[ptr % 10])
     app.processEvents(QtCore.QEventLoop.ProcessEventsFlag.AllEvents)
-    elapsed.append(perf_counter() - t_start)
+    elapsed.append(pg.time() - t_start)
```
NilsNemitz, line 37: Sorry, what happened here?
Does this swap from perf_counter back to pg.time?

pyqtgraph/imageview/ImageView.py
```diff
@@ -483,12 +483,12 @@ def evalKeyState(self):
             if key == QtCore.Qt.Key.Key_Right:
                 self.play(20)
                 self.jumpFrames(1)
-                self.lastPlayTime = ptime.time() + 0.2  ## 2ms wait before start
+                self.lastPlayTime = perf_counter() + 0.2  ## 2ms wait before start
```
NilsNemitz, line 486: Is this 200 ms?

j9ac9k: yeah, I'm thinking that the comment here is wrong...I should look at the git history to try and get a little more context.

pyqtgraph/SignalProxy.py
```diff
@@ -76,7 +76,7 @@ def flush(self):
             return False
         args, self.args = self.args, None
         self.timer.stop()
-        self.lastFlushTime = time()
+        self.lastFlushTime = perf_counter()
         self.sigDelayed.emit(args)
         return True
```
NilsNemitz, line 82: I guess this needs perf_counter (and not just time.time() ) to allow precise rate limiting at short intervals?
 
Unrelated to this PR:
Below this is a the disconnect function with a comment regarding a workaround for PySide on Python 2.
Do you know that that is about and if it can be removed or simplified now?

NilsNemitz (approved):
Dear @j9ac9k , I looked at all the changes and did not see anything really problematic.
I also ran all the modified examples, and they all seem to give FPS values in the expected range. So I think this can be merged.

Before that, I left a few small comments to look at.

pyqtgraph/imageview/ImageView.py
j9ac9k: @outofculture you have insight here?  Looks like this manipulates the timing during key-strokes, I'm assuming the comment is wrong, but should the value being added be modified?

examples/infiniteline_performance.py
```diff
@@ -4,7 +4,8 @@
 from pyqtgraph.Qt import QtGui, QtCore
```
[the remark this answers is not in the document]
j9ac9k: I don't think so? I'm starting to think we should better compartmentalize our benchmarks vs. examples ...this is basically PlotSpeedTest.py with a bunch of infinitelines drawn.

examples/multiplePlotSpeedTest.py
```diff
@@ -71,7 +72,7 @@ def fastPlot():
     item.setPen(pg.mkPen('w'))
     plt.addItem(item)
     
-    dt = pg.ptime.time() - start
+    dt = perf_counter() - start
     print("Create plots took: %0.3fms" % (dt*1000))
```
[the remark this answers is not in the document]
j9ac9k: fair point, easy enough to change!

pyqtgraph/SignalProxy.py
j9ac9k: Fairly sure it can be simplified/removed now... although that should probably be its own commit in case we have to git-bisect our way back.  PySide1 has almost no resemblance to PySide2 of today (besides the name)

j9ac9k: looking a bit more on it, this can be simplified a great deal now with the use of [`weakref.finalize`](https://docs.python.org/3/library/weakref.html#weakref.finalize) which is new in python 3.4

examples/ImageItem.py
```diff
@@ -48,7 +48,7 @@ def updateData():
     i = (i+1) % data.shape[0]
 
     timer.start(1)
-    now = ptime.time()
+    now = perf_counter()
```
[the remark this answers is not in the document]
j9ac9k: > How do we do this elsewhere?

We're all over the map on this, and I suspect we will need to have a collective discussion on how we should do it, and apply the change elsewhere...

pyqtgraph/SignalProxy.py
j9ac9k: actually while I just tried this, I realized there is a gotcha with weakref.finalize, you cannot reference the object being finalized within the method being called...which is a bit problematic due to how we handle the signals... might be worth revisiting more later, but going to file this under "if it's stupid, and it works, it's not stupid".

pyqtgraph/imageview/ImageView.py
NilsNemitz: This very much looks like an outdated comment. If we can't figure it out, maybe just drop the "2ms" value and keep it as  `# wait before start` ?

examples/ImageItem.py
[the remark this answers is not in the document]
NilsNemitz: I am good with leaving this as it is here for now, since it doesn't actually output anything by default.

examples/infiniteline_performance.py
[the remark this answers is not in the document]
NilsNemitz: It seems that it isn't so useful to have this if it isn't reference anywhere, but I am good with keeping it in its current hiding place. I think I would add it to the "benchmark" section of the example app, though.

pyqtgraph/SignalProxy.py
NilsNemitz: If your reading of the comment is "This implementation was chosen because it also works on Pyside" then I agree.
At first glance the current code looked reasonable enough. It also did not look like this wouldn't typically be part of anything time-critical. So, seems good not to touch it unless you see a nice way to make it elegant!

pyqtgraph/imageview/ImageView.py
j9ac9k: I'll give @outofculture a day or so to chime in but yeah otherwise drop the numerical value in the comment I think is the best way to go forward.

outofculture: This comment is from a Luke-omnibus commit 10 years ago, and it's definitely incorrect compared to the code. `lastPlayTime` is only used in the `ImageView.timeout` method, which controls the auto-advancing play behavior and has a no-op when "`lastPlayTime` is in the future". If I had to give this line a comment, it would be "effectively pauses playback for 0.2s".

pyqtgraph/SignalProxy.py
j9ac9k: I've been thinking of drafting a PR to try and remove some of these Qt4 work-arounds/band-aids one commit at a time (for git bisect purposes)... I'll be sure to try and clean this up then!

examples/scrollingPlots.py
```diff
@@ -78,7 +79,7 @@ def update2():
 
 def update3():
     global p5, data5, ptr5, curves
-    now = pg.ptime.time()
+    now = perf_counter()
```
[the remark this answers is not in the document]
j9ac9k: Created #1916 from this comment

PR #1920, opened 2021-07-24
arrayToQPath can handle empty paths

### LINKED ISSUE #1888: `arrayToQPath` fails when drawing an empty path and boolean array `connect` argument
(pypi version 0.12.2 release)
```python
import numpy as np
import pyqtgraph as pg
x = np.zeros((0, 2))
isfinite = np.ones(x.shape[0], bool)
pg.arrayToQPath(*x.T, connect=isfinite)
```

Traceback:
```
  File "C:\Users\njessurun\Miniconda3\envs\py39\lib\site-packages\IPython\core\interactiveshell.py", line 3441, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-2-28fd2e751002>", line 5, in <module>
    pg.arrayToQPath(*x.T, connect=isfinite)
  File "C:\Users\njessurun\Miniconda3\envs\py39\lib\site-packages\pyqtgraph\functions.py", line 2047, in arrayToQPath
    arr[0]['c'] = 0  # the first vertex has no previous vertex to connect
IndexError: index 0 is out of bounds for axis 0 with size 0
```

ntjess:
- [X] Closes #1888
- [X] Increases test coverage across all supported dtypes

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1925,6 +1925,8 @@ def arrayToQPath(x, y, connect='all', finiteCheck=True):
 
     path = QtGui.QPainterPath()
     n = x.shape[0]
+    if n == 0:
+      return path
 
     connect_array = None
     if isinstance(connect, np.ndarray):
@@ -1936,7 +1938,7 @@ def arrayToQPath(x, y, connect='all', finiteCheck=True):
     isfinite = None
     if connect == 'finite':
         isfinite = np.isfinite(x) & np.isfinite(y)
-        if not finiteCheck or n == 0:
+        if not finiteCheck:
             # if user specified to skip finite check, then that forces use_qpolygonf
             use_qpolygonf = True
         else:
@@ -2047,8 +2049,7 @@ def arrayToQPath(x, y, connect='all', finiteCheck=True):
     else:
         raise ValueError('connect argument must be "all", "pairs", "finite", or array')
 
-    # Use 0:1 instead of 0 since it behaves when there are no points in the array
-    arr[0:1]['c'] = 0  # the first vertex has no previous vertex to connect
+    arr[0]['c'] = 0  # the first vertex has no previous vertex to connect
 
     # create QDataStream object and stream into QPainterPath
     path.strn = backstore
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1936,7 +1936,7 @@ def arrayToQPath(x, y, connect='all', finiteCheck=True):
     isfinite = None
     if connect == 'finite':
         isfinite = np.isfinite(x) & np.isfinite(y)
-        if not finiteCheck:
+        if not finiteCheck or n == 0:
             # if user specified to skip finite check, then that forces use_qpolygonf
             use_qpolygonf = True
         else:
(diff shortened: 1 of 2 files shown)


ntjess:
I'm definitely not qualified enough in my understanding to figure out what the problem is... #WellItPassedOnMyMachine

pijyoi:
Why not bail out early?
```python
path = QtGui.QPainterPath()
n = x.shape[0]
if n == 0:
    return path
```

But the cause of the failure seems to be from ```ndarray_from_qpolygonf()```. With some testing, it seems that on Qt6, QPolygonF::data() returns a nullptr if empty.
On PySide6, the ```shiboken6.VoidPtr``` gets marked as readonly resulting in a readonly numpy array.
On PyQt6, ```polyline.data()``` returns None instead of a voidptr, resulting in an AttributeError.

```python
In [8]: from PySide6 import QtCore, QtGui, QtWidgets
In [9]: import shiboken6
In [10]: poly = QtGui.QPolygonF()
In [11]: shiboken6.VoidPtr(poly.data(), 0, True)
Out[11]: shiboken6.shiboken6.VoidPtr(0x0000000000000000, 0, False)

In [30]: buffer = shiboken6.VoidPtr(poly.data(), 0, True)
In [31]: np.frombuffer(buffer, np.double).flags.writeable
Out[31]: False
```

```python
In [18]: from PyQt6 import QtCore, QtGui, QtWidgets
In [19]: poly = QtGui.QPolygonF()
In [20]: poly.data() is None
Out[20]: True
```

pijyoi:
You could update ```ndarray_from_qpolygonf()``` as follows.
```python
def ndarray_from_qpolygonf(polyline):
    nbytes = 2 * len(polyline) * 8
    if QT_LIB == "PySide2":
        buffer = Qt.shiboken2.VoidPtr(polyline.data(), nbytes, True)
    elif QT_LIB == "PySide6":
        if polyline.capacity() == 0:
            polyline.reserve(1)
        buffer = Qt.shiboken6.VoidPtr(polyline.data(), nbytes, True)
    else:
        buffer = polyline.data()
        if buffer is None:      # can happen for PyQt6
            # emulate reserve(1)
            polyline.insert(0, QtCore.QPointF())
            polyline.remove(0)
            buffer = polyline.data()
        buffer.setsize(nbytes)
    memory = np.frombuffer(buffer, np.double).reshape((-1, 2))
    return memory
```

ntjess:
@pijyoi early bail-out was my first thought, but I wasn't sure if this was a symptom of a condition that could occur elsewhere too. However, if it is only due to an empty buffer it makes sense to simply check for that condition early.

From a performance perspective, I know when a loop rarely fails a `try` statement is more efficient than an `if` statement. Would you prefer `try`ing to convert the data, or simply enacting early termination?

pijyoi:
I am not sure if your comment refers to arrayToQPath or ndarray_from_qpolygonf.

Early bail out was the easy way out. (It was marked as good first issue after all) Not bailing out early helped to expose the implementation detail difference between Qt5 and Qt6.

As ndarray_from_qpolygonf serves as a compatibility layer for the various bindings, I thought it would be good to fix it. Fixing it here removes the exceptions from being raised.

j9ac9k:
I think I contributed to the confusion here; I did indeed add the "good first issue" thinking an early exit would be the correct solution here; but if we can make ndarray_from_qpolygonf a bit more robust that is definitely ideal.  I suspect there will be more usage of `arrayToQPolygonF` in `PColorMeshItem`; we construct some QPolygonF objects there in a pretty inefficient matter, so I expect that method to get more usage.

pijyoi:
In the same fashion, ```qimage_to_ndarray``` could be made robust against unallocated images. It could also be renamed to ```ndarray_from_qimage``` for consistency. Using ```from``` hopefully hints that no reference to the qimage is being held.

j9ac9k:
Hi @ntjess with #1936 merged, I think we can wrap up this PR with the test code changes and an early exit.  I don't think we should have to mess with modifying the array.  Would you mind trying that and we can see what the test suite does.

PR #1921, opened 2021-07-25
Factor out palette-related code from "named colors" PR

NilsNemitz:
This is work-in-progress code that extracts (and hopes to clean up) the main palette management code from my earlier PR #1625.
Palette colors and UI monochromes are now sampled from color maps, which means that generation of an arbitrary number of plot colors by `functions.intColor()` and the monochrome generation for float values passed to `mkColor()` can be maintained while matching the alternative color schemes.

Once the code is in better shape, it should allow color presets (to be selected before plots are initialized) with less complexity and bugs than the original proposal... and will hopefully be less painful to review :)

### DIFF
--- a/examples/paletteDisplay.py
+++ b/examples/paletteDisplay.py
@@ -1,0 +1,467 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+"""
+Adjust color palettes and apply them to a running plot
+"""
+
+## Add path to library (just for examples; you do not need this)
+import initExample
+
+import numpy as np
+
+from pyqtgraph.Qt import mkQApp, QtCore, QtGui, QtWidgets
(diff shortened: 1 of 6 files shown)


j9ac9k:
Hi @NilsNemitz this PR looks like it needs a rebase given all the reshuffling of stuff we've done over the last few months, sorry for the head-ache!  Let me know if I can help with that effort.

NilsNemitz:
Heh, thanks for the concern, but unfortunately, this is the least of the worries for this PR.

I put this here as a PR to have a basis for discussion, but in all honesty, work has stalled on this a little, and I am still trying to figure out what bits need to go where in order to give this at least a semblance of elegance.

I think we might close this PR for now, keep the numbers looking good.
I will make a new one when I have my squids in a row.

PR #1927, opened 2021-07-27
Reduce ColorMap inefficiencies

NilsNemitz:
This implements some of @pijyoi 's suggestions from the discussion on #1853:

- Gradient generation should ask for data in QColor format to avoid repeated conversion.
- Where conversion is needed, it can be accelerated by explicit `.tolist()` conversion of the numpy data structures.

I have also added a short comment to the generation of gradients and gradient-derived pens and brushes: These are not free operations, and for best performance should only be done as much as needed.

While we could possibly speed this up by caching the result internally, I think the gradient generation is basically a cool toy which realistically sees relatively little use. My impression is that the code for this should remain as simple as possible to save future maintainers some headaches. Or is there an obvious performance optimization that would clearly benefit from integration into the main ColorMap code?

@pijyoi, if you fall over any more improvements, we can throw those in here, too :)

### DIFF
--- a/pyqtgraph/colormap.py
+++ b/pyqtgraph/colormap.py
@@ -605,10 +605,10 @@ class ColorMap(object):
 
         # Convert to QColor if requested
         if mode == self.QCOLOR:
-            if np.isscalar(data):
+            if np.isscalar(data):                
                 return QtGui.QColor(*interp)
             else:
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks @NilsNemitz 

Just wanted to publicly state my appreciation for your work on making the colormap capabilities better and better.

Merging!

PR #1928, opened 2021-07-27
Restore previous signature on TargetItem.setPos()

outofculture:
The previous signature was before 5a08650853a339c383281fd531fe059b74e1bbcd and accepted two args, the same as ever other `UIGraphicsItem.setPos`. This further simplifies to just use the `Point.__init__` to do the type checking.

### DIFF
--- a/pyqtgraph/graphicsItems/TargetItem.py
+++ b/pyqtgraph/graphicsItems/TargetItem.py
@@ -143,29 +143,27 @@ class TargetItem(UIGraphicsItem):
         )
         return self.sigPositionChangeFinished
 
-    def setPos(self, pos):
+    def setPos(self, *args):
         """Method to set the position to ``(x, y)`` within the plot view
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -143,29 +143,27 @@ def sigDragged(self):
...
-        elif isinstance(pos, (QtCore.QPointF, QtCore.QPoint)):
-            newPos = Point(pos.x(), pos.y())
-        else:
-            raise TypeError
+        try:
+            newPos = Point(*args)
+        except TypeError:
+            raise
+        except Exception:
+            raise TypeError(f"Could not make Point from arguments: {args!r}")
```
j9ac9k, line 165: I've never seen this syntax in a f-string before!

j9ac9k:
docs failing!

j9ac9k:
LGTM, thanks for the PR @outofculture !

Issue #1926, opened 2021-07-26
[ENH] Set Antialiasing optional for PlotCurveItem with OpenGL?


marsipu:
## Short Description
For PlotCurveItems plotting with OpenGL enabled, antialiasing is currently always enabled by default. Would it be possible to make it optional?

## Long Description
In `GLGridItem`, the following lines (l.69-72) are depending on the parameter `antialias`:
```python
glEnable(GL_LINE_SMOOTH)
glEnable(GL_BLEND)
glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA)
glHint(GL_LINE_SMOOTH_HINT, GL_NICEST)
```
For `PlotCurveItem` currently the same lines (l.631-643) aren't optional thus antialiasing is always enabled.
I tried commenting them out and it had the desired effect of turning antialiasing off while the plot was still working.

**What do you think about making antialiasing optional here too?
Or is there a specific reason why it has to be always enabled?**

pijyoi:
The code in question was last modified in 2013 2980f8335c1745564e89259a026e83167bb33e96. There probably wasn't any particular reason to it not being a settable option. Why don't you submit a PR to add this feature?

marsipu:
@pijyoi Done:)

j9ac9k:
Think we can close this as #1932 was merged.  Thanks @marsipu

PR #1932, opened 2021-07-29
make antialiasing optional for paintGL in PlotCurveItem

marsipu:
## Summary
This PR makes antialiasing optional when OpenGL is used with PlotCurveItem as described in #1926.

I checked for `antialias` in `self._exportOpts` as it is done in `paint`. Is this necessary for this to be inside paint/paintGL or could it also be in e.g. `updateData` to improve performance a little bit?

### DIFF
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -641,9 +641,6 @@ class PlotCurveItem(GraphicsObject):
                     gl.glHint(gl.GL_LINE_SMOOTH_HINT, gl.GL_NICEST)
                 else:
                     gl.glDisable(gl.GL_LINE_SMOOTH)
-                    gl.glDisable(gl.GL_BLEND)
-                    gl.glBlendFunc(gl.GL_ONE, gl.GL_ZERO)
-                    gl.glHint(gl.GL_LINE_SMOOTH_HINT, gl.GL_DONT_CARE)
 
                 gl.glDrawArrays(gl.GL_LINE_STRIP, 0, int(pos.size / pos.shape[-1]))
             finally:
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -628,10 +628,18 @@ class PlotCurveItem(GraphicsObject):
                     width = 1
                 gl.glPointSize(width)
                 gl.glLineWidth(width)
-                gl.glEnable(gl.GL_LINE_SMOOTH)
-                gl.glEnable(gl.GL_BLEND)
-                gl.glBlendFunc(gl.GL_SRC_ALPHA, gl.GL_ONE_MINUS_SRC_ALPHA)
-                gl.glHint(gl.GL_LINE_SMOOTH_HINT, gl.GL_NICEST)
+
+                # enable antialiasing if requested
+                if self._exportOpts is not False:
+                    aa = self._exportOpts.get('antialias', True)
+                else:
+                    aa = self.opts['antialias']
+                if aa:
+                    gl.glEnable(gl.GL_LINE_SMOOTH)
+                    gl.glEnable(gl.GL_BLEND)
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -628,10 +628,18 @@ def paintGL(self, p, opt, widget):
...
+                if self._exportOpts is not False:
+                    aa = self._exportOpts.get('antialias', True)
+                else:
+                    aa = self.opts['antialias']
+                if aa:
+                    gl.glEnable(gl.GL_LINE_SMOOTH)
+                    gl.glEnable(gl.GL_BLEND)
+                    gl.glBlendFunc(gl.GL_SRC_ALPHA, gl.GL_ONE_MINUS_SRC_ALPHA)
+                    gl.glHint(gl.GL_LINE_SMOOTH_HINT, gl.GL_NICEST)
+
```
pijyoi, line 642: If aa is False, you would need to glDisable whatever got enabled when aa is True.
I tried out this PR with #1910 and toggling AA from False->True->False would get a plot different from the initial plot.

marsipu: Thank you for the hint, I haven't thought of that. I am no OpenGL-expert so I just reverted to the default-values from the OpenGL-documentation. Now this lets me toggle between aa-states in your (really nice) example. Does this do now what you would expect?

pijyoi: There are 3 other places that setup GL_LINE_SMOOTH: GLAxisItem, GLGridItem, GLLinePlotItem. GLAxisItem doesn't enable blending, but would probably get it enabled by GLGridItem. A proper fix might require all GLGraphicsItem(s) to disable everything that they have enabled before they exit their paint.

I would suggest to leave them as is. Don't modify things that we aren't able to test ourselves. Or raise a separate issue.

pijyoi: Probably after painting with antialiasing on, GL_LINE_SMOOTH should be disabled.
If we take a peek at: https://code.woboq.org/qt6/qtbase/src/opengl/qopenglpaintengine.cpp.html#_ZN28QOpenGL2PaintEngineExPrivate12resetGLStateEv
GL_LINE_SMOOTH happens to be one of the things that Qt library doesn't disable for us.
On the other hand, GL_BLEND is disabled for us, so we didn't actually need to disable it for non-antialiased mode.

marsipu: Ok, so should I remove those lines again?
```python
gl.glDisable(gl.GL_BLEND)
gl.glBlendFunc(gl.GL_ONE, gl.GL_ZERO)
gl.glHint(gl.GL_LINE_SMOOTH_HINT, gl.GL_DONT_CARE)
```

marsipu: That seems to work too, but performance doesn't seem to be influenced by just removing the lines. Do you think we could introduce an attribute like `_aa_state' to PlotCurveItem which then calls the `glEnable`/`glDisable`-functions only when the state changes?

pijyoi: The problem is that when you have called ```beginNativePainting()```, Qt makes little guarantee about the state. So things that we need, we need to set again for each frame. E.g. What would happen if we created 2 PlotWidgets in a single QMainWindow, one with antialiasing enabled and the other disabled? I would guess that because of the shared OpenGL context, they would see the last set setting of GL_LINE_SMOOTH.

marsipu: So should I just remove the above lines to have a bit less to call when AA is disabled?

pijyoi: Yes. Based on my understanding, the only line needed would be
```python
gl.glDisable(gl.GL_LINE_SMOOTH)
```
Btw, it's not so much for performance to reduce the number of lines here. Code like this tends to get copied around, so if we can, we keep the code to only the bits necessary.

j9ac9k:
Thanks for the PR @marsipu  and thanks for the review/feedback @pijyoi !

PR #1935, opened 2021-07-31
Many unused import cleanups

ksunden:
Ignored some star imports, some vendored code in colorama, only looked within pyqtgraph the library, not e.g. examples

Should clean up a good portion of Code QL complaints


j9ac9k:
Thanks for the PR @ksunden 

We've largely used the static code checker to check for issues involving diffs from PRs but few changes have been made with the intention of knocking down the total count of issues flagged by the static checker, thanks for taking this on!

Issue #1882, opened 2021-07-08
"from PyQt5 import sip" fails on Ubuntu 20.04


pijyoi:
Trying to run pyqtgraph using Ubuntu's distro-packaged PyQt5 (5.14.x) fails due to failure of
```python
from PyQt5 import sip, uic
```

Background:
Since PyQt5 5.11, (https://www.riverbankcomputing.com/static/Docs/PyQt5/incompatibilities.html) ```from PyQt5 import sip``` has been the documented way to import sip module, rather than the old ```import sip```.
However, Ubuntu 20.04 somehow packages it such that only ```import sip``` works.
https://packages.ubuntu.com/focal/amd64/python3-sip/filelist

Searching around, this was reported in https://github.com/pytest-dev/pytest-qt/issues/369 and https://github.com/pytest-dev/pytest-qt/pull/370

The question is whether pyqtgraph should be working around this.

j9ac9k:
ahh, yeah we don't test on the bistro-packaged versions of the bindings; but we should certainly try and support that, especially if we know what the fix looks should be and it is relatively non-invasive.

pijyoi:
Just slightly related, I have a Windows Anaconda installation that comes by default with PyQt5 5.9.2 Qt 5.9.7, nothing installed from conda-forge. I noticed that the enums were in fact accessible using the long spelling (#1818). So just to see how much was still working, I tried running the test suite.

Modifications needed to ```Qt/__init__.py```:
1) change ```from PyQt5 import sip``` to ```import sip```
2) bypass the Qt >= 5.12 check

Result:
1 test failed: test_ImageItemFormat.py::test_uint16
Due to use of Format_RGBA64 available only since Qt 5.12. Minor, since that's an uncommon usage.

j9ac9k:
This has come up enough times now I'm thinking we should make the version
check be a warning; letting users proceed, making it clear we don't
support/test < 5.12 (and conda-forge rolls out qt 5.15 we may make that the
minimum version)

This topic has also brought up discussions of documenting
`PYQTGRAPH_QT_LIB`; and also maybe redoing the preferred order of the
library imports.

PR #1936, opened 2021-07-31
Handle empty QPolygonF

pijyoi:
#1920 exposed that ```ndarray_from_qpolygonf()``` did not handle empty QPolygonF on Qt6. Specifically, it did not handle the return of a null pointer. i.e the Qt library had not allocated any underlying memory yet.

Some included cleanups:
- import shiboken{2,6} as shiboken. This allows to merge some PySide{2,6} codepaths
- ```import sip``` if ```import PyQt5.sip``` fails. This addresses #1882

Additional changes:
- make qimage_to_ndarray() support non-tightly packed QImage
- add QImage zero-copy and write-through test from https://github.com/pyqtgraph/pyqtgraph/pull/1795#issuecomment-853084414

### DIFF
--- a/examples/PlotSpeedTest.py
+++ b/examples/PlotSpeedTest.py
@@ -17,12 +17,10 @@ import pyqtgraph.functions as fn
 import itertools
 import argparse
 
-if QT_LIB == 'PySide2':
-    wrapinstance = pg.Qt.shiboken2.wrapInstance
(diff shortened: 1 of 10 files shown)


j9ac9k:
Looks good to me @pijyoi

Thanks for the extra tests!

PR #1919, opened 2021-07-24
Organize paramtypes

ntjess:
To prevent the extraordinary merge hassle of last parameter type PR, this organization:
* Makes a new file for each parameter type
* Allows parameter items to be registered without belonging to a parameter
* Splits `WidgetParameterItem` implementations into actually separate items like `BoolParameterItem`, `ColorParameterItem`, etc.
* A script to auto-generate parameter types documentation

For reviewers: The content of each Parameter/Item is unchanged -- the only modifications are to the imports

PR #1939, opened 2021-07-31
Remove all usage of python2_3.py

ksunden:
Technically these functions were exported at the top level of the library, this removes them without warning... If we want to we can bring them back for there and add deprecation warnigns, but I honestly don't think its needed, as we are py3 only now and have been for multiple releases.

This may introduce a number of 'useless cast' or similar but those were always happening anyway

This PR brought to you by `sed`


j9ac9k:
Nice!

On mobile so only skimmed a small portion of the diff.  Saw some cases assigned `range = ...` which we probably shouldn't assign Python keywords to variables

pyqtgraph/Qt/__init__.py
```diff
@@ -73,7 +72,7 @@ def write(self, data):
         self.data.append(data)
         
     def getvalue(self):
-        return ''.join(map(asUnicode, self.data)).encode('utf8')
+        return ''.join(map(str, self.data)).encode('utf8')
```
j9ac9k, line 75: we can remove the `.encode('utf8')` bit, as that's default behavior.

pyqtgraph/configfile.py
```diff
@@ -57,7 +56,7 @@ def readConfigFile(fname):
     try:
         #os.chdir(newDir)  ## bad.
         with open(fname) as fd:
-            s = asUnicode(fd.read())
+            s = str(fd.read())
```
j9ac9k, line 59: do we need to specify `str` here, we may not if we can modify the context manager to open with text mode.

```python
with open(fname, 'rt') as fd:
   s = fd.read()
```

pyqtgraph/console/CmdInput.py
```diff
@@ -25,10 +25,10 @@ def keyPressEvent(self, ev):
             self.execCmd()
         else:
             super().keyPressEvent(ev)
-            self.history[0] = asUnicode(self.text())
+            self.history[0] = str(self.text())
```
j9ac9k, line 28: fairly sure we can ditch `str` casting here.

pyqtgraph/dockarea/Dock.py
```diff
@@ -125,7 +124,7 @@ def title(self):
         """
         Gets the text displayed in the title bar for this dock.
         """
-        return asUnicode(self.label.text())
+        return str(self.label.text())
```
j9ac9k, line 127: can ditch `str` casing here.

pyqtgraph/exporters/Exporter.py
```diff
@@ -63,13 +62,13 @@ def fileSaveDialog(self, filter=None, opts=None):
...
     def fileSaveFinished(self, fileName):
-        fileName = asUnicode(fileName)
+        fileName = str(fileName)
         global LastExportDirectory
         LastExportDirectory = os.path.split(fileName)[0]
         
         ## If file name does not match selected extension, append it now
         ext = os.path.splitext(fileName)[1].lower().lstrip('.')
-        selectedExt = re.search(r'\*\.(\w+)\b', asUnicode(self.fileDialog.selectedNameFilter()))
+        selectedExt = re.search(r'\*\.(\w+)\b', str(self.fileDialog.selectedNameFilter()))
```
j9ac9k, line 71: can ditch `str` casting here too.

pyqtgraph/functions.py
```diff
@@ -64,7 +63,7 @@
     's': QtGui.QColor(100,100,150,255),
 }  
 
-SI_PREFIXES = asUnicode('yzafpnµm kMGTPEZY')
+SI_PREFIXES = str('yzafpnµm kMGTPEZY')
```
j9ac9k, line 66: can ditch str casting here.

ksunden:
literally all I did was `sed -i s/asUnicode/str/g` and similar , not surprised there are some unnecessary casts...

pyqtgraph/Qt/__init__.py
ksunden: is it? encode converts it from str to bytes

j9ac9k: oh, right; yeah disregard me here!

```
$ python
Python 3.9.6 (default, Jul  1 2021, 09:28:20)
[Clang 12.0.5 (clang-1205.0.22.11)] on darwin
>>> test = "123"
>>> test.encode("utf8")
b'123'
```

j9ac9k:
my only comment is regarding the unnecessary casts... besides that, this LGTM.

pyqtgraph/functions.py
```diff
@@ -163,7 +162,7 @@ def siParse(s, regex=FLOAT_REGEX, suffix=None):
         contains a suffix, it is discarded. This enables interpreting
         characters following the numerical value as an SI prefix.
     """
-    s = asUnicode(s)
+    s = str(s)
```
j9ac9k, line 165: `s` is expected to be a string already, we can remove this line altogether.

pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -1070,7 +1068,7 @@ def generateDrawSpecs(self, p):
                 if s is None:
                     rects.append(None)
                 else:
-                    br = p.boundingRect(QtCore.QRectF(0, 0, 100, 100), QtCore.Qt.AlignmentFlag.AlignCenter, asUnicode(s))
+                    br = p.boundingRect(QtCore.QRectF(0, 0, 100, 100), QtCore.Qt.AlignmentFlag.AlignCenter, str(s))
```
j9ac9k, line 1071: `strings` can either be `None` or of type `str`, since we already filter out the `None` we can ditch this casting too.

pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -1112,7 +1110,7 @@ def generateDrawSpecs(self, p):
                 vstr = strings[j]
                 if vstr is None: ## this tick was ignored because it is out of bounds
                     continue
-                vstr = asUnicode(vstr)
+                vstr = str(vstr)
```
j9ac9k, line 1113: As earlier, we already filter out the `None` case, leaving us with the `str` case, so we can remove this line altogether.

pyqtgraph/graphicsItems/ViewBox/ViewBoxMenu.py
```diff
@@ -238,7 +237,7 @@ def setViewList(self, views):
             
         for i in [0,1]:
             c = self.ctrl[i].linkCombo
-            current = asUnicode(c.currentText())
+            current = str(c.currentText())
```
j9ac9k, line 240: since `c` is a `QComboBox` we can be sure that `QComboBox.currentText()` returns a `str` so we can ditch this casting.

pyqtgraph/metaarray/MetaArray.py
```diff
@@ -469,7 +468,7 @@ def _interpretIndexes(self, ind):
         return tuple(nInd)
       
     def _interpretAxis(self, axis):
-        if isinstance(axis, basestring) or isinstance(axis, tuple):
+        if isinstance(axis, str) or isinstance(axis, tuple):
```
j9ac9k, line 471: I know not your change, but this one we can simplify to

```python
if isinstance(axis, (str, tuple)):
```

pyqtgraph/parametertree/parameterTypes.py
```diff
@@ -231,11 +231,11 @@ def updateDisplayLabel(self, value=None):
             value = self.param.value()
         opts = self.param.opts
         if isinstance(self.widget, QtWidgets.QAbstractSpinBox):
-            text = asUnicode(self.widget.lineEdit().text())
+            text = str(self.widget.lineEdit().text())
```
j9ac9k, line 234: We can be assured this will return a `str` as well: http://doc.qt.io/qt-5/qlineedit.html#text-prop

pyqtgraph/widgets/ComboBox.py
```diff
@@ -64,7 +63,7 @@ def value(self):
         """
         if self.count() == 0:
             return None
-        text = asUnicode(self.currentText())
+        text = str(self.currentText())
```
j9ac9k, line 66: `QComboBox.currentText()` can have `str` casting removed too...

pyqtgraph/widgets/SpinBox.py
```diff
@@ -298,7 +297,7 @@ def selectNumber(self):
         Select the numerical portion of the text to allow quick editing by the user.
         """
         le = self.lineEdit()
-        text = asUnicode(le.text())
+        text = str(le.text())
```
j9ac9k, line 300: `QLineEdit::text()` will return a string too: http://doc.qt.io/qt-5/qlineedit.html#text-prop

pyqtgraph/widgets/ComboBox.py
```diff
@@ -131,7 +130,7 @@ def indexChanged(self, index):
         # current index has changed; need to remember new 'chosen text'
         if self._ignoreIndexChange:
             return
-        self._chosenText = asUnicode(self.currentText())
+        self._chosenText = str(self.currentText())
```
j9ac9k, line 133: `QComboBox::currentText()` can remove `str` casting

pyqtgraph/widgets/ComboBox.py
```diff
@@ -230,7 +229,7 @@ def saveState(self):
             except AttributeError:
                 pass
         if data is None:
-            return asUnicode(self.itemText(ind))
+            return str(self.itemText(ind))
```
j9ac9k, line 232: QComboBox::itemText(index) returns a string as well, so we can remove the casting here too

pyqtgraph/widgets/TableWidget.py
```diff
@@ -311,22 +310,22 @@ def serialize(self, useSelection=False):
         if self.horizontalHeadersSet:
             row = []
             if self.verticalHeadersSet:
-                row.append(asUnicode(''))
+                row.append('')
             
             for c in columns:
-                row.append(asUnicode(self.horizontalHeaderItem(c).text()))
+                row.append(str(self.horizontalHeaderItem(c).text()))
```
j9ac9k, line 316: `QTableWidgetIem::text` returns a string, so we can ditch the casting here as well

tests/widgets/test_spinbox.py
```diff
@@ -30,7 +30,7 @@ def test_SpinBox_formatting(value, expected_text, opts):
     sb.setValue(value)
 
     assert sb.value() == value
-    assert pg.asUnicode(sb.text()) == expected_text
+    assert str(sb.text()) == expected_text
```
j9ac9k, line 33: QSpinBox::text returns `str` type as well.

j9ac9k:
Alright, I think I'm done 😬  had to do a lot of doc lookups for the Qt stuff.... so thankful for Dash docs letting me lookup the stuff insta-fast.

j9ac9k:
@ksunden thanks for the PR, this LGTM, merging!

pyqtgraph/parametertree/parameterTypes.py
```diff
@@ -146,8 +146,8 @@ def makeWidget(self):
             w = QtWidgets.QLineEdit()
             w.setStyleSheet('border: 0px')
             w.sigChanged = w.editingFinished
-            w.value = lambda: asUnicode(w.text())
-            w.setValue = lambda v: w.setText(asUnicode(v))
+            w.value = w.text
+            w.setValue = w.setText
```
ntjess, line 150: @j9ac9k I'm almost inclined to leave `setValue` as casting to string. While you can never get a non-string from `text()`, you can certainly set with e.g. an `int` and this would safely convert. What are your thoughts? It plays into #1919 merging.

ntjess: The new definition of `_interpretValue()` in 1919 should handle a call to `setValue(10)`, but this would also work with `item.widget.setValue(10)`

ntjess: Eh, the widget should inherently function as a text setter, so I'm fine with leaving it as `w.setValue = w.setText`.

j9ac9k: Feel free to correct things in #1919 as you see fit, again, sorry for the trouble!

Our PR count is low enough we're getting more aggressive about small changes in lots of the library... this is going to keep happening until we effectively zero out the PR queue :(   again, sorry for the headache!

PR #1940, opened 2021-08-01
fix log mode by reverting to previous formulation

### LINKED ISSUE #1933: Log mode broken in 0.12.2
### Short description
Log mode shows wrong sign in the exponent, completely wrong numbers, and inverted axis.
Everything works as expected in 0.12.1

### Code to reproduce
```
import pyqtgraph as pg

app = pg.mkQApp()
pw = pg.PlotWidget(enableMenu=False)

pw.plotItem.setLogMode(x=False, y=True)
x = range(8)
y = [1e-6,2e-6,3e-6,4e-6,5e-6,6e-6,7e-6,8e-6]
pw.plot(x, y, symbol='o')

pw.show()
# Start Qt event loop unless running in interactive mode or using pyside.
if __name__ == '__main__':
    import sys
```

### Expected behavior
logarithmic y axis

### Real behavior
Wrong sign in the exponent and inverted scale

### Tested environment(s)

 * PyQtGraph version: 0.12.2
 * Qt Python binding: pyqt5
 * Python version: 3.8

pijyoi:
this formulation is in the documentation for setLogMode() https://pyqtgraph.readthedocs.io/en/latest/graphicsItems/plotdataitem.html?highlight=setlogmode#pyqtgraph.PlotDataItem.setLogMode

fixes #1933 

Powers of 10 should plot a straight line
```python
import pyqtgraph as pg
pg.mkQApp()
pw = pg.PlotWidget()
pw.show()
pi = pw.getPlotItem()
pi.setLogMode(x=False, y=True)
x = range(-10, 11)
y = [(10.**p) for p in x]
pi.plot(x, y, symbol='o')
pg.exec()
```

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -701,7 +701,7 @@ class PlotDataItem(GraphicsObject):
                     eps = np.finfo(y.dtype).eps
                 else:
                     eps = 1
-                y = np.copysign(np.log10(np.abs(y)+eps), y)
+                y = np.sign(y) * np.log10(np.abs(y)+eps)
 
         ds = self.opts['downsample']
         if not isinstance(ds, int):


j9ac9k:
ooof... you would think, give my comment on the original PR for that implementation (about being interested in the range between 0 and 1, I would have recognized the difference between `np.sign(y) * np.log10(....)` and `np.copysign(np.log10(...), y)`

thanks for patching up my mistake @pijyoi, this LGTM, merging.

simonzihlmann:
May I ask why you have chosen `y = np.sign(y) * np.log10(np.abs(y)+eps)` to handle negative values?

For me this does not make sense. The logarithm of a negative number is not defined first of all. Then to go around that problem, I kind of agree to take the absolute value of the number and then to take the log. Because this tells you how "big" a number is. Multiplying with the sign after taking the log does not make sense at all since there is a lot of information in the sign of the output of the log. It tells you whether the number was bigger or smaller than the base 10.

Let me do a small example.
y = 1e2
--> in log mode: y = 1*(2+eps) ~2 (neglecting eps)

y = -1e-5
--> in log modde: y = -1*(-5+eps)~5 (neglecting eps)

Therefore, -1e5 appears to be 3 orders of magnitude bigger than 1e2 even though the absolute value is 7 orders of magnitude smaller. Therefore,  `y = np.sign(y) * np.log10(np.abs(y)+eps)` leads to very strange plots that are really misleading if very small negative numbers are involved.

I would propose to use `y = np.log10(np.abs(y)+eps)`

I am also questioning the use of `eps`. This could also lead to strange behaviors. Why not setting the log of y to `np.nan` if it is zero? Strictly speaking, log(0)-->-inf, which cannot be plotted.

pijyoi:
May I suggest that you put your queries into an issue / discussion?

I myself may not agree with all aspects of the formulation (both sign and eps), but the purpose of this PR was to revert the formulation in response to reports of breakage with respect to 0.12.1.

simonzihlmann:
Sure, I will open an issue then. Thanks a lot for the fast response and let's discuss it there.

Issue #895, opened 2019-05-25
[CI-fail] Segfault with pytest --cov


ixjlyons:
Moving some notes over from slack:

We're seeing a segmentation fault when running `pytest --cov pyqtgraph` both locally and in CI. It occurs right after the tests are completed (just before the coverage report is produced).

Environment:

- PySide2 5.12
- Python 3.7
- Either venv or conda
- pytest-cov 2.7

I noticed this occurs when using bash but *not* fish. @j9ac9k reported the segfault with zsh as well.

Running `coverage run -m pytest` does *not* result in a segfault.

`pytest-cov`'s README says this:

> Consistent pytest behavior. If you run `coverage run -m pytest` you will have slightly different `sys.path` (CWD will be in it, unlike when running `pytest`).

Edit: added pytest-cov version to list

j9ac9k:
Also should add that on Python 2.7, I am unable to replicate a segfault

j9ac9k:
A good first step may be run `pytest --cov` with a very limited test set and output `PATH` and see what the difference is.

ixjlyons:
I just printed `sys.path` and `os.environ['PATH']` from within `pytest_cov.engine.summary` and I'm not seeing a difference between bash/fish. Also, the fact that it's happening with PySide2 and not PyQt5 seems suspiciously like it has more to do with subprocess management than paths.

Putting this here though I haven't fully read/understood it: https://pytest-cov.readthedocs.io/en/latest/subprocess-support.html

ixjlyons:
Another note: adding `--collect-only` lets the coverage report through without segfault: `pytest --collect-only --cov pyqtgraph`

j9ac9k:
The "custom signal handling" section in the link you provided has some promising examples.

ixjlyons:
Ok really scratching my head here now. I get no segfaults if I run each test package separately:

```bash
#!/bin/bash

for d in $(find pyqtgraph -name tests -type d)
do
    pytest --cov pyqtgraph $d
done
```

j9ac9k:
Think it's time to cross post an issue to the pytest-cov repo?

ixjlyons:
Yeah they may have insight. Might also be worth getting this running with gdb to get a backtrace.

hhslepicka:
Give it a try with the following requirements:

pytest>=3.6
pytest-cov<2.6.0

[...]

PR #1943, opened 2021-08-02
Remove python 2 code paths

ntjess:
A search for `sys.version_info` showed several locations still checking for Python 2. This PR removes those branches and assumes Python 3 throughout.

### DIFF
--- a/pyqtgraph/__init__.py
+++ b/pyqtgraph/__init__.py
@@ -129,9 +129,6 @@ def renamePyc(startDir):
                 os.rename(fileName, name2)
                 
 path = os.path.split(__file__)[0]
-if __version__ is None and not hasattr(sys, 'frozen') and sys.version_info[0] == 2: ## If we are frozen, there's a good chance we don't have the original .py files anymore.
-    renamePyc(path)
-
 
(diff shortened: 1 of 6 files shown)


ksunden:
https://github.com/asottile/pyupgrade will find a lot of these (and convert a lot of `%` format to f strings/`.format`)

It was going to be something I looked to do soon.

ksunden (approved):

j9ac9k:
> https://github.com/asottile/pyupgrade will find a lot of these (and convert a lot of `%` format to f strings/`.format`)
> 
> It was going to be something I looked to do soon.

Just want to point out [@asottile has been a friend to this project](https://github.com/pyqtgraph/pyqtgraph/issues/895#issuecomment-495968175), while not a direct contributor, he has assisted with troubleshooting some of the most difficult of issues to troubleshoot in the past; while live on his twitch stream.  Still super appreciative of your time/efforts!

j9ac9k:
would definitely encourage folks to let pyupgrade handle f-string conversions as if we tried to do that manually, we would be here all day...

Thanks for the PR @ntjess !

PR #1944, opened 2021-08-02
use more of QColor functions / methods

pijyoi:
This PR makes more use of QColor functions / methods
- static QColor.fromXXX()
- QColor.getRgbF()
- QColor.name()

Several helper functions in pg.functions are effectively trivial reimplementations of the above functions / methods
- class Color
- colorTuple
- colorStr
- glColor

The following line was left unconverted because it seems that the format string does not match the data types. (%f to integer, %d to float [0, 1])
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/exporters/SVGExporter.py#L122

### DIFF
--- a/pyqtgraph/colormap.py
+++ b/pyqtgraph/colormap.py
@@ -249,8 +249,7 @@ def makeHslCycle( hue=0.0, saturation=1.0, lightness=0.5, steps=36 ):
     lgt_vals = np.linspace(lgtA, lgtB, num=steps+1)
     color_list = []
     for hue, sat, lgt in zip( hue_vals, sat_vals, lgt_vals):
-        qcol = QtGui.QColor()
-        qcol.setHslF( hue%1.0, sat, lgt )
+        qcol = QtGui.QColor.fromHslF( hue%1.0, sat, lgt )
         color_list.append( qcol )
     name = f'Hue {hueA:0.2f}-{hueB:0.2f}'
     return ColorMap( None, color_list, name=name )
@@ -311,8 +310,7 @@ def makeMonochrome(color='neutral'):
     l_vals = np.linspace(l_min, l_max, num=16)
     color_list = []
     for l_val in l_vals:
-        qcol = QtGui.QColor()
-        qcol.setHslF( h_val, s_val, l_val )
+        qcol = QtGui.QColor.fromHslF( h_val, s_val, l_val )
         color_list.append( qcol )
     return ColorMap( None, color_list, name=name, linearize=True )
 
@@ -581,7 +579,7 @@ class ColorMap(object):
             mode = self.enumMap[mode.lower()]
 
         if mode == self.QCOLOR:
-            pos, color = self.getStops(self.BYTE)
+            pos, color = self.getStops(self.FLOAT)
         else:
             pos, color = self.getStops(mode)
 
@@ -606,9 +604,9 @@ class ColorMap(object):
         # Convert to QColor if requested
         if mode == self.QCOLOR:
             if np.isscalar(data):
-                return QtGui.QColor(*interp)
+                return QtGui.QColor.fromRgbF(*interp)
             else:
-                return [QtGui.QColor(*x.tolist()) for x in interp]
+                return [QtGui.QColor.fromRgbF(*x.tolist()) for x in interp]
         else:
             return interp
 
@@ -762,8 +760,10 @@ class ColorMap(object):
                 color = color.astype(float) / 255.
             elif mode == self.QCOLOR:
                 if color.dtype.kind == 'f':
-                    color = (color*255).astype(np.ubyte)
-                color = [QtGui.QColor(*x.tolist()) for x in color]
+                    factory = QtGui.QColor.fromRgbF
+                else:
+                    factory = QtGui.QColor.fromRgb
+                color = [factory(*x.tolist()) for x in color]
             self.stopsCache[mode] = (self.pos, color)
         return self.stopsCache[mode]
 
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -763,9 +763,8 @@ class PlotItem(GraphicsWidget):
 
             for item in self.curves:
                 if isinstance(item, PlotCurveItem):
-                    color = fn.colorStr(item.pen.color())
-                    opacity = item.pen.color().alpha() / 255.
-                    color = color[:6]
+                    color = item.pen.color()
+                    hrrggbb, opacity = color.name(), color.alphaF()
                     x, y = item.getData()
                     mask = (x > xRange[0]) * (x < xRange[1])
                     mask[:-1] += mask[1:]
@@ -780,9 +779,9 @@ class PlotItem(GraphicsWidget):
                     # fh.write('<g fill="none" stroke="#%s" '
                     #          'stroke-opacity="1" stroke-width="1">\n' % (
                     #           color, ))
-                    fh.write('<path fill="none" stroke="#%s" '
+                    fh.write('<path fill="none" stroke="%s" '
                              'stroke-opacity="%f" stroke-width="1" '
-                             'd="M%f,%f ' % (color, opacity, x[0], y[0]))
+                             'd="M%f,%f ' % (hrrggbb, opacity, x[0], y[0]))
                     for i in range(1, len(x)):
                         fh.write('L%f,%f ' % (x[i], y[i]))
 
@@ -798,15 +797,14 @@ class PlotItem(GraphicsWidget):
                         pos = point.pos()
                         if not rect.contains(pos):
                             continue
-                        color = fn.colorStr(point.brush.color())
-                        opacity = point.brush.color().alpha() / 255.
-                        color = color[:6]
+                        color = point.brush.color()
+                        hrrggbb, opacity = color.name(), color.alphaF()
                         x = pos.x() * sx
                         y = pos.y() * sy
 
-                        fh.write('<circle cx="%f" cy="%f" r="1" fill="#%s" '
+                        fh.write('<circle cx="%f" cy="%f" r="1" fill="%s" '
                                  'stroke="none" fill-opacity="%f"/>\n' % (
-                                    x, y, color, opacity))
+                                    x, y, hrrggbb, opacity))
 
             fh.write("</svg>\n")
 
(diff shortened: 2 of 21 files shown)


pyqtgraph/colormap.py
```diff
@@ -606,9 +604,9 @@ def map(self, data, mode=BYTE):
         # Convert to QColor if requested
         if mode == self.QCOLOR:
             if np.isscalar(data):
-                return QtGui.QColor(*interp)
+                return QtGui.QColor.fromRgbF(*interp)
             else:
-                return [QtGui.QColor(*x.tolist()) for x in interp]
+                return [QtGui.QColor.fromRgbF(*x.tolist()) for x in interp]
         else:
```
NilsNemitz, line 610: Thank you, that is very much what I had in mind!

I am glad that you implemented this; I am sorry that I didn't find any time to add this after you pointed out the factory functions.

pyqtgraph/colormap.py
```diff
@@ -762,8 +760,10 @@ def getStops(self, mode=BYTE):
                 color = color.astype(float) / 255.
             elif mode == self.QCOLOR:
                 if color.dtype.kind == 'f':
-                    color = (color*255).astype(np.ubyte)
-                color = [QtGui.QColor(*x.tolist()) for x in color]
+                    factory = QtGui.QColor.fromRgbF
+                else:
+                    factory = QtGui.QColor.fromRgb
+                color = [factory(*x.tolist()) for x in color]
```
NilsNemitz, line 766: neat :)

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -763,9 +763,8 @@ def writeSvgCurves(self, fileName=None):
 
             for item in self.curves:
                 if isinstance(item, PlotCurveItem):
-                    color = fn.colorStr(item.pen.color())
-                    opacity = item.pen.color().alpha() / 255.
-                    color = color[:6]
+                    color = item.pen.color()
+                    hrrggbb, opacity = color.name(), color.alphaF()
```
NilsNemitz, line 767: Out of curiosity, what does the `h` in `hrrggbb` stand for?

pijyoi: hex or hash. The name() method returns a str of the form '#rrggbb'. It's to remind that a hex or hash has already been prepended.

pyqtgraph/opengl/items/GLMeshItem.py
```diff
@@ -181,7 +181,7 @@ def paint(self):
                     if self.colors is None:
                         color = self.opts['color']
                         if isinstance(color, QtGui.QColor):
-                            glColor4f(*fn.glColor(color))
+                            glColor4f(*color.getRgbF())
```
NilsNemitz, line 184: As the code checker points out, it seems you have successfully eliminated the dependency on pg.functions !

pijyoi: There are inconsistencies with the opengl items' APIs. Some take in float colors, some take in an argument for mkColor.

j9ac9k:
@pijyoi thank you for this!  Been thinking there was likely room to implement the QColor methods a bit more directly.

The one gotcha, which it looks like you've taken care to avoid, is that `QColor::setNamedColor` wants `#AARRGGBB` (if you're going to specify alpha), whereas pyqtgraph expects `#RRGGBBAA`.

As you pointed out, we have some inconsistencies across the board for the opengl item APIs; not sure if trying to modify those would be part of this PR or not (I'm not opposed to it).

j9ac9k:
Also I'm totally good with putting deprecation warnings (or should we be putting future warnings, I'm always uncertain) on some of the stuff that's rendered more or less useful (like `Color` class, `colorTuple`, `colorStr` and `glColor`).

NilsNemitz:
@pijyoi , thank you for this clean-up! I also see this as a great improvement, since it removes a lot of redundancy and alternative definitions in favor of simply using the Qt conventions.

Fortunately it turns out that the data formats are mostly compatible; Much more so than I would have expected. That makes the now redundant function definitions so simple that there is not much cost* to keeping them. They actually serve as excellent demonstrations of direct replacement code - for those who look at the source.
(* for "cost" I would consider potential for user and developer confusion most important; this seems near zero here.)

My suggestion would be to take a quick look over the documentation for the colors:
The first step to getting the legacy functions out of user code would be to no longer advertise them, and to make sure that alternatives are easy to find... This might include some hint towards the (somewhat hidden) Qt factory functions.

A simple approach might be to move the now-redundant set of functions to a "legacy functions" section of the page for now and basically copy-and-paste in an example of the equivalent Qt dialect as a suggested replacement?

pijyoi:
>As you pointed out, we have some inconsistencies across the board for the opengl item APIs; not sure if trying to modify those would be part of this PR or not (I'm not opposed to it).

From what I can see, the various items can be grouped thus:

passed to mkColor
- GLBoxItem
- GLGridItem
- GLTextItem (although only documented to take an array of R,G,B,A)

2D-ndarray of float colors if different color for each element
4-tuple of float or QColor if single color for every element
- GLMeshItem
- GLScatterPlotItem
- GLSurfacePlotItem
- GLLinePlotItem (in addition takes a str passed to mkColor for scalar (undocumented))

GLGradientLegendItem
- fontColor : 4-tuple of float only

The only glaring inconsistencies are:
- GLGradientLegendItem. This looks like a bug. It should be changed to take ```mkColor()```
   * Unless you consider GLGradientLegendItem to be "correct" and GLBoxItem, GLGridItem, GLTextItem to be "wrong"
- GLLinePlotItem taking a ```str``` overload. Probably should just leave it as is.

NilsNemitz:
As stated, I have some tolerance for just keeping the old definitions around without aggressive deprecation.

But I do believe we should deprecate the `#RRGGBBAA` format as soon as possible. There are many other options to specify alpha, and I really hope that this is not so commonly used.

Once we have officially removed this usage, we can re-add the Qt compatible `#AARRGGBB`. 
I am sure there are good reasons for one or the other, but after this PR is merged, PyQtGraph will do color "the Qt way" basically everywhere. And that places us -at least from my point of view- very firmly on the `#AARRGGBB` side of the fence.

It would also remove at least one case distinctions from the painfully long list in mkColor. 🤯

j9ac9k:
> Once we have officially removed this usage, we can re-add the Qt compatible `#AARRGGBB`.
> I am sure there are good reasons for one or the other, but after this PR is merged, PyQtGraph will do color "the Qt way" basically everywhere. And that places us -at least from my point of view- very firmly on the `#AARRGGBB` side of the fence.

This change is a tough sell IMO.  There is no good way of rolling out this change without silently breaking code for existing users.  I suppose we could have that conversation if we threw in a long enough of a FutureWarning when passing hex values of RGBA, but I think at this point we should likely keep the hex ordering consistent with the library.

pyqtgraph/opengl/items/GLGradientLegendItem.py
```diff
@@ -18,7 +18,8 @@ def __init__(self, **kwds):
             gradient: a pg.ColorMap used to color the colorbar
             labels: a dict of "text":value to display next to the colorbar.
                 The value corresponds to a position in the gradient from 0 to 1.
-            fontColor: sets the color of the texts
+            fontColor: sets the color of the texts.
+                Arguments are the same as those accepted by functions.mkColor().
```
j9ac9k, line 22: To do a link within this docs, you can change to

```python
:func:`~pyqtgraph.mkColor`
```

pijyoi:
I agree with @NilsNemitz that the existing trivial functions / classes can be left behind without immediate deprecation.

I am not so sure about the #AARRGGBB vs #RRGGBBAA. #AARRGGBB feels very much like we are getting exposed to the internal Qt representation. i.e. alpha in MSB and blue in LSB of a 32-bit word.
```python
In [6]: hex(QtGui.QColor(0x33, 0x66, 0x99).rgba())
Out[6]: '0xff336699'
```

NilsNemitz:
@j9ac9k , I *was* advocating a deprecation period. :)

Maybe a compromise would be to have the documentation and docstrings state that 4-byte hex-strings are not ideal for unambiguously giving RGB + alpha, since the byte ordering is less well-defined as for RGB. That might make it easier to switch in the future, and doesn't really hurt anything in the present.

j9ac9k:
> @j9ac9k , I _was_ advocating a deprecation period. :)
> 
> Maybe a compromise would be to have the documentation and docstrings state that 4-byte hex-strings are not ideal for unambiguously giving RGB + alpha, since the byte ordering is less well-defined as for RGB. That might make it easier to switch in the future, and doesn't really hurt anything in the present.

I'd be good with that; At the very least we should explicitly state that we do not follow the Qt API.

@pijyoi The benefit of adopting `#AARRGGBB` would be not for the sake of the hex conversion, but the code in `mkColor` can be simplified in a way such that...

```python
if isinstance(arg, str):
    return QColor.setNamedColor(arg)
```

That would give us the benefit of not just being able to offload the hex logic (which isn't particularly complex); but also lets us access the [SVG color keyword names](https://www.w3.org/TR/SVG11/types.html#ColorKeywords); which I think would be useful.

Granted, we can easily accommodate both (get the color by SVG name, and handle our own hex string ordering).

pyqtgraph/opengl/GLViewWidget.py
```diff
@@ -138,9 +138,9 @@ def initializeGL(self):
     def setBackgroundColor(self, *args, **kwds):
         """
         Set the background color of the widget. Accepts the same arguments as
-        pg.mkColor() and pg.glColor().
+        pg.mkColor().
```
j9ac9k, line 141: I know this wasn't placed there before, but we can link the function here too:

```
:func:`~pyqtgraph.mkColor`
```

pijyoi:
>That would give us the benefit of not just being able to offload the hex logic (which isn't particularly complex); but also lets us access the SVG color keyword names; which I think would be useful.

I added it in as an experimental feature. Perhaps to become officially documented when the deprecated non-#-prefixed hex strings gets removed in 0.13?

```python
In [4]: pg.mkColor('tomato').getRgb()
Out[4]: (255, 99, 71, 255)
```

j9ac9k:
> I added it in as an experimental feature. Perhaps to become officially documented when the deprecated non-#-prefixed hex strings gets removed in 0.13?

That's a great idea.

j9ac9k:
@pijyoi this LGTM, I see no problem merging this right now.  Are you planning on adding anything else to this PR?

pijyoi:
> Are you planning on adding anything else to this PR?

No

j9ac9k:
Ha, oddly enough while working on that IEEE paper, I think I came across that SVG background bug, not sure why I didn't write up an issue on that given I opened numerous other issues that I encountered while generating plots for that... oh well.

This LGTM, thanks so much @pijyoi the code is so much simpler in so many places!

Issue #1378, opened 2020-09-21
New Parameter Items: FilePicker, ActionWithShortcut, PopupLineEditor


ntjess:
I ended up making these for one of my application and am proposing them here for integration into pyqtgraph, provided they are general enough. What do you think? Is it worth the effort of a pull request / should I post code in this issue / something else?

* The file dialogs are, of course, native to the OS since it is a `QFileDialog`.
* Folder and file pickers are the same parameter, where a boolean option `asFolder` determines whether files or folders are selected
* Similar to `list`, the options for `popuplineeditor` are set in the `limits` parameter. A `QListModel` with a `completer` are the core elements.

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/23620506/93737986-9de4f600-fbb2-11ea-8239-b365a12d25f6.gif)

ntjess:
FYI the source code for these parameter types is located in `pgregistered.py` within s3a ([here](https://gitlab.com/ficsresearch/s3a/-/blob/development/s3a/parameditors/pgregistered.py))

ntjess:
Also introducing `SliderParameter`:

```python
children = [
  dict(name='a', type='slider', value=5, limits=[0, 10]),
  dict(name='b', type='slider', value=0.1, limits=[-5, 5], step=0.1),
  dict(name='c', type='slider', value=2, span=np.linspace(0, 2*np.pi, 1000)),
]

p = Parameter.create(name='test', type='group', children=children)
t = ParameterTree()
t.setParameters(p, showTop=True)
```
![HUFpoYVAhN](https://user-images.githubusercontent.com/23620506/99867653-3125b200-2b89-11eb-8141-9a5f590545f0.gif)

ntjess:
@j9ac9k should I create a pull request with these? Or do you anticipate it will be rejected until usage is standardized / agreed upon?

ntjess:
@j9ac9k I see there was some recent attention on this ticket. Just for reference, the implementation for all these parameters is now [here](https://gitlab.com/ntjess/utilitys/-/blob/master/utilitys/params/pgregistered.py).

Of note, I have an implementation for `PgParamDelegate` as well that lets a table cell of a certain type have that `Parameter`'s delegate (i.e. an `int` table field would receive the `int` Parameter spinbox as a delegate). Do you think this is worth considering implementing?

[...]

PR #1952, opened 2021-08-05
Adds checklist parameter

ntjess:
The current suite of parameter types does not allow setting a list-like parameter with options. This is enabled with a `checklist` parameter.

Options:
  * `exclusive`: If true, only one option can be selected at a time. This makes the functionality similar to a `list` parameter type

Todo:
 - [x] Allow `checklist` to work with dict-like `limits`, similar to `list` behavior
 - [x] Apparently, `exclusive` decided to stop working. #FIXME


ntjess:
c.f. #1378

j9ac9k:
image

Just saw this when testing with dark mode on macOS.  I'm looking through the code-base and I'm not seeing a place that the style information is explicitly set though; so I don't think this is a blocker, but just sharing here in case you can think of someplace that might be setting the palette in a way that it's not inheriting from the system palette properly.

j9ac9k:
Had a little time, and I think I got to the bottom of this...

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/parametertree/parameterTypes/basetypes.py#L319-L338

Need to figure out how to inherit from the system palette and modify accordingly... granted this is a bit out of scope for this PR, so I'll create a separate issue.

ntjess:
FWIW, using a darkstyle palette seems to make the right choices on ubuntu, are you able to attempt this one as well? The attachment has to be `txt` or GH rejects it... [parametertree.txt](https://github.com/pyqtgraph/pyqtgraph/files/6949590/parametertree.txt)

image

pyqtgraph/parametertree/parameterTypes/checklist.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    def __init__(self, param, depth):
+        self.btnGrp = QtWidgets.QButtonGroup()
+        self.btnGrp.setExclusive(False)
+        self._constructMetaBtns()
+
+        super().__init__(param, depth)
+
+    def _constructMetaBtns(self):
+        self.metaBtnWidget = QtWidgets.QWidget()
```
ntjess, line 20: @j9ac9k Notice `_constructMetaBtns` doesn't assign a parent to the widget buttons, is this why the palette change isn't detected? I can't reproduce either way on my laptop, but I'm curious if the button text is correct for you if you change this to `QWidget(self)`.

I figured line 42 (`tw.setItemWidget(self, 1, self.metaBtnWidget)`) would properly assign the parent, but maybe I'm wrong

j9ac9k: changing

```python
        self.metaBtnWidget = QtWidgets.QWidget()
```

to

```python
        self.metaBtnWidget = QtWidgets.QWidget(self)
```

Results in a type-error.

j9ac9k:
With the example (not pyqt6 compatible), I do render like the screenshot you included:

image

That's not really how it should be either... I commented out the `self.setForeground` and  `self.setBackground` lines  in `GroupParameterItem.updateDepth` and got the color as it should be IMO.

image


This just indicates that the issue is outside of this PR.

j9ac9k:
image

While testing with exclusive=True, I noticed the checkboxes look like they're disabled/grayed out.

pyqtgraph/parametertree/parameterTypes/checklist.py
```diff
@@ -0,0 +1,169 @@
...
+        names = [self.reverse[1][self.reverse[0].index(val)] for val in cmpVals]
+        if exclusive and len(names) > 1:
+            names = [names[0]]
+        elif exclusive and not len(names) and len(self.forward):
+            # An option is required during exclusivity
+            names = [self.reverse[1][0]]
+        for chParam in self:
+            checked = chParam.name() in names
+            chParam.setValue(checked, self._onSubParamChange)
+            chParam.setOpts(enabled=not (exclusive and checked))
```
j9ac9k, line 168: Commenting out this line fixed the issue with the checkboxes looking disabled as they're being selected when `exclusive` is `True`.

ntjess: The disabling was by design, the goal being when `exclusive` is true exactly one option must be selected (like `list`). This keeps the currently checked box from becoming unchecked

ntjess: @j9ac9k Do you think it would be better to have custom logic that leaves the checkbox looking enabled, while also disallowing the user from having a `None` selection? I'm open to this, but I think leaving it disabled is a clearer indicator that something must be selected

j9ac9k: Sorry been slow to respond, I'm a bit under the weather as is one of my kids.  

_Generally_ speaking, when you can only select one entry, you don't have checkboxes, you get [radio buttons](https://doc.qt.io/qt-5/qradiobutton.html).  The exclusive with check-boxes feel a little out of place; but without knowing how much code would have to change to accommodate the radio buttons vs. the checkbox; I'm ok with it.

That said, having the checkboxes appear disabled, and yet be selectable definitely feels like an odd combination.

I guess I'll turn this around a bit, how difficult would it be to swap out checkboxes for QRadioButton's if `'exclusive': True` is passed?

ntjess: Probably not that hard -- I am already making a button group in the parameter item, so I'll play with it a bit

ntjess:
@j9ac9k I'm not great at fixing Qt signal connect/disconnect errors, so it's entirely possible the `dontPropagateChanges` logic I added is unnecessary. Basically, during `exclusive` options one parameter's `setValue` influences another, which causes problems trying to disconnect the same signal multiple times. I got around this by using the `dontPropagateChanges` fences during the potentially recursive operation.

Other than that, it seems to work pretty well and maintains radio/checkbox buttons depending on exclusivity.

j9ac9k:
I think you can call `.setValue(val, blockSignal=True)`

ntjess:
@pijyoi I tried the recommendation of blocking signals before `setValue`: 
```python
# self.dontPropagateChanges = True
blocked = self.blockSignals(True)
self.setValue(val)
self.blockSignals(blocked)
# self.dontPropagateChanges = False
```
However, this means that `sigValueChanged` doesn't fire when limits are updated. Notice how if you change `exclusive` from True to False or vice versa, there is no printout of the new value:
Using `dontPropagate`
image

Using `blockSignals`:
image

j9ac9k:
Failing CI checks are unrelated, so don't worry about those; I'll try and take a look at the diff here shortly.  Thanks for updating!

j9ac9k:
Thanks for the PR @ntjess sorry this took me so long to press the merge button!

PR #1953, opened 2021-08-05
Better parameter tree testing

ntjess:
This will fail until #1952 is merged. Normally I would wait to submit, but if folks have time to review this while 1952 is evaluated I can make changes to both at the same time.

The current parameter tree example suffers from two main issues. Since more parameter types have been added, 1) Organization is somewhat confusing, and 2) not all options are being tested in the example suite. This PR does the following:

* Organizes parameter types -- all available parameter types are together and collapsed to make the tree more digestible.
* All options for all types are exposed for playing -- The user can edit any option and it will propagate to the widget under test.
* An assertion checks that all types are instantiated -- if more parameter types are added without an example for them, the assertion will fail.

Some considerations: Since the `examples` text box doesn't show the code for `makeAllParamTypes`, it is unclear how these base types are instantiated. However,

* The rest of the parameter tree clearly shows how to make parameters with options
* Each option in `makeAllParamTypes` directly corresponds to an option with the same name for that parameter, so the intuition for applying the example should still be applicable.


j9ac9k:
> Some considerations: Since the `examples` text box doesn't show the code for `makeAllParamTypes`, it is unclear how these base types are instantiated.

I think we can get away with putting a comment in the example file.

This diff LGTM, only thing I would suggest is addressing the issues the static code checker turned up (I know some of the issues are not new; if you recognize a good fix for those, feel free to implement, otherwise feel free to skip).

While not related to your PR I don't think, the `cfg` file, while useful, annoys me a bit that we're implementing our own file format, and not using some of the standards that python embeds a parser for such as [`ini` files](https://docs.python.org/3/library/configparser.html)

ntjess:
@j9ac9k I would agree with you regarding the config parsing. The only major drawbacks to ini are the lack of nested parsing and arbitrary evals (with global/local scopes), which the "home-baked" solution gives. Yaml or some more sophisticated format could handle both, but of course is not in the standard library.

I considered these things when writing the PR, but figured it would take a long time to overload ConfigParser for evals and nested statement parsing so I just... didn't :grimacing:

j9ac9k:
> @j9ac9k I would agree with you regarding the config parsing. The only major drawbacks to ini are the lack of nested parsing and arbitrary evals (with global/local scopes), which the "home-baked" solution gives. Yaml or some more sophisticated format could handle both, but of course is not in the standard library.
> 
> I considered these things when writing the PR, but figured it would take a long time to overload ConfigParser for evals and nested statement parsing so I just... didn't 😬

I mean, if we want nesting, would json be the end of the world?

ntjess:
> I mean, if we want nesting, would json be the end of the world?

json makes it much harder to understand what's going on (I experienced this with my larger config files). I also have no idea how to parse values with `eval` in json (which is arguably bad practice anyway, but for this application it works well).

Alternatively, I could make the config a `py` file that defines a python dict

j9ac9k:
> Alternatively, I could make the config a `py` file that defines a python dict

Considering I can't think of a reason to groan or complain about this option off the top of my head; this would probably be my preferred option. Other folks (@outofculture ?) should probably weigh in tho.

examples/_buildParamTypes.py
```diff
@@ -69,8 +64,8 @@ def makeAllParamTypes():
     # Also minor tweak to meta opts
     meta = params.child('Applies to All Types')
     infoChild = meta.child('Extra Information')
-    for ch in meta.children()[1:]:
-        ch.sigValueChanged.connect(lambda _param, _val: infoChild.setOpts(**{_param.name(): _val}))
+    for child in meta.children()[1:]:
+        child.sigValueChanged.connect(lambda _param, _val: infoChild.setOpts(**{_param.name(): _val}))
```
j9ac9k, line 68: @outofculture I know we've had trouble with signals being connected the lambda functions, I'm having a nagging memory that there was a 3rd factor that had to meet, like the lambda function had to reference `self`?  Or am I remembering wrong?

outofculture: Correct. Lambdas are fine so long as they don't reference `self`.

j9ac9k:
Hi @ntjess should have merged this some time ago, my apologies.  Merging now.

PR #1954, opened 2021-08-05
Deprecate `values` opt for list parameter

ntjess:
List parameter accepts a `values` argument during initialization, but doesn't listen for this anywhere else (i.e. during an opt change, limit update, etc.). The clearest path forward I believe is to deprecate `values` and encourage `limits` moving forward, rather than to allow option aliasing.


pijyoi:
There are some more examples using "values": ```PlotSpeedTest.py```, ```ScatterPlotSpeedTest.py```, ```relativity.py```.

I used List parameter several years ago and I would be confident to say that I used "values" because the way I learnt how to use it was through ```examples/parametertree.py```.
So the question is whether this deprecation is necessary? In the sense that, does it help the library by deprecating it?

ntjess:
I thought I ran a search across the library, but it looks like I was accidentally scoped to the `pyqtgraph` folder -- will make the adjustments tomorrow.

I believe the deprecation is necessary because

  * To say `values` is synonymous with `limits` to me implies that you can `setOpts` with values similarly. However, there is no listener for this in the `list` code -- it only looks for the `values` key on init. Moreover, limits is an established paradigm through function/signal names also in other parameters, while `list` is the only one that uses `values` (again, only during initialization). Thus, either every parameter supporting limits should also support values, or the support for `values` should be reduced. I'm in favor of avoiding the alias creep.
  * `values` is visually similar to `value`, so personally I see a readability improvement by specifying `limits`.
  * No other parameter option supports aliasing (enabled, expanded, etc.), so it seems like a strange break in pattern to have it here and nowhere else. Thus, deprecating the alias helps future improvements to the library confirm to a consistent naming scheme.

If these sentiments aren't shared by others, I can submit a different PR to at least fix the `setOpts` bug

pijyoi:
May I suggest increasing the deprecation runway?
```python
  if 'values' in opts:
      warnings.warn('Using "values" to set limits is deprecated. Use "limits" instead.',
          DeprecationWarning, stacklevel=2)
      opts['limits'] = opts['values']
```
If there are only a couple of lines dealing with handling ```"values"```, there may not even be a need to set a removal deadline. See @NilsNemitz 's comment at https://github.com/pyqtgraph/pyqtgraph/pull/1944#issuecomment-891534490. i.e. stop advertising undesired features but no need to remove them outright.

ntjess:
That makes a lot of sense to me. I've removed the timeline specification, thanks.

pijyoi:
Would it be a bug in the (original) examples/parametertree.py that ```'value'``` is set to a value not within ```'limits'```?
```python
 {'name': 'Named List', 'type': 'list', 'limits': {"one": 1, "two": "twosies", "three": [3,3,3]}, 'value': 2},
```

ntjess:
Less of a 'bug' and more 'undocumented'. A code path checks for values that aren't in limits. When this is the case, the first limit is selected. Note that #1953 changes the logic to use `pg.eq` in the event values are numpy-like.

https://github.com/pyqtgraph/pyqtgraph/blob/f16125c3785cde231a73e923c46412be289f67a5/pyqtgraph/parametertree/parameterTypes/list.py#L36-L37

j9ac9k:
Thanks for the PR @ntjess and thanks for the feedback to it @pijyoi Sorry this took me so long to merge, doing so now.

PR #1956, opened 2021-08-07
arrayToQPath: use QByteArray as backing store

pijyoi:
this avoids:
1) allocation and zero-ing of bytearray
2) copy of bytearray to QByteArray

Using the benchmarking script in #1817,

test | time (s) / master | time (s) / this PR
-- | -- | --
pairs | 0.394 | 0.360
array | 0.394 | 0.365

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1957,12 +1957,13 @@ def arrayToQPath(x, y, connect='all', finiteCheck=True):
         backstore = create_qpolygonf(n)
         arr = np.frombuffer(ndarray_from_qpolygonf(backstore), dtype=[('x', 'f8'), ('y', 'f8')])
     else:
-        backstore = bytearray(4 + n*20 + 8)
+        backstore = QtCore.QByteArray()
+        backstore.resize(4 + n*20 + 8)      # contents uninitialized
+        backstore.replace(0, 4, struct.pack('>i', n))
+        # cStart, fillRule (Qt.FillRule.OddEvenFill)
+        backstore.replace(4+n*20, 8, struct.pack('>ii', 0, 0))
         arr = np.frombuffer(backstore, dtype=[('c', '>i4'), ('x', '>f8'), ('y', '>f8')],
             count=n, offset=4)
-        struct.pack_into('>i', backstore, 0, n)
-        # cStart, fillRule (Qt.FillRule.OddEvenFill)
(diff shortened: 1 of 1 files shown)


j9ac9k:
9% improvement for "free"; nice!  ...also getting away from if statements based on Qt binding versions; this diff is fantastic.

Regarding the two points, does this mean we likely have some memory consumption benefits as well (at least temporarily until the function returns?).

pijyoi:
>does this mean we likely have some memory consumption benefits as well

Yes, there is one less memory allocation.

In the pre-#1796 code: (3x allocs + 2x memcpy for PySide; 2x allocs + 1x memcpy for PyQt)
1) one ndarray.empty() (alloc-only)
2) one conversion to bytes (alloc + memcpy) (for PySide codepath)
3) one QByteArray (alloc + memcpy)

In master: (2x allocs + 1x zero + 1x memcpy)
1) one bytearray (alloc + zero)
2) one QByteArray (alloc + memcpy)

This PR: (1x alloc)
1) one QByteArray (alloc-only)

j9ac9k:
> > does this mean we likely have some memory consumption benefits as well
> 
> Yes, there is one less memory allocation.
> 
> In the pre-#1796 code: (3x allocs + 2x memcpy for PySide; 2x allocs + 1x memcpy for PyQt)
> 
> 1. one ndarray.empty() (alloc-only)
> 2. one conversion to bytes (alloc + memcpy) (for PySide codepath)
> 3. one QByteArray (alloc + memcpy)
> 
> In master: (2x allocs + 1x zero + 1x memcpy)
> 
> 1. one bytearray (alloc + zero)
> 2. one QByteArray (alloc + memcpy)
> 
> This PR: (1x alloc)
> 
> 1. one QByteArray (alloc-only)

Oh wow, as part of an IEEE paper we submitted some time ago (which did not get accepted) we did some [benchmarking of plot refresh rates vs. number of points in a line](https://github.com/pyqtgraph/IEEEVIS2021-pyqtgraph/blob/main/figures/lineBenchmark.pdf); ....this was pre #1796; we will be trying to resubmit that paper elsewhere (let me know if you're interested in joining that effort); but I'm going to be curious as to the impact of not only the speed boost but if memory consumption will play a role as well (basically, will _really_ large line-plots have significant gains...  If/when I rerun those benchmarks, I'll be sure to update you

j9ac9k:
Thanks for the PR @pijyoi this LGTM, Merging!

PR #1959, opened 2021-08-08
Extend docs template to add a banner for the user survey

ixjlyons:
Submitting this before the survey goes out to collaborate on the details.

Currently it's just a paragraph on top of the body of ~every page. I think we could do a dismissible message/banner type thing, but that might be more annoying? Starting simple in any case. Suggestions welcome on anything about it really.

https://pyqtgraph--1959.org.readthedocs.build/en/1959/

@NilsNemitz @j9ac9k

### DIFF
--- a/doc/source/_templates/page.html
+++ b/doc/source/_templates/page.html
@@ -1,6 +1,6 @@
 {% extends "layout.html" %}
 {% block body %}
-    <p style="border:2px solid Tomato;font-weight:bold">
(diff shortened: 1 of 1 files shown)


j9ac9k:
I think that's perfect, I'm good with making this annoying

Achilles1515:
I think adding some padding to keep the paragraph from being suffocated by the border would look nice.

Current:
image

----------------------------------

vs `padding: 1em;`, for example:
image

Issue #1967, opened 2021-08-14
The scale argument from the ImageView setImage does not work as expected with PyQt6


Anatoly1010:
### Short description
The scale argument from the ImageView setImage method does not work as expected when using PyQt6.

### Code to reproduce

```python
import numpy as np
# force to use pyqt6 
import PyQt6
from pyqtgraph.Qt import QtGui, QtWidgets
import pyqtgraph as pg

def main():
    app = QtWidgets.QApplication([])
    
    # dummy data
    kwargs = {}
    data = np.random.normal( size = (2, 600, 600), loc = 1024, scale = 64 )

    win = QtGui.QWidget()
    win.setLayout( QtGui.QHBoxLayout() )
    win.resize(800,800)

    plot_item = view = pg.PlotItem()
    img_view = pg.ImageView( view = view )
    item = plot_item.getViewBox()
    item.invertY(False)
    
    # pos argument works correctly
    kwargs['pos'] = (500, 500)
    # scale gives rise to an error
    kwargs['scale'] = (10, 1)
    img_view.setImage( data, **kwargs )
    
    win.layout().addWidget(img_view)
    win.show()
    win.raise_()
    app.exec()
    
if __name__ == "__main__":
    main()
```

### Expected behavior

The scale argument should change a scale of the image.

### Real behavior

When using PyQt6, the scale argument returns an error. In the case of PyQt5, everything works as expected.
```
Traceback (most recent call last):
  File "image_view_error.py", line 33, in <module>
    main()
  File "image_view_error.py", line 25, in main
    img_view.setImage( data, **kwargs )
  File ".../python3.8/site-packages/pyqtgraph/imageview/ImageView.py", line 363, in setImage
    self.imageItem.scale(*scale)
TypeError: scale(self): too many arguments
```

### Tested environment(s)

 * PyQtGraph version: 0.12.2
 * Qt Python binding: PyQt6 Qt 6.1.2 // PyQt5 Qt 5.14.1 
 * Python version: 3.8.10
 * NumPy version: 1.18.3
 * Operating system: Linux Ubuntu 20.04
 * Installation method: pip

pijyoi:
Hi, thank you for reporting. ImageView is using a deprecated monkey-patched method QGraphicsItem.scale(). This monkey-patching was deliberately left out for Qt6 and usage of this method within the library was removed. You have found a leftover.

It would be appreciated if you could test out #1968 to see whether it fixes this issue.

[...]

PR #1968, opened 2021-08-14
fix: ImageView calling deprecated QGraphicsItem.scale()

pijyoi:
This PR is meant to address #1967. A leftover usage of QGraphicsItem.scale() within the library.

In addition, it fixes another problem. The documentation for ImageView says that the ```transform``` option overrides ```pos``` and ```scale```. 
```
        pos                Change the position of the displayed image
        scale              Change the scale of the displayed image
        transform          Set the transform of the displayed image. This option overrides *pos*  and *scale*.
```
However, because ```setPos()``` was used, which acts independently of ```setTransfrom()```, this was not actually true.
(Related to the discussions in #1726)

Now, the question is whether fixing the code to match the docs is in fact the right thing to do.

### DIFF
--- a/pyqtgraph/imageview/ImageView.py
+++ b/pyqtgraph/imageview/ImageView.py
@@ -358,13 +358,15 @@ class ImageView(QtGui.QWidget):
         
         profiler()
 
-        self.imageItem.resetTransform()
-        if scale is not None:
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks @pijyoi for the quick patch!

PR #1969, opened 2021-08-17
bugfix: result of clip_array() not assigned

pijyoi:
change it to operate in-place.


j9ac9k:
Nice catch! Thanks for the PR @pijyoi

Issue #1970, opened 2021-08-17
PlotWidget (and other plotting widgets) unhides all Items after removing single item; Again


sem-geologist:
The issue is exact as previously reported by me in issue #985
which was fixed with #987.

I had bisected regression down to this PR: #1764;

The previously introduced fix was kind of reverted at commit fa1be1e5bf50c9da27cc9473b04f2399319dc324 ,
at method `updateDecimation`, where inserted unnecessary `curve.show()`, which ignores the fact that some curves intentionally could be hidden, makes everything show if anything is removed or added.

sem-geologist:
I looked over to this:
```python
...
    def updateDecimation(self):
        if not self.ctrl.maxTracesCheck.isChecked():
            # numCurves = len(self.curves)
            return
        else:
            numCurves = self.ctrl.maxTracesSpin.value()

        for i, curve in enumerate(self.curves):
            if i < numCurves:
                curve.show()
            elif self.ctrl.forgetTracesCheck.isChecked():
                curve.clear()
                self.removeItem(curve)
            else:
                curve.hide()
```
first of all
`updateDecimation` method is not documented well, it took me a while to understand why it is called in different places of the code and what it is intended to do.
I think that 3rd line `numCurves = len(self.curves)` should be simply replaced with `return as in example`.
That way there would be no iterations over curves and no calling `show()` method even when maxTraces is not checked, which is now being done.
@outofculture , what do you think? Would maxTraces would still work (I have no working example or need/workflow for need of that feature, where automatic unhidding of curves breaks my work, and can't find easy workaround)

sem-geologist:
I looked in tooltip for `forgetTraceCheck` and it sounds like that:

> If MaxTraces is checked, remove curves from memory after they are hidden (saves memory, but traces can not be un-hidden).

so I think it is enough to do check if maxTracesCheck is checked and return from function early as possible if it is not.

sem-geologist:
and maybe ui file should be as well updated with `forgetTracesCheck` and `maxTracesSpin` being by default disabled, and being enabled only after ticking Max Traces...

PR #1971, opened 2021-08-17
fix unhiding of curves when Max Traces is not checked.

sem-geologist:
return early from updateDecimation if Max Traces checkbox is not checked.


sem-geologist:
this fixes issue #1970

sem-geologist:
I probably should upload the PR from separate branch...

PR #1973, opened 2021-08-17
Fix: updateDecimate should not unhide intentionaly hidden curves

### LINKED ISSUE #1970: PlotWidget (and other plotting widgets) unhides all Items after removing single item; Again

sem-geologist:
make update decimate to not unhide curves when items added/removed,
while preserving the Max Traces well behavior.

this fixes #1970 

- [x] implementation of bug-fix
- [x] add test if hidden curves stays hidden while adding/removing other item.
- [x] doc for MaxTrace and mention external possibilities to control curve visibility. (or add doc string to `updateDecimate`)

### DIFF
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -246,7 +246,7 @@ class PlotItem(GraphicsWidget):
         
         if labels is None:
             labels = {}
-        for label in list(self.axes.keys()):b
+        for label in list(self.axes.keys()):
             if label in kargs:
                 labels[label] = kargs[label]
                 del kargs[label]
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -1006,7 +1006,7 @@ class PlotItem(GraphicsWidget):
     
     def _max_trace_wrapper_slot(self, check_state):
         if check_state:
-            self.updateDecimation(self)
+            self.updateDecimation()
         else:
             for curve in self.curves:
                 curve.show()
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -1003,7 +1003,6 @@ class PlotItem(GraphicsWidget):
     def clipToViewMode(self):
         return self.ctrl.clipToViewCheck.isChecked()
     
-    
     def _handle_max_traces_toggle(self, check_state):
         if check_state:
             self.updateDecimation()
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -240,13 +240,13 @@ class PlotItem(GraphicsWidget):
         self.ctrl.avgParamList.itemClicked.connect(self.avgParamListClicked)
         self.ctrl.averageGroup.toggled.connect(self.avgToggled)
         
-        self.ctrl.maxTracesCheck.toggled.connect(self.updateDecimation)
+        self.ctrl.maxTracesCheck.toggled.connect(self._max_trace_wrapper_slot)
         self.ctrl.forgetTracesCheck.toggled.connect(self.updateDecimation)
         self.ctrl.maxTracesSpin.valueChanged.connect(self.updateDecimation)
         
         if labels is None:
             labels = {}
-        for label in list(self.axes.keys()):
+        for label in list(self.axes.keys()):b
             if label in kargs:
                 labels[label] = kargs[label]
                 del kargs[label]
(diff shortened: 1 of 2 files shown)


pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -240,13 +240,13 @@ def __init__(self, parent=None, name=None, labels=None, title=None, viewBox=None
...
         
-        self.ctrl.maxTracesCheck.toggled.connect(self.updateDecimation)
+        self.ctrl.maxTracesCheck.toggled.connect(self._max_trace_wrapper_slot)
         self.ctrl.forgetTracesCheck.toggled.connect(self.updateDecimation)
         self.ctrl.maxTracesSpin.valueChanged.connect(self.updateDecimation)
         
         if labels is None:
             labels = {}
-        for label in list(self.axes.keys()):
+        for label in list(self.axes.keys()):b
```
j9ac9k, line 249: looks like a typo resulting in a syntax error?

sem-geologist: yes :/ navigating with thinkpad trackpoint...

sem-geologist:
The idea of this PR is to make Max Track work as intended, while preventing the intently hidden curves from showing up.
These are a bit orthogonal behaviors, as  implementation of Max Tracks does not care if hide, show methods of curves (PlotItem) are used outside of its scope. updateDecimate is being called in different places in the code.
We want updateDecimate to show all curves only if its state was previously toggled=True; that is why that wrapper function is added, and connection from the checkbox of Max Track redirected to that wrapper function. I am not sure about naming of that function, and if it would be "private" (with underscore).

outofculture:
Can we add a test? `tests.graphicsItems.PlotItem.test_PlotItem.test_PlotItem_maxTraces` already has some assertions about the Max Traces checkbox, but you should add something that says "when max traces is unchecked, removing curves will not change the visibility of any of the existing curves". You already have example code that breaks, so it should be quick to put an `assert` in front of the correct spot, make sure it fails with the old code, and secure this behavior from future regressions.

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -1002,10 +1002,18 @@ def setClipToView(self, clip):
         
     def clipToViewMode(self):
         return self.ctrl.clipToViewCheck.isChecked()
-
+    
+    
+    def _max_trace_wrapper_slot(self, check_state):
```
outofculture, line 1007: I would rename this to be more verb-y, e.g. `_handle_max_traces_toggle`, but that's a style choice for which we don't have a guideline.

sem-geologist:
> You already have example code that breaks,

What do you mean? after all those my typo fixed, the last test run had passed. So it does not brake current MaxTraces behavior (it is showing all curves when Max Traces checkbox is unchecked as is intended by your fix).

My usage case is as below in screenshot:
HussariX_screenshot
Where curves are created for all positions on the left with position checked in the bottom list, and visibility is toggled and changed with tree state changes on the left (selection highlights, ticking/unticking checkbox shows/hides curve). So basically the visibility of curves are handled in my case externally from the pyqtgraph PlotItem, directly onto curves. In my case that makes sense, as creating destroying curves is expensive, and workflow with spectral data needs checking different curves against other curves. Alternative would be to add/removeItems with left tree check state changes, but adding/removing items does autorange and I need my plot coordinates to stay the same between item showing/hiding.

PyQtGraph documentation is silent about handling the curve visibility externally, and probably my usage case was the only one which came across that "bug". I could add test to see that if MaxTrace state is unchecked, adding hidden items would not show hidden items. But I think then documentation should also be extended to include possibility of external control of curve visibility, else this additional test would test not documented behavior of pyqtgraph. I am not sure where to document such usage case (Under documentation of `PlotItem`?)

As for renaming additional function I think Your proposed name is more clear and I should change that.

sem-geologist:
@j9ac9k , I heard that there is going to be patch, I would be interested this fix to be included in next minor version of pyqtgraph. What needs to be done more for this PR for it to be merged?

j9ac9k:
hi @sem-geologist 

As a general point, before doing any releases, I try and merge the ready to merge PRs; which this PR likely meets that criteria.  I'll let @outofculture comment on merge status (or merge this PR himself).  This makes doing patch releases take a bit longer, but I feel that when a contributor makes a PR and does the leg-work of getting it ready to me merged, their contribution should be present in the following release.

I've been taking a bit of a break from maintainer duties so I apologize for the longer queue on incoming PRs.  FWIW (nearly) zeroing out the outstanding PR queue has been a goal of the maintainers here for some time (we used to have over 160 outstanding PRs!), so please don't feel like the PR queue is a place where contributions go to die.

outofculture (approved):
Looks ready to merge.

> I think then documentation should also be extended to include possibility of external control of curve visibility

The docstring you added is great for this. Nice touch.

j9ac9k:
Thanks for the PR @sem-geologist sorry this was in the queue for so long!  Thanks for the review/feedback @outofculture

PR #1974, opened 2021-08-19
Fix ImageView levelMode

### LINKED ISSUE #1769: ImageView(levelMode='rgba') throws exception
### Short description

Instantiating ImageView widget as ImageView(levelMode='rgba') causes an exception in HistogramLUTItem.
Providing an ImageItem upon instantation doesn't help.

Changing the levelMode to rgba after instantiation does work as shown in the "workaround" codepath.

### Code to reproduce

```python
import pyqtgraph as pg
import numpy as np

data = np.random.randint(256, size=(256,256,3))

app = pg.mkQApp()

imgitem = pg.ImageItem(data)
if 1:
    # fails
    imv = pg.ImageView(imageItem=imgitem, levelMode='rgba')
else:
    # workaround
    imv = pg.ImageView(imageItem=imgitem)
    imv.getHistogramWidget().item.setLevelMode('rgba')
imv.show()

app.exec_()
```

### Expected behavior

Program starts with no exception thrown

### Real behavior

```
  File "d:\github\pyqtgraph\pyqtgraph\imageview\ImageView.py", line 130, in __init__
    self.ui.histogram.setLevelMode(levelMode)
  File "d:\github\pyqtgraph\pyqtgraph\graphicsItems\HistogramLUTItem.py", line 309, in setLevelMode
    self._showRegions()
  File "d:\github\pyqtgraph\pyqtgraph\graphicsItems\HistogramLUTItem.py", line 337, in _showRegions
    xdif = 1.0 / nch
UnboundLocalError: local variable 'nch' referenced before assignment
```

### Tested environment(s)

 * PyQtGraph version: master 
 * Qt Python binding: PySide6 6.0.4 Qt 6.0.4
 * Python version: 3.8.9
 * NumPy version: 1.20.2 

magnium:
fixes #1769

### DIFF
--- a/pyqtgraph/graphicsItems/HistogramLUTItem.py
+++ b/pyqtgraph/graphicsItems/HistogramLUTItem.py
@@ -443,7 +443,7 @@ class HistogramLUTItem(GraphicsWidget):
             self.regions[i].setVisible(False)
 
         if self.levelMode == 'rgba':
-            imax = 4
+            nch = 4
             if self.imageItem() is not None:
                 # Only show rgb channels if connected image lacks alpha.
                 nch = self.imageItem().channels()
--- a/pyqtgraph/imageview/ImageView.py
+++ b/pyqtgraph/imageview/ImageView.py
@@ -127,7 +127,6 @@ class ImageView(QtGui.QWidget):
         self.ui = ui_template.Ui_Form()
         self.ui.setupUi(self)
         self.scene = self.ui.graphicsView.scene()
-        self.ui.histogram.setLevelMode(levelMode)
         
         self.ignorePlaying = False
         
@@ -147,6 +146,7 @@ class ImageView(QtGui.QWidget):
         self.currentIndex = 0
         
         self.ui.histogram.setImageItem(self.imageItem)
+        self.ui.histogram.setLevelMode(levelMode)
         
         self.menu = None
         
--- a/pyqtgraph/imageview/ImageView.py
+++ b/pyqtgraph/imageview/ImageView.py
@@ -138,16 +138,6 @@ class ImageView(QtGui.QWidget):
         self.view.setAspectLocked(True)
         self.view.invertY()
         
-        if imageItem is None:
-            self.imageItem = ImageItem()
-        else:
-            self.imageItem = imageItem
-        self.view.addItem(self.imageItem)
-        self.currentIndex = 0
-        
-        self.ui.histogram.setImageItem(self.imageItem)
-        self.ui.histogram.setLevelMode(levelMode)
-        
         self.menu = None
         
         self.ui.normGroup.hide()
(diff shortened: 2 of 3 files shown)


pijyoi:
Could you rebase your branch onto pyqtgraph master?

magnium:
Done. Sorry, I forked first when master didn't exist.

pyqtgraph/graphicsItems/HistogramLUTItem.py
```diff
@@ -443,7 +443,7 @@ def _showRegions(self):
             self.regions[i].setVisible(False)
 
         if self.levelMode == 'rgba':
-            imax = 4
+            nch = 4
```
outofculture, line 446: Nice catch.

outofculture:
This changeset does fix the exception reported in #1769, but as you point out, the ROI graph doesn't get populated appropriately. Calling `ImageView.setImage(data, levelMode='rgba')` _does_ fill in the ROI, but that method only accepts the raw data, not an `ImageItem` as @pijyoi attempted. I think this PR could be merged, and the original issue closed, but then we'll need to open an additional ticket to address the broken ROI graph. 

Notably, adding:
```python
if imageItem is not None:
    self.setImage(imageItem.image)
```
to the end of the `ImageView.__init__` method is enough to get the ROI graph back. I don't know what the usecase is for passing in a pre-initialized ImageItem, though, so I don't know if that's an acceptable workaround. @pijyoi any thoughts there?

This also could get a test without too much trouble. `tests/imageview/test_imageview.py` could have this added:
```python

def test_init_with_mode_and_imageitem():
    data = np.random.randint(256, size=(256, 256, 3))
    imgitem = pg.ImageItem(data)
    pg.ImageView(imageItem=imgitem, levelMode="rgba")
```
Though that test doesn't touch on the missing ROI graph.

pijyoi:
> I don't know what the usecase is for passing in a pre-initialized ImageItem,

One reason to pass in an ImageItem is to be able to preset ```autoDownsample```.
Alternatively, one would have to call ```imv.getImageItem().setAutoDownsample(True)``` as done in
https://github.com/pyqtgraph/pyqtgraph/issues/1512#issuecomment-768970303

The issue being that some settings can only be set via an ```ImageItem``` instance.

magnium:
> is enough to get the ROI graph back

Not quite. ImageView.setImage calls self.roiClicked() which needs self.roi to be initialized. Also, if autoRange is True the image gets zoomed. Additionally, since we are using imageItem, we don't want its attributes modified implicitly (ImageItem.transform). Hence the new commit, with the test proposed by @outofculture.

outofculture (approved):

j9ac9k:
Thanks for the PR @magnium sorry this took us a bit to merge.  Thanks for the review @outofculture and @pijyoi

Issue #1984, opened 2021-09-05
ImageItem float luts "regression"


pijyoi:
ImageItem weakly documents that the lut that it accepts is of type ubyte:
https://pyqtgraph.readthedocs.io/en/latest/graphicsItems/imageitem.html#pyqtgraph.ImageItem.setLookupTable
https://pyqtgraph.readthedocs.io/en/latest/functions.html#pyqtgraph.makeARGB

However, in pyqtgraph <= 0.12.1, float luts scaled to 255 are silently accepted and work.
Since pyqtgraph >= 0.12.2, such float luts no longer work in certain cases. (lut more than 256 entries, image data is floating)
luts that are ubyte work as always. 

Sample demonstrating code. (Matplotlib returns float luts)
```python
import pyqtgraph as pg
import numpy as np
import matplotlib.pyplot as plt

app = pg.mkQApp()
glw = pg.GraphicsLayoutWidget()
glw.show()

# affected types: float32, float64, uint16
ramp = np.arange(1000, dtype=np.float32)
ramp = ramp.reshape((1, -1))
cmap = plt.get_cmap("viridis")

for idx, num_colors in enumerate([256, 257]):
    lut = cmap(np.linspace(0, 1, num_colors)) * 255
    # lut = lut.astype(np.uint8)

    imgitem = pg.ImageItem(ramp, axisOrder='row-major')
    imgitem.setLookupTable(lut)

    vb = glw.addViewBox(row=idx, col=0)
    vb.addItem(imgitem)

app.exec_()
```
Output for pyqtgraph >= 0.12.2
For pyqtgraph <= 0.12.1, both bars are smooth gradients.
![Capture](https://user-images.githubusercontent.com/2657027/132125347-58bd01dc-c9ca-4763-9bf5-473f17dad438.PNG)

Possible "fixes":
1) Do nothing
2) Fail noisily if lut is not uint8
3) Warn if lut is not uint8
4) Cast the float ndarray to uint8 within the library

Note that even though float luts work now if the number of colors is <= 256, the library is forced to go through the slower ```makeARGB()``` codepath.

j9ac9k:
@outofculture  what do you think?

pijyoi:
This issue was motivated when I chanced upon https://github.com/acconeer/acconeer-python-exploration/issues/88
After experimenting some more, I think I have managed to reproduce the error encountered in the referenced issue.
i.e. an exception is thrown rather than the wrong colorbar being displayed.

[...]

PR #1985, opened 2021-09-06
delegate float luts to makeARGB()

pijyoi:
This is a workaround for #1984.

While it could be considered a usage error for luts to not be uint8, such a usage did after all work in pyqtgraph <= 0.12.1. Rather than correct this usage error, we can simply delegate processing to the codepath used in pyqtgraph <= 0.12.1, i.e. ```makeARGB()```.

### DIFF
--- a/pyqtgraph/graphicsItems/ImageItem.py
+++ b/pyqtgraph/graphicsItems/ImageItem.py
@@ -535,7 +535,15 @@ class ImageItem(GraphicsObject):
         levels = self.levels
         augmented_alpha = False
 
-        if image.dtype.kind == 'f':
+        if lut is not None and lut.dtype != self._xp.uint8:
+            # Both _try_rescale_float() and _try_combine_lut() assume that
+            # lut is of type uint8. It is considered a usage error if that
+            # is not the case.
+            # However, the makeARGB() codepath has previously allowed such
+            # a usage to work. Rather than fail outright, we delegate this
+            # case to makeARGB().
+            pass
+        elif image.dtype.kind == 'f':
             image, levels, lut, augmented_alpha = self._try_rescale_float(image, levels, lut)
             # if we succeeded, we will have an uint8 image with levels None.
             # lut if not None will have <= 256 entries
--- a/pyqtgraph/graphicsItems/ImageItem.py
+++ b/pyqtgraph/graphicsItems/ImageItem.py
@@ -1,6 +1,4 @@
 # -*- coding: utf-8 -*-
-from __future__ import division
-
 import numpy
 
 from .GraphicsObject import GraphicsObject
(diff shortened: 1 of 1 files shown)


outofculture:
None of the tests, for either ImageItem or makeARGB, use a float LUT. We should add one of those, at least. The code from #1984 should work for that.

pijyoi:
The library code has always assumed that the lut values range from 0-255. The required dtype of uint8 was documented in makeARGB(). If it used to work for float dtype luts, it was by accident. 
(Note that conventionally, float colors would range from 0.0-1.0, which in this case it does not)

The current workaround here was to restore any old behavior that was being relied upon. Correct library usage would still be uint8 luts. The examples given in #1984 are in fact examples of wrong usage.
If it would be preferable, this PR could be changed to reject non-uint8 luts.

Do you want the tests to check that the library accepts or rejects non-uint8 luts?

outofculture:
I don't have a strong opinion about this behavior - my app doesn't use it - but I wish I could ask all the users whether the matplotlib-supplied LUT is a common use-case. I would opine then that it's cheap to keep, but we should add a deprecation warning that says "Using float LUTs is an undocumented accidental feature and may be removed at some point in the future. Please open an issue if you instead believe this to be worthy of protected inclusion in pyqtgraph." or some such.

pijyoi:
>but I wish I could ask all the users whether the matplotlib-supplied LUT is a common use-case.

Prior to pyqtgraph 0.12.0, there was no easy way to get a colormap. So yes, using a matplotlib colormap was probably the way to go.

Old way 1: get it from GradientEditorItem
- using pyqtgraph internals (probably not recommended!)
- limited choice of colormaps
```python
cmap = pg.ColorMap(*zip(*Gradients["viridis"]["ticks"]))
lut = cmap.getLookupTable(0, 1, num_colors)
```

Old way 2: get it from an external source, e.g. matplotlib
- user needs to ensure it is in the right format
```python
import matplotlib.pyplot as mpl_plt
cmap = mpl_plt.get_cmap("viridis")
lut = (cmap(np.linspace(0, 1, num_colors)) * 255).astype(np.uint8)
```

With a new pyqtgraph 0.12.x, we can now do:
```python
import pyqtgraph as pg
cmap = pg.colormap.get("viridis")
lut = cmap.getLookupTable(0, 1, num_colors)
```
or
```python
import pyqtgraph as pg
cmap = pg.colormap.get("viridis", source="matplotlib")
lut = cmap.getLookupTable(0, 1, num_colors)
```

j9ac9k:
Thanks @pijyoi for this and for tracking down the issue in #1984 ...btw what do you think we should do w/ that issue now?

pijyoi:
> but I wish I could ask all the users whether the matplotlib-supplied LUT is a common use-case.

In https://stackoverflow.com/questions/69367352/matplotlib-pyplot-clim-counterpart-in-pyqtgraph, the following is done:
```python
colormap = cm.get_cmap("hot")
colormap._init()
lut = (colormap._lut * 255).view(np.ndarray)
...
img.setLookupTable(lut)
```

After investigating, besides passing in floats to pyqtgraph, there's another level of wrong in this usage.
Matplotlib's _private_ ```_lut``` contains 3 extra entries at the end to indicate the colors to use for values that are:
1) less than low level
2) more than high level
3) masked

Thus, passing in the whole ```_lut``` to pyqtgraph would colorize the image wrongly.

Issue #1988, opened 2021-09-09
How to take the log of a negative number


simonzihlmann:
I am opening this issue to ask about how to take the log of a negative number as I already mentioned here: https://github.com/pyqtgraph/pyqtgraph/pull/1940

I personally find it a bit strange how it is currently done. If I understand it correctly, the log is taken like this: `y = np.sign(y) * np.log10(np.abs(y)+eps)`  with `eps = np.finfo(y.dtype).eps` or `eps = 1` if y is not a float.

For me this does not make sense. The logarithm of a negative number is not defined first of all. Then to go around that problem, I kind of agree to take the absolute value of the number and then to take the log. Because this tells you how "big" a number is. Multiplying with the sign after taking the log does not make sense at all since there is a lot of information in the sign of the output of the log. It tells you whether the number was bigger or smaller than the base 10.

Let me do a small example.
y = 1e2
--> in log mode: y = 1*(2+eps) ~2 (neglecting eps)

y = -1e-5
--> in log modde: y = -1*(-5+eps)~5 (neglecting eps)

Therefore, -1e5 appears to be 3 orders of magnitude bigger than 1e2 even though the absolute value is 7 orders of magnitude smaller. Therefore, y = np.sign(y) * np.log10(np.abs(y)+eps) leads to very strange plots that are really misleading if very small negative numbers are involved.

I would propose to use y = np.log10(np.abs(y)+eps)

I am also questioning the use of eps. This could also lead to strange behaviors. Why not setting the log of y to np.nan if it is zero? Strictly speaking, log(0)-->-inf, which cannot be plotted.

This is my opinion on this and I came across this since I had some strange plots when using log for the y-scale. Maybe there is a good reasoning behind how you do. I would be curious to know. Thanks anyway for the great library!

j9ac9k:
I'll first say I'm open to a different implementation, the implementation that #1940 re-implements was created in #1476 so that may be of interest regarding reference.  If we do change up behavior here, I think we should get more input from other folks; so going to tag @NilsNemitz @ixjlyons @outofculture @ksunden and @pijyoi

NilsNemitz:
Hi @simonzihlmann !

I don't have much time to look at code these days, but I have a few quick comments on the log implementation:

[...]

PR #1992, opened 2021-09-15
Traditional log10 mode for PlotDataItem (by adding "mapped data" stage)

### LINKED ISSUE #1989: Pyqtgraph internal error handling
When I used PlotDataItem I get the following error. it's not an issue actually but I want to report it. Thank you so much.

>   File "C:\Users\HIbrahim\anaconda3\envs\moldyn\lib\site-packages\pyqtgraph\graphicsItems\GraphicsObject.py", line 24, in itemChange
>     self.parentChanged()
>   File "C:\Users\HIbrahim\anaconda3\envs\moldyn\lib\site-packages\pyqtgraph\graphicsItems\GraphicsItem.py", line 471, in parentChanged
>     self._updateView()
>   File "C:\Users\HIbrahim\anaconda3\envs\moldyn\lib\site-packages\pyqtgraph\graphicsItems\GraphicsItem.py", line 527, in _updateView
>     self.viewRangeChanged()
>   File "C:\Users\HIbrahim\anaconda3\envs\moldyn\lib\site-packages\pyqtgraph\graphicsItems\PlotDataItem.py", line 939, in viewRangeChanged
>     self.updateItems(styleUpdate=False)
>   File "C:\Users\HIbrahim\anaconda3\envs\moldyn\lib\site-packages\pyqtgraph\graphicsItems\PlotDataItem.py", line 639, in updateItems
>     x,y = self.getData()
>   File "C:\Users\HIbrahim\anaconda3\envs\moldyn\lib\site-packages\pyqtgraph\graphicsItems\PlotDataItem.py", line 781, in getData
>     top_exc =-(self._drlLastClip[0]-view_range.bottom()) / view_height
> ZeroDivisionError: float division by zero

NilsNemitz:
This attempts to address the problems ( #1988 ) with correct log-scaling mode in PlotDataItem.

The failure mode was that the `dataRect()` function describes the bounds of the original, non-log-scale data set. As a result, the dynamic range limiting function of PlotDataItem was working with very wrong assumptions and this resulted in an infinite loop of clipping changes which locked up the entire code.

The fix attempted here is to add additional data structures that hold information on the "mapped" data. Right now, "mapped" data just means log-scaled, but I hope this is compatible with a more modular approach in the future.

When log-scale is off, these `xMapped` and `yMapped` arrays should just be an additional reference to the existing data.
From the data held in `xMapped` an `yMapped`, a new `dataRectMapped` is calculated, and the dynamic range limiter seems to work as before when using this data (but see below...).

Additional changes:
- I changed the log-scale functions to simply be `np.log10()`, since I think this is what we want to use for now.
- I silenced a warning in the scaling of data coordinates to screen coordinates (in `functions`) which was triggered when working with infinities in the data. We accept Nan as silently missing points here, and I think we want the same for inf.
- The dataRect functions now return the bounds of the plottable data: This is the same as before when the data does not contain any infinities. But it does not return infinite bounds.
- I removed the "in-place clipping" method from the dynamic range limiter. I was able to repeatedly trigger a sequence that would set the display cache to the original data, and then clip a display buffer that shared memory with the original data. As far as I can tell, that bug was already there before.
- The clipping operation is now skipped when the view height is zero, which might possibly fix #1989 . Or not.

I would appreciate it if someone could try and break this, I am a little short on time currently 😢 

This needs a little more cleanup, removing of comments and print statements, and probably some tests, so I am leaving it as a draft for now.

A known "feature" is that normally, the PlotCurveItem lines skip the infinite points. But when the limiter triggers, it that clips these points to a far-away but finite value, and the lines towards "infinity" suddenly become visible. If that requires fixing, it would probably be the best to map all infinities to Nan *once*. Otherwise we end up searching the entire data set for such values repeatedly.

### DIFF
--- a/examples/logAxis.py
+++ b/examples/logAxis.py
@@ -5,34 +5,38 @@ Test programmatically setting log transformation modes.
 import initExample ## Add path to library (just for examples; you do not need this)
 
 import numpy as np
-from pyqtgraph.Qt import QtGui, QtCore
 import pyqtgraph as pg
-
 
 app = pg.mkQApp("Log Axis Example")
 
 w = pg.GraphicsLayoutWidget(show=True)
-w.setWindowTitle('pyqtgraph example: logAxis')
-p1 = w.addPlot(0,0, title="X Semilog")
+w.resize(800,800)
+w.setWindowTitle('pyqtgraph example: Log Axis, or How to Recognise Different Types of Curves from Quite a Long Way Away')
+
+p0 = w.addPlot(0,0, title="Linear")
+p1 = w.addPlot(0,1, title="X Semilog")
 p2 = w.addPlot(1,0, title="Y Semilog")
-p3 = w.addPlot(2,0, title="XY Log")
-p1.showGrid(True, True)
-p2.showGrid(True, True)
-p3.showGrid(True, True)
+p3 = w.addPlot(1,1, title="XY Log")
 p1.setLogMode(True, False)
 p2.setLogMode(False, True)
 p3.setLogMode(True, True)
+
+x = np.logspace(-1, 1, 1000) # 1000 points from 0.01 to 100
+plotdata = (
+    ('1 / 3x'    , '#ff9d47', 1./(3*x) ),
+    ('sqrt x'    , '#b3cf00', 1/np.sqrt(x) ),
+    ('exp. decay', '#00a0b5', 5 * np.exp(-x/1) ), 
+    ('-log x'    , '#a54dff', - np.log10(x) )
+)
+p0.addLegend( offset=(-20,20) )
+for p in (p0, p1, p2, p3):
+    p.showGrid(True, True)
+    p.showAxes(True, size=(40,None))
+    for name, color, y in plotdata:
+        pen = pg.mkPen(color, width=2)
+        p.plot( x,y, pen=pen, name=name )
+
 w.show()
-
-y = np.random.normal(size=1000)
-x = np.linspace(0, 1, 1000)
-p1.plot(x, y)
-p2.plot(x, y)
-p3.plot(x, y)
-
-
-
-#p.getAxis('bottom').setLogMode(True)
 
 if __name__ == '__main__':
     pg.exec()
--- a/examples/logAxis.py
+++ b/examples/logAxis.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-Test programmatically setting log transformation modes.
+Demonstrate programmatic setting of log transformation modes.
 """
 import initExample ## Add path to library (just for examples; you do not need this)
 
@@ -17,22 +17,24 @@ p0 = w.addPlot(0,0, title="Linear")
 p1 = w.addPlot(0,1, title="X Semilog")
 p2 = w.addPlot(1,0, title="Y Semilog")
 p3 = w.addPlot(1,1, title="XY Log")
+# configure logarithmic axis scaling:
 p1.setLogMode(True, False)
 p2.setLogMode(False, True)
 p3.setLogMode(True, True)
 
-x = np.logspace(-1, 1, 1000) # 1000 points from 0.01 to 100
-plotdata = (
+# 1000 points from 0.1 to 10, chosen to give a compatible range of values across curves:
+x = np.logspace(-1, 1, 1000) 
+plotdata = ( # legend entry, color, and plotted equation:
     ('1 / 3x'    , '#ff9d47', 1./(3*x) ),
     ('sqrt x'    , '#b3cf00', 1/np.sqrt(x) ),
     ('exp. decay', '#00a0b5', 5 * np.exp(-x/1) ), 
     ('-log x'    , '#a54dff', - np.log10(x) )
 )
-p0.addLegend( offset=(-20,20) )
-for p in (p0, p1, p2, p3):
-    p.showGrid(True, True)
-    p.showAxes(True, size=(40,None))
-    for name, color, y in plotdata:
+p0.addLegend(offset=(-20,20)) # include legend only in top left plot
+for p in (p0, p1, p2, p3):    # draw identical numerical data in all four plots
+    p.showGrid(True, True)    # turn on grid for all four plots
+    p.showAxes(True, size=(40,None)) # show a full frame, and reserve identical room for y labels
+    for name, color, y in plotdata:  # draw all four curves as defined in plotdata
         pen = pg.mkPen(color, width=2)
         p.plot( x,y, pen=pen, name=name )
 
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -156,9 +156,11 @@ class PlotDataItem(GraphicsObject):
         """
         GraphicsObject.__init__(self)
         self.setFlag(self.GraphicsItemFlag.ItemHasNoContents)
-        self.xData = None
+        self.xData = None # original data
         self.yData = None
-        self.xDisp = None
+        self.xMapped = None # data after applying a possible mapping, e.g. log-scale.
+        self.yMapped = None
+        self.xDisp = None # data in the form to be displayed
         self.yDisp = None
         self.curve = PlotCurveItem()
         self.scatter = ScatterPlotItem()
@@ -168,7 +170,7 @@ class PlotDataItem(GraphicsObject):
         self.curve.sigClicked.connect(self.curveClicked)
         self.scatter.sigClicked.connect(self.scatterClicked)
         self.scatter.sigHovered.connect(self.scatterHovered)
-
+        
         # self._xViewRangeWasChanged = False
         # self._yViewRangeWasChanged = False
         # self._styleWasChanged = True # force initial update
@@ -180,6 +182,7 @@ class PlotDataItem(GraphicsObject):
         self.setProperty('styleWasChanged', True) # force initial update
 
         self._dataRect = None
+        self._dataRectMapped = None
         self._drlLastClip = (0.0, 0.0) # holds last clipping points of dynamic range limiter
         #self.clear()
         self.opts = {
@@ -258,7 +261,8 @@ class PlotDataItem(GraphicsObject):
         if self.opts['fftMode'] == mode:
             return
         self.opts['fftMode'] = mode
-        self.xDisp = self.yDisp = None
+        self.xDisp   = self.yDisp   = None
+        self.xMapped = self.yMapped = None
         self.updateItems(styleUpdate=False)
         self.informViewBoundsChanged()
 
@@ -276,7 +280,8 @@ class PlotDataItem(GraphicsObject):
         if self.opts['logMode'] == [xMode, yMode]:
             return
         self.opts['logMode'] = [xMode, yMode]
-        self.xDisp = self.yDisp = None
+        self.xMapped = self.yMapped = None # invalidata mapped data
+        self.xDisp   = self.yDisp   = None # invalidate display data
         self.updateItems(styleUpdate=False)
         self.informViewBoundsChanged()
 
@@ -586,9 +591,10 @@ class PlotDataItem(GraphicsObject):
             if not isinstance(x, np.ndarray):
                 x = np.array(x)
             self.xData = x.view(np.ndarray)  # one last check to make sure there are no MetaArrays getting by
-        self._dataRect = None
-        self.xDisp = None
-        self.yDisp = None
+        self._dataRect = None # invalidate bounds for original data
+        self._dataRectMapped = None # invalidate bounds for mapped (e.g. log-scale) data
+        self.xDisp   = self.yDisp   = None
+        self.xMapped = self.yMapped = None
         profiler('set data')
 
         self.updateItems( styleUpdate = self.property('styleWasChanged') )
@@ -654,6 +660,19 @@ class PlotDataItem(GraphicsObject):
         else: # ...hide if not.
             self.scatter.hide()
 
+    def _applyLogMapping(self, x, y):
+        ### Updates the mapped (e.g. log-scale) data if needed ###
+        if self.xMapped is None:
+            self.xMapped = x
+            if self.opts['logMode'][0]:
+                with np.errstate(divide='ignore'):
+                    self.xMapped = np.log10(x)
+        if self.yMapped is None:
+            self.yMapped = y
+            if self.opts['logMode'][1]:
+                with np.errstate(divide='ignore'):
+                    self.yMapped = np.log10(y)
+        return (self.xMapped, self.yMapped)
 
     def getData(self):
         if self.xData is None:
@@ -692,16 +711,24 @@ class PlotDataItem(GraphicsObject):
         if self.opts['phasemapMode']:  # plot dV/dt vs V
             x = self.yData[:-1]
             y = np.diff(self.yData)/np.diff(self.xData)
-                
-        with np.errstate(divide='ignore'):
-            if self.opts['logMode'][0]:
-                x = np.log10(x)
-            if self.opts['logMode'][1]:
-                if np.issubdtype(y.dtype, np.floating):
-                    eps = np.finfo(y.dtype).eps
-                else:
-                    eps = 1
-                y = np.sign(y) * np.log10(np.abs(y)+eps)
+        old_y = y
+        x,y = self._applyLogMapping(x,y)
+        # print('yMapped:', self.yMapped)  
+        # with np.errstate(divide='ignore'):
+        #     if self.opts['logMode'][0]:
+        #         self.xMapped = np.log10(x)
+        #     else:
+        #         self.xMapped = x
+        #     if self.opts['logMode'][1]:
+        #         if np.issubdtype(y.dtype, np.floating):
+        #             eps = np.finfo(y.dtype).eps
+        #         else:
+        #             eps = 1
+        #         # testing:
+        #         # y = np.sign(y) * np.log10(np.abs(y)+eps)
+        #         y = np.log10(y)
+            # if self._dataRectMapped is None:
+            #     _update_dataRectMapped(x,y)
 
         ds = self.opts['downsample']
         if not isinstance(ds, int):
@@ -762,15 +789,19 @@ class PlotDataItem(GraphicsObject):
                 y = y1.reshape(n*2)
 
         if self.opts['dynamicRangeLimit'] is not None:
+            # print('mapped:', self.yMapped )
             if view_range is not None:
-                data_range = self.dataRect()
+                data_range = self.dataRectMapped()
                 if data_range is not None:
+                    # print('drange:', data_range)
+                    # print('vrange:', view_range)
                     view_height = view_range.height()
                     limit = self.opts['dynamicRangeLimit']
                     hyst  = self.opts['dynamicRangeHyst']
                     # never clip data if it fits into +/- (extended) limit * view height
                     if ( # note that "bottom" is the larger number, and "top" is the smaller one.
-                        not data_range.bottom() < view_range.top()     # never clip if all data is too small to see
+                        view_height > 0                                # never clip if the view does not show anything and would cause division by zero
+                        and not data_range.bottom() < view_range.top() # never clip if all data is too small to see
                         and not data_range.top() > view_range.bottom() # never clip if all data is too large to see
                         and data_range.height() > 2 * hyst * limit * view_height
                     ):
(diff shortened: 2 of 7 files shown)


pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -790,58 +821,97 @@ def getData(self):
...
+        selection = np.isfinite(x) # We are looking for the bounds of the plottable data set. Infinite and Nan are ignored. 
+        try:
+            xmin = np.min( x[selection] ) # find minimum of all finite values
+            xmax = np.max( x[selection] ) # find maximum of all finite values
+        except ValueError: # is reaised then there are no finite values
+            xmin = np.nan
+            xmax = np.nan
+        return QtCore.QRectF(
+            QtCore.QPointF(xmin,ymin),
+            QtCore.QPointF(xmax,ymax) )
```
j9ac9k, line 859: I was just testing this in the interpreter, thinking that constructing a QRectF with QPoints that have `np.nan` values would blow up, but it seems to work. 👍🏻

pijyoi:
The following script belongs more to the fuzzing category than to an actual use-case.
After the plot appears, try to pan or zoom it. The scatter plots points will get corrupted / disappear.
This effect is present in both master and this PR.

```python
import numpy as np
import pyqtgraph as pg
app = pg.mkQApp()
pw = pg.PlotWidget()
pw.show()

pw.setLogMode(x=False, y=True)
x = np.arange(-20, 21, dtype=np.float64)
yp = 10**x
x[20] = np.inf    # nan is fine, however

pw.plot(x, yp, symbol='x')
app.exec()
```

Note:
I don't think people have an expectation that non-finites would be handled by a plotting library to begin with.
If such non-finite values are not generated by the library itself as a result of transforming otherwise reasonable inputs, the library need not do anything.

j9ac9k:
> The fix attempted here is to add additional data structures that hold information on the "mapped" data. Right now, "mapped" data just means log-scaled, but I hope this is compatible with a more modular approach in the future.

Noticed this as I was going through the code, this is great for supporting different log-bases, sqrt, cubed-root and all sorts of transformations.  I know we've talked about having this be stored in ViewBox.py before, but I think this implementation belongs in PlotDataItem, just as you have it 👍🏻 

> A known "feature" is that normally, the PlotCurveItem lines skip the infinite points. But when the limiter triggers, it that clips these points to a far-away but finite value, and the lines towards "infinity" suddenly become visible. If that requires fixing, it would probably be the best to map all infinities to Nan once. Otherwise we end up searching the entire data set for such values repeatedly.

First off, I'm not sure we can even draw a line with a point at +/- inf even if we wanted to ... like, I assume that would be a vertical line from the previous point going either upward or downward, and a second vertical line to the point after... I'm curious what other plotting libraries do.

I would rather not get bogged down with feature creep here, if there isn't an obvious solution for drawing lines with points at -inf or inf, I'm all with going w/ @pijyoi 's input:

> I don't think people have an expectation that non-finites would be handled by a plotting library to begin with.

and move on...

NilsNemitz:
Dear @j9ac9k , 

I was going to agree with you on not overcomplicating things, but somehow the code went in a different direction...

@pijyoi 's fuzzing attack showed something pretty relevant: 
The ScatterPlotItem code doesn't seem to handle infinities well. We can basically set a "no infinities" policy, but then we shouldn't violate it ourselves when log scaling is turned on. The code now searches the mapped data and replaces any non-finites with `np.nan`, which seems to be handled appropriately.

From this search, we now *know* if the entire (log-scale mapped) dataset is finite, or if there are NaN values. We can use this information to automatically switch from the fast connection mode "all" to "finite", which can indicate missing points by interruptions in the plotted line. This has bugged me about the previous implementation: The great appeal of having a context menu switch to log scaling is data exploration. But then it doesn't seem like a good idea to silently drop the invalid points with no indication.

Keeping track of the search for non-finites required a new property `containsNonfinite`, that is True/False when the result is know, and None if the search hasn't been done. But with this added, the code started to look way too much like basic or assembler, with a growing number of arbitrary variables scattered all over the name space. To handle this, all the `xData`, `yData`, `x/yDisp` and `x/yMapped` properties have been moved to a new `PlotDataset` object that also manages the `dataRect` data and the new `containsNonfinite` flag. Compatibility methods and `@property` decorators make sure that the old interface still works.

Summary:
- log mode is now simple log10
- datasets for original data, mapped data, and display data are now handled by three instances of a new `PlotDataset` object
- dataset object can keep track of metadata, like the presence of non-finite data.
- at calculation of log-scaled data, Inf values are replaced by NaN, and information on non-finites is populated from the search.
- the line connection mode in PlotDataItem now defaults to 'auto': This falls back from 'all' to 'finite' when there are nonfinites.

Please let me know what you think. I believe these changes should not affect performance in normal, non-log-scale operation, but I am not really set up for testing that... In log mode, there is some additional overhead, but everything should be a lot lighter than the log10 calculation itself.

I am open to a less invasive implementation, but I think this adds functionality without making the code a lot messier. Tests also seem to pass :) 

To do list:
- Remove all debug outputs and dead code.
- Better tests are still needed for non-finite data, and for the modified data structures
- The 'auto' mode for connection in PlotDataItem needs documentation
- I lost the suppression of numpy warnings in the actual log10 calculation, that needs to go back in.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -38,6 +38,85 @@ class PlotDataItem(GraphicsObject):
     sigClicked = QtCore.Signal(object, object)
     sigPointsClicked = QtCore.Signal(object, object, object)
     sigPointsHovered = QtCore.Signal(object, object, object)
+    
+    class PlotDataset(object):
+        """ Holds collected information for a plotable dataset """
+        def __init__(self, x, y):
+            super().__init__()
+            assert x is not None
+            assert y is not None
```
j9ac9k, line 47: `asset` statements in functional code is generally a bad idea; they are completely ignored in the case of `python -O` so best to keep them strictly for test code.  If we want checks here; probably should raise a `RuntimeError`.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -38,6 +38,85 @@ class PlotDataItem(GraphicsObject):
     sigClicked = QtCore.Signal(object, object)
     sigPointsClicked = QtCore.Signal(object, object, object)
     sigPointsHovered = QtCore.Signal(object, object, object)
+    
+    class PlotDataset(object):
```
j9ac9k, line 42: I think this might be the first place in the library we a re using nested classes.  These are much tougher to test, so unless there is a good reason to test them, we should probably break this out into its own separate class.  

Also, if there is a chance we might want this class to emit signals, we should inherit from QObject as well, but since I'm not seeing a signal being emitted, probably no need to modify that.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -38,6 +38,85 @@ class PlotDataItem(GraphicsObject):
...
+            all_x_are_finite = selection.all() # False if all values are finite, True if there are any non-finites
+            try:
+                xmin = np.min( self.x[selection] ) # find minimum of all finite values
+                xmax = np.max( self.x[selection] ) # find maximum of all finite values
+            except ValueError: # is reaised then there are no finite values
+                xmin = np.nan
+                xmax = np.nan
+                
+            self._dataRect = QtCore.QRectF( QtCore.QPointF(xmin,ymin), QtCore.QPointF(xmax,ymax) )
+            self.containsNonfinite = not (all_x_are_finite and all_y_are_finite) # This always yields a definite True/False answer
```
j9ac9k, line 83: nit-pick level 1000 here, should it be `containsNonFinite` ?

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -38,6 +38,85 @@ class PlotDataItem(GraphicsObject):
...
+                self._updateDataRect()
+            return self._dataRect
+            
+        def applyLogMapping(self, logMode):
+            """
+            Applies log10 as a mapping transformation according to logmode.
+            logmode is a two-element list of Booleans requesting log mode for x and y axis.
+            """
+            all_x_finite = False
+            if logMode[0]:
```
j9ac9k, line 101: wonder if we should move this if block into it's own "private" function, since the code is just repeated right below.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -156,10 +235,15 @@ def __init__(self, *args, **kargs):
...
         self.setFlag(self.GraphicsItemFlag.ItemHasNoContents)
-        self.xData = None
-        self.yData = None
-        self.xDisp = None
-        self.yDisp = None
+        # self.xData = None # original data
+        # self.yData = None
+        self.dataset       = None # will hold a PlotDataset for the original data
+        self.datasetMapped = None # will hold a PlotDataset for data after mapping transforms (e.g. log scale)
+        self.datasetDisp   = None # will hold a PlotDataset for data downsampled and limited for display
```
j9ac9k, line 242: nit-pick level 1001 

I'm trying to avoid introducing new variable names that are abbreviated; may I suggest renaming to `datasetDisplay` ?

Also, these 3 datasets are closely coupled, I keep wondering if they should be members of a dictionary... but I'm not strongly opinionated here.

j9ac9k:
Just did my first pass-through the diff.  Looks like you took care to preserve API backwards compatibility 👍🏻

Other issues are mostly involve commented out code, I think there is an unused import, but that's about it.

Next up would definitely be to try and break this some more by playing w/ it.

j9ac9k:
> The following script belongs more to the fuzzing category than to an actual use-case.
> After the plot appears, try to pan or zoom it. The scatter plots points will get corrupted / disappear.
> This effect is present in both master and this PR.

so... I got a bit curious about this, and realized that in 0.11.0, while it still failed, the scatter plots would display fine if I didn't pan to the right (only panned to the left, or panned to the right up to the original starting point).  If I panned further to the right than when I started, they would disappear on me.

I git bisected when the scatter points still render when going to the left vs. not, and zero'd in on this commit: 

https://github.com/pyqtgraph/pyqtgraph/commit/71a92a6fd2421ea8204ee0dfe135337f6e069349

While the diff is small here, the diff involves some changes to `ViewBox.py` which is very much known for "here be dragons" territory.

I'm curious if what broke it was my change to Viewbox.resizeEvent, the change to padding from 0 to 0.0 or the change in how I compute the bounds in `mapFromViewToItem`

pyqtgraph/graphicsItems/PlotDataItem.py
NilsNemitz: These were added to test if any part of the test suite manages to instantiate a PlotDataset with data set to None.
To avoid replicating unneeded code everywhere, the convention throughout PlotDataItem is that a PlotDataset will only be instantiated when there is actual data. Otherwise the reference is kept as None instead.

I think this can simply be removed. I added comments to the code explaining the convention, which should only be relevant to developers working inside PlotDataItem.

pyqtgraph/graphicsItems/PlotDataItem.py
NilsNemitz: I *think* "nonfinite" is a word, at least as an adjective.
I see a parallel here between 'infinite' // 'nonfinite' and would consider NonFinite to be even less elegant :)

pyqtgraph/graphicsItems/PlotDataItem.py
NilsNemitz: Oops. This moved to PlotDataset rather late, and this copy should no longer be needed.

pyqtgraph/graphicsItems/PlotDataItem.py
NilsNemitz: --> changed to datasetDisplay.

I was trying not to jump the degree of encapsulation from zero to max in a single step. I think the individual properties are a little easier to track (compared to stuffing them all into one dictionary) while the refactored structures are still somewhat fragile. I think these are not really user-facing, and we should still be able to clean them up later.
--> renamed to "private" namespace as `_dataset`, `_datasetMapped` and `_datasetDisplay`

pyqtgraph/graphicsItems/PlotDataItem.py
j9ac9k: I'm good w/ removing them 👍🏻

pyqtgraph/graphicsItems/PlotDataItem.py
j9ac9k: Haha, I'm convinced, resolved!

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -982,159 +1175,3 @@ def dataType(obj):
 
 def isSequence(obj):
     return hasattr(obj, '__iter__') or isinstance(obj, np.ndarray) or (hasattr(obj, 'implements') and obj.implements('MetaArray'))
-
-
-
-#class TableData:
```
NilsNemitz, line 988: If I read the git log correctly, this has been commented out since 2012.

tests/graphicsItems/test_PlotDataItem.py
```diff
@@ -71,6 +72,35 @@ def test_setData():
...
     assert pdi.yData is None
+    
+def test_nonfinite():
+    def _assert_equal_arrays(a1, a2):
+        assert a1.shape == a2.shape
+        for ( xtest, xgood ) in zip( a1, a2 ):
+            assert( (xtest == xgood) or (np.isnan(xtest) and np.isnan(xgood) ) ) 
+        
+    x = np.array([-np.inf, 0.0, 1.0,  2.0  , np.nan,   4.0 , np.inf])
+    y = np.array([    1.0, 0.0,-1.0, np.inf,   2.0 , np.nan,   0.0 ])
```
NilsNemitz, line 83: This should look for crashes with combinations of infs, NaNs and regular values in normal and log mode.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -254,59 +384,80 @@ def setAlpha(self, alpha, auto):
...
-        self.xDisp = self.yDisp = None
+        self.opts['phasemapMode'] = state
+        self._datasetMapped  = None  # invalidata mapped data
+        self._datasetDisplay = None  # invalidate display data
         self.updateItems(styleUpdate=False)
         self.informViewBoundsChanged()
 
-    def setPointMode(self, mode):
-        if self.opts['pointMode'] == mode:
+    def setPointMode(self, state):
```
NilsNemitz, line 445: Any memory what this is supposed to do?
I could not find any actual function, but PlotItem seems to actually call it.

j9ac9k: Looks like it references a checkbox in PlotItem's ui template in the `plotConfigTemplate_<binding>`, something involving `pointsGroup`, supported values are None, True and False.

None `pointsGroup.isChecked()` and `autoPointsCheck.isChecked()`
True `pointsGroup.isChecked()` and _not_ `autoPointsCheck.isChecked()`
False _not_ `pointsGroup.isChecked()`

always hated those template files, not sure what uses that particular one.

j9ac9k: I'll try and take a deeper dive into the git history and figure out the source of those, and see if it's something left-behind  from some past cleanup effort.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -10,18 +9,127 @@
 from .. import debug as debug
 from .. import getConfigOption
 
-__all__ = ['PlotDataItem']
+__all__ = ['PlotDataItem', 'PlotDataset']
```
j9ac9k, line 12: If `PlotDataset` is not used outside of this file, we should probably remove it from `__all__` (as `__all__` handles the case where you input `*`)

tests/graphicsItems/test_PlotDataItem.py
```diff
@@ -71,6 +72,35 @@ def test_setData():
...
+        y_log = np.log10(y)
+    x_log[ ~np.isfinite(x_log) ] = np.nan
+    y_log[ ~np.isfinite(y_log) ] = np.nan
+
+    pdi.setLogMode(True, True)
+    dataset = pdi.getDisplayDataset()
+    print( 'good x', x_log)
+    print( 'data x', dataset.x)
+    print( 'good y', y_log)
+    print( 'data y', dataset.y)
```
j9ac9k, line 101: do we want the test suite printing stuff? (I know it does in other places), often pytest can suppress output, and I often have pytest _not_ suppress output when debugging.  Not saying this is a deal breaker; but just want to verify that the extra noise in the console here is worthwhile

NilsNemitz: Next time, just say "Hey! Print statements!"   (facepalm)

No, there are there because it took me a while to figure out why arrays are never considered equal when there are any np.nan values in there:   np.nan != np.nan.

j9ac9k: haha well right now print statements is how we do our logging 😬  could be completely legit reasons they're there.

NilsNemitz: --> stripped out

pyqtgraph/graphicsItems/PlotDataItem.py
NilsNemitz: This was primarily added to make sphinx find it when coming from the TOC list.
And without the TOC list entry, a sphinx warning ("no TOC entry") failed the CI tests.
--> Flagging the PlotDataset docstring as `:orphan:` seems to avoid this.

NilsNemitz:
I think this now has all the changes I wanted to make at this time. I would now appreciate attempts to break it :)

@pijyoi, do you have some time to fuzz it some more?

pijyoi:
I tried out the various snippets that had been posted before, and they are well behaved. (no multiple cycles to stabilize, infinites get clipped)

I am not sure if https://github.com/pyqtgraph/pyqtgraph/pull/1992#issuecomment-920507286 was intended to be fixed, but I see in the diff that the snippet https://github.com/pyqtgraph/pyqtgraph/pull/1992#issuecomment-924582802 got included.
However, I still see corruption when panning left and when zooming.
Note: I had no expectation that there was anything to fix for that script.

I do consider log10(0) as a valid use case, and it does plot nicely now.
```python
import numpy as np
import pyqtgraph as pg
pg.mkQApp()
pw = pg.PlotWidget()
pw.show()

pw.setLogMode(x=False, y=True)

xxx = np.linspace(-3, 3, 601)
yyy = np.sinc(xxx)**2

# quantize
yyy = (yyy * 16384).astype(np.int16).astype(np.float32) / 16384
print(yyy[::100])  # nulls

item = pg.PlotDataItem()
# item = pg.PlotCurveItem()

item.setData(xxx, yyy)
pw.addItem(item)
pg.exec()
```

I tend to use PlotCurveItem more often than PlotDataItem though.

j9ac9k:
> I am not sure if [#1992 (comment)](https://github.com/pyqtgraph/pyqtgraph/pull/1992#issuecomment-920507286) was intended to be fixed, but I see in the diff that the snippet [#1992 (comment)](https://github.com/pyqtgraph/pyqtgraph/pull/1992#issuecomment-924582802) got included.
> However, I still see corruption when panning left and when zooming.
> Note: I had no expectation that there was anything to fix for that script.

huh, weird! I just reran this example on the qt 5.15 and qt 6.1 bindings (both pyqt and pyside) and things looked good on zoom/pan (left) on macOS, I'll test out on Windows later.

```python
import numpy as np
import pyqtgraph as pg
app = pg.mkQApp()
pw = pg.PlotWidget()
pw.show()

pw.setLogMode(x=False, y=True)
x = np.arange(-20, 21, dtype=np.float64)
yp = 10**x
x[20] = np.inf    # nan is fine, however

pw.plot(x, yp, symbol='x')
pg.exec()
```

> Note: I had no expectation that there was anything to fix for that script.

Yeah, this is in the "nice to have" category for me, definitely not a primary objective.  

I think @outofculture said he would try and break this, so I'll give him some time to give it a go before merging.

NilsNemitz:
Sorry, I just copied and pasted the `update` method override, and only checked that it didn't cause any additional errors. :)

It does not actually improve the behavior on my system: If there is an np.inf value in the x-data, the scatter plot stops drawing entirely when the left-most symbol hits the left edge of the screen.
```
sys.platform: win32
```
I'll open a separate bug for that. Since it already works when panning to the *right*, maybe there is a simple, proper fix.

But addition of `prepareForPaint` seems to fix this on OSX, I would propose keeping that in for now, unless there is a performance concern.

@pijyoi, thank you for the sinc function example, that's a nice example of where zeros might actually come up in a real-world log-plot. The log routine now seems to fail more gracefully than before.     (...) I also have the problem of using PlotCurveItem by default; and then I am wondering why the dynamic range issue still breaks my plots when zooming in 😁

pijyoi:
If you export the example in https://github.com/pyqtgraph/pyqtgraph/pull/1992#issuecomment-927013749 to Matplotlib, the Y-axis lists only the exponents. Probably a separate issue.
image

NilsNemitz:
Yes, that very much seems like a separate issue.
It would seem that the exporter is not aware of special labeling that happens to the axes in log mode.

How do we actually want this to work?
A) Do we special-case export in log mode to read the raw data from PlotDataItem, send that to the MPL code, and activate the log mode there?
B) Or do we send the log10 values (as seems to happen now) and adjust the MPL axis ticks so that it displays the same as in pyqtgraph?

For (A) we would get to use all the work put into MPL for log display. But (B) is easier (possibly more robust) and it would be able to handle any other mapping functions we may implement...

pijyoi:
There seems to be a precedent for B.
from ```tests/exporters/test_matplotlib.py```:
```python
import pyqtgraph as pg
scale = 1e9
plt = pg.plot([0, 1, 2], [(i+1)*scale for i in range(3)])
# set the label so autoSIPrefix works
plt.setLabel('left', 'magnitude')
pg.exec()
```
image2

I didn't even know there was an ```autoSIPrefix``` feature.

ixjlyons:
Oops, see #1282 for some background on the SI prefix issue

pijyoi:
Something looks broken with ```examples/logAxis.py```
Capture

NilsNemitz:
Hmm. Is it possible that this is no longer a good example data set?
The top plot looks like the data is random over the range of +/-2. That wouldn't work really well with a log transform...

pijyoi:
So the existing example relied on np.abs() being applied, which the new implementation doesn't do any more.

How about changing the example to the following. Can visually see that it's the "expected" plot.
```python
y = np.logspace(-6, 6, 13)
x = np.logspace(-6, 6, 13)
p1.plot(x, y, symbol='x')
p2.plot(x, y, symbol='x')
p3.plot(x, y, symbol='x')
```
Capture3

NilsNemitz:
How about _something completely different_?

In #1992, I argued that the appeal of an interactive log-mode toggle is identifying properties of unknown data.
I have replaced the log plot example with one based on that argument, showing a number of plots that have somewhat similar character in a linear plot, but are easily identified by their shape in the log or semi-log plots.
Log axis example

examples/logAxis.py
```diff
@@ -5,34 +5,38 @@
...
 p2.setLogMode(False, True)
 p3.setLogMode(True, True)
-w.show()
 
-y = np.random.normal(size=1000)
-x = np.linspace(0, 1, 1000)
-p1.plot(x, y)
-p2.plot(x, y)
-p3.plot(x, y)
+x = np.logspace(-1, 1, 1000) # 1000 points from 0.01 to 100
```
pijyoi, line 24: the comment doesn't match the code

NilsNemitz: Awww crap.  :)
Thank you for spotting that, I hope the updated version is more correct.

j9ac9k:
Thanks for tackling this @NilsNemitz this was a doozy! Thanks for reviewing @pijyoi this LGTM, merging!

PR #1995, opened 2021-09-17
Change hardcoded padding into initialization parameter

Achilles1515:
The hardcoded `ColorButton` padding value of 6 is generally okay, but starts to become undesirable when the button is very small (e.g. when the button is 15px high/wide, the padding takes up 12 of these pixels, leaving almost nothing for the actual color portion.

So let's move the hardcoded value to a default value, and allow the user to be able to change this padding value and also allow defining the value on a per-side basis (via passing in a tuple or list of inputs to supply to the `QRect.adjusted()` method).


j9ac9k:
Hi @Achilles1515 

Thanks for the PR, sorry I didn't reply earlier, lost track of things while reviewing 1992.

This PR LGTM, will let the test suite do it's thing, then merge if it comes back good.

PR #1998, opened 2021-09-24
Allow ThreadTrace to save to a file

outofculture:
This small feature enhancement came out of a debug session. It saves the file relative to the working directory, and either way the output format has slightly more readable use whitespace.

### DIFF
--- a/pyqtgraph/debug.py
+++ b/pyqtgraph/debug.py
@@ -1132,10 +1132,11 @@ class ThreadTrace(object):
     Used to debug freezing by starting a new thread that reports on the 
     location of other threads periodically.
     """
-    def __init__(self, interval=10.0):
+    def __init__(self, interval=10.0, logFile=None):
         self.interval = interval
         self.lock = Mutex()
         self._stop = False
+        self.logFile = logFile
         self.start()
 
     def stop(self):
@@ -1151,12 +1152,21 @@ class ThreadTrace(object):
         self.thread.start()
 
     def run(self):
+        iter = 0
+        if self.logFile is None:
+            printFile = sys.stdout
+        else:
+            printFile = open(self.logFile, 'w', encoding='utf8')
+
         while True:
             with self.lock:
                 if self._stop is True:
                     return
-                    
-            print("\n=============  THREAD FRAMES:  ================")
+
+            if self.logFile is not None:
+                print(f"\n=============  THREAD FRAMES {iter}:  ================")
+
+            printFile.write(f"\n=============  THREAD FRAMES {iter}:  ================\n")
             for id, frame in sys._current_frames().items():
                 if id == threading.current_thread().ident:
                     continue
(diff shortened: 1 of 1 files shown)


j9ac9k:
I'm good with this PR but this screams for use of the logging module; but I know that's a whole separate effort.

pyqtgraph/debug.py
```diff
@@ -1151,12 +1152,21 @@ def start(self, interval=None):
         self.thread.start()
 
     def run(self):
+        iter = 0
+        if self.logFile is None:
+            printFile = sys.stdout
+        else:
+            printFile = open(self.logFile, 'w', encoding='utf8')
```
j9ac9k, line 1159: We open the file, but it is never closed.

outofculture:
> I'm good with this PR but this screams for use of the logging module; but I know that's a whole separate effort.

Yeah, being able to have all the logging configured to go to any or all of the log-ui, a file, stdout or stderr would be nice.

pyqtgraph/debug.py
```diff
@@ -1151,35 +1155,51 @@ def start(self, interval=None):
...
-            with self.lock:
-                if self._stop is True:
-                    return
-                    
-            print("\n=============  THREAD FRAMES:  ================")
-            for id, frame in sys._current_frames().items():
-                if id == threading.current_thread().ident:
-                    continue
+        iter = 0
+        if self.logFile is None:
```
j9ac9k, line 1159: sys.stdout behaves like a file object, so we should be able to do something like....

```python
printFile = sys.stdout if self.logFile is None else self.logFile
```

we can nick all the `if self.logFile` checks and then use `printFile.write` everywhere.

j9ac9k:
was reading the python docs and looks like we can treat `std.stdout` as a FileObject; which could simplify a lot of this code (and make the static code checker happy too :D )... is there some example that you happen to have on hand that I could test this out with?

outofculture:
I've been using it in ACQ4, which is nothing at all like a minimal example. Here's something like a reasonable test, though:

```python
from threading import Thread
from time import sleep

from pyqtgraph.debug import ThreadTrace


def sleep_a_little():
    for _ in range(3):
        sleep(1)


thread = Thread(target=sleep_a_little, daemon=True)

trace = ThreadTrace(interval=1, logFile="tt.log")
thread.start()
sleep(5)
trace.stop()
thread.join()
```

However, this script fails to put any text into the `tt.log` file, like I would expect it to, so I think I broke it.

j9ac9k:
if you're actively using this in ACQ4, and have it more or less working there as is; would you be willing to try and run a different branch of pyqtgraph on there? and see if it blows up.

tl;dr I think we can ditch both print statements and `if logfile` checks everywhere here and leverage the FileObject interface to `sys.stdout`  ... I don't mind trying to make that work if you can run within ACQ4 and make sure it works as intended.

j9ac9k:
that false positive on the static code checker is irking me... I'll mess w/ it on my branch to see if I can call `close` in a way that it convinces codeql that it will actually close the file reliably.

PR #2002, opened 2021-09-29
Run isort and pycln over entire repo - upgrade pre-commit config

j9ac9k:
This PR makes some changes such that isort and pycln can be run over the entire repository safely, and also incorporates the entire repository as well.

Some things to note, import order is important with respect to pyopengl, that needs to be imported first, so on any file that uses that as an import, I have the following at the very top.  `isort` knows not to reshuffle the order if I have the `# noqa` comment and it is before other imports.

```python
from OpenGL.GL import *  # noqa
```


j9ac9k:
Static code-checker isn't necessarily _happy_ but it's happier

image

j9ac9k:
I thought CodeQL would ignore the unused import if I had a `# noqa` next to it; but that does not appear to be the case (looking at you `initExample` unused import warning); doing another look at docs, it _might_ need to be `#noqa` (without the space) ...trying that now.

j9ac9k:
On further reflection of this, I think removing whitespaces as part of this PR will likely cause merge conflicts w/ the remaining PRs so we probably want not want to include that as part of this PR.

ksunden:
We could consider taking the examples _out_ of the CodeQL coverage (at least for now) When I've gone on my "rampage through codeQL identified problems" I've largely ignored the examples anyway, personally.

Thats just a matter of adding an "ignore" path to the action.

You can also ignore CodeQL problems via the web ui, that should cause them to not come back unless the lines are touched, I think.

j9ac9k:
> We could consider taking the examples _out_ of the CodeQL coverage (at least for now) When I've gone on my "rampage through codeQL identified problems" I've largely ignored the examples anyway, personally.
> 
> Thats just a matter of adding an "ignore" path to the action.
> 
> You can also ignore CodeQL problems via the web ui, that should cause them to not come back unless the lines are touched, I think.

...that's actually a really good idea.

j9ac9k:
@outofculture I updated this PR so that isort/pycln and such run w/o issue, removed the encoding bit, updated pre-commit (and even added `pyproject.toml` so we have some consistent settings).  I rebased `master` on this branch recently as well.

The only thing I haven't done is change the Qt namespace bits, which  of course take us back to how do we want to deal with the modules that changed their parent namespaces.  If memory serves, @pijyoi made an argument a while back about minimizing the modifications we made to the parent modules (as they could impact user code outside of pyqtgraph).

pijyoi:
It's okay to do things like QtWidgets.QAction = QtGui.QAction because we have a fake QtWidgets and QtGui.

Not okay would be QMouseEvent.position = QMouseEvent.pos as that would be visible to other libraries

j9ac9k:
> It's okay to do things like QtWidgets.QAction = QtGui.QAction because we have a fake QtWidgets and QtGui.
> 
> Not okay would be QMouseEvent.position = QMouseEvent.pos as that would be visible to other libraries

Thanks for the clarification 👍🏻

j9ac9k:
Alright folks, merging this is going to cause some merge conflicts, sorry about that...shortly after merging this we'll merge #1915 ...which is also likely to cause some merge conflicts.

PR #2007, opened 2021-10-08
skip tests using scipy if not installed

pijyoi:
scipy is an optional dependency for pyqtgraph but is used in the test suite for verifying interpolation results.
This PR skips those 2 tests if scipy is not installed.

This leaves ```numpy``` as the only mandatory dependency for running ```pytest tests```.
```pyopengl``` is additionally required to run ```pytest examples```.


j9ac9k:
Thanks @pijyoi ! Merging!

Issue #533, opened 2017-08-16
Poor performance of plot with pen width larger than 1


jenshnielsen:
I recently noticed that some of our pyqtgraph plots were a lot slower than I expected it turns out that plot performs very bad for me with a marker size larger than 1.

In https://github.com/pyqtgraph/pyqtgraph/blob/develop/examples/PlotSpeedTest.py
I can sustain around 300 fps with the default setting. However changing the pen as in 
doing ```curve = p.plot(pen=pg.mkPen(width=1.01))``` the performance drops to 2/3 fps
with a width=1 or less everything is fine.

Python 3.6
PyQt 5.9 installed from pip similar for 5.6 from Anaconda
PyQtGraph 0.10 and latest master
both windows as osx

bheklilr:
Commenting to say that I experienced this problem as well ([see stackoverflow post I wrote up about it](https://stackoverflow.com/questions/46206296/performance-issues-with-pyqtgraph)).  I'm fine with just sticking with `width=1` for now, but would really like to apply some more styling in the future to have some nicer looking plots.  This was on Windows 7 running Python 2.7, PyQt 4.10.4, and pyqtgraph 0.10.0 all installed from conda.  I am willing to help provide debugging and testing on my system if needed.

Devligue:
I don't know how it was before, but the difference between width 1 and 2 is still massive for me. At width 2 I can barely move the graph with mouse due to fps drop.

campagnola:
This is a known issue with Qt; see for example: http://www.qcustomplot.com/
index.php/support/forum/1008

There is a suggestion that using OpenGL instead can solve this problem, but
I am unable get Qt to use OpenGL on my system. Perhaps you could have
better luck:

    import pyqtgraph as pg
    pg.setConfigOptions(useOpenGL=True)

danilomendesdias:
@campagnola it worked for me. thanks.

nvaytet:
How can I re-enable the line anti-aliasing when using `useOpenGL=True`?
`pg.setConfigOptions(antialias=True)` no longer seems to have any effect and the lines are not so pretty.
Thanks!

jaxankey:
When I try to use `setConfigOptions(useOpenGL=True)` on my system I get a core dump as soon as I run something like `pg.plot([1,2,1,2,3])`. Anyone else getting the same or know of a solution?

Linux Mint 20
Anaconda Python 3.7.7

I have pyopengl installed if that helps at all.

[...]

PR #2011, opened 2021-10-09
experimental line drawing mode for thick lines

pijyoi:
Proof of concept using drawLines code from PlotSpeedTest.py.
Drawing thick lines probably involves a polygon flood fill by Qt, which gets slower the more complicated the polygon gets.
So the idea here is to draw many many short line segments. This has its own overhead but is far less than before.

To see it in action in PlotSpeedTest.py, set to line width more than 1 and enable skipFiniteCheck.

It's only meant to support fully finite data.

I don't use thick lines myself, so feel free to take over this PR.


j9ac9k:
@markotoplak this was something you previously mentioned being interested in.  If you could give this PR a spin that would be greatly appreciated!

pijyoi:
From https://github.com/pyqtgraph/pyqtgraph/issues/533#issuecomment-329490971, there is a link to https://stackoverflow.com/questions/46206296/performance-issues-with-pyqtgraph

By adding ```skipFiniteCheck=True``` to that example (to activate the changes made in this PR), "Plotting time" went from 3.925 secs to 0.106 secs
```python
plot_item = pg.PlotCurveItem(x, y, pen=pen, skipFiniteCheck=True)
```

markotoplak:
Thanks for working on this. I'll check it out!

pijyoi:
>The round caps will probably degrade performance somewhat, but they should remove the corner artifacts.

PlotSpeedTest.py (width=2) showed a drop from 97 fps to 67 fps when switching to RoundCap.

I am thinking perhaps this mode should be made even more opt-in by way of ```setConfigOption('enableExperimental', True)```, so that any surprising behavior is at the request of the user.

pijyoi:
If we compare with QtCharts, their implementation uses ```drawLine``` and the only condition they use is ```SolidLine```.
https://github.com/pyqtgraph/pyqtgraph/pull/1910#issuecomment-886204430

I was able to trigger the various other plotting artifacts (Alpha and CapStyle) with PlotSpeedTest_qtcharts.py (https://gist.github.com/pijyoi/c8a2eb963a36a5d5bb14cc3b75a8bab0)

NilsNemitz:
My opinion is that the gain is large enough to make this the default behavior when the conditions (solid, opaque) allow it. This covers the majority of things I would like to draw! The documentation should probably explain the different modes and requirements for the speed-up. I can help write that once we have the code worked out.

I do not think there is a user expectation for line end caps. This is not even an option that can be set through `mkPen`. And we have basically been telling everybody that thick lines don't work. :)   So I would say that the main user expectation is fast plotting, and endcap styling is you get what you get. It would be nice if we could detect a "don't care" setting for the endcaps, but since we basically expect the user to supply a QPen, we cannot differentiate between "SquareCap as the default", and "SquareCap because the user picked it". The best we can do is the default in mkPen, since that can still be overridden manually by modifying the QPen before passing it on. (And we might add a parameter for the endcaps later on.)

The slowdown from RoundCap seems acceptable to me: This would only apply when drawing in the "thick pen", segmented line mode. And in that case it drops the speed to 70% of a 37-fold speed-up. If I calculate right, that means the "slow" segmented line mode with CircleCap still speeds up drawing by more than 25 times relative to the previous code.

But we can do better than my earlier suggestion:
How about defaulting to RoundCaps only when it makes a significant difference?
For a 45 degree line connection, the cap edges will stick out by about (sqrt(2) - 1) * w/2 
For a width w <= 4, this is significantly less than a pixel.

Thus we could make `mkPen` default to RoundCap only when width > 4.

That gives the best speed for all "normal" linewidths, and avoids clearly visible artifacts for very wide lines. Below, orange and green are width=4 lines with square endcaps, purple and blue are width = 5 with round endcaps:
round caps at 5

Random thoughts: (( Do you have a good idea how to design a general API for optimization flags?
Otherwise, it might be helpful to just add a flag that controls this directly:
`optimize_wide_line = None` might be the default that checks for thickness, solid, opaque.
True might force this code path and allow using it with e.g. translucent lines. False would disable it.

In the long run, it would certainly make sense to have an overall "optimize = False" setting. That would make it easier to test if some specific bug results from the optimizations. It would also allow the user to turn them off if there are conflicting settings, or to keep their code working until we can fix a bug. ))

But fundamentally, this PR effectively enables a use case that wasn't very practical before, so I would see backwards compatibility as somewhat of a secondary concern. But I would love to hear some other opinions!

pijyoi:
The modified-to-use thick pens examples/crosshair.py is an example where this PR makes user-interaction feasible.

j9ac9k:
`stepMode` ....didn't even cross my mind; yeah that's going to be an issue.  

I fully support the idea of moving the logic/implementation to `PlotDataItem`.

> Maybe in another PR.

Definitely, but this change should happen sooner rather than later, we're going to need a long deprecation period here.  Suppose we should make sure there is a consensus that we migrate the `stepMode` logic to `PlotDataItem`.

sem-geologist:
`stepMode` to `PlotDataItem`? I just planned to use that (and I use mostly CurveItem, PlotDataItem is a bit too heavy for me) in my software, good I still had not. What I use is ShadowPen (for highlighting) will it be affected by this implementation?

NilsNemitz:
I think the expectation would be that ShadowPen should become quite a bit more usable, since it is kind of pointless with a single-point width.  :)   (We should double-check that, though. I played around with it, but didn't benchmark for speed.)

@sem-geologist , could you let us know what application you might have for a stepMode plot with many points?
I thought about possible use cases, and most of what I could think of revolved around histograms, or similar plots with no more than a few hundred points. In that case, the overhead of PlotDataItem shouldn't matter much.

pijyoi:
>I think the expectation would be that ShadowPen should become quite a bit more usable, since it is kind of pointless with a single-pixel width. :) (We should double-check that, though. I played around with it, but didn't benchmark for speed.)

There is one usage of shadow pen in examples/PlotWidget.py with number of samples set to 100.
Try setting it to 10000 samples.

pijyoi:
>Others may feel differently, but I think weirdness should be expected with NaN values, and there is a limitation for how much we should try and work-around issues with our dependencies (Qt in this case). If there is some obvious work-around here that doesn't impact performance, then sure, we should look into implementing it, but I don't think something like this should prevent this feature from being implemented.

It is well-defined how non-finite values are supposed to be dealt with in "all" and "finite" modes.
"all": drop non-finite points. the before and after points will join together.
"finite": non-finite points cause breaks in the line.

What's not spelt out is the behavior for "pairs" and "array". So you could say that it is left as undefined behavior. i.e. We don't handle non-finites in "pairs" and "array". You get what you get if you pass in non-finites.
As implemented in ```arrayToQPath```, probably as a side effect, they behave in an "all"-ish manner.

When I was adding non-finite support for thick pens mode for "pairs" and "array", I was originally implementing them in an "finite"-ish manner, as I thought that seemed to make more sense.

pijyoi:
After due consideration, **_for this PR_**, I am going to make thickpen "pairs" behave the same way as ```arrayToQPath```.

Any change to the "pairs" and/or "array" **behavior** in the face of non-finites should be done **_in another PR_** (if at all). It should be applied consistently to both thickpen and ```arrayToQPath```. Any change would in fact require an update to the images found in ```tests/images/plotcurveitem```.

The goal of this PR is only to enable the usage of thick pens in a **compatible** manner to users.

pijyoi:
>>It fails test_PlotCurveItem.py because of tiny line edge differences.

>Would you like me to make a new image?

No, it was expected to fail. The segments are correct, it's just the joints are very slightly different.

j9ac9k:
> For edge-case testing this PR, considering modifying `1.0` to `0.0` in the following snippet in `PlotCurveItem.py` and giving it a run on your pyqtgraph programs.
> 
> ```python
>     def _shouldUseDrawLineSegments(self, pen):
>         return (
>             pen.widthF() > 1.0
> ```
> 
> It fails `test_PlotCurveItem.py` because of tiny line edge differences.


Changed this in the application I develop internally at my company; it works exactly the same 👍🏻 

image

pijyoi:
>The one question I have is what do you think about moving LineSegments to functions.py (or someplace more accessible) so we can draw thick lines using drawLines with greater ease from other parts of the library besides PlotCurveItem.

functions.py is currently too bloated. There are some large tables and some specialised niche functions in it that should move out to slim it down. 

What parts of the library do you have in mind? Couldn't they use PlotCurveItem?

I suppose this decision to refactor can be left to the first PR that wants to make use of it? 

The class is also slightly low-level. It returns a list of objects that don't own their memory. So we may not want to make it so accessible to outside the library.

NilsNemitz:
I just wanted to say that I very much agree with @pijyoi 's assessment here. functions.py has a lot disconnected stuff in it, and we should stop adding to without need. I have run into trouble with circular imports when I really only wanted one or two of the functions that functions.py brings in. Even if PlotDataItem might take over some or all of this later, the code could still live in PlotCurveItem.py, since that will always need to get imported anyway. 

I don't really see much need to generate the current stepmode datasets outside the internal code. If we decide to offer that functionality later, it should probably bring a wider set of options and an extended API; I don't think it would be very productive to accidentally formalize the current API by putting it somewhere accessible.

pijyoi:
Sneaking in some changes to PlotSpeedTest.py
1) add "fill" parameter to demonstrate the huge performance hit due to it
2) remove drawLines proof of concept code

(When merging, please use squash)

j9ac9k:
@pijyoi I can't think you enough for tackling this issue.  When we were soliciting for feedback with what areas of the library could use improvement, this issue was the most referenced, by far.  Thanks for jumping on this.  Code looks good, and I couldn't break it so I'm going to squash and merge.

PR #2013, opened 2021-10-09
Add python publishing workflow

ksunden:
This will hopefully make releasing even easier.

Still will need to do changelog/update the version string, but now we can have github actions take over from the point of the tag.

I use this on several of my other projects, has worked fine.

Does require adding a variable to the secrets (with a pypi token)

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,33 @@
+# This workflows will upload a Python Package using flit when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install wheel twine setuptools
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: __token__
+        # The PYPI_PASSWORD must be a pypi token with the "pypi-" prefix with sufficient permissions to upload this package
+        # https://pypi.org/help/#apitoken
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        python setup.py sdist bdist_wheel
+        twine upload dist/*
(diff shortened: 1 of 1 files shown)


.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,33 @@
+# This workflows will upload a Python Package using flit when a release is created
```
j9ac9k, line 1: Looks like we don't use `flit` but use `setuptools` and `twine` ?

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,33 @@
...
+        python -m pip install --upgrade pip
+        pip install wheel twine setuptools
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: __token__
+        # The PYPI_PASSWORD must be a pypi token with the "pypi-" prefix with sufficient permissions to upload this package
+        # https://pypi.org/help/#apitoken
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        python setup.py sdist bdist_wheel
```
j9ac9k, line 32: Do we want/need to upload `sdist` distributions since we use universal wheels.  I'm not opposed to uploading them if there is some use-case that's beneficial.

.github/workflows/python-publish.yml
ksunden: Ah, holdover from converting the script...

ksunden: (I started with flit, converted to setuptools for some projects that use setuptools, so I know it works)

ksunden:
```suggestion
# This workflows will upload a Python Package using setup.py/twine when a release is created
```

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,33 @@
+# This workflows will upload a Python Package using flit when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
```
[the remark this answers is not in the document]
j9ac9k: ahh, release on tag would be nice; but this is still such a major improvement still 👍🏻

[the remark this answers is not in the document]
ksunden: To be clear, you can set the action to run on tag, but a raw tag doesn't show up as a release in github and thus you get in a situation where you have links to old versions on the main project page and it doesn't show up in the releases tab, so its just easier to make the kickoff to do a "release"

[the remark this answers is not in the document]
j9ac9k: got it; at some point should see if we can migrate the relevant section of `CHANGELOG` to the "releases" page...  but that's a whole separate thing.  Having something a CI event that does a lot of the work w/ a release would be really handy.

j9ac9k:
This LGTM! I'll let you merge at your own convenience.

PR #2012, opened 2021-10-09
Prep for 0.12.3 Release

j9ac9k:
So far this PR contains just the change-log for 0.12.3 

I kept thinking this was a small release and  am once again blown away by the number of contributions by everyone.

Thank you all for helping make pyqtgraph better!

PR #2014, opened 2021-10-09
Fix Some Import Issues

j9ac9k:
This PR addresses two issues.

1. Some parameter types used non-relative imports which does not conform to the rest of the library standards.
2. Some imports were done in such a way that when `isort` redoes the ordering, they would break.


j9ac9k:
whops, accidentally added an unused import, removing that and force pushed... when this passes I'll merge and update #2012

Issue #1937, opened 2021-07-31
showGrid with showAxis('right', ...) causes displacement between mouse mode selection area and mouse cursor position


swvanbuuren:
### Short description

For a plot with x- and y-grid lines as well as the right axis enabled, the selection area's bottom right corner (created for mouse interaction mode set to RectMode) is (stronlgy) displaced from the mouse cursor position.

### Code to reproduce

```python
import pyqtgraph as pg
from pyqtgraph.Qt import QtGui
import numpy as np

win = pg.GraphicsLayoutWidget(show=True)
win.resize(800,350)
win.setWindowTitle('pyqtgraph example: Histogram')
plt1 = win.addPlot()
plt2 = win.addPlot()

# preset mouse mode to 1 button for quick testing
plt1.getViewBox().setMouseMode(pg.ViewBox.RectMode)
plt2.getViewBox().setMouseMode(pg.ViewBox.RectMode)

# create data
x = np.arange(100)
y = np.random.normal(size=100)

plt1.plot(x, y)
plt2.plot(x, y)

plt1.showAxis('right', show=True)
plt2.showAxis('right', show=True)

plt1.showGrid(x=False, y=False)
plt2.showGrid(x=True, y=True)

QtGui.QApplication.instance().exec_()
```

### Expected behavior

The bottom right corner of the selection area coincides with the mouse cursor position.
![20210731_bug_expected](https://user-images.githubusercontent.com/5203275/127743930-e7f5dc2b-fce0-48b0-9987-7cef80c2954c.png)

### Real behavior

The bottom right corner of the selection area is (strongly) displaced from the mouse cursor position.
![20210731_bug_actual](https://user-images.githubusercontent.com/5203275/127743933-c99ea7a7-2fa9-4a51-86ec-ae0aa7b57fca.png)

### Tested environment(s)

 * PyQtGraph version: 0.12.2 
 * Qt Python binding: PySide2 5.15.1 Qt 5.15.1 
 * Python version: 3.8.10
 * NumPy version: 1.19.2 
 * Operating system: XUbuntu 20.04.2 LTS
 * Installation method: pip 

### Additional context
The displacement magnitude and direction between selection area and mouse cursor seems to depend on the activation of the x- and y- grid lines as well as the plot's right axis. For example, only activating x-grid lines leads to a strong displacement in vertical direction (contrary to the example above, where the displacement is in horizontal direction).

PR #2017, opened 2021-10-13
fix displace of the selectioon area and mouse position in "Rect Mode".

leo603222:
For a plot with x- and y-grid lines as well as the right axis enabled, the selection area's bottom right corner (created for mouse interaction mode set to RectMode) is (stronlgy) displaced from the mouse cursor position.
Detail information see #1937.
I track into the source code，and found some problems in `ViewBox.mouseDragEvent()` and `ViewBox.updateScaleBox()`.
`ev.buttonDownPos()` in` mouseDragEvent()` return a pos  in the coordinate of an AxisItem.
`r = self.childGroup.mapRectFromParent(r)` in `updateScaleBox()` needs to use a pos in ViewBox's coordinate.
These two coordinate is not consistent, and this problem cause the displace of the selectioon area and mouse position.
I fix this problem by using scene coordinate in these two places, And everything works fine to me.

### DIFF
--- a/pyqtgraph/graphicsItems/LegendItem.py
+++ b/pyqtgraph/graphicsItems/LegendItem.py
@@ -65,7 +65,7 @@ class LegendItem(GraphicsWidget, GraphicsWidgetAnchor):
         """
         GraphicsWidget.__init__(self)
         GraphicsWidgetAnchor.__init__(self)
-        self.setFlag(self.GraphicsItemFlag.ItemIgnoresTransformations)
+        # self.setFlag(self.GraphicsItemFlag.ItemIgnoresTransformations)
         self.layout = QtGui.QGraphicsGridLayout()
         self.layout.setVerticalSpacing(verSpacing)
         self.layout.setHorizontalSpacing(horSpacing)


pyqtgraph/graphicsItems/LegendItem.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, size=None, offset=None, horSpacing=25, verSpacing=0,
         """
         GraphicsWidget.__init__(self)
         GraphicsWidgetAnchor.__init__(self)
-        self.setFlag(self.GraphicsItemFlag.ItemIgnoresTransformations)
+        # self.setFlag(self.GraphicsItemFlag.ItemIgnoresTransformations)
```
j9ac9k, line 68: Hi @leo603222 

Can you either delete this line, or undo the comment and make the removal of this line part of a separate PR?

Thanks!

j9ac9k:
I've verified the PR works as intended; I'm good to merge.  I would prefer to leave the change to `LegendItem.py` to be part of a separate PR for tracking purposes.  👍🏻

pyqtgraph/graphicsItems/LegendItem.py
leo603222: Sorry, I'm not familiar with git.
I have undo the change of LegendItem, and rename my branch name too.

PR #2022, opened 2021-10-17
Added remove button option to the GroupParameterItem class

samcondon4:
Modified the GroupParameterItem base class so that when a user subclasses the GroupParameter type to create an expandable parameter group, they can add a button that will remove the latest parameter that was added to the group. 

I would like the user to be able to remove the currently selected item within the group however I have not been able to figure out how to retrieve the currently selected child within a GroupParameter instance... I wanted to open the PR now to propose this idea and ask how one can retrieve the currently selected child within a GroupParameter?


j9ac9k:
Hi @samcondon4 

I'm going to close this PR as I think @ntjess 's point addresses the issue and we haven't heard back from you.  If there is some functionality missing in the parameter trees (or other parts of the library) that you think should be present, please don't hesitate to reach out!

PR #2023, opened 2021-10-18
Optimize PColorMeshItem

pijyoi:
Generally, a ~4x speedup was obtained on the example.
The remaining bottleneck is that 3 method calls are still required per polygon.
Even if Qt _had_ a batch ```drawConvexPolygons```, it wouldn't help; because each polygon potentially has a different color.
It might be possible to iterate on the outer loop by color, but that would require treating each polygon as having unique vertices, rather than being in a regular grid.

Changes:
1) allow user to pass in a pg.Colormap
   * old limited str selection from GradientEditorItem is still permitted but not advertised
2) reuse Qt primitive objects
   * QPointF
   * QBrush
3) general code cleanup
4) add a timing printout to the example
5) change the example framerate delay to take into account processing time


pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -48,7 +86,7 @@ def __init__(self, *args, **kwargs):
 
             "ASCII from: <https://matplotlib.org/3.2.1/api/_as_gen/
                          matplotlib.pyplot.pcolormesh.html>".
-        cmap : str, default 'viridis
+        cmap : pg.ColorMap, default 'viridis'
```
j9ac9k, line 89: this type isn't quite correct, this should be `pg.ColorMap or str` no?

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -66,27 +104,31 @@ def __init__(self, *args, **kwargs):
...
-                self.cmap = kwargs['cmap']
+        self.cmap = kwargs.get('cmap', None)
+        if self.cmap is None:
+            self.cmap = colormap.get('viridis')
+
+        # Allow specifying str for backwards compatibility
+        # but this will only use colormaps from Gradients.
+        # Note that the colors will be wrong for the hsv colormaps.
+        if isinstance(self.cmap, str):
+            if self.cmap in Gradients:
```
j9ac9k, line 122: Sort of hate that we're limiting ourselves to these when referencing by string instead of using the functionality in `colormap.py` but I suppose that's for another PR.

pijyoi: "thermal", "flame", "yellowy", "bipolar" are the colormaps available in ```GradientEditorItem``` but not in ```colors/maps```.
Can we simply check for those 4 to fetch from GradientEditorItem and the rest from ColorMap?
Alternatively, ```PColorMeshItem``` being fairly new as it were, can we just ditch the use of ```GradientEditorItem```?

j9ac9k: Nice catch on the list of colors there that are not available in `colors/maps`

I'm 100% onboard with ditching `GradientEditorItem`, but I feel like this might be a bit out of scope for this PR, especially if we need to move some colors into `colors/maps` in order to retain equivalent capabilities.

j9ac9k:
I'm good with this PR as is @pijyoi only question is if we should try and remove the use of `GradientEditorItem` and migrate over the pg.ColorMap exclusively.  I feel it's a bit out of scope for this PR, but certainly wouldn't reject this PR if it included that change either.  

If we did that change, we should see about migrating the colors in `GradientEditorItem` to `colors/maps` ...but again doesn't need to part of this PR.

Sure wish the source in `GradientEditorItem.py` referenced where those color maps came from.

pyqtgraph/graphicsItems/PColorMeshItem.py
NilsNemitz: But I am also good with having to write out
`PColorMeshItem( colorMap = pg.colormap.get('viridis') )` instead of the string shorthand. It keeps the API simpler (once the legacy string gets deprecated). And this is also what `ColorBarItem` does now, so it might be a good idea for consistency.
https://pyqtgraph.readthedocs.io/en/latest/graphicsItems/colorbaritem.html#pyqtgraph.ColorBarItem

Can I bring up the option of making the officially advertised parameter `colorMap` instead of the abbrevated `cmap`?
That's what we have in ColorBarItem now, where we deprecated `cmap` with a target of summer 2022:

Since this is using `**kwargs` already, maybe checking for both variants is not too painful.

We could also keep `cmap` as the hidden legacy parameter that accepts the old strings, while the `colorMap` keyword wants a `ColorMap`. But it isn't clear to me if that is simpler/better.  :)

pijyoi: ```examples/NonUniformImage.py``` _inserts_ a custom color map into ```GradientEditorItem```. Let's hope people don't do that in the wild...

pijyoi:
Reducing the number of method calls from 3 to 2 within the inner-most loop yielded a speedup of 3/2.

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -66,27 +104,35 @@ def __init__(self, *args, **kwargs):
...
-        if 'cmap' in kwargs.keys():
-            if kwargs['cmap'] in Gradients.keys():
-                self.cmap = kwargs['cmap']
-            else:
+        if 'colorMap' in kwargs:
+            cmap = kwargs.get('colorMap')
+            if not isinstance(cmap, colormap.ColorMap):
+                raise ValueError('colorMap argument must be a ColorMap instance')
+            self.cmap = cmap
+        elif 'cmap' in kwargs:
```
j9ac9k, line 119: Thanks for including this @pijyoi the only question that I have here is do we want to emit a deprecation warning?

pijyoi: Okay, I have added a deprecation warning with a runway of 1 year.
But generally, I think it would be sufficient to provide a new API that is more featureful to gently nudge users over. e.g. pg.ColorMap has way more colors. Although in this case, PColorMeshItem is fairly new, so shouldn't have many users to begin with.

j9ac9k:
LGTM, thanks for this @pijyoi !

PR #2026, opened 2021-10-18
optimize ``create_qpolygonf`` for PySide2 and PyQt5

danielhrisca:
remove ``create_qpolygonf``: PyQt and PySide support the constructor with the size argument so this function is not needed anymore.

This change brings a small performance boots (~2us / call)

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -2184,9 +2184,12 @@ def ndarray_from_qpolygonf(polyline):
     return memory
 
 def create_qpolygonf(size):
-    if QT_LIB in ('PyQt6', 'PySide6'):
+    if QT_LIB == 'PyQt6':
         polyline = QtGui.QPolygonF()
         polyline.fill(QtCore.QPointF(), size)
+    elif QT_LIB == 'PySide6':
+        polyline = QtGui.QPolygonF()
+        polyline.resize(size)
     else:
         polyline = QtGui.QPolygonF(size)
     return polyline
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -2184,7 +2184,7 @@ def ndarray_from_qpolygonf(polyline):
     return memory
 
 def create_qpolygonf(size):
-    if QT_LIB == 'PyQt6':
+    if QT_LIB in ('PyQt6', 'PySide6'):
         polyline = QtGui.QPolygonF()
         polyline.fill(QtCore.QPointF(), size)
     else:
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -43,7 +43,7 @@ __all__ = [
     # 'ndarray_from_qimage',
     'imageToArray', 'colorToAlpha',
     'gaussianFilter', 'downsample', 'arrayToQPath',
-    # 'ndarray_from_qpolygonf', 'arrayToQPolygonF',
+    # 'ndarray_from_qpolygonf', 'create_qpolygonf', 'arrayToQPolygonF',
     'isocurve', 'traceImage', 'isosurface',
     'invertQTransform',
     'pseudoScatter', 'toposort', 'disconnect', 'SignalBlock']
@@ -1919,7 +1919,7 @@ def _arrayToQPath_all(x, y, finiteCheck):
 
     if numchunks < minchunks:
         # too few chunks, batching would be a pessimization
-        poly = QtGui.QPolygonF(n)
+        poly = create_qpolygonf(n)
         arr = ndarray_from_qpolygonf(poly)
 
         if backfill_idx is None:
@@ -1989,7 +1989,7 @@ def _arrayToQPath_finite(x, y, isfinite=None):
 
     # create a single polygon able to hold the largest chunk
     maxlen = max(len(chunk) for chunk in xchunks)
-    subpoly = QtGui.QPolygonF(maxlen)
+    subpoly = create_qpolygonf(maxlen)
     subarr = ndarray_from_qpolygonf(subpoly)
 
     # resize and fill do not change the capacity
@@ -2183,6 +2183,14 @@ def ndarray_from_qpolygonf(polyline):
     memory = np.frombuffer(buffer, np.double).reshape((-1, 2))
     return memory
 
+def create_qpolygonf(size):
+    if QT_LIB == 'PyQt6':
+        polyline = QtGui.QPolygonF()
+        polyline.resize(size)
+    else:
+        polyline = QtGui.QPolygonF(size)
+    return polyline
+
 def arrayToQPolygonF(x, y):
     """
     Utility function to convert two 1D-NumPy arrays representing curve data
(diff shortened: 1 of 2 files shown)


danielhrisca:
> I'd be shocked if this worked across the board, this change was implemented due to oddities between different bindings versions if memory serves me right, but as @pijyoi implemented this, I'll let him speak to the specifics.
> 
> EDIT: btw good to see you again @danielhrisca thanks for the PR 👍🏻

I checked the PySide2 and PyQt5 bindings and all looked good, but then Qt6 came with all the backwards incompatible changes 😞

j9ac9k:
> > I'd be shocked if this worked across the board, this change was implemented due to oddities between different bindings versions if memory serves me right, but as @pijyoi implemented this, I'll let him speak to the specifics.
> > EDIT: btw good to see you again @danielhrisca thanks for the PR 👍🏻
> 
> I checked the PySide2 and PyQt5 bindings and all looked good, but then Qt6 came with all the backwards incompatible changes 😞

yeah stuff like this practically requires `tox` to be tested locally...  ~2us / call is pretty small and within the range that I'm not sure that the difference is actually what is being measured; I'd be curious to know if PlotSpeedTest.py (or whatever benchmarks that plot line differences) can detect this change.

danielhrisca:
> > > I'd be shocked if this worked across the board, this change was implemented due to oddities between different bindings versions if memory serves me right, but as @pijyoi implemented this, I'll let him speak to the specifics.
> > > EDIT: btw good to see you again @danielhrisca thanks for the PR 👍🏻
> > 
> > 
> > I checked the PySide2 and PyQt5 bindings and all looked good, but then Qt6 came with all the backwards incompatible changes 😞
> 
> yeah stuff like this practically requires `tox` to be tested locally... ~2us / call is pretty small and within the range that I'm not sure that the difference is actually what is being measured; I'd be curious to know if PlotSpeedTest.py (or whatever benchmarks that plot line differences) can detect this change.

I did profile on an live data application were the lines are update each 100ms (~200 lines). That's were I saw the time delta (trying to remove all possible wasted CPU cyles)

j9ac9k:
> > > > I'd be shocked if this worked across the board, this change was implemented due to oddities between different bindings versions if memory serves me right, but as @pijyoi implemented this, I'll let him speak to the specifics.
> > > > EDIT: btw good to see you again @danielhrisca thanks for the PR 👍🏻
> > > 
> > > 
> > > I checked the PySide2 and PyQt5 bindings and all looked good, but then Qt6 came with all the backwards incompatible changes 😞
> > 
> > 
> > yeah stuff like this practically requires `tox` to be tested locally... ~2us / call is pretty small and within the range that I'm not sure that the difference is actually what is being measured; I'd be curious to know if PlotSpeedTest.py (or whatever benchmarks that plot line differences) can detect this change.
> 
> I did profile on an live data application were the lines are update each 100ms (~200 lines). That's were I saw the time delta (trying to remove all possible wasted CPU cyles)

Fair enough, this is quite performance sensitive bit of code.

For purposes of seeing the impact on pyqtgraph, could you suggest some alternative values to `nPlots` and `nSamples` in `examples/MultiPlotSpeedTest.py` that match up with your use-case (I still have managed to avoid pretty much ever using ParameterTree, I should modify that test to incorporate the ParameterTree just like the other line plot speed test).

j9ac9k:
I'm benchmarking this code right now with modifying the Multiple Plots to `nPlots=200` and `nSamples=2000`; I have the example automatically exit after 100 updates so the profiling results can be compared.  On the FPS counter, I couldn't notice anything different.

Running the profiler on the various configs, this is the result I got.  Values are runtimes of `create_qpolygonf` in `ms`.  This was also run on macOS.  The Qt5 5.15 bindings were on python 3.8.12 and the 5.12 bindings were installed on python 3.7.10.

| branch | master | PR |
| ----- | --- | ---- |
| pyside2 5.15 | 85 | 106 |
| pyqt5 5.15 | 124  | 137 |
| pyside2 5.12 | 110 | 66 |
| pyqt5 5.12 | 142  | 81 |

While this PR does show a significant performance boost on 5.12, it is at some minor expense of 5.15 bindings (I'm guessing difference in time here is due to the extra `elif` clause, amounting to 1 or 2 more if-checks).  Given we will likely be killing Qt 5.12 support before Qt 5.15, I'm not really convinced this is a net-gain.  If performance is that much of a premium, moving up to Qt 5.15 would likely yield the best results across the entire library.  We support 5.12 right now because that's the newest version of Qt you can get via conda-forge, and I know they're working on having PySide2 5.15 before too much longer.

It's possible other platforms besides macOS will show a more drastic improvement.

Anyone else have thoughts here?

For reference, here is the file I ran for the profiling

```python
#!/usr/bin/python
# -*- coding: utf-8 -*-
"""
Test the speed of rapidly updating multiple plot curves
"""

import initExample


import numpy as np
import pyqtgraph as pg

from time import perf_counter
# pg.setConfigOptions(useOpenGL=True)
app = pg.mkQApp("MultiPlot Speed Test")

plot = pg.plot()
plot.setWindowTitle('pyqtgraph example: MultiPlotSpeedTest')
plot.setLabel('bottom', 'Index', units='B')

nPlots = 200
nSamples = 2000
curves = []
for idx in range(nPlots):
    curve = pg.PlotCurveItem(pen=({'color': (idx, nPlots*1.3), 'width': 1}), skipFiniteCheck=True)
    plot.addItem(curve)
    curve.setPos(0,idx*6)
    curves.append(curve)

plot.setYRange(0, nPlots*6)
plot.setXRange(0, nSamples)
plot.resize(600,900)

rgn = pg.LinearRegionItem([nSamples/5.,nSamples/3.])
plot.addItem(rgn)


data = np.random.normal(size=(nPlots*23,nSamples))
ptr = 0
lastTime = perf_counter()
fps = None
count = 0
def update():
    global curve, data, ptr, plot, lastTime, fps, nPlots, count
    count += 1

    for i in range(nPlots):
        curves[i].setData(data[(ptr+i)%data.shape[0]])

    ptr += nPlots
    now = perf_counter()
    dt = now - lastTime
    lastTime = now
    if fps is None:
        fps = 1.0/dt
    else:
        s = np.clip(dt*3., 0, 1)
        fps = fps * (1-s) + (1.0/dt) * s
    plot.setTitle('%0.2f fps' % fps)

    if count == 100:
        pg.mkQApp().quit()
    #app.processEvents()  ## force complete redraw for every plot
timer = QtCore.QTimer()
timer.timeout.connect(update)
timer.start(0)

if __name__ == '__main__':
    pg.exec()
```

j9ac9k:
> I've only tested it on PyQt 5.14, so the 5.15 behavior looks like a big regression in the Qt code.

I haven't done that much testing with 5.14 specifically, but generally speaking, we see significantly better performance for line plots on PyQt5 5.15 over PySide2 (any version) and PyQt5 5.12.  This specific function might be slower on 5.15, but I suspect there are some other changes that speed it up; this function consumes 1% of the overall runtime for 100 line plot updates of 200 lines with 2000 points each; so slight performance regressions (or improvements) here would be difficult to detect.  If memory serves me right, we've seen bigger performance gains with Qt6 (both PySide and PyQt); so for performance sensitive folks, the advice currently stands to migrate to the newest Qt bindings you can use.

I'll try and benchmark against 5.14 later on.

PR #2032, opened 2021-10-20
speed up PlotCurveItem fillLevel

pijyoi:
To see it in action, run ```examples/PlotSpeedTest.py``` and check ```fill```. Compared to master, enabling ```fill``` will no longer be catastrophically slow.

What's not supported (i.e. falls back to old existing code):
1) any connect mode other than 'all'
   * no meaning to filling disjoint lines
2) ~~fillOutline = True~~
   * ~~this option seems only useful for histograms?~~
   * ~~CLARIFICATION: filling using the brush will take the fast path, it's drawing the outline using the pen _when fillOutline=True_ that is unchanged.~~
     o ~~i.e. if the pen is a thick pen, it does not take the thick pen fast path implemented in #2011~~
3) fillLevel = 'enclosed'
   * this is some undocumented option added in #1188

Help needed for testing it out.

Implementation details:
```fillPath``` has catastrophic slowdown when the path gets complex. i.e. the slowdown does not appear to be proportional to length. So this PR breaks up the filling into smaller chunks.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -57,8 +57,8 @@ class LineSegments:
         if connect == 'all':
             if not all_finite:
                 # remove non-finite points, if any
-                x = x[mask]
-                y = y[mask]
+                x = x.compress(mask)
+                y = y.compress(mask)
 
         elif connect == 'finite':
             if not all_finite:
@@ -394,6 +394,7 @@ class PlotCurveItem(GraphicsObject):
         """Set the level filled to when filling under the curve"""
         self.opts['fillLevel'] = level
         self.fillPath = None
+        self._fillPathList = None
         self.invalidateBounds()
         self.update()
 
@@ -506,6 +507,7 @@ class PlotCurveItem(GraphicsObject):
 
         self.path = None
         self.fillPath = None
+        self._fillPathList = None
         self._mouseShape = None
         self._renderSegmentList = None
 
@@ -535,7 +537,7 @@ class PlotCurveItem(GraphicsObject):
         profiler('emit')
 
     @staticmethod
-    def _generateStepModeData(stepMode, x, y, fillLevel):
+    def _generateStepModeData(stepMode, x, y, baseline):
         ## each value in the x/y arrays generates 2 points.
         if stepMode == "right":
             x2 = np.empty((len(x) + 1, 2), dtype=x.dtype)
@@ -550,19 +552,18 @@ class PlotCurveItem(GraphicsObject):
             x2[:] = x[:, np.newaxis]
         else:
             raise ValueError("Unsupported stepMode %s" % stepMode)
-        if fillLevel is None:
+        if baseline is None:
             x = x2.reshape(x2.size)[1:-1]
             y2 = np.empty((len(y),2), dtype=y.dtype)
             y2[:] = y[:,np.newaxis]
             y = y2.reshape(y2.size)
         else:
-            ## If we have a fill level, add two extra points at either end
+            # if baseline is provided, add vertical lines to left/right ends
             x = x2.reshape(x2.size)
             y2 = np.empty((len(y)+2,2), dtype=y.dtype)
             y2[1:-1] = y[:,np.newaxis]
             y = y2.reshape(y2.size)[1:-1]
-            y[0] = fillLevel
-            y[-1] = fillLevel
+            y[[0, -1]] = baseline
         return x, y
 
     def generatePath(self, x, y):
@@ -571,7 +572,7 @@ class PlotCurveItem(GraphicsObject):
                 self.opts['stepMode'],
                 x,
                 y,
-                self.opts['fillLevel']
+                baseline=self.opts['fillLevel']
             )
 
         return fn.arrayToQPath(
@@ -589,6 +590,7 @@ class PlotCurveItem(GraphicsObject):
             else:
                 self.path = self.generatePath(*self.getData())
             self.fillPath = None
+            self._fillPathList = None
             self._mouseShape = None
 
         return self.path
@@ -616,7 +618,7 @@ class PlotCurveItem(GraphicsObject):
                     self.opts['stepMode'],
                     x,
                     y,
-                    self.opts['fillLevel']
+                    baseline=self.opts['fillLevel']
                 )
 
             self._renderSegmentList = self._lineSegments.arrayToLineSegments(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by pijyoi** [comment]
An artificial example that chokes on master (slow to come up with plot, panning not usable) but works fine on this PR.
master works fine if the line ```y[::2] *= -1``` is commented out. i.e. the complexity of the curve plays a part.
I really have no idea how people are making use of ```fillLevel```, but it seems to me that for many kinds of real-world plots, master would work just fine.
```python
import pyqtgraph as pg
import numpy as np

def gaussian(x, mu, sig):
    return np.exp(-np.power(x - mu, 2.) / (2 * np.power(sig, 2.)))

pg.mkQApp()
pw = pg.PlotWidget()
pw.show()
item = pg.PlotCurveItem()
item.setPen(QtGui.QPen(QtCore.Qt.PenStyle.NoPen))
pw.addItem(item)

x = np.linspace(-3, 3, 10001)
mid = (x[:-1] + x[1:])/2
y = gaussian(mid, 0, 1)
y[::2] *= -1
item.setData(x, y, stepMode="center", fillLevel=0, brush='b')

pg.exec()
```

j9ac9k:
> I really have no idea how people are making use of `fillLevel`, but it seems to me that for many kinds of real-world plots, master would work just fine.

Whenever I'm curious about this sort of thing, I typically use github search; only "example" I found with filllevel not being 0 or None was this guy:

https://github.com/trieschlab/PymoNNto/blob/759e629568531d35b8370ad8058d4bdaadea00d0/Exploration/Network_UI/Basic_Tabs/fourier_tab.py#L132-L135

What this looks like GUI wise, I'm not sure

pijyoi:
@goodboy, you have mentioned before on Slack that you use ```stepMode``` together with ```fillLevel```. The artificial example above runs smoothly at 10,000 data points. Would you be able to try out this PR on your code?

pijyoi:
>just a note on the example, we should probably not merge red and green in such a way, they can be difficult to distinguish for individuals that are colorblind.

Thanks, I have modified it.

The example is less an example of how to use and more a check that this PR is maintaining prior behavior (intended or otherwise).
There _is_ a change in behavior for the 4th plot, although it seems safe to assume that the previous behavior of no shadow pen for the ```fillOutline``` horizontal segment was an over-sight / bug .

j9ac9k:
> The example is less an example of how to use and more a check that this PR is maintaining prior behavior (intended or otherwise).

That's completely fair.  My broader point is I don't think the naming of these options is not clear what the result _should_ look like, I think this is a use-case where just a blurb in the docs in insufficient, and where a picture speaks a thousand words...   Added benefit is that they can be checks when making changes (perhaps we should add this to one images in the test suite?

pijyoi:
I agree that having plots in the documentation would help in illustrating their use. The issue I have is that there is an inconsistency between stepMode and non-stepMode as shown in the script below.

So it might be a good idea to hold off the documentation till we have a more consistent api as discussed in https://github.com/pyqtgraph/pyqtgraph/pull/2032#issuecomment-952544754

Interestingly (for me), the bottom-row, 2nd-col plot shows that ```dataBounds()``` "knows" about ```fillLevel```. Actually, this particular plot should probably be considered a non-valid use-case. 

```python
import pyqtgraph as pg
import numpy as np

pg.mkQApp()
glw = pg.GraphicsLayoutWidget(show=True)

x = np.arange(12)
y = np.array([10, 9, 8, 7, 6, 5, 6, 7, 8, 9, 10])
xc = (x[1:] + x[:-1])/2

pen = pg.mkPen('w', width=1)
spen = pg.mkPen((0,80,0), width=10)

PCI = pg.PlotCurveItem
common = dict(pen=pen, shadowPen=spen, stepMode="center")
glw.addPlot().addItem(PCI(x, y, **common, fillLevel=None))
glw.addPlot().addItem(PCI(x, y, **common, fillLevel=0))
glw.addPlot().addItem(PCI(x, y, **common, fillLevel=0, brush='b'))
glw.addPlot().addItem(PCI(x, y, **common, fillLevel=0, brush='b', fillOutline=True))
glw.nextRow()
common = dict(pen=pen, shadowPen=spen)
glw.addPlot().addItem(PCI(xc, y, **common, fillLevel=None))
glw.addPlot().addItem(PCI(xc, y, **common, fillLevel=0))
glw.addPlot().addItem(PCI(xc, y, **common, fillLevel=0, brush='b'))
glw.addPlot().addItem(PCI(xc, y, **common, fillLevel=0, brush='b', fillOutline=True))

pg.exec()
```
stepMode_compare

NilsNemitz:
It seems strangely unsatisfying that the top and bottom version of the third plot (filled, no outline) come out differently in terms of the vertical edge lines.

@pijyoi, can your code give us direct access to switch if these lines are plotted not?
I think defaulting to the version shown in your example is a good idea, but it would be nice to have the option of plotting a filled in stepped plot without the lines.

An example use might be a display for digital I/O signals? A vertical line at start/end might then falsely imply a transition. (I admit that that is not a great example, and I'd just use the current version anyway.)

There would be a nice sense of symmetry if the lines could also be manually turned on for the non-stepped plot... I can't currently come up with a reasonable example for that, but my impression from
https://github.com/pyqtgraph/pyqtgraph/issues/2032#issuecomment-952693980_
was that you think the edge lines can be factored out of the stepmode code to be a separate processing step.

From an API perspective, I still think the edge line switching could be done through the existing `fillOutline` parameter, since it seems that it does not make any sense to switch on the edge lines AND the full outline.

But this is all nit-picking. If you don't see a way to handle this in an easy AND satisfying way, then I am good with the functionality as you have it now.

Documentation-wise, I think it might be acceptable to merge the functionality first, since the code seems to provide the same output with the same parameter settings in all typical use cases. I might be able to with a documentation path after that, but I'd prefer not to add any confusion to the code updates before then.

@j9ac9k , what do you think is the best merging procedure?

pijyoi:
So the thing is that PlotCurveItem is suitable for plotting series data (where the x-ordinate increases monotonically), but it can also be used for plotting arbitrary (x, y) and in particular can be used for drawing closed shapes.

From this point of view, ```stepMode``` is redundant. The user could just as well provide the (x, y) created by ```_generateStepModeData()```.

Even if the current API may seem non-orthogonal, there are users using ```stepMode="center", fillLevel=True``` and ```stepMode="center", fillLevel=True, fillOutline=True```. The latter usage probably stems from ```examples/histogram.py```.

>There would be a nice sense of symmetry if the lines could also be manually turned on for the non-stepped plot

So this would also be redundant for the same reason.

>But this is all nit-picking. If you don't see a way to handle this in an easy AND satisfying way, then I am good with the functionality as you have it now.

It's not easy because the existing behavior is like this:
1) ```self.path``` : data provided by the user and takes part in mouse shape.
2) ```self.fillPath``` : includes path needed to close the shape in order to fill it; takes part in dataBounds computation.

So ```stepMode``` vertical lines belong to ```self.path``` due to its original implementation.
Adding vertical lines for non-```stepMode``` should probably belong to ```self.fillPath``` because that's not data provided by the user.

Definitely not satisfying... 

As for adding more API: https://xkcd.com/927/

NilsNemitz:
**(Idle API discussion)**
I think the API is actually pretty good. Only the names of the parameters leave room for improvement.
If 'filllevel' was 'baseline', then it makes sense that this sets the shape of the edge lines, and even that auto-scale wants to include this value.
If 'fillOutline' was 'drawOutline' or 'closeOutline', then it would be clearer that the main function is to close the outline, not to fill it.
In this convention, it would be specifically 'brush=QBrush' that activates filling, since 'baseline' is disconnected from the concept of filling. That still matches the previous behavior, where -it seems- setting a brush was always required to activate filling.

**(Main PR discussion)**
Can we look at this from the point of view that
- non-step mode is now healthy and without too many strange compatibility patches (such as patching in the edge lines)
- step mode includes some compromises to keep compatibility
?

Then our long-term strategy might indeed be to find a convenient way to move the entire step mode system out of `PlotCurveItem`, either by moving the functionality to `PlotDataItem`, or by making a fully-featured API for `_generateStepModeData()`.

Once we figure something out that is good enough to justify the "one more standard" syndrome, we would slowly move to deprecate having this stepmode functionality hard-coded into PlotCurveItem.

In that case, are there still open issues to be solved?
Or is this good to go?

goodboy:
FWIW we ended up implementing a version of the `stepMode` with similar perf to our version of a "fast update" version of `PlotCurveItem` and also found this `fillLevel` stuff to be horrendously slow for any moderate sequence length.

I'm very interested in seeing if these improvements can be swapped in.
Code is in the above linked PR.

pijyoi:
>I think that further supports 'fillLevel is not None' as the main toggle switch for handling the area under the curve. (Rather than the brush setting.)

I found an in-library usage that assumes the above:
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/HistogramLUTItem.py#L191-L211

j9ac9k:
> As for adding more API: https://xkcd.com/927/

I need some kind of chrome extension for obscuring xkcd urls as I knew which one this was from the number alone.



> > I think that further supports 'fillLevel is not None' as the main toggle switch for handling the area under the curve. (Rather than the brush setting.)
> 
> I found an in-library usage that assumes the above: https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/HistogramLUTItem.py#L191-L211

I suspect this is probably what we would want (independent of the HistogramLUTItem; but having that be in practice elsewhere certainly helps).

j9ac9k:
Since creating #2079 to discuss the API inconsistency issues, I think we can merge this PR 👍🏻 

Thanks for the PR @pijyoi I think you might be running out of areas to speed up 😆

PR #2034, opened 2021-10-22
Fix displace between selection area and mouse pos

leo603222:
For a plot with x- and y-grid lines, the selection area's bottom right corner (created for mouse interaction mode set to RectMode) is (stronlgy) displaced from the mouse cursor position.
Detail information see #1937.
There's some problems in `ViewBox.mouseDragEvent()` and `ViewBox.updateScaleBox()`.
When gird is on, the mouse drag event is sent by the` AxisItem`, when grid is off, the mouse drag event is sent by `ViewBox`.
So with grid off, `ev.buttonDownPos()` in `mouseDragEvent()` return a position in the coordinate of `ViewBox`, which is right for `r = self.childGroup.mapRectFromParent(r)` in `updateScaleBox()`.
But with grid on, `ev.buttonDownPos()` in `mouseDragEvent()` return a position in the coordinate of `AxisItem`, which is not consistent with the need of `self.childGroup.mapRectFromParent(r)` in `updateScaleBox()`.
This problem is fixed by using scene coordinate in these two places.

### DIFF
--- a/pyqtgraph/graphicsItems/LegendItem.py
+++ b/pyqtgraph/graphicsItems/LegendItem.py
@@ -65,7 +65,7 @@ class LegendItem(GraphicsWidget, GraphicsWidgetAnchor):
         """
         GraphicsWidget.__init__(self)
         GraphicsWidgetAnchor.__init__(self)
-        self.setFlag(self.GraphicsItemFlag.ItemIgnoresTransformations)
+        # self.setFlag(self.GraphicsItemFlag.ItemIgnoresTransformations)
         self.layout = QtGui.QGraphicsGridLayout()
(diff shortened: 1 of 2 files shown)


leo603222:
#2017 is closed by accident, maybe because of renaming branch name. 
I changed the branch name and remove the change to LegendItem in this PR.

j9ac9k:
Thanks for the PR, long time no see :D

Diff looks good, I'm on mobile, assuming CI comes back green I'll merge when I get home.

j9ac9k:
Thanks for the PR @leo603222 Sorry for the confusion w/ branches and such.  Generally I'm happy to try and help w/ git issues like that.

This LGTM, merging!

PR #2041, opened 2021-10-24
allow unsetting options in PlotDataItem and PlotCurveItem

### LINKED ISSUE #2039: some PlotCurveItem options not resettable by PlotDataItem
### Short description

The PlotCurveItem options ```fillLevel```, ```brush``` and ```shadowPen``` are not able to be reset to their non-active state of None from PlotDataItem keyword arguments. Some of them can only be deactivated by calling methods.

```brush``` and ```shadowPen``` could be deactivated by calling ```PlotDataItem``` methods ```setBrush(None)``` and ```setShadowPen(None)```, although this only changes their styles to ```NoBrush``` and ```NoPen```, so may not be strictly equivalent to being able to reset to ```None```.

### Code to reproduce

```python
import pyqtgraph as pg
import numpy as np

def print_opts(title, opts):
    print(title)
    print("fillLevel :", opts['fillLevel'])
    brush = opts['brush']
    if isinstance(brush, QtGui.QBrush):
        print("brush :", brush.style(), brush.color().name())
    else:
        print("brush :", brush)
    pen = opts['shadowPen']
    if isinstance(pen, QtGui.QPen):
        print("shadowPen :", pen.style(), pen.color().name())
    else:
        print("shadowPen :", pen)
    print()

pg.mkQApp()

if 1:
    item = pg.PlotDataItem()
    curve = item.curve
else:
    # try out using a PlotCurveItem directly
    item = pg.PlotCurveItem()
    curve = item

# setup initial values
item.setData(np.arange(10), fillLevel=0.0, brush='b', shadowPen='g')
print_opts('INITIAL', curve.opts)

# none of the 3 options can be reset via keywords
item.setData(np.arange(10), fillLevel=None, brush=None, shadowPen=None)
print_opts('MODIFY VIA KEYWORDS', curve.opts)

# fillLevel cannot be reset via method call either
item.setFillLevel(None)
item.setBrush(None)
item.setShadowPen(None)
item.setData(np.arange(10))
print_opts('MODIFY VIA METHODS', curve.opts)
```

### Expected behavior

```
INITIAL
fillLevel : 0.0

MODIFY VIA KEYWORDS
fillLevel : None

MODIFY VIA METHODS
fillLevel : None
```

### Real behavior

OUTPUT:
```
INITIAL
fillLevel : 0.0
brush : BrushStyle.SolidPattern #0000ff
shadowPen : PenStyle.SolidLine #00ff00

MODIFY VIA KEYWORDS
fillLevel : 0.0
brush : BrushStyle.SolidPattern #0000ff
shadowPen : PenStyle.SolidLine #00ff00

MODIFY VIA METHODS
fillLevel : 0.0
brush : BrushStyle.NoBrush #000000
shadowPen : PenStyle.NoPen #c8c8c8
```
### Tested environment(s)

 * PyQtGraph version:  0.12.3
 * Python version: 3.8.10
 * NumPy version:  1.21.3
 * Installation method:  pip install -e .

### Additional context

```fillLevel``` can be reset to None by calling ```PlotCurveItem.setFillLevel(None)``` but _not_ by calling ```PlotDataItem.setFillLevel(None)```.

### PR DESCRIPTION
**Pull request by NilsNemitz**
This is a quick attempt to address #2039: Various plotting parameters can be set, but cannot be unset, especially when going through the `PlotDataItem` interfaces.

The origin of the problem is that much of the code implicitly assumes that a new element is initialized from zero, and there are explicit checks that cancel out of the update early when parameters are `None`.

This problem is present in both `PlotDataItem` and `PlotCurveItem`, but the faulty shortcut checks are particularly used in the `setData()` methods. I think the problem is worse when going through a `PlotDataItem`, because this tends to call `PlotCurveItem.setData()` with any changes, 

I made a quick pass over the code and removed problematic shortcut checks for `None`, as well as unconditional `mkBrush` / `mkPen` calls that prevented setting these to explicit None values.

The tests seem to pass, and this *seems* to address the test cases described in #2039.
But I have probably missed a few things, so some additional testing and feedback would be appreciated!

@pijyoi , do you immediately see anything that this misses?

Closes #2039 ... once it works.

### CODE REVIEW

**Comment by j9ac9k** [comment]
@NilsNemitz do you think `skipFiniteCheck` should be a sticky option along w/ the others?

**Comment by NilsNemitz** [comment]
Hmm. So far, I always read "skipFiniteCheck" as a hint for just the current dataset.

But there is an argument to be made for turning it sticky:
In many cases, your data source will never produce non-finite values. Being forced to set the hint for each `setData()` call is needless boilerplate then. 
Doing that is only useful if 
- non-finite values occur only occasionally
- the user code already knows when they are present. And where that rare case applies, a sticky flag can still be updated with every call.
So I think a sticky flag might be better.

File: pyqtgraph/graphicsItems/PlotCurveItem.py [hunk 1]
```diff
@@ -514,18 +536,18 @@ def updateData(self, *args, **kargs):
...
             self.opts['fillOutline'] = kargs['fillOutline']
-        if 'brush' in kargs and kargs['brush'] is not None:
+        if 'brush' in kargs:
             self.setBrush(kargs['brush'])
         if 'antialias' in kargs:
             self.opts['antialias'] = kargs['antialias']
-
-        self.opts['skipFiniteCheck'] = kargs.get('skipFiniteCheck', False)
+        if 'skipFiniteCheck' in kargs:
+            self.opts['skipFiniteCheck'] = kargs['skipFiniteCheck']
```
NilsNemitz, line 550: This updates skipFiniteCheck only on request. A default False value will be used if it is never explicitly set.

NilsNemitz:
The latest commit should make `skipFiniteCheck` a sticky flag that can be set once, and will still be effective during later setData() calls. It also adds a `setSkipFiniteTest()` method (for both `PlotDataItem` and `PlotCurveItem`) to allow setting/unsetting this outside of `setData()`.

I have updated the documentation for the logic and effect, and also added some text that explains the recently added handling of wide lines.

I would also suggest adding a final one-year expiration period for the long-deprecated options `stepMode=True`, `decimate` and `identical`. I have added that for now. The main motivation would be to keep the code simple. If there is a reason to keep these mostly non-functional parameters, please let me know!

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -259,18 +261,37 @@ def __init__(self, *args, **kargs):
...
+        **Notes on performance:**
+        
+        Plotting lines with the default single-pixel width is the fastest available option. For such lines,
+        translucent colors (`alpha` < 1) do not result in a significant slowdown.
+        
+        Wider lines increase the complexity due to the overlap of individual line segments. Translucent colors
+        require merging the entire plot into a single entity before the alpha value can be applied. For plots with more
+        than a few hundred points, this can result in excessive slowdown.
+
+        Since version 0.12.3, this slowdown is automatically avoided by an algorithm that draws line segments
```
pijyoi, line 288: Since 0.12.4 (or whichever version comes after 0.12.3)

pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -496,7 +515,10 @@ def updateData(self, *args, **kargs):
         if self.opts['stepMode'] in ("center", True):  ## check against True for backwards compatibility
             if self.opts['stepMode'] is True:
                 import warnings
```
j9ac9k, line 517: I know you didn't just add this, but would you mind removing this import here given the static code checker is complaining about it?

pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -416,53 +421,56 @@ def setSkipFiniteCheck(self, skipFiniteCheck):
...
         connect         Argument specifying how vertexes should be connected
-                        by line segments. Default is "all", indicating full
-                        connection. "pairs" causes only even-numbered segments
-                        to be drawn. "finite" causes segments to be omitted if
-                        they are attached to nan or inf values. For any other
-                        connectivity, specify an array of boolean values.
+                        by line segments. 
+                        
+                            | 'all' (default) indicates full connection. 
+                            | 'pairs' omits even-numbered segments.
```
pijyoi, line 456: The previous docs say that even-numbered segments are drawn, presumably counting from zero.

NilsNemitz: Hmm. If you naturally start counting from zero, then neither of these descriptions seems particularly helpful.

"  'pairs' draws one separate line segment for each two points given, starting with the first point.   " ?

pijyoi: I suppose ``` 'pairs' draws one separate line segment for each two points given.``` would be good enough. Don't think there is any doubt about it starting from the first point.

NilsNemitz: Agreed, and changed.

j9ac9k:
Thanks @NilsNemitz you clearly have been busy ...love the documentation updates too 👍🏻  merging.

PR #2045, opened 2021-10-28
Moved examples inside project directory

### LINKED ISSUE #2042: Imports of "examples" break in distributed package
### Short description
Our examples, in particular noticed with parametertree.py, import the examples as a top level import designed to work when run from the top level of the repo.

We do Path mangling to insert the folder containing `examples` into the search path for the 
### Code to reproduce

From an _installed_ version of pyqtgraph (non editable install), Not running from a directory that contains the `examples` folder
```python
python -m pyqtgraph.examples
```
Select Parameter Tree example

### Expected behavior

Parameter tree example should run
### Real behavior

fails to run with Import Error 

```
  File "/home/blaise/miniconda3/envs/edge/lib/python3.9/site-packages/pyqtgraph/examples/parametertree.py", line 15, in <module>
    from examples._buildParamTypes import makeAllParamTypes
ModuleNotFoundError: No module named 'examples'
```

### Tested environment(s)

 * PyQtGraph version: 0.12.3 (conda)
 * Qt Python binding: Not fully sure, should be orthogonal anyway (not actually my machine)
 * Python version: 3.9
 * NumPy version: Not fully sure, should be orthogonal anyway (not actually my machine)
 * Operating system: Debian
 * Installation method: conda

### Additional context

I am personally in favor of just moving the examples into the library proper and doing relative imports rather than path wrangling and imports assuming "examples" is top level.

I know the hesitation in the past has been for ease of finding the examples... and that the conclusion was that if we could reliably have symlinks the ideal solution would be to have the canonical version inside the library with a symlink to it from the top level of the repo... but symlinks are problematic on NTFS and therefore windows...

Moving it in the repo would make `python -m pyqtgraph.examples` work with editable installs as well as non-editable installs, and significantly simplify setup.py

My proposal is that we keep the examples folder in the top level, but with just a README.md that contains an index with links and a description of how to run the examples using `python -m` This keeps the discoverability of examples, while simplifying packaging and the examples themselves to always be able to do relative imports without editing path. I still think perhaps the examples have good reason to do full imports for pyqtgraph, but relative for things only within the examples submodule.

The examples assume installed anyway.

I guess the other consideration was for vendoring purposes, you can leave out the examples, but I'm not too sure that is a concern, if vendoring projects are concerned about size, it is pretty trivial to remove the folder, otherwise it has no effect.

j9ac9k:
Fixes #2042

This PR moves the examples from the `{project_dir}/examples` to `{project_dir}/pyqtgraph/examples`. 

This change was prompted by #2042 which @ksunden pointed out that some of the examples didn't work when pyqtgraph was installed in non-editable mode and the examples were run from outside the project directory.  Given the extensive path manipulation via `import initExample` on every example, and in other places, figured this would be a good time to take out some of the custom code written for pyqtgraph's `setup.py` and use standard `setuptools` functionality instead.

I also took the opportunity to update `MANIFEST.in` which only effects source distributions.

A by-product of this change means that examples may not work if pyqtgraph is not installed via pip.  Given that I'd like to phase out support for pyqtgraph to be used when _not_ via a package manager, I'm good with this change (but others may not be).

The testing I did was as follows

I launched the example app via `python -m pyqtgraph.examples` and then ran the parameter tree example ensuring it ran successfully (this example is good because it imports other python modules in the example directory, and imports from pyqtgraph)

I did this for editable and non-editable installs, from both the root project directory and a directory not within the project.

Furthermore, I ran through the test suite `pytest -n auto pyqtgraph/examples`

~What no longer works is running `python -m pyqtgraph.examples` from the project directory (example app pops up) but almost all examples fail as they attempt to import `pyqtgraph` which reports package not found.~ this works now

The only use-case I can identify where examples do not work is if a user attempts to run an example directly (ex: `python pyqtgraph/examples/PlotSpeedTest.py` _and_ pyqtgraph is not installed; the user will get an ImportError.

EDIT: slight oddity; somewhere in the development process here, I managed to break pyopengl to the point that I would get an import error; after force-reinstalling it, I could no longer replicate

```
Python 3.9.7 (default, Sep 17 2021, 12:53:27)
[Clang 12.0.5 (clang-1205.0.22.11)] on darwin
>>> from OpenGL.GL import *
  File "/Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.9.7/envs/pyqtgraph-pyqt_62-py39/lib/python3.9/site-packages/OpenGL/GL/__init__.py", line 3, in <module>
    from OpenGL import error as _error
  File "/Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.9.7/envs/pyqtgraph-pyqt_62-py39/lib/python3.9/site-packages/OpenGL/error.py", line 12, in <module>
    from OpenGL import platform, _configflags
  File "/Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.9.7/envs/pyqtgraph-pyqt_62-py39/lib/python3.9/site-packages/OpenGL/platform/__init__.py", line 21, in <module>
    from OpenGL import _configflags
  File "/Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.9.7/envs/pyqtgraph-pyqt_62-py39/lib/python3.9/site-packages/OpenGL/_configflags.py", line 2, in <module>
    from OpenGL import (
ImportError: cannot import name 'ERROR_CHECKING' from 'OpenGL' (unknown location)
>>> exit()
```


j9ac9k:
I think this PR is ready for merging; I'll give folks a bit to see if they spot any issues; otherwise I'll merge over the weekend.

pyqtgraph/examples/ExampleApp.py
```diff
@@ -470,18 +474,21 @@ def loadFile(self, edited=False):
...
-
+        else:
+            env = dict(os.environ)
+        example_path = os.path.abspath(os.path.dirname(__file__))
+        path = os.path.dirname(os.path.dirname(example_path))
+        env['PYTHONPATH'] = f'{path}'
         if edited:
-            path = os.path.abspath(os.path.dirname(__file__))
             proc = subprocess.Popen([sys.executable, '-'], stdin=subprocess.PIPE, cwd=path, env=env)
-            code = str(self.ui.codeView.toPlainText()).encode('UTF-8')
```
NilsNemitz, line 477: This encoding still seems to be required, see issue #2060 .

PR #2051, opened 2021-10-30
RemoteGraphicsView.py : implement pickle protocol

pijyoi:
RemoteGraphicsView does manual serialization and deserialization of mouse events because not all the Qt objects are picklable and different Qt bindings have different degrees of pickability.
This PR reimplements the serialization and deserialization to make use of the Python pickle protocol. In particular, it makes the deserialization transparent to the Renderer class.


j9ac9k:
Love this diff @pijyoi definitely looks cleaner.  Thanks for the PR!

PR #2052, opened 2021-10-31
make PlotItem's average pen and shadow pen accessible

### LINKED ISSUE #2040: Plot Options -> Average invokes thick pen plotting slow path
### Short description

Since #2011, thick pen plotting has become more usable. Plot Options -> Average uses a pen that is not supported by #2011, resulting in a very slow plot.

### Code to reproduce

Run ```examples/crosshair.py```.
On the upper plot, right-click -> context menu -> ```Plot Options``` -> ```Average``` -> check ```average```

### Expected behavior

No slow-down

### Real behavior

Panning (amongst other operations) will become very slow, to the extent of being unusable.

### Tested environment(s)

 * PyQtGraph version:  git ef458d3bbae513f6e58604332acf3c901f304e9e
 * Python version: 3.10.0
 * NumPy version:  1.21.3
 * Installation method:  pip install -e .

### Additional context
In ```pyqtgraph/graphicsItems/PlotItem/PlotItem.py```, a shadow pen of color (0, 0, 0, 100), width=3 is created. Non-opaque pens are not supported by #2011, hence the slow-down.

Not sure what visual effect a non-opaque black pen is supposed to achieve.

NilsNemitz:
#2040 points out that the average curve that can be activated from the context menu of `PlotItem` includes a wide shadowPen . This is by default set to be partially opaque, which prevents fast drawing as individual line segments.

This PR implements a minimalistic fix to this problem by redefining the shadow pen to be a solid black. This should still ensure the visibility of the average line over any kind of background.

To allow somewhat improved access to the colors of the average line, it defines these as `Plotitem.avgPen` and `PlotItem.avgShadowPen`, which can be replaced by user code (as long as this is done before the averaging mode is activated).

Finally, it applies this restyling to the crosshair.py example to improve the chance to find this option if needed.

crosshair

A follow-up step would be to create a proper API, but this would probably need to include code to handle redrawing the line when restyled, and would thus be more complex than this quick fix :)

This closes #2040: With the solid black shadow pen that is now the default, the example plot pans smoothly even with the average enabled.


j9ac9k:
Thanks for the PR @NilsNemitz guess this one was easy to overlook as we had written off thick-lines.

PR #2053, opened 2021-10-31
Disable and comment the prepareForPaint call in ViewBox.update()

### LINKED ISSUE #2029: Bug introduced in setState in version 0.11.x, 0.12.3 (but it's working on 0.12.0!)
### Short description
Prior to version of pyqtgraph 0.11.x, I heavily used the **setState** property in order to be able to keep the pan and zoom settings the user defined. For example, some of my application have the imageView on top and a slider at the bottom. The user click on the slider to display a different image. But if he pan or zoom into the image, switching to another image keeps the pan and zoom settings. Other applications I wrote modified the image and refresh it, still keeping the pan and zoom settings as shown in this animation. 

without_issue

But if I'm using pyqtgraph 0.11.x or later, the view auto-reset. and a single action on the mouse (little pan, little zooming) correctly reset the image view to the range previously defined. 

withIssue

I tested the issue with various pyqtgraph version and came up with the following conclusion:
0.10.x -> it's working
0.11.0 -> not working
0.12.0 -> it's working
0.12.3 -> not working

### Code to reproduce

Here is some of the code that I use to record the image settings before plotting a new image, and then resetting the 
view according to the parameters saved

```python

[...]

 _view = self.parent.ui.image_view.getView()
 _view_box = _view.getViewBox()
 _state = _view_box.getState()

  self.parent.ui.image_view.setImage(_image)
  _view_box.setState(_state)

[...]

```

### Expected behavior
The new plot should have the same view state as the previous plot

### Real behavior
The new plot state is reset. 

```
An error occurred?
```

### Tested environment(s)

 * PyQtGraph version: tested with 0.10.x, 0.11.0, 0.12.0, 0.12.3
 * Python version:  3.7
 * NumPy version: 1.21.0
 * Operating system:  Mac OS, Linux
 * Installation method: Using conda environment for most librairies. Used pip for pyqtgraph.

NilsNemitz:
This addresses #2029, where a recent change to ViewBox was found to result in an unwanted reset of viewing range with each update of data. All the credit for experimentation and triangulating the problem goes to @pijyoi , @j9ac9k and @JeanBilheux .

But since the problem was introduced as part of my PR #1992, I thought I should propose a PR to fix it.

This change reverts the addition of a `prepareForPaint()` call in `ViewBox.update()`. This was added since it appeared to result in a more user-friendly failure mode for the basically unsupported case of data with infinite values, at least on OSX. This addition was not part of the main functionality of #1992, but no side effects were known or expected at the time. 

But since such side effects clearly exist after all, I believe a revert to the previous state is the best solution.
This was found to solve the issue in https://github.com/pyqtgraph/pyqtgraph/issues/2029#issuecomment-952408770_

Closes #2029

### DIFF
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -1468,10 +1468,17 @@ class ViewBox(GraphicsWidget):
 
         bounds = QtCore.QRectF(range[0][0], range[1][0], range[0][1]-range[0][0], range[1][1]-range[1][0])
         return bounds
-        
-    def update(self, *args, **kwargs):
-        self.prepareForPaint()
-        GraphicsWidget.update(self, *args, **kwargs)
+
+    # Including a prepareForPaint call is part of the Qt strategy to
+    # defer expensive redraw opertions until requested by a 'sigPrepareForPaint' signal
+    # 
+    # However, as currently implemented, a call to prepareForPaint as part of the regular 
+    # 'update' call results in an undesired reset of pan/zoom:
+    # https://github.com/pyqtgraph/pyqtgraph/issues/2029
+    #
+    # def update(self, *args, **kwargs):
+    #     self.prepareForPaint()
+    #     GraphicsWidget.update(self, *args, **kwargs)
 
     def updateViewRange(self, forceX=False, forceY=False):
         ## Update viewRange to match targetRange as closely as possible, given


j9ac9k:
Thanks for this @NilsNemitz 

I appreciate the comment you put in the code here; so much of the code in ViewBox.py has so much in the way of subtle impacts.  

Only question I have is if there is an "easy" test to add to test for that catches the error that was detailed to `test_ViewBox.py`.  I'll try and do some experimentation today.

j9ac9k:
I was trying to do an "easy" test, but noticed the axis coordinates were changing w/ the zoom behavior, but the image was not, so short of some `assertImageMatch` business; which I'm not really inclined to do, I can't think of a good way to test for this.  Given there is an existing/documented issue and we have confidence in this fix, I'll merge this.  Thanks @NilsNemitz for the PR!

PR #2054, opened 2021-10-31
Maintain limits on ViewBox scaling

### LINKED ISSUE #2001: Axis can be zoomed out until program becomes unresponsive
When working in log mode, it is very easy to zoom out until `AxisItem` can no longer handle the magnitude of the numbers involved.
An easy fix might be to set a default limit on the maximum range?

### Steps to reproduce

- Run the code below to show a plot in y-axis logarithmic mode.
- Point at the vertical axis.
- Spin the mousewheel down. 
- Spin it again. On my system, this increases the displayed range to more then 1E-300 to 1E+300. Any values outside this range display as "inf" and "0".
- The following warning starts appearing: `..\pyqtgraph\graphicsItems\AxisItem.py:845: RuntimeWarning: overflow encountered in power estrings = ["%0.1g"%x for x in 10 ** np.array(values).astype(float) * np.array(scale)]`
- Spin the wheel one more time.
- The only labels are now "0", "1" and "inf".
- The application becomes unresponsive. 
- If it recovers, spin down again, it will lock up completely eventually.

```python
import numpy as np
import pyqtgraph as pg
pg.mkQApp()
pw = pg.PlotWidget()
pw.setLogMode(x=False, y=True)
pw.show()

xxx = np.linspace(-3, 3, 601)
yyy = np.sinc(xxx)**2
yyy = (yyy * 16384).astype(np.int16).astype(np.float32) / 16384 # quantize to make nulls 0

item1 = pg.PlotDataItem(pen='r')
item1.setData(xxx,yyy)
pw.addItem(item1)
pg.exec()
```

### Expected behavior

The program should not become unresponsive.

Beyond that, I am hoping for comments :)
To me, it does not seem very useful to try and display values outside the +/-1.7E+308 (and E-308) range that a double can handle. So it might be sufficient not to allow any larger range than that, even if working in log mode?

### Tested environment(s)
```
sys.platform: win32
```

NilsNemitz:
As described in #2001 , the absolute limits of values represented by double-precision floating-point values can be reached very quickly by zooming out in logarithmic mode: The `AxisItem`s start calculating with out of range values when the logarithmic representation exceeds +/-308, which can be reached with two or three spins of the mouse wheel.

In normal mode, it takes a little longer to get there, but it is similarly possible to create failures by simply zooming out too far.

This PR initialized the range limits in normal mode to be +/-1E307 (close to the absolute range limits, with a slight extra margin to avoid earlier out-of-range values in internal calculations. It also applies additional limits of about +/308 to the allowed ViewBox scaling range when in logarithmic mode, corresponding to log10 of the range limit.

Downsides:
- ViewBox now has to be made aware when logarithmic mapping is activated for one of the axes, which was not the case previously.
- The limiting code now executes on every update, although I do not expect the limited number of additional calculations to present a noticeable limit to performance.

I am hoping for comments.

This closes #2001.

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -210,9 +210,15 @@ class AxisItem(GraphicsWidget):
         If *log* is True, then ticks are displayed on a logarithmic scale and values
         are adjusted accordingly. (This is usually accessed by changing the log mode
         of a :func:`PlotItem <pyqtgraph.PlotItem.setLogMode>`)
+        The linked ViewBox will be informed of the change.
         """
         self.logMode = log
         self.picture = None
+        if self._linkedView is not None:
+            if self.orientation in ('top', 'bottom'):
+                self._linkedView().setLogMode('x', log)
+            elif self.orientation in ('left', 'right'):
+                self._linkedView().setLogMode('y', log)
         self.update()
 
     def setTickFont(self, font):
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -164,11 +164,14 @@ class ViewBox(GraphicsWidget):
             'wheelScaleFactor': -1.0 / 8.0,
 
             'background': None,
+            
+            'logMode': [False, False],
 
             # Limits
-            'limits': {
-                'xLimits': [None, None],   # Maximum and minimum visible X values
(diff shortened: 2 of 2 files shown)


j9ac9k:
Upon further check for adding a test, I noticed the axes change, but the rendering doesn't; short of some kind of `assertImageMatch` or something like that; this might be a tougher case to test for; and given this PR fixed an issue, I'm going to merge this as is.

Thanks for the PR @NilsNemitz and thanks for generating the example of the issue in #2001 @pijyoi

Issue #1963, opened 2021-08-11
Jupyterlab support


j-bac:
### Short description
pyqtgraph does not seem to work in Jupyterlab

### Code to reproduce
```python
%gui qt
import pyqtgraph as pg
pg.plot([1,2,3,2,3])
```

### Expected behavior
Graph display
### Real behavior
Kernel dies

### Tested environment(s)

 * PyQtGraph version: 
 * Qt Python binding: 'PyQt5 5.15.4 Qt 5.15.2'
 * Python version: 3.7.9
 * NumPy version: 
 * Operating system: Ubuntu
 * Installation method: pip

j9ac9k:
We've had some requests to have better support for jupyter, I generally don't use that environment so I don't know much about it.  We're certainly open to expanding support for jupyter lab, but I'm not even sure where to begin in that regard!

I just experimented w/ the jupyter qt console, and things work provided you set the QT_API variable to the correct bindings

```
import os
os.environ['QT_API']='pyside2'

%gui qt
import pyqtgraph as pg
pg.plot([1, 2, 3])
Out[6]: <pyqtgraph.widgets.PlotWidget.PlotWidget(0x7fe3b9bec450) at 0x12b28f940
```

I'm not sure what jupyter lab does differently here though.  Any input/suggestions would be helpful!

ksunden:
I just tried your exact example in jupyterlab and it worked as expected... perhaps something is different with different qt apis? (I was using pyqt5 without specifying it, `%gui qt` set  `QT_API=pyqt5`, even if I had previously set it to pyside2...)

j-bac:
Thanks for the prompt reply. That is so weird. I tried running your examples, setting the environment variable but it doesn't work.

After running %gui qt in a separate cell, the kernel dies when executing the next line or cell that is executed

```
jupyter core     : 4.6.3
jupyter-notebook : 6.1.1
qtconsole        : 5.1.1
ipython          : 7.17.0
ipykernel        : 5.3.4
jupyter client   : 6.1.6
jupyter lab      : 2.2.9
nbconvert        : 5.6.1
ipywidgets       : 7.5.1
nbformat         : 5.0.7
traitlets        : 4.3.3
```

````
PyQt5==5.15.4
PyQt5-Qt5==5.15.2
PyQt5-sip==12.9.0
PyQt6==6.1.1
PyQt6-Qt6==6.1.2
PyQt6-sip==13.1.0
[...]

PR #2055, opened 2021-11-02
Jupyter PlotWidget example

pijyoi:
This PR is a continuation of https://github.com/pyqtgraph/pyqtgraph/issues/1963#issuecomment-955871254. Github does not allow attaching ```ipynb```, so creating a PR seemed the easier way to continue the discussion.

A standalone Jupyter Notebook is in this PR that implements a Jupyter PlotWidget and example usage.
What works:
1) resizing
2) mouse events (panning / zooming)
3) running under WSL2. Set the environment variable ```QT_QPA_PLATFORM=offscreen``` before running ```jupyter-notebook```

What doesn't work well:
1) mouse wheel zoom
   * easy to accidentally trigger it while intending to scroll the notebook page.

What doesn't work:
1) context menu caused a crash so is disabled
   * doesn't fit into the role of a notebook widget anyway

Eventually, the PlotWidget class could reside somewhere else within pyqtgraph codebase. But right now, the easiest way to try it out now is to just download the file and run it with your existing installation of pyqtgraph. (plus ```jupyter_rfb``` and ```pillow```)

Some notes:
1) ```show()```-ing the underlying ```GraphicsView``` widget causes a window to pop up (on Windows platform anyway). Fortunately, rendering to a QImage works fine without the ```show()```.
2) That it works without having to run the Qt event loop is nice. Note that the mouse events are sent synchronously using ```sendEvent```


j9ac9k:
I just tried this out, my goodness, this is amazing @pijyoi ... I had completely written off this functionality some time ago...

This works as intended in Chrome, in Safari none of the interactive bits were working, and I got the following error in the console:

```
RuntimeError: Cannot enter into task <Task pending name='Task-47' coro=<HTTP1ServerConnection._server_request_loop() running at /Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.9.7/envs/pyqtgraph-pyqt_62-jupyter-py39/lib/python3.9/site-packages/tornado/http1connection.py:823> wait_for=<Future finished result=b'GET /static...flate\r\n\r\n'> cb=[IOLoop.add_future.<locals>.<lambda>() at /Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.9.7/envs/pyqtgraph-pyqt_62-jupyter-py39/lib/python3.9/site-packages/tornado/ioloop.py:688]> while another task <Task pending name='Task-2' coro=<KernelManager._async_start_kernel() running at /Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.9.7/envs/pyqtgraph-pyqt_62-jupyter-py39/lib/python3.9/site-packages/jupyter_client/manager.py:337>> is being executed.
ERROR:asyncio:Exception in callback <TaskWakeupMethWrapper object at 0x105801a60>(<Future finis...late\r\n\r\n'>)
handle: <Handle <TaskWakeupMethWrapper object at 0x105801a60>(<Future finis...late\r\n\r\n'>)>
  File "/Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.9.7/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
RuntimeError: Cannot enter into task <Task pending name='Task-48' coro=<HTTP1ServerConnection._server_request_loop() running at /Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.9.7/envs/pyqtgraph-pyqt_62-jupyter-py39/lib/python3.9/site-packages/tornado/http1connection.py:823> wait_for=<Future finished result=b'GET /nbexte...flate\r\n\r\n'> cb=[IOLoop.add_future.<locals>.<lambda>() at /Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.9.7/envs/pyqtgraph-pyqt_62-jupyter-py39/lib/python3.9/site-packages/tornado/ioloop.py:688]> while another task <Task pending name='Task-2' coro=<KernelManager._async_start_kernel() running at /Users/ogi/.zinit/plugins/pyenv---pyenv/versions/3.9.7/envs/pyqtgraph-pyqt_62-jupyter-py39/lib/python3.9/site-packages/jupyter_client/manager.py:337>> is being executed.
```

More errors, I don't think this is an issue with our end of the code, so I'm mostly putting this for record keeping purposes.

pijyoi:
I found somewhere and copied it down (but now I can't find the webpage again) that the following will workaround the asyncio error. But the error doesn't seem to prevent the notebook from working, so I just launch it the simple way now.
```
# avoid asyncio error
jupyter notebook --NotebookApp.kernel_manager_class=notebook.services.kernels.kernelmanager.AsyncMappingKernelManager
```

j9ac9k:
nit-pick part 2, on macOS I'm getting the "bouncing icon" after I run the first cell...

image

This normally occurs when launching an application; I can try and investigate this one a bit more since I'm on this platform.

pijyoi:
>mouse wheel zoom
easy to accidentally trigger it while intending to scroll the notebook page.

Seems like there is some web standard for mouse-wheel : https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent
1) wheel alone --> scroll
2) wheel + CTRL --> zoom

pyqtgraph, as a Desktop application, doesn't consider the CTRL modifier for mouse wheel, which makes sense for it since ```GraphicsView (is a) Re-implementation of QGraphicsView that removes scrollbars```

almarklein:
> pyqtgraph, as a Desktop application, doesn't consider the CTRL modifier for mouse wheel

I think most viz applications prefer to keep modifiers to themselves :)  there are other ways though.

pyqtgraph/jupyter/GraphicsView.py
```diff
@@ -0,0 +1,127 @@
...
+                btns = get_buttons(event["buttons"])
+                mods = get_modifiers(event["modifiers"])
+                typ = TYPLUT[event_type]
+                evt = QtGui.QMouseEvent(typ, pos, pos, btn, btns, mods)
+                QtCore.QCoreApplication.sendEvent(self.gfxView.viewport(), evt)
+                self.request_draw()
+        elif event_type == "wheel":
+            pos = QtCore.QPointF(event["x"], event["y"])
+            pixdel = QtCore.QPoint()
+            scale = -1.0    # map JavaScript wheel to Qt wheel
```
almarklein, line 88: From what I've seen, Qt uses steps of 120, versus 96 of JS, so useing `-120/96 == -1.25`  here it should be more precise.

pyqtgraph/jupyter/GraphicsView.py
```diff
@@ -0,0 +1,127 @@
...
+            angdel = QtCore.QPoint(int(event["dx"] * scale), int(event["dy"] * scale))
+            btns = get_buttons([])
+            mods = get_modifiers(event["modifiers"])
+            phase = QtCore.Qt.ScrollPhase.NoScrollPhase
+            inverted = False
+            evt = QtGui.QWheelEvent(pos, pos, pixdel, angdel, btns, mods, phase, inverted)
+            QtCore.QCoreApplication.sendEvent(self.gfxView.viewport(), evt)
+            self.request_draw()
```
almarklein, line 96: Typically the event will be emitted into the application, and if the application reacts to it with changes that need a redraw, it will call `request_draw`. Then again, this should probably work too.

pijyoi: The implementation done here has jupyter_rfb using pyqtgraph as a library; rather than pyqtgraph using jupyter_rfb as a backend. i.e. pyqtgraph has no idea that it is being used in jupyter_notebook, and would not be able to call ```request_draw()```

I did initially start off with a class inheriting from both pyqtgraph's GraphicsView and RemoteFrameBuffer; but got this error with PySide6: ```TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases```

pyqtgraph/jupyter/GraphicsView.py
pijyoi: I tried out with a mouse wheel. On Qt Desktop, each step generated a delta of 120. But on jupyter_rfb (on Windows Chrome), each step generated a delta of 200 (unscaled). But you are saying that you are getting 96 per step?

pijyoi:
Based on @almarklein 's comment on not calling ```request_draw()``` upon any mouse event, I connected 2 of ```ViewBox```'s signals (```sigRangeChangedManually``` and ```sigStateChanged```) to ```request_draw()```. Having a "pointer_move" trigger a ```request_draw()``` doesn't sound correct.

There are some other signals (```sigYRangeChanged```, ```sigXRangeChanged```, ```sigRangeChanged```, ```sigTransformChanged```). Not being familiar with ```ViewBox```, I am not sure if they should be connected.
Should I connect them all just to be safe? Or should I only connect those known to be needed?

pyqtgraph/jupyter/GraphicsView.py
almarklein: Yes, I get 96 on Windows Firefox. I assumed that this was more or less standard. Arg. I'll do some more research.

almarklein: I got 100 on Chrome. I wonder why you get 200. Do you happen to be using a high-rez display?

pijyoi: Oh yes, I was on a laptop with dpr 2.0.

almarklein: I suppose you "feel" the difference in scrolling too. What about using the touchpad, does it feel more sensitive too?

almarklein: > Oh yes, I was on a laptop with dpr 2.0.

Could you try on Firefox too please? I don't have access to my hidpi screen right now :)

pijyoi: On the same Windows laptop with dpr 2.0:
Chrome: 200
Firefox: 96

almarklein: I made an issue, with a little test html to get some more details: https://github.com/vispy/jupyter_rfb/issues/50

pijyoi:
Ok, so the downside with not requesting draw on any mouse event is that ViewBox RectMode no longer works since the frame is not redrawn to draw the rubber band.

almarklein:
I don't know the internals of pyqtgraph, but does it no have an update/request_draw method or something that you can tap into? Requesting a new draw on user interaction can be a practical solution, but its not airtight, because there are other events (thinking mostly of timers for animations) that may require a redraw.

pijyoi:
As far as I can tell, pyqtgraph internals is Qt. In Desktop Qt, ```update()``` is the equivalent of ```request_draw()```, but is non-virtual, so we can't override it.

pijyoi:
>Ok, so the downside with not requesting draw on any mouse event is that ViewBox RectMode no longer works since the frame is not redrawn to draw the rubber band.

So I thought I had a working workaround for the above, but that only handles actions taken by the ViewBox. There are other graphicItems that accept clicks and hover events. (The included ScatterPlot example is one of them) The simplest solution for now is still to request_draw() on each pointer_* event.

pijyoi:
Compared to the Desktop Qt version, the ScatterPlot example doesn't show tooltip on hover.

There are 2 main items that support clicks, PlotCurveItem and ScatterPlotItem. Clicks don't need special handling, the user, if they hook onto the click signal, could just call request_draw() within their callback.

ScatterPlotItem in addition supports hover and the library automatically creates a tooltip on hover, which is already not working in this rfb implementation.

If we reduce the scope of what we want to achieve in this implementation, we could remove the request_draw() on pointer_move.

j9ac9k:
Hi @pijyoi 

This looks good, I did run into one weird behavior, if you run the plotting notebook, and manipulate the linear region item in the 8th plot, the end result isn't quite right.

Also super nit-picky, but probably should remove the encoding pragma.

pijyoi:
I added the packages that pyqtgraph adds in its CI, but the kernel still crashes upon instantiation of the QApplication. You can try it out here:
https://mybinder.org/v2/gh/pijyoi/pyqtgraph/jupyter_rfb?labpath=pyqtgraph%2Fexamples%2Fnotebooks

>if you run the plotting notebook, and manipulate the linear region item in the 8th plot, the end result isn't quite right.

I don't see a difference with the Desktop version though.

j9ac9k:
> > if you run the plotting notebook, and manipulate the linear region item in the 8th plot, the end result isn't quite right.
> 
> I don't see a difference with the Desktop version though.

I'll capture some video shortly.

pijyoi:
okay, to get it to work on binder, I had to set
```python
import os
os.environ["QT_QPA_PLATFORM"] = "offscreen"
```

pijyoi:
Okay, the binder link above does work now!

j9ac9k:
The LinearRegionItem behavior I experienced was only occurring in the Safari web browser, which makes me think this potentially a `jupyter_rfb` issue, not a pyqtgraph one.  Being able to replicate the issue in binder is fantastic.

j9ac9k:
I think the only thing this PR would need is to add a blurb to the README, and perhaps to the documentation somewhere, but given we don't reference jupyter anywhere in the docs, I'm not sure where the place to put it would be.

pijyoi:
Just a reminder, before any merge, I would need to point the binder configuration to fetch from pyqtgraph repo.

pijyoi:
Okay, I can replicate the ```ROIExamples.ipynb``` being truncated on binder also when running on WSL2 with ```QT_QPA_PLATFORM=offscreen```. So it probably has something to do with using ```offscreen```.

j9ac9k:
@pijyoi do you want to see if this is a PyQt5 vs. PySide6 difference or you think we should just merge?  FWIW this looks good to me.

pijyoi:
Okay, ready to merging.
1) Moved binder badge location
2) Point to pyqtgraph repo
3) Restore CI workflow

j9ac9k:
This LGTM, I would be interested in having a binder config that doesn't depend on conda (so we're not stuck w/ ancient versions of qt bindings); but given this works and a supported environment; let's leave it as is 👍🏻 ...I suppose notebook environments are more likely to use a conda distribution so probably for the best to replicate that environment.

pijyoi:
>do you want to see if this is a PyQt5 vs. PySide6 difference

I did some testing locally with WSL2. It's a Qt5 vs Qt6 difference.
As the provided container is Ubuntu 18.04, we cannot use Qt6.

j9ac9k:
> As the provided container is Ubuntu 18.04, we cannot use Qt6.

While I believe you, this really surprised me, and I double-checked the documentation, and sure enough, [Ubuntu 20.04 is the supported configuration](As the provided container is Ubuntu 18.04, we cannot use Qt6.)

PR #2061, opened 2021-11-08
re-add UTF-8 encoding to example app code loader

### LINKED ISSUE #2060: Example app fails to run examples on Windows
When running the example application on Windows, examples run when clicking on the `Run Example` button.
However, trying to run an edited example, or starting an example by double clicking the name fails with

```
  File "C:\e\prog\pg_dev\pyqtgraph\pyqtgraph\examples\ExampleApp.py", line 485, in loadFile
    proc.stdin.write(code)
TypeError: a bytes-like object is required, not 'str'
```

### Tested environment(s)
sys.platform: win32
pyqtgraph: 0.12.3; None

### Additional context
It seems this is the result of a change in #2045 :
In line 485 of exampleApp.py, the codeview text is now directly written to the stdin pipe. Previously UTF-8 encoding was specifically applied here. It seems this is still required, at least on (my specific version of) Windows.

NilsNemitz:
This re-adds the UTF-8 encoding used by the example app to run code from the editor.

Works on my machine  :) 
I think this just restores the previous code, but it would be great if someone can test on different architectures.

Closes #2060

### DIFF
--- a/pyqtgraph/examples/ExampleApp.py
+++ b/pyqtgraph/examples/ExampleApp.py
@@ -481,7 +481,7 @@ class ExampleLoader(QtWidgets.QMainWindow):
         env['PYTHONPATH'] = f'{path}'
         if edited:
             proc = subprocess.Popen([sys.executable, '-'], stdin=subprocess.PIPE, cwd=path, env=env)
-            code = self.ui.codeView.toPlainText()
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for the PR @NilsNemitz  Forgot that some windows platforms require this 😬  sorry for the trouble

PR #2064, opened 2021-11-08
ViewBox quantization limit

### LINKED ISSUE #2056: Plot does not do show constant values for value above 4.5e15 (roughly)
### Short description
If you run the example attached, you will see a constant value curve, if you change the value of `constant `variable from `4.5e15` to `4.6e15`, constant value curve is no longer shown. Presumably this is true for any value >= 4.6e16.

### Code to reproduce

```python
import numpy as np
import pyqtgraph as pg

app = pg.mkQApp()
mw = QtGui.QMainWindow()
cw = QtGui.QWidget()
mw.setCentralWidget(cw)
l = QtGui.QVBoxLayout()
cw.setLayout(l)

pw = pg.PlotWidget(name='Plot1') 
l.addWidget(pw)

mw.show()

## Create an empty plot curve to be filled later, set its pen
p1 = pw.plot()

constant = 4.5e15 # From 4.6 onward, it fails, that is line is not shown
pw.plot(y=[constant,]*2, x=list(range(2)))

if __name__ == '__main__':
    pg.exec()

```

### Expected behavior

Curve is shown for `constant = 4.6e15`
### Real behavior

See short description. No error occurs, but curve is not shown. Auto range issue ?

### Tested environment(s)

 * PyQtGraph version: 0.12.2
 * Qt Python binding: 
 * Python version: PyQt5 5.15.4 Qt 5.15.2
 * NumPy version: 1.20.1
 * Installation method: pip

NilsNemitz:
This is meant to address #2056, where the vertical axis collapses when a list of large, constant values is plotted.

The reason for this is that the default 1.0 span of the auto-scaling eventually becomes smaller than the effective (~3e-16) resolution of a double float, and the range calculations break down.

This PR sets the minimum displayed range to 6E-15 of the center value, which makes sure that at least 20 quantization steps are on screen. At this level, the errors become obvious, but the plot remains visible, and the axis manages to calculate ticks and labels.

For values < 1.6e+14, the behavior should be unchanged, *except* that there is now a hard lower limit on the plot range: It is no longer possible to deliberately collapse the axis range to zero. However, I could not come up with a good reason to do so.

Here is some code to play around with, based on the report in #2056:
```python
import pyqtgraph as pg

app = pg.mkQApp()
mw = QtGui.QMainWindow()
cw = QtGui.QWidget()
mw.setCentralWidget(cw)
l = QtGui.QVBoxLayout()
cw.setLayout(l)

pw = pg.PlotWidget(name='Plot1') 
l.addWidget(pw)
mw.show()
constant = 8e15 # now keeps working until other things break for >8e130 
p1 = pw.plot(y=[constant,constant])

# uncomment this to see quantization errors:
# p1 = pw.plot(y=[constant,constant+1,constant+2])

# uncomment this to see hard limit on axis range:
# pw.setYRange(constant-0, constant+0, padding=0)

if __name__ == '__main__':
    pg.exec()
```

Closes #2056, I hope.

### DIFF
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -575,7 +575,10 @@ class ViewBox(GraphicsWidget):
                 mn -= dy*0.5
                 mx += dy*0.5
             # Make sure that the range include a usable number of quantization steps:
-            quantization_limit = (mn+mx) * 0.5 * 3e-15 # ~10 discrete steps of double resolution
+            #    approx. eps  : 3e-16
+            #    * min. steps : 10
+            #    * mean value : (mn+mx)*0.5 
+            quantization_limit = (mn+mx) * 1.5e-15 # +/-10 discrete steps of double resolution
             if mx-mn < 2*quantization_limit:
                 mn -= quantization_limit
                 mx += quantization_limit
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -574,7 +574,7 @@ class ViewBox(GraphicsWidget):
                     dy = 1
                 mn -= dy*0.5
                 mx += dy*0.5
-            # Make sure that the range include a usable number of quantization steps:
+            # Make sure that the range includes a usable number of quantization steps:
             #    approx. eps  : 3e-16
             #    * min. steps : 10
             #    * mean value : (mn+mx)*0.5 
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -574,6 +574,11 @@ class ViewBox(GraphicsWidget):
                     dy = 1
                 mn -= dy*0.5
                 mx += dy*0.5
+            # Make sure that the range include a usable number of quantization steps:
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/ViewBox/ViewBox.py
```diff
@@ -574,6 +574,14 @@ def setRange(self, rect=None, xRange=None, yRange=None, padding=None, update=Tru
...
                 mn -= dy*0.5
                 mx += dy*0.5
+            # Make sure that the range includes a usable number of quantization steps:
+            #    approx. eps  : 3e-16
+            #    * min. steps : 10
+            #    * mean value : (mn+mx)*0.5 
+            quantization_limit = (mn+mx) * 1.5e-15 # +/-10 discrete steps of double resolution
+            if mx-mn < 2*quantization_limit:
+                mn -= quantization_limit
+                mx += quantization_limit
```
msmttchr, line 584: It looks good to me. Thanks. I just have a very minor point: regarding the comment `approx. eps`, stands for approximated double resolution ?

NilsNemitz: Yes, numpy seems to call it that. I think C spells it out as epsilon, typically.
https://numpy.org/doc/stable/reference/generated/numpy.finfo.html
According to that page: `eps = 2**-52`, approximately 2.22e-16.
But there is no real benefit to being ultra-accurate here; I am just trying to explain the magic cutoff value.

j9ac9k:
thanks for the PR @NilsNemitz, merging!

PR #2069, opened 2021-11-10
add HussariX to the list of Used By in README

sem-geologist:
This is currently in very early WIP, with only single feature/ type of files opening.

It heavily relies on `pyqtgraph` provided PlotWidget, PlotDataItem, Docking widget/window framework, InfiniteLine and some other widgets as ColorButton and more is coming in further implementation.
It can also showcase customization of AxisItem (zooming in one direction as there are no negative values in spectra).


j9ac9k:
Thank you!

PR #2073, opened 2021-11-12
implement glInfo without PyOpenGL

pijyoi:
Github Actions has moved to macOS Big Sur for some runners. This causes:
1) an early failure on Python 3.7
   * this is because the CI workflow has a dependency on ```PyOpenGL``` when it executes ```pyqtgraph.opengl.glInfo```.
2) OpenGL examples to be skipped on Python 3.8
   * however, the OpenGL load failure had been fixed since 3.8.10

This PR fixes (1) by reimplementing glInfo without using PyOpenGL. (2) is fixed by updating the skip constraints.
It adds some more useful info like VENDOR, RENDERER and isOpenGLES.

However, it loses the maximum TEXTURE size info.
This is because PySide bindings do not have a working glGetIntegerv function.

Note: The file was moved from ```pyqtgraph/opengl/glInfo.py``` to ```pyqtgraph/util/glinfo.py``` because ```pyqtgraph/opengl/__init__.py``` already triggers import of PyOpenGL.


j9ac9k:
Thanks for this @pijyoi this LGTM 👍🏻

pijyoi:
Finally did get `glGetIntegerv` working on the PySide{2,6} bindings. But I guess the TEXTURE size info isn't very important.

pijyoi:
PySide6 seems to be getting a working `glGetIntegerv` function.
https://code.qt.io/cgit/pyside/pyside-setup.git/commit/?id=9fb437cd2d017648f4bfc522b47f9488b825909e

Issue #997, opened 2019-07-17
Adding an AxisItem to a PlotItem causes an exception


Vieler:
### Code to reproduce

```python
import pyqtgraph as pg

pg.mkQApp()

p1 = pg.plot([1,2])

ax2 = pg.AxisItem("right")
p1.addItem(ax2)

pg.QAPP.exec_()
```

### Expected behavior
You should be able to add an `AxisItem` to a `PlotItem`
### Real behavior
```
Traceback (most recent call last):
  File "/home/vieler/issue.py", line 8, in <module>
    p1.addItem(ax2)
  File "/home/vieler/.virtualenvs/general/lib/python3.7/site-packages/pyqtgraph/graphicsItems/PlotItem/PlotItem.py", line 526, in addItem
    item.setLogMode(self.ctrl.logXCheck.isChecked(), self.ctrl.logYCheck.isChecked())
TypeError: setLogMode() takes 2 positional arguments but 3 were given
```

### Tested environment(s)

 * PyQtGraph version: 0.10.0
 * Qt Python binding: PyQt5 5.12.1 Qt 5.12.2
 * Python version: 3.7.3
 * NumPy version: 1.16.4
 * Operating system: Arch Linux x64
 * Installation method: pip

j9ac9k:
Can reproduce this from the current develop branch.

j9ac9k:
Ran this through the debugger, this is really weird, put the breakpoint right before the call, verified I'm getting booleans from the `self.ctrl.log[X|Y]Check.isChecked()` but it's still saying it's getting 3 positional arguments.

Then I called it myself with

```
(Pdb) item.setLogMode(False, False)
*** TypeError: setLogMode() takes 2 positional arguments but 3 were given
```

j9ac9k:
As soon as I typed that issue, I figured out the problem

`AxisItem.setLogMode` takes in arguments of `self` and `log`, whereas `PlotItem.setLogMode`, takes arguments of `self`, `x` and `y`.

ixjlyons:
[...]

PR #2081, opened 2021-11-13
Separate x and y flags for AxisItem.setLogMode

max-radin:
This fix allows the example code from #997 to be run. I also checked that the logAxis.py example looks the same after this change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,7 +68,7 @@ PyQtGraph developers are highly encouraged to (but not required) to use [`pre-co
 To run the test suite, after installing the above dependencies run
 
 ```bash
-python -m pytest examples tests
+python -m pytest pyqtgraph/examples tests
 ```
 
 ### Tox
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -205,20 +205,24 @@ class AxisItem(GraphicsWidget):
         self.prepareGeometryChange()
         self.update()
 
-    def setLogMode(self, log):
+    def setLogMode(self, x=None, y=None):
         """
-        If *log* is True, then ticks are displayed on a logarithmic scale and values
+        Set log scaling for x and/or y axes.
+        If an axis is set to log scale, ticks are displayed on a logarithmic scale and values
         are adjusted accordingly. (This is usually accessed by changing the log mode
         of a :func:`PlotItem <pyqtgraph.PlotItem.setLogMode>`)
         The linked ViewBox will be informed of the change.
         """
-        self.logMode = log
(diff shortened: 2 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -68,7 +68,7 @@ PyQtGraph developers are highly encouraged to (but not required) to use [`pre-co
 To run the test suite, after installing the above dependencies run
 
 ```bash
-python -m pytest examples tests
+python -m pytest pyqtgraph/examples tests
```
j9ac9k, line 71: nice catch!

NilsNemitz:
Hi @max-radin,

That is interesting, because I think the code that seems to break here is newer than the issue #997.
Ooops. Might have broken something there.

I am currently looking at the log-mode propagation from another angle... I currently think that it might be better to tell the ViewBox about the requested log mode settings from inside `PlotItem.updateLogMode` that should have the information for both axes. The AxisItems don't really have a need to know about both orientations.

Doesn't each of the four potential axes set both x and y mode individually, now?

NilsNemitz:
Ah, sorry, I missed the actual broken bit here.

While I am not sure if `setLogMode` is ever called directly, do you think this can work so that 
- two arguments are treated as `x` and `y`, and AxisItem.logMode is extracted from the correct one, like you have now
- a single argument directly controls `AxisItem.logMode` as before
?

max-radin:
Hi @NilsNemitz, one way we could support the old single-argument behavior would be to do something like
```python
def setLogMode(self, *args, **kwargs):
  if len(args) == 1 or 'log' in kwargs:
    # do the old thing
  else:
    # do the new thing
```
That way calling `setLogMode(True)` will control log mode along the direction of the axis while still allowing `setLogMode(x=True)` to control log mode in the x direction.

Pros:
* Backwards compatibility with any user code that called `AxisItem.logMode`
* Provides an easy way to set the log mode along an axis without having to check its orientation

Cons:
* The behavior of `AxisItem.logMode` will be different from `PlotItem.logMode`.
* The function signature and logic in `AxisItem.logMode` will be a bit more complicated.

Do you think this makes sense to do? I'm pretty new to the codebase so I don't really have a sense of how these methods typically get used.

NilsNemitz:
I think that is a good idea.
You probably do not need to check for the `log` keyword, that is probably not commonly used in a single-parameter function call.
So the simple distinction of one argument / two arguments might make it work.
Then it has the same functionality as other items when called by PlotItem (with two parameters).
And the old single parameter, possibly user-accessible interface still works to set log mode along the axis alignment.

max-radin:
@NilsNemitz just pushed a commit that adds backwards compatibility for the single argument case. thanks!

j9ac9k:
@max-radin thanks for the PR! This LGTM, merging.  @NilsNemitz thanks for reviewing!

Issue #1686, opened 2021-04-02
PlotItem.addLine() does not take logMode into account 


j9ac9k:
### Short description

If you have a PlotIem, and have log-mode enabled, and call `PlotItem.addLine()` the position you specify will not match up with the axis, instead you have to take the log of the value yourself before adding it.

### Code to reproduce

```python
import pyqtgraph as pg
import numpy as np
from math import log10

plotWidget = pg.plot()

points = 100
x = np.linspace(1, 100, points, endpoint=True)

plotWidget.plotItem.plot(x=x, y=np.exp(x), name="e^x")
plotWidget.plotItem.addLine(x=10)
# plotWidget.plotItem.addLine(x=log10(10))  # comment in for appropriate placement
plotWidget.plotItem.setLogMode(x=True, y=True)

pg.mkQApp().exec_()
```

### Expected behavior

The line should be added at the position x=10

### Real behavior

The line is added at the position x=10^10

```
An error occurred?
Post the full traceback inside these 'code fences'!
```

### Tested environment(s)

 * PyQtGraph version: 0.12.0 + master
 * Qt Python binding: All Bindings

### Additional context

![image](https://user-images.githubusercontent.com/646398/113383623-30700a00-9339-11eb-9bf9-54e99ffada5f.png)

smishy05:
Hey @j9ac9k ! Is this bug resolved? If not, can you please assign me this?

j9ac9k:
Hi @smishy05 

Have at it!  There is another issue that is very much related, #1684  While fixing that other issue is not necessary to merge a PR, figured I would bring it to your attention.

Thanks for tackling this!

smishy05:
Thanks a lot for the reply! 

Thanks a lot for the link to the other issue!

Yup, I will have a go at it!

Issue #1470, opened 2020-12-18
mapSceneToView return wrong mouse coordinate in log mode


edumur:
### Short description
The mouse coordinates return by `mapSceneToView` does not work when one of the axis is in log mode.

### Code to reproduce
```python
import numpy as np
from pyqtgraph.Qt import QtGui, QtCore
import pyqtgraph as pg
app = QtGui.QApplication([])
w = pg.GraphicsLayoutWidget(show=True)
p = w.addPlot()


p.setLogMode(True, False) # Line to comment or not to see the issue
y = np.random.normal(size=1000)
x = np.linspace(0, 1, 1000)
p.plot(x, y)
def mouseMoved(pos):
    print(p.vb.mapSceneToView(pos))
p.scene().sigMouseMoved.connect(mouseMoved)
w.show()
## Start Qt event loop unless running in interactive mode or using pyside.
if __name__ == '__main__':
    import sys
    if (sys.flags.interactive != 1) or not hasattr(QtCore, 'PYQT_VERSION'):
        QtGui.QApplication.instance().exec_()
```

### Expected behavior
The `mapSceneToView` function should return the coordinate hovered by the mouse.

### Real behavior
When one of the axis is in log mode the `mapSceneToView` function return incorrect mouse coordinates.

### Tested environment(s)

 * PyQtGraph version: 0.11.0
 * Qt Python binding: PyQt5 5.9.2 Qt 5.9.6
 * Python version: 3.7
 * NumPy version: 1.18.1
 * Operating system: windows 10
 * Installation method: pip

### Additional context

I would be happy to help but after some research I was not able to come up with the slightest idea on how to correct that issue.

j9ac9k:
Unfortunately #646 does not appear to resolve this issue :(

elisey:
I faced with the same problem.
I have this results:

```shell
2, instead of 1 * 10²
2.3, instead of 2 * 10²
3, instead of 2 * 10³
```

Workaround here is to raise 10 to a power of x, where x is your value.

10**2 == 1 * 10²
101**2.3 == 2 * 10²

edumur:
Hello,

@elisey: Thanks a lot! Following your post I implemented a workaround on your data viewer and it saved the day:
```python
        # We implement a workaround for the log mode of 1D plot.
        # See: https://github.com/pyqtgraph/pyqtgraph/issues/1470#issuecomment-864568004
        if self.plotType=='1d':
[...]

PR #2084, opened 2021-11-13
Fix log display of various plot items

### LINKED ISSUE #22: InfiniteLine issue when toggling logarithmic axes
Hello,

If I add a horizontal infinite line at let's say y=50, when I toggle the y scale to be logarithmic, the line appears at 10^50 instead of at y=50. Tee sample behaviour is observed with the x axis.

Should it be considered a bug or should I reposition myself the lines when changing from linear to logatirhmic scales?

Armando

NilsNemitz:
This PR makes InfiniteLine, LinearRegionItem and TargetItem functional in combination with log mode.

It moves away from hardcoded normal/log mode, towards a more flexible setup, where a new PlotDataMapping object holds the transform and the associated limits to possible values. For example, in log mode, a view space value of 307 correspond to 1E307 in data space. A that point, calculations start breaking down.

The current goal is to make the InfiniteLine example functional when switching one or both axes to log mode.

PlotDataItem still uses its hardcoded log functionality, but once this is updated (in a different PR), we will have support for arbitrary mapping functions, such as cubic root, or a signed log transform similar to what was the default for a while.

To do / help wanted list:
- Some functions still don't work when adding items with the plot already in log mapping mode, that means #1686 is not fixed yet.
- ~~LinearRegionItem still breaks autoscaling during change to logarithmic mapping~~
- angle behavior of infinite lines in mixed lin/log mode is strange
- input would be welcome on how to expose the (reverse) mapping transform to the user, which would address #1470

closes #22


Bluenix2:
Why is there a lot of uncommented code?

j9ac9k:
> Why is there a lot of uncommented code?

Not uncommon in draft/WIP PRs where you're doing a lot of experimentation ... when the PR is closer to be ready for merging, all of us nit-pickers come out of the woodwork; just ask the author 😆 

test suite appears to be in `test_ViewBoxZoom.py` and across all platforms, so should hopefully be reproducible locally.

j9ac9k:
Posted in the slack workspace, but figure I should post here for completeness/documentation.

Right now, the Infinite Line example works well, but on auto-range it shows _way_ too much to the right; this is due to the autoRange feature in ViewBox thinks that the ViewBox has a child in the value of ~70-80 in the x-axis when in log-mode.

As you identified, the LinearRegionItem is the culprit.  I was able to zero-in on the issue when stepping through the debugger.

In `Viewbox.childrenBounds` when it loops over each item, and when it gets to the LinearRegionItem, even after the log-x transformation, it thinks the linear region item has an x-range from 70 to 80.

The LinearRegionItem reports this value as in this case it returns the result of `LinearRegionItem.getRegion()`, which is made up of this function

```python
    def getRegion(self):
        """Return the values at the edges of the region."""

        r = (self.lines[0].value(), self.lines[1].value())
        if self.swapMode == 'sort':
            return (min(r), max(r))
        else:
            return r
```

The issue here is that we're grabbing the InfiniteLine `.value()` method, and not `.pos()` (and subsequently checking if we want the x or y point).  So question is, should we also change the result of `InfiniteLine.value()` when we undergo a log-transformation?  That, I'm not sure, I'll leave to you @NilsNemitz

sem-geologist:
Am I reading this right? InfiniteLine will be properly bent and draggable and dynamically reshaping curve in Log space if deg is not one of (0,90,180,270) ?

pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -205,6 +207,18 @@ def setGrid(self, grid):
...
+    def setMapping(self, mapping):
+        """ 
+        Sets data mapping function. Currently, only `plotDataMapping.LogMapping` 
+        is supported, others will be treated as indentity
+        """
+        self.picture = None
+        if isinstance(mapping, plotDataMappings.LogMapping):
+            self.logMode = True
+        else:
+            self.logMode = True
```
sem-geologist, line 219: should not it be False?

NilsNemitz: Errr... That would appear to be so.

NilsNemitz:
> Am I reading this right? InfiniteLine will be properly bent and draggable and dynamically reshaping curve in Log space if deg is not one of (0,90,180,270) ?

That was not *that* high on the list of priorities. :)
For the moment, it fails in a way that is somewhat usable, and it seems to make sense again when in log/log mode.

The infinite line is primarily defined by a single handle point, and *that* point survives the transformations, no matter what angle is set. If we figure out an actual use case for a helper line in semi-log space, we can try to figure out what to do in that case.

j9ac9k:
> Am I reading this right? InfiniteLine will be properly bent and draggable and dynamically reshaping curve in Log space if deg is not one of (0,90,180,270) ?

Ha, no, that would be nice, but nowhere near that yet, for that, we would need to convert our InfiniteLine item to a custom PlotCurveItem, ...as @NilsNemitz said, we would just be transforming the point the infinite line is defined at.

pyqtgraph/plotDataMappings.py
```diff
@@ -0,0 +1,182 @@
...
+def register(name, mapping):
+    """
+    Registers a new PlotDataMapping object
+    """
+    if not isinstance(mapping, PlotDataMapping):
+        raise TypeError('added object must subclass PlotDataMapping')
+    MAPPINGS[name] = mapping
+    
+
+class PlotDataMapping(object):
```
j9ac9k, line 25: nit-pick level 1000

no need to inherit `object` in modern python3 code, this line can just be `class PlotDataMapping`

pyqtgraph/graphicsItems/InfiniteLine.py
```diff
@@ -334,18 +327,7 @@ def vsValue(self):
             return self.vs_p[0]
         else:
             return self.vs_p
-        
 
-    ## broken in 4.7
```
j9ac9k, line 283: I actually think we should not only leave this code in, but we should un-comment it (and make it python3 compatible).  It was commented out to deal w/ a bug that I'm fairly sure no longer exists.  But we probably should investigate closer.

j9ac9k:
This might be a macOS issue; but look at the x-axis after I do the log transforms...those numbers get crowded, is it like that on windows?

image

EfremBraun:
I'm noticing a bug with this, which is that in some cases, setting the axes to log mode hides the InfiniteLine.

For example, run the following, and immediately go to Plot Options -> Transforms -> Log Y, and you'll see the InfiniteLine disappear. It reappears after transforming back to the linear scale.

Not sure why it happens in cases like this but not in the example case given with PyQtGraph.

```
import sys
from PySide6.QtWidgets import *
import pyqtgraph as pg

class Window(QMainWindow):

    def __init__(self):
        super().__init__()

        self.UiComponents()
        self.show()

    def UiComponents(self):

        # creating a plot window
        plot = pg.plot()
        inf1 = pg.InfiniteLine(movable=True, angle=90)
        inf1.setPos(5)
        plot.addItem(inf1)

        # setup layout 
        widget = QWidget()
        layout = QGridLayout()
        widget.setLayout(layout)
        layout.addWidget(plot, 0, 1)
        self.setCentralWidget(widget)

App = QApplication(sys.argv)
window = Window()
sys.exit(App.exec())
```

Issue #784, opened 2018-11-08
ImageItem.drawAt() method ignores "axisOrder" config option


honkomonk:
Symptom:
If you use the "draw feature" of a ImageItem while the axisOrder is set to "`row-major`" the image will be transposed, but the coordiates of the mouse pointer are not.

Example:
Take the default "GraphicsItems/ImageItem - draw" example under `pyqtgraph/examples/draw.py` and add the line `pg.setConfigOptions(imageAxisOrder='row-major')` after the imports. Drawing and mouse pointer are now out of sync.

Suggested fix:
In the method ´drawAt()´ replace the line:
`pos = [int(pos.x()), int(pos.y())]`
with
`pos = [int(pos.y()), int(pos.x())] if self.axisOrder == 'row-major' else [int(pos.x()), int(pos.y())]`

Issue #2071, opened 2021-11-11
Unexpected behavior with drawKernel + row-major image item


ntjess:
### Short description
ImageItem's drawKernel doesn't respect `imageAxisOrder`, leading to unexpected results when it is employed (unless there's another setting that I am missing?)

### Code to reproduce

```python
import pyqtgraph as pg
from pyqtgraph.Qt import QtCore, QtGui, QtWidgets
import numpy as np

# Comment out this line for expected behavior
pg.setConfigOptions(imageAxisOrder='row-major')
class VectorPaintImage(pg.ImageItem):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.setDrawKernel(np.full((3, 3), 255))
        self.setBorder('r')

class UserPainter(pg.PlotWidget):
    """
    Allows user to draw vector shapes by pressing and holding the mouse
    """
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.imgItem = VectorPaintImage()
        self.addItem(self.imgItem)
        self.getViewBox().invertY()
        self.setAspectLocked(True)
    
pg.mkQApp()
up = UserPainter()
up.setBackground('w')
up.imgItem.setImage(np.zeros((100, 100)))
up.show()
pg.exec()
```

### Expected behavior
drawing happens under the mouse

### Real behavior
x- and y-axes are swapped during drawing

https://user-images.githubusercontent.com/23620506/141327288-bec97855-88de-4170-bf6b-54268363bdec.mp4

### Tested environment(s)

 * PyQtGraph version: 0.12.3

PR #2085, opened 2021-11-14
reverse coordinates when drawing on row-major images

outofculture:
This addresses #2071

### DIFF
--- a/pyqtgraph/graphicsItems/ImageItem.py
+++ b/pyqtgraph/graphicsItems/ImageItem.py
@@ -1039,7 +1039,10 @@ class ImageItem(GraphicsObject):
         #print(ev.pressure())
 
     def drawAt(self, pos, ev=None):
-        pos = [int(pos.x()), int(pos.y())]
+        if self.axisOrder == "col-major":
+            pos = [int(pos.x()), int(pos.y())]
+        else:
+            pos = [int(pos.y()), int(pos.x())]
         dk = self.drawKernel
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/ImageItem.py
```diff
@@ -1039,7 +1039,10 @@ def tabletEvent(self, ev):
         #print(ev.pressure())
 
     def drawAt(self, pos, ev=None):
-        pos = [int(pos.x()), int(pos.y())]
+        if self.axisOrder == "col-major":
```
j9ac9k, line 1042: change looks fine to me, but do you know if there is a situation where it would be preferable to do 

```python
point  = pos.toPoint()
[point.x(), point.y()]
```

Or the better way of phrasing that question is is there a situation where QPointF -> `QPoint` conversion is what we would want instead of `int(QPointF.x())` ?

clearly this has worked well for us in the library, so not requesting a change, just something that crossed my mind when I looked at this.

outofculture: Ooo, interesting. The documentation at https://doc.qt.io/qt-5/qpointf.html#toPoint says it would round the values, whereas `int` is going to floor them. Rounding is better, but investigating, I also found problems. To be continued...

outofculture:
This also fixes #784

j9ac9k:
Thanks @outofculture Love it when the "simple" fix does the trick.

PR #2087, opened 2021-11-14
update imports in the broken example demos

Wubbzi:


j9ac9k:
Hi @Wubbzi 

Thanks for the PR! I know I suggested this fix on the discord, but after taking a closer look, I believe there is a better fix.  Specifically in this line:

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/examples/ExampleApp.py#L484

We should replace `cwd=path` to `cwd=example_path`.  

Would you mind undoing the changes in this PR, and modify that subprocess call instead?

j9ac9k:
Thank you @Wubbzi !

Issue #1702, opened 2021-04-09
AttributeError: 'NoneType' object has no attribute 'apoptose'


SchrodingersGat:
### Short description

When closing a Dock (programatically, not using the "close" button), throws the error as described in the title.

### Expected behavior

Dock window should close

### Real behavior

Throws AttributeError, dock does not close

```
  File "C:\...\Python\Python38-32\lib\site-packages\pyqtgraph\dockarea\Dock.py", line 236, in close
    self._container.apoptose()
AttributeError: 'NoneType' object has no attribute 'apoptose'
```

### Tested environment(s)

 * PyQtGraph version: 0.12.1
 * Python version: 3.6
 * NumPy version: 1.19.4
 * Operating system: Windows 10
 * Installation method: PIP

### Additional context

`self._container` is None, but there are no checks for a null container in `Dock.close()`

Also, in `Dock.__init__`, there are two separate lines `self._container = None` (might be worth deleting one of these)

j9ac9k:
I've run into this very issue when trying to close an already closed Dock (which admittedly, should not raise an exception either).  I am curious how this error occurred if the dock was already open.  Was there anything else odd about it; like ...was there no widget within it?

SchrodingersGat:
There are two pg.PlotWidget attached to the dock

j9ac9k:
How do you have two plot widgets?  The Dock can only have one widget at most 1 widget at a time.  Do you mean you have a GraphicsWidgetLayout with two plots in it?

SchrodingersGat:
Interestingly, this behaviour does not exhibit if there is only a single widget attached to the dock..

With a single widget, the dock closes correctly when .close() is called programatically

max-radin:
Has anyone been able to reproduce this error? The example below worked ok for me both on master and pyqtgraph-0.12.1.

```python
import numpy as np
import pyqtgraph as pg
from pyqtgraph.dockarea.Dock import Dock
from pyqtgraph.dockarea.DockArea import DockArea
from pyqtgraph.Qt import QtWidgets

app = pg.mkQApp("DockArea Example")
win = QtWidgets.QMainWindow()
area = DockArea()
win.setCentralWidget(area)
win.resize(1000,500)
win.setWindowTitle('dock with two plots')


d1 = Dock("Dock1", size=(1, 1))
d2 = Dock("Dock5", size=(500,200))
area.addDock(d1, 'left')
area.addDock(d2, 'right', d1)

[...]

PR #2089, opened 2021-11-14
Don't raise exception when close method of an already-closed dock is called

max-radin:
See #1702

### DIFF
--- a/pyqtgraph/dockarea/Dock.py
+++ b/pyqtgraph/dockarea/Dock.py
@@ -227,7 +227,7 @@ class Dock(QtWidgets.QWidget, DockDrop):
     def close(self):
         """Remove this dock from the DockArea it lives inside."""
         if self._container is None:
-            warnings.warn(f"Cannot close dock {self} because it is not open.")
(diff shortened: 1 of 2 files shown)


pyqtgraph/dockarea/Dock.py
```diff
@@ -227,6 +226,10 @@ def raiseDock(self):
 
     def close(self):
         """Remove this dock from the DockArea it lives inside."""
+        if self._container is None:
+            warnings.warn(f"Cannot close dock {self} because it is not open.")
```
j9ac9k, line 230: nice touch on the warning! I think you need to add `stacklevel=2` to capture the correct context

```python
warnings.warn(f"Cannot close dock {self} because it is not open.", RuntimeWarning, stacklevel=2)
```

max-radin: good point! fixed in 53fe2b38dd994dfcd7d64f9b5b5848f016bc461b

j9ac9k:
Thanks @max-radin merging!

PR #2090, opened 2021-11-14
more convenience methods for color maps and bars

### LINKED ISSUE #2067: Enhancement: allow string or colormap to be passed to ImageItem directly
I find matplotlib's ability to accept a string `cmap` to be extremely convenient during plotting, i.e. `plt.imshow(data, cmap='red')`. It would be fantastic to enable the same functionality with pyqtgraph's `pg.colormap`, i.e. `item.setImage(data, lut='viridis')` or `lut=pg.colormap.get('viridis')`.

The current workaround is `lut=pg.colormap.get('viridis').getLookupTable(nPts=X)`. The trick with a str lut is how to determine X (i.e. during autolevels, should nPts always be set to image range?), what should happen if the image has negative values, and probably a host of options I'm not even considering. But it would be quite convenient :slightly_smiling_face:

NilsNemitz:
This is meant to address #2067.

The `ImageItem` false color options were initially completely designed around control through `HistogramLUTItem`.
This adds a `setColorMap()` method to ImageItem, and the option to use this through a `colorMap` argument during initialization, `setImage()` or `setOpts()`.

The PR also adds the option to create and add a `ColorBarItem` through a new `addColorBar()` method of ImageItem.
This is somewhat inelegant, since the user needs to manually pass a reference to the plot. I could not find a means for the ImageItem to determine its parent. An alternative would have been to include `addColorBar()` as a method of the `PlotItem`, but this results in a circular import.

I have updated the ColorBarItem example to use the new methods for demonstration. The code to add an image with color map and interactive color bar is now down to just
```python
i1 = pg.ImageItem(image=data)
p1.addItem( i1 )
p1.addColorBar( i1, colorMap='CET-L9', values=(0, 30_000))
```
Might close #2067 ...

### DIFF
--- a/pyqtgraph/graphicsItems/ImageItem.py
+++ b/pyqtgraph/graphicsItems/ImageItem.py
@@ -139,9 +139,9 @@ class ImageItem(GraphicsObject):
         Parameters
         ----------
             levels: list_like
-                - [`blackLevel`, `whiteLevel`] 
+                - ``[blackLevel, whiteLevel]`` 
                   sets black and white levels for monochrome data and can be used with a lookup table.
-                - [[`minR`, `maxR`], [`minG`, `maxG`], [`minB`, `maxB`]] 
+                - ``[[minR, maxR], [minG, maxG], [minB, maxB]]``
                   sets individual scaling for RGB values. Not compatible with lookup tables.
             update: bool, optional
                 Controls if image immediately updates to reflect the new levels.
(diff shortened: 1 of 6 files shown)


j9ac9k:
That cyclic import is going to be tough to work-around, unless you move the import of `ColorBarItem` into the `addColorBar` method; which, while ugly, might not be the worst idea.

pyqtgraph/graphicsItems/ImageItem.py
```diff
@@ -161,6 +163,44 @@ def getLevels(self):
...
+            raise TypeError("'colorMap' argument must be ColorMap or string")
+        self.setLookupTable( self._colorMap.getLookupTable(nPts=256) )
+        
+    def getColorMap(self):
+        """
+        Returns the assigned color map, or None if not available
+        """
+        return self._colorMap
+        
+    # def addColorBar(self, colorMap=None, plot=None, **kargs):
```
j9ac9k, line 190: Looks like just need to delete this commented out code and this is probably good to merge? I should take this diff for a spin myself first I suppose.

j9ac9k:
Thanks for all the documentation tidying!  This LGTM, merging.

PR #2092, opened 2021-11-15
Add some helpful exceptions for invalid inputs to methods

ksunden:
Raise ValueError/TypeError instead of relying on NameError being raised later on in less helpful contexts to avoid uninitialized variables.


j9ac9k:
Love fixing static code error issues!  Thanks @ksunden for the PR!

PR #2094, opened 2021-11-15
Handle/remove unused variables

ksunden:
All found via CodeQL

### DIFF
--- a/pyqtgraph/flowchart/library/functions.py
+++ b/pyqtgraph/flowchart/library/functions.py
@@ -327,7 +327,6 @@ def removePeriodic(data, f0=60.0, dt=None, harmonics=10, samples=4):
     
     ## determine frequencies in fft data
     df = 1.0 / (len(data1) * dt)
-    freqs = np.linspace(0.0, (len(ft)-1) * df, len(ft))
     
     ## flatten spikes at f0 and harmonics
     for i in range(1, harmonics + 2):
--- a/pyqtgraph/flowchart/library/functions.py
+++ b/pyqtgraph/flowchart/library/functions.py
@@ -325,7 +325,7 @@ def removePeriodic(data, f0=60.0, dt=None, harmonics=10, samples=4):
     
     ft = np.fft.fft(data1)
     
-    ## determine frequencies in fft data
+    ## determine frequency step in fft data
     df = 1.0 / (len(data1) * dt)
     
     ## flatten spikes at f0 and harmonics
--- a/pyqtgraph/graphicsItems/LinearRegionItem.py
+++ b/pyqtgraph/graphicsItems/LinearRegionItem.py
@@ -1,3 +1,4 @@
+from .. import debug
 from .. import functions as fn
 from ..Qt import QtCore, QtGui
 from .GraphicsObject import GraphicsObject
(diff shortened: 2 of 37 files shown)


ksunden:
Test failure caused by something that isn't actually new, and is addressed in #2095

pyqtgraph/flowchart/library/functions.py
```diff
@@ -327,7 +327,6 @@ def removePeriodic(data, f0=60.0, dt=None, harmonics=10, samples=4):
     
     ## determine frequencies in fft data
     df = 1.0 / (len(data1) * dt)
-    freqs = np.linspace(0.0, (len(ft)-1) * df, len(ft))
```
NilsNemitz, line 330: This might be good to keep as a comment, assigning the correct frequencies in FFT is always a pain.

ksunden: I personally hate leaving commented code committed to the repo... we have a lot in this library as it is (many of which have python2 style prints)

[np.fft.fftfreq](https://numpy.org/doc/stable/reference/generated/numpy.fft.fftfreq.html) exists for a reason

pyqtgraph/graphicsItems/LinearRegionItem.py
```diff
@@ -269,7 +268,6 @@ def boundingRect(self):
         return br
         
     def paint(self, p, *args):
-        profiler = debug.Profiler()
```
NilsNemitz, line 272: Don't these profiler assignments actually do something?

pyqtgraph/flowchart/library/functions.py
NilsNemitz: Fair enough. The following code is also more readable than I was expecting.
Change L 328 to 
`## determine frequency step in fft data`
maybe? You are removing the list of frequencies that the comment now refers to.

pyqtgraph/graphicsItems/LinearRegionItem.py
ksunden: as far as I'm aware, these are debugging tools that probably shouldn't be scattered in production code, no?

j9ac9k: So, I'm inclined to keep them; they're a signal that this is a _very_ performance sensitive function, and when implementing new features, it's trivial to re-enable them so they spew output in the console (either via setting environment variables, or modifying the line to add `disabled=False` and `delayed=False`).

That said; they're not crucial on their own, when profiling/debugging, the developer can re-add them as needed, and if we're really wanting to do profiling, we probably shouldn't be dependent on these buy use a tool that can capture runtimes and average them out sort of thing.

j9ac9k: What might be a fair compromise here is to remove the assignment for cases where profilers are not used throughout the function/method they're in such as this case.  So instead of 

```python
profiler = debug.Profiler()
```

we can have

```python
debug.profiler()
```

This serves the same thing and should make the static code checker happy.

ksunden:
I believe this is ready for review

j9ac9k:
Should have merged this a while back, sorry about that @ksunden LGTM, thanks!

PR #2095, opened 2021-11-15
avoid redundant assignment of pos=pos

ksunden:
While I'm at it, I got rid of the equality comparisons to None literal


j9ac9k:
LGTM!  Thank you @ksunden

PR #2096, opened 2021-11-15
Be lazier about importing h5py

ksunden:
We discovered recently bad interactions between pytables and h5py resulting in the inability to import one after the other had been imported. This takes a step towards pyqtgraph not being a confounding variable in that situation when it doesn't need to be. Note that h5py is still imported in MetaArray, so until that is removed, this doesn't actually solve the issue of h5py being loaded when pyqtgraph is imported, but once metaarray is removed as planned, this is the only remaing library (non-example) code that uses h5py directly.

### DIFF
--- a/pyqtgraph/exporters/HDF5Exporter.py
+++ b/pyqtgraph/exporters/HDF5Exporter.py
@@ -1,3 +1,5 @@
+import importlib
+
 import numpy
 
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks @ksunden !  Merging!

PR #2097, opened 2021-11-15
Add a proxy delay to checklist parameter changes via children edits

ntjess:
Randomly testing with a touchpad I found 1s proxy delay was usually adequate. Input from others would be nice on this

https://user-images.githubusercontent.com/23620506/141720037-ab4d8c2f-80d5-4cd8-8751-8f98d662f097.mp4

.

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes/checklist.py
+++ b/pyqtgraph/parametertree/parameterTypes/checklist.py
@@ -1,5 +1,5 @@
 from ... import functions as fn
-from ...Qt import QtWidgets, QtCore
+from ...Qt import QtWidgets
 from ...SignalProxy import SignalProxy
 from ..ParameterItem import ParameterItem
 from . import BoolParameterItem, SimpleParameter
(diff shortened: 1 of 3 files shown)


ntjess:
This is similar to how the SpinBox parameter (int/float) doesn't emit value changes until after the delay is reached. I just wasn't sure about hard-coding 1s delay, so I just made it an option.

j9ac9k:
Hi @ntjess I haven't worked w/ this much, but what is the issue that this solves?

ntjess:
Same issue that is solved in spinbox. It lets individual check boxes act as sigValueChanging while valueChanged only fires when the user is done editing

j9ac9k:
Oh, so in this case, when you're checking the checkboxes, instead of running the function that's being called on each time it's checked _as_ its checked, it waits a second to see if a user checks anything else, and then runs the slot that's connected to the `stateChanged` signal (in the case of the checkboxes)?

ntjess:
Yep! Just like changing vs changed on a spinbox (waiting until the spinner is motionless for awhile)

pyqtgraph/parametertree/parameterTypes/checklist.py
```diff
@@ -1,5 +1,6 @@
 from ... import functions as fn
-from ...Qt import QtWidgets
+from ...Qt import QtWidgets, QtCore
```
j9ac9k, line 2: Can we get rid of the QtCore import here?

j9ac9k:
LGTM, thanks @ntjess for the PR!

ntjess:
Looking at https://github.com/pyqtgraph/pyqtgraph/pull/2086#issuecomment-972824641, should this logic be changed to "when no child is focused" rather than "after 1 second"?

j9ac9k:
I can see a case for being made for if lose focus, emit signal right away, else wait for the delay period.

PR #2098, opened 2021-11-15
Remove polluting imports using *

ksunden:
Define `__all__` on additional files


pyqtgraph/GraphicsScene/GraphicsScene.py
```diff
@@ -6,7 +6,7 @@
 from .. import getConfigOption
 from ..Point import Point
 from ..Qt import QT_LIB, QtCore, QtGui, QtWidgets, isQObjectAlive
-from .mouseEvents import *
+from .mouseEvents import HoverEvent, MouseClickEvent, MouseDragEvent
```
j9ac9k, line 9: Since we define `__all__` in this case, can we undo this change?

j9ac9k:
Thanks @ksunden LGTM, merging.

PR #2099, opened 2021-11-15
Convert == None checks to is None

ksunden:


j9ac9k:
There are a few places in the repo we have `!= None` which would probably be appropriate to grab.

I see two instances in `ROI.py`, `ParameterItem.py` (granted that one is part of a string in an exception) and `test_dock.py`

ksunden:
The ROI ones were addressed by #2095

ksunden:
I do not care about the exception one, I think its current state is readable and gets the point across, and if anything changing it would be _less_ readable in this case.

j9ac9k:
That's legit, LGTM, another PR can capture the `==/!= True/False` bits   Merging this, thanks @ksunden !

PR #2100, opened 2021-11-15
Avoid reusing variables in nested loops

ksunden:


j9ac9k:
LGTM, thanks @ksunden !

Issue #1958, opened 2021-08-07
ParameterTree GroupItem Style Doesn't Agree With DarkMode


j9ac9k:
While looking at #1952 I noticed that the group parameters didn't look good with dark mode.

![image](https://user-images.githubusercontent.com/646398/128609421-37143816-a602-4e85-aa36-fdd0629b348d.png)

After looking around a little bit, sure enough, there is code that manually sets the background and foreground color.  This definitely is an indication that things will not work as intended.  The code should be modified to grab colors  from the [application palette ](http://doc.qt.io/qt-5/qapplication.html#palette) and modify those accordingly instead of hard-coding values.

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/parametertree/parameterTypes/basetypes.py#L319-L338

To test/replicate this issue any version of pyqtgraph/numpy/qt will work, but you need to be on macOS with dark mode enabled.

PR #2101, opened 2021-11-15
change GroupParameterItem palette to address issue in darkmode on mac

Wubbzi:
This partially addresses #1958 

These changes will still need to be tested on windows/Linux.

Group title colors were changed to the color of QPalette.TextRole but it could easily be changed back to the original in dark mode.

Light
unknown-1
Dark
unknown-2


j9ac9k:
A fellow macOS user has entered the chat!

Thanks for tracking this down @Wubbzi 

Over the next few days I'll try and test out this change on the other platforms.

sem-geologist:
I had tested this on Linux KDE Neon with breeze (light) and breeze dark themes (runing example with systems python3):
Parameter_tree_kde_neon_light

Parameter_tree_kde_neon_breeze_dark


Then also tested with conda (which uses a bit different theminng, more compact, quite similar to KDE system's default breeze, but there are some differences. It is also one of these possibilities to force dark theme on windows using conda's "dark_theme":
Parameter_tree_kde_neon_conda_light
Parameter_tree_kde_neon_conda_dark

everything looks good and readable.

sem-geologist:
One thing, which is bothering me is why on Your Mac at light theme the font at 0 level gets dark dark, and on my linux KDE neon and conda it is not getting dark at all. It is still readable, but the font color does not kick in, Your code suppose to change that but it does not. I tested if app.property("darkMode") is working, it is. As far as I recall from my other issues (with my software) somehow the font color in my cases could not be directly changed. Changing Font color worked through stylesheets only. Probably it is Qt for linux thing... I am surprised that set color is ignored, but Qt still manages to set color which is enough contrasting.

j9ac9k:
@sem-geologist macOS does some shenanigans here, and manipulates colors that it _thinks_ the application has specified to be inline with system defaults, to make them appear to match the rest of the system theme; and I believe that's what's happening with some of the transparent backgrounds for some of the widgets (which is why we see labels written twice for example).

Annoyingly, this means you need to be on macOS to replicate the issue.

@Wubbzi thanks for the PR, I'll hop on your branch and do some testing on my machine ...it might take me a day or two.

Wubbzi:
@sem-geologist Are you sure you're running the code from the PR? The pale lavender color in the outermost groups was part of the code that was replaced. I'd assume that you should only see black or white, seeing as how the text color should be using the TextRole from the application palette. 

Setting the text color to the TextRole is redundant, I only defined the color and left the code for `setForeground` in case someone wanted to play around with using the previous color. If you were to comment out the `setForeground` line, you shouldn't see a difference, since we're applying the TextRole color to text and the app palette is deciding which color TextRole color is.

The darkMode property is pretty nifty, and its actually part of pg. You can find the code in `pyqtgraph/Qt/__init__.py` on line 422.

sem-geologist:
oh... I see... I am an idiot - had cloned Your tree, and had not changed the branch 🤦.
Here are the real view, and colors looks same as on Your Mac pictures (I am lazy and not trying conda, just breeze and breeze dark with system python3):
Parameter_tree_kde_neon_light
Parameter_tree_kde_neon_breeze_dark

I am leaving the wrong pictures uploaded above (will comment that are old before this PR), as now I see clear difference. These changes are not only still readable, but looks much better IMHO.

Wubbzi:
@sem-geologist Thank you for taking the time to check - I was beginning to worry that macOS had gotten the best of me(again).

ntjess:
> @sem-geologist macOS does some shenanigans here, and manipulates colors that it _thinks_ the application has specified to be inline with system defaults, to make them appear to match the rest of the system theme; and I believe that's what's happening with some of the transparent backgrounds for some of the widgets (which is why we see labels written twice for example).
> 
> Annoyingly, this means you need to be on macOS to replicate the issue.
> 
> @Wubbzi thanks for the PR, I'll hop on your branch and do some testing on my machine ...it might take me a day or two.

This problem also shows on non-mac platforms with really long action names. The label will 'leak' out from underneath the button. I was able to solve this in my application with `self.setText(0, '')` in the ActionParameterItem constructor, but it would be good to make sure this is also robust against i.e. title changes and so on.

I was trying to find a reference to someone's message in the discord sprint chat but I'm not sure how to search the archived chat

j9ac9k:
> I was trying to find a reference to someone's message in the discord sprint chat but I'm not sure how to search the archived chat

I'll ask the mods there if we can get the archive.

ntjess:
Here was the message I remembered seeing:

> So, using a palette didn't seem to work and I'm not sure why. I was able to change the color of the button text, but not the background.
> Setting a stylesheet would work, if you want to go that route.
> If we change the __init__ in ParameterItem, that also seems to do the trick. 
> ```python
> # QtWidgets.QTreeWidgetItem.__init__(self, [param.title(), ''])      
> QtWidgets.QTreeWidgetItem.__init__(self, ['', ''])
> ```

I think this change would work better just in the override of widgets that don't have a name column, rather than at the ParameterItem level. So I might still recommend the `self.setText` approach vs. the `ParameterItem` change, but I'm not sure if it matters much either way

ntjess:
image


It's not working on qdarkstyle.load_stylesheet (a use case for several users, there's a [dedicated pyqtgraph section](https://github.com/ColinDuquesnoy/QDarkStyleSheet#usage-in-applications) in the repo. Is that something that can be changed in the palette here, or something qdarkstyle should alter?

image

j9ac9k:
ooooo nice call on checking against qdarkstyle; we may need to get some input from them on how to "best" go about this.

j9ac9k:
@ntjess wait... does qdarkstyle render this PR differently from master?  If not, there is likely nothing that we can do on our end.

Wubbzi:
@ntjess @j9ac9k 

I didn't even think about qdarkstyle but I added a condition to check for a stylesheet and apply a nicer looking palette in the case one exists.

Screen Shot 63

Wubbzi:
> image
> 
> It's not working on qdarkstyle.load_stylesheet (a use case for several users, there's a [dedicated pyqtgraph section](https://github.com/ColinDuquesnoy/QDarkStyleSheet#usage-in-applications) in the repo. Is that something that can be changed in the palette here, or something qdarkstyle should alter?
> 
> image

Which OS/Qt version are you using in the dark screenshot for this PR? I'm not sure why the `Example Parameters` group is completely white. Did you notice if `Save/Restore functionality` and the other groups where depth=0 were like this also?

NilsNemitz:
I seem to remember that qdarkstyle just restyles everything by stylesheet, and does not even try to touch the QPalette.
Is there a working way to set/detect dark mode on Windows now?
Would it be useful to force the system into the dark mode palette when qdarkstyle is detected?

ntjess:
It looks like qdarkstyle indeed is limited to the stylesheet with no palette integrations: https://github.com/ColinDuquesnoy/QDarkStyleSheet/blob/26c3878fff742c535058bf5d08643d0135cde2af/qdarkstyle/__init__.py#L275

@Wubbzi I'm not sure about how to load official dark themes, so I used an unofficial gist here: https://gist.github.com/lschmierer/443b8e21ad93e2a2d7eb

Running on Zorin OS 16 with PyQt5 5.15.2. List of recognized styles just shows "Fusion" and "Windows".

Thus, I'm not confident if my representation is truly indicative of a dark style.

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -319,24 +319,37 @@ def __init__(self, param, depth):
...
+        This allows highest-level groups to be displayed more prominently.
+        """
+        app = QtWidgets.QApplication.instance()
+        palette = app.palette()
+        textColor = palette.text().color()
+        backgroundColor = palette.button().color().darker(110)
+        altBackgroundColor = palette.dark().color()
+
+        # if stylesheet and 'QDarkStyleSheet' in stylesheet:
+        if app.styleSheet(): # Assume the stylesheet is qdarkstyle
```
ntjess, line 333: Thanks for checking for a stylesheet --I recommend looking for "qdarkstyle" explicitly in the sheet, rather than just checking it is nonzero in length. The header of qdarkstyle's sheet is

```
/* ---------------------------------------------------------------------------
    WARNING! File created programmatically. All changes made in this file will be lost!
    Created by the qtsass compiler v0.3.0
    The definitions are in the "qdarkstyle.qss._styles.scss" module
--------------------------------------------------------------------------- */
/* Light Style - QDarkStyleSheet ------------------------------------------ */
/*
See Qt documentation:
```

So there are a few magic strings you can check for

j9ac9k:
Is there a way to extract the color code programmatically from the style
sheet that's applied?

NilsNemitz:
Sorry if it isn't helpful, but I've been running a QPalette like this for darkmode on windows. It isn't perfect yet, but it kind of matches qdarkstyle in colors.

Would it make sense to apply that manually if a desire for dark mode is detected on Windows?

``` python
def dark_QPalette():
    # color definitions roughly match QDarkstyle
    BLACK      = QColor('#000000')
    BG_LIGHT   = QColor('#505F69')
    BG_NORMAL  = QColor('#32414B')
    BG_DARK    = QColor('#19232D')
    FG_LIGHT   = QColor('#F0F0F0')
    FG_NORMAL  = QColor('#AAAAAA')
    FG_DARK    = QColor('#787878')
    SEL_LIGHT  = QColor('#148CD2')
    SEL_NORMAL = QColor('#1464A0')
    # SEL_DARK   = QColor('#14506E')
    qpal = QPalette( QColor(BG_DARK) )
    for ptype in (  QPalette.Active,  QPalette.Inactive ):
        qpal.setColor( ptype, QPalette.Window    , QColor( COLORS['DS_bg_normal'] ) )
        qpal.setColor( ptype, QPalette.WindowText, QColor( COLORS['DS_fg_light' ] ) )
        qpal.setColor( ptype, QPalette.Base      , QColor( COLORS['DS_bg_dark'  ] ) )
        qpal.setColor( ptype, QPalette.Text, FG_LIGHT )
        qpal.setColor( ptype, QPalette.AlternateBase, BG_DARK )
        qpal.setColor( ptype, QPalette.ToolTipBase, BG_LIGHT )
        qpal.setColor( ptype, QPalette.ToolTipText, FG_LIGHT )
        qpal.setColor( ptype, QPalette.Button, BG_DARK )
        qpal.setColor( ptype, QPalette.ButtonText, FG_LIGHT )
        qpal.setColor( ptype, QPalette.Link, SEL_NORMAL )
        qpal.setColor( ptype, QPalette.LinkVisited, FG_NORMAL )
        qpal.setColor( ptype, QPalette.Highlight, SEL_LIGHT )
        qpal.setColor( ptype, QPalette.HighlightedText, BLACK )
    qpal.setColor( QPalette.Disabled, QPalette.Button, BG_NORMAL )
    qpal.setColor( QPalette.Disabled, QPalette.ButtonText, FG_DARK )
    qpal.setColor( QPalette.Disabled, QPalette.WindowText, FG_DARK )
    return qpal
```

ntjess:
@NilsNemitz Your solution is more robust than that gist -- thanks. Here is the new screenshot, everything appears to be in order

image

pyqtgraph/parametertree/parameterTypes/basetypes.py
ntjess: With load_stylesheet(palette=LightPalette), it looks a bit wonky:
image

Note that "DarkStyle" is commented out in this case as opposed to Light Style

Wubbzi:
> Is there a way to extract the color code programmatically from the style sheet that's applied?
> […](#)

We could extract the entire style sheet with `app.styleSheet()`. Since the styleSheet would be applied to the application and not specific widgets, we would probably need to use regex or manually parse the app's style sheet to pull out the colors for specific widgets.

If we changed the style sheet for any widget we would have to replace everything(margin,border,etc) not just the colors. 

I do have some old code that was used to do something very similar that could be adapted if needed. 

@NilsNemitz I can't test this right now, but I think we could use the current method for windows although it might need to be tweaked.
```py
def onPaletteChange(palette):
    color = palette.base().color().name()
    app = QtWidgets.QApplication.instance()
    app.setProperty('darkMode', color.lower() != "#ffffff")
```

@ntjess
Ahh, okay, I guess the palette explains why the colors are a bit off. 

Originally i was checking for 'QDarkStyleSheet' at the top of the style sheet but I wasn't sure how many groups one of these trees might have in the real world and I was concerned about slowing down the startup with all of the `in` checks so I changed it back. I'm happy to change it though.

NilsNemitz:
Does `onPaletteChange` ever happen on Windows, though?
The current code seems to check explicitly for white. Maybe check for lightness > 0.5 instead? Somebody might set their palette for something a little off-white in the future.

Wubbzi:
@NilsNemitz 

Really glad you asked about this! When I was looking for the answer I realized `qApp.paletteChanged` is emitted any time [mkQApp](https://github.com/pyqtgraph/pyqtgraph/blob/f1c62c0a4b346be540be9b794440be5b95a19898/pyqtgraph/Qt/__init__.py#L419https://github.com/pyqtgraph/pyqtgraph/blob/f1c62c0a4b346be540be9b794440be5b95a19898/pyqtgraph/Qt/__init__.py#L419) is used to create or access the running instance. 

This means we need to run `onPalletteChanged()`(or the code from within the function) before colors are chosen. If we rely on that function running 100% of the time, the apps that create a `QApplication()` instead of using `mkQApp` won't have pretty GroupParameterItems.

The `darkMode` property only seems to be used in this PR, the example app and and the syntax example, so its at least worth testing if Windows does allow changing the base color like this, as it won't break much if there is some edge case where it doesn't work.

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: image

Hows this? I feel like the text could be better on the outermost groups.. maybe not quite so dark or something with color.

sem-geologist:
This (opens pop-corns slowly) is fun to watch :) . Jokes aside, I think most of well thought themes use no absolute white or absolute black. i.e. Breeze Dark (which I won't hide like a lot) of KDE5: https://develop.kde.org/hig/style/color/dark/
look the brightest color (under selection section) called "paper white" for `ForegroundActive` is still not 6 f's, but a bit a tiny bit toned down white (`#fcfcfc`), where normal foreground is tiny-winy blued toned-down white `#eff0f1`.
I like how Qt applications looks on the KDE5. (light breeze looks also good, but I am quite a caveman, and sit with dark themes on). Using `.lighten()` instead of using OS colors. I wonder how much those are exposed to Qt, I guess for Qt on KDE5 it is alot,  where problem could be with other OS. While it looks good, It seem to force colors a bit out from theme. I think if those theme colors are not passed to Qt during intiation, if global Qt settings could be found, it would be nice as third option (or rather it should be checked at very first) to use global Qt theme for foreground, background, altBackground and so on instead of creating the colors on the fly. This PR then goes a bit out from scope and is going to include other OS and Qt theming.

j9ac9k:
> We could extract the entire style sheet with `app.styleSheet()`. Since the styleSheet would be applied to the application and not specific widgets, we would probably need to use regex or manually parse the app's style sheet to pull out the colors for specific widgets.

That's probably a bag of pain we want no part of.

I think going w/ hardcoded palette values that @NilsNemitz detailed above for this PR is the way to go.  It seems to look good in the most configurations, which is an improvement over the current implementation.  That meets the criteria for merging IMO.

I certainly didn't intend for the scope of this to blow up so much, and if we want to revisit proper detection of themes on KDE/Windows/macOS, we can certainly do so, but I don't think this PR is the place to do it.

Wubbzi:
> I think going w/ hardcoded palette values that @NilsNemitz detailed above for this PR is the way to go. It seems to look good in the most configurations, which is an improvement over the current implementation. That meets the criteria for merging IMO.

Sounds like a plan. 

I'm thinking that we can check for a qdarkstyle stylesheet(light or dark), if it exists, let the stylesheet take over, if there is no qdarkstyle stylesheet and the darkMode property is set, we apply the palette above to the ParameterTree.

Should I keep the lightMode colors from the initial commit in cases where there is no stylesheet and no darkMode?

I'll wait until tomorrow afternoon to commit the changes in case there are any objections/suggestions.

j9ac9k:
> I'm thinking that we can check for a qdarkstyle stylesheet(light or dark), if it exists, let the stylesheet take over, if there is no qdarkstyle stylesheet and the darkMode property is set, we apply the palette above to the ParameterTree.

If this check can be done w/o difficulty/complexity; I think that's fine... if users are specifying their own style sheet, they know what they want and would probably serve us to respect that.

> Should I keep the lightMode colors from the initial commit in cases where there is no stylesheet and no darkMode?

While those colors definitely look a bit dark for a light mode theme, those colors are definitely intended to be darker than  normal, so I personally think they're fine for their intended purpose.

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -319,24 +319,37 @@ def __init__(self, param, depth):
...
-                font.setBold(True)
+        """
+        Change item's appearance based on its depth in the tree
+        This allows highest-level groups to be displayed more prominently.
+        """
+        app = QtWidgets.QApplication.instance()
+        palette = app.palette()
+        textColor = palette.text().color()
+        backgroundColor = palette.button().color().darker(110)
+        altBackgroundColor = palette.dark().color()
```
NilsNemitz, line 330: I like that this pulls colors from the palette now!

Is it necessary to mix and match the color roles here?
There are already `base()` and `alternateBase()` colors, how far can we get with those?
Mixing in `button()` as an additional background color might work, but we don't really know how that might relate to the base colors depending on platform or theme.

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -319,24 +319,37 @@ def __init__(self, param, depth):
...
+
+        # if stylesheet and 'QDarkStyleSheet' in stylesheet:
+        if app.styleSheet(): # Assume the stylesheet is qdarkstyle
+            # #455364 is the background-color of the QHeaderView, #E0E1E3 is lineEdit text from QDarkStyle
+            # lighter/darker are applied to the colors to maintain a good contrast with the item:hover color
+            backgroundColor = QtGui.QColor('#455364').darker(100)
+            altBackgroundColor =  backgroundColor.lighter(130)
+            textColor = QtGui.QColor('#E0E1E3')
```
NilsNemitz, line 338: Can we get around hardcoding colors here, in the local code where it cannot be overridden?
It seems to me that that is what caused the original issue: The hardcoded values that you patched out in L326.

If I see it correctly, this is supposed to address one specific issue:
When running on windows, qdarkstyle is the best way to realize a dark mode color scheme. The problem? qdarkstyle individually skins all elements by stylesheet, and never bothers to set up a matching QPalette. The widgets provided by pyqtgraph are not in that stylesheet, so they fall back to the palette and appear basically color swapped.

So the problem in that case is the missing palette.

Why don't we fix *that* instead?
Since we already rely on mkApp to set 'darkMode' for us, can't we just sanity check the palette there?
E.g. if the stylesheet is qdarkstyle, but the base color is white... Then we manually change the application palette to a dark one, like I posted above.

Yes, it's still hardcoded. But the user can override it, and that would automatically change the appearance of GroupParameterItem, too. Even if it is done at runtime.

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -319,24 +319,37 @@ def __init__(self, param, depth):
...
+        elif app.property('darkMode'):
+            backgroundColor = palette.button().color().lighter(115)
+            altBackgroundColor = palette.light().color().lighter(190)
```
NilsNemitz, line 341: I don't want to complain too much here (Did I say thank you for your work yet? 👍 ), but one awesome thing about a palette is that it makes everything look consistent. If every bit of the application creates its own set of variations on the color set, that is somewhat less effective.
As I already asked above, how far can we get with the base() and alternateBase() colors that seem to exist exactly for this purpose?

pyqtgraph/parametertree/parameterTypes/basetypes.py
NilsNemitz, line 333: Although ntjess already pointed it out:
This seems like an overly broad condition to check. I do run a global stylesheet on my apps, and I might not be the only one. Having some elements of the app switch color as soon as you apply ANY stylesheet, even a completely empty one, seems like a good way to give a major headache to someone, somewhere, sometime 😆

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: This is exactly what should be used according to the documentation and it does work as described in light mode. 
image

However, in darkMode I tried using `base` and `alternateBase` but the color for `alternateBase` is _much_ brighter than it should be. 

image

There are other quirks when using the palette. You can see how initially the `altBackgroundColor` is set to `palette.dark()`, this works in light mode, but the color of `dark` doesn't change when in `darkMode`, so you end up with something like `#888`, which is fairly bright in `darkMode`.

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: From what I remember choosing a color palette in the windows settings really only affects the Microsoft apps(maybe native applications too) so Qt apps would not pick up these changes in the palette. 

The only problem I see for doing this in `mkQApp` is the users who use their own `QApplication` won't get see the changes. If I can get a conformation that this is okay, i think this would be a good way to handle this, although like you said, this would change the entire application palette, not just the parameterTree.

This would be a better option than applying the palette directly to the parameterTree because the window stays gray but the parameterTree is dark blue, which looks a bit weird.

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: I responded above, the colors for `base` and `alternatingBase` don't appear as they should. This might only be the case in darkMode on macOS.

pyqtgraph/parametertree/parameterTypes/basetypes.py
j9ac9k: Which bindings/versions of qt are you using for this testing @Wubbzi ? 

I'm thinking that `AlternateBase` is what we want, but the fact that it's not specified well by Qt itself kinda makes me want to say "not our problem"...

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: I've made the changes to this already but haven't pushed a new commit yet. The code now checks for both a light and dark version of the QDarkStyle stylesheet.

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: I switched to PySide2 5.15.2 when for testing QDarkStyle since they don't seem to support v6 yet. PySide6.2.1 looks the same. 

If everyone is fine with `base` and `alternateBase` across the board, this should be much less complicated. 😄

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: I was just thinking about this and realized that we aren't going to be able to check for a stylesheet here. `mkQApp` creates the QApplication and one needs to exist before a stylesheet can be applied to it.

pyqtgraph/parametertree/parameterTypes/basetypes.py
NilsNemitz: I looked at the picture you posted, but I couldn't quite figure it out. Could you give me a quick summary of how base and alternateBase colors change between light and dark mode on Mac?
Does base change correctly (dark in dark mode, light in light mode), but alternateBase does not? alternateBase is always that medium gray in your picture?

If that is the case, then that *might* be an argument to move back towards a color adjustment for the less intense alternate color. Not making up a completely new shade, but modifying the palette base shade as little as possible.

Maybe pull `backgroundColor` from the palette and then make the alternative color as
``` python
h, s, l = backgroundColor.getHsl()
l = 0.8 * (l - 0.5) + 0.5 # move towards gray
altBackgroundColor = QColor().setHsl(h,s,l)
```
(for illustration only, there's probably a few errors)
Taking out a little contrast and/or saturation should work for both light and dark schemes, so it should also be able to avoid the case distinction.

pyqtgraph/parametertree/parameterTypes/basetypes.py
NilsNemitz: Hmm. Good point.
Since the stylesheet gets set on the QApplication itself, it seems that Qt did not bother to include an event callback; other widgets have a `changeEvent()` method that should be called for this change.

That is a pity. 
A) We could give mkQApp a parameter to pass in style sheets, then the code can handle that correctly.
  Then add some documentation why mkQApp is great.
B) We could figure out if there is a signal for a style change that CAN be accessed even when we only have the QApplication
C) We could subclass QApplication with something that looks intercepts and handles the `setStyleSheet()` call.
D) We could monkey-patch QAPP to replace the setStyleSheet call with our own; which does stuff

Ideas? (A) might be reasonably safe...

Wubbzi: Option A does sound nice compared to what I've done

I'm not aware of a way to override something like `setStyleSheet()` without a subclass and when I was looking through the change log I noticed we had QApplication sub-classed but decided to change it to this function because QApplication is already a singleton. There was also a comment in `Qt/__init__.py` about segfaults being caused by a subclass of QApplication on PySide2.6/Python 3.8.7.

> Since the stylesheet gets set on the QApplication itself, it seems that Qt did not bother to include an event callback; other widgets have a `changeEvent()` method that should be called for this change.
actually this is kind of what I went with.

I figured since `QApplication` inherits from `QObject` I could just patch the `eventFIlter` and listen for `QEvent.StyleChange`

This is what I ended up with: https://gist.github.com/Wubbzi/50eea2eceae6d4220a0003ae5815fab9

For some reason  `QEvent.ApplicationPaletteChange` triggers `QEvent.StyleChange`, so If someone applies their own palette and the onPaletteChanged runs again, sees that the base color isn't pure white, their palette will be overridden with the darkPalette. 

Right now, using a global as a flag to keep onPaletteChanged from running more than once is the best way I can think of to solve this. 

Let me know what you think.

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: In Light mode when colors are set to base/alternating base, there are only 2 colors used for backgrounds. In dark mode when I use alternate base, it isn't actually the alternate base color that the tree is using to alternate every other item in the tree, its a third color that is much brighter than the 2 colors the tree is using. 

the alternating base is set on Example Parameters , and base on sample list, sample file, etc. Nothing changes except for the Light/Dark modes
Screen Shot 67
image

Wubbzi: I just went looking for which role Qt is using and it seems to be the Window color in place of alternate base in dark mode.

Screen Shot 68

Wubbzi:
This commit will detect dark mode and apply the dark_QPalette from Nils. It also overrides the QApplication.setStylesheet to set the stylesheet on the App. The code assumes that the stylesheet was designed around the original darkMode palette and will remove the dark_QPalette if a stylesheet has been set. 

I tried doing what Nils suggested, which was to add a parameter to mkQApp to accept ab argument for the stylesheet. 
```py
app = pg.mkQApp(name="Parameter Tree Example",stylesheet=qdarkstyle.load_stylesheet(palette=qdarkstyle.DarkPalette))
```
This worked well except for in the case above, `qdarkstyle.load_stylesheet` expects an instance of QApplication to exist. Since this isn't the case when `qdarkstyle.load_stylesheet`was called, it caused a warning to appear in the console: 

`No QCoreApplication instance found. Application patches not applied. You have to call load_stylesheet function after instantiation of QApplication to take effect.`

Despite the warning, everything worked as expected but when I looked for more info on the warning, I found some cases where this had caused issues when running a debugger on the code.

5yutan5:
Hi.
I don't like to use the color of qdarkstyle as default.
I think this change is outside the range of this PR.

NilsNemitz:
Hi @5yutan5 , don't worry, we are still trying to figure out how to control the color scheme so that it does not cause any sudden changes.

The idea above is to detect if the user has manually set up the qdarkstyle stylesheet. Then we know that they want a dark style, and we can set up colors to make that look better. If you don't set up that stylesheet, then you would not be force to the dark colors.

On some operating systems we can also detect if a dark or light style is selected system-wide. Ideally, we would also want to match that. But for an easy transition, we might need to keep the current color scheme as the default, and require a parameter to activate the automatic mode.

Could you tell us what system you are using? And maybe if you apply a global stylesheet?
Would you like pyqtgraph to try and match your system's color scheme, or do you prefer colors that are the same everywhere, even if it means the application will not match the normal light/dark choice?

5yutan5:
Hi @NilsNemitz, thank you for your reply.

Sorry, I talked about only dark mode on mac. I didn't make it clear enough.
I tested this branch on mac(m1, dark mode). and I'm not using global stylesheet.
In light mode, the default style is used, so there is no problem, but in dark mode, the default style is set to qdarkstyle, which confused me a little.

I want to use system's color scheme, but if it is difficult, it is no need.

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -319,24 +317,38 @@ def __init__(self, param, depth):
...
+        h, s, l = color.hue(), color.saturation(), color.lightness()
+
+        stylesheet =  app.styleSheet()
+        textColor = None
+        if '/* Light Style' in stylesheet or '/* Dark Style' in stylesheet:
+            background = QtGui.QColor(255,255,255,0)
+            altBackground = background
+            textColor = QtGui.QColor('#000000')
+            if '/* Light Style' in stylesheet:
+                textColor = QtGui.QColor('#F0F0F0')
```
NilsNemitz, line 335: Doesn't this set a light text color for "Light Style" mode?

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -319,24 +317,38 @@ def __init__(self, param, depth):
...
+        Change set the item font to bold and increase the font size on outermost groups.
+        """
+        app = QtWidgets.QApplication.instance()
+        palette = app.palette()
+        color = palette.window().color()
+        h, s, l = color.hue(), color.saturation(), color.lightness()
+
+        stylesheet =  app.styleSheet()
+        textColor = None
+        if '/* Light Style' in stylesheet or '/* Dark Style' in stylesheet:
```
NilsNemitz, line 330: In my version of the QDarkStyle style sheet, I did not find these terms, so the code never triggered.

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -319,24 +317,38 @@ def __init__(self, param, depth):
...
-                font.setBold(True)
-                font.setPointSize(font.pointSize()+1)
-                self.setFont(c, font)
+        """
+        Change set the item font to bold and increase the font size on outermost groups.
+        """
+        app = QtWidgets.QApplication.instance()
+        palette = app.palette()
+        color = palette.window().color()
+        h, s, l = color.hue(), color.saturation(), color.lightness()
```
NilsNemitz, line 326: There's a `color.getHslF()` method that returns a tuple of h, s, l, alpha in float format. That seems the most convenient here.

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -319,24 +317,38 @@ def __init__(self, param, depth):
...
+        stylesheet =  app.styleSheet()
+        textColor = None
+        if '/* Light Style' in stylesheet or '/* Dark Style' in stylesheet:
+            background = QtGui.QColor(255,255,255,0)
+            altBackground = background
+            textColor = QtGui.QColor('#000000')
+            if '/* Light Style' in stylesheet:
+                textColor = QtGui.QColor('#F0F0F0')
+        elif app.property('darkMode'):
+            background = QtGui.QColor.fromHsl(h, s, l * .6, 255)
```
NilsNemitz, line 337: It might be better to work with the float version of the Hsl methods, should be `fromHslF()` here.

I would also suggest to stick to the original palette background here and change as few colors as possible.

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -319,24 +317,38 @@ def __init__(self, param, depth):
...
         else:
-            for c in [0,1]:
-                self.setBackground(c, QtGui.QBrush(QtGui.QColor(220,220,220)))
-                self.setForeground(c, QtGui.QBrush(QtGui.QColor(50,50,50)))
-                font = self.font(c)
-                font.setBold(True)
-                #font.setPointSize(font.pointSize()+1)
-                self.setFont(c, font)
+            background = QtGui.QColor.fromHsl(h, s, l)
+            altBackground = QtGui.QColor.fromHsl(h, s, l * .95, 255)
```
NilsNemitz, line 341: You can do this without the light mode / dark mode detection by something like
`l = 0.5 + 0.8 * (l-0.5)`  , to pull any color a little closer to the 0.5 lightness value.

j9ac9k:
Hi @Wubbzi 

I just ran this branch, this worked beautifully 🎊   Handled switch between light/dark mode seamlessly!

I would recommend adopting the suggested changes @NilsNemitz regarding `QColor.fromHslF` and `QColor.getHslF`.  

Lastly, looks like you're using a python 3.8 feature, `cached_property`.  Unfortunately this is at odds with [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html), which has us supporting python 3.7 for all releases until December 26th of this year.  So we can certainly keep the `cached_property` feature, but if we do, we're going to have to hold tight on merging this PR until it's a bit closer to the end of the year.

Looking at the diff, the `cached_property` has to do w/ the style-sheet detection; in which case I would recommend moving that bit of code into a separate PR that we can merge later.

Wubbzi:
> I would recommend adopting the suggested changes @NilsNemitz regarding QColor.fromHslF and QColor.getHslF.

I've already made these changes to my local branch, I just need to push them to github

> Lastly, looks like you're using a python 3.8 feature, cached_property. Unfortunately this is at odds with NEP-29, which has us supporting python 3.7 for all releases until December 26th of this year. So we can certainly keep the cached_property feature, but if we do, we're going to have to hold tight on merging this PR until it's a bit closer to the end of the year.

Unless the tree has 1000s of groups, the difference with and without the caching isn't really noticeable, so I'm fine with moving it out, or changing it to use functools lru cache, if you'd like have some sort of caching until we're able to use cached_property.

j9ac9k:
> > I would recommend adopting the suggested changes @NilsNemitz regarding QColor.fromHslF and QColor.getHslF.
> 
> I've already made these changes to my local branch, I just need to push them to github

👍🏻 

> > Lastly, looks like you're using a python 3.8 feature, cached_property. Unfortunately this is at odds with NEP-29, which has us supporting python 3.7 for all releases until December 26th of this year. So we can certainly keep the cached_property feature, but if we do, we're going to have to hold tight on merging this PR until it's a bit closer to the end of the year.
> 
> Unless the tree has 1000s of groups, the difference with and without the caching isn't really noticeable, so I'm fine with moving it out, or changing it to use functools lru cache, if you'd like have some sort of caching until we're able to use cached_property.

I'm good w/o caching.

Wubbzi:
This update includes a light and dark palette meant to be used with a QDarkStyle stylesheet. 

Either the Palette.ID from the `QDarkStyle` or the style of the palette as a string`light`/`dark` will work. 
```py
pg.applyPalette(app, qdarkstyle.DarkPalette.ID)
app.setStyleSheet(qdarkstyle.load_stylesheet(palette=qdarkstyle.DarkPalette))

pg.applyPalette(app, 'light')
app.setStyleSheet(qdarkstyle.load_stylesheet(palette=qdarkstyle.LightPalette))
```

pyqtgraph/Qt/__init__.py
```diff
@@ -415,11 +415,13 @@ def mkQApp(name=None):
     ============== ========================================================
     """
     global QAPP
-    
+    global PALETTE
```
NilsNemitz, line 418: is this still in use?

pyqtgraph/Qt/__init__.py
```diff
@@ -415,11 +415,13 @@ def mkQApp(name=None):
...
+
     def onPaletteChange(palette):
         color = palette.base().color().name()
         app = QtWidgets.QApplication.instance()
-        app.setProperty('darkMode', color.lower() != "#ffffff")
+        darkMode = color.lower() < "#e0e0e0"
```
NilsNemitz, line 423: Does that work? Doesn't it look at the red channel, mostly?
How about `color.LightnessF() < 0.5` ?

pyqtgraph/__init__.py
```diff
@@ -14,6 +14,7 @@
 
 ## 'Qt' is a local module; it is intended mainly to cover up the differences
 ## between PyQt and PySide.
+from . import palette
```
NilsNemitz, line 17: Nitpick:
This file might want to live in the (already existing) 'colors' folder?

pyqtgraph/palette.py
```diff
@@ -0,0 +1,72 @@
+from .Qt import QtGui
+
+
+def dark_QPalette():
+    RED = QtGui.QColor('#F00000')
```
NilsNemitz, line 5: Oops, I only defined that for debugging :)
Could you delete it? 
Probably "BLACK", too, Although that has at least some usability.

pyqtgraph/palette.py
```diff
@@ -0,0 +1,72 @@
...
+    qpal.setColor(QtGui.QPalette.Disabled, QtGui.QPalette.Button, BG_NORMAL)
+    qpal.setColor(QtGui.QPalette.Disabled, QtGui.QPalette.ButtonText, FG_DARK)
+    qpal.setColor(QtGui.QPalette.Disabled, QtGui.QPalette.WindowText, FG_DARK)
+    qpal.setColor(QtGui.QPalette.Disabled, QtGui.QPalette.Text, FG_DARK)
+    qpal.setColor(QtGui.QPalette.Disabled, QtGui.QPalette.Highlight, BG_LIGHT)
+    qpal.setColor(QtGui.QPalette.Disabled, QtGui.QPalette.HighlightedText, FG_DARK)
+    return qpal
+
+
+def light_QPalette():
```
NilsNemitz, line 40: Since we are keeping these private, should we give them more specific names?
`getQDarkStyleDarkQPalette()` and `getQDarkStyleLightQPalette()`, maybe?
I think we might want to keep the namespace open for our own version of just "light" and "dark" palettes. 
The very dark colors in QDarkStyle do not work well with e.g. black outlines of UI elements and the black shading applied according to button state. It really needs the stylesheet.
I'd ultimately like to have a dark mode palette that works without depending on an external stylesheet, and I would want to refer to that as just "dark". :)

pyqtgraph/parametertree/ParameterItem.py
```diff
@@ -158,13 +158,20 @@ def nameChanged(self, param, name):
...
-        fm = QtGui.QFontMetrics(self.font(0))
-        textFlags = QtCore.Qt.TextFlag.TextSingleLine
-        size = fm.size(textFlags, self.text(0))
-        size.setHeight(int(size.height() * 1.35))
-        size.setWidth(int(size.width() * 1.15))
-        self.setSizeHint(0, size)
+
+        # when changing from light to dark mode updateDepth(GroupParamItem) will insert what I think is the
+        # invisible root node. Most of the time, in the example app the title for this node
+        # will be empty, occasionally its named params.
```
NilsNemitz, line 164: "it is name 'params'"? 
Since the comment is long already, maybe include what actually happens?
"In this case, we hide it again by setting sizeHint to zero."? (Is that what happens?)

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -1,11 +1,10 @@
 import builtins
+from functools import  partial
```
NilsNemitz, line 2: do we still use this?

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -1,11 +1,10 @@
...
 
-from ... import functions as fn
-from ... import icons
-from ...Qt import QtCore, QtGui, QtWidgets
 from ..Parameter import Parameter
 from ..ParameterItem import ParameterItem
-
+from ... import functions as fn, icons
```
NilsNemitz, line 6: I am sure somebody strong has opinions on how many lines this should use.  :)
I don't, but I find that this reads a little curiously when glancing over it and see "as fn, icons"

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -274,22 +274,28 @@ def __init__(self, *args, **kargs):
...
     """
     Group parameters are used mainly as a generic parent item that holds (and groups!) a set
     of child parameters. It also provides a simple mechanism for displaying a button or combo
     that can be used to add new parameters to the group.
     """
+
     def __init__(self, param, depth):
         ParameterItem.__init__(self, param, depth)
+        # prevent 0 depth item's font from growing on every call of updateDepth
+        self.fontPointSize = partial(self.font(0).pointSize)
```
NilsNemitz, line 298: I have no experience with this. 
If this defines a `fontPointSize` method for WidgetParameterItem, maybe the boring, verbose way of going through `def fontPointSize` might be more readable?

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -319,24 +325,26 @@ def __init__(self, param, depth):
...
+        """
+        Change set the item font to bold and increase the font size on outermost groups.
+        """
+        app = mkQApp()
+        palette = app.palette()
+        color = palette.base().color()
+        h, s, l, a = color.getHslF()
+        lightness = 0.5 + (l - 0.5) * .8
+        background = QtGui.QColor.fromHslF(h, s, lightness, a)
+        altBackground = color
```
NilsNemitz, line 337: No problem, but you might be able to help future devs to figure this out faster.
What you now call "color" is the default background color.
What you call "background" is a modified background color.

Maybe swapping the names around makes that easier to follow (and saves a line):
``` python
background = palette.base().color()
h, s, l, a = background.getHslF()
lightness = 0.5 + (l - 0.5) * .8
altBackground = QtGui.QColor.fromHslF(h, s, lightness, a)
```

pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -319,24 +325,26 @@ def __init__(self, param, depth):
...
+
+        for c in [0, 1]:
+            font = self.font(c)
+            font.setBold(True)
+            if depth == 0:
+                background = altBackground
+                font.setPointSize(self.fontPointSize() + 1)
+            self.setBackground(c, background)
```
NilsNemitz, line 345: If you swap above, these need to swap, too :)

pyqtgraph/__init__.py
```diff
@@ -463,3 +464,14 @@ def stack(*args, **kwds):
     except NameError:
         consoles = [c]
     return c
+
+
+def applyPalette(app,name):
```
NilsNemitz, line 469: Needs documentation before merge, but we can work that out when we know that this is the way we want to do things.
(I like it.)

pyqtgraph/__init__.py
```diff
@@ -463,3 +464,14 @@ def stack(*args, **kwds):
...
+    name = name.lower()
+    if name == "light":
```
NilsNemitz, line 471: I would like to advertise (in the docs) a name that is specific to QDarkStyle here.
'qdarkstyle' and 'qdarkstylelight' come to mind.
We can default 'dark' and 'light' to these, but we might point these simple names at palettes that do not depend on a stylesheet in the future?
To clarify
`if name in ['qdarkstylelight', 'light']:`
and
`if name in ['qdarkstyle','dark']: `
seem more appealing to me.

The docs will only mention the qdarkstyle names for now.

pyqtgraph/Qt/__init__.py
Wubbzi: Done

pyqtgraph/Qt/__init__.py
Wubbzi: No, I went too far back when refactoring.

pyqtgraph/__init__.py
Wubbzi: moved to colors, but we might want to think about making it a package and renaming the module to qdarkstylePalettes so palettes.py is available in the future for general use palettes

pyqtgraph/__init__.py
Wubbzi: I'm changing the `name` parameter to `style`. Do you think we should set a default where `style = qarkstyleDark`?

pyqtgraph/__init__.py
Wubbzi: To avoid confusion in the future about why the `'dark'` palette suddenly looks different with the qdarkstyle dark palette stylesheet, should we just stick with names specific to qdarkstyle?
```
if name == "qdarkstylelight":
    ...
elif name =="qdarkstyledark"
    ...
```

pyqtgraph/palette.py
Wubbzi: I was wondering what those were but I didn't want to make any changes. 

Done

pyqtgraph/palette.py
Wubbzi: I'm all for `QDarkStyleDarkQPalette()` and `QDarkStyleLightQPalette()` but pep8 doesn't like the java style getters/setter names. We might even want to make these properties.

> I'd ultimately like to have a dark mode palette that works without depending on an external stylesheet, and I would want to refer to that as just "dark".

Agree!  [more about this here](https://github.com/pyqtgraph/pyqtgraph/pull/2101/files/4a565507d760cc8cd917134a32d322e9832786e2#r755812152)

pyqtgraph/parametertree/ParameterItem.py
Wubbzi: >"it is name 'params'"?

changed it to `occasionally the title is 'params'.`, if that helps to be more clear. 

> "In this case, we hide it again by setting sizeHint to zero."? (Is that what happens?)

Not exactly, it appears to be hidden by default, unless given a size hint by the code below. I'm making a small change to the code so its easier to tell whats happening. 

```py
        title = self.param.title()
        if not title or title == 'params':
            return
        self.setText(0, title)
        fm = QtGui.QFontMetrics(self.font(0))
        textFlags = QtCore.Qt.TextFlag.TextSingleLine
        size = fm.size(textFlags, self.text(0))
        size.setHeight(int(size.height() * 1.35))
        size.setWidth(int(size.width() * 1.15))
        self.setSizeHint(0, size)
```

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: Yes, its being used here in `GroupParameterItem` and is being used to always return the original font size: The current font size is returned and incremented by 1 for items where depth=0. Since `updateDepth` is now run every time(to paint the backgrounds), when switching from dark/light system themes, the font size would increase 1 point, every time the theme is switched without it(or something similar). 

```
    def __init__(self, param, depth):
        ParameterItem.__init__(self, param, depth)
        # prevent 0 depth item's font from growing on every call of updateDepth
        self.fontPointSize = partial(self.font(0).pointSize)
```

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: > I am sure somebody strong has opinions on how many lines this should use. :)

I certainly don't.

I can see by some of the other changes in this commit that I must have accidentally run the code cleanup on this file before committing. It will automatically organize imports, remove trailing white space, etc. 

I've changed it back to 
```
from ... import icons
```

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: In this specific case, this is basically the same as `self.fontPointSize = lambda: self.font(0).pointSize()` but is marginally faster. 
I didn't define a method because I didn't want to add any methods to the object that weren't necessary elsewhere in the code. 

I think this is pretty readable as is but I'll change this into a property instead. 

```py
    @ property
    def fontPointSize(self):
        return self.font(0).pointSize()
```

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: I guess I did it this way because I didn't want to call `setBackground(altBackground)`, pick between using `setBackground` more than once, or writing an else.

Wubbzi: I think I'd be more confused seeing `setBackground(altBackground)` and having to trace my way through the code to figure out why it wasn't called with `background`

pyqtgraph/__init__.py
NilsNemitz: I think we want people to read the docs on this, so I think it is better not to set a default.
We should accept a QPalette for the `style` parameter, though, and we should check the base color to update the `darkMode` flag!

pyqtgraph/__init__.py
NilsNemitz: I think those would be good canonical names.
I would tend towards "qdarkstyle" for the dark style, that seems less redundant.
I wouldn't mind an alias to "qdarkstyledark". 

You are probably right about not having a 'dark' label until we think about that some more.

pyqtgraph/palette.py
NilsNemitz: I think content of palette.py needs some more thought.
This might better use some internal dictionary that is easy to add to. Ideally, the user should be able to throw in their own palette info into the colors folder and grab it from there in multiple applications.

At this time, the strategy might be to keep everything in here private and marked experimental. `applyPalette` is the only supported access to this... for now.

NilsNemitz: Regarding the naming: 
We are basically following the Qt CamelCase naming conventions, at least on the API side. So we regularly deviate from python standards. @j9ac9k has the best overview here, though.

pyqtgraph/parametertree/ParameterItem.py
NilsNemitz: How about just sticking a comment on the setSizeHint line?
`# This makes sure that the label remains visible`  
...or invisible, because I still don't know what this is actually trying to do :)

pyqtgraph/parametertree/parameterTypes/basetypes.py
NilsNemitz: Oh, I find that much easier to read :)
"I will give you this if asked for fontPointSize()"

Last question:
This is not solved by simply overriding / providing a matching `pointSize()` method here?

pyqtgraph/__init__.py
Wubbzi: >  We should accept a QPalette for the style parameter, though, and we should check the base color to update the darkMode flag!

Do you mean in addition to accepting a string like this?
```py
def applyPalette(app, style):
    if isinstance(style, str):
        style = style.lower()
        if style in ['qdarkstylelight', 'light']:
            p = palette.getQDarkStyleLightQPalette()
        elif style in ['qdarkstyle', 'dark']:
            p = palette.getQDarkStyleDarkQPalette()
        else:
            raise ValueError(f'no palette by the name {style} exists')
    elif isinstance(style, QtGui.QPalette):
        p = style
        app.setPalette(p)
    else:
        raise TypeError('style either be a string or QPalette')
    app.setPalette(p)
 ```

File: pyqtgraph/parametertree/parameterTypes/basetypes.py (hunk 20 above)
Comment (NilsNemitz) [reply pyqtgraph/parametertree/parameterTypes/basetypes.py:345]: Maybe just put 
``` python
 if depth == 0:
    font.setPointSize(self.fontPointSize() + 1)
    self.setBackground(c, background)
else:
    self.setBackground(c, altBackground)`
```
Then it should be clear what background goes where.

pyqtgraph/__init__.py
Wubbzi: Okay, the changed to:

``` 
        if style == 'qdarkstylelight':
            p = palette.getQDarkStyleLightQPalette()
        elif style in ['qdarkstyle','qdarkstyledark']:
            p = palette.getQDarkStyleDarkQPalette()
```

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: Done.

pyqtgraph/parametertree/parameterTypes/basetypes.py
pijyoi: Would it have been the same as ```self.fontPointSize = self.font(0).pointSize```?

pyqtgraph/palette.py
j9ac9k: @Wubbzi we generally deviate from pep8 when it comes to naming convention to public portions of our API as @NilsNemitz suggested.  We will go `snake_case` for methods intended for internal methods, and use the leading underscore to designate if something should be considered private or not.

On top of using `camelCase` for public API methods; we try and follow Qt naming conventions when feasible (for example we do this with `.setData()` `.getData()` and a variety of other methods that we feel are analogous to other parts of the Qt API.

Of course the above are just guidelines.  ... regarding `applyPalette` would it be more appropriate to use `.setPalette()`?  That method is used in other parts of the Qt framework: 
image

pyqtgraph/parametertree/ParameterItem.py
Wubbzi: The only thing I've changed here is in this method is:

```py
        # This makes sure that items with no title or the title 'params' remain invisible
        
        title = self.param.title()
        if not title or title == 'params':
            return
```
If there isn't a title or the title is ''params'` then we don't want to run the code below to give it a size. It seems to default to 0,0 width/height, but I didn't look very deeply into it because simply checking that the title wasn't empty did what we wanted. 

> How about just sticking a comment on the setSizeHint line?
This isn't my code, but the code on the size hint line sets a size hint of `size` for column 0. The size is set to the width/height *1.35 of a single line of whichever font self.font(0) returns

pyqtgraph/palette.py
Wubbzi: > We are basically following the Qt CamelCase naming conventions, at least on the API side

I don't have any problem with the case conventions pg uses. The snake_case/true_property stuff that was added in pyside6 looks so wrong, and I hope it never catches on. 

I was trying to stick with the Qt conventions. for example, Qt has `setMaximumHeight()` but when you want to get it, its just `maximumHeight()` not `getMaximumHeight()` I stuck with the capital Q because I was afraid it would look weird. Qt does lowercase their methods that begin with q, like `qAbs`,`qRound` but you don't see those used in Python very often.

pyqtgraph/parametertree/ParameterItem.py
NilsNemitz: I maybe missed this bit then
--> `# This makes sure that items with no title or the title 'params' remain invisible`
That was the thing that wasn't clear to me.

Sorry! It just felt weird that there was a three line comment that seemed to build up to something and then didn't tell me what it was actually doing in the end :)

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: I don't think I follow..  If you're asking "Why does this need to be a property, wouldn't a method work just the same?" the answer is yes, it would work but ` font.setPointSize(self.fontPointSize() + 1)` looks weird compared to `                font.setPointSize(self.fontPointSize + 1)`

pyqtgraph/__init__.py
NilsNemitz: Yes. Except for not setting the palette twice, maybe.

And with the overall addition of the basecolor lightness check from mkQApp.

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: > Would it have been the same as self.fontPointSize = self.font(0).pointSize?

The behavior would have been the same. The difference to me is, when an attribute is defined as in `__init__` generally you wouldn't expect that its required to be called later. Seeing it defined in a lambda, or with partial lets you know that it is a callable.

pyqtgraph/parametertree/parameterTypes/basetypes.py
NilsNemitz: This is nitpicky and not a large concern. Overall, we *typically* don't reach into other objects to grab their properties. Even though we are not Java, it *seems* that we more commonly go through accessor methods. 

I think my viewpoint is
- if this is used internally only (as in your example), then do we even need it over accessing `self.font(0).pointSize()` directly where applicable?
- if this provides a regular interface for other parts of the code, going through a normal method definition is the easiest to find and document. And it clearly sells "use this to get this information". This would be particularly true if you are basically reporting the effective pointsize for this `ParameterItem`. Other objects do that through `pointSize()`, wouldn't it be nice to match that?

pyqtgraph/__init__.py
NilsNemitz: I like that :)

pyqtgraph/palette.py
Wubbzi: @j9ac9k oooh, another dash user!

> We will go snake_case for methods intended for internal methods, and use the leading underscore to designate if something should be considered private or not.

Oh, I didn't realize thats what was meant by 
> the strategy might be to keep everything in here private and marked experimental. 

> regarding applyPalette would it be more appropriate to use .setPalette()
Yep. I was trying to stick as closely to what Nils suggested as I could, so after I copied it out of chat I never changed it, but it has now been changed.

NilsNemitz: Well, my thought process wasn't that sophisticated.
More like "make sure nobody uses it, whatever we put now, we'll break it later when we put more functionality there"

j9ac9k: > another dash user!

if you're on macOS and a developer and _not_ using Alfred + Dash ...that's a crime 😆

pyqtgraph/parametertree/parameterTypes/basetypes.py
Wubbzi: > if this is used internally only (as in your example), then do we even need it over accessing self.font(0).pointSize() directly where applicable?

Actually, now that I think about it, the method or lambda won't work either. 

```
# we define the font as
font = self.font(0/1)
# we increment based on the current size
font.setPointSize(self.font(0/1).pointSize() + 1)  # this will grow when palette is changed. 
# same goes for anything that calls the method of self.font().pointSize()
# to get the current size, as its increased every time we set the font
self.setFont(0/1,font)
```

If we set the values in the `__init__` 

```

self.fontPointSize = partial(self.font(0).pointSize) # set only once when class initialized 
self.fontPointSize = self.font(0).pointSize          # set only once when class initialized
# we increment with the same pointSize every time updateDepth is run
font.setPointSize(self.fontPointSize()+1) 
# when the font is set it will always use size 14(or whatever the actual size is)
```

pyqtgraph/__init__.py
Wubbzi: > Yes. Except for not setting the palette twice, maybe.

I just wanted to make sure it was really set. Like really set... 
Fixed!

>
And with the overall addition of the basecolor lightness check from mkQApp.
That would go just before or after the final setPalette call, when we know that p contains a QPalette.

Gotcha will emit the palette through paletteChanged, though... is the darkMode property used other than the example app?

pyqtgraph/parametertree/parameterTypes/basetypes.py
pijyoi: So the issue then is that ```self.font(0)``` returns a *copy of* (instead of a reference to) a ```QFont```?
From the Qt docs:
```c++
QFont QTreeWidgetItem::font(int column) const
```

pyqtgraph/__init__.py
NilsNemitz: Ah, if we catch the change there and set the flag, then we're all good!

Let's assume for now that *some* part of the code *somewhere* probably does *something* with the darkMode flag. So it's nice to always have it up-to-date.

pyqtgraph/parametertree/ParameterTree.py
```diff
@@ -130,6 +134,16 @@ def contextMenuEvent(self, ev):
         if hasattr(item, 'contextMenuEvent'):
             item.contextMenuEvent(ev)
             
+    def updatePalette(self):
+        """
+        - called when application palette changes -
```
NilsNemitz, line 139: Pretty sure that this is what makes sphinx confused, and that
"Called when the application palette changes"
would not cause any trouble.

pyqtgraph/colors/palette.py
```diff
@@ -1,9 +1,7 @@
-from .Qt import QtGui
+from pyqtgraph.Qt import QtGui
```
j9ac9k: we use relative imports because one of the supported ways this library "can" be used (although we don't recommend it) is by placing a pyqtgraph directory into a project.  Would you mind switching back to a relative import?

j9ac9k: whops, saw this on the commit, not on the final result, disregard!

pyqtgraph/parametertree/parameterTypes/basetypes.py
j9ac9k: I think this isn't used any more with the most recent commit; once this import is removed, I think we can merge this 🎉

PR #2103, opened 2021-11-15
Fix stuck ColorBarItem

### LINKED ISSUE #2102: ColorBarItem lower axis value stuck
### Short description

After dragging lower handle down till lower limit is hit on the axis, dragging it up no longer shifts the axis values.

### Code to reproduce

1) Run the following script
2) Drag the lower handle downwards. The axis lower bound gets locked at -0.2.
3) Drag the lower handle upwards. The axis lower bound stays locked at -0.2.

```python
import pyqtgraph as pg
import numpy as np

pg.mkQApp()

pw = pg.PlotWidget()
pw.show()
imgitem = pg.ImageItem(axisOrder='row-major')
pw.addItem(imgitem)
xy = np.linspace(-3, 3, 1000)
data = np.sinc(xy)[:,None] * np.sinc(xy)
cmap = pg.colormap.get('viridis')
limits = data.min(), data.max()
bar = pg.ColorBarItem(values=limits, colorMap=cmap, limits=limits, rounding=0.1)
bar.setImageItem(imgitem, insert_in=pw.getPlotItem() )
imgitem.setImage(data)

pg.exec()
```

### Expected behavior

The axis lower bound value should not get stuck at -0.2

### Real behavior

The axis lower bound value gets stuck at 0.2

### Tested environment(s)

 * PyQtGraph version: pyqtgraph 0.12.3
 * Qt Python binding: 
 * Python version: 3.8.10
 * NumPy version: 1.21.2 

NilsNemitz:
This address #2102:

The test case there (Thank you @pijyoi!) triggers a case I had not considered in the range adjustment code of ColorBarItem:
- Set the (top) range and limit to a value that is slightly below a rounding step.
- Start adjusting the bottom range

Result:
The top end of the range is detected out of range and corrected down, but restored to the original value by rounding.
This is (continuously) interpreted as both top and bottom end moving, a "region move" normally triggered by dragging the center of the bar.
In that case, the previous size of the region is held fixed to avoid squishing the range against the top or bottom limit.
Which is the bug encountered in #2102 : The region actively resists any attempt to make it smaller. 

This PR 
- slightly expands the limits to make them consistent with the rounding step
- double checks for a region move before blocking the region size.

I hope that eliminates the issue in the future and closes #2102

### DIFF
--- a/pyqtgraph/graphicsItems/ColorBarItem.py
+++ b/pyqtgraph/graphicsItems/ColorBarItem.py
@@ -72,14 +72,15 @@ class ColorBarItem(PlotItem):
         self.horizontal = bool( orientation in ('h', 'horizontal') )
 
         self.lo_prv, self.hi_prv = self.values # remember previous values while adjusting range
-        if limits is None:
-            self.lo_lim = None
-            self.hi_lim = None
-        else:
+        self.lo_lim = None
+        self.hi_lim = None
+        if limits is not None:
             self.lo_lim, self.hi_lim = limits
             # slightly expand the limits to match the rounding steps:
-            self.lo_lim = self.rounding * math.floor( self.lo_lim/self.rounding )
-            self.hi_lim = self.rounding * math.ceil(  self.hi_lim/self.rounding )
+            if self.lo_lim is not None:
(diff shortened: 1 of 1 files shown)


j9ac9k:
LGTM, thanks @NilsNemitz !

PR #2104, opened 2021-11-15
change binder to requirements.txt

pijyoi:
As hinted in https://github.com/pyqtgraph/pyqtgraph/pull/2055#issuecomment-968403025, non-conda environment is preferred, which at least allows us to get PyQt5 5.15. We can't use Qt6 as mybinder.org provides Ubuntu 18.04 containers.

The previous usage of ```environment.yaml``` was less of a choice and more of a copy of existing usage in other projects, while struggling to get the binder up and working.


j9ac9k:
Thanks @pijyoi I should have caught this and given appropriate feedback during #2055 my apologies!

PR #2106, opened 2021-11-16
set PLATFORM=offscreen in binder/start

pijyoi:
1) Found the correct place to set environment variables for binder.
2) Update simple image example to have a ColorBarItem
    * better demonstrates the interactive capability


j9ac9k:
Hi @pijyoi this LGTM.  I noticed that the example with the notebook sliders to adjust the colors of the image was removed, I assume this was done because scipy was removed as a dependency in binder?   How would you feel about re-introducing an example like that (doesn't have to be part of this PR).

pijyoi:
>I noticed that the example with the notebook sliders to adjust the colors of the image was removed, I assume this was done because scipy was removed as a dependency in binder?

The dependency on scipy was one reason. Another was that it was a poor man's version of ```examples/HistogramLUT.py```, therefore not very satisfactory.

The image example in ```simple.ipynb``` hopes to better demonstrate the viewing of "scientific" images, which are:
1) often in floating point.
2) often have data ranges that aren't clipped to some storage format limitation. e.g 0-255, 0-65535, 0.0-1.0, etc
   * i.e. with pyqtgraph, you are not obligated to first do some contrast stretching before displaying it. you can just display it and then interactively play around with the levels.

Issue #2020, opened 2021-10-15
Zoom Issues with DateTime Axis


agjaeger:
### Short description
The DateTime Axis has rendering issues with zoomed in past the 5ish minute level. 
Either the rendered plot is missing data or during panning the axis and plot lose sync.

I have attempted to use the optimization options such as clipToView, however they do not seem to provide benefit.
There is a [thread](https://www.mail-archive.com/pyqtgraph@googlegroups.com/msg01054.html) outlining zoom issues from floating point rounding. Maybe this is a similar issue?

The DateTime axis accepts seconds since epoch as its time units, is it possible to use time_ns() to get nanoseconds since epoch or milliseconds since epoch?

### Code to reproduce

```python

import sys
import time
import os

import numpy as np
import pyqtgraph as pg
from pyqtgraph.Qt import QtCore, QtGui, QtWidgets, QT_LIB, loadUiType

def make_data ():
    now = time.time()
    npoints = int(1e5)
    seconds_in_day = 86400

    x = np.linspace(now - 6*30*seconds_in_day, now, npoints)
    y = np.random.rand(npoints)

    return x, y

if QT_LIB.startswith('PyQt'):
    wrapinstance = pg.Qt.sip.wrapinstance
else:
    wrapinstance = pg.Qt.shiboken.wrapInstance

# ----------
# set options
pg.setConfigOption('background', 'w')
pg.setConfigOption('foreground', 'k')

pg.setConfigOption('useOpenGL', True)
pg.setConfigOption('enableExperimental', True)

# don't limit frame rate to vsync
sfmt = QtGui.QSurfaceFormat()
sfmt.setSwapInterval(0)
QtGui.QSurfaceFormat.setDefaultFormat(sfmt)

# ----------
# setup ui
app = pg.mkQApp("Plot Speed Test")
default_pen = pg.mkPen(width=2)
pw = pg.PlotWidget()

splitter = QtWidgets.QSplitter()
splitter.addWidget(pw)
splitter.show()

pw.setWindowTitle('zoom test')
pw.setLabel('bottom', 'Index', units='B')

# ----------------
# setup plot
lcurve = pw.plot(
    *make_data(), 
    connect="pairs",
    pen=default_pen
)
pw.addItem(lcurve)

# ----------------
# set plot options
pw.setMouseEnabled(x=True, y=False)
pw.setYRange(0, 1, padding=0.5, update=False)
[...]

PR #2111, opened 2021-11-18
PlotCurveItem OpenGL : avoid automatic conversion from float64 to float32

pijyoi:
There is an inefficiency in the current PlotCurveItem OpenGL implementation:
1) temporary float64 buffer created for vertices
2) tell PyOpenGL to use it as float32 (thus triggering a float64 - > float32 conversion) 

This PR changes step (1) to use a temporary float32 buffer right from the start.
This gains a few fps with 1,000,000 points: ```examples\PlotSpeedTest.py --opengl --frames 3 --nsamples 1000000```

Alternatively, if it is considered a bug that ```glVertexPointerf``` was used, then that should be changed instead to ```glVertexPointer(2, gl.GL_DOUBLE, 0, pos)```. However, this would then be a change in behavior to use double precision, and in addition, the calls to setting up the stencil should also be changed to double precision.

### DIFF
--- a/doc/source/config_options.rst
+++ b/doc/source/config_options.rst
@@ -29,7 +29,16 @@ exitCleanup        bool                True               Attempt to work around
 useOpenGL          bool                False              Enable OpenGL in GraphicsView.
 useCupy            bool                False              Use cupy to perform calculations on the GPU. Only currently applies to
                                                           ImageItem and its associated functions.
+useNumba           bool                False              Use numba acceleration where implemented.
 enableExperimental bool                False              Enable experimental features (the curious can search for this key in the code).
+                                                          In combination with useOpenGL, this makes PlotCurveItem use PyOpenGL
+                                                          for curve drawing.
+                                                          
+
+                                                          **Caveats**
+
+                                                            * Only a very limited subset of the full options of PlotCurveItem is implemented.
+                                                            * Single precision is used. This may cause drawing artifacts.
 crashWarning       bool                False              If True, print warnings about situations that may result in a crash.
 ================== =================== ================== ================================================================================
 
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -879,7 +879,7 @@ class PlotCurveItem(GraphicsObject):
 
         try:
             x, y = self.getData()
-            pos = np.empty((len(x), 2))
+            pos = np.empty((len(x), 2), dtype=np.float32)
             pos[:,0] = x
             pos[:,1] = y
             gl.glEnableClientState(gl.GL_VERTEX_ARRAY)
@@ -906,7 +906,7 @@ class PlotCurveItem(GraphicsObject):
                 else:
                     gl.glDisable(gl.GL_LINE_SMOOTH)
 
-                gl.glDrawArrays(gl.GL_LINE_STRIP, 0, int(pos.size / pos.shape[-1]))
+                gl.glDrawArrays(gl.GL_LINE_STRIP, 0, pos.shape[0])
             finally:
                 gl.glDisableClientState(gl.GL_VERTEX_ARRAY)
         finally:


j9ac9k:
Nice find @pijyoi 

I'm not sure what the consequences are of going to double-precision, so I'm inclined to think that we should probably maintain the single point precision given that's how it's been.

NilsNemitz:
I can't help but feel that single-wide floats with a resolution of only around 1E-7 are a little weak for showing arbitrary data.

On the other hand, the placement of OpenGL in the library currently seems to be "fast at all costs", so sticking with single floats is probably justified. The viewing range and axis tick calculations would still work CPU-side, using doubles, I guess. So at least this should not require any special treatment of range limitations (like in the various breakages we get when zooming out too far).

How reliable is OpenGL support for double-wide float vertices? Are there platforms that we want to support where `glVertexPointer(2, gl.GL_DOUBLE, 0, pos)` is not available? That would certainly settle the question.

On a somewhat modern system, is there still a performance penalty for operation with double vertices?
I guess moving the data to graphics card memory will take twice as long...

pijyoi:
>I can't help but feel that single-wide floats with a resolution of only around 1E-7 are a little weak for showing arbitrary data.

Yes, that's correct. #2020 has an example with artifacts when using OpenGL.

>On the other hand, the placement of OpenGL in the library currently seems to be "fast at all costs", so sticking with single floats is probably justified.

And also the OpenGL codepath has been untouched for ages. So not changing its behavior is the easiest course of action.

>How reliable is OpenGL support for double-wide float vertices? Are there platforms that we want to support where glVertexPointer(2, gl.GL_DOUBLE, 0, pos) is not available? That would certainly settle the question.

I remember reading language somewhere that an implementation is only needed to provide 24-bits of precision.

>On a somewhat modern system, is there still a performance penalty for operation with double vertices?
I guess moving the data to graphics card memory will take twice as long...

If you use CUDA with consumer graphics cards, there is a 32x performance penalty to using double precision versus single precision. So I would suspect that that would be true when using OpenGL too. Either that or the driver truncates it to float32 anyway, which still makes it compliant.

NilsNemitz:
I'd say that is a fairly strong point for sticking to single-wide floats, then. There is a still a lot that can be done with 23 bits of mantissa.

Is there a good place to mention the single float limitation in the docs?

j9ac9k:
32x performance penalty?!?!?! that's _brutal_ 

When I started reading more about cross-platform openGL code, I was coming across some articles (granted, they were ancient, like most openGL content) indicating that 64 bit float support within opengl drivers was fairly inconsistent... anyway that's a lot of reasons listed above to keep things as they are.

I am 👍🏻  on @NilsNemitz 's suggestion on adding a blurb in the docs about this limitation.

pijyoi:
See the comments in the following link regarding double precision:
https://stackoverflow.com/questions/59740148/opengl-floating-point-precision

Basically, double precision is not used for processing even if the API allows you to provide double precision values.

j9ac9k:
> See the comments in the following link regarding double precision: https://stackoverflow.com/questions/59740148/opengl-floating-point-precision
> 
> Basically, double precision is not used for processing even if the API allows you to provide double precision values.

I was just chatting with the [moderngl](https://github.com/moderngl/moderngl) maintainers and they explained more or less the same thing to me

pijyoi:
>adding a blurb in the docs about this limitation.

Where would be a good place to put blurb? ```enableExperimental``` is effectively an undocumented option.

j9ac9k:
I was thinking in the `useOpenGL` option here https://github.com/pyqtgraph/pyqtgraph/blob/master/doc/source/config_options.rst

Broadly speaking, and this is _way_ outside the scope of this PR, but we should likely add a separate section in the docs about benefits/consequences of enabling openGL for GraphicsView.

j9ac9k:
Thanks @pijyoi !  I'm not an rst wizard, but can we get ordering/numbering to not be in-line in the rendered document?

https://pyqtgraph--2111.org.readthedocs.build/en/2111/config_options.html

pijyoi:
would need some help here...

j9ac9k:
I'll take a look I'm a bit; once I figure it out would you like me to push to your branch?

pijyoi:
>once I figure it out would you like me to push to your branch?

Yes, please go ahead.

j9ac9k:
welp, I got it, it's ugly as we run into a [known issue](https://github.com/readthedocs/sphinx_rtd_theme/issues/289) with the `sphinx_rtd_theme`

but the end result isn't offensive at least...

image

j9ac9k:
@pijyoi since I requested the documentation bit, and it's added, I'm going to merge as soon as the CI comes back green.  Thanks for the PR!

pijyoi:
Could you open a new issue on this?

Issue #2107, opened 2021-11-17
Jupyter ROIExamples plot truncated when run on binder


pijyoi:
### Short description

This issue is to document an unresolved anomaly. 
The ROIExamples.ipynb plot looks truncated when run on binder with QT_QPA_PLATFORM=offscreen.

### Code to reproduce

Go to the following link
https://mybinder.org/v2/gh/pyqtgraph/pyqtgraph/HEAD?labpath=pyqtgraph%2Fexamples%2Fnotebooks%2FROIExamples.ipynb
and then run the notebook. The plot will seem be truncated on the right.

### Expected behavior

Plot Should not be truncated.

### Real behavior

Plots are truncated.

![roi_truncated](https://user-images.githubusercontent.com/2657027/142130282-952ec020-cb96-4776-abf5-02da2da92eb7.png)

### Tested environment(s)

 * PyQtGraph version: master 
 * Qt Python binding: PyQt5 5.15 
 * Python version: 3.7 (binder)
 * NumPy version: latest from pip 
 * Operating system: Ubuntu 18.04 
 * Installation method: binder requirements.txt 

### Additional context

This could be reproduced on WSL2 Ubuntu 20.04 also using QT_QPA_PLATFORM=offscreen. The truncation appears on PyQt5, PySide2 but not on PyQt6, PySide6.

In https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/jupyter/GraphicsView.py, there is a line:
```self.gfxView.render(painter, self.gfxView.viewRect(), self.gfxView.rect())```.

```self.gfxView.rect()``` specifies the source rect. If I increase the size of that source rect, then more of the plot gets rendered. However the mouse coordinates will be wrong. It seems like internally the correct size of the plot has not been propagated. (Note that the unmodified value of self.gfxView.rect() is the correct size)

**Comment by pijyoi** [comment]
So I think the issue has to do with ```GraphicsLayout``` not being able to shrink items with text in them.
In the following script, the center plot is given as much space as is needed to fit the long title text.
```python
import numpy as np
import pyqtgraph as pg

text = "The quick brown fox jumps over the lazy dog"

pg.mkQApp()
glw = pg.GraphicsLayoutWidget(show=True)
p1 = glw.addPlot(y=np.arange(10))
p2 = glw.addPlot(y=np.arange(10), title=text)
p3 = glw.addPlot(y=np.arange(10))
pg.exec()
```

[...]

PR #2112, opened 2021-11-19
use smaller font size for ROIExamples.ipynb

pijyoi:
This is a quick-and-dirty fix for #2107.

From https://doc.qt.io/qt-5/qapplication.html#setFont:
```On application start-up, the default font depends on the window system. It can vary depending on both the window system version and the locale.```

The default font when running offscreen on PyQt5 is for some reason largish, causing the issue listed in #2107. By changing the font size to something smaller, we don't get the truncation.

There are possibly other ways to achieve the same thing, like using style sheets.


j9ac9k:
Given this is an example, I think a quick and dirty fix here is appropriate. Thanks @pijyoi merging!

PR #2116, opened 2021-11-21
Param tree style, modified

NilsNemitz:
Modified version of #2101 for visible discussion. 
I hope some of this approach can be merged into #2101, and this PR can be deleted. :)

The idea is not to hardcode any colors, but to always pull from the application palette.

This automatically works on Mac for light/dark mode. (Except -right now- it doesn't allow live switching)
Windows is always on a light palette. This also works there.

The most common way to get dark mode on Windows uses the QDarkStyle style sheet. The code detects if the user specifically applies this stylesheet, but did not provide a matching dark palette. Only in this case, the code sets a dark palette that matches the QDarkStyle color scheme.

Big thanks to @Wubbzi for digging into this and getting all the puzzle pieces nicely lined up on the table!

### DIFF
--- a/pyqtgraph/Qt/__init__.py
+++ b/pyqtgraph/Qt/__init__.py
@@ -485,9 +485,10 @@ def mkQApp(name=None, style='system'):
 
         QAPP._setStyleSheet = QAPP.setStyleSheet
         QAPP.setStyleSheet = patched_setStyleSheet.__get__(QAPP.setStyleSheet, QtWidgets.QApplication)
-        if style == 'dark':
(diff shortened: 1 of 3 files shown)


NilsNemitz:
Modified version of #2101 for visible discussion. 
I hope some of this approach can be merged into #2101, and this PR can be deleted. :)

NilsNemitz:
I think all the proposals I wanted to make here have been either absorbed into #2101, or found to be impractical.

I am closing this, and hope that we will sonn have #2101 ready to merge.

PR #2121, opened 2021-11-24
Bugfix: Handle example search failure due to bad regex

ntjess:
This fixes an error message when the user is halfway through typing a regex and a `re.error` is raised. The filter box turns red to indicate an issue and no search is performed until the issue is resolved.

Hoping @NilsNemitz can get the palette working instead of manhandling the stylesheet... I had no luck calling `setPalette` directly.

### DIFF
--- a/pyqtgraph/examples/ExampleApp.py
+++ b/pyqtgraph/examples/ExampleApp.py
@@ -325,6 +325,9 @@ class ExampleLoader(QtWidgets.QMainWindow):
             if self.curListener is not None:
                 self.curListener.disconnect()
             self.curListener = textFil.textChanged
+            # In case the regex was invalid before switching to title search,
+            # ensure the "invalid" color is reset
+            self.ui.exampleFilter.setStyleSheet('')
             if searchType == 'Content Search':
                 self.curListener.connect(self.filterByContent)
             else:
--- a/pyqtgraph/examples/ExampleApp.py
+++ b/pyqtgraph/examples/ExampleApp.py
@@ -372,13 +372,14 @@ class ExampleLoader(QtWidgets.QMainWindow):
         validRegex = True
         try:
             re.compile(text)
-            background = app.palette().background().color()
+            self.ui.exampleFilter.setStyleSheet('')
         except re.error:
             colors = DarkThemeColors if app.property('darkMode') else LightThemeColors
-            background = pg.mkColor(colors.Red)
-            background.setAlpha(100)
+            errorColor = pg.mkColor(colors.Red)
             validRegex = False
-        self.ui.exampleFilter.setStyleSheet(f'background: {pg.mkColor(background).name(QtGui.QColor.HexArgb)}')
+            errorColor.setAlpha(100)
+            # Tuple prints nicely :)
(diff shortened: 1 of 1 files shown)


pyqtgraph/examples/ExampleApp.py
```diff
@@ -369,7 +369,20 @@ def filterByTitle(self, text):
...
+        validRegex = True
+        try:
+            re.compile(text)
+            background = app.palette().background().color()
+        except re.error:
+            colors = DarkThemeColors if app.property('darkMode') else LightThemeColors
+            background = pg.mkColor(colors.Red)
+            background.setAlpha(100)
+            validRegex = False
+        self.ui.exampleFilter.setStyleSheet(f'background: {pg.mkColor(background).name(QtGui.QColor.HexArgb)}')
```
pijyoi, line 383: QtGui.QColor.NameFormat.HexArgb?

NilsNemitz: Does it work to just set 
``` python
try:
    re.compile(text)
    self.ui.exampleFilter.setStyleSheet('')
except:
    colors = DarkThemeColors if app.property('darkMode') else LightThemeColors
    qcol = QtGui.QColor(colors.Red)
    qcol.setAlpha(100)
    hexcol = qcol.name(QtGui.QColor.NameFormat.HexArgb)
    self.ui.exampleFilter.setStyleSheet(f'background: {hexcol})
    return
``` 
?
Going through mkColor twice just to prefix #64 to the string seems like the long way around.

And I think clearing the stylesheet should revert to the original colors, without explicitly setting one. Did not test, though.

j9ac9k: @NilsNemitz beat me to it, but yeah, get the `colors` via accessing `DarkThemeColors` or `LightThemeColors` based on the `darkMode` property.

Since the colors come in string form, I don't think we should even convert to qcolor.

```python
colors = DarkThemeColors if app.property('darkMode') else LightThemeColors
red = colors.Red
self.ui.exampleFilter.setStyleSheet(f'background : {red}00')
```

NilsNemitz: Qt's stylesheet seems to be AARRGGBB, though. Not sure if 
`self.ui.exampleFilter.setStyleSheet(f'background : #64{red[1:]}')` 
is more readable than the detour through QColor.  
:)

j9ac9k: oh yeah, `rgba()` is likely far more readable

```python
colors = DarkThemeColors if app.property('darkMode') else LightThemeColors
error_color = pg.mkColor(colors.Red)
self.ui.exampleFilter.setStyleSheet(f'background : rgba({error_color.red()}, {error_color.green()}, {error_color.blue(), {error_color.alpha()})
```

ntjess:
image
Does anyone else's error box change color with an annoying white outline?

j9ac9k:
Ugh, this PR is 9-10 months old..this is embarrassing.  Thanks for the PR @ntjess LGTM, merging.

PR #2124, opened 2021-11-27
Fix deprecation warnings

pijyoi:
Fixes various warnings that have arisen due to previous code cleanups.


j9ac9k:
Thanks for the tidying up @pijyoi  ...surprised the `pg.time` one snuck by.

PR #2125, opened 2021-11-27
win32 np.clip slowness fixed in numpy

pijyoi:
#1641 added a workaround for ```np.clip``` slowness when running on Windows platform.

The cause was found and fixed in https://github.com/numpy/numpy/pull/20134.
https://github.com/numpy/numpy/pull/20234 is also relevant. (bump compiler to MSVC 2019).

The two above-mentioned PRs are included in numpy 1.22. Thus the workaround can and _should_ be avoided when using numpy >= 1.22. Note that with numpy 1.22, ```np.clip``` now runs equally fast on both Windows and Linux (WSL2) on the same machine, as would be expected. 

Note that the slowness was not a Windows platform issue, it was a code + compiler issue.
For testing the speed difference between different versions of numpy, the script in https://github.com/numpy/numpy/issues/18673 can be used.
Numpy 1.22.0rc1 can currently be downloaded from pypi.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1196,6 +1196,13 @@ def clip_scalar(val, vmin, vmax):
     """ convenience function to avoid using np.clip for scalar values """
     return vmin if val < vmin else vmax if val > vmax else val
 
+# umath.clip was slower than umath.maximum(umath.minimum).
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for sticking w/ this issue @pijyoi !  LGTM, merging!

PR #2132, opened 2021-12-04
Fix PySide6 6.2.2 breaking change

j9ac9k:
This fixes an issue that came up in PySide6 6.2.2 where calls to `GraphicsObject.parentChanged` were passed to `QGraphicsObject.parentChanged` instead of `GraphicsItem.parentChanged`

Looks like we may have other similar issues with:

* `sceneTransform`
* `childItems`
* `setParentItem`
* `parentItem`
* `pos`

but the test suite doesn't appear to have caught those...


j9ac9k:
I'm not convinced that this is the ideal fix; there is something not write to me with having the signal/method conflict via multiple-inheritance.  

Given that `GraphicsItem.parentChanged()` just called `GraphicsItem._updateView()` should we instead do something like

```python
# in GraphicsObject.__init__
self.parentChanged.connect(self._updateView)
```

Thoughts?

pijyoi:
So the following would be the MWE. Python rules say that M is searched before Q.
The puzzling thing is: how does PySide6 6.2.2 manage to subvert the Python resolution machinery?
```python
from PySide6 import QtCore

class Q(QtCore.QObject):
    signal = QtCore.Signal()

    def method(self):
        print('Q::method')

class M:
    def signal(self):
        print('M::signal')

    def method(self):
        print('M::method')

class C(M, Q):
    def __init__(self):
        Q.__init__(self)
        M.__init__(self)

c = C()
c.method()    # okay
c.signal()    # problem on PySide6 6.2.2
```

pijyoi:
Documenting here that calling ```GraphicsItem.parentChanged(self)``` instead of ```self.parentChanged()``` prevents ```ScaleBar::parentChanged``` method from getting called.

pijyoi:
I have a workaround that will successfully call `ScaleBar.parentChanged`.
Change in `GraphicsObject.py`:
```python
           self.parentChanged()
```
to:
```python
            import types
            if isinstance(self.parentChanged, types.MethodType):
                self.parentChanged()
            else:
                # workaround PySide6 6.2.2
                getattr(self.__class__, 'parentChanged')(self)
```

The `else` branch actually works for all bindings, but given that this appears to be a bug of PySide6 6.2.2 which may perhaps get fixed in the future, writing it this way documents the why.

j9ac9k:
Holding off on merging this until I can submit a bug-report with the pyside folks so I can reference the specific issue number; unfortunately I can't set 6.2.2 as the "affected version" for the submission process...hopefully I can do that tomorrow.

j9ac9k:
Created https://bugreports.qt.io/browse/PYSIDE-1730 with @pijyoi 's minimum reproducible example, added a comment to this issue, think this is ready for merging now 👍🏻

pijyoi:
The following is a better test of any eventual fix in PySide6 and reflects pyqtgraph's usage. (`ScaleBar`)
In particular, on PySide6 6.2.2, `print(d.signal)` yields `<PySide6.QtCore.SignalInstance object at ...>`, whereas on PySide6 <= 6.2.1 and other bindings, it yields `<bound method D.signal of ...>'

It seems from the current discussion on https://bugreports.qt.io/browse/PYSIDE-1730 that a fixed version might still yield `<PySide6.QtCore.SignalInstance object at ...>`.

```python
from PySide6 import QtCore

class Q(QtCore.QObject):
    signal = QtCore.Signal()

class M:
    def signal(self):
        # default implementation
        print("M::signal")

class C(M, Q):
    pass

class D(C):
    def signal(self):
        # overridden implementation
        print("D::signal")

c = C()
print(getattr(c.__class__, 'signal'))
print(c.signal)

d = D()
print(getattr(d.__class__, 'signal'))
print(d.signal)

c.signal()
d.signal()
```

PR #2147, opened 2021-12-10
Fire correct signal type for checklist value changing

ntjess:
Not sure how I made it this far without noticing... Forwarding changing signals from checklist children show "true"/"false" as changing values, not self's value. This fixes the problem.

Also, "value" should represent the stored `self.opts['value']` instead of the transient widget value. I fixed this with a separate `childrenValue` function which contains the transient widget value


j9ac9k:
Thanks @ntjess  this looks good to me!

PR #2148, opened 2021-12-11
avoid comparing string with ndarray

### LINKED ISSUE #2027: PlotDataItem with ```connect=ndarray``` raises FutureWarning
### Short description

PlotDataItem with ```connect=ndarray``` raises FutureWarning

### Code to reproduce

```python
import pyqtgraph as pg
import numpy as np
pg.mkQApp()
pw = pg.PlotWidget()
pw.show()
item = pg.PlotDataItem(np.arange(10), connect=np.ones(10, dtype=bool))
pw.addItem(item)
pg.exec()
```

### Real behavior

```
c:\work\github\pyqtgraph\pyqtgraph\graphicsItems\PlotDataItem.py:842: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  if curveArgs['connect'] == 'auto': # auto-switch to indicate non-finite values as interruptions in the curve
```

### Tested environment(s)

 * PyQtGraph version:  0.12.3
 * Python version: 3.8.10
 * NumPy version:  1.20.3
 * Installation method:  pip

### Additional context
No warning occurs if PlotCurveItem is used instead.

pijyoi:
This fixes #2027 by extracting the fix made in #1453 (6624f7e88a08b437021cba664e96354c857a5f3c)
As this bug was introduced in 0.12.3, I figured it would be good to fix it before 0.12.4, even if it is marked as a "good first issue".

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -880,8 +880,8 @@ class PlotDataItem(GraphicsObject):
             self.opts['pen'] is not None 
             or (self.opts['fillBrush'] is not None and self.opts['fillLevel'] is not None)
             ): # draw if visible...
-            # print('connect is', curveArgs['connect'], 'expect nonfinites:', dataset.containsNonfinite)
-            if curveArgs['connect'] == 'auto': # auto-switch to indicate non-finite values as interruptions in the curve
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks @pijyoi forgot about this one!

PR #2149, opened 2021-12-11
Fix automatic int-casting code for Python 3.10 compatibility

JamesWrigley:
Python 3.10 no longer allows automatically converting some floats to ints, see
the fourth point here:
https://docs.python.org/3/whatsnew/3.10.html#other-language-changes

(I see there's already been a bunch of other Python 3.10 fixes too)


j9ac9k:
Hi @JamesWrigley 

Thanks for the PR, reminded me that I need to add Python 3.10 to our CI!

Great call on casting to `QPointF` instead of just `int()` for the `drawText` method.  This LGTM, merging!

PR #2150, opened 2021-12-11
Add Python 3.10 to CI - Re-Enable pytest-xdist in CI

j9ac9k:
This PR adds python 3.10 to our CI suite.  In addition, we use pytest-xdist to speedup our test suite.  Previously we removed pytest-xdist because it made it a bit more difficult to determine which test or example was being problematic in context to segfaults and things like that.   Given our CI has been stable for some time now, I think we can safetly reintroduce pytest-xdist here.

In addition, I remove the automatic upgrade of `wheel` and `setuptools`, as that's not really needed.   I also removed the `python -m pip install --use-feature=in-tree-build .` line as that is now the default, it was generating a warning.

In addition, updating the README accordingly.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -3,9 +3,6 @@ name: codeql
 on: [push, pull_request]
 
 jobs:
-  concurrency: 
-    group: ${{ github.head_ref || github.run_id }}
-    cancel-in-progress: true
   analyze:
     name: analyze
     runs-on: ubuntu-latest
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -3,6 +3,9 @@ name: codeql
 on: [push, pull_request]
 
 jobs:
+  concurrency: 
+    group: ${{ github.head_ref || github.run_id }}
+    cancel-in-progress: true
   analyze:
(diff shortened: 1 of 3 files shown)


j9ac9k:
Well that's concerning, the CSV exporter test crashed, on rerun it ran fine...

j9ac9k:
Just noticed that installing numba results in numpy being downgraded from 1.21.4 to 1.20.3 ...hmm..that's not ideal; should probably re-think how I can do this more robustly...

pijyoi:
PyQt5 5.15 can be pip installable and does run on Python 3.10

j9ac9k:
> PyQt5 5.15 can be pip installable and does run on Python 3.10

Nice catch, fixed!

pijyoi:
There is now a Python 3.10 compatible wheel for PySide2 5.15.2.1.
https://pypi.org/project/PySide2/

j9ac9k:
CI will now cancel existing runs on a new push to the same branch, this should help keep things from getting clogged up when new commits are added before the CI runs finish.

I'm also leaving the `numba` to an optional install for 3.10.  When I originally wrote the diff, `numba` was not available for python 3.10.  It is available now, but if we install `numba` to all pipelines, we won't test the non-numba code-paths.

PR #2151, opened 2021-12-11
console font change

Wubbzi:
change default font to speed up console widget launch


j9ac9k:
Looks like you may not have created your branch from master, let me know if you need help cleaning this up some.

Wubbzi:
The changes made here behave the same as the current master. 
PySide2 5.12+ works on 3.7 
PySide2 5.14+  works on 3.8
PySide2 < 5.14 fails on 3.8

j9ac9k:
Hi @Wubbzi the pyside2 bindings < 5.14 are a bit complicated due to the bug you ran into; that bug looks to have been fixed in the conda-forge distributions so the line above isn't _quite_ accurate (to be fair to you, on our own README we indicate pyside2 5.12 doesn't work w/ python 3.8+)...anyway provides the pyside2 5.15 generated template files work on Qt 5.12 (regardless of python version), that's good enough 👍🏻

Wubbzi:
Ahh, Sorry. I didn't even think about conda.

Let me know if you think we should add a few more pixels of margin to the input lineEdit layout. From the screenshot I saw, it looked like apple has added more border radius to the windows in newer versions of MacOs.

j9ac9k:
Hi @Wubbzi , I cherry-picked some a few of the commits here and applied them to #2435 due to a migration of the template files.  Since we no longer support Qt5 < 5.15, some of the issues are no longer deal breakers for us.  Anyway I'm going to close out this issue and merge that one.  Thanks again for this PR!

PR #2153, opened 2021-12-15
Fix more int-casting code for Python 3.10 compatibility

JamesWrigley:
Same thing as #2149.


j9ac9k:
Thank you @JamesWrigley  this looks good to me!

PR #2154, opened 2021-12-15
Fix typos and formatting errors

JamesWrigley:
Also changed the docs of `TargetItem` to point to `ScatterPlotItem.setSymbol()`
in reference to what symbols are accepted, since the documentation for this
method lists all the symbols.


j9ac9k:
Thanks @JamesWrigley !  Nice catches on all these!

PR #2160, opened 2021-12-22
Add lru cache support to `AxisItem.tickStrings()`

goodboy:
This is still very much a WIP/testing patch but on my current use case
(real time updates to a 40k+ datums set at around display rate) results
in approximately a 2x speedup according to the profiler output included
here.

Interested to see my results are not common to other users.

UPDATE:
This is due to the [block i commented](https://github.com/pyqtgraph/pyqtgraph/pull/2160/commits/66700eb39c742063c6800548ec9fcbc1b338df44#diff-1e50558db426ca1859cf7ebfd1472571d8450c94cad8dd07059b080bc0a4f4fdR1108) in 66700eb39c742063c680 which I'm pretty sure we can eventually eliminate (see comments below).

Also I once in a while can get into an error state where I'll see the following error repeatedly:

```python
TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'
    > Entering PriceAxis.generateDrawSpecs
      init: 0.4193 ms
      compute ticks: 0.2095 ms
    < Exiting PriceAxis.generateDrawSpecs, total time: 108.0801 ms
  File "/home/goodboy/repos/pyqtgraph/pyqtgraph/graphicsItems/AxisItem.py", line 652, in paint
    specs = self.generateDrawSpecs(painter)
  File "/home/goodboy/repos/pyqtgraph/pyqtgraph/graphicsItems/AxisItem.py", line 1174, in generateDrawSpecs
    rect = QtCore.QRectF(tickStop+offset, x-(height/2), width, height)
TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'
```

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -1,3 +1,4 @@
+import functools
 import warnings
 import weakref
 from math import ceil, floor, isfinite, log, log10
@@ -21,7 +22,21 @@ class AxisItem(GraphicsWidget):
     If maxTickLength is negative, ticks point into the plot.
     """
 
-    def __init__(self, orientation, pen=None, textPen=None, linkView=None, parent=None, maxTickLength=-5, showValues=True, text='', units='', unitPrefix='', **args):
+    def __init__(
+        self,
+        orientation,
+        pen=None,
+        textPen=None,
+        linkView=None,
+        parent=None,
+        maxTickLength=-5,
+        showValues=True,
+        text='',
+        units='',
+        unitPrefix='',
+        lru_cache_tick_strings: bool = True,
+        **args,
+    ):
         """
         =============== ===============================================================
         **Arguments:**
@@ -113,8 +128,9 @@ class AxisItem(GraphicsWidget):
             self._linkToView_internal(linkView)
 
         self.grid = False
-        
-        #self.setCacheMode(self.DeviceCoordinateCache)
+
+        if lru_cache_tick_strings:
+            self.tickStrings = functools.lru_cache(maxsize=2**20)(self.tickStrings)
 
     def setStyle(self, **kwds):
         """
@@ -903,7 +919,10 @@ class AxisItem(GraphicsWidget):
         be drawn, then generates from this a set of drawing commands to be
         interpreted by drawPicture().
         """
-        profiler = debug.Profiler()
+        profiler = debug.Profiler(
+            disabled=False,
+            delayed=False,
+        )
         if self.style['tickFont'] is not None:
             p.setFont(self.style['tickFont'])
         bounds = self.mapRectFromParent(self.geometry())
@@ -1034,7 +1053,6 @@ class AxisItem(GraphicsWidget):
                 tickSpecs.append((tickPen, Point(p1), Point(p2)))
         profiler('compute ticks')
 
-
         if self.style['stopAxisAtTick'][0] is True:
             minTickPosition = min(map(min, tickPositions))
             if axis == 0:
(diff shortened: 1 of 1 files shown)


goodboy:
Oh and btw you can see the cache stats using:

```python
print(self.tickStrings.cache_info())
```

goodboy:
I feel like there's some work here to be done to avoid all the `strings: list` mutation stuff especially since it seems using the roundtrip `tuple` **plus** the `lru_cache` is still faster then the old implementation.

j9ac9k:
Hi @goodboy  Thanks for the PR, any opportunities to speed-up `paint` calls are appreciated for sure!

I'll be taking a closer look later, but at first glance, that `maxsize` for the cache is _way_ too large; I would try and do something like `2**8`.

FYI we're getting close to releasing 0.12.4, so I'm going to not try and merge this before that release (given that I've delayed that release due to the holidays).

That None instead of `float` bit is likely due to the screen not being rendered, you can likely have an if-check for checking if it's shown and return/exit early if that's the case....actually, this might be problematic with respect to the caching; the caching should only take effect if the axis is actually being displayed.

pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -903,7 +919,10 @@ def generateDrawSpecs(self, p):
         be drawn, then generates from this a set of drawing commands to be
         interpreted by drawPicture().
         """
-        profiler = debug.Profiler()
+        profiler = debug.Profiler(
```
j9ac9k, line 922: we're not going to want to leave these settings here in the library or else it would print to everyone in the console; a safer alternative for testing might be to try and setup `.env` file with `PYQTGRAPHPROFILE=pyqtgraph.graphicsItems.AxisItem.setStyle` (I think that should work, haven't tested, the docstring for the Profiler has the blurb about setting environment variables).

pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -113,8 +128,9 @@ def __init__(self, orientation, pen=None, textPen=None, linkView=None, parent=No
             self._linkToView_internal(linkView)
 
         self.grid = False
-        
-        #self.setCacheMode(self.DeviceCoordinateCache)
+
+        if lru_cache_tick_strings:
+            self.tickStrings = functools.lru_cache(maxsize=2**20)(self.tickStrings)
```
j9ac9k, line 133: didn't know you could declare it this way without the decorator, neat!

goodboy: Yah that's what `@decorator` is transformed to iirc.

goodboy: Yah, check [the docs](https://docs.python.org/3.10/glossary.html#term-decorator)

pyqtgraph/graphicsItems/AxisItem.py
goodboy: Yah of course. This patch is **no where near meant for landing**.

This is just to get some peeps testing and see if they figure if the cache usage also speeds up their systems.

goodboy:
> I'll be taking a closer look later, but at first glance, that `maxsize` for the cache is _way_ too large; I would try and do something like `2**8`.

Yah I think that's the default. More then happy to tweak whatever settings y'all want.

> FYI we're getting close to releasing 0.12.4, so I'm going to not try and merge this before that release (given that I've delayed that release due to the holidays).

No rush whatsoever; I'd just like to get feedback on this mostly from other low latency users so see if we can work towards maybe even simplifying the tick generation logic since I don't think it's aiding much with speed.

>That None instead of float bit is likely due to the screen not being rendered, you can likely have an if-check for checking if it's shown and return/exit early if that's the case....actually, this might be problematic with respect to the caching; the caching should only take effect if the axis is actually being displayed.

Yes. I think it's actually something to do with the the axis' *in view bounding rect* internal calcs (which yes appear to only generate tick strings if *that part* of the axis is displayed).

> the caching should only take effect if the axis is actually being displayed.

Well there's two cases:
- ticks haven't been generated yet **and** that part of the axis is not in view -> so don't generate the ticks/strings
- those tick strings were already generated and cached from some prior state where they **were in view** in which case the detection logic is fruitless since the values are already cached.

pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -1075,14 +1093,19 @@ def generateDrawSpecs(self, p):
             ## Get the list of strings to display for this level
             if tickStrings is None:
                 spacing, values = tickLevels[i]
-                strings = self.tickStrings(values, self.autoSIPrefixScale * self.scale, spacing)
+
+                strings = list(self.tickStrings(
```
goodboy, line 1097: Yeah so as mentioned in comments, ideally I'd like to strip out the `list` casting since I don't really think we benefit much from the *initial pass* style of *avoiding generating ticks for graphics not in view* that's going on here. 

Further even if we do want to keep that, after the initial generation avoiding re-generating is mostly fruitless (especially with caching) afaict unless someone is expecting a dynamic domain 😂

I dunno though, maybe that's a use case?

goodboy:
Yeah, so far use in "production" has been great and especially so after doing the work for mult-axis stuff in #2162 and https://github.com/pikers/piker/pull/257.

So I'll make the changes suggested above and let's see if we can get this landed in yah?

PR #1359, opened 2020-09-08
Added multi axis plots

0000matteo0000:
Added management for multiple axis on the same plot

**issues and things that are still broken:**
- [x] ~~example~~
- [x] ~~mouse events propagation~~ courtesy of @herodotus77
- [x] ~~prepareForPaint issue fix~~ courtesy of @outofculture
- [x] ~~documentation~~
- [x] ~~code quality~~ courtesy of  @j9ac9k
- [x] ~~lambdas leak when used with self~~ with the help of a workaround found by @pijyoi
- [x] ~~lambda with self leak~~
- [x] ~~axis labels not padded correctly~~
- [x] ~~double autoscale button when using the main plotitem axis~~ 
- [x] ~~broken signals when selecting charts in a different order than creation~~
- [x] ~~initial autorange is not correct~~
- [x] ~~multiple signals received by the charts in the back when interacting with the top level one making them move more than expected~~
- [x] ~~moving axis and then rescaling results in axis not receiving mouse events outside their old rect size (as if they did not rescale)~~

PR #2162, opened 2021-12-28
`PlotItemOverlay`: an alt approach for #1359

goodboy:
This is a competing approach to #1359 for "multi-axis" support.

The main difference is not introducing any new apis or inherited types (`PlotWidget`) and to instead use existing subsystems (with minor internal changes) and a composed `PlotItem` overlay approach which keeps support for custom user defined subtypes (such as `ViewBox`, other `GraphicsObject`, etc.).

---
#### Notes:
- this **is not yet at** feature parity with #1359
  - [x] no `'left'`, `'bottom'` axis layout stacking yet (but this is trivial)
    - prettty sure the [`QGraphicsAnchorLayout`](https://doc.qt.io/qt-5/qgraphicsanchorlayout.html) is the right tool for this job 😉 
  - [x] no label retrieval and layout insertion yet (also trivial)
  - [x] there is a **bug** with click-drag interaction on axes which seems to be something that needs to be fixed in the `ViewBox`-`AxisItem` rats nest of "linked" handlers 😂 
- **yes i know you probably don't like the Python 3.9+ annotations**, we can remove them once we're ready to make the code hard to read and ready to land 😉 

---
#### Improvements over #1359
- each plot item keeps it's own `ViewBox` which means multiple implementations can be use together (or passed in by the user)
- no ad-hoc rerouting of interaction signals (where possible) to some singleton view
- [linkable axes controls](https://github.com/pyqtgraph/pyqtgraph/pull/2162/files#diff-da1deb0e90c1b92d20e3158b950bc6b59cce99510e741d3fd3c768ff436d7d50R83) for either x, y or both

The current issues there last I checked were:
- [ ] initial autorange is not correct
- [ ] multiple signals received by the charts in the back when interacting with the top level one making them move more than expected
  - figured this out! turns out you just have to ignore calls to any handler that was a broadcasted call from some source `ViewBox`, the details of why this is an issue is gets fairly involved:
    - using `ViewBox.linkView()` does it's own set of signal/slot-broadcasting configuration which can result in handlers being called       **twice**, which in this case **is true** since [`ViewBox.linkView()`](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ViewBox/ViewBox.py#L1000) will [**also call `.setXRange()`**](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ViewBox/ViewBox.py#L1098).. 

Afaict both of these are solved in this approach. Minus unimplemented features, the only bug here is the axis click-drag support which seems to be an internal bug to do with signal relaying.

The trick for overlaying view box code was definitely used from #1359 which you'll see commented.

See the [list of out-standings in the module](https://github.com/pyqtgraph/pyqtgraph/pull/2162/files#diff-da1deb0e90c1b92d20e3158b950bc6b59cce99510e741d3fd3c768ff436d7d50R9) for more details.

---
#### Changes to `ViewBox` and `PlotItem`
I will make comments alongside the code below. 
 
 ---
Very interested to get a detailed discussion on this approach 🏄🏼 😂 🥸 
ping @j9ac9k @outofculture @pijyoi

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -13,6 +13,7 @@ import struct
 import sys
 import warnings
 from collections import OrderedDict
+import weakref
 
 import numpy as np
 
@@ -3235,3 +3236,59 @@ class SignalBlock(object):
     def __exit__(self, *args):
         if self.reconnect:
             self.signal.connect(self.slot)
+
+def connect_lambda(bound_signal, self, func, **kwargs):
+    """Convenience function for connecting a function to a signal
+    passing self as argument avoid lambda "leaks".
+    from: `Kovid Goyal <https://riverbankcomputing.com/pipermail/pyqt/2018-July/040604.html>`_
+    Example: connect_lambda(signal, self, lambda self, arg: self.method(arg))
+
+    Saving a rereference to an object inside the
+    object itself or one of it's children creates a
+    reference loop that breaks garbage collection.
+    Method instances (like self.disableAxisAutoRange)
+    store a reference to self.
+    Lambdas and dynamic functions also store references
+    to the objects used inside them.
+    This means that using self in lambdas or dynamic
+    functions and using them as slots will break
+    garbage collection because PySide and PyQt store
+    the slots connected to signals in a way that
+    break garbage collection too.
+
+    Parameters
+    ----------
+    bound_signal : QSignal
+        The signal to connect func to.
+    self : object
+        The object to be passed to the func.
+        It will be wrapped transparently in a weakref
+        to resolve garbage collection issues.
+    func : callable
+        The callable to connect to bound_signal.
+        Should expect at least one parameter: self.
+        The function will be called like this: func(self, *args),
+        args are the parameters passed by the signal.
+    kwargs : dict, optional
+        extra keyword arguments to be passed to the .connect method
+        of the bound_signal.
+
+    Returns
+    -------
+    Connection
+        The newly created Connection QMetaObject
+        returned by the .connect method of the bound_signal.
+    """
+    ref = weakref.ref(self)
+    num_signal_args = func.__code__.co_argcount - 1
+    if num_signal_args < 0:
+        raise TypeError("lambda must take at least one argument")
+
+    def slot(*args):
+        ctx = ref()
+        if ctx is not None:
+            if len(args) != num_signal_args:
+                args = args[:num_signal_args]
+            func(ctx, *args)
+
+    return bound_signal.connect(slot, **kwargs)
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -1,3 +1,4 @@
+import functools
 import warnings
 import weakref
 from math import ceil, floor, isfinite, log, log10
@@ -21,7 +22,21 @@ class AxisItem(GraphicsWidget):
     If maxTickLength is negative, ticks point into the plot.
     """
 
-    def __init__(self, orientation, pen=None, textPen=None, linkView=None, parent=None, maxTickLength=-5, showValues=True, text='', units='', unitPrefix='', **args):
+    def __init__(
+        self,
+        orientation,
+        pen=None,
+        textPen=None,
+        linkView=None,
+        parent=None,
+        maxTickLength=-5,
+        showValues=True,
+        text='',
+        units='',
+        unitPrefix='',
+        lru_cache_tick_strings: bool = True,
+        **args,
+    ):
         """
         =============== ===============================================================
         **Arguments:**
@@ -113,8 +128,9 @@ class AxisItem(GraphicsWidget):
             self._linkToView_internal(linkView)
 
         self.grid = False
-        
-        #self.setCacheMode(self.DeviceCoordinateCache)
+
+        if lru_cache_tick_strings:
+            self.tickStrings = functools.lru_cache(maxsize=2**20)(self.tickStrings)
 
     def setStyle(self, **kwds):
         """
@@ -903,7 +919,10 @@ class AxisItem(GraphicsWidget):
         be drawn, then generates from this a set of drawing commands to be
         interpreted by drawPicture().
         """
-        profiler = debug.Profiler()
+        profiler = debug.Profiler(
+            disabled=False,
+            delayed=False,
+        )
         if self.style['tickFont'] is not None:
             p.setFont(self.style['tickFont'])
         bounds = self.mapRectFromParent(self.geometry())
@@ -1034,7 +1053,6 @@ class AxisItem(GraphicsWidget):
                 tickSpecs.append((tickPen, Point(p1), Point(p2)))
         profiler('compute ticks')
 
-
         if self.style['stopAxisAtTick'][0] is True:
             minTickPosition = min(map(min, tickPositions))
             if axis == 0:
@@ -1075,17 +1093,22 @@ class AxisItem(GraphicsWidget):
             ## Get the list of strings to display for this level
             if tickStrings is None:
                 spacing, values = tickLevels[i]
-                strings = self.tickStrings(values, self.autoSIPrefixScale * self.scale, spacing)
+
+                strings = self.tickStrings(
+                    tuple(values),
+                    self.autoSIPrefixScale * self.scale,
+                    spacing,
+                )
             else:
                 strings = tickStrings[i]
 
             if len(strings) == 0:
                 continue
 
-            ## ignore strings belonging to ticks that were previously ignored
-            for j in range(len(strings)):
-                if tickPositions[i][j] is None:
-                    strings[j] = None
+            # ignore strings belonging to ticks that were previously ignored
+            # for j in range(len(strings)):
+            #     if tickPositions[i][j] is None:
+            #         strings[j] = None
 
             ## Measure density of text; decide whether to draw this level
             rects = []
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -1094,11 +1094,11 @@ class AxisItem(GraphicsWidget):
             if tickStrings is None:
                 spacing, values = tickLevels[i]
 
-                strings = self.tickStrings(
+                strings = list(self.tickStrings(
                     tuple(values),
                     self.autoSIPrefixScale * self.scale,
                     spacing,
-                )
+                ))
             else:
                 strings = tickStrings[i]
 
@@ -1106,9 +1106,9 @@ class AxisItem(GraphicsWidget):
                 continue
 
             # ignore strings belonging to ticks that were previously ignored
-            # for j in range(len(strings)):
-            #     if tickPositions[i][j] is None:
-            #         strings[j] = None
+            for j in range(len(strings)):
+                if tickPositions[i][j] is None:
+                    strings[j] = None
 
             ## Measure density of text; decide whether to draw this level
             rects = []
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -3,6 +3,7 @@ import importlib
 import os
 import warnings
 import weakref
+from typing import Optional
 
 import numpy as np
 
@@ -153,7 +154,7 @@ class PlotItem(GraphicsWidget):
         
         # Initialize axis items
         self.axes = {}
-        self.setAxisItems(axisItems)
+        self.setAxisItems(axisItems) #, add_to_layout=False)
         
         self.titleLabel = LabelItem('', size='11pt', parent=self)
         self.layout.addItem(self.titleLabel, 0, 1)
@@ -284,18 +285,65 @@ class PlotItem(GraphicsWidget):
                 return getattr(self.vb, name)(*args, **kwargs)
             method.__name__ = name
             return method
-        
+
         locals()[m] = _create_method(m)
-        
+
     del _create_method
 
+    def removeAxis(
+        self,
+        name: str,
+        unlink: bool = True,
+
+    ) -> Optional[AxisItem]:
+        """
+        Remove an axis from the contained axis items
+        by ```name: str```.
+
+        This means the axis graphics object will be removed
+        from the ``.layout: QGraphicsGridLayout`` as well as unlinked
+        from the underlying associated ``ViewBox``.
+
+        If the ``unlink: bool`` is set to ``False`` then the axis will
+        stay linked to its view and will only be removed from the
+        layoutonly be removed from the layout.
+
+        If no axis with ``name: str`` is found then this is a noop.
+
+        Return the axis instance that was removed.
+
+        """
+        entry = self.axes.pop(name, None)
+
+        if not entry:
+            return
+
+        axis = entry['item']
+        self.layout.removeItem(axis)
+        axis.scene().removeItem(axis)
+        if unlink:
+            axis.unlinkFromView()
+
+        self.update()
+
+        return axis
+
+    # NOTE: seriously.. y'all
+    # Why do we need to always have all axes created? I don't understand
+    # this at all. Everything seems to work if you just always apply the
+    # set passed to this method **EXCEPT** for some super weird reason
+    # the view box geometry still computes as though the space for the
+    # `'bottom'` axis is always there **UNLESS** you always add that
+    # axis but hide it? Why in tf would this be the case!?!?
     def setAxisItems(
         self,
-        axisItems=None,
+        # XXX: yeah yeah, i know we can't use type annots like this yet.
+        axisItems: Optional[dict[str, AxisItem]] = None,
         add_to_layout: bool = True,
     ):
         """
-        Place axis items as given by `axisItems`. Initializes non-existing axis items.
+        Place axis items as given by `axisItems`. Initializes
+        non-existing axis items.
 
         ==============  ==========================================================================================
         **Arguments:**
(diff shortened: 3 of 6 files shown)


pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -113,8 +128,9 @@ def __init__(self, orientation, pen=None, textPen=None, linkView=None, parent=No
             self._linkToView_internal(linkView)
 
         self.grid = False
-        
-        #self.setCacheMode(self.DeviceCoordinateCache)
+
+        if lru_cache_tick_strings:
```
goodboy, line 132: Ignore this stuff since it's from the history in #2160.

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -284,65 +285,155 @@ def method(self, *args, **kwargs):
...
+        axis.scene().removeItem(axis)
+        if unlink:
+            axis.unlinkFromView()
+
+        self.update()
+
+        return axis
+
+    # NOTE: seriously.. y'all
+    # Why do we need to always have all axes created? I don't understand
```
goodboy, line 332: This is a pretty serious design thing I noticed.
I don't grok at all why all axes must always be created **other then** that when you don't create them the `'bottom'` axis geometry seems to still be used to in the view calculations resulting in `.viewRect()` style calcs to compute wrong?!

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -284,65 +285,155 @@ def method(self, *args, **kwargs):
...
+    # this at all. Everything seems to work if you just always apply the
+    # set passed to this method **EXCEPT** for some super weird reason
+    # the view box geometry still computes as though the space for the
+    # `'bottom'` axis is always there **UNLESS** you always add that
+    # axis but hide it? Why in tf would this be the case!?!?
+    def setAxisItems(
```
goodboy, line 338: The semantics of `set` IMO should be write, not *mutate*. If you want to *add* then we should have a `addAxisItems()`.

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -284,65 +285,155 @@ def method(self, *args, **kwargs):
...
-                oldAxis = self.axes[k]['item']
-                self.layout.removeItem(oldAxis)
-                oldAxis.scene().removeItem(oldAxis)
-                oldAxis.unlinkFromView()
-            
+        # visibleAxes = ['left', 'bottom']
+        # Note that it does not matter that this adds
+        # some values to visibleAxes a second time
+
+        # XXX: uhhh wat^ ..?
```
goodboy, line 365: I don't understand any of these comments.
There's weird defaults for weird defaults, special cases for skipping this axis and that axis.
It's all highly confusing and doesn't represent the semantics of the method name.

pyqtgraph/graphicsItems/ViewBox/ViewBox.py
```diff
@@ -235,6 +252,26 @@ def __init__(self, parent=None, border=None, lockAspect=False, enableMouse=True,
         if name is None:
             self.updateViewLists()
 
+        self.allow_signal_relay: bool = allow_signal_relay
+
+    # TODO: make this a decorator so that sub-types
```
goodboy, line 257: We can probably just call a decorator version of this around all (user-defined) signal handlers right?

pyqtgraph/widgets/_plotitemoverlay.py
```diff
@@ -0,0 +1,256 @@
...
+    separate (non-colliding) axes apply ``ViewBox`` signal broadcasting
+    such that all overlaid items respond to input simultaneously.
+
+    '''
+    def __init__(
+        self,
+        root_plotitem: PlotItem
+    ) -> None:
+        self.root_plotitem: PlotItem = root_plotitem
+        root_plotitem.vb.allow_signal_relay
```
goodboy, line 65: Woops was supposed to be set to `True`.

pyqtgraph/widgets/_plotitemoverlay.py
```diff
@@ -0,0 +1,256 @@
...
+        count = len(self.overlays)
+        assert count
+
+        for name, axis_info in plotitem.axes.items():
+            axis = axis_info['item']
+            axis_view = axis.linkedView()
+
+            # Remove old axis
+            # plotitem.removeAxis(axis, unlink=False)
+            if not axis.isVisible():
```
goodboy, line 153: We should be able to drop this if we make `PlotItem.setAxisItems()` actually be a pure write operation.

pyqtgraph/widgets/_plotitemoverlay.py
```diff
@@ -0,0 +1,256 @@
...
+
+            plotitem.removeAxis(axis, unlink=False)
+
+            # vb.linkView(0, "")
+            # vb.linkView(1, "")
+            # scene = axis.scene()
+            # if scene:
+            #     scene.removeItem(axis)
+
+            # XXX: DON'T unlink it since we the original ``ViewBox``
```
goodboy, line 164: Can all be dropped.

pyqtgraph/widgets/_plotitemoverlay.py
```diff
@@ -0,0 +1,256 @@
...
+            plotitem,
+            lambda plotitem,
+            vb: plotitem.setGeometry(vb.sceneBoundingRect())
+        )
+        # ensure the overlayed view is redrawn on each cycle
+        root.scene().sigPrepareForPaint.connect(vb.prepareForPaint)
+
+        # vb.sigStateChanged.emit(vb)
+
+        # focus state sanity
```
goodboy, line 216: can also all be dropped.

goodboy:
Gah missing some patches from #1359, notably `ViewBox.sigMouseDragged` and `functions.connect_lambda`.

goodboy:
Lol, Ok so apparently the history from #1359 was more extensive then I thought and is needed to make the example work out of the box.

Might try to get this working on master if not will bring in that history.

goodboy:
There it's back to where I expected.

goodboy:
> there is a bug with click-drag interaction on axes which seems to be something that needs to be fixed in the ViewBox-AxisItem rats nest of "linked" handlers

Ahh think I know what this is; we need a relay signal for xaxis adjustments and likely independently as in #1359 as well. Though I'm confused why it didn't work on that history either?

pyqtgraph/functions.py
```diff
@@ -3235,3 +3236,59 @@ def __enter__(self):
     def __exit__(self, *args):
         if self.reconnect:
             self.signal.connect(self.slot)
+
+def connect_lambda(bound_signal, self, func, **kwargs):
```
goodboy, line 3240: Pulled verbatim from original commit on #1359.
Still need to cherry the last couple commits from that branch that adjusted docs iirc.

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -284,65 +285,155 @@ def method(self, *args, **kwargs):
...
-        
+        Remove an axis from the contained axis items
+        by ```name: str```.
+
+        This means the axis graphics object will be removed
+        from the ``.layout: QGraphicsGridLayout`` as well as unlinked
+        from the underlying associated ``ViewBox``.
+
+        If the ``unlink: bool`` is set to ``False`` then the axis will
+        stay linked to its view and will only be removed from the
```
goodboy, line 308: typo: duplicate wording.

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -284,65 +285,155 @@ def method(self, *args, **kwargs):
...
+            ('bottom', (3,1)),
+            ('left', (2,0)),
+            ('right', (2,2))
+        ):
+            if name in self.axes and name in axisItems:
+                # we already have an axis entry for this name
+                # so remove the existing entry.
+                self.removeAxis(name)
+
+            # elif name not in axisItems:
```
goodboy, line 384: I want to see it work like this.

0000matteo0000:
i looked at your code and tested your example, it seems it needs some work still, and it's not quite mature, but i think there is some good intent.
fixing the spaghetti in axisitem and viewbox i agree should be done, but we need to keep in mind that changing the api is probably not an option since it would mean reworking many other classes that depend on them.
i still think #1359's approach is better for the simple reason it almost does not touch the internals of the library except for adding some signals and some minor edits.
the most critical point in simplifying my approach is changing how events and signals are routed,
events some time just call other event-handling methods in other classes instead of emitting signals that are then connected to such methods properly, and some signals do not include information to support multiple axes correctly.
this is why implementing multiple axis was such a pain and still the issues in my pr remain.
i think your approach is a little over-confident and you should not try to remake what i made already and instead you focus on fixing the pipework in the core library so that the widget in my pr does not need the _connect_signals _magic_.
mostly this would mean, as i previously hinted, making each event throw signals, connect those signals to the correct slots, making sure that viewbox and axisitem are actually separate entities.
another reason to implement a multi-plotitem instead of multi-viewbox is that this means all the features in the plotitem are pretty much supported out of the box with minor tweaks, this means legend, auto-scale button, grids, and many other things that are eye candy maybe but needed for shure and currently working in my pr (you surely noticed i almost do not touch the plotitems after creation).
supporting these features fully would mean changing plotitem in such a way so that it supports basically two modes: master and slave, so that the master plotitem will accept other slave plotitems with most of their features delegated to the master one that should then track the state of the slave ones and act accordingly

goodboy:
> i looked at your code and tested your example, it seems it needs some work still, and it's not quite mature, but i think there is some good intent.

Thanks 🏄🏼 !

> i think your approach is a little over-confident and you should not try to remake what i made already and instead you focus on fixing the pipework in the core library so that the widget in my pr does not need the _connect_signals magic.

Yes! I totally agree. I actually have narrowed down the main issue to a couple things:
- `pyqtgraph` basically *modifies* the normal behaviour of `Qt`'s event routing by nearly re-implementing all of [`QGraphicsScene` and event processing semantics](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/GraphicsScene/GraphicsScene.py#L23) 😂 (the original reason for this seems very very legacy). This re-implementation is partially the source of why my overlay/relay approach doesn't work as easily as you'd expect - [`QEvent.accept()` actually short circuits the custom (written in python) event loop in `pyqtgraph`](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/GraphicsScene/GraphicsScene.py#L356) which is not the normal `Qt` processing behaviour IIRC.
- **Further**, all of ^ could be bypassed if we can overlay the view boxes in a given order - currently this is happening in the exact opposite of the desired overlay order (i.e. in my example `'Data 2'` is the last overlay and its `ViewBox` receives events *first* from the event loop...)
  - if we can get the *root* to be the view box that accepts input first then this view box relay approach will work very well.
  - there is some additional odd behavior in the event loop i haven't found the source of where if you re-`.emit()` an event from one from one of the views it is actually cutting the mouse *drag state* short? Super super weird.

> mostly this would mean, as i previously hinted, making each event throw signals, connect those signals to the correct slots, making sure that viewbox and axisitem are actually separate entities.
another reason to implement a multi-plotitem instead of multi-viewbox is that this means all the features in the plotitem are pretty much supported out of the box with minor tweaks, this means legend, auto-scale button, grids, and many other things that are eye candy maybe but needed for shure and currently working in my pr (you surely noticed i almost do not touch the plotitems after creation).

Totally agree. This was exactly my original quiff with the other PR. We already have APIs for all the axis management and graph annotations/labels - we should leverage that and just make the event and layout system work around it.

>  making each event throw signals, connect those signals to the correct slots, making sure that viewbox and axisitem are actually separate entities.

I haven't pushed it yet, but the signals you added for mouse drag and wheel I think should be `mouseDragRelay` and mouseWheelRelay` and these can then be hooked by any widget looking to plug into the view's input handling. In this case it's other views and we can simply add a optional `kwarg` to the `mouseDragEvent` such as `relayed_from: ViewBox` which the handler can use to detect if the signal was sent from some other view (ideally a lone broadcaster / producer).

> supporting these features fully would mean changing plotitem in such a way so that it supports basically two modes: master and slave, so that the master plotitem will accept other slave plotitems with most of their features delegated to the master one that should then track the state of the slave ones and act accordingly.

Yup exactly. I would prefer to call it `focussed` and `overlaid` plot items since there should be one top level widget that broadcasts to the others.

> this changes in conjunction with the, i think, simple to use api of the widget in my pr should make the multi axis feature complete.

Sounds good 😎 . I want to keep working on this anyway just to improve my own knowledge of the `pyqtgraph` code base core and I need it for my own parent project.

I'll keep you posted on progress for sure :)
Also, it might be good if you hopped in the slack room so we can interact in real-time if you want.

goodboy:
Yeah just to clarify the main problem that i'm seeing with the event loop even just to remind myself:
- the scenes containing each view are layered in the view in the opposite desired order such that the last overlayed view is receiving events **first**.
  - no idea yet how to change this, but if we can set a scene order up front then this whole event loop issue (to be described below) should go away
- when the last overlayed view receives the first event, if it `MouseDragEven.ignore()`'s it, that will short circuilt the processing *for that view*. It also seems that the event loop keeps memory of that and won't re-fire that same event to that view due to the some kind of either "drag items" state tracking or bad event loop logic (though I don't want to claim too much yet until I've stepped through it proper).

0000matteo0000:
look at my pr, i fixed the "first receiving the event" issues by setting the z index for the viewbox i wanted in front

goodboy:
> look at my pr, i fixed the "first receiving the event" issues by setting the z index for the viewbox i wanted in front

Oh nice. Yeah funny enough I solved the same thing yesterday aft 😂 

It's all working now on this branch.
Afaict all the zooming and broadcasting stuff is fixed and i've factored it into a helper func which I'd like to transform into a class method decorator to get a very non-invasive event broadcast solution 😎 

Give this latest commit a shot and let me know if the interaction is as you'd expect.

goodboy:
> i still think #1359's approach is better for the simple reason it almost does not touch the internals of the library except for adding some signals and some minor edits.

this PR will actually touch less internals if I am able to do the decorator API I'd like.
Further, I'm starting to think the overlay layout should be distinct from `PlotItem.layout` potentially.
I actually think that `PlotItem` itself should have more decoupling of *underlying graphics components* from *layout of those graphics* it uses.

pijyoi:
I noticed that many of the existing examples have lost their axis display on this PR.

goodboy:
> I noticed that many of the existing examples have lost their axis display on this PR.

Yeah haven't touched any existing tests or examples yet; just been trying to get base functionality to work.
This is likely due to the `PlotItem.setAxisItems()` tweaks which, frankly, I think it's wild so much relies on such strange implicit behavior.

Can you maybe point out some simple ones I can get started on to begin to approach that method's changes.
I may even break that bit out to a new PR since it's no longer required in this now anyway other then the new `.removeAxis()` which shouldn't have an effect.

goodboy:
@pijyoi ahh also click drags will be broken for now since the view box has been modified to be broadcast only in this PR.

I need to figure out a way to enable that and then disable it by default.
I was actually thinking this whole relay system might be better suited on `GraphicsItem` since it operates on those APIs.

goodboy:
Actually so missing axes is actually due to removal of the `visibleAxes = ['left', 'bottom']` default thing inside `.setAxisItems()`, which I've now fixed and will push shortly.

goodboy:
Lol, ok so I added a method `@eventrelay` [a few commits ago](https://github.com/pyqtgraph/pyqtgraph/pull/2162/commits/64beeaf52ff4531c4e2ce0bb7575f8a12e87a904#diff-39b99d73a5dd4c001f9fe2d6c3e6276ad2bcec3f5493ea65873450be51bed7cbR80) which does work for a pseudo-non-invasive approach for all this but, obviously it requires a permanent change to `ViewBox` methods.

Then I stripped it back to a method overriding system which actually **can't** work with explanation as follows:

- The main problem still remains that that `MouseDragEvent`/`MouseClickEvent`/`MouseHoverEvent` all have this `.accept()` which seems to be **required** in (at least `ViewBox.mouseDragEvent()`) handlers to halt the event loop in such a way that *sibling* widgets will not receive the event after called.
  - I don't know why this decision was made since [Qt developers should really never need to worry about calling `.accept()` by default](https://doc.qt.io/archives/qq/qq11-events.html#acceptorignore):
    - https://doc.qt.io/qt-5/eventsandfilters.html
    - https://doc.qt.io/archives/qq/qq11-events.html#acceptorignore
    - https://doc.qt.io/qt-5/qcoreapplication.html#notify
  - **even weirder** is that if the `.mouseDragEvent()` handler **does not call** `.accept()` then it breaks our click-drag interaction - for example, it will halt a panning interaction short after an initial short movement!?
    - Can i presume *this is a bug*?
    - this seems to be fixed by moving `ev.accept()` into the left button [panning branch](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ViewBox/ViewBox.py#L1309) but still does not seem like the correct underlying solution to me
  
---
#### An even further alternative approach: don't use `Signal.emit()` and method overloading
After doing some digging into the constraints of the signal/slots system and how it really doesn't bring any benefit to python code (verified this with Qt for Python core dev), **and** the fact that `Signal()` must be declared on `GraphicsObject` *before* class construction (since it's construction is done as part of meta-class machinery), I'm tempted to propose that to get our own version of *event broadcasting* we should just roll our own method call loop.

Premises:
- defining custom `mouseBlahEventRelay` attrs is cumbersome and involves modifying the target widget definitions
- using `Signal.emit()` can be tricky (in terms of signature matching) and it the exact same as calling python functions in a loop (except Qt does it for you)
- we're not using the standard event system anyway

Doing our own broadcast loop would make this method overriding / monkey patching approach easier to accomplish since it can be applied over any handler set and will mostly avoid core changes to the handlers themselves.

0000matteo0000:
i think you are really over-complicating things, why not just changing the event handlers to simple signal throwers and then connecting those signals to a almost-copy-paste version of the handler and all the objects that should receive that event?
signals are fine imho they abstract a lot of the code, just standardize the emission of the emitting object or it's name and the signal name to make it easy to manage in handlers and i think that's all that's needed.

goodboy:
> i think you are really over-complicating things, why not just changing the event handlers to simple signal throwers and then connecting those signals to a almost-copy-paste version of the handler and all the objects that should receive that event?

Well that's what I **did do** 😂 

> signals are fine imho they abstract a lot of the code, just standardize the emission of the emitting object or it's name and the signal name to make it easy to manage in handlers and i think that's all that's needed.

This is already how it works as of the code right now?

I'm just proposing an additional approach of avoiding signals all together (for relaying) since they don't really add anything from a Python perspective and possibly even make things more complicated then necessary. All we really need is a fan-out method call into overlayed widgets, plugging into this correctly will require an API like `PlotItemOverlay` anyway so why bother with adding extra signals that are superfluous.

Interested to see what others think.

goodboy:
Oh right and alternatively if we did want to *hack around* requiring `Signal` to be defined at class compile time we can use [this recipe for `Signal` implemented as python descriptor](https://github.com/epcpower/stlib/blob/b09fc9ad3940d049c78d4dc6c69cabc51a400d84/epyqlib/utils/qt.py#L1194-L1232), thought as mentioned, I'm much more of a fan of just avoiding the whole system.

pijyoi:
>Oh right and alternatively if we did want to hack around requiring Signal to be defined at class compile time we can use this recipe for Signal implemented as python descriptor.

Note that the file that you linked to has the following license:
```
__copyright__ = "Copyright 2017, EPC Power Corp."
__license__ = "GPLv2+"
```
https://github.com/epcpower/stlib/blob/master/COPYING

goodboy:
> Note that the file that you linked to has the following license:

Yeah I'm honestly not super fond of this approach - I think we can greatly simplify by simply doing a method call loop for all overlays.

Using a `Signal` only makes it harder to understand the implementation (which is **very adhoc**) and I can't think of another reason besides overlaying plots for a relay style signal we'd want expose as part of the public api? Also, we can always add the signals later if needed.

goodboy:
BTW y'all I would say this is ready for experimental testing though probably a small PR to update the `ViewBox` default handlers is going to be needed before this is ready to land.

I would like to get preliminary feedback though before pushing hard on all that.
For me, this works as expected in terms of the absolute minimal feature set (not including what's outstanding above).

pyqtgraph/functions.py
goodboy: Actually, so now having tested this a bunch in a parent project, I'm pretty sure it's not needed?

@0000matteo0000 what was the original leak you found which motivated using this?

goodboy:
Ok I'm giving y'all until end of week to give feedback on this approach or I'm closing it and AGPL-ing the recent improvements we've made and moving them out into a client project.

I realize this stuff is probably useful (and actually the new stuff is very generalized) but the project that requires it doesn't have much bandwidth for worrying about non-modern python requirements.

If and when there's a want for this more then happy to re-license and bring it back whenever other's are interested 😎

j9ac9k:
> Ok I'm giving y'all until end of week to give feedback on this approach or I'm closing it and AGPL-ing the recent improvements we've made and moving them out into a client project.

Hi @goodboy.  First off, thank you for the PR with the alternative proposal.  This PR does address one issue that #1359 has not been able to address to date, and we will be examining for differences, hoping to adopt what makes zoom behavior work here that #1359 has missed.

I do feel the need to respond to the timeline you've imposed on us. I should start by stating the obvious; pyqtgraph maintainers are volunteers, none of us are paid to work on this library. I believe all active maintainers are all currently employed or in a full time educational program. Many of us also have responsibilities at home on top of that.  With the most recent surge in covid cases; the entirety of my household have contracted covid, despite all of us of age being vaccinated. I share this to illustrate that while I recognize pyqtgraph development is important, it isn't anywhere near the top priority for me.

I recognize pyqtgraph development can be slow, #1359 is well over a year old now! Trying to impose deadlines on maintainers for code-reviews/feedback on your contributions is unlikely to get the reaction you are hoping for.  I request you please make an effort to not try and exert extraneous pressure on us.

goodboy:
@j9ac9k it's not so much a *deadline* as an alternative way to move forward with development of this feature (set).

>  Trying to impose deadlines on maintainers for code-reviews/feedback on your contributions is unlikely to get the reaction you are hoping for.

It's not really "imposing" as much as it is a *if, then* about my plan with this feature set going forward.
I apologize if it came off that way, but it was not my intention.

Further the way I see this approach for newer features is actually less stressful for maintainers since,
- there is no downside to `pyqtgraph` we will gladly upstream the changes whenever they are desired and are able to be scrutinized
- i can move the feature support into an issue that will track the client project and core devs and users can give it attention whenever they decide it's more important 😃 

The client project has its own demands (such as pleasing user feature requests, staying competitive with alternative platforms, not requiring support for non-modern Python as a speed bump, etc..) and so I'm sure you can understand that while `pyqtgraph` might not have all those same demands this project **does** which is a similar form of "pressure" that needs to be addressed.

I think what I've proposed is actually pretty *flexible*; we move forward with this feature in another project and when users of `pyqtgraph` more generally decide they **need** this feature, we work together through up-streaming it.

Alternatively if there's isn't core dev leeway maybe it's worth considering a set of experimental features to start introducing to the project that are opt-in where contributors like myself can have more of an ability to move things forward? As an example working with the code base in many parts is difficult not only due to poor legacy formatting of docs and code but also the lack of type annotations for tracking the many different layers that exist in the heavy type-inheritance style design. I personally have my own set of rolling patches just to make parts of many 2d components readable when I have to understand how they work again and again 😉 

I'm not trying to put undue pressure on any core devs whatsoever, I'm just trying to get *any* feedback and ideally keep development momentum high in the project that needs this feature. I'm sure you can understand that. It is not meant to cause stress or be an ultimatum it's just my plan for how to move forward I'm trying to be up front about it. I'm sure you already know I've done this prior with many "high performance" graphics needs already 😉 (for which I've created multiple issues and consequently implemented the features in the client project).

> despite all of us of age being vaccinated. 

On that note, I hope you and your family recover quickly and are well soon 🙏🏼 
If it helps with encouragement I actually wrote this original patch set while I had covid over the end of last year so I understand the annoyance of such a situation 😸

goodboy:
I ended up taking a slightly different approach with the label stacking (since I personally think the defaults are far too big and invasive) but for all other intents and purposes this is working as expected minus one outstanding issue: when you link all the axes panning via click-drag results in `N` times the movement per mouse input move. I haven't quite figured out the source of this issue but it's likely due to recursive calls in the `ViewBox` mouse handler that somehow aren't filtered correctly.

For those interested the first draft implementation is in https://github.com/pikers/piker/pull/257.
I will keep this open because the example is likely a key driver to getting this brought back into this repo but for now we will not be working on this branch as the client project it is needed for is much faster moving with fewer legacy Python support constraints.

Please feel free to reach out when ready and we'll do our best to resume this 😄

pyqtgraph/functions.py
0000matteo0000: storing self in a lambda or in any object and then storing that object in self (`self.var = {"self": self}`) will break python garbage collection. this is how i test it:
`

import weakref
import gc

import numpy as np
import pyqtgraph as pg
app = pg.mkQApp()
pg.setConfigOption('background', 'w')
pg.setConfigOption('foreground', 'k')

mpw = pg.MultiAxisPlotWidget()

# LEGEND
mpw.addLegend(offset=(0, 0))
# TITLE
mpw.setTitle('MultiAxisPlotWidget Example')
# AXYS
mpw.addAxis('sx1', 'bottom', text='Samples1', units='sx1')
mpw.addAxis('sx2', 'bottom', text='Samples2', units='sx2')
mpw.addAxis('sx3', 'bottom', text='Samples3', units='sx3')
mpw.addAxis('sy1', 'left', text='Data1', units='sy1')
mpw.addAxis('sy2', 'left', text='Data2', units='sy2')
mpw.addAxis('sy3', 'left', text='Data3', units='sy3')
# CHARTS
mpw.addChart('Dataset 1', xAxisName='sx1', yAxisName='sy1')[0].setData(np.array(np.sin(np.linspace(0, 2 * np.pi, num=1000))))
mpw.addChart('Dataset 2', xAxisName='sx2', yAxisName='sy1')[0].setData(np.array(np.sin(np.linspace(0, 2 * np.pi, num=1000))) * 2)
mpw.addChart('Dataset 3', xAxisName='sx2', yAxisName='sy2')[0].setData(np.array(np.sin(np.linspace(0, 4 * np.pi, num=500))) * 3)
mpw.addChart('Dataset 4', xAxisName='sx3', yAxisName='sy3')[0].setData(np.array(np.sin(np.linspace(0, 4 * np.pi, num=500))) * 3)
# make and display chart
mpw.makeLayout()
mpw.update()
mpw.show()
pg.exec()

refs = [
    weakref.ref(mpw),
    weakref.ref(mpw.pi),
    weakref.ref(mpw.vb),
    weakref.ref(mpw.layout),
    weakref.ref(mpw.pi.legend),
    weakref.ref(mpw.charts['Dataset 1']),
    weakref.ref(mpw.axes['sx1']),
]
del mpw
for ref in refs:
    print(ref(), 'should be None')
gc.collect()
for ref in refs:
    print(ref(), 'should be None')
`

goodboy:
As an update, I've reworked this original code to discard all signal/slot rewiring to instead use a python broadcast loop in https://github.com/pikers/piker/pull/414.

This turned out to be vastly easier to understand and implement and appears to have zero performance hit.
The new implementation solves all prior interaction *quirks*/bugs and clears up the differences between axis linking versus event handler broadcasting allowing for much finer grained controls over pub-sub mechanics particularly when an event handler itself fires signals.

PR #2168, opened 2022-01-06
lazy create the rectangle selection item

danielhrisca:
In the current implementation for each ViewBox there will be a rectangle selection item. If the mouse mode is set to PanMode then this rectangle is unused.

With this PR the rectangle is only created if it is really needed.

### DIFF
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -205,6 +205,7 @@ class ViewBox(GraphicsWidget):
         self.borderRect.setPen(self.border)
 
         self.reScaleBox = None
+        self._add_rectangle_selection()
 
         ## show target rect for debugging
         self.target = QtWidgets.QGraphicsRectItem(0, 0, 1, 1)
@@ -347,7 +348,12 @@ class ViewBox(GraphicsWidget):
         if mode not in [ViewBox.PanMode, ViewBox.RectMode]:
             raise Exception("Mode must be ViewBox.PanMode or ViewBox.RectMode")
         self.state['mouseMode'] = mode
-        if self.reScaleBox is None and mode == ViewBox.RectMode:
+        self._add_rectangle_selection()
+
+        self.sigStateChanged.emit(self)
+
+    def _add_rectangle_selection(self):
+        if self.reScaleBox is None and self.state['mouseMode'] == ViewBox.RectMode:
             ## Make scale box that is shown when dragging on the view
             self.rbScaleBox = QtWidgets.QGraphicsRectItem(0, 0, 1, 1)
             self.rbScaleBox.setPen(fn.mkPen((255, 255, 100), width=1))
@@ -355,8 +361,6 @@ class ViewBox(GraphicsWidget):
             self.rbScaleBox.setZValue(1e9)
             self.rbScaleBox.hide()
             self.addItem(self.rbScaleBox, ignoreBounds=True)
-
-        self.sigStateChanged.emit(self)
 
     def setLeftButtonAction(self, mode='rect'):  ## for backward compatibility
         if mode.lower() == 'rect':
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/ViewBox/ViewBox.py
```diff
@@ -353,8 +348,20 @@ def setMouseMode(self, mode):
         if mode not in [ViewBox.PanMode, ViewBox.RectMode]:
             raise Exception("Mode must be ViewBox.PanMode or ViewBox.RectMode")
         self.state['mouseMode'] = mode
+        self._add_rectangle_selection()
+
         self.sigStateChanged.emit(self)
 
+    def _add_rectangle_selection(self):
+        if self.reScaleBox is None and self.state['mouseMode'] == ViewBox.RectMode:
```
pijyoi, line 356: `self.reScaleBox` should be a typo for `self.rbScaleBox`?
As it is, a new `QGraphicsRectItem` is instantiated each time `mouseMode` changes to `RectMode`

danielhrisca: Yes, that was a typo 🤦‍♂️

j9ac9k:
Hi @danielhrisca thanks for the PR.

Only comment I have is should that hidden method remove the rectangular selection box if setMouseMode is set to PanMode, after initially being in RectMode.  This PR I don't think will attempt to remove the selection rectangle in that condition.

danielhrisca:
> Hi @danielhrisca thanks for the PR.
> 
> Only comment I have is should that hidden method remove the rectangular selection box if setMouseMode is set to PanMode, after initially being in RectMode. This PR I don't think will attempt to remove the selection rectangle in that condition.

My idea was that the CPU cycles were already used to create it, so removing it would only save some memory. Do you think it is better to remove it if the mouse mode is set to PanMode?

j9ac9k:
> My idea was that the CPU cycles were already used to create it, so removing it would only save some memory. Do you think it is better to remove it if the mouse mode is set to PanMode?

We're talking few cycles and little memory here, so I don't think it makes a noticeable difference either way.  I remember for your use-case you overlap many viewboxes, so i'm inclined to follow your lead here... removing the scale box would only occur on changing of the mouse mode, which I suspect is a time we have some cycles to spare.  Either way, not going to hold up this PR on this, was just throwing out the idea is all 👍🏻

j9ac9k:
Blerg, should have merged this long ago, sorry for the wait @danielhrisca Thanks for the PR.

danielhrisca:
Absolutely no problem 👍

j9ac9k:
> 

FYI PyQt6 6.3.1 just introduced a new `sip.array` type which pyqtgraph master branch makes use of.  It allows for noticeable faster ScatterPlots's and much faster calls to `drawLines`; and should be zero-copy operations for minimizing memory usage.  I know you've started rolling out your own drawing mechanism, but if you're using `QPainter.drawPixmapFragments` or `QPainter.drawLines` and are using newer versions of PyQt6, you should absolutely take a look at our implementation.

Also if you want to get ahold of us on not the issue tracker, we now have a discord channel on the main python discord server here: https://discord.com/channels/267624335836053506/898139460821192724

danielhrisca:
Thanks for the tip. In my code I use `drawPoints` with a circular cap for the scatter plots, and `drawPath` with a cached version of `generatePath` https://github.com/danielhrisca/asammdf/blob/18c71009ef8b4784a55a560675eec66911cb1419/asammdf/gui/widgets/plot.py#L4589

In my work project we also use `drawPolygon`

j9ac9k:
Now that #2324 is merged, if you set `enableExperimental` to `True`, it's possible `drawPainterPath` will outperform `drawPolygon` now.   While the `sip.array` object will work w/ `drawPoints`, the `drawPoints` method accepts a `QPolygonF` object which are are already able to construct very efficiently, so probably not _much_ benefit in trying to use `sip.array` there with pyqtgraph.

PR #2169, opened 2022-01-09
Introduce EasyWidget

ntjess:
The goal is to reduce the barrier to entry for creating nice applications *quickly*. I feel this is best demonstrated by example. With the PR, a comparison of two plots with a splitter between can be created in just a few lines of code:

```
from pyqtgraph.widgets.EasyWidget import EasyWidget as ew
from pyqtgraph.Qt import QtWidgets, QtCore, QtGui
import pyqtgraph as pg

pg.mkQApp()
pw = pg.PlotWidget()
lbl = QtWidgets.QLabel('hello world')
lbl.setAlignment(QtCore.Qt.AlignmentFlag.AlignHCenter)
pw2 = pg.PlotWidget()

win = ew.buildMainWin([lbl, ew([pw, pw2], splitter=True)])

win.show()
pg.exec()
```

Obviously, if people are receptive to this PR I will create much more documentation and more examples.

Currently, the entry points are `EasyWidget.buildMainWin`, `EasyWidget.buildWidget`, and `EasyWidget.fromParameter`. The last creates a label and widget from a `paramtertree.Parameter`, which I find useful when I want to use bound widgets outside a tree context and still have `sigValueChanging`/etc. coupled signals.

This avoids the need for creating a main win, contents widget, layout, splitter, adding widgets to the splitter, adding the splitter to the layout, setting the central widget, and so on. For me, that was relatively confusing to get the hang of as a beginner.

## Making your own examples
Every list nesting level changes from a horizontal to vertical layout. You can explicitly force a layout by wrapping the list inside an `EasyWidget` (like the sample code above) and supplying a layout before calling `buildMainWin` or `buildWidget`. Similarly, wrapping with `EasyWidget` can introduce splitters at arbitrary child levels.

### DIFF
--- a/pyqtgraph/__init__.py
+++ b/pyqtgraph/__init__.py
@@ -261,6 +261,7 @@ from .widgets.ComboBox import *
 from .widgets.DataFilterWidget import *
 from .widgets.DataTreeWidget import *
 from .widgets.DiffTreeWidget import *
+from .widgets.EasyWidget import EasyWidget
 from .widgets.FeedbackButton import *
 from .widgets.FileDialog import *
 from .widgets.GradientWidget import *
--- a/pyqtgraph/widgets/EasyWidget.py
+++ b/pyqtgraph/widgets/EasyWidget.py
@@ -173,7 +173,7 @@ class EasyWidget:
         return retWidget
 
     @classmethod
-    def fromPgParam(cls, param: Parameter=None, layout='H', **opts):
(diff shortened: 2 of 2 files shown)


ntjess:
Also note that if this PR receives any traction, it should be good grounds for eliminating many `uic` files. Rather, the windows can be created quite easily using the EasyWidget paradigm. However, I didn't want to perform these operations if there were API changes requested or if EasyWidget was not favored.

PR #2170, opened 2022-01-10
Fix formatting on minimum value of GradientLegend.

CanisUrsa:
The GradientLegend has a bug with the minimum label being divided by 10 and converted from int to float.

### DIFF
--- a/pyqtgraph/graphicsItems/GradientLegend.py
+++ b/pyqtgraph/graphicsItems/GradientLegend.py
@@ -42,7 +42,7 @@ class GradientLegend(UIGraphicsItem):
             g.setColorAt(x, colors[i])
         self.setGradient(g)
         if 'labels' not in kargs:
-            self.setLabels({str(minVal/10.): 0, str(maxVal): 1})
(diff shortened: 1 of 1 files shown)


pijyoi:
While testing it out, another issue was found. The code to `GradientLegend.setIntColorScale` suggests that a `labels` argument could be supplied, but in fact an exception will be thrown.

```python
import pyqtgraph as pg
pg.mkQApp()
pw = pg.PlotWidget()
pw.show()
gl1 = pg.GradientLegend([25, 300], [20, -30]) 
gl1.setIntColorScale(10, 20, labels=['A', 'B'])
pw.addItem(gl1)
pg.exec()
```

It's not clear to me though how `setIntColorScale` should be used. The method generates its own colors, which seems backwards. How would the user get back those colors to false-color their image?

campagnola:
Thanks @CanisUrsa!
@pijyoi you are correct, this code is a mess. Seems like setIntColorScale should just generate an appropriate color map, set it with setColorMap, and then return it.

PR #2172, opened 2022-01-12
Added projects to Used by list

jakimowb:
This PR adds two projects which are very happy that pyqtgraph exists


sem-geologist:
As time-to-time user of QGis, despite I had never had a need  of dealing with hyperspectral earth maps, I still should tell that these two projects looks so awesome!

PR #2181, opened 2022-01-25
Used the power of `blockIfUnchanged` decorator

StSav012:
If the signals are blocked externally, keep the state. Before the patch, the block got released unconditionally. There has been a `blockIfUnchanged` decorator for this, but its code has been re-used with a logical error, so the decorator (although used) couldn't have worked.

### DIFF
--- a/pyqtgraph/widgets/ComboBox.py
+++ b/pyqtgraph/widgets/ComboBox.py
@@ -106,19 +106,9 @@ class ComboBox(QtWidgets.QComboBox):
         If a dict is given, then the keys are used to populate the combo box
         and the values will be used for both value() and setValue().
         """
-        prevVal = self.value()
-        
-        self.blockSignals(True)
-        try:
-            self.clear()
-            self.addItems(items)
-        finally:
-            self.blockSignals(False)
-            
-        # only emit if we were not able to re-set the original value
-        if self.value() != prevVal:
-            self.currentIndexChanged.emit(self.currentIndex())
-        
+        self.clear()
+        self.addItems(items)
+
     def items(self):
         return self.items.copy()
         


pijyoi:
An MWE to test out this PR:
```python
import pyqtgraph as pg

pg.mkQApp()

class MyComboBox(pg.ComboBox):
    def indexChanged(self, index):
        print('indexChanged', index)
        super().indexChanged(index)

cb = MyComboBox()
print('setting A')
cb.setItems(['A'])
print('setting A')
cb.setItems(['A'])
print('setting B')
cb.setItems(['B'])
```

Before this PR,
1) the signal gets fired _twice_ when items have changed. (BAD)
2) no signal gets fired when no change in items. (OK)
```
setting A
indexChanged 0
indexChanged 0
setting A
setting B
indexChanged 0
indexChanged 0
```

With this PR,
1) the signal gets fired _once_ when items have changed. (GOOD)
2) no signal gets fired when no change in items. (OK)
```
setting A
indexChanged 0
setting A
setting B
indexChanged 0
```

StSav012:
@pijyoi, does your comment prove my commit useful? If so, I expect it to be merged. If not, please explain what's wrong with it.

pijyoi:
I am only a user of the library like yourself and can't do any merging. In fact, I don't even use pyqtgraph's ComboBox.

That said, the maintainers of this library are doing so on their own free time, so please be patient.

j9ac9k:
Hi @StSav012 

Sorry that I missed this earlier; these extra signals firing are a constant thorn in my side, and while I don't use pyqtgraph's combo box in my work, I'm glad this part is improving.  Merging.

PR #2184, opened 2022-01-27
[Bugfix] Update import in parametertree.py example

Tillsten:
Change relative import to absolute import. The relative import currently breaks running the example in the runner.


j9ac9k:
Hi @Tillsten 

Thanks for the PR, I think this was already fixed in #2087 and the fix will be part of the next release.

Tillsten:
Yep, sorry for the noise.

j9ac9k:
> Yep, sorry for the noise.

Not a problem, appreciate the PR none-the-less, it lets us know people care.  I know we're delayed with 0.12.4 being merged, I've been hoping to get another PR into the library first but if it doesn't look like it's happening I should try and release 0.12.4 with this fix. 

Thanks again for the PR 👍🏻

Issue #2178, opened 2022-01-18
PR #2011 + anti-aliasing enabled leads to line artifacts


swvanbuuren:
### Short description

PR #2011 introduced an optimized way to plot thick lines. However, when anti-aliasing is enabled, this
leads to line artifacts.

### Code to reproduce
``` python
import pyqtgraph as pg
pg.setConfigOptions(antialias=True)
pg.setConfigOption('background', 'w')
pg.setConfigOption('foreground', 'k')
import pyqtgraph.Qt.QtWidgets as QtWidgets
import numpy as np

win = pg.GraphicsLayoutWidget(show=True)
win.resize(800,350)
win.setWindowTitle('pyqtgraph example: #2011 test')
plt1 = win.addPlot(title='With anti-aliasing and alpha = 1')
plt2 = win.addPlot(title='With anti-aliasing and alpha != 1')

x = np.linspace(0, 2*np.pi, 25)
y = np.sin(x)

WIDTH = 4.0
color = (0, 113, 188)
pen = pg.mkPen(color + (255,), width=WIDTH)
pen_alpha = pg.mkPen(color + (254,), width=WIDTH)

plt1.plot(x, y, pen=pen)
plt2.plot(x, y, pen=pen_alpha)

QtWidgets.QApplication.instance().exec_()
```

### Expected behavior
Fallback to the old drawing method whenever anti-aliasing is enabled (just as is the case if alpha < 1).

### Real behavior
Line artifacts are seen for thicker lines.

### Tested environment(s)

 * PyQtGraph version: commit Id 8436457 and newer
 * Qt Python binding: PySide2 5.15.1 Qt 5.15.1
 * Python version: 3.8.10
 * NumPy version: 1.19.2
 * Operating system: XUbuntu 20.04.2 LTS
 * Installation method: pip

pijyoi:
Could you post some images to show the extent of the artifacts?
What happens if you set WIDTH to 5.0?

swvanbuuren:
Here are some screenshots.

WIDTH = 2.0
![width2](https://user-images.githubusercontent.com/5203275/150201687-86ae3392-a087-4e57-ab6a-8de878aa4d78.png)

WIDTH = 4.0
![width4](https://user-images.githubusercontent.com/5203275/150201720-d26591db-697d-414c-a756-3608661936ef.png)

WIDTH = 5.0
[...]

PR #2185, opened 2022-01-27
Introduce API option to control whether lines are drawn as segmented lines

swvanbuuren:
This PR proposes an API interface for the segmented line drawing method introduced in  PR #2011 (as discussed in issue #2178).

The commit features the following:
- Adds a global options to set mode for drawing segmented lines
- Mode can be 'auto' (use existing method),  'on' (always draw segmented lines), 'off' (never draw segmented lines)
- Global option can be overridden for given ``PlotCurveItem`` instance using setter method
- **[optional]** Added activation of anti-aliasing as criterion to decide whether to draw segmented lines in 'auto' mode (specifically [these lines](https://github.com/swvanbuuren/pyqtgraph/blob/segmented-lines-api-option/pyqtgraph/graphicsItems/PlotCurveItem.py#L669-L671))

I would like to emphasize that the last point is still under debate and can be left out if required.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -637,12 +637,12 @@ class PlotCurveItem(GraphicsObject):
                ``'auto'`` (default) segmented lines are drawn if the pen's width > 1, pen style is a
                solid line, the pen color is opaque and anti-aliasing is not enabled.
 
-               ``'on'`` or ``True`` lines are always drawn as segmented lines
+               ``'on'`` lines are always drawn as segmented lines
 
-               ``'off'`` or ``False`` lines are never drawn as segmented lines, i.e. the drawing
+               ``'off'`` lines are never drawn as segmented lines, i.e. the drawing
                method with continuous lines is used
         """
-        if mode not in ('auto', 'on', 'off', True, False):
+        if mode not in ('auto', 'on', 'off'):
             raise ValueError(f'segmentedLineMode must be "auto", "on" or "off", got {mode} instead')
         self.opts['segmentedLineMode'] = mode
         self.invalidateBounds()
@@ -650,9 +650,9 @@ class PlotCurveItem(GraphicsObject):
 
     def _shouldUseDrawLineSegments(self, pen):
         mode = self.opts['segmentedLineMode']
-        if mode in ('on', True):
+        if mode in ('on'):
             return True
-        if mode in ('off', False):
+        if mode in ('off'):
             return False
         return (
             pen.widthF() > 1.0
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -424,49 +424,51 @@ class PlotCurveItem(GraphicsObject):
 
     def setData(self, *args, **kargs):
         """
-        =============== =================================================================
+        ================= =================================================================
         **Arguments:**
-        x, y            (numpy arrays) Data to display
-        pen             Pen to use when drawing. Any single argument accepted by
-                        :func:`mkPen <pyqtgraph.mkPen>` is allowed.
-        shadowPen       Pen for drawing behind the primary pen. Usually this
-                        is used to emphasize the curve by providing a
-                        high-contrast border. Any single argument accepted by
-                        :func:`mkPen <pyqtgraph.mkPen>` is allowed.
-        fillLevel       (float or None) Fill the area under the curve to
-                        the specified value.
-        fillOutline     (bool) If True, an outline surrounding the `fillLevel`
-                        area is drawn.
-        brush           Brush to use when filling. Any single argument accepted
-                        by :func:`mkBrush <pyqtgraph.mkBrush>` is allowed.
-        antialias       (bool) Whether to use antialiasing when drawing. This
-                        is disabled by default because it decreases performance.
-        stepMode        (str or None) If 'center', a step is drawn using the `x`
-                        values as boundaries and the given `y` values are
-                        associated to the mid-points between the boundaries of
-                        each step. This is commonly used when drawing
-                        histograms. Note that in this case, ``len(x) == len(y) + 1``
+        x, y              (numpy arrays) Data to display
+        pen               Pen to use when drawing. Any single argument accepted by
+                          :func:`mkPen <pyqtgraph.mkPen>` is allowed.
+        shadowPen         Pen for drawing behind the primary pen. Usually this
+                          is used to emphasize the curve by providing a
+                          high-contrast border. Any single argument accepted by
+                          :func:`mkPen <pyqtgraph.mkPen>` is allowed.
+        fillLevel         (float or None) Fill the area under the curve to
+                          the specified value.
+        fillOutline       (bool) If True, an outline surrounding the `fillLevel`
+                          area is drawn.
+        brush             Brush to use when filling. Any single argument accepted
+                          by :func:`mkBrush <pyqtgraph.mkBrush>` is allowed.
+        antialias         (bool) Whether to use antialiasing when drawing. This
+                          is disabled by default because it decreases performance.
+        stepMode          (str or None) If 'center', a step is drawn using the `x`
+                          values as boundaries and the given `y` values are
+                          associated to the mid-points between the boundaries of
+                          each step. This is commonly used when drawing
+                          histograms. Note that in this case, ``len(x) == len(y) + 1``
                         
-                        If 'left' or 'right', the step is drawn assuming that
-                        the `y` value is associated to the left or right boundary,
-                        respectively. In this case ``len(x) == len(y)``
-                        If not passed or an empty string or `None` is passed, the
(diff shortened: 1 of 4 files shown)


pijyoi:
Could you modify examples/PlotSpeedTest.py to allow interactive changing of this segmented line mode parameter?

swvanbuuren:
I've added an option called 'segmented lines' to examples/PlotSpeedTest.py.

Does this fit your expectations?

pijyoi:
I think you should model it after onPenChanged. Currently setSegmentedLineMode() is getting called upon every line update.

swvanbuuren:
Now it's implemented analogous to the pen options.

pijyoi:
Was exercising this new feature through a `PlotDataItem` and realized that it is not interchangeable with a `PlotCurveItem`. (Note: I am just checking out the ways to use this API, not saying there's anything needed to change here.)

If you look at a `PlotDataItem`, you will find that a lot of code is spent propagating options to the underlying `PlotCurveItem` or `ScatterPlotItem`. In my opinion, the ability to change this particular option via keyword options of `setData` (whether for `PlotCurveItem` or `PlotDataItem`) doesn't seem useful. After all, this option is the sort that probably doesn't need to change for the life-time of the `PlotCurveItem`. 

```python
import numpy as np
import pyqtgraph as pg

pg.mkQApp()

pw = pg.PlotWidget()
pw.show()
x = np.linspace(0, 2*np.pi, 25)
pen = pg.mkPen('r', width=4.0)

# we can create PlotDataItem with a particular mode
# by changing the global config option temporarily
oldopt = pg.getConfigOption('segmentedLineMode')
pg.setConfigOption('segmentedLineMode', 'on')
pdi = pw.plot(pen=pen)
assert pdi.curve.opts['segmentedLineMode'] == 'on'
pg.setConfigOption('segmentedLineMode', oldopt)

# or we could set it later programmatically
pdi.curve.setSegmentedLineMode('off')
assert pdi.curve.opts['segmentedLineMode'] == 'off'

# Can't set directly through PlotDataItem
# pdi.setSegmentedLineMode('on')

# this fails, PlotDataItem doesn't propagate to PlotCurveItem
pdi.setData(x, np.sin(x), segmentedLineMode='on')
assert pdi.curve.opts['segmentedLineMode'] == 'on'

pg.exec()
```

swvanbuuren:
I can extend the API so that `PlotDataItem` also propagates the segmented line option to `PlotCurveItem`.

But you're saying that, in this case, that's not necessary?

pijyoi:
In my opinion,

1) it would be over- engineering things to expose every option of
PlotCurveItem to PlotDataItem.

2) allowing the mode to be changed via setData keyword argument also seems
excessive.

swvanbuuren:
Ok, so I've removed the option to change the segmentedLineMode via setData in `PlotCurveItem`

Is there anything else that should be changed or improved, before this pull request can be granted?

j9ac9k:
Hi @swvanbuuren Sorry it's taken so long to get a response from a maintainer.  This is a very useful PR.  Thank you @pijyoi for the feedback here, I agree with everything you said.

The only suggestion I would have is that the global configuration option should also accept booleans `True` and `False` (not just strings for `'on'` and `'off'`).  I think `'auto'` should be the default as it currently is.  I mention this as many of the existing options works with booleans.

Regarding the anti-alias option here, can you provide an example (ideally using one of the examples) that shows the issue when anti-aliasing is enabled?

Thanks again for the PR.

swvanbuuren:
The global configuration option now also accepts `True` and `False`

For the anti-alias option; I have shown examples in issue #2178 that illustrate the behavior with segmented lines and anti-aliasing enabled.

These examples and their results were the reason why I disabled segmented lines by default, if anti-aliasing is enabled.

IMHO it's better to disable segmented lines if anti-aliasing is enabled (after all, it's also disabled if alpha < 1).
However, I am willing to retract this behavior, if this does not fit the general opinion of the maintainers/community of pyqtgraph.

pijyoi:
Wouldn't it be troublesome for a user to getConfigOption and have to test it against both boolean and str?
Perhaps we should just stick to either all strings or (True, False, None)?

j9ac9k:
> Wouldn't it be troublesome for a user to getConfigOption and have to test it against both boolean and str? Perhaps we should just stick to either all strings or (True, False, None)?

Good point, True/False/None might be the right combination here....

@swvanbuuren out of curiosity is there a particular is there a motivating factor that got you thinking you should use strings instead of booleans + None?

I should test locally the anti-alias options (not that I don't take your word for it)

pijyoi:
> out of curiosity is there a particular is there a motivating factor that got you thinking you should use strings instead of booleans + None?

https://github.com/pyqtgraph/pyqtgraph/issues/2178#issuecomment-1019395518
That would be on me.
I think I was taking reference from the `connect` parameter of `PlotDataItem` which takes str.
There's also the `stepMode` option, which used to take booleans but then was expanded to take `str` to give it more flexibility.

j9ac9k:
> [#2178 (comment)](https://github.com/pyqtgraph/pyqtgraph/issues/2178#issuecomment-1019395518)
> That would be on me.
> I think I was taking reference from the `connect` parameter of `PlotDataItem` which takes str.
> There's also the `stepMode` option, which used to take booleans but then was expanded to take `str` to give it more flexibility.

Someone should talk to the maintainer and get some kind of consistency across the library on this sort of thing 😬

swvanbuuren:
Now I am confused.

Which set of options for segmented line mode is preferred?

1. `'on'`, `'off'` and `'auto'`
2. `True`, `False` and `None`
3. `'on'`/`True`, `'off'`/`False` and `'auto'`/`None`

I prefer option 1, since it most clearly expresses what each options does.

j9ac9k:
> Now I am confused.
> 
> Which set of options for segmented line mode is preferred?
> 
> 1. `'on'`, `'off'` and `'auto'`
> 2. `True`, `False` and `None`
> 3. `'on'`/`True`, `'off'`/`False` and `'auto'`/`None`
> 
> I prefer option 1, since it most clearly expresses what each options does.

For the record, these kinds of minor stuff are the things I dread most.  I'm going to take some time later today and browse through the rest of the library and identify similar functionality and see how we handle it there.  I agree that `on`, `off`, and `auto` more clearly express the intended functionality.  At a glance earlier, I got the sense that `True`, `False` and `None` were what the rest of the library API usually expects for manual overrides.  Either way, give me a day or so and I'll chime back in.

Meanwhile, @campagnola @outofculture @ixjlyons if you all have input, please weigh in.

outofculture:
> 1. `'on'`, `'off'` and `'auto'`

This is my preference as well. Explicitness is a good default.

j9ac9k:
Alright @swvanbuuren we've talked it over internally, and there is a clear majority that prefer `on`, `off` and `auto` so we'll go w/ that, let me give this another look over, thanks for your patience.

pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -624,7 +625,35 @@ def getPath(self):
...
+        mode : str
+               ``'auto'`` (default) segmented lines are drawn if the pen's width > 1, pen style is a
+               solid line, the pen color is opaque and anti-aliasing is not enabled.
+
+               ``'on'`` or ``True`` lines are always drawn as segmented lines
+
+               ``'off'`` or ``False`` lines are never drawn as segmented lines, i.e. the drawing
+               method with continuous lines is used
+        """
+        if mode not in ('auto', 'on', 'off', True, False):
```
j9ac9k, line 645: so we can nuke the `True` / `False` bit here, sorry for the run-around.

j9ac9k:
Hi @swvanbuuren 

I put an in-line comment, the only other request I have is to add the option to [`config_options.rst`](https://github.com/pyqtgraph/pyqtgraph/blob/master/doc/source/config_options.rst)

This PR is ready for merging after that IMO.

swvanbuuren:
Also have added the `'segmentedLineMode'` option to the table with global configuration options in [config_options.rst](https://github.com/pyqtgraph/pyqtgraph/blob/master/doc/source/config_options.rst) in commit [76a7e08](https://github.com/pyqtgraph/pyqtgraph/pull/2185/commits/76a7e08f929b72d4c88cde13a4582f9955be8acd).

j9ac9k:
Thank you @swvanbuuren for this PR and thank you for your patience with us, this should have been merged long ago, and for that, please accept my apologies.

Merging!

pijyoi:
```python
def _shouldUseDrawLineSegments(self, pen):
    mode = self.opts['segmentedLineMode']
    if mode in ('on'):
        return True
    if mode in ('off'):
        return False
```
The `('on')` and `('off')` should be `('on',)` and `('off',)` respectively.
Otherwise, we have the following behaviour:
```python
In [1]: 'o' in ('off')
Out[1]: True

In [2]: 'of' in ('off')
Out[2]: True

In [3]: 'ff' in ('off')
Out[3]: True

In [4]: 'off' in ('off')
Out[4]: True
```

j9ac9k:
Nice catch @pijyoi, that slipped by all of us!

PR #2186, opened 2022-01-30
Reverted to real data to export CSV correctly

StSav012:
When the axes are in log mode, what's saved is log10 of values instead of real data. The view should _never alter_ the data to export.

Please check how the change can affect other code. I haven't found anything.

Please check other item types that have their own `getData()` function implementations. I don't have time for that right now.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -1065,12 +1065,11 @@ class PlotDataItem(GraphicsObject):
 
     def getData(self):
         """
-        Returns the displayed data as the tuple (`xData`, `yData`) after mapping and data reduction.         
+        Returns the data as the tuple (`xData`, `yData`).
         """
-        dataset = self.getDisplayDataset()
-        if dataset is None:
+        if self._dataset is None:
             return (None, None)
-        return dataset.x, dataset.y
+        return self._dataset.x, self._dataset.y
 
     # compatbility method for access to dataRect for full dataset:
     def dataRect(self):


StSav012:
MWE:

```python
# coding: utf-8
import pyqtgraph as pg


class Window(pg.QtWidgets.QMainWindow):
    def __init__(self) -> None:
        super().__init__()
        plot: pg.PlotWidget = pg.PlotWidget(self)
        line: pg.PlotDataItem = plot.plot([1, 2, 4], [10, 20, 40])
        self.setCentralWidget(plot)
        canvas: pg.PlotItem = plot.plotItem
        print('data in linear mode:', line.getData())
        canvas.setLogMode(True, True)
        print('data in log mode:', line.getData())  # what gets exported to CSV
        line.getData = lambda: (line.xData, line.yData)
        print('correct data in log mode:', line.getData())


app = pg.QtWidgets.QApplication()
w = Window()
w.show()
app.exec()
```

pijyoi:
Just to clarify, this is new behavior that you are requesting?
The old behavior seems to go as far back as 0.11.0 (June 2020). Older than that doesn't run on my Python 3.8 environment.

```python
import pyqtgraph as pg

pg.mkQApp()

plot = pg.PlotWidget()
line = plot.plot([1, 10, 100, 1000], [1, 10, 100, 1000])
canvas = plot.plotItem
canvas.setLogMode(True, True)
print('data in log mode:', line.getData())  # what gets exported to CSV
```

| version | output | 
|-------|---------|
|0.12.3 | data in log mode: (array([0., 1., 2., 3.]), array([0., 1., 2., 3.]))  |
|0.12.2 | data in log mode: (array([0., 1., 2., 3.]), array([0.30103   , 1.04139269, 2.00432137, 3.00043408]))  |
|0.12.1 | data in log mode: (array([0., 1., 2., 3.]), array([0.30103   , 1.04139269, 2.00432137, 3.00043408]))  |
|0.12.0 | data in log mode: (array([0., 1., 2., 3.]), array([0.30103   , 1.04139269, 2.00432137, 3.00043408]))  |
|0.11.1 | data in log mode: (array([0., 1., 2., 3.]), array([0., 1., 2., 3.]))   |
|0.11.0 | data in log mode: (array([0., 1., 2., 3.]), array([0., 1., 2., 3.]))   |

StSav012:
Yes, that's the behavior I consider to be the appropriate one. It's been an unpleasant surprise for me to see that the exported data depend on the axes' mode. I strongly believe that the data I export to a CSV file should be the same as what I've plotted.

However, I can't match the code you posted with the results. What I get from your code with the proposed patch is 
```plaintext
data in log mode: (array([   1,   10,  100, 1000]), array([   1,   10,  100, 1000]))
```
And without the patch
```plaintext
data in log mode: (array([0., 1., 2., 3.]), array([0., 1., 2., 3.]))
```
Why do your results differ for the two axes?

pijyoi:
If you get `data in log mode: (array([0., 1., 2., 3.]), array([0., 1., 2., 3.]))`, you are either using version <= 0.11.1 or >= 0.12.3

StSav012:
Shouldn't the data replicate the initial values? I believe they should. I expect the exported data to be `array([   1,   10,  100, 1000])` for both axes. It's bizarre to see the `getData()` results differ for x and y axes. So, I think that with neither of your results are the behavior users expect.

pijyoi:
> It's bizarre to see the getData() results differ for x and y axes. So, I think that with neither of your results are the behavior users expect.

`getData()` returns the same values for both axes since pyqtgraph >= 0.12.3 as a result of #1992, fixing a behavior that only existed between 0.12.0 to 0.12.2

Which version of pyqtgraph are you using?

StSav012:
And still, none of the versions return the data actually plotted! The export to a CSV file is useless with such inconsistency. All I'm asking is to export the correct data.

I'm using the version from the repo of 2 days ago.

NilsNemitz:
Hi @StSav012 !

I think the misalignment in the current discussion is that contrary to the title, this PR does not *revert* to a previous behaviour.

As far as the version history above goes, `getData()` has always returned the displayed data, not the user's original data.
The only reason that some of the data above differs is that there was a code change in 12.0 that resulted in somewhat undesirable output on screen and -again identically- in what `getData()` returned. This was reverted in 12.3.

The original thought process for the CSV export *may* have been that all the other (graphics) export options export exactly what is on the screen, so why should CSV export be different? It might also be a reasonable assumption that the user code already has the data that was passed to PlotDataItem, so that retrieving a spare copy is not a high priority feature. The user does not have the transformations calculated by pyqtgraph. (Although as long as the transform is a simple `log10`, there isn't much gain here either.)

For my typical use case, exporting the raw values would indeed be convenient, even if only to abuse pyqtgraph's context menu to lazily avoid setting up my own menus. So I would personally support a change to the CSV exporter in this direction. I believe the original data is already available through the `xData` and `yData` properties of `PlotDataItem`. Maybe a more reasonable approach would be to adjust the CSV exporter to collect the data in this way? That would not require changing the established return of `getData()`.

NilsNemitz:
@StSav012 ,

`CurvePoint` and `PlotItem` seem to call `getData()` on the associated curves. I am not sure if those can be of type `PlotDataItem`, but it does support the assumption that the expected return is indeed a screen coordinate.

I had been looking into more changes to the map-to-screen functions, of which log10 is the only one we have at this point. I have, however, run out of time right now.

A consideration at the current stage was indeed how to access mapped ("log") vs unmapped ("original") data. 
The main candidates were either
(a) `getData( coord='orig' )`
or 
(b) `getOriginalData()`

While (a) avoids some optical clutter, (b) has the advantage that is is easy to check if a given element implements this method... and fall back to `getData` when it doesn't. And then ideally apply the missing transform manually.

If your investigations give you any ideas on this, please let us know!

PR #2187, opened 2022-02-02
Apply the log mode to the displayed dataset, export the mapped one

StSav012:
My next try to export the actual data plotted in log mode.

I believe that the log mode is something of displaying, not of the actual data. So, I moved the `applyLogMapping` function call down to `self._datasetDisplay`.

Next, I return the data saved into `self._datasetMapped` in `PlotDataItem.getData` function. It's a kind of expected behavior when it comes to returning the data, not the position of screen points.

I realize that it better be two tiny commits.


StSav012:
Oops, somehow the whole syntax got changed. Sorry about that. I'll redo the PR with only the actual changes.

PR #2192, opened 2022-02-07
Add an option to makeARGB to disable masking NaNs

JamesWrigley:
Sometimes it's useful for NaNs not to be masked by transparent pixels.

(originally written by @cammilleCC)

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1353,7 +1353,7 @@ def makeRGBA(*args, **kwds):
     return makeARGB(*args, **kwds)
 
 
-def makeARGB(data, lut=None, levels=None, scale=None, useRGBA=False, output=None):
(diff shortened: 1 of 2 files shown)


outofculture:
Great! Can you throw a quick test in, too? In `tests/test_makeARGB.py:test_makeARGB_with_human_readable_code`, you could add something like:

```python
    inputWithNaNs = ...
    expectedOutputWithNaNs = ...
    checkImage(_makeARGB(inputWithNaNs, maskNans=False), expectedOutputWithNaNs)
```

I also wonder if we should spell it `maskNaNs`... That's my first impulse, but I'd be willing to hear arguments against it.

JamesWrigley:
Sure, added in 0b32fb8b. WRT the naming, I lean towards `maskNans` because it's easier to type (:stuck_out_tongue:), but I don't feel very strongly about it.

outofculture:
I suppose numpy spells it `nan`, so we're not doing anything too weird.

PR #2194, opened 2022-02-09
numpy deprecated binary use of fromstring

outofculture:
Deprecated as of 1.14.


j9ac9k:
MetaArray updates!!!! 

Thanks @outofculture LGTM, merging.

PR #2198, opened 2022-02-14
improve GraphicsWidget paint speed by caching the bounding rect and the path used by the shape method

danielhrisca:
By caching the bounding rect the painting is increased significantly. On a live updating plot with ~160 curves the timings show the following:

#### With this patch 0.95 us/call
image

#### Without 1.99 us/call
image

Overall the painting speed went from 43.5ms/frame to 39.3ms/frame

### DIFF
--- a/pyqtgraph/graphicsItems/GraphicsWidget.py
+++ b/pyqtgraph/graphicsItems/GraphicsWidget.py
@@ -18,6 +18,7 @@ class GraphicsWidget(GraphicsItem, QtWidgets.QGraphicsWidget):
 
         # cache bouding rect and geometry
         self._br = self._geometry = None
+        self._path = None
         
         ## done by GraphicsItem init
         #GraphicsScene.registerObject(self)  ## workaround for pyqt bug in graphicsscene.items()
@@ -50,6 +51,8 @@ class GraphicsWidget(GraphicsItem, QtWidgets.QGraphicsWidget):
     def boundingRect(self):
         geometry = self.geometry()
         if geometry != self._geometry:
+            self._path = None
+            
             br = self.mapRectFromParent(geometry).normalized()
             self._br = br
             self._geometry = geometry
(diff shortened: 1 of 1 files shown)


danielhrisca:
I've also added the caching of the path used by the ``shape`` method.

With ``shape`` caching:

``shape`` takes 0.18 us/call and the painting takes 36ms/frame
image

Without ``shape caching
``shape`` takes 2.48 us/call and the painting takes 39ms/frame
image

pyqtgraph/graphicsItems/GraphicsWidget.py
```diff
@@ -45,12 +49,25 @@ def width(self):
...
         return br
-        
+
     def shape(self):  ## No idea why this is necessary, but rotated items do not receive clicks otherwise.
-        p = QtGui.QPainterPath()
-        p.addRect(self.boundingRect())
-        #print "shape:", p.boundingRect()
+        p = self._path
+        if p is None:
+            p = QtGui.QPainterPath()
```
j9ac9k, line 68: can probably put this in one line...

```python
self._path = p = QtGui.QPainterPath()
```

j9ac9k:
hi @danielhrisca 

Thanks for this PR and the profiling results.   I would ask you rename the variables to better reflect that they hold cached values.  This LGTM otherwise!

danielhrisca:
> I would ask you rename the variables

Done

pyqtgraph/graphicsItems/GraphicsWidget.py
danielhrisca: done

pyqtgraph/graphicsItems/GraphicsWidget.py
```diff
@@ -45,12 +49,25 @@ def width(self):
...
+            p.addRect(self.boundingRect())
+            self._path = p
+
+        # print "shape:", p.boundingRect()
```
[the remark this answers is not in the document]
danielhrisca: done

danielhrisca:
@j9ac9k can you trigger a rerun since there was a network problem that caused the failure ?

j9ac9k:
LGTM, thanks @danielhrisca

PR #2199, opened 2022-02-14
avoid unnecessary call if autorange is disabled

danielhrisca:
slight increase in performance for live updating plots in case autorange is disabled


j9ac9k:
Nice find @danielhrisca LGTM, merging.

PR #2201, opened 2022-02-15
micro-optimization for plotcurveitem paint

danielhrisca:
avoids calling ``get`` for the ``pen`` keys since this is always in the dictionary.


j9ac9k:
Hi @danielhrisca 

Thanks for the PR.  I do believe this would qualify as a micro-optimization as we're removing a O(1) lookup 😄 

LGTM, merging.

danielhrisca:
Yes, this was a minor finding as I was profiling the code base for reaching 60 FPS on a live plotting application (driving ~200 curves with 2 samples / horizontal pixel for the min and max values of the culled signal). The real bottlenecks were the ones found in #2202, #2198 and @2199

PR #2202, opened 2022-02-16
cache the ViewBox view pixel size

danielhrisca:
For a live data streaming application I needed to call ``ViewBox.viewPixelSize`` multiple times (large Curve count and high frame rate), which become very costly.

With this change the size tuple is cached and updated when the ViewBox is resized or when the ranges are changed.


j9ac9k:
Thanks for the PR @danielhrisca

PR #2211, opened 2022-03-05
Remove Python 3.7 from CI - Add Conda Package Testing

j9ac9k:
Per NEP-29, the next release of PyQtGraph will no longer support Python 3.7 or Numpy 1.18; so the CI pipelines using those configurations are removed.  

This puts PySide2 5.12.x into an unfortunate situation.  While there exists a PySide2 5.12 wheel on pip for Python 3.8, [PYSIDE-1140](https://bugreports.qt.io/browse/PYSIDE-1140) prevents us from being able to use it.  That means, that if a user wants to use PySide 5.12, they _must_ use the conda-forge pyside2 package; which we currently do not test for.

This PR adds CI tests to test conda pyside2 packages on each OS on Python 3.8.  

Furthermore, this PR also updates the `README.md`, `setup.py`, `pyproject.toml` and `.pre-commit-config.yml` to reflect the new python required versions.

On my fork, I have had the `test-conda` macOS job hang at the test stage.  I'm hoping that's just a temporary CI hick-up, if not, we should consider removing the `pytest-xdist` plugin from the conda jobs.


.github/workflows/main.yml
```diff
@@ -7,39 +7,31 @@ concurrency:
...
-            numpy-version: "~=1.18.0"
           - python-version: "3.8"
             qt-lib: "pyqt"
-            qt-version: "PyQt5~=5.15.0"
+            qt-version: "PyQt5~=5.12.0"
             numpy-version: "~=1.19.0"
           - python-version: "3.8"
             qt-lib: "pyside"
-            qt-version: "PySide2~=5.15.0"
+            qt-version: "PyQt5~=5.15.0"
```
pijyoi, line 26: qt-lib: "pyside" should be using a PySide binding

j9ac9k:
Debugging the macOS / conda issue is going to be tough... guess I have the setup to do it, but will require significant changes to my local setup (probably better off creating a new user on my mac).

.github/workflows/main.yml
j9ac9k: 🤦🏻

j9ac9k: Fixed now, nice catch.

j9ac9k:
I could not reproduce a segfault on my local mac machine w/ an environment that I attempted to match up to the CI to the best of my ability.  What I did get were a _ton_ of warnings everywhere we used an enum flag instead of an integer, so I told pytest to filter those warnings out, and hoping that does the trick, but I somehow doubt it.

j9ac9k:
I've identified one test causing a segfault here, `test_histogramlutwidget`, but there is at least one more there test triggering this, and at least one example (the examples are a bit trickier to debug).  I'm going to try and develop an exhaustive list of the tests causing issues over the next day or so.

https://github.com/pyqtgraph/pyqtgraph/blob/master/tests/widgets/test_histogramlutwidget.py#L13-L46

j9ac9k:
As I'm re-enabling tests, I'm finding quite a few tests seem to be triggering segfaults; so many so that I'm considering just abandoning the conda test suite on macOS.  I'm going to spend some more time trying to narrowly define the tests that result in segfault, but as I can't replicate locally this may be more trouble than its worth.

j9ac9k:
Issue seems to be `QtWidgets.QApplication.processEvents()` calls ...which would explain why the segfaults are so prevalent ... I'm going to attempt to see if it will work with some of the flags we can pass to the `processEvents()` method, but I'm not hopeful.

pijyoi:
>While there exists a PySide2 5.12 wheel on pip for Python 3.8, [PYSIDE-1140](https://bugreports.qt.io/browse/PYSIDE-1140) prevents us from being able to use it.

I don't see any Python 3.8 wheel for PySide2 5.12 at https://pypi.org/project/PySide2/5.12.6/#files

j9ac9k:
> > While there exists a PySide2 5.12 wheel on pip for Python 3.8, [PYSIDE-1140](https://bugreports.qt.io/browse/PYSIDE-1140) prevents us from being able to use it.
> 
> I don't see any Python 3.8 wheel for PySide2 5.12 at https://pypi.org/project/PySide2/5.12.6/#files

It installed the one labeled for `PySide2-5.12.6-5.12.6-cp35.cp36.cp37-abi3-macosx_10_12_intel.whl`

```bash
❯ pyenv virtualenv 3.8.12 pyside2_512-py38
❯ pyenv activate pyside2_512-py38
pyside2_512-py38 ❯ pip install --upgrade pip setuptools wheel
pyside2_512-py38 ❯ pip install pyside2~=5.12.0
Collecting pyside2~=5.12.0
  Using cached PySide2-5.12.6-5.12.6-cp35.cp36.cp37-abi3-macosx_10_12_intel.whl (115.1 MB)
Collecting shiboken2==5.12.6
  Using cached shiboken2-5.12.6-5.12.6-cp35.cp36.cp37-abi3-macosx_10_12_intel.whl (779 kB)
Installing collected packages: shiboken2, pyside2
Successfully installed pyside2-5.12.6 shiboken2-5.12.6
pyside2_512-py38 ❯ python3
Python 3.8.12 (default, Oct 17 2021, 10:10:09)
[Clang 13.0.0 (clang-1300.0.29.3)] on darwin
>>> from PySide2 import QtCore
>>>
```

j9ac9k:
I'm starting to lean more towards giving up on conda-forge macOS CI runs; doing a github advanced search for other projects I got 8 results, 6 of which are outdated forks, one project is [QtPy](https://github.com/spyder-ide/qtpy), and the other, which I had never heard of is [labelme](https://github.com/wkentaro/labelme).  Neither of the projects seems to have `QApplication.processEvents()` in their code-base.

Going to give myself a little more time to see if I can sort this out.

pijyoi:
I seem to understand from the thread in PYSIDE-1140 that being able to pip install pyside2_512 on Python 3.8 is itself a bug.

The first pypi wheel to support Python 3.8 being pyside2_514.

j9ac9k:
> 

Yeah, we're in a tough spot.  5.12 LTS support has ended, and given that there are newer versions of Qt bindings available on pypi, I'm _all for_ making the oldest supported version 5.15.

Unfortunately, some users (I really should get back to writing that user survey so I can at least approximate how many) are dependent on conda packages.  Anaconda defaults channel distributes Qt 5.9 (which is used in `matplotlib`), and the newer PySide2 (5.13.2 build with Qt 5.12.8, which I didn't know you could cross-build like that) is available via conda-forge.  The conda-forge pyside2 bindings supposedly work through Python 3.10 ([link to files](https://anaconda.org/conda-forge/pyside2/files)).

While I would love to ditch Qt 5.12/5.13, I'm not sure that is in the cards while that's the newest version available to users that for whatever reason, need to stick to conda-packages. @ccordoba12 mentioned in an issue a few months back that there was work being done to make PySide2 5.15 available on conda-forge; but I have no clue what the status there is.

ccordoba12:
Hey @j9ac9k, the work on updating Qt/PyQt to 5.15 is in progress. I hope it'll be over by the end of the month, or a month and a half at most. The main bottleneck at the moment is the long times the Spyder developer working on this has to wait until the conda-forge core team uploads his packages (which can't be compiled on CIs) to the channel.

After that, updating PySide2 should take a couple of weeks more. But we're not planning to work on it, so it's up to the community.

ccordoba12:
> If memory serves, the build-time for the Qt bindings was a major hurdle with respect to getting it available on conda-forge.

That's right. That's why Qt was split in qt-main and qt-webengine, so that the first can be compiled in CIs (except aarch64, which runs out of time). The second needs to compiled locally because it's huge (it involves compiling Chromium and then the Qt bindings for it).

j9ac9k:
I'm happy w/ this PR in this state, the only other feature I would like to try (again) to implement is the caching of python environments.  the install dependencies stage is one of the longest portions of the runtime.

.github/workflows/main.yml
```diff
@@ -115,8 +119,8 @@ jobs:
     - name: Run Tests
       run: |
         mkdir $SCREENSHOT_DIR
+        pytest pyqtgraph/examples -v -n 2
```
pijyoi, line 122: Shouldn't tests come before examples since the former takes a lot less time than latter? (fast fail)

j9ac9k:
I think I'm going to call it good, a couple of things I wish I could sort out but couldn't figure out

* Newer MESA driver than 19.2.7
* `dependabot` to update `.github/workflows/etc/requirements.txt`

Both of these things are pretty far into the nice to have, so I'm going to skip.

I'm still getting intermittent failures on the pytest suite on macOS when using pytest-xdist, so I'm going to skip that for now and just eat the longer runtime.

.github/workflows/main.yml
j9ac9k: sounds reasonable, order changed 👍🏻

j9ac9k:
I’m indifferent to the order but the reasoning makes sense.  I’ll change
the order.

PR #2224, opened 2022-03-12
Remove old Qt workarounds

pijyoi:
Remove workarounds for old versions of Qt.
Some old Python 2 compatibility code also removed.


j9ac9k:
Thanks for the PR @pijyoi

Diffs look good, merging

Issue #2173, opened 2022-01-12
Item border QGraphicsRectItem not properly removed from GraphicsLayoutWidget after child item is removed


theMashUp:
### Short description
When an item is added to a GraphicsLayoutWidget using GraphicsLayoutWidget.addItem(), it is added to the layout's contents alongside a border QGraphicsRectItem. When this item is removed using GraphicsLayoutWidget.addItem(), the border QGraphicsRectItem stays in the layout's contents. For each added and removed item, a superfluous QGraphicsRectItem persists.

The impacts are as follow:
1. Small memory leak
2. In a personal project, affects the display of items added after having been previously removed.

### Code to reproduce
```python
import pyqtgraph as pg

graphics_layout_widget = pg.GraphicsLayoutWidget()

# One item in GraphicsLayoutWidget: GraphicsLayout
items = graphics_layout_widget.items()
print(f"{len(items)} items: {items}")

# Add a PlotItem, many items are added
plot_item = pg.PlotItem()
graphics_layout_widget.addItem(plot_item)
items = graphics_layout_widget.items()
print(f"{len(items)} items: {items}")

# Remove the PlotItem, GraphicsLayout is still present (as expected),
# but we have superfluous QGraphicsRectItem remaining.
graphics_layout_widget.removeItem(plot_item)
items = graphics_layout_widget.items()
print(f"{len(items)} items: {items}")
```

### Expected behavior
GraphicsLayoutWidget.items() should contain only the instance of GraphicsLayout when all items have been removed.

### Real behavior
GraphicsLayoutWidget.items() contains one instance GraphicsLayout as well as one+ instance(s) of QGraphicsRectItem when all items have been removed.

### Tested environment(s)

 * PyQtGraph version: 0.12.3
 * Qt Python binding: PySide2 5.15.2 Qt 5.15.2
 * Python version: 3.8
 * Operating system: Linux Mint 20.3, XFCE
 * Installation method: pip in virtual environment

PR #2225, opened 2022-03-12
remove border QGraphicsRectItem from scene

pijyoi:
This fixes issue #2173 where the QGraphicsRectItem representing the border for the item is leaked when the item is removed.
This fix can be tested with the MWE given in #2173


j9ac9k:
Love getting rid of `del` statements.  Thanks for the PR @pijyoi

Issue #2179, opened 2022-01-19
No module named 'pyqtgraph.graphicsItems.ViewBox.axisCtrlTemplate_pyqt5' while executing exe


mnwato:
`
 No module named 'pyqtgraph.graphicsItems.ViewBox.axisCtrlTemplate_pyqt5'
`

Facing with this error while executing exe file which is created via pyinstaller module.

- python -version: 3.8
- pyqtgraph -version: 0.12.3
- pyinstaller -version: 4.2

j9ac9k:
Hi @mnwato 

This is a known issue, the problem is that the freezing portion of pyinstaller doesn't recognize those modules as needing to be imported.  Here is a comment not long ago for the same issue: 

https://github.com/pyqtgraph/pyqtgraph/issues/1981#issuecomment-912031468

tl;dr you need to tell pyinstaller to grab a handful of files explicitely in the `.spec` file.

If you are familiar with pyinstaller hooks, we would greatly appreciate a PR to https://github.com/pyinstaller/pyinstaller-hooks-contrib specifying the supplemental files needed.

agausmann:
Was having the same issue, but the most recent PyInstaller 4.8 release is working for me.

mnwato:
Hi @j9ac9k

Thanks. I tried to make exe with `--paths=pyqtgraph.graphicsItems.PlotItem.plotConfigTemplate_pyside2` but it still stuck on the following error.

Here is `.spec` file:
```
# -*- mode: python ; coding: utf-8 -*-


block_cipher = None


a = Analysis(['NobitexMarketWatch_noToken.py'],
             pathex=['C:/Users/mostafa/AppData/Local/Programs/Python/Python310/Lib/site-packages/pyqtgraph/graphicsItems/ViewBox/axisCtrlTemplate_pyside2.py'],
             binaries=[],
             datas=[],
             hiddenimports=[],
             hookspath=[],
             hooksconfig={},
             runtime_hooks=[],
             excludes=[],
             win_no_prefer_redirects=False,
             win_private_assemblies=False,
             cipher=block_cipher,
             noarchive=False)
pyz = PYZ(a.pure, a.zipped_data,
             cipher=block_cipher)

exe = EXE(pyz,
          a.scripts,
          a.binaries,
          a.zipfiles,
          a.datas,  
          [],
          name='NobitexMarketWatch_noToken',
          debug=False,
          bootloader_ignore_signals=False,
          strip=False,
[...]

Issue #2122, opened 2021-11-25
issue when using py2exe to build a PySide6 application containing the latest pyqtgraph v0.12.3


RidaShamasneh:
I am trying to use py2exe to build a PySide6 application containing the latest pyqtgraph v0.12.3; and I have the following issue after building the application ...

Traceback (most recent call last):
  File "q_application.py", line 46, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 627, in _load_backward_compatible
  File "<frozen zipimport>", line 259, in load_module
  File "q_matrix3D.pyc", line 7, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 627, in _load_backward_compatible
  File "<frozen zipimport>", line 259, in load_module
  File "pyqtgraph\__init__.pyc", line 194, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 627, in _load_backward_compatible
  File "<frozen zipimport>", line 259, in load_module
  File "pyqtgraph\graphicsItems\GraphicsLayout.pyc", line 6, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 627, in _load_backward_compatible
  File "<frozen zipimport>", line 259, in load_module
  File "pyqtgraph\graphicsItems\ViewBox\__init__.pyc", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
[...]

Issue #1981, opened 2021-09-02
Pyinstaller issue when running exe


myue891:
Hello,

I am getting an issue only when I try to create an exe using pyinstaller and running that exe file. Everything works fine in my IDE and the GUI works as expected. This is the error I get when I try to run my executable. I have pyqtgraph installed correctly with pip and I can't figure out why this error occurs.

Thanks

<img width="496" alt="Screenshot 2021-09-02 132448" src="https://user-images.githubusercontent.com/55855607/131911242-99482cc0-44a2-4ef4-a169-b38f24909a49.png">

j9ac9k:
Hi @myue891 

This has been discussed in a few places in the issue tracker, if you search for pyinstaller you'll likely see some longer explanation.

Basically because of how pyqtgraph does importing of templates, pyinstaller doesn't know to bundle them.  In my application (I use pyside2 primarily), I have to manually specify the files for pyinstaller to grab:

```
        "pyqtgraph.console.template_pyside2",
        "pyqtgraph.graphicsItems.ViewBox.axisCtrlTemplate_pyside2",
        "pyqtgraph.graphicsItems.PlotItem.plotConfigTemplate_pyside2",
        "pyqtgraph.imageview.ImageViewTemplate_pyside2"
```

You likely need to modify your spec file to grab these.

If you're familiar with pyinstaller, and have time/ability to contribute to pyqtgraph, we would love for someone to create a custom hook that would go here: https://github.com/pyinstaller/pyinstaller-hooks-contrib  This has been on the todo for us, but we haven't gotten to it.

myue891:
Hi @j9ac9k 
Thank you for the reply, I found the solution in a similar post.

labbeast2020:
@j9ac9k Hi I have the same issue and this is my spec file 

`# -*- mode: python ; coding: utf-8 -*-
from PyInstaller.utils.hooks import collect_data_files
import os
import site

ROOT_PATH = os.getcwd()
PACKAGE_SITE= site.getusersitepackages()

print(ROOT_PATH)
print(PACKAGE_SITE)

datas = [(f'{ROOT_PATH}/config', './config'), (f'{ROOT_PATH}/assets', './assets'), (f'{ROOT_PATH}/log', './log')]
datas += collect_data_files('qt_material')

block_cipher = None


a = Analysis([f'{ ROOT_PATH }/main.py'],
[...]

Issue #1934, opened 2021-07-30
Float divison by zero in ButtonItem.py - Found temporary fix


TroyBrandt:
### Short description

Application crashes due to exception in pyqtgraph\graphicsItems\ButtonItem.py | Found a temporary fix by changing this:
`       if width is not None:
                s = float(width) / self.pixmap.width()
                self.setScale(s)`
to this:
`        if width is not None:
            try:
                s = float(width) / self.pixmap.width()
                self.setScale(s)
            except:
                pass`
on line 18

### Real behavior
Application crashes due to a ZeroDivisionError exception, this only happens when the program is compiled with pyinstaller.
```
An error occurred?
Post the full traceback inside these 
'Traceback (most recent call last):
  File "main.py", line 37, in <module>
  File "templateCode\botApplicationInit.py", line 18, in __init__
  File "PyQt5\uic\__init__.py", line 238, in loadUi
  File "PyQt5\uic\Loader\loader.py", line 66, in loadUi
  File "PyQt5\uic\uiparser.py", line 1037, in parse
  File "PyQt5\uic\uiparser.py", line 828, in createUserInterface
  File "PyQt5\uic\uiparser.py", line 806, in traverseWidgetTree
  File "PyQt5\uic\uiparser.py", line 273, in createWidget
  File "PyQt5\uic\uiparser.py", line 806, in traverseWidgetTree
  File "PyQt5\uic\uiparser.py", line 273, in createWidget
  File "PyQt5\uic\uiparser.py", line 806, in traverseWidgetTree
  File "PyQt5\uic\uiparser.py", line 273, in createWidget
  File "PyQt5\uic\uiparser.py", line 806, in traverseWidgetTree
  File "PyQt5\uic\uiparser.py", line 264, in createWidget
  File "PyQt5\uic\uiparser.py", line 228, in setupObject
  File "PyQt5\uic\objcreator.py", line 118, in createQObject
  File "PyQt5\uic\Loader\qobjectcreator.py", line 137, in instantiate
  File "pyqtgraph\widgets\PlotWidget.py", line 56, in __init__
  File "pyqtgraph\graphicsItems\PlotItem\PlotItem.py", line 122, in __init__
  File "pyqtgraph\graphicsItems\ButtonItem.py", line 19, in __init__
ZeroDivisionError: float division by zero
[33304] Failed to execute script 'main' due to unhandled exception!'!
```

### Tested environment(s)

 * PyQtGraph version: 0.12.2
[...]

PR #2226, opened 2022-03-15
use single generic template for all bindings

pijyoi:
Currently, each `.ui` file has a separate generated `.py` file for each of the 4 bindings.

This PR demonstrates that the currently supported bindings are sufficiently similar that a single generic "generated" file can be used for all bindings. The generic "generated" file is actually a hand-modified version of the files churned out by PyQt6's `pyuic6`.

One benefit of doing this would be to make it easier for users who use `py2exe`, `pyinstaller` or similar programs. These programs seem to have issues with bundling files imported using the following statement. (#2122, #2179, #1981, #1934, #1811, #1801)
```python
ui_template = importlib.import_module(
    f'.exportDialogTemplate_{QT_LIB.lower()}', package=__package__)
```

Disclaimer: Not being a user of any of these freezing / bundling programs, I can't say for certain that this PR would make those programs include the needed files.

Note: flowchart, canvas, console modules were left out since they are not much used.


pijyoi:
By deleting the non-generic files, it becomes clear from the git history that the generic files are just a rename of the *_pyqt6.py files with a one-line modification.

j9ac9k:
The only other thing I can think of that should probably be part of this PR is that `rebuildUi.py` should be updated to _only_ work on PyQt6 bindings, write out to `*_generic.py` and change the import statements so they import from pyqtgraph's abstraction layer.

pijyoi:
The automatic rewrite of the import line would need to handle different depth relative imports (for the library) and absolute imports (for the examples).

It would also be relying on pyuic6 output to remain the same style.

j9ac9k:
Forgot about the relative import, that would indeed be trickier (and frankly just add un-needed complexity here).  Probably not worth the effort.

pijyoi:
On a whim, I tried using `pyinstaller` (for the first time) on one of my scripts in a clean environment with `pyqtgraph 0.12.4`, `PySide6` and `numpy`. It seemed to just work (* with a little tweak) and I didn't encounter the issues listed before about missing template files and missing icon files.

(*) The colormap files were missing, so I changed my script to not use them.

Digging a bit, it seems that someone has taught `pyinstaller` about `pyqtgraph` template files.
https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/1649c5e412e02f5ebee4c542093064c8ba1c96bb/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyqtgraph.py

j9ac9k:
Thanks for this @pijyoi 

Hopefully, this will make life easier for the folks doing freezing.

Issue #2230, opened 2022-03-16
0.12.4 regression of GL transformation


cwiede:
### Short description
I'm drawing an image in a 3D view with a specific transformation and I realize there are differences between 0.12.3 and 0.12.4. The image is appearing flipped compared to 0.12.3.

### Additional context
EDIT: When the changes from https://github.com/pyqtgraph/pyqtgraph/commit/36b2e49cd3b0ba38a2169e8cb3b2d2d7c8b5e619 are reverted, the image is displayed in its original pose. So I guess there is an issue with this commit.

pijyoi:
Hello, thank you for the report. The commit in question was part of #2131 and there was a script included https://github.com/pyqtgraph/pyqtgraph/pull/2131#issue-1071166647 used for verifying that the "fix" had not changed the pose.

Running the script in both 0.12.3 and 0.12.4 seems to show the same pose, so I would say that there's something missing about my reasoning there.

Is your image data row-major or column-major? (The test script uses a row-major image)
By "flipped", do you mean a transpose or a left-right flip?
Would it be possible for you to provide an MWE?

cwiede:
The relevant code snippet is

```python
        T = np.array([
            [0          , (y0-y1)/w,0 ,y1],
            [(x0 - x1)/h, 0        ,0 ,x1],
            [0          , 0        ,0 ,z ],
            [0          , 0        ,0 ,1 ],
        ])
        logger.info("T=%s", T)
        t = Transform3D(*T.flatten())
        self.resetTransform()
        self.applyTransform(t, True)
        self.setData(rgba)
        self.setVisible(True)
```
where 

```
    T = [[ 0.          0.05        0.         -5.00000007]
           [ 0.05        0.          0.         -5.00000007]
           [ 0.          0.          0.         -0.01      ]
           [ 0.          0.          0.          1.        ]]
```
RGBA has the following properties:

```
rgba.shape=(200, 200, 4)
rgba.flags=  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```

Results of 0.12.3 (the image is the black white projection of the point cloud to the floor, the point cloud is also displayed in color coding for reference; i have asserted that only the version of pyqtgraph has changed, everything else is exactly same)
[...]

PR #2131, opened 2021-12-04
avoid PyOpenGL automatic array conversion

pijyoi:
Following #2111, this PR tries to eliminate PyOpenGL automatic array conversions in the OpenGL items by making the conversions explicitly. The places where conversion was taking place was discovered by setting ```OpenGL.ERROR_ON_COPY = True``` in the examples and running them.  With this PR, all included OpenGL examples are able to run without error with the above attribute set.

GLImageItem and GLVolumeItem deserve a mention. Pyqtgraph uses Fortran-major axes and thus transposes the texture data before uploading them. GLImageItem was fixed to avoid the transpose by changing the texture coordinate code. The texture coordinate code of GLVolumeItem was too complicated to fix, thus a memory copy is forced.

The script below was used to verify that GLImageItem draws the same before and after this PR.
```python
import pyqtgraph as pg
import pyqtgraph.opengl as gl
import numpy as np
import scipy.misc

pg.mkQApp()

glv = gl.GLViewWidget()
glv.setCameraParams(elevation=90, azimuth=0)
# X points down, Y points right
glv.show()

image = scipy.misc.face()
tex, _ = pg.makeRGBA(image)
glii = gl.GLImageItem(tex)
s = 1/128
glii.scale(s, s, 1)
w, h = tex.shape[:2]    # GLImageItem uses transposed coords?
w *= s
h *= s
glii.translate(-w/2, -h/2, 0)
glv.addItem(glii)

spots = np.array([[-w/2, -h/2, 0, 1, 0, 0, 0.5],
                  [-w/2,  h/2, 0, 0, 1, 0, 0.5],
                  [ w/2,  h/2, 0, 0, 0, 1, 0.5],
                  [ w/2, -h/2, 0, 0, 1, 1, 0.5]])
glspi = gl.GLScatterPlotItem(pos=spots[:,:3], size=50, color=spots[:,3:])
glv.addItem(glspi)

pg.exec()
```

This PR may fix #825.

PR #2232, opened 2022-03-17
Fix GLImageItem regression

pijyoi:
#2131 caused a regression to GLImageItem as reported in #2230.
This PR reverts the offending commit 36b2e49cd3b0ba38a2169e8cb3b2d2d7c8b5e619 and applies only the cast to contiguous array operation that was intended in #2131.


j9ac9k:
Hi @pijyoi 

Thanks for patching up this regression so quickly.  Given all the opengl methods in the paint mechanism expect floats, almost wonder if it would be worth it to use the integer equivalents, but suppose we shouldn't go looking to fix what isn't broken.

j9ac9k:
Fixes #2230

PR #2235, opened 2022-03-23
Track when ScatterPlotItem clears the tooltip, only clear when needed

### LINKED ISSUE #2135: ScatterPlotItem: hoverable=True does not create Tooltip
Tested with python -m pyqtgraph.examples

_--> GraphicsItems --> Scatter Plot_

Changed:
`g.ScatterPlotItem(size=10, pen=pg.mkPen(None), brush=pg.mkBrush(255, 255, 255, 120))`
to:
`g.ScatterPlotItem(size=10, pen=pg.mkPen(None), brush=pg.mkBrush(255, 255, 255, 120), hoverable=True)`
or:
`g.ScatterPlotItem(size=10, pen=pg.mkPen(None), brush=pg.mkBrush(255, 255, 255, 120), hoverable=True, tip='x: {x:.3g}\ny: {y:.3g}\ndata={data}'.format)`

Expected behaviour: When hovering over the data points a tooltip should be shown.
Experienced behaviour: Tooltip is **not** shown.

Tested on:
0.12.2 on Windows
0.11.1 on Linux (Arch)

The respective code lies in ScatterPlotItem.py (l. 1224):
```
           # Show information about hovered points in a tool tip
            vb = self.getViewBox()
            if vb is not None and self.opts['tip'] is not None:
                cutoff = 3
                tip = [self.opts['tip'](x=pt.pos().x(), y=pt.pos().y(), data=pt.data())
                       for pt in points[:cutoff]]
                if len(points) > cutoff:
                    tip.append('({} others...)'.format(len(points) - cutoff))
                print('\n\n'.join(tip)) # ADDED to TEST
                vb.setToolTip('\n\n'.join(tip))

            self.sigHovered.emit(self, points, ev)
```
The added line shows that the tip is created, but it is never shown.

ixjlyons:
Fixes #2135, which has a nice script to test the changes: https://github.com/pyqtgraph/pyqtgraph/issues/2135#issuecomment-1075062167

### DIFF
--- a/pyqtgraph/graphicsItems/ScatterPlotItem.py
+++ b/pyqtgraph/graphicsItems/ScatterPlotItem.py
@@ -443,6 +443,10 @@ class ScatterPlotItem(GraphicsObject):
 
         #self.setCacheMode(self.DeviceCoordinateCache)
 
+        # track when the tooltip is cleared so we only clear it once
+        # this allows another item in the VB to set the tooltip
+        self._toolTipCleared = True
+
     def setData(self, *args, **kargs):
         """
         **Ordered Arguments:**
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks @ixjlyons for the PR this LGTM.  Merging!

PrimozGodec:
Thank you for this fix. It also solves my issue. Would it be possible to make a new release of the package that would include the fix?

PR #2236, opened 2022-03-23
Fixed the app crash on right clicked

Cosmicoppai:
Fixed the application crash when QGraphicsTextItem(node-label) is right-clicked

### DIFF
--- a/pyqtgraph/flowchart/Node.py
+++ b/pyqtgraph/flowchart/Node.py
@@ -467,7 +467,7 @@ class TextItem(QtWidgets.QGraphicsTextItem):
         if ev.button() == QtCore.Qt.MouseButton.LeftButton:
             self.setTextInteractionFlags(QtCore.Qt.TextInteractionFlag.TextEditorInteraction)
         elif ev.button() == QtCore.Qt.MouseButton.RightButton:
-            self.setTextInteractionFlags(QtCore.Qt.NoTextInteraction)
+            self.setTextInteractionFlags(QtCore.Qt.TextInteractionFlag.NoTextInteraction)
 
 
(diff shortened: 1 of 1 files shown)


pyqtgraph/flowchart/Node.py
```diff
@@ -462,6 +462,12 @@ def keyPressEvent(self, ev):
                 self.on_update()
                 return
         super().keyPressEvent(ev)
+        
+    def mousePressEvent(self, ev):
+        if ev.button() == QtCore.Qt.MouseButton.LeftButton:
+            self.setTextInteractionFlags(QtCore.Qt.TextInteractionFlag.TextEditorInteraction)
+        elif ev.button() == QtCore.Qt.MouseButton.RightButton:
+            self.setTextInteractionFlags(QtCore.Qt.NoTextInteraction)
```
j9ac9k, line 470: Slight change request, I think this should be 

```python
self.setTextInteractionFlags(QtCore.Qt.TextInteractionFlag.NoTextInteraction)
```

PyQt6 mandates that the full enum namespace is used.  On PySide2/6, and PyQt5, the abbreviated namespace works.  If you would make the change that would be great.

Cosmicoppai: Done.

Cosmicoppai: Also added the code to set **focus on text** https://github.com/pyqtgraph/pyqtgraph/pull/2236/commits/b0880ea1b723e9cb7064e3d36d07b4f0b929c987

j9ac9k:
LGTM, thanks @Cosmicoppai !

PR #2242, opened 2022-03-31
Add glow example

edumur:
This pull request add an example adding a glow and underglow effect, see https://github.com/dhaitz/mplcyberpunk.

We may discuss if we want to keep things as they are or if we want to add/remove stuff from this example.

I know mypy is not of concern here but the file is mypy compliant except for an error on line 39 that I do not understand (and I spent 0 time about it):
```python
splitter = QtWidgets.QSplitter()
```

```bash
"Module has no attribute "QSplitter"
```

Example:

glow_lines


j9ac9k:
Hi @edumur 

Thanks for this so much, this looks great.  

> I know mypy is not of concern here but the file is mypy compliant except for an error on line 39 that I do not understand (and I spent 0 time about it):

This is a known issue, we have a `__init__.pyi` stub file that goes in the `./pyqtgraph/Qt/` directory that a number of the maintainers have passed around to get type-completion working in PyCharm.  We should really add it to the library at some point.  Type-hinting is something we would like to add to the library, but I need to do more research to find out if we want to try and adopt a style/format that is compatible w/ [pydantic ](https://pydantic-docs.helpmanual.io/) or not.  If there is a pyqtgraph contributor that is reading this that has experience with pydantic, please speak up, I'd love to chat with you.  Ok, enough about type-hinting.

Nit-pick comment, the example doesn't end in a new-line character, if you could update `glow.py` so that it has a newline at the end that would be great.

The rest of the code at a glance looks good to me, I'm going to try and play w/ the example here later today and see if I can break it or see if there is something relevant to add.

Thanks so much for adding this and thank you @dhaitz for the inspiration.

j9ac9k:
I did manage to make this blow up; let me see if I can give steps to reproduce:

```python
pyqtgraph-pyqt_62-py39 ❯ python -m pyqtgraph.examples
  File "/Users/ogi/Developer/pyqtgraph/pyqtgraph/examples/glow.py", line 66, in update_plot
    ys.append(np.sin(xs[-1])*xs[-1]*10-i/3+noises[i])
```

edumur:
It has been solved.

j9ac9k:
Hi @edumur sorry this took so long for me to test, this LGTM merging.  Thanks so much for creating this!

On an unrelated note, when submitting a PR, it's helpful to use a different branch name than what exists in the target repo.  For example, I couldn't just do something like

```
git remote add edumur https://github.com/edumur/pyqtgraph.git
git fetch edumur
git switch add-glow-example  # hypothetical name
```

With submitting the PR w/ `master` as the branch name, I needed to run `git checkout -b edumur-master edumur/master` which isn't a big deal but I have to google how to checkout a local branch with a different branch name from the remote branch 😆 

As I said, not a big deal, and thanks again for submitting this PR!

PR #2247, opened 2022-04-06
Added opts to InfiniteLine so it can be included in Legend

aaron-ott:
I was in a situation where I needed to be able to add an InfiniteLine item to a graph legend, but was unable to. Perhaps this was a design choice, but I changed InfiniteLine to now have the opts component needed for Legend to create an item out of it.

### DIFF
--- a/pyqtgraph/graphicsItems/InfiniteLine.py
+++ b/pyqtgraph/graphicsItems/InfiniteLine.py
@@ -3,6 +3,7 @@ from math import atan2, degrees
 import numpy as np
 
 from .. import functions as fn
+from .. import getConfigOption
 from ..Point import Point
 from ..Qt import QtCore, QtGui
 from .GraphicsItem import GraphicsItem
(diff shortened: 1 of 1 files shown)


pijyoi:
I don't think you should copy all the options from `PlotDataItem`, particularly since most of them are non-applicable to an `InfiniteLine`.

From https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/LegendItem.py#L348-L381, it seems that the only options needed for plotting the legend would be `antialias` and `pen`.

I see that currently `InfiniteLine` is always drawn with antialiasing, so I am not sure if `opts['antialias']` should be unconditionally set to `True`.
`setPen` would need to be updated to keep the `opts['pen']` option in sync.

j9ac9k:
@aaron-ott I'm with @pijyoi on this, we likely only need a handful of the options, not all of them.

j9ac9k:
Hi @aaron-ott 

I'm going to close out this PR, as making InfiniteLine like PlotDataItem is not the right way to go about this.

Looking at the source code, looks like LegendItem by default expects a PlotDataItem's (only) to be added to it; which I do agree is overly restrictive.  There is the `itemSample` bit of code which appears to allow for adding more types to be drawn. 

That said, I don't think that's the right way to go about it either.  We likely will want to create a typing protocol, and specify it so it's a bit more clear what attributes (or better yet, methods) LegendItem needs to see in order to work.

PR #2248, opened 2022-04-08
Fixes for API changes in Qt

campagnola:
In my most recent environment (PyQt5 5.12.3 Qt 5.12.9), I started seeing errors in comparisons between MouseButton and MouseButtons (we used to be able to use `&` to determine whether a MouseButton was included in a MouseButtons, but now this generates `TypeError: unsupported operand type(s) for &: 'MouseButton' and 'MouseButtons'`).

This PR fixes a few instances of this comparison, plus a similar comparison between KeyboardModifier and KeyboardModifiers. 

Also snuck in a fix for the HDF5 exporter; seems like importlib.util is no longer imported by default.

### DIFF
--- a/pyqtgraph/examples/ExampleApp.py
+++ b/pyqtgraph/examples/ExampleApp.py
@@ -517,7 +517,7 @@ class ExampleLoader(QtWidgets.QMainWindow):
 
     def keyPressEvent(self, event):
         ret = super().keyPressEvent(event)
-        if not QtCore.Qt.KeyboardModifier.ControlModifier & event.modifiers():
+        if int(QtCore.Qt.KeyboardModifier.ControlModifier) & int(event.modifiers()):
             return ret
         key = event.key()
         Key = QtCore.Qt.Key
--- a/pyqtgraph/exporters/HDF5Exporter.py
+++ b/pyqtgraph/exporters/HDF5Exporter.py
@@ -1,4 +1,4 @@
-import importlib
+import importlib.util
 
 import numpy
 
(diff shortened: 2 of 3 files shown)


j9ac9k:
I may be remembering wrong, but `int` casting I don't think is a reliable way anymore to handle the enum/flags in newer versions of Qt.

I'd have to dig into the library to figure out what we do elsewhere.

pijyoi:
I am not getting the error though.
Putting `MouseButtons` on the left hand side seems to allow avoiding the deprecation warning (on Python 3.8).
The resulting expression can be used where a boolean is needed.

```python
In [21]: QtCore.PYQT_VERSION_STR
Out[21]: '5.12.3'

In [22]: MB = QtCore.Qt.MouseButton

In [23]: btns = MB.LeftButton | MB.RightButton

In [24]: btns & MB.LeftButton
Out[24]: <PyQt5.QtCore.Qt.MouseButtons at 0x204a452c820>

In [25]: MB.LeftButton & btns
<ipython-input-25-27340288dfca>:1: DeprecationWarning: an integer is required (got type MouseButtons).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  MB.LeftButton & btns
Out[25]: <PyQt5.QtCore.Qt.MouseButtons at 0x204a4495580>

In [26]: bool(btns & MB.LeftButton)
Out[26]: True

In [27]: bool(btns & MB.MiddleButton)
Out[27]: False
```

campagnola:
Closing in favor of #2250

PR #2249, opened 2022-04-08
Fix UFuncTypeError when plotting integer data on windows

campagnola:
When plotting integer arrays on windows, I ran into the following:
```
  File "pyqtgraph\pyqtgraph\graphicsItems\PlotDataItem.py", line 1057, in getDisplayDataset
    y = fn.clip_array(y, min_val, max_val)
  File "pyqtgraph\pyqtgraph\functions.py", line 1223, in clip_array
    out = np.core.umath.minimum(arr, vmax, out=out)
numpy.core._exceptions.UFuncTypeError: Cannot cast ufunc 'minimum' output from dtype('float64') to dtype('int32') with casting rule 'same_kind'
```
This is because we are asking (on ine 1223) for the minimum of an integer array and a float scalar, but the output array has already been allocated as integer. The fix here instead asks numpy to select a compatible dtype between the two.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1218,9 +1218,10 @@ def clip_array(arr, vmin, vmax, out=None):
         return np.core.umath.maximum(arr, vmin, out=out)
     elif _win32_clip_workaround_needed:
         if out is None:
-            out = np.empty_like(arr)
+            out = np.empty(arr.shape, dtype=np.find_common_type([arr.dtype], [type(vmax)]))
         out = np.core.umath.minimum(arr, vmax, out=out)
         return np.core.umath.maximum(out, vmin, out=out)
+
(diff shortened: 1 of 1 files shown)


j9ac9k:
Hi @campagnola would you mind renaming your PR title to something a bit more descriptive? This will be used in the automatic changes generation when we do the next release 👍🏻

pijyoi:
Do you think it would be better to just remove the win32 codepath? Since numpy 1.22, it is no longer needed.

And in fact, the workaround is still slower than the other platforms.

j9ac9k:
> 

We'll remove it soon enough when we can stop supporting number < 1.22 (June 23, 2023).  For now, we should probably keep it in there given the huge performance disparity we saw to begin with.

PR #2250, opened 2022-04-09
test enum using : "enums & enum"

pijyoi:
It appears that some warnings for PyQt5 5.12 and Python 3.8 were being deliberately masked.
These masked warnings seem to be the source of trouble that #2248 was attempting to fix.

This PR attempts to fix the underlying issue and allows us to remove the masked warnings.
The fix is by changing `enum & enums` to `enums & enum`.
On PyQt5 5.12, `enum` is an integer and thus does the `&` operation by coercing the rhs to an integer, which is deprecated or no longer allowed with newer Python.

@campagnola , could you test whether this fixes the issue in #2248?

### DIFF
--- a/pyqtgraph/examples/ExampleApp.py
+++ b/pyqtgraph/examples/ExampleApp.py
@@ -517,7 +517,7 @@ class ExampleLoader(QtWidgets.QMainWindow):
 
     def keyPressEvent(self, event):
         ret = super().keyPressEvent(event)
-        if not QtCore.Qt.KeyboardModifier.ControlModifier & event.modifiers():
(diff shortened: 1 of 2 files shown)


pijyoi:
MWE:
1) using PyQt5 5.12 and Python 3.8
2) `python -Wdefault -m pyqtgraph.examples`
3) press Ctrl-F

On master branch (but not with this PR), the following warning will be emitted:
```python
c:\XXX\pyqtgraph\pyqtgraph\examples\ExampleApp.py:520: DeprecationWarning: an integer is required (got type KeyboardModifiers).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  if not QtCore.Qt.KeyboardModifier.ControlModifier & event.modifiers():
```

PR #2253, opened 2022-04-13
Make auto downsample factor calculation more robust

StSav012:
The proposed changes enable ignoring the infinite and NaN values of `x` and avoiding a division by zero if `self.opts['autoDownsampleFactor'] == 0`.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -977,11 +977,9 @@ class PlotDataItem(GraphicsObject):
             if view_range is not None and len(finite_x) > 1:
                 dx = float(finite_x[-1]-finite_x[0]) / (len(finite_x)-1)
                 if dx != 0.0:
-                    x0 = (view_range.left()-finite_x[0]) / dx
-                    x1 = (view_range.right()-finite_x[0]) / dx
                     width = self.getViewBox().width()
                     if width != 0.0:  # autoDownsampleFactor _should_ be > 1.0
-                        ds_float = max(1.0, (x1 - x0) / (width * self.opts['autoDownsampleFactor']))
+                        ds_float = max(1.0, abs(view_range.width() / dx / (width * self.opts['autoDownsampleFactor'])))
                         if math.isfinite(ds_float):
                             ds = int(ds_float)
                     ## downsampling is expensive; delay until after clipping.
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -979,8 +979,10 @@ class PlotDataItem(GraphicsObject):
                     x0 = (view_range.left()-finite_x[0]) / dx
                     x1 = (view_range.right()-finite_x[0]) / dx
                     width = self.getViewBox().width()
-                    if width*self.opts['autoDownsampleFactor'] != 0.0 and np.isfinite(x1-x0):
-                        ds = int(max(1, int((x1-x0) / (width*self.opts['autoDownsampleFactor']))))
+                    if width * self.opts['autoDownsampleFactor'] != 0.0:
+                        ds_float = max(1.0, (x1 - x0) / (width * self.opts['autoDownsampleFactor']))
+                        if np.isfinite(ds_float):
+                            ds = int(ds_float)
                     ## downsampling is expensive; delay until after clipping.
 
         if self.opts['clipToView']:
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -971,13 +971,15 @@ class PlotDataItem(GraphicsObject):
 
         if self.opts['autoDownsample']:
             # this option presumes that x-values have uniform spacing
-            if view_range is not None and len(x) > 1:
-                dx = float(x[-1]-x[0]) / (len(x)-1)
+
+            finite_x = x[np.isfinite(x)]  # ignore infinite and nan values
+            if view_range is not None and len(finite_x) > 1:
+                dx = float(finite_x[-1]-finite_x[0]) / (len(finite_x)-1)
                 if dx != 0.0:
-                    x0 = (view_range.left()-x[0]) / dx
-                    x1 = (view_range.right()-x[0]) / dx
+                    x0 = (view_range.left()-finite_x[0]) / dx
+                    x1 = (view_range.right()-finite_x[0]) / dx
                     width = self.getViewBox().width()
-                    if width != 0.0:
+                    if width*self.opts['autoDownsampleFactor'] != 0.0 and np.isfinite(x1-x0):
                         ds = int(max(1, int((x1-x0) / (width*self.opts['autoDownsampleFactor']))))
                     ## downsampling is expensive; delay until after clipping.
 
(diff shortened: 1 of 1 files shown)


j9ac9k:
Hi @StSav012 

Thanks for the PR, this will definitely make an improvement.  I'm a little concerned with the equality check at the bottom of your diff 

```python
if width*self.opts['autoDownsampleFactor'] != 0.0 and np.isfinite(x1-x0):
```

I know the code you're replacing was `if width != 0.0` but I'm wondering if you think using something like `not math.isclose()` would be more appropriate here.  I'm ok with merging as is, but wanted to get your thoughts on the equality check first.

StSav012:
Hi @j9ac9k 

It's a good point. I'm not confident how to tell a sufficient tolerance from a bad one. After your question, I'd rather test that `(x1-x0) / (width*self.opts['autoDownsampleFactor'])` is still finite. This way there is no need for any tolerance. Kind of like this:

```python
                    if width * self.opts['autoDownsampleFactor'] != 0.0:
                        ds_float = (x1 - x0) / (width * self.opts['autoDownsampleFactor'])
                        if np.isfinite(ds_temp):
                            ds = int(max(1, int(ds_float)))
```

And do we require both `int`s at the line above?

j9ac9k:
`width` should always be of type float,  and `autoDownsampleFactor` can be float or integer.  `autoDownsampleFactor` _should_ be > 1.0, so we shouldn't run into any weirdness, so this is likely a non-issue.

Also for singular values, I generally prefer to use `math.isfinite` vs. `numpy.isfinite` but there was an issue some time ago where it turns out the two behave differently (I should try and search the issue/pr tracker for more specifics).

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -971,14 +972,18 @@ def getDisplayDataset(self):
...
+                dx = float(finite_x[-1]-finite_x[0]) / (len(finite_x)-1)
                 if dx != 0.0:
-                    x0 = (view_range.left()-x[0]) / dx
-                    x1 = (view_range.right()-x[0]) / dx
+                    x0 = (view_range.left()-finite_x[0]) / dx
+                    x1 = (view_range.right()-finite_x[0]) / dx
                     width = self.getViewBox().width()
-                    if width != 0.0:
-                        ds = int(max(1, int((x1-x0) / (width*self.opts['autoDownsampleFactor']))))
+                    if width != 0.0:  # autoDownsampleFactor _should_ be > 1.0
```
j9ac9k, line 983: Hi @StSav012 

I think for this equality comparison to 0.0, I think we should use `math.isclose(width, 0.0, abs_tol=1e-9)`

j9ac9k:
Hi @StSav012 

I chatted w/ @outofculture today, and we agreed on the equality comparison to 0.0 being problematic and settled on a tolerance we felt was appropriate.  If youi don't mind making that change, I'll merge this right after 👍🏻

pyqtgraph/graphicsItems/PlotDataItem.py
outofculture: With a `not` in front of it.

StSav012: Hi, @j9ac9k @outofculture 

I can't agree with the approach you propose. Not yet, at least.

The check for infinity takes care of the cases when `width` is not exactly zero, but close to it. I'm afraid, with `isclose`, the behavior would not be straightforward; it would depend on the tolerance picked essentially at random. Say, if the `x` span is huge (like `1e308`), we might still get the infinity, even if `width` is significantly non-zero. However, if `width` is exactly `0.0`, the function fails with `ZeroDivisionError`, and the rest of the function gets ignored. On the other hand, if `view_range.right() - view_range.left()` (why do we need `x0` and `x1` at all?) is small (say, equal to the `isclose` tolerance), the picked tolerance might be far too rough.

j9ac9k: My understanding is that `width` here wasn't in the view range coordinates, but corresponds to `QGraphicsWidget.geometry().width()` which should return the `QRectF` that represents the position and size of the widet, and not the view space. 

That said, the tolerance picked effectively at random so you're not wrong there...

outofculture: Hmm. Looking further, `QRect.width()` should be an [integer](https://doc.qt.io/qtforpython/PySide6/QtCore/QRect.html#PySide6.QtCore.PySide6.QtCore.QRect.width), so I don't think this protection will ever matter either way. We could wrap it in a `try`/`except ZeroDivisionError: pass`, even, if that's all we're guarding against, with the added possible bonus of protecting against `self.opts['autoDownsampleFactor'] == 0`, too. That leaves the question of whether we need to treat e.g. `1 / ((0.3 - 0.2) - 0.1)` as a divide-by-zero, which we would normally want to do. In this case, if a application sets `self.opts['autoDownsampleFactor']` to e.g. 1e-15, they're going to get bad behavior ("please _up-sample_ my data"?), so we can probably safely ignore this aspect of the problem. Our knee-jerk resistance to comparing floats to specific values likely doesn't matter either way here, but the sentiment is a good one, so it might be worthwhile to put in the `isclose` just to uphold that principle.

j9ac9k: > Hmm. Looking further, `QRect.width()` should be an [integer](https://doc.qt.io/qtforpython/PySide6/QtCore/QRect.html#PySide6.QtCore.PySide6.QtCore.QRect.width), so I don't think this protection will ever matter either way. 


`QRect.width()` should absolutely be an integer, but we're dealing w/ a `QRectF` here ;)

https://doc-snapshots.qt.io/qt6-dev/qgraphicslayoutitem.html#geometry

> We could wrap it in a `try`/`except ZeroDivisionError: pass`, even, if that's all we're guarding against, with the added possible bonus of protecting against `self.opts['autoDownsampleFactor'] == 0`, too. 

nit-pick level 100; if we do that we should use 

```python
import contextlib

...

with contextlib.suppress(ZeroDivisionError):
    # do the division here
```

> That leaves the question of whether we need to treat e.g. `1 / ((0.3 - 0.2) - 0.1)` as a divide-by-zero, which we would normally want to do. In this case, if a application sets `self.opts['autoDownsampleFactor']` to e.g. 1e-15, they're going to get bad behavior ("please _up-sample_ my data"?), so we can probably safely ignore this aspect of the problem. Our knee-jerk resistance to comparing floats to specific values likely doesn't matter either way here, but the sentiment is a good one, so it might be worthwhile to put in the `isclose` just to uphold that principle.

Yeah, @StSav012 is modifying existing code here that has some code I absolutely hate, and I'm requesting they change something else that was already in the library; probably not the most fair ask.  I'm good w/ leaving it for this PR, and addressing the float equality comparisons in another PR.

j9ac9k:
Closing/Opening to trigger new CI run.

j9ac9k:
CI all green, merging, sorry for the wait @StSav012 thanks for the PR!

Mriv31:
```np.isfinite``` is actually extremely slow and is needlessly called each time we recompute the ds parameter (ie update the viewrange). 

If you have a very large datasets (in my case 10^7 pts) and are navigating into small portions of it using cliptoview and autodownsample, you do not want to recompute ```np.isfinite`` of the whole dataset everytime you change the x range. 

In my case removing the isfinite check allows a gain of 0.4s when updating the region of my dataset I am looking at. Replacing the two occurences of searchsorted in the whole dataset removes another 0.4s (but only works if x proportional to the index).

Those two functions are the main performance problem when zooming in and navigating into very large datasets.

Maybe ```isfinite```  be checked only once for a given dataset -  upon loading ? and  only finite values kept for what happens next. What do you thing ?

PR #2254, opened 2022-04-13
Make ValueLabel formatting consistent with SpinBox

StSav012:
That's the ValueLabel I'd like to have.


j9ac9k:
Going to tag @ixjlyons on this one since SpinBox has become his baby

j9ac9k:
Hi @StSav012 

Sorry I haven't given you a response/review in some time here... thanks for the PR! 

I still need to go over this closer, but I want to point out that pyqtgraph generally doesn't use the python property/getter/setters, as we adopt the Qt API naming convention by using methods with the associated property name like

```python

def brush(self) -> QBrush:
   return self._brush

def setBrush(self, brush: QBrush) -> None:
    self._brush = brush
    return None
```

The pyqtgraph code-base does have the use of the property decorator in a few places, but that's primarily for backward compatibility issues.  Anyway, I'll try and read through this and give you some more thorough feedback shortly.

ixjlyons:
These additions make sense to me. Aside from the getter/setter stuff mentioned above, I also wonder if it makes sense to refactor this and `SpinBox` to share some code.

j9ac9k:
@StSav012 can you convert the python properties to Qt style setter and getter methods, I'll merge this.

For example, for the `value` property it would be 

```python
def setValue(self, value):
    self._value = value 
    ....

def value(self):
    return self._value
```

StSav012:
@j9ac9k, I used Python properties to keep the class compatible with old code. In `pyqtgraph==0.13.1`, I can change `averageTime`, `suffix`, `siPrefix`, and `formatStr` of a `ValueLabel` instance directly. As for `value`, `format`, `decimals`, `prefix`, and `fancyMinus`, it would not hurt to get rid of the properties indeed. However, then the syntax would not be consistent, and I'd like that much less.

In theory, I can compose monstrous classes that would behave like a built-in Python float or string, but could be called and, then, syntax like `self.value()` would be possible. The following code is not for weakhearted.

```python
def make_callable(x):
    if callable(x):
        return x  # don't break the callable

    class _Callable(type(x)):
        def __call__(self) -> type(x):
            return self

    return _Callable(x)


class Foo:
    def __init__(self):
        self._egg = 1
        self._spam = 'python'

    @property
    def egg(self):
        return make_callable(self._egg)

    @egg.setter
    def egg(self, new_value) -> None:
        self._egg = new_value

    def setEgg(self, new_value) -> None:
        self._egg = new_value

    @property
    def spam(self):
        return make_callable(self._spam)

    @spam.setter
    def spam(self, new_value) -> None:
        self._spam = new_value

    def setSpam(self, new_value) -> None:
        self._spam = new_value


foo = Foo()

# access any way
print(foo.egg, foo.spam())
print(foo.egg(), foo.spam)

# set as a property
foo.egg = 2
foo.spam = foo.spam().capitalize()
print(foo.egg, foo.spam())
print(foo.egg(), foo.spam)

# set in Qt style
foo.setEgg(foo.egg + 1)
foo.setSpam(foo.spam[::-1])
print(foo.egg, foo.spam())
print(foo.egg(), foo.spam)
```

PR #2259, opened 2022-04-15
Avoid import error in HDF5 exporter

campagnola:
`importlib.util` is no longer imported by default with `importlib`; this has led to an import error in the HDF5 exporter.


ksunden (approved):

j9ac9k:
Thanks @campagnola !

PR #2260, opened 2022-04-15
Fix : QPoint() no longer accepts float arguments

campagnola:


j9ac9k:
Thanks for the PR @campagnola LGTM, merging.

pijyoi:
Strangely, the deprecation warning appears only on PyQt{5,6} but not on PySide{2,6}.
It might be cleaner to implement it as:
```python
pos = ev.screenPos()
menu.popup(pos.toPoint())
```
Although the meaning would be slightly different, since `toPoint()` does a round. https://doc.qt.io/qt-5/qpointf.html#toPoint

PR #2263, opened 2022-04-18
add support for PySide6 6.3.0 QOverrideCursorGuard

pijyoi:
This fixes `BusyCursor` such that `test_busycursor.py` passes for PySide6 6.3.0.

A similar fix would be required for `ProgressDialog` which also makes use of `setOverrideCursor()`. Not being familiar with the workings of `ProgressDialog`, this PR limits itself to `BusyCursor` only.

### DIFF
--- a/pyqtgraph/widgets/BusyCursor.py
+++ b/pyqtgraph/widgets/BusyCursor.py
@@ -18,10 +18,15 @@ def BusyCursor():
     """
     app = QtCore.QCoreApplication.instance()
     in_gui_thread = (app is not None) and (QtCore.QThread.currentThread() == app.thread())
+    need_cleanup = in_gui_thread
     try:
         if in_gui_thread:
-            QtWidgets.QApplication.setOverrideCursor(QtGui.QCursor(QtCore.Qt.CursorShape.WaitCursor))
+            guard = QtWidgets.QApplication.setOverrideCursor(QtGui.QCursor(QtCore.Qt.CursorShape.WaitCursor))
+            if hasattr(QtGui, 'QOverrideCursorGuard') and isinstance(guard, QtGui.QOverrideCursorGuard):
+                # on PySide6 6.3.0, setOverrideCursor() returns a QOverrideCursorGuard context manager
+                # object that calls restoreOverrideCursor() for us
+                need_cleanup = False
         yield
     finally:
-        if in_gui_thread:
+        if need_cleanup:
             QtWidgets.QApplication.restoreOverrideCursor()


pijyoi:
Due to `test_parametertypes.py` causing a segfault before `test_busycursor.py` gets a chance to run, the CI won't show a positive result for this PR.

j9ac9k:
Great news, I can replicate the segfault locally; but that's another issue.

Thanks for the PR @pijyoi merging.

pijyoi:
Looks like I misunderstood how context managers are supposed to work.

The `__enter__` and `__exit__` methods are only executed when used with the `with` keyword. (i.e. line [3] below).

In line [4] below, the `__enter__` method would not have been executed. Yet a `guard` object is returned that does its cleanup when it loses its last reference. This is the behavior that this PR was relying on.

```python
In [1]: from PySide6 import QtCore, QtGui, QtWidgets

In [2]: app = QtGui.QGuiApplication([])

In [3]: with app.setOverrideCursor(QtCore.Qt.CursorShape.WaitCursor) as ctx:
   ...:     print(ctx)
   ...:
<PySide6.QtGui.QOverrideCursorGuard object at 0x000002308B288240>

In [4]: guard = app.setOverrideCursor(QtCore.Qt.CursorShape.WaitCursor)

In [5]: print(guard)
<PySide6.QtGui.QOverrideCursorGuard object at 0x000002308B0E73C0>
```

pijyoi:
From the sources:
https://github.com/qtproject/pyside-pyside-setup/blob/6.3.0/sources/pyside6/PySide6/qtguihelper.h
https://github.com/qtproject/pyside-pyside-setup/blob/6.3.0/sources/pyside6/PySide6/QtGui/typesystem_gui_common.xml#L2674
https://github.com/qtproject/pyside-pyside-setup/blob/6.3.0/sources/pyside6/PySide6/QtGui/typesystem_gui_common.xml#L3113
https://github.com/qtproject/pyside-pyside-setup/blob/6.3.0/sources/pyside6/PySide6/glue/qtgui.cpp#L663

It does appear that the implementation has changed the meaning of `QtGui.QGuiApplication.setOverrideCursor()` to return a `QOverrideCursorGuard` object which does self-cleanup when the object gets destroyed.
In the case where the returned object doesn't get assigned to a variable, it gets destroyed immediately. Since existing code expects `setOverrideCursor() -> None`, they wouldn't have assigned it to a variable. i.e. Existing code becomes broken.

pijyoi:
> i.e. Existing code becomes broken.

The broken behavior was reported by someone here: https://bugreports.qt.io/browse/PYSIDE-2039
And subsequently fixed here: https://code.qt.io/cgit/pyside/pyside-setup.git/commit/?id=287dc1365251a15e95363366d9c86bb8f5962ecb

The current pyqtgraph `BusyCursor` code (patched in #2349) should continue to work however.

Issue #2270, opened 2022-04-23
PyQtGraph Largely Incompatible with PyPy


j9ac9k:
### Short description

Getting segfaults when using widgets based on GraphicsView, Widgets using QOpenGLWidget appear to function as intended.

### Code to reproduce

When running examples that use `pyqtgraph.GraphicsWidget` we get a segfault/

```bash
python pyqtgraph/examples/PlotSpeedTest.py
```

### Expected behavior

No segfault

### Real behavior

```
pyqtgraph-pypy ❯ python pyqtgraph/examples/PlotSpeedTest.py
zsh: segmentation fault  python pyqtgraph/examples/PlotSpeedTest.py
```

### Tested environment(s)

 * PyQtGraph version: `master` commit: 74fa36da37d36a3da52019f673898ff0fe165de0
 * Qt Python binding: PySide6 6.3 PyPy
 * Python version: 3.8.13 - pypy
 * NumPy version: 1.22.3 - pypy
 * Operating system: macOS
 * Installation method: pip

### Additional context

PySide6 installed via

```bash
pip install pyside6 --index-url=https://download.qt.io/official_releases/QtForPython/pypy/
```

NumPy was built from source.

j9ac9k:
Walking through the debugger, the segfault occurs on this line:

https://github.com/pyqtgraph/pyqtgraph/blob/74fa36da37d36a3da52019f673898ff0fe165de0/pyqtgraph/graphicsItems/GraphicsWidget.py#L16

```python
> /Users/ogi/Developer/pyqtgraph/pyqtgraph/graphicsItems/GraphicsWidget.py(17)__init__()
-> QtWidgets.QGraphicsWidget.__init__(self, *args, **kargs)
(Pdb) widget = QtWidgets.QGraphicsWidget()
(Pdb) widfget = QtWidgets.QGraphicsWidget.__init__(self)
zsh: segmentation fault  python pyqtgraph/examples/PlotSpeedTest.py
```

j9ac9k:
Ok, so this is not a pyqtgraph specific issue; but I think we can work-around it with some effort/work:

```python
from PySide6.QtWidgets import QGraphicsWidget

class A:
	pass

class B(A, QGraphicsWidget):
	pass

class C(QGraphicsWidget):
	pass

class D(QGraphicsWidget, A):
	pass

if __name__ == "__main__":
	# b = B()  # segfault
	c = C()  # ok
	d = D()  # ok
	print("All done!")
```

PR #2271, opened 2022-04-24
try getting PyPySide to work

pijyoi:
With regards to #2270, experiment with swapping inheritance order to get GraphicsItem(s) to work on PyPy.

~Except for PySide6 (tested on 6.2.4 and 6.3.0), the patches here cause all other CPython bindings to fail or crash.
Note: The combination of Python 3.10 + PySide6 6.3.0 already fails for other reasons.~

The various GraphicItem(s) seem to work.
Dragging of ROI Handles don't work.

The existing codebase does assume the method resolution order of inheriting `GraphicsItem` before `QGraphicsWidget` / `QGraphicsObject`.


pijyoi:
So the reason for #2132 is also the reason why CPython bindings were crashing once the inheritance order was swapped.

j9ac9k:
The ROI Issue seems limited to the `Handle` class, and not the parent `UIGraphicsItem` class, as `TargetItem` (which also inherits from `UIGraphicsItem`) behaves as intended.

pijyoi:
As far as I can tell, both `Handle` and `_PolyLineSegment` are not getting their `hoverEvent` called on PyPy.
In the extract of `ROIExamples.py` below, you can neither drag the existing handles nor add new handles.

```python
import pyqtgraph as pg

pg.mkQApp()
pw = pg.PlotWidget()
pw.show()
    
r2a = pg.PolyLineROI([[0,0], [10,10], [10,30], [30,10]], closed=True)
pw.addItem(r2a)
r2b = pg.PolyLineROI([[0,-20], [10,-10], [10,-30]], closed=False)
pw.addItem(r2b)
pw.disableAutoRange('xy')
pw.autoRange()

pg.exec()
```

pijyoi:
By adding a judicious `print` to `GraphicsScene.sendClickEvent()`, it appears that the ROI `Handle` object got decayed to a `QGraphicsItem` on PyPy, and thus doesn't get any mouse events delivered to it.

CPython PySide 6.3.0
```python
(py38_pyside6) PS C:\work\repos\pyqtgraph> python C:\work\pyqtgraph\simple_roi.py
<PySide6.QtWidgets.QGraphicsRectItem(0x21742530780, parent=0x21742892a00, pos=0,0, z=1000) at 0x0000021743462880> False
<pyqtgraph.graphicsItems.PlotItem.PlotItem.PlotItem(0x217428919f0, pos=0,0, flags=(ItemUsesExtendedStyleOption|ItemSendsGeometryChanges)) at 0x000002174345EC80> False
<pyqtgraph.graphicsItems.ROI.Handle(0x21742aa7160, parent=0x21742a91c50, pos=15,-15, z=11, flags=(ItemSendsGeometryChanges|ItemSendsScenePositionChanges)) at 0x00000217460E7E80> True
<ViewBox(0x217428929f0, parent=0x21742891a00, pos=23.7344,1, z=-100, flags=(ItemIsFocusable|ItemClipsChildrenToShape|ItemUsesExtendedStyleOption|ItemSendsGeometryChanges)) at 0x000002174345EFC0> True
```

PyPy PySide6 6.3.0
```python
(pypy) C:\work\repos\pyqtgraph>python \work\pyqtgraph\simple_roi.py
<PySide6.QtWidgets.QGraphicsRectItem(0x2c4d8ab3cd0, parent=0x2c4d7517f90, pos=0,0, z=1000) at 0x000002C4D8AB3C90> False
<pyqtgraph.graphicsItems.PlotItem.PlotItem.PlotItem(0x2c4d7473790, pos=0,0, flags=(ItemUsesExtendedStyleOption|ItemSendsGeometryChanges)) at 0x000002C4D8AB3010> False
<PySide6.QtWidgets.QGraphicsItem(0x2c4d7c6a260, parent=0x2c4d7c65ac0, pos=15,-15, z=11, flags=(ItemSendsGeometryChanges|ItemSendsScenePositionChanges)) at 0x000002C4D8CF8F20> False
<ViewBox(0x2c4d7517f80, parent=0x2c4d74737a0, pos=23.7344,1, z=-100, flags=(ItemIsFocusable|ItemClipsChildrenToShape|ItemUsesExtendedStyleOption|ItemSendsGeometryChanges)) at 0x000002C4D8AB3110> True
```

j9ac9k:
I guess that would explain it; probably need to sprinkle in some `print(type(self))` statements throughout and see if we can see when the type changes.

j9ac9k:
Most confusion ...with a slight modification of your example:

```python
from PySide6 import QtWidgets
import contextlib
import pyqtgraph as pg

scene = QtWidgets.QGraphicsScene(0, 0, 400, 400)
view = QtWidgets.QGraphicsView(scene)

roi = pg.LineSegmentROI([[15, -15], [30, -15]])
scene.addItem(roi)

items = view.scene().items()
for item in items:
    print(f"{type(item)=}")
print([type(handle['item']) for handle in roi.handles])
print([(handle['item'].scene() is view.scene()) for handle in roi.handles])
```

This indicates the handles are still of type `Handle`, not `QGraphicsItem`.  Furthermore, it shows that they are in the same scene, not sure why `QGraphicsScene.items()` is not showing them there...

pijyoi:
If we do a `shiboken6.dump(item)` on the returned items, on PyPy, we see something like
```python
type(item)=<class 'PySide6.QtWidgets.QGraphicsItem'>
C++ address....... PySide6.QtWidgets.QGraphicsItem/0000016EAA88D530
hasOwnership...... 0
containsCppWrapper 0
validCppObject.... 1
wasCreatedByPython 0
```
where on CPython, `containsCppWrapper` and `wasCreatedByPython` are both 1.

pijyoi:
I am only just "discovering" that PyQt but not PySide does the equivalent of a `super()`, thus automatically calling the `__init__` methods of inherited classes coming after any `QObject`s.
Thus in the snippet below, putting mixin classes after a `QObject` would have the mixin's `__init__` method being called twice on PyQt.
It does happen that a `GraphicsItem` having its `__init__` being called twice has no side-effects.

To avoid `M.__init__` being called twice if we *really* wanted to move the `M` class before the `QObject`, we would need to *not* call `M.__init__` manually on PyQt. 

```python

class M:
    def __init__(self):
        print("M.__init__")

class C1(M, QtCore.QObject):
    def __init__(self):
        M.__init__(self)
        QtCore.QObject.__init__(self)

class C2(QtCore.QObject, M):
    def __init__(self):
        QtCore.QObject.__init__(self)
        M.__init__(self)

print('C1')
c1 = C1()
print('C2')
c2 = C2()
```

j9ac9k:
> I am only just "discovering" that PyQt but not PySide does the equivalent of a `super()`, thus automatically calling the `__init__` methods of inherited classes coming after any `QObject`s.
> Thus in the snippet below, putting mixin classes after a `QObject` would have the mixin's `__init__` method being called twice on PyQt.

Huh, I had no idea about this.  I wonder if this explains some past weird behavior. 

> It does happen that a `GraphicsItem` having its `__init__` being called twice has no side-effects.

I'm wondering if this should be explicitly stated somewhere, like a comment in the `GraphicsItem` docstring or something.

pijyoi:
>I'm wondering if this should be explicitly stated somewhere, like a comment in the GraphicsItem docstring or something.

The current code in master has `GraphicsItem` as the first class to be inherited, so its `__init__` will only get called once.
It's only in this PR that `GraphicsItem` is put as the second class; this results in `__init__` getting called twice on PyQt.

I think maybe for `DockDrop`, having its `__init__` being called twice (on PyQt) might have a side effect; as it instantiates a `DockAreaOverlay`. The second call to `__init__` replaces the first instance. Note that this is already the case on master.

pijyoi:
There's something wrong on PyPySide6 when running the `dockarea.py` example.
To reproduce: Drag dock6 to somewhere else other than dock4. Then drag dock6 to dock4.
```python
  File "c:\work\repos\pyqtgraph\pyqtgraph\widgets\GraphicsView.py", line 137, in paintEvent
    return super().paintEvent(ev)
TypeError: 'PySide6.QtWidgets.QGraphicsView.paintEvent' called with wrong argument types:
  PySide6.QtWidgets.QGraphicsView.paintEvent(QPainter)
Supported signatures:
  PySide6.QtWidgets.QGraphicsView.paintEvent(PySide6.QtGui.QPaintEvent)
```

pijyoi:
`TypeError: 'QByteArray' does not support the buffer interface`

This affects `arrayToQPath` with `connect="pairs"` (ErrorBarItem) or `connect="array"`.
A quick workaround here is to enable the experimental mode of #2324.
Alternatively, we could configure `segmentedLineMode="on"` (added in #2185).

j9ac9k:
Probably should open a bug-report with the pyside folks on this.  @stonebig it seem you have a vested interest in pypy support, would you mind creating a bug report on the pyside ticket tracker for the issue @pijyoi detailed here

> `TypeError: 'QByteArray' does not support the buffer interface`

?

pijyoi:
The `QByteArray` is a small issue.

The big issue is the crashing when the QWidget is not listed as the left-most class in multiple inheritance. This breaks the recommended code pattern of placing Mix-In classes on the left-hand-side.

j9ac9k:
> The `QByteArray` is a small issue.
> 
> The big issue is the crashing when the QWidget is not listed as the left-most class in multiple inheritance. This breaks the recommended code pattern of placing Mix-In classes on the left-hand-side.

I got it in my head that that may have been patched in 6.3.1; probably should file a bug-report for that too; I alerted the pyside maintainer that has been working on this on the gitter chat, but it may have been lost in the noise.

pijyoi:
A minimal Parameter Tree test-case that passes on CPython but fails on PyPy.
This is actually caught in the existing tests (`test_parametertypes.py::test_types`), but the tests were modified in this PR to bypass it for PyPy.

The issue can be described as follows:
For the "str" Parameter type, setting a brush or a pen will end up with an empty string when the value is retrieved.
Setting a color works though.
In addition, the issue only occurs when the "str" Parameter is part of a tree, not when it is standalone.

The baffling thing is that all objects are converted to str type when it gets set to the Parameter anyway, so it shouldn't have mattered at all what the original object was.
Tagging @ntjess.

Of course, it probably doesn't make practical sense to be setting a brush or a pen to a "str" Parameter to begin with.

```python
import pyqtgraph as pg
import pyqtgraph.parametertree as pt

app = pg.mkQApp()

def check(param):
    objs = [pg.mkColor('k'), pg.mkBrush('k'), pg.mkPen('k')]
    results = []
    for obj in objs:
        param.setValue(obj)
        results.append(str(obj) == param.value())
    return results

def test_param():
    param = pt.Parameter.create(name='params', type='str')
    results = check(param)
    assert all(results), results

def test_tree():
    root = pt.Parameter.create(name='params', type='group', children=[
        dict(name='str', type='str')
    ])
    tree = pt.ParameterTree()
    tree.setParameters(root)

    param = root.child('str')
    results = check(param)
    assert all(results), results
```

pijyoi:
I have a more minimal test-case to demonstrate the ROI-Handle decay issue in `tests_pypy/test_roi_handle_decay.py`.
In essence, it boils down to: "a `GraphicsObject` that has a parent gets decayed to `QGraphicsItem`".
In the pyqtgraph library, it appears that the only `GraphicsObject` that has a parent is a `ROI.Handle`.

pijyoi:
Triggered a CI-run with PyPySide 6.3.2. No change in failing test cases.
Note: one of the failing test-cases is related to https://bugreports.qt.io/browse/PYSIDE-1991

j9ac9k:
I profiled this branch with 1000 iterations of PlotSpeedTest.py to try and better see where the discrepancy in performance is.

Results are a bit puzzling, the call tree in CPython looks like what I would expect, but in PyPy it's a decent amount different.  Also I'm not seeing entries for methods in the PyPy table that correspond to `arrayToQPath` and such

## CPython 

cpython-plotspeedtest

Top 10 entries sorted by "own time"

image


## PyPy

pypy-plotspeedtest

Top 10 entries sorted by "own time"

image

pijyoi:
Triggered a CI run with PyPySide 6.4.0 using the PyPy 3.9 wheels available on pypi.org.
Numpy wheels on pypi.org are for PyPy 3.8, so this causes a build from source of numpy.

No change in failing test cases.

pijyoi:
A fix for the crash has been merged (https://bugreports.qt.io/browse/PYSIDE-2264) but will only arrive in PySide6 6.5

pijyoi:
The current PySide release is 6.8.0, but the last binary pypy release in https://download.qt.io/official_releases/QtForPython/pyside6-essentials/ is for PySide 6.5.2 only.

PR #2274, opened 2022-04-27
Added possibility to use custom dock labels

ardiloot:
Changing the style of dock labels is currently very limited, only font size can be adjusted (i.e colors etc cannot be adjusted). To allow more customization of the labels, this pull request:

1. Includes `DockLabel` class to `dockarea` module so it could be imported and customized by subclassing
2. Added `label` argument to `Dock` class so that custom label can be specified.

There are no breaking changes, only one extra parameter (`label`) for `Dock` class.

Example:
```python
from pyqtgraph.dockarea import Dock, DockLabel

class CustomDockLabel(DockLabel):
    def updateStyle(self):
        ...

label = CustomDockLabel("name")
dock = Dock("name", label=label)

```

### DIFF
--- a/pyqtgraph/dockarea/Dock.py
+++ b/pyqtgraph/dockarea/Dock.py
@@ -10,14 +10,17 @@ class Dock(QtWidgets.QWidget, DockDrop):
     sigStretchChanged = QtCore.Signal()
     sigClosed = QtCore.Signal(object)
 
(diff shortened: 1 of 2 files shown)


j9ac9k:
Hi @ardiloot 

Thanks for the PR!  IMO Docks are an underutilized portion of this library, likely in large part due to the missing functionality that this PR aims to address.  Thanks so much!

jobusch:
This was really useful. Can you give me a hint how I can add some padding above and below the text of the label. by subclassing Or in another words: Increasing the height of the label without changing the text font size.

PR #2275, opened 2022-04-27
Fixed division by zero when no pixmap is loaded

StSav012:
`QtGui.QPixmap().width() == 0`. That's an omission.


j9ac9k:
Thanks for the PR @StSav012 this LGTM 👍🏻

PR #2281, opened 2022-04-29
Add a `ms_threshold` flag to profiler: filter blocks "slower then" a ms value, fix recursive profiling

goodboy:
I found this super useful for narrowing down latency issues (from slowest to fastest) much easier then having a whole sale or "tag based" selection of sections enabled.

If noone is interested in this I'll stick with our ongoing approach of pulling into our dependent project and y'all can come ask for it later if wanted 🖖🏼 

### Upate: new changes to aid with *recursive profiling*,
Taken from latest commit msg:
Further this commit makes some larger alterations:
- drops the class variable `._msgs` since it was causing issues on
  nested profiler use. My presumption is that this code was written with
  the expectation that stacks could unwind (per thread) out of order?
  I don't see why else you'd have `.flush()` only print on a `._depth
  < 1` condition..
- drops `.flush()`, [since `.finish()` is the only place this method is
  called (even when scanning the rest of the code base)](https://github.com/pyqtgraph/pyqtgraph/search?q=.flush%28%29&type=code) thus dropping
  a call frame.
- formatting adjustments around this `class` as per `flake8` linter.

Details, in order from bullets above:
- when a top level (aka `._depth == 0`) profiler was deleted by the GC,
  all messages from all underling (aka recursive) profilers were also
  wiped, meaning you'd get a weird case where only the "Exit" msg was
  showing. Now we instead expect each instance to clear it's own msg
  set. There should also be no further reason for the `._depth < 1`
  check since each instance unwinds and emits its own messages,
  presumably as the call stack unwinds via the GC.
- I see no reason to have the `.flush()` when it was a trivial method
  that can be inlined to `.finish()` and I doubt there's much of a use
  case for "incrementally flushing" unless the user is using one "god
  profiler" throughout their entire stack..

### Some notes (old)
- ~found the `.__del__()` finishing to be highly unreliable~ **now solved by making `._msgs` and instance var as per above** (particularly in async code driving sections with profiling)
- ~"nested profiling" seems to be something really handy but i'm still not entirely sure how it works; do you need to pass in the caller's instance to make it work or should it work with new instances always?~ **seems to work great now, even with using the new `ms_threshold` param after the latest patch**.

---
### TODO:
- [ ] should we add a context manager interface to this api?

### DIFF
--- a/pyqtgraph/debug.py
+++ b/pyqtgraph/debug.py
@@ -484,8 +484,6 @@ class GarbageWatcher(object):
     def __getitem__(self, item):
         return self.objs[item]
 
-    
-
 
 class Profiler(object):
     """Simple profiler allowing measurement of multiple time intervals.
@@ -519,39 +517,72 @@ class Profiler(object):
 
     _profilers = os.environ.get("PYQTGRAPHPROFILE", None)
     _profilers = _profilers.split(",") if _profilers is not None else []
-    
+
     _depth = 0
-    _msgs = []
-    disable = False  # set this flag to disable all or individual profilers at runtime
-    
+
+    # XXX: is this needed if the instance
+    # defines it?
+    # _msgs = []
+
+    # set this flag to disable all or individual profilers at runtime
+    disable = False
+
     class DisabledProfiler(object):
         def __init__(self, *args, **kwds):
             pass
+
         def __call__(self, *args):
             pass
+
         def finish(self):
             pass
+
         def mark(self, msg=None):
             pass
+
     _disabledProfiler = DisabledProfiler()
-        
-    def __new__(cls, msg=None, disabled='env', delayed=True, gt: float = 0):
+
+    def __new__(
+        cls,
+        msg=None,
+        disabled='env',
+        delayed=True,
+        ms_threshold: float = 0.0,
+    ):
         """Optionally create a new profiler based on caller's qualname.
+
+        ``ms_threshold`` can be set to value in ms for which, if the
+        total measured time  of the lifetime of this profiler is **less
+        than** this value, then no profiling messages will be printed.
+        Setting ``delayed=False`` disables this feature since messages
+        are emitted immediately.
+
         """
-        if disabled is True or (disabled == 'env' and len(cls._profilers) == 0):
+        if (
+            disabled is True
+            or (
+                disabled == 'env'
+                and len(cls._profilers) == 0
+            )
+        ):
             return cls._disabledProfiler
-                        
+
         # determine the qualified name of the caller function
         caller_frame = sys._getframe(1)
         try:
             caller_object_type = type(caller_frame.f_locals["self"])
-        except KeyError: # we are in a regular function
+
+        except KeyError:  # we are in a regular function
             qualifier = caller_frame.f_globals["__name__"].split(".", 1)[-1]
-        else: # we are in a method
+
+        else:  # we are in a method
             qualifier = caller_object_type.__name__
         func_qualname = qualifier + "." + caller_frame.f_code.co_name
-        if disabled == 'env' and func_qualname not in cls._profilers: # don't do anything
+
+        if disabled == 'env' and func_qualname not in cls._profilers:
+            # don't do anything
             return cls._disabledProfiler
+
         # create an actual profiling object
         cls._depth += 1
         obj = super(Profiler, cls).__new__(cls)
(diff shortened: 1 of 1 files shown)


pyqtgraph/debug.py
```diff
@@ -535,7 +535,7 @@ def mark(self, msg=None):
             pass
     _disabledProfiler = DisabledProfiler()
         
-    def __new__(cls, msg=None, disabled='env', delayed=True):
+    def __new__(cls, msg=None, disabled='env', delayed=True, gt: float = 0):
```
j9ac9k, line 538: probably should make the default value be `0.` or `0.0` if it's going to be type-annotated to a float.

j9ac9k:
Hi @goodboy  Thanks for the PR.  The feature looks good to me, I agree with its usefulness. 

I would suggest renaming the keyword argument from `gt` to `threshold` or something along those lines.

goodboy:
@j9ac9k sounds good!

I have a couple more commits to address issues with nested profilers that i'm going to append here. I will also make the suggested changes 🏄🏼

pyqtgraph/debug.py
goodboy: good idea, are we cool with introducing type annots now?

goodboy:
@j9ac9k changed the name and added more changes that you'll likely want to audit 😂

pyqtgraph/debug.py
```diff
@@ -519,39 +517,72 @@ def function(...):
...
     _disabledProfiler = DisabledProfiler()
-        
-    def __new__(cls, msg=None, disabled='env', delayed=True):
+
+    def __new__(
+        cls,
+        msg=None,
+        disabled='env',
+        delayed=True,
+        ms_threshold: float = 0.0,
```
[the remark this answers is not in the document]
goodboy: I honestly am indifferent but personally don't love the camel case stuff.
It's up to y'all.

goodboy:
> I'm also a little confused at the issue you reported with recursive profiling, can you give me a small code example to understand? I tried the current Profiler implementation with the following and didn't face any issues:

Yeah it probably is confusing since in the client system we're using to call this code is all async driven and it seems the whole unwinding via garbage collection isn't playing great. 

I couldn't even originally see half of the profile stack levels until i made this change. Not entirely sure why.

I can probably whip up something but honestly for such a small change as this we'll probably just end up copying this code out since it's getting resistance for merge yet again 😂 

It looks like you already have something more extensive in #2322 anyway 😎

ntjess:
I wouldn't be opposed to the simpler solution (I mentioned in my PR the current implementation is quite slow if milliseconds matter)

The only thing giving me pause on this PR is the reverse order printout for the MWE I made. The hesitation is certainly not against this PR, I think it's a good idea 👍

> we're using to call this code is all async driven

Do you mean `async def` functions or something else? It would be nice to replicate if indeed it is a bug/limitation in the current implementation, I'd be happy to help

goodboy:
See my comment in #2322 but I think we might just not worry too much about keeping this and go with a more "serious" solution in the longer run for our code base.

https://github.com/plasma-umass/scalene

I really just was using this profiler for graphics stuff, but honestly it would make more sense to be profiling things properly (even `Qt` C++ code) instead of this counter based stuff.

goodboy:
> Do you mean async def functions or something else? It would be nice to replicate if indeed it is a bug/limitation in the current implementation, I'd be happy to help

Yes, we have a single-thread graphics loop entirely driven by async data flows from multiple processes.

goodboy:
> The only thing giving me pause on this PR is the reverse order printout for the MWE I made. The hesitation is certainly not against this PR, I think it's a good idea +1

Sorry missed this. Yeah I mean if it's something obvious can you just revert what you think is causing it, PR it to this branch and then I'll give it a shot.

Sorry i'm not in graphics land at the moment but will definitely come back this!

Also thanks for all the feedback 😸

goodboy:
@ntjess just went back to this again and I think you're right. It's the change of `._msgs` to an instance var that broke it.

Not sure why I couldn't get it working before (maybe other issues in the threshold logic before it was working properly?) but seems to be doing about what I'd expect now.

I'll push up this last little change and you can tell me if it matches your expectations 😎

goodboy:
> I mentioned in my PR that the `logging` implementation is quite slow if milliseconds matter

Oh, and yes, for our purposes we're trying to get under 10ms per full graphics render cycle on interactions with >= 1M datums..so yeah it kinda does matter. I'm already dealing with a bit much skew due to the current overhead latency this original impl contains 😂

Issue #2283, opened 2022-04-30
Exception for polyline ROI region  with value more then 1000


sasha-sem:
Hi,
I found a thing that I don't understand.
For some reason if I usestriangle or polyline ROI, I can't get region of ROI with width of its containing rectangle more than 1000. In this case python throws Exception if i change this value to any other it works with no problems.
https://github.com/pyqtgraph/pyqtgraph/blob/6f6e86e0deff0e41745396c18bdba8883e0eee6a/pyqtgraph/graphicsItems/ROI.py#L1200

That's why, I have a question. Is there any reason for this limitation and is there way to change/remove it?
Thank you!

j9ac9k:
Hi @sasha-sem 

That change was introduced quite some time ago as part of this commit: https://github.com/pyqtgraph/pyqtgraph/commit/bb507cf6d089a6aa7dcbbd8656a75969b754815f

As there is no comment there, I would be ok with removing that if-check provided the test suite still passes.

j9ac9k:
@campagnola you happen to have any memory of that if-check?

campagnola:
Most likely I added this to debug some issue and should never have committed it.

j9ac9k:
> Most likely I added this to debug some issue and should never have committed it.

I've _never_ done that 😆

PR #2285, opened 2022-05-01
#2283 delete limitation by rectangle width

sasha-sem:
Deleted strange exception for region of ROI with width of its containing rectangle more than 1000 that was discussed in Issue #2283


j9ac9k:
Hi @sasha-sem 

Thanks for the PR, this LGTM based on the discussion in #2283.  Would you mind deleting the line above, that method no longer uses the `br` variable so no need to get it.

PR #2286, opened 2022-05-03
avoid double __init__ of DockDrop

pijyoi:
The `DockDrop` class has its `__init__` method being called twice on PyQt{5, 6}.

According to the following references, PyQt{5,6} (but not PyQt4) added cooperative inheritance.
Besides minor changes, `DockDrop` was last modified in 2012. i.e. the code was written based on PyQt4 behavior.
https://docs.huihoo.com/pyqt/PyQt5/pyqt4_differences.html#cooperative-multi-inheritance
https://doc.bccnsoft.com/docs/PyQt5/multiinheritance.html
https://www.riverbankcomputing.com/pipermail/pyqt/2017-January/038650.html

This PR avoids the double `__init__` by differentiating between PyQt{5,6} and PySide{2,6}.
PySide{2,6} do not support cooperative inheritance. i.e. They have PyQt4 behavior.

An attempt was made to use `DockDrop` by composition instead of inheritance, thus avoiding all this trouble. But it turned out that there is quite a bit of coupling present. e.g. there is an assumption that a `DockDrop` instance *is* a `QWidget`.

### DIFF
--- a/pyqtgraph/dockarea/Dock.py
+++ b/pyqtgraph/dockarea/Dock.py
@@ -1,6 +1,6 @@
 import warnings
 
-from ..Qt import QtCore, QtGui, QtWidgets
+from ..Qt import QT_LIB, QtCore, QtGui, QtWidgets
 from ..widgets.VerticalLabel import VerticalLabel
 from .DockDrop import DockDrop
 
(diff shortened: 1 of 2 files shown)


j9ac9k:
Thanks @pijyoi for this PR!  This LGTM!

PR #2304, opened 2022-05-18
GridItem: Fix pen for usage of dash-pattern

bbc131:
Hi, I wanted to use a custom dash-pattern for the grid. Currently this doesn't work, the drawn dash pattern changes with the zoom of the plot.
See the attached screenshot showing different vertical and horizontal grid lines, due to the different ranges set at the end of the example.
```
...
app = mkQApp("Grid bug demo")
plot = PlotWidget()
plot.show()

grid = GridItem()
grid.setPen(dash=[4,4,8,4])
plot.addItem(grid)

plot.setXRange(min=0.0, max=10.0)
plot.setYRange(min=0.0, max=1.0)
```
grafik

### DIFF
--- a/pyqtgraph/graphicsItems/GridItem.py
+++ b/pyqtgraph/graphicsItems/GridItem.py
@@ -163,11 +163,7 @@ class GridItem(UIGraphicsItem):
 
                 bx = (ax+1) % 2
                 for x in range(0, int(nl[ax])):
-                    linePen.setCosmetic(False)
-                    if ax == 0:
-                        linePen.setWidthF(self.pixelWidth())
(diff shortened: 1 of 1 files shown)


bbc131:
Note, this should also fixe #1709.
Now, something like `grid.setPen(width=4)` should work.

ixjlyons (approved):
This appears to work well to me. I suspect that was *some* reason it was `setCosmetic(False)`, but I'm not finding one. Zoom, window resize, export, all seem fine with this change on the examples that use a `GridItem`. I'm not sure if there's anything else that should be checked.

PR #2308, opened 2022-05-19
Allow passing in a pre-allocated path for reuse

goodboy:
If we don't reuse the `QPainterPath` that is be "reserved" (i.e. by
calling `.reserve()` on it to avoid re-allocating memory and using
`.clear()` instead of generating a new path) the mem allocs will be
somewhat pointless. Instead make it possible to pass in a previously
generated path, (maybe?) clear it, and rewrite to the reserved space.

Relevant docs:
- https://doc.qt.io/qt-5/qpainterpath.html#reserve
- https://doc.qt.io/qt-5/qpainterpath.html#clear

Source code:
- https://codereview.qt-project.org/gitweb?p=qt/qtbase.git;a=blob;f=src/gui/painting/qpainterpath.cpp;h=bfe86b251d90cd996affc2208ee4b766af74c5d9;hb=423b6509382c3bcc36eca78ced0254bfb463d017#l626

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -2043,14 +2043,20 @@ def _arrayToQPath_finite(x, y, isfinite=None):
     return path
 
 
-def arrayToQPath(x, y, connect='all', finiteCheck=True):
+def arrayToQPath(
+    x,
+    y,
+    connect='all',
+    finiteCheck=True,
+    path=None,
+):
     """
     Convert an array of x,y coordinates to QPainterPath as efficiently as
     possible. The *connect* argument may be 'all', indicating that each point
     should be connected to the next; 'pairs', indicating that each pair of
     points should be connected, or an array of int32 values (0 or 1) indicating
     connections.
-    
+
     Parameters
     ----------
     x : (N,) ndarray
@@ -2058,7 +2064,7 @@ def arrayToQPath(x, y, connect='all', finiteCheck=True):
     y : (N,) ndarray
         y-values to be plotted, must be same length as `x`
     connect : {'all', 'pairs', 'finite', (N,) ndarray}, optional
-        Argument detailing how to connect the points in the path. `all` will 
+        Argument detailing how to connect the points in the path. `all` will
         have sequential points being connected.  `pairs` generates lines
         between every other point.  `finite` only connects points that are
         finite.  If an ndarray is passed, containing int32 values of 0 or 1,
@@ -2067,12 +2073,12 @@ def arrayToQPath(x, y, connect='all', finiteCheck=True):
         When false, the check for finite values will be skipped, which can
         improve performance. If nonfinite values are present in `x` or `y`,
         an empty QPainterPath will be generated.
-    
+
     Returns
     -------
     QPainterPath
         QPainterPath object to be drawn
-    
+
     Raises
     ------
     ValueError
@@ -2091,7 +2097,7 @@ def arrayToQPath(x, y, connect='all', finiteCheck=True):
     1(i4)   x(f8)   y(f8)    <-- 1 means this vertex connects to the previous vertex
     ...
     cStart(i4)   fillRule(i4)
-    
+
     see: https://github.com/qt/qtbase/blob/dev/src/gui/painting/qpainterpath.cpp
 
     All values are big endian--pack using struct.pack('>d') or struct.pack('>i')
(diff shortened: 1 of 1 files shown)


pijyoi:
```python
import numpy as np
import pyqtgraph as pg 
from time import perf_counter

nelems = 1_000_000

def count_reallocs(nelems, reserve):
    capas = set()
    path = QtGui.QPainterPath()
    if reserve:
        path.reserve(nelems)
    for x in range(nelems):
        path.lineTo(x, x)
        capas.add(path.capacity())
    return sorted(list(capas))

# demonstrate that reserve reduces the number of re-allocs
# during construction of the qpath
for reserve in [False, True]:
    capas = count_reallocs(nelems, reserve=reserve)
    print(f'{reserve=} {capas=}')

def add_stream(ba, nelems, reserve):
    path = QtGui.QPainterPath()
    if reserve:
        path.reserve(nelems)
    ds = QtCore.QDataStream(ba)
    ds >> path
    return path

ramp = np.arange(nelems, dtype=np.float64)
path = pg.functions.arrayToQPath(ramp, ramp)
ba = QtCore.QByteArray()
ds = QtCore.QDataStream(ba, QtCore.QIODevice.OpenModeFlag.WriteOnly)
ds << path

# demonstrate that reserve is not point-less
for reserve in [False, True]:
    t0 = perf_counter()
    path = add_stream(ba, nelems, reserve=reserve)
    t1 = perf_counter()
    assert path.elementCount() == nelems
    print(f'{reserve=} {t1-t0:.6f}')
```

Output when run on PySide6:
```
reserve=False capas=[16, 20, 42, 84, 170, 340, 682, 1364, 2730, 5460, 10922, 21844, 43690, 87380, 174762, 349524, 699050, 1398100]
reserve=True capas=[1000000]
reserve=False 0.156873
reserve=True 0.146354
```

You could also run `asv` on https://github.com/pyqtgraph/pyqtgraph/blob/master/benchmarks/arrayToQPath.py

j9ac9k:
Hi @goodboy 

Thanks for the PR!

I've been looking at the code and the diff for a while, and the only potential benefit I see is not so much in how we use `.reserve(int)` and `.clear()` but in potentially reusing QPainterPath instances; but I'm not convinced that there actually would be a measurable improvement and if there was, what would be the memory cost.

@pijyoi was kind enough to provide a benchmark script that demonstrates our current usage of `.reserve()` does in fact lead to a slight performance improvement, so I don't think we're using it incorrectly in its current form.  If you can demonstrate via a similar script that this PR would lead to a performance improvement I'd like to see it.

goodboy:
Ahh I see now how in this simplest case (where `n` datums maps to `n` path elements, at least i think) you get an improvement - so yeah part of the change i made isn't right. I didn't actually realize that you get a speedup by pre-allocating on before the first path write since I assumed the mem would be allocated based on the path generation size which wouldn't always be possible to calc ahead of time? - but i guess in this simplest case `n` datums from `numpy` **is equal** to the # of elements in the path? I need to grok this a little better.

I think I was hitting an edge case I guess? maybe i'm confused though:

> reserve=False capas=[16, 20, 42, 84, 170, 340, 682, 1364, 2730, 5460, 10922, 21844, 43690, 87380, 174762, 349524, 699050, 1398100]

I understand this output shows incremental mem size increases but i'm pretty sure the main issue is that `n` doesn't necessarily map to the element size of the path *generated by n `numpy` datums* for all connection types when you look at how `backingstore` is filled out in `.arrayToQPath()`.

In other words, an output path from `n` elements from the `numpy` array can be **be greater then** the number of input datums, this is the value `n` that `.reserve()` is being called with inside the func. So unless you always make `n >= .capacity()` (which you can't guarantee until you've rendered the path and/or pre-calculated it or have some kind of transform calc), I think, you can always hit this edge case no?

Let me play with @pijyoi's script a bit and try to replicate what i previously thought Irew found.

goodboy:
> but i'm pretty sure the main issue is that n doesn't necessarily map to the element size of the path generated by n numpy datums for all connection types when you look at how backingstore is filled out in .arrayToQPath().

Ok so I double checked and yes, it's the case where you **do not have** a simple line / curve graphic path.

An example would be the OHLC sample scheme bars I'm using:

screenshot-2022-05-27_10-00-17

In this case the `n` from [`x.shape[0]`](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L2101) is not equal to the # of line elements per datum.

The reference array generation can bee [seen here](https://github.com/pikers/piker/blob/incremental_update_paths/piker/ui/_pathops.py#L83) and results in a `connect == 'array'` case which is not handling the case to pre-allocate for a larger number of elements.

There's 2 ways I can see to approach this:
- on first path allocation render and then `.reserve(path.capacity()` as I've done here and allow passing in the same path object for reuse - so, use the same mem allocs for a new data set that is of the same `n` datums -> `N` elements.
- allow passing in some kind of function to generate the `.reserve()` argument based on the user's knowledge of their own graphic element count?

I will come back and hopefully update the script to show my scenario in simpler terms.

j9ac9k:
Hi @goodboy 

I'm going to close out this PR.  I'm certainly open to trying to get some faster performance in our QPainterPath generation, but the code presently in the library has been used very extensively and covers a _wide_ variety of edge cases.  To incorporate a change to `arrayToQPath` there would need to be a justifiable case; which would mean both code to demonstrate the performance improvement, and potentially additional test code to ensure nothing breaks.  This is a very sensitive bit of code to the library.

I would note that I'm skeptical that reusing an already instantiated QPainterPath will save that much more time vs. creating a new one and allowing the older path to be garbage collected (but I could certainly be mistaken!)

PR #2309, opened 2022-05-19
Update exception handling to catch exceptions in threads (py3 change)

campagnola:
In python 3, unhandled exceptions inside threads no longer invoke `sys.excepthook`; instead they invoke `threading.excepthook`. Additionally, the signature of `threading.excepthook` is incompatible with `sys.excepthook`, which necessitates adding new callback registration functions and deprecating the old ones.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,7 +4,7 @@
 version: 2
 
 python:
-  version: 3
+  version: 3.8
   install:
     - requirements: doc/requirements.txt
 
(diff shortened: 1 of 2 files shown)


j9ac9k:
This is the first time in a while I've seen the readthedocs CI fail.  Also Py310-PySide6 binding failures are unrelated to this PR.

campagnola:
Oo, threading.excepthook was added in python 3.8. I guess readthedocs is using something older..

j9ac9k:
Thanks @campagnola for the PR!

Issue #2207, opened 2022-02-23
visual glitches with InfinteLine


danielhrisca:
### Code to reproduce

```python
import pyqtgraph as pg
import numpy as np
```

### Real behavior

I use a movable InfinteLine to simulate a cursor. With PyQt5 there is no virtual glitch, but with PySide6 I see the following glitches:

![glitches](https://user-images.githubusercontent.com/20952040/155315637-bc796f2a-b675-49c4-a17b-65bdb5627841.gif)

```
An error occurred?
Post the full traceback inside these 'code fences'!
```

### Tested environment(s)

 * PyQtGraph version:  0.12.3 (latest master code)
 * Qt Python binding:  PySide6 6.2.3 Qt 6.2.3
 * Python version: 3.9 x64
 * NumPy version:  1.21.5
 * Operating system: Windows 10 x64
 * Installation method:  from source

danielhrisca:
It looks like the problem is related to the DPI setting. With 100% scaling there is no error. With 125% scaling the glitches show up.

It doesn't matter if the DPI is at 125% before the application is started.

![glitches_dpi](https://user-images.githubusercontent.com/20952040/155330580-55f5a43a-efd3-46d9-a0d6-b12fb6cf84f2.gif)

danielhrisca:
The solution is to set the environment variable 

```
os.environ["QT_ENABLE_HIGHDPI_SCALING"] = "0"
```

this reverts to the behavior found in Qt5 according to the [Qt6 documentation](https://doc.qt.io/qt-6/highdpi.html#environment-variable-reference)

j9ac9k:
oof this is going to hurt; 

Here is where we attempt to handle hidpi to the best of our ability

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/Qt/__init__.py#L426-L437

Enabling HIDPI support is definitely not something we're going to turn off; enabling that fixed a large variety of issues.

@danielhrisca in your application, do you use `pg.mkQApp` or do you instantiate your own `QApplication`? If you instantiate your own, before `exec`-ing it, can you try running

```python
QtWidgets.QApplication.setHighDpiScaleFactorRoundingPolicy(QtCore.Qt.HighDpiScaleFactorRoundingPolicy.PassThrough)
```

danielhrisca:
[...]

PR #2310, opened 2022-05-20
ScatterPlotItem point masking fix

ardiloot:
If `ScatterPlotItem` is transformed by `item.setTransform(...)` then it starts to act weird (e.g. symbols partly disappearing when zooming in).
Discovered that scatter points are incorrectly filtered by untransformed view box (`self.getViewBox().viewRect()`), however, view box in the local coordinate system of the item (`self.viewRect()`) should be used instead (as data is given in the  local coordinate system of the item).


j9ac9k:
Hi @ardiloot this must have been a _nightmare_ to debug!  This PR LGTM, I'll merge shortly.

I wonder if this same fix can be used to resolve #2207 or #2312

PR #2311, opened 2022-05-23
Make parameter tree read-only values selectable and copiable

ardiloot:
If `ParameterTree` is used with read-only parameters then it makes sense that it would be possible to select and copy the value of the parameter. Right now the value is not selectable or copiable in any way.

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes/basetypes.py
+++ b/pyqtgraph/parametertree/parameterTypes/basetypes.py
@@ -230,7 +230,7 @@ class WidgetParameterItem(ParameterItem):
         if 'readonly' in opts:
             self.updateDefaultBtn()
 
-            if opts["readonly"]:
+            if opts['readonly']:
                 self.displayLabel.setTextInteractionFlags(QtCore.Qt.TextSelectableByMouse)
             else:
                 self.displayLabel.setTextInteractionFlags(QtCore.Qt.LinksAccessibleByMouse)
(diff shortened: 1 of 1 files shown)


pyqtgraph/parametertree/parameterTypes/basetypes.py
```diff
@@ -229,6 +229,12 @@ def optsChanged(self, param, opts):
 
         if 'readonly' in opts:
             self.updateDefaultBtn()
+
+            if opts['readonly']:
+                self.displayLabel.setTextInteractionFlags(QtCore.Qt.TextSelectableByMouse)
+            else:
+                self.displayLabel.setTextInteractionFlags(QtCore.Qt.LinksAccessibleByMouse)
```
j9ac9k, line 236: For PyQt6 support you need to specify the full enum namespace, so it should be 

```python
            if opts['readonly']:
                self.displayLabel.setTextInteractionFlags(QtCore.Qt.TextInteractionFlag.TextSelectableByMouse)
            else:
                self.displayLabel.setTextInteractionFlags(QtCore.Qt.TextInteractionFlag.LinksAccessibleByMouse)
```

Luckily the full name enum namespace works for all the other bindings we support too 👍🏻

j9ac9k:
Hi @ardiloot 

Thanks for the PR.  I left an in-line comment explaining why this is failing on PyQt6, and showed you what adjustment you need to make to make it work for all the bindings pyqtgraph supports.

Thanks!

ntjess:
I played with this for a bit, seems ok to me! My only thought is that some widgets (i.e. calendar or other thing with a display label) won't get these benefits, but subitems are a bit hard to generalize and that can always come later.

ntjess (approved):

j9ac9k:
Merging, thanks for the PR @ardiloot !

PR #2313, opened 2022-05-26
test_PlotCurveItem: unset skipFiniteCheck for next test

pijyoi:
`test_PlotCurveItem.py` has a bug as follows:

At the time (8997cfa07c7a0ff4209f60702b028a81f46e14ca) of the addition of the 2nd last test with `skipFiniteCheck=True`, `skipFiniteCheck` was not a sticky option. Thus, the last test automatically reverts to `skipFiniteCheck=False`.

Since the other options of `PlotCurveItem` were sticky in nature, for consistency, 21d77dbea5225fcce448df2f9149174e877db472 changed `skipFiniteCheck`  to also be a sticky option. This made the last test to be run with `skipFiniteCheck=True` even though it was intended to be executed with `skipFiniteCheck=False`.

Note: With the bug, a blank image actually gets generated for the last test but due to the low number of "on" pixels in the reference image, it passes the tolerance test.


j9ac9k:
Nice catch @pijyoi   Thanks for the PR!

PR #2315, opened 2022-05-26
Fix PlotCurveItem errors when pen=None

campagnola:
I received errors when using a `PlotCurveItem` with `pen=None`. This used to work; I'm not sure what changed. 

MWE:
```
import pyqtgraph as pg
plt = pg.plot()
c = pg.PlotCurveItem([1, 2, 3, 4], [5, 2, 4, 3], pen=None)
plt.addItem(c)
```


j9ac9k:
While I question the utility of plotting a curve without a pen, I admit pyqtgraph shouldn't blow up if that happens haha.  LGTM, thanks @campagnola

PR #2317, opened 2022-05-29
ScatterPlotItem: use Format_ARGB32_Premultiplied

pijyoi:
From https://doc.qt.io/qt-5/qimage.html: "Rendering is best optimized to the Format_RGB32 and Format_ARGB32_Premultiplied formats"

### DIFF
--- a/pyqtgraph/graphicsItems/ScatterPlotItem.py
+++ b/pyqtgraph/graphicsItems/ScatterPlotItem.py
@@ -97,7 +97,8 @@ def renderSymbol(symbol, size, pen, brush, device=None):
     ## Render a spot with the given parameters to a pixmap
     penPxWidth = max(math.ceil(pen.widthF()), 1)
     if device is None:
-        device = QtGui.QImage(int(size+penPxWidth), int(size+penPxWidth), QtGui.QImage.Format.Format_ARGB32)
+        device = QtGui.QImage(int(size+penPxWidth), int(size+penPxWidth),
+            QtGui.QImage.Format.Format_ARGB32_Premultiplied)
         device.fill(QtCore.Qt.GlobalColor.transparent)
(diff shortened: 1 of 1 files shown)


j9ac9k:
@danielhrisca @ibrewster you two seem to have some extreme scatter plot performance needs, curious if this change will make a noticeable impact for you.

pijyoi:
Oh, there should be no measurable impact because the QImage(s) being rendered to are very small.
This is more to do with code correctness (as far as Qt docs are concerned).

j9ac9k:
LGTM @pijyoi nice catch in the docs!  Merging.

danielhrisca:
> @danielhrisca @ibrewster you two seem to have some extreme scatter plot performance needs, curious if this change will make a noticeable impact for you.

I've had to step away from the default plotting functionality provided by pyqtgraph.

Essentially the following goes on:
* there is only one viewbox instead of individual viewbox/curve
* I use the curve objects as storage, but I don't add them to the plotitem 
* in the reimplemented paintEvent the curves are scaled and drawn on the viewport. In case of scatter plots I use QPainter.drawPoints and set the pen width and cap style to get de desired pixel size and shape
* there is also the functionality to have different curve colors depending on certain conditions (eq. y > 300 => red color). This is done by drawing the segments on top of the default curve polygon
* the painted curves are stored to a qpixmap that only gets updated if the user changes the x or y ranges or if curves are disabled or enabled. With this approach the pixmap gets painted and then I can have an infinte line acting as a cursor that can be moved around with minimal performance impact (in the older implementation moving the cursor would trigger the repaint of all the curves)

If anybody is interested the code is [here](https://github.com/danielhrisca/asammdf/blob/d210a2112ea530cdaaaa7d9f6e66ee09b1046443/asammdf/gui/widgets/plot.py#L4442).

PR #2319, opened 2022-05-31
Enhance testing for creating parameters from saved states

ntjess:
Before this PR, parameters like e.g. checklist would save states that couldn't reproduce their parameter layout. Fix this for the specific fail types of checklist and calendar while ensuring all other parameter types can successfully recreate themselves.

During testing, a few other issues with checklists came up  and were resolved, so the branch name is a bit of a misnomer now
- [x] `exclusive` checklists now propagate changing values correctly
- [x] `Clear All`, `Select All`, and default buttons now force changes even when there is a pending proxy signal

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes/calendar.py
+++ b/pyqtgraph/parametertree/parameterTypes/calendar.py
@@ -51,5 +51,6 @@ class CalendarParameter(Parameter):
     def saveState(self, filter=None):
         state = super().saveState(filter)
         fmt = self._interpretFormat()
-        state['value'] = state['value'].toString(fmt)
+        if state['value'] is not None:
+            state['value'] = state['value'].toString(fmt)
         return state
--- a/tests/parametertree/test_parametertypes.py
+++ b/tests/parametertree/test_parametertypes.py
@@ -173,3 +173,11 @@ def test_pen_settings():
     # Opts from changing child
     p["width"] = 10
     assert p.pen.width() == 10
+
+
+def test_recreate_from_savestate():
+    from pyqtgraph.examples import _buildParamTypes
+    created = _buildParamTypes.makeAllParamTypes()
+    state = created.saveState()
+    created2 = pt.Parameter.create(**state)
+    assert pg.eq(state, created2.saveState())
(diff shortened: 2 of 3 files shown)


tests/parametertree/test_parametertypes.py
```diff
@@ -173,3 +173,11 @@ def test_pen_settings():
...
     p["width"] = 10
     assert p.pen.width() == 10
+
+
+def test_recreate_from_savestate():
+    from pyqtgraph.examples import _buildParamTypes
+    created = _buildParamTypes.makeAllParamTypes()
+    state = created.saveState()
+    created2 = pt.Parameter.create(**state)
+    assert pg.eq(state, created2.saveState())
```
j9ac9k, line 183: add new-line here

pyqtgraph/parametertree/parameterTypes/calendar.py
```diff
@@ -51,5 +51,6 @@ def _interpretValue(self, v):
     def saveState(self, filter=None):
         state = super().saveState(filter)
         fmt = self._interpretFormat()
-        state['value'] = state['value'].toString(fmt)
+        if state['value'] is not None:
```
j9ac9k, line 54: under what conditions would this be `None`?  Also is there a case where the key `value` would be missing and we should use `state.get('value')` instead?

ntjess: `value` is guaranteed to exist in the state and is `None` if nothing is initially specified.

I see two easy options for a better initial value if the user doesn't specify anything:
- `QDate()` would give a null/invalid (but date-like) value
- `QDate.currentDate()` is what the calendar widget's selected day defaults to. I'd lean toward this one since it's intuitive based on the calendar being displayed. Of course, a minor downside is potentially non-reproducible results overtime due to a changing default value.

j9ac9k:
Going to close/re-open to trigger a new CI test.

pyqtgraph/parametertree/parameterTypes/calendar.py
j9ac9k: I think `QDate.currentDate()` is a pretty good default, but I think changing this might be a bit out of scope for this PR.

PR #2322, opened 2022-06-03
Allow profiler to utilize logging capabilities

ntjess:
Offers many more features than the default pyqtgraph profiler! Stream logging, filtering messages, log levels for scope change vs. profiling, ...

The only downside: It's **slow** (maybe just an implementation issue 🤞). I'm hoping someone smarter than me can figure out how to streamline it more effectively.

My current benchmark:

```python

def fib_bad_recursive(n):
    """
    A bad recursive implementation of fibonacci.
    """
    if n < 2:
        return n
    profiler = Profiler("fib_bad_recursive", disabled=False)
    return fib_bad_recursive(n - 1) + fib_bad_recursive(n - 2)

if __name__ == "__main__":
    n = 6
    Profiler = OldDefinitionProfiler
    fib_bad_recursive(n) # Takes about 0.13ms on my machine

    print("\n=====================\n")

    Profiler = NewDefinitionProfiler
    fib_bad_recursive(n) # Takes about 16ms!!!
```

If this works, I think it will add more functionality and be easier to customize than #2281
@goodboy you're welcome to browse this implementation; the goal was to allow more flexibility than just time filtering if changes are happening. This also allows users to customize messaging and everything else that logging exposes.

### DIFF
--- a/pyqtgraph/debug.py
+++ b/pyqtgraph/debug.py
@@ -11,6 +11,7 @@ import contextlib
 import cProfile
 import gc
 import inspect
+import logging
 import os
 import re
 import sys
@@ -37,6 +38,7 @@ if sys.version.startswith("3.8") and QT_LIB == "PySide2":
         )
 from .util.mutex import Mutex
 
+logger = logging.getLogger(__name__)
 
 # credit to Wolph: https://stackoverflow.com/a/17603000
 @contextlib.contextmanager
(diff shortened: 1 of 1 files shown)


goodboy:
Fwiw we're probably going to start playing with a larger project:
https://github.com/plasma-umass/scalene

So i'll report back with some demo code if y'all are interested.

PR #2323, opened 2022-06-05
Add sphinx gallery to docs

ixjlyons:
https://pyqtgraph--2323.org.readthedocs.build/en/2323/examples/index.html

I'm thinking once this is working relatively well, we can start tackling some other improvements like categorizing examples in a separate PR.

- [x] Fix merge conflicts
- [ ] Track down the examples that are giving "RuntimeError: Internal C++ object (ViewBox) already deleted". Ideally we can fix them, but for now either PyQt5 or PyQt6 will treat them like warnings and still output a rendering that looks reasonable. PyQt6 appears to not be installable on readthedocs for now...
- [ ] Take another look at how VisPy's scraper works - is it feasible for us to capture/render animations? e.g. https://vispy.org/gallery/gloo/animate_images.html#sphx-glr-gallery-gloo-animate-images-py
- [ ] Work around the `__file__` issue in some of the examples (https://sphinx-gallery.github.io/stable/faq.html#why-is-file-not-defined-what-can-i-use)
- [ ] ~Can we get intersphinx working with Qt/bindings somehow so things like `QMainWindow` used in an example links to the Qt API page for it? That would be neeeeat, but I suspect unlikely with our abstraction package~ (edit: haven't gotten it to work with https://doc.qt.io/qtforpython/ thought it might be feasible?)
- [ ] Get GL examples working


pyqtgraph/__init__.py
```diff
@@ -60,6 +60,8 @@
                                   # 'off' or False: lines are never plotted in segments
 }
 
+DEFAULT_CONFIG_OPTIONS = CONFIG_OPTIONS.copy()
```
ixjlyons, line 63: I think this could/should go in the sphinx `conf.py` actually

pyqtgraph/examples/_RunExampleApp.py
```diff
@@ -4,7 +4,7 @@
 not run twice.
 """
 
-from ExampleApp import ExampleLoader
+from _ExampleApp import ExampleLoader
```
ixjlyons, line 7: I guess I should check - this doesn't break one of the various ways of running the example application does it?

pyqtgraph/examples/CLIexample.py
```diff
@@ -1,20 +1,25 @@
 """
-Display a plot and an image with minimal setup. 
+CLI Example
```
ixjlyons, line 2: "Tutorialized" this example https://pyqtgraph--2323.org.readthedocs.build/en/2323/examples/CLIexample.html#sphx-glr-examples-cliexample-py

j9ac9k: Was wondering if this example even worked anymore now that interactive mode doesn't necessarily work with some bindings.

ixjlyons: That's a good point, looks like this usage currently only works with PyQt{5,6}. So maybe not the best example to iterate on, but I *believe* this would still work with PySide at least for the docs build. I use PySide2 for qtgallery and the [iterative example](https://qtgallery.readthedocs.io/en/latest/auto_examples/iterative.html) works.

ixjlyons:
Playing with the CLI example, I found that with PyQt5, running that example interactively in a REPL, right clicking on the image, then exiting caused a `RuntimeError: Internal C++ object...` error from `ViewBoxMenu`. I'd seen some errors like this in the gallery build, so after some digging, I found that wrapping these two lines in a `try / except RuntimeError` got rid of it and building the docs with PyQt5 and PyQt6 gives no errors:

https://github.com/pyqtgraph/pyqtgraph/blob/4b802de5fab2468dfb193ae79d6fcb4cf957f68e/pyqtgraph/graphicsItems/ViewBox/ViewBoxMenu.py#L239-L240

I do still see a bunch of these types of errors with PySide:

```
  File "/home/krlyons/src/pyqtgraph/pyqtgraph/examples/DataSlicing.py", line 25, in <module>
    imv1 = pg.ImageView()
  File "/home/krlyons/src/pyqtgraph/doc/source/../../pyqtgraph/imageview/ImageView.py", line 129, in __init__
    self.ui.setupUi(self)
  File "/home/krlyons/src/pyqtgraph/doc/source/../../pyqtgraph/imageview/ImageViewTemplate_pyside6.py", line 42, in setupUi
    self.histogram = HistogramLUTWidget(self.layoutWidget)
  File "/home/krlyons/src/pyqtgraph/doc/source/../../pyqtgraph/widgets/HistogramLUTWidget.py", line 22, in __init__
    self.item = HistogramLUTItem(*args, **kargs)
  File "/home/krlyons/src/pyqtgraph/doc/source/../../pyqtgraph/graphicsItems/HistogramLUTItem.py", line 107, in __init__
    self.vb = ViewBox(parent=self)
  File "/home/krlyons/src/pyqtgraph/doc/source/../../pyqtgraph/graphicsItems/ViewBox/ViewBox.py", line 236, in __init__
    self.updateViewLists()
  File "/home/krlyons/src/pyqtgraph/doc/source/../../pyqtgraph/graphicsItems/ViewBox/ViewBox.py", line 1699, in updateViewLists
    nv = sorted(ViewBox.NamedViews.values(), key=view_key)
  File "/home/krlyons/src/pyqtgraph/doc/source/../../pyqtgraph/graphicsItems/ViewBox/ViewBox.py", line 1696, in view_key
    return (view.window() is self.window(), view.name)
RuntimeError: Internal C++ object (ViewBox) already deleted.
```

But I'm somewhat hopeful a similar fix could work and might end up getting rid of some of these errors that occasionally pop up in regular usage (or does that still actually happen?)

ixjlyons:
Re: comment above

Yep, wrapping this in a try/except gets rid of the rest of the errors for PyQt and PySide

https://github.com/pyqtgraph/pyqtgraph/blob/bb1b7714c3fa509b9f2db1d5b1346d3dd2609ce4/pyqtgraph/graphicsItems/ViewBox/ViewBox.py#L1689

The difference here is that from the traceback I pasted above, this is happening during setup and not teardown, so this is probably not the *right* approach. Maybe there's something in qtgallery to fix in this case.

j9ac9k:
Hi @ixjlyons I just made some documentation changes which caused some merge conflicts, sorry about that! 

Once addressed, would you mind re-running with error on warning enabled again? I'd like to try and replicate and solve those if possible!

ixjlyons:
Updated with master and resolved conflicts. It's running now with `fail_on_warning: true` though those warnings are generated regardless.

The last thing I had done was play around with the environment to see if I could get all the OpenGL examples working reliably. I'm not really sure what's wrong but I do get some GL errors locally too.

For the record, here's are the patches I mentioned above that give me an error-free build locally (aside from the OpenGL stuff):

```diff
diff --git a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
index 3ca843ce..8454b26c 100644
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -1693,7 +1693,10 @@ class ViewBox(GraphicsWidget):
             return (view.window() is self.window(), view.name)
 
         ## make a sorted list of all named views
-        nv = sorted(ViewBox.NamedViews.values(), key=view_key)
+        try:
+            nv = sorted(ViewBox.NamedViews.values(), key=view_key)
+        except RuntimeError:
+            return
 
         if self in nv:
             nv.remove(self)
diff --git a/pyqtgraph/graphicsItems/ViewBox/ViewBoxMenu.py b/pyqtgraph/graphicsItems/ViewBox/ViewBoxMenu.py
index cf468ed3..90bd6314 100644
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBoxMenu.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBoxMenu.py
@@ -236,8 +236,11 @@ class ViewBoxMenu(QtWidgets.QMenu):
             self.viewMap[name] = v
             
         for i in [0,1]:
-            c = self.ctrl[i].linkCombo
-            current = c.currentText()
+            try:
+                c = self.ctrl[i].linkCombo
+                current = c.currentText()
+            except RuntimeError:
+                continue
             c.blockSignals(True)
             changed = True
             try:
```

doc/requirements.txt
```diff
@@ -1,5 +1,10 @@
-sphinx==5.0.2
+<<<<<<< HEAD
```
j9ac9k, line 1: some merge conflict noise left-over

j9ac9k:
Heh, macOS seemed to have an issue 😆 

image

j9ac9k:
this Runtime error that is being encountered is _screaming_ race condition to me; I think it would be beneficial if we tracked down the source (whether it's in qtgallery or in pyqtgraph).  I'll poke around some.

ixjlyons:
> Heh, macOS seemed to have an issue laughing


Hmm, not sure what to make of this. I suspect it's because `QScreen.grabWindow` supports grabbing contents that aren't part of the application - there's a note [here](https://doc.qt.io/qt-6/qscreen.html#grabWindow) about iOS at least.

The original motivation for using `QScreen.grabWindow` over `QWidget.grab` was to support QML applications in qtgallery. It may be worth reconsidering that switch or coming up with something better.

> Also some more issues on macOS:
> 
> ```
> Qt api: PySide2
> WARNING: Could not find Xvfb
> ```
>
> also macOS does not have Xvfb, we probably should use it if we see it, but not try and force the issue if it is not present.

That's pretty much what is happening - it's a non-fatal warning.

However, I just noticed that disabling the virtual display, the renderings are mostly *not* the plot window contents but are seemingly random grabs of my display contents (behind the window?). Looking more at the `grabWindow` docs, there's a warning about X11, but it doesn't seem to be describing this situation exactly. Speaking of race condition...

> this Runtime error that is being encountered is screaming race condition to me

Yep, haven't been able to track it down to this point.

doc/source/conf.py
```diff
@@ -101,6 +110,153 @@
...
+    if script_vars["src_file"].endswith("hdf5.py"):
+        return ["test.hdf5", "1000000"]
+    else:
+        return []
+
+
+# sphinx gallery config
+sphinx_gallery_conf = {
+    "examples_dirs": os.path.join(
+        os.path.dirname(__file__), "..", "..", "pyqtgraph", "examples"
```
j9ac9k, line 240: probably should wrap with `os.path.abspath` so the console/log output doesn't have directories with `..` throughout.

j9ac9k:
been poking around, in just about all cases, the ones that are raising errors are plots that have QTimers.  I would guess what's happening is that the QTimer is still connected to the update methods, and that timer is still firing after `qtgallery` shuts down the QApplication.  I'm trying to think of what would be a good way to "stop" the QTimer's in each of these examples before having qtgallery kill the QApplication instance.

j9ac9k:
> Hmm, not sure what to make of this. I suspect it's because QScreen.grabWindow supports grabbing contents that aren't part of the application - there's a note here about iOS at least.

I don't think this PR should attempt to address this issue.  Was mostly bringing it up because it surprised me.

pyqtgraph/examples/MultiPlotWidget.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/python
```
ixjlyons, line 1: TODO: re-remove this example

PR #2327, opened 2022-06-13
Update PColorMeshItem.py

LarsVoxen:
self.x and self.y should not be set before detecting if view bounds have changed. 
self._prepareData(args) is now called after the view bounds check

### DIFF
--- a/pyqtgraph/graphicsItems/PColorMeshItem.py
+++ b/pyqtgraph/graphicsItems/PColorMeshItem.py
@@ -205,9 +205,6 @@ class PColorMeshItem(GraphicsObject):
                          matplotlib.pyplot.pcolormesh.html>".
         """
 
-        # Prepare data
-        self._prepareData(args)
-
         # Has the view bounds changed
         shapeChanged = False
         if self.qpicture is None:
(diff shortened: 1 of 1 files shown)


j9ac9k:
Hi @LarsVoxen 

Thanks for the PR.  I agree with your assessment, but I'm curious what issue you ran into that led you to debug this.  Regardless, I'm merging, but I'd be curious to hear more about the specifics of your issue 👍🏻

LarsVoxen:
Hi @j9ac9k 

I have a tool that do live signal analysis of noise and vibration measurement data. I use Pyqtgraph to show results live as we measure.

pcolormesh is used to show a special spectrogram where the spacing is not uniform on the x-axis. It was not updated correctly when new data was measured and added to  Mesh in PcolorMesh. 

I am and have been a happy user of Pyqtgraph for years. Thanks was merging. Makes we feel like I can contribute I little back to a great and very useful project.

j9ac9k:
Ha, several careers and academic programs ago, I did vibration modal analysis and had to write my code in matlab, including some GUI code that is so horrendous I am glad I don't have a copy around anymore.

I wouldn't have thought to use PColorMesh for a spectrogram (most of my recent signal processing work is in analyzing speech), but I can see why it would work well in your use case.

Glad to hear pyqtgraph has worked out for you well for so long.  PColorMesh is a relatively new feature to the library; @pijyoi recently did some significant performance overhauls on it as well; hopefully those improvements have been noticeable in your usage.

If your project is open source, or has some kind of public availability, would love for it to be added to the [used-by](https://github.com/pyqtgraph/pyqtgraph#used-by) section in the README.

pijyoi:
There's also the `NonUniformImage` graphicsItem (using `drawRect`), which appears to have significant overlap with the functionality of `PColorMeshItem` (using `drawConvexPolygon`).

PR #2329, opened 2022-06-17
add support for PySide6's usage of Python Enums

pijyoi:
PySide6 is moving to implement Qt enums as Python Enums (like PyQt6).
in PySide6 6.3.1, it is opt-in by setting environment variable `PYSIDE63_OPTION_PYTHON_ENUM=1`

This PR adds the support by changing the previously PyQt6-specific codepaths to become Python-Enum-specific codepaths. The older implementation of enums on PySide6 continues to work.


j9ac9k:
Thanks @pijyoi for the PR! 

The changes for the remote graphics view renderView method seem small enough that they don't need their own commit (much less PR).

Merging!

PR #2330, opened 2022-06-17
add Atomize to the list of Used By in README.md

Anatoly1010:
Atomize uses pyqtgraph as its main plotting library which allows real-time 1D, 2D and 3D plotting together with a number of useful widgets.


j9ac9k:
I had to scroll far too much before I got to the screenshots :D 

Looks like a great tool, hopefully pyqtgraph has served you well!

PR #2333, opened 2022-06-18
test PyQt version before using sip.array

pijyoi:
it is not sufficient to test for presence of sip.array.
newer versions of PyQt[56]-sip with sip.array can be mixed with older PyQt[56] that doesn't have support.

In particular, a fresh installation of PyQt5 5.12.x (that doesn't have support for sip.array) would pull in the newer PyQt5-sip 12.11.0 with sip.array.


j9ac9k:
Thanks for the follow up on this!

PR #2334, opened 2022-06-19
Update CI Config

j9ac9k:
This PR does a number of small changes

* Expands conda based pipelines to test both pyqt5 and pyside2 bindings
* Uses mambaforge instead of miniconda (much faster environment resolution)
* On conda configs, explicitly state qt-main to not download qt-webengine
* On conda configs install pytest via pip (hopefully have slightly faster dependency resolution)
* On pip, use pyside6-essentials for the current pyside6 pipeline
* Bump required numpy version to 1.20 per NEP-29
* Add dependabot config file to keep our github actions config up to date.

### DIFF
--- a/.github/workflows/etc/environment-pyqt.yml
+++ b/.github/workflows/etc/environment-pyqt.yml
@@ -8,6 +8,4 @@ dependencies:
   - nomkl
   - scipy
   - pyopengl
-  - pytest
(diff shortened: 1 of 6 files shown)


j9ac9k:
Love seeing all the green on the CI 😄

PR #2346, opened 2022-06-22
specify python version for mybinder

pijyoi:
By default, Python 3.7 is provided, which means that the binder launch button has not been working since 4bd94ccc4cede2de0b3b8ddc64681762c5963d24.
Since there's only ~10 months left for NEP-29 to bump up to Python 3.9, we choose to jump from Python 3.7 to 3.9.


j9ac9k:
Bah, should have added binder checks to that PR template I just created

PR #2350, opened 2022-06-27
Add helpful comment about transformation order

gary-harlow:
Transformation operations are non-commutable. I spent ages trying to convert old code that used the .scale and .setPos and was acting strangely. In my case I needed to call translate first and then scale.

### DIFF
--- a/doc/source/images/gen_example_imageitem_transform.py
+++ b/doc/source/images/gen_example_imageitem_transform.py
@@ -17,6 +17,7 @@ class MainWindow(pg.GraphicsLayoutWidget):
         # Example: Transformed display of ImageItem
 
         tr = QtGui.QTransform()  # prepare ImageItem transformation:
+        #Order of transforms are important here, you might want the other way around. 
         tr.scale(6.0, 3.0)       # scale horizontal and vertical axes
         tr.translate(-1.5, -1.5) # move 3x3 image to locate center at axis origin
 


pijyoi:
The relevant section in the Qt docs: https://doc.qt.io/qt-6/qtransform.html#combining-transforms

What may be confusing:
If you use the multiply operator, the order of application of transforms is from left to right. (As in the docs above)
But if you use the dot operator, then the order is from right to left.
```python
In [38]: QTransform().scale(0.5, 1.0) * QTransform().rotate(45) * QTransform().translate(50, 50)
Out[38]: PySide6.QtGui.QTransform(0.353553, 0.353553, 0.000000, -0.707107, 0.707107, 0.000000, 50.000000, 50.000000, 1.000000)

In [39]: QTransform().translate(50, 50).rotate(45).scale(0.5, 1.0)
Out[39]: PySide6.QtGui.QTransform(0.353553, 0.353553, 0.000000, -0.707107, 0.707107, 0.000000, 50.000000, 50.000000, 1.000000)
```

j9ac9k:
I'm good with accepting a comment like this into the code but the comment should link to the upstream documentation which @pijyoi was kind enough to provide

NilsNemitz:
Shouldn't the example start by demonstrating `setRect()` which exists so that the user does not need to deal with transformations and their ordering for most normal applications?

j9ac9k:
> Shouldn't the example start by demonstrating `setRect()` which exists so that the user does not need to deal with transformations and their ordering for most normal applications?

probably, I would argue that's a bit out of scope for a PR that aims to add a comment tho 😆

NilsNemitz:
Agreed, just pointing out that we might be able to spare some users the hassle of dealing with transforms manually :)

pijyoi:
For the record, the code in question here is:
```python
tr = QtGui.QTransform()  # prepare ImageItem transformation:
#Order of transforms are important here, you might want the other way around. 
tr.scale(6.0, 3.0)       # scale horizontal and vertical axes
tr.translate(-1.5, -1.5) # move 3x3 image to locate center at axis origin
```
And it's doing translation first followed by scale. (Using the dot operator)

The same thing is being done in the following example:
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/examples/imageAnalysis.py
```python
# set position and scale of image
tr = QtGui.QTransform()
img.setTransform(tr.scale(0.2, 0.2).translate(-50, 0))
```
Where an image of width x height (100, 200) is displayed at x = [-10, 10] and y = [0, 40]

So perhaps in Python code, it's more "natural" to chain the SRT transforms together with the dot operator.

j9ac9k:
On further reflection, I'm going to close this PR.  I'm not opposed to having better documentation regarding the appropriate usage of `QTransform` (or ideally `setRect`). The place to do that is not an inline comment in one of the examples.  It should go someplace in the documentation, with a link to upstream docs as well.

PR #2356, opened 2022-07-09
Use property to lazily declare rectangle

### LINKED ISSUE #2352: Lazy rectangle selection box contruction breaks subclassed ViewBoxes
### Short description
#2168 introduced creating selection boxes lazily. This is, however, compatibility breaking for any classes that extend `ViewBox` and expect the attribute to behave as it did. For example, we use these in Orange here and I am sure this is not the only place we touch it:
https://github.com/biolab/orange3/blob/269d7fa75c0f36f69b318d925c080b3626d65837/Orange/widgets/visualize/utils/plotutils.py#L156

Should we work around the issue or are you perhaps planning on reverting this? Were there any substantial gains achieved by that particular optimization? Does anyone open thousands of viewboxes quickly?

j9ac9k:
Fixes #2352

This PR redoes #2168 in a way that we hope should make everyone happy.  Thank you to @ksunden 's suggestion [here](https://github.com/pyqtgraph/pyqtgraph/issues/2352#issuecomment-1178362902) for a good way of going about it.

@markotoplak @danielhrisca Please take the opportunity to test this branch and see if this works for your use-case.

Thanks!

### DIFF
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -204,8 +204,7 @@ class ViewBox(GraphicsWidget):
         self.borderRect.setZValue(1e3)
         self.borderRect.setPen(self.border)
 
-        self.rbScaleBox = None
-        self._add_rectangle_selection()
+        self._rbScaleBox = None
 
         ## show target rect for debugging
         self.target = QtWidgets.QGraphicsRectItem(0, 0, 1, 1)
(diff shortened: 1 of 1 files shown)


ksunden:
depending on how it is used, you may add a setter

j9ac9k:
> depending on how it is used, you may add a setter

Yeah I thought about that but wasn't sure what that should look like.  We never supported setting your own QRectF instance, so I didn't think we should try supporting `self.rbScaleBox =...` but maybe we should...

j9ac9k:
Now a setter is supported; I'm thinking that in a future PR we will want to deprecate `rbScaleBox` attribute and instead use the standard Qt getter and setter methods; but that's for another time.

pyqtgraph/graphicsItems/ViewBox/ViewBox.py
```diff
@@ -232,8 +231,34 @@ def __init__(self, parent=None, border=None, lockAspect=False, enableMouse=True,
...
+    def rbScaleBox(self):
+        if self._rbScaleBox is None:
+            # call the setter with the default value
+            scaleBox = QtWidgets.QGraphicsRectItem(0, 0, 1, 1)
+            scaleBox.setPen(fn.mkPen((255, 255, 100), width=1))
+            scaleBox.setBrush(fn.mkBrush(255, 255, 0, 100))
+            scaleBox.setZValue(1e9)
+            scaleBox.hide()
+            self._rbScaleBox = scaleBox
+            self.addItem(scaleBox)
```
pijyoi, line 244: should also require `ignoreBounds=True` to preserve old behavior.

j9ac9k:
Thanks for  checking @markotoplak !

PR #2357, opened 2022-07-10
Invalidate cached properties on geometryChanged signal

### LINKED ISSUE #2345: Axes zoom area does not resize in 0.12.4
### Short description
When a plot is first generated with a given size, say `width ` and `height` in pixels, the entire axes areas on the left and bottom of the plot are zoomable. When the plot size is resized such that the plot is larger (e.g. the window is grabbed along an edge or corner and expanded), only the top `height` portion of the y-axes or the left `width` portion of the x-axis retain the ability to zoom the plot using the scroll wheel. The axes area outside (i.e. the lower portion of the y-axis or the right portion of the x-axis) are not zoomable. If hovering over the plot, not on an axes, there is no issue with zooming. Reverting to 0.12.3 fixes issue.

I was able to reproduce this in both custom plots and many of the pyqtgraph.example scripts.

### Tested environment(s)

 * PyQtGraph version: 0.12.4
 * Qt Python binding: PySide2 5.15.2.1 Qt 5.15.2
 * Python version: 3.7
 * NumPy version: 1.21.6
 * Operating system: Windows 10 Enterprise 21H2
 * Installation method: pip

j9ac9k:
Fixes #2345 

Since #2198 GraphicsWidget attempts to cache properties such as boundingRect; unfortunately it does not appear those caches were being appropriately reset.  Luckily the fix was straightforward as `QGraphicsWidget` has a `geometryChanged` signal that I used to trigger a cache invalidation.

Took the opportunity to clean up some formatting issues with the file so pycharm can stop having a cow.

### DIFF
--- a/pyqtgraph/graphicsItems/GraphicsWidget.py
+++ b/pyqtgraph/graphicsItems/GraphicsWidget.py
@@ -3,36 +3,43 @@ from .GraphicsItem import GraphicsItem
 
 __all__ = ['GraphicsWidget']
 
+
 class GraphicsWidget(GraphicsItem, QtWidgets.QGraphicsWidget):
     
     _qtBaseClass = QtWidgets.QGraphicsWidget
-    def __init__(self, *args, **kargs):
+
+    def __init__(self, *args, **kwargs):
         """
         **Bases:** :class:`GraphicsItem <pyqtgraph.GraphicsItem>`, :class:`QtWidgets.QGraphicsWidget`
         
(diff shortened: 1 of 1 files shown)


j9ac9k:
@danielhrisca if you have some time and want to verify this does not penalize the performance improvement you introduced in #2198 that would be appreciated.

@Seaspider1 if you are able to test to verify this PR fixes your issue, that would be greatly appreciated.

PR #2358, opened 2022-07-11
Add a pyi stub file to import best-guess pyqt type hints

outofculture:
This will help most editors with guessing how the Qt objects should behave.


j9ac9k:
I should add that `pylance` does not read in `.pyi` files, so this seems to only help out when using PyCharm, but still better than nothing!

ntjess:
Just saw this, neat to know it was accepted :)

I added extra things to my `pyi` file to fully replicate what is available from the created `Qt` module. Here is everything after the `ImportError` matrix:

```python
def mkQApp(name=None) -> QtWidgets.QApplication: ...
def exec_() -> int: ...

QAPP: Optional[QtWidgets.QApplication]
QT_LIB: str

USE_PYSIDE: bool
USE_PYQT4: bool
USE_PYQT5: bool

PYSIDE: str
PYSIDE2: str
PYSIDE6: str
PYQT4: str
PYQT5: str
PYQT6: str

def __getattr__(name) -> Any: ...
```

Is it worth including these in the `pyi`
A few things:
  - __Note__ that without `__getattr__`, accessing any of these extra attributes will be seen by PyCharm as an error.
  - QAPP can be `None`, so I think `Optional` is appropriate to add
  - `exec_` returns `int`, not `QApplication`

PR #2359, opened 2022-07-12
fix: super().__init__ does not need self

pijyoi:
`super().__init__(self)` results in a call to `UIGraphicsItem.__init__(self, bounds=self, parent=None)`

UIGraphicsItem expects `"bounds"` to be a `QRectF` and assigns it to `"self._bounds"`.
(`"self._bounds"` ends up not being used at all, but that's another issue...)

At worst, the bug here causes a cyclic reference.


j9ac9k:
Nice catch @pijyoi  Merging!

PR #2385, opened 2022-08-04
Add style option 'hideOverlappingLabels'

bbc131:
* With 'hideOverlappingLabels==False' nothing changes.
  Tick labels might be drawn overlapping with tick labels from
  neighboring plots or overlap with other items or might be simply
  clipped.
* With 'hideOverlappingLabels==True', only tick labels, which are
  fully contained within the items geometry rectangle are drawn.

### Fixes the following problem

The tick labels of `AxisItem` can be drawn outside of `GraphicsWidget.geometry()` rectangle, therefore, they might overlap with neighboring items, e.g. tick labels from other plots or simply be clipped. 

It's very easy to reproduce this with the `GraphicsLayout` example, just zoom and pan the view, to provoke some labels with long values at the axis edges. See also the **following screenshot with overlapping / clipped tick labels**.

overlapping_axis_labels

In comparison to that, the **screenshot below shows the result of this fix** which means the **labels which would overlap or be clipped haven't been drawn**.

overlapping_axis_labels_afterwards

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -59,6 +59,7 @@ class AxisItem(GraphicsWidget):
             'tickTextHeight': 18,
             'autoExpandTextSpace': True,  ## automatically expand text space if needed
             'autoReduceTextSpace': True,
+            'hideOverlappingLabels': True,
             'tickFont': None,
             'stopAxisAtTick': (False, False),  ## whether axis is drawn to edge of box or to last tick
             'textFillLimits': [  ## how much of the axis to fill up with tick text, maximally.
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -1143,6 +1148,7 @@ def generateDrawSpecs(self, p):
                 #self.textHeight = height
                 offset = max(0,self.style['tickLength']) + textOffset
 
+                rect = QtCore.QRectF()
```
j9ac9k, line 1151: are you instantiating here so mypy doesn't have a cow or is there some other reason?

If it's to make type-checking/editors happy, you can probably just do 

```python
rect: QtCore.QRectF  # no actual assignment
```

j9ac9k:
Hi @bbc131 thanks for the PR!

Code diff looks fine, I just left one in-line comment.  Read the docs doesn't like the changes to the docstring, here is the blurb that it is emitting warnings/errors about.

```
/home/docs/checkouts/readthedocs.org/user_builds/pyqtgraph/checkouts/2385/doc/source/../../pyqtgraph/graphicsItems/AxisItem.py:docstring of pyqtgraph.graphicsItems.AxisItem.AxisItem.setStyle:6: ERROR: Unexpected indentation.
/home/docs/checkouts/readthedocs.org/user_builds/pyqtgraph/checkouts/2385/doc/source/../../pyqtgraph/graphicsItems/AxisItem.py:docstring of pyqtgraph.graphicsItems.AxisItem.AxisItem.setStyle:8: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/docs/checkouts/readthedocs.org/user_builds/pyqtgraph/checkouts/2385/doc/source/../../pyqtgraph/graphicsItems/AxisItem.py:docstring of pyqtgraph.graphicsItems.AxisItem.AxisItem.setStyle:12: ERROR: Unexpected indentation.
/home/docs/checkouts/readthedocs.org/user_builds/pyqtgraph/checkouts/2385/doc/source/../../pyqtgraph/graphicsItems/AxisItem.py:docstring of pyqtgraph.graphicsItems.AxisItem.AxisItem.setStyle:13: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/docs/checkouts/readthedocs.org/user_builds/pyqtgraph/checkouts/2385/doc/source/../../pyqtgraph/graphicsItems/AxisItem.py:docstring of pyqtgraph.graphicsItems.AxisItem.AxisItem.setStyle:15: ERROR: Unexpected indentation.
/home/docs/checkouts/readthedocs.org/user_builds/pyqtgraph/checkouts/2385/doc/source/../../pyqtgraph/graphicsItems/AxisItem.py:docstring of pyqtgraph.graphicsItems.AxisItem.AxisItem.setStyle:17: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/docs/checkouts/readthedocs.org/user_builds/pyqtgraph/checkouts/2385/doc/source/../../pyqtgraph/graphicsItems/AxisItem.py:docstring of pyqtgraph.graphicsItems.AxisItem.AxisItem.setStyle:45: WARNING: Definition list ends without a blank line; unexpected unindent.
/home/docs/checkouts/readthedocs.org/user_builds/pyqtgraph/checkouts/2385/doc/source/../../pyqtgraph/graphicsItems/AxisItem.py:docstring of pyqtgraph.graphicsItems.AxisItem.AxisItem.setStyle:45: CRITICAL: Unexpected section title or transition.
```

ixjlyons (commented):
This looks like a nice improvement to me.

I wonder if it makes sense to expose the allowable overlap value itself. I notice matplotlib gives axis tick labels a little extra space outside the axis line, but that could be because of some padding at an Axes or Figure level or something, I'm not sure. I struggle to think of a good name for such a parameter though, which probably means it's better as on/off.

I also don't know about changing behavior to default 0. It seems like a more reasonable default than 15 though I'd say.

pyqtgraph/graphicsItems/AxisItem.py
bbc131: This is only because I don't like having a variable instantiated within an `if` block, which doesn't have an `else` case.

bbc131:
> I wonder if it makes sense to expose the allowable overlap value itself. I notice matplotlib gives axis tick labels a little extra space outside the axis line, but that could be because of some padding at an Axes or Figure level or something, I'm not sure. I struggle to think of a good name for such a parameter though, which probably means it's better as on/off.

I think it makes sense to have this value being equal to the surrounding layout's border margins. So if you use different margins, you may want to change this value.

pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -1159,6 +1165,11 @@ def generateDrawSpecs(self, p):
                 textFlags = alignFlags | QtCore.Qt.TextFlag.TextDontClip    
                 #p.setPen(self.pen())
                 #p.drawText(rect, textFlags, vstr)
+
+                br = self.boundingRect()
+                if not br.contains(rect):
```
bbc131, line 1170: I just realized, that my note in the commit message 
> With 'hideOverlappingLabels==False' nothing changes.

isn't true. Originally, the labels get clipped, if they go over the bounding rect. After my change, they will be hidden, independent of `hideOverlappingLabels`.
So, to get really the same behaviour as originally with `hideOverlappingLabels==False`, one should change the following.

```suggestion
                if self.style['hideOverlappingLabels'] and not br.contains(rect):
```

But, as painting outside of the bounding rect is not allowed (see https://doc.qt.io/qt-5/qgraphicsitem.html#paint), this clipping should additionally be implemented or we simply keep the hiding enabled if it goes out ot the bounding rect.

j9ac9k: Given that painting outside boundingRect I'm good w/ how it currently is

j9ac9k:
> > I wonder if it makes sense to expose the allowable overlap value itself. I notice matplotlib gives axis tick labels a little extra space outside the axis line, but that could be because of some padding at an Axes or Figure level or something, I'm not sure. I struggle to think of a good name for such a parameter though, which probably means it's better as on/off.
> 
> I think it makes sense to have this value being equal to the surrounding layout's border margins. So if you use different margins, you may want to change this value.

that's fairly sensible, would you be up for changing the `0` default value to [`self.layout.getContentsMargins`](https://doc.qt.io/qt-6/qgraphicslayout.html#getContentsMargins).

bbc131:
> > > I wonder if it makes sense to expose the allowable overlap value itself. I notice matplotlib gives axis tick labels a little extra space outside the axis line, but that could be because of some padding at an Axes or Figure level or something, I'm not sure. I struggle to think of a good name for such a parameter though, which probably means it's better as on/off.
> > 
> > 
> > I think it makes sense to have this value being equal to the surrounding layout's border margins. So if you use different margins, you may want to change this value.
> 
> that's fairly sensible, would you be up for changing the `0` default value to [`self.layout.getContentsMargins`](https://doc.qt.io/qt-6/qgraphicslayout.html#getContentsMargins).

I quickly looked into this using the GraphicsLayout example and it seems to be not so straight forward as I thought:

For example the bottom `AxisItem` of "Plot 1" is embedded in a layout handled by the `PlotItem` with default margins of `1`. The visible margins are added by the superior `GraphicsLayout`, named `l` in the example.

So, first problem: It's not easy to decide from which layout the margins should be taken.

Further, this aforementioned superior layout `l` also contains a label on the left side (text: "Long vertical label"). And also on the right side of "Plot 1" is another plot located. So, obviously, using the `contentsMargins` is not really what we need if there are neighboring elements in the layout, one rather should consider the spacing between the elements in this case.

So all in all I would say, that it would be too much to consider all these possibilities. I suggest to stick with a constant default value.

j9ac9k:
Thanks for investigating @bbc131 this LGTM, merging.

PR #2387, opened 2022-08-04
Fix flickering, when panning/scrolling in a fully zoomed-out view

bbc131:
If a `ViewBox` is fully zoomed out (i.e. the current view range equals to the limits), it's still possible to pan the view. 
The values by which the views range changes are extremely small (less than 1e-14), but still this is source of at least some imperfections. Issues I observed are that the `ViewBox.sigRangeChanged` is emitted and that plot or axis might flicker.

### Fix

In `ViewBox.updateViewRange` the desired view range values are compared with the current view range values and if they differ, the `ViewBox.sigRangeChanged` is emitted.
To solve the issue, I replaced the direct comparison of the view-range-values with a threshold based comparison.
So only if the difference between new and old values relative to the total visible span exceeds a value of 1.0e-9 it's considered as a actual change.

### mwe to reproduce the problem

To reproduce the problems, run the mwe below and follow these steps:
1. The view should already be fully zoomed-out but you can ensure it trying to further zoom-out.
2. Now click'n'drag and try to pan the view a few times to the left and a few times to the right.
You'll observe that the "0"- and "1"-tick-labels at the x-axis sometimes (dis)appear, depending on which direction you tried to pan the view.
Further, the printed x-range shows values like 
`x-range: [3.41740524767431e-16, 1.0]` or `x-range: [0.0, 0.9999999999999996]`.

```{python}
import pyqtgraph as pg

app = pg.mkQApp()
plot = pg.PlotWidget()
plot.show()

plot.setMouseEnabled(x=True, y=False)
plot.setLimits(xMin=0.0, xMax=1.0, maxXRange=1.0)
plot.setXRange(min=0.0, max=1.0, padding=0.0)

def print_x_range(plotwidget,view_range):
    print(f"x-range: {view_range[0]}")

plot.sigRangeChanged.connect(print_x_range)

if __name__ == '__main__':
    pg.exec()
```

### DIFF
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -1622,8 +1622,13 @@ class ViewBox(GraphicsWidget):
                     changed[0] = True
                 viewRange[0] = rangeX
 
-
-        changed = [(viewRange[i][0] != self.state['viewRange'][i][0]) or (viewRange[i][1] != self.state['viewRange'][i][1]) for i in (0,1)]
+        # Consider only as 'changed' if the differences are larger than floating point inaccuracies
+        thresholds = [(viewRange[axis][1] - viewRange[axis][0]) * 1.0e-9 for axis in (0,1)]
+        changed = [
+            (abs(viewRange[axis][0] - self.state["viewRange"][axis][0]) > thresholds[axis])
+            or (abs(viewRange[axis][1] - self.state["viewRange"][axis][1]) > thresholds[axis])
+            for axis in (0, 1)
+        ]
         self.state['viewRange'] = viewRange
 
         if any(changed):
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -1622,7 +1622,9 @@ class ViewBox(GraphicsWidget):
                     changed[0] = True
                 viewRange[0] = rangeX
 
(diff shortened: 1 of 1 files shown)


j9ac9k:
Hi @bbc131 

Thanks for the PR, your description of the issue, and the proposed fix.  

My first thought is we could probably do better than 1.0e-9 as the multiplier, and instead use [`self.pixelSize`](https://github.com/pyqtgraph/pyqtgraph/blob/74fa36da37d36a3da52019f673898ff0fe165de0/pyqtgraph/graphicsItems/GraphicsItem.py#L306-L311) but when I was looking that over, I got to thinking it might be too expensive of a call to make.

I would love to get your thoughts on that.

bbc131:
Hi @j9ac9k,

sorry for replying only now, was absent for a while.

I guess you mean something like "consider as changed if view differs by more than a x-th of a pixel". At the first glance I would say this would be easier to understand, but in essence it's more or less the same. You also can't circumvent to define some factor.
Since the origin lies in floating point errors, it's not crucial to use the exact pixel size, so I would stick to the simple version.

j9ac9k:
Hi @bbc131 

No need to apologize!  This looks good to me, only change I would request then is to explicitly state in your comment that 1.0e-9 was chosen arbitrarily to address floating point errors.  When I see constants like that in the code-base I periodically wonder if there was a specific reason that was chosen or not.

I'm good w/ not using `pixelSize` due to your reasoning.

j9ac9k:
LGTM, thanks for the PR!

PR #2389, opened 2022-08-06
Add GraPhysio to the list of projects using pyqtgraph.

jaj42:
This adds GraPhysio to the list of projects using pyqtgraph in README.md. GraPhysio is a GUI program which makes makes it easy to visualize and modify time series data from different input formats. GraPhysio makes heavy use of pyqtgraph.
While here, remove trailing whitespace from the previous line in the file.


j9ac9k:
Thanks for the PR @jaj42 You should add some screenshots to your readme!

jaj42:
Yes you're right, I should add some screenshots. Thanks for the merge!

PR #2395, opened 2022-08-14
map pyqtgraph symbols to a matplotlib equivalent

pijyoi:
This PR maps all the pyqtgraph scatterplot symbols to some matplotlib equivalent such that the `Symbols.py` example can export something to Matplotlib:

master:
before

this PR:
image


j9ac9k:
Nice, having these sorts of things match up between libraries would definitely be something our users appreciate.

LGTM, merging!

PR #2397, opened 2022-08-16
`getHistogramRange` for `HistogramLUTItem`

kremeyer:
`pyqtgraph.HistogramLUTItem` already had the setter `setHistogramRange`, but was missing the getter `getHistogramRange`. I added it.


j9ac9k:
Hi @kremeyer 

Thanks for the PR.  _Normally_ I would say we should move away from `def get<Thing>` naming of methods, as they don't match up w/ the Qt framework's naming scheme, but given that the `HistogramLUTItem` already has numerous methods that follow that naming scheme, I think this is the one place that I'm ok with it.

Thanks for the PR, this LGTM, merging!

PR #2400, opened 2022-08-24
remove MatplotlibWidget from pg namespace

pijyoi:
This PR partially undoes #2366, specifically its import of `MatplotlibWidget` into the `pg` namespace.
Firstly, it adds unnecessary overhead to the import of `pyqtgraph` for users who do not use the Matplotlib Exporter functionality.
Secondly, this import seems to be causing the segfault in one of the CI pipelines (Python 3.9.13 / Linux / PySide2 5.15) for the not-yet-merged #2318.


j9ac9k:
Thanks for your help debugging this and for the PR @pijyoi

PR #1805, opened 2021-05-27
CSVExporter now can export x- and y-axis errors

3fon3fonov:
CSVExporter now can export x- and y-axis errors, if such are introduced by the ErrorBarItem. This is my fix, and it works well for the Exo-Striker tool. Yet, a more elegant code solution could exist.

PR #2405, opened 2022-09-03
Have CSV exporter export error bar information

### LINKED ISSUE #1365: Error bars must be also exported with .csv export
I like the option to export a plot to a .csv so I can make an alternative plot offline (i.e. matplotlib).

However, the .csv export only exports x-,y- and e.g. model. Without the uncertainties, this export is meaningless.
I understand that not all pyqtgraph plots use the "ErrorBarItem", but when it does, the errors must be exported too.

Please, include this in pyqtgraph! :-)

j9ac9k:
This PR is a re-implementation of #1805.  It allows for error bar information to be exported as part of the CSV exporter.

While working on this PR, I replaced the string-concatenation method of creating the CSV file with use of the `csv` module, which results in code that's much easier to read.

Fixes #1365

### DIFF
--- a/pyqtgraph/exporters/CSVExporter.py
+++ b/pyqtgraph/exporters/CSVExporter.py
@@ -1,4 +1,9 @@
-from .. import PlotItem
+import csv
+import itertools
+
+import numpy as np
+
+from .. import ErrorBarItem, PlotItem
 from ..parametertree import Parameter
 from ..Qt import QtCore
 from .Exporter import Exporter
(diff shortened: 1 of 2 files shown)


j9ac9k:
After sleeping on this PR and re-reviewing it.  A couple of things stand-out.

1. The export method is _way_ too long and should be broken up into different functions.
2. The ErrorBarItem extractor doesn't actually get the x and y positions, just the top, bottom, left and right error limits 😬

PR #2406, opened 2022-09-03
QFileSystemModel was incorrectly placed in the QtWidgets module

j9ac9k:
On PySide6, QFileSystemModel is placed in the QtWidgets module, despite the C++ documentation indicating it should be in the QtGui module.  This PR shims PySide6.QtGui so that QFileSystemModel is accessible from it.

### DIFF
--- a/pyqtgraph/Qt/__init__.py
+++ b/pyqtgraph/Qt/__init__.py
@@ -258,6 +258,12 @@ if QT_LIB in [PYQT6, PYSIDE6]:
 
     if not isinstance(QtOpenGLWidgets, FailedImport):
         QtWidgets.QOpenGLWidget = QtOpenGLWidgets.QOpenGLWidget
+
+    # PySide6 incorrectly placed QFileSystemModel inside QtWidgets
+    if QT_LIB == PYSIDE6 and hasattr(QtWidgets, 'QFileSystemModel'):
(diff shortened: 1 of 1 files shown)


pyqtgraph/Qt/__init__.py
```diff
@@ -258,6 +258,12 @@ def _copy_attrs(src, dst):
 
     if not isinstance(QtOpenGLWidgets, FailedImport):
         QtWidgets.QOpenGLWidget = QtOpenGLWidgets.QOpenGLWidget
+
+    # PySide6 incorrectly placed QFileSystemModel inside QtWidgets
+    if QT_LIB == "PYSIDE6" and hasattr(QtWidgets, 'QFileSystemModel'):
```
pijyoi, line 263: Remove the quotes from PYSIDE6.
Everywhere else in the library, we (inconsistently) compare directly against 'PySide6', etc. (Case sensitive)

pijyoi (changes requested):

pyqtgraph/Qt/__init__.py
j9ac9k: 🤦🏻  ... yeah a future change should be to convert all the string representations to lower-case (or upper case) before any evaluation.

PR #2404, opened 2022-09-02
Interactive refactoring

ntjess:
See commit messages for more information. Very minor housekeeping that is easier to perform before widespread adoption of the `Interactor` feature set

PR #2408, opened 2022-09-07
Bugfix: `interact` on decorated method that uses `self`.

ntjess:
The modified example in `tests/parametertree` used to fail before this PR. A result of #2404 was that `inspect.signature` no longer realized that `self` was a bound variable in class-defined, outside-decorated functions. Fix this simply by calling `functionToParameterDict` on the original function, rather than the interactive version.

Also expands `runOpts` to `runOptions` to remain in keeping with fully-expanded argument names elsewhere. `setOpts` and related terms are now the only non-expanded items, since they match the already-present `Parameter` API

### DIFF
--- a/doc/source/parametertree/interactiveparameters.rst
+++ b/doc/source/parametertree/interactiveparameters.rst
@@ -73,24 +73,24 @@ code below is functionally equivalent to above):
 There are several caveats, but this is one of the most common scenarios
 for function interaction.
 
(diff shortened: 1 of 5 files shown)


j9ac9k:
congratulations @ntjess for your first merge to the repo! ...while you're at it ... (kidding!) Thanks for the PR!

PR #2412, opened 2022-09-09
no need to specially disable test_loadUiType for conda

pijyoi:
Probably due to the availability of PySide2 5.15, `test_loadUiType` now works on conda.


j9ac9k:
Nice find @pijyoi  indeed we can nuke this thanks to newer conda packages.

PR #2413, opened 2022-09-12
Fix doubling labels in DateAxisItem

bbc131:
### In short
* Due to floating point errors, comparison for equality sometimes fails to detect doubling values. Use approximative comparison instead. Code is copied from AxisItem.tickValues().

### In detail

In some situations `DateAxisItem` draws two different tick labels (of different levels) for the same value. See the following screenshot, where "60.0" and "02:01:00" are drawn. "60.0" should be skipped and only "02:01:00" should be drawn.
This can be reproduced, using the `DateAxisItem` example, simply zoom in like in the screenshot. There, the visible range is something between 0.01s and 0.004s. It seems to happen at each full minute. Note, that the occurence depends also on the minimum axis value, meaning it may be necessary to pan the axis a bit to the left or right always keeping the full minute within the visible range.

grafik

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -820,7 +820,11 @@ class AxisItem(GraphicsWidget):
             ## remove any ticks that were present in higher levels
             ## we assume here that if the difference between a tick value and a previously seen tick value
             ## is less than spacing/100, then they are 'equal' and we can ignore the new tick.
-            values = list(filter(lambda x: np.all(np.abs(allValues-x) > spacing/self.scale*0.01), values))
+            close = np.any(
+                np.isclose(allValues, values[:, np.newaxis], rtol=0, atol=spacing/self.scale*0.01)
+                , axis=-1
+            )
+            values = values[~close]
             allValues = np.concatenate([allValues, values])
             ticks.append((spacing/self.scale, values))
 
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -826,7 +826,7 @@ class AxisItem(GraphicsWidget):
             )
             values = values[~close]
             allValues = np.concatenate([allValues, values])
-            ticks.append((spacing/self.scale, values))
+            ticks.append((spacing/self.scale, values.tolist()))
 
         if self.logMode:
             return self.logTickValues(minVal, maxVal, size, ticks)
--- a/pyqtgraph/graphicsItems/DateAxisItem.py
+++ b/pyqtgraph/graphicsItems/DateAxisItem.py
@@ -170,7 +170,6 @@ class ZoomLevel:
             # remove any ticks that were present in higher levels
             # we assume here that if the difference between a tick value and a previously seen tick value
             # is less than min-spacing/100, then they are 'equal' and we can ignore the new tick.
-            ticks = np.array(ticks)
             close = np.any(
                 np.isclose(allTicks, ticks[:, np.newaxis], rtol=0, atol=minSpc * 0.01),
                 axis=-1,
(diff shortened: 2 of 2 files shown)


j9ac9k:
Hi @bbc131 

Thanks so much for your work on DateAxisItem, it's one of the relatively newer parts of the library, and clearly has some rough corners.  I appreciate your work on improving it.

Slight nitpick suggestion, no need to cast to `list` the `filter` object; extending a list with a filter object will address that conversion for you. 

An alternative would be to have `allTicks` be a numpy array (like `allValues` in AxisItem), and concatenate the array instead of extending it.

pijyoi:
A numpy broadcasted version:
```python
ticks = np.array(ticks)
close = np.any(np.isclose(allTicks, ticks[:, np.newaxis], rtol=0, atol=minSpc*0.01), axis=-1)
ticks = ticks[~close].tolist()
```

bbc131:
Hi @j9ac9k and @pijyoi,
Thanks for your fast response and feedback.
I like the suggestion of @pijyoi, and implemented it, also for `AxisItem`. I like it, if the code of both classes differs as little as possible.

pyqtgraph/graphicsItems/DateAxisItem.py
```diff
@@ -168,9 +168,16 @@ def tickValues(self, minVal, maxVal, minSpc):
             # reposition tick labels to local time coordinates
             ticks += self.utcOffset
             # remove any ticks that were present in higher levels
-            tick_list = [x for x in ticks.tolist() if x not in allTicks]
-            allTicks.extend(tick_list)
-            valueSpecs.append((spec.spacing, tick_list))
+            # we assume here that if the difference between a tick value and a previously seen tick value
+            # is less than min-spacing/100, then they are 'equal' and we can ignore the new tick.
+            ticks = np.array(ticks)
```
pijyoi, line 173: I suppose this line is redundant since `ticks` is already an ndarray, something that wasn't obvious to me at the time.

bbc131: You're right, I updated it.

pyqtgraph/graphicsItems/DateAxisItem.py
```diff
@@ -168,9 +168,15 @@ def tickValues(self, minVal, maxVal, minSpc):
...
-            valueSpecs.append((spec.spacing, tick_list))
+            # we assume here that if the difference between a tick value and a previously seen tick value
+            # is less than min-spacing/100, then they are 'equal' and we can ignore the new tick.
+            close = np.any(
+                np.isclose(allTicks, ticks[:, np.newaxis], rtol=0, atol=minSpc * 0.01),
+                axis=-1,
+            )
+            ticks = ticks[~close]
+            allTicks = np.concatenate([allTicks, ticks])
+            valueSpecs.append((spec.spacing, ticks))
```
pijyoi, line 179: To respect the description in the method docstring, `ticks` would need to be converted to a Python list here.

pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -820,7 +820,11 @@ def tickValues(self, minVal, maxVal, size):
...
             ## we assume here that if the difference between a tick value and a previously seen tick value
             ## is less than spacing/100, then they are 'equal' and we can ignore the new tick.
-            values = list(filter(lambda x: np.all(np.abs(allValues-x) > spacing/self.scale*0.01), values))
+            close = np.any(
+                np.isclose(allValues, values[:, np.newaxis], rtol=0, atol=spacing/self.scale*0.01)
+                , axis=-1
+            )
+            values = values[~close]
             allValues = np.concatenate([allValues, values])
             ticks.append((spacing/self.scale, values))
```
pijyoi, line 829: To respect the description in the method docstring, `ticks` would need to be converted to a Python list here.

bbc131: Thanks for being so attentive. I see, you had this correct already in your very first comment, but I removed it silently ;)

j9ac9k:
LGTM, thanks @bbc131 for the PR thanks @pijyoi  for the review/feedback!

PR #2414, opened 2022-09-13
Promote interactive `Run` action to group level

ntjess:
Much less screen real estate is required if an `interactive` parameter's "Run" button is associated with the group rather than as a subchild (similar to the `pen` parameter):
image
A very nice additional benefit is being able to run the function even when all parameters are not expanded:
image

Because "Run" buttons would no longer clutter a parameter and make unnecessary children, there is minimal drawback to making `[RunOptions.ON_CHANGED, RunOptions.ON_ACTION]` the default instead of `RunOptions.ON_CHANGED`.

To facilitate this change, minor refactoring of `action`'s button meant more of its options (`icon`, `shortcut`, other button properties as desired) could be exposed as parameter `opts`.

**Note:** The current `shortcut` behavior means an action only listens to the shortcut when any of its items are visible. So if a tree is hidden (or no parameter items were spawned), the shortcut doesn't work. There are many other ways of implementing parameter shortcuts, but this is the most pain-free (doesn't require the user to specify a widget parent in `opts`, doesn't depend on shortcut parent's life cycle). If this behavior is unintuitive, my next plan would be to force the user to provide a `shortcutParent` (or something similar) if `shortcut` is specified. This is the widget whose focus determines whether the shortcut will be listening. The downside is if this object is deleted, it can lead to confusing behavior. However, it would allow the shortcut to be tied to the parameter itself, rather than an instantiated button. _(I could also simply remove `shortcut` from the list of available options, but I find it useful enough)_

- [x] TODO: Update `ActionParameter` docs to incorporate new options once the discussion regarding `shortcut` is settled.

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes/action.py
+++ b/pyqtgraph/parametertree/parameterTypes/action.py
@@ -3,6 +3,39 @@ from ..Parameter import Parameter
 from ..ParameterItem import ParameterItem
 
 
+class ParameterControlledButton(QtWidgets.QPushButton):
+    settableAttributes = {"title", "tip", "icon", "shortcut", "enabled", "visible"}
+
+    def __init__(self, parameter=None, parent=None):
+        super().__init__(parent)
+        if not parameter:
+            return
+        parameter.sigNameChanged.connect(self.onNameChange)
+        parameter.sigOptionsChanged.connect(self.updateOpts)
+        self.clicked.connect(parameter.activate)
+        self.updateOpts(parameter, parameter.opts)
+
+    def updateOpts(self, param, opts):
+        # Of the attributes that can be set on a QPushButton, only the text
+        # and tooltip attributes are different from standard pushbutton names
+        nameMap = dict(title="text", tip="tooltip")
+        # Special case: "title" could be none, in which case make it something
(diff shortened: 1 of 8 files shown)


j9ac9k:
I think shortcuts not working if a parameter is hidden is fine; especially for an initial go.  If people complain, we can certainly revisit!

pyqtgraph/parametertree/parameterTypes/action.py
```diff
@@ -29,26 +68,24 @@ def treeWidgetChanged(self):
...
-    def buttonClicked(self):
-        self.param.activate()
-
 
 class ActionParameter(Parameter):
     """Used for displaying a button within the tree.
 
     ``sigActivated(self)`` is emitted when the button is clicked.
+
+    ============== ============================================================
```
j9ac9k, line 79: Would you mind changing this docstring to numpy style docstring?  We are hoping to migrate the rest of the repo to that, but it may be a while.

Also I think it should be 

```
:class:`QIcon`  and :class`QKeySequence`
```

Our docs are now smart enough to link to the upstream Qt docs for Qt types like that 👍🏻

PR #2418, opened 2022-09-17
Add iterations argparse argument to benchmarks

j9ac9k:
Add `--iterations` argument to the benchmark examples allowing the benchmark to exit on its own cleanly after `n` updates.  This is useful for profilers which need the application to exit before computing results.

EDIT: Below was the original PR post, decided to walk back the scope of this PR by scrapping the changes to try and compute the `setData` + `paint` duration

Add functionality to allow for the benchmark to stop after a specific number of iterations.  By default, will run indefinitely as it currently does, however, if an iteration argument is passed with a numerical value, the update timer will stop, and the example will exit cleanly.

~~This is my first attempt at using QSignalMapper, while I would normally use `self.sender()`, the update method is a function, not a method, where `self` is not an argument being passed in.~~

QEventLoop to block the python code until the paint function finishes.  This should provide for a far more accurate duration calculation than `QApplication.processEvents()`

Redo the FPS calculation to take into account the entirety of the Qt EventLoop, not just the call to paint.

### DIFF
--- a/pyqtgraph/examples/MultiPlotSpeedTest.py
+++ b/pyqtgraph/examples/MultiPlotSpeedTest.py
@@ -2,12 +2,21 @@
 """
 Test the speed of rapidly updating multiple plot curves
 """
+import argparse
+import itertools
 
 import numpy as np
+from utils import FrameCounter
 
 import pyqtgraph as pg
 from pyqtgraph.Qt import QtCore
-from utils import FrameCounter
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--iterations', default=float('inf'), type=float,
+    help="Number of iterations to run before exiting"
+)
+args = parser.parse_args()
+iterations_counter = itertools.count()
 
 # pg.setConfigOptions(useOpenGL=True)
 app = pg.mkQApp("MultiPlot Speed Test")
@@ -37,7 +46,10 @@ data = np.random.normal(size=(nPlots*23,nSamples))
 ptr = 0
 def update():
     global ptr
-
+    if next(iterations_counter) > args.iterations:
+        timer.stop()
+        app.quit()
+        return None
     for i in range(nPlots):
         curves[i].setData(data[(ptr+i)%data.shape[0]])
 
--- a/pyqtgraph/examples/PlotSpeedTest.py
+++ b/pyqtgraph/examples/PlotSpeedTest.py
@@ -4,15 +4,15 @@ Update a simple plot as rapidly as possible to measure speed.
 """
 
 import argparse
+import itertools
 
 import numpy as np
+from utils import FrameCounter
 
 import pyqtgraph as pg
 import pyqtgraph.functions as fn
 import pyqtgraph.parametertree as ptree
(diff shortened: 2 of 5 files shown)


j9ac9k:
I should probably update VideoSpeedTest.py in a similar fashion.

ntjess:
Something scary that happens on Windows 11

image

j9ac9k:
@ntjess what Qt bindings were you using to create this issue?

j9ac9k:
Also on `ScatterPlotSpeedTest.py` when I switch to simulating hover events, I get the following console output:

```
QEventLoop::exec: instance 0x6000013995e0 has already called exec()
```

Debating if this is "ok" or if I should add a little more logic to try and account for this.

ntjess:
> @ntjess what Qt bindings were you using to create this issue?

PyQt5 5.15.7 Qt 5.15.2

pijyoi:
I am getting the same error on Windows 11, Qt5 and Qt6.

The following 2 changes seems to make things work.

```python
flags = QtCore.QEventLoop.ProcessEventsFlag.AllEvents   # equivalent to no flags
```
```python
curve.sigPaintFinished.connect(eventLoop.exit, QtCore.Qt.ConnectionType.QueuedConnection)
```

The problem as I see it is nested callbacks, the chain of which gets broken by using `QueuedConnection`.
This is by no means the "right" solution, nor do I know what a "right" solution would be.

I do encounter such issues when using event loops. i.e. wanting to execute some piece of code _after_ the current callback has completed. Qt doesn't have a convenient ready-made "CallAfter" function.

j9ac9k:
If I understand things correctly, using `QtCore.Qt.ConnectionType.QueuedConnection` would be the same as using `QApplication.processEvents()`.  I want to know how long it takes from the time we call `setData` to the time the following screen render is done (I know the _actual_ rendering on the screen may happen a bit after the `paint` method finishes, but that's the best I can do). 

Using the `QueuedConnection` (or `QApplication.processEvents()` would add the processing of other events in the Qt event loop which I don't really want to muddy the water with the duration benchmark (the FPS calculation should take that into account).

That said, there may not be a good way to capture the performance here short of much fancier signal/slot logic, in which case I should probably unwind the bulk of the diff that this PR has.

j9ac9k:
I'm away from my windows machine for a few days but some potential solutions I thought of

1. After we collect our measurement info in the `update` method, we can then run `QApplication.processEvents()`
2. After we `sigPaintFinished` is emitted, we then disconnect (or block) that signal, until the update method is called again, where we then re-connect (or unblock) it.

The first option _may_ impact the fps calculation in a way that is not accurate.

ntjess:
Still happens on 4c9b9268

j9ac9k:
Ok, after merging #2752 decided to revisit this and re-limit the scope. 

This PR only adds the `--iterations` argument to the benchmarks, allowing users to specify the number of iterations before the benchmark exits automatically.

This is useful when running a benchmark within a profiler, and developers need the application to exit before you can view the results.

PR #2419, opened 2022-09-17
Add GLTextItem to docs

### LINKED ISSUE #2037: Text in 3D Graphics
API change #1525 makes `GLViewWidget` inherit `QOpenGLWidget` instead of `QGLWidget`.
The `QOpenGLWidget` does not have text rendering capabilities.

In my application I make use of displaying text in a 3D Graphic: I use it to label some axis of a 3D plot. 
I did this, based on solution from [Stackoverflow](https://stackoverflow.com/questions/64600904/plot-text-in-3d-plot-that-does-not-scale-or-move), which provides a `GLTextItem` class.
Under the hood, this class relies on the `QGLWidget.renderText` method for drawing text.

* Are there any methods for rending text in the 3D Graphics?
* Would it make sense to have a GLTextItem in your 3D API, alongside the [current items](https://pyqtgraph.readthedocs.io/en/latest/3dgraphics/index.html)?

jebguth:
Fixes #2037.

Google did not give me the suggestion to use GLTextItem when trying to find a way to get text in a 3D scene. Documentation didn't help either. After an hour or two I ended up in #2037 with great happiness. Thought I should fix it to help anyone else in the same situation.

This is my first contribution ever to an open source project. Even if this is a miniscule change, please tell me if I am doing something wrong.

### DIFF
--- a/doc/source/3dgraphics/gltextitem.rst
+++ b/doc/source/3dgraphics/gltextitem.rst
@@ -1,0 +1,7 @@
+GLTextItem
+==========
+
+.. autoclass:: pyqtgraph.opengl.GLTextItem
+    :members:
+
+    .. automethod:: pyqtgraph.opengl.GLTextItem.__init__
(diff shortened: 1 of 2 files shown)


j9ac9k:
Oh my, completely forgot about this! Thanks for the PR @jebguth !

PR #2420, opened 2022-09-18
deprecate GraphicsObject::parentChanged method

pijyoi:
QGraphicsObject has a signal named `parentChanged`.
Although Qt bindings "permit" a method name to be the same as a Qt signal name, it may not be a good idea to make use of this "feature".
This "feature" has been broken before: #2132
It also relies on the inheritance order of the parent classes of `GraphicsObject`.

This PR renames the method name while emitting a deprecation warning if it detects that the old name is being used by user sub-classes.

One way to test out this PR is to modify `examples/customGraphicsItem.py`.
By adding the following method:
```python
def parentChanged(self):
    super().parentChanged()
```
We can verify that running this example with `python -Wdefault` emits a deprecation warning.

### DIFF
--- a/pyqtgraph/graphicsItems/GraphicsItem.py
+++ b/pyqtgraph/graphicsItems/GraphicsItem.py
@@ -452,13 +452,16 @@ class GraphicsItem(object):
                 #print "   --> ", ch2.scene()
             #self.setChildScene(ch2)
 
-    def parentChanged(self):
+    def changeParent(self):
         """Called when the item's parent has changed. 
         This method handles connecting / disconnecting from ViewBox signals
         to make sure viewRangeChanged works properly. It should generally be 
         extended, not overridden."""
         self._updateView()
-        
+
+    def parentChanged(self):
+        # deprecated version of changeParent()
+        GraphicsItem.changeParent(self)
 
     def _updateView(self):
         ## called to see whether this item has a new view to connect to
--- a/pyqtgraph/graphicsItems/GraphicsObject.py
+++ b/pyqtgraph/graphicsItems/GraphicsObject.py
@@ -1,3 +1,4 @@
+import warnings
 from ..Qt import QT_LIB, QtCore, QtWidgets
 from .GraphicsItem import GraphicsItem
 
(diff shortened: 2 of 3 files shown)


pijyoi:
The detection only works for the simplest case where the method is at the last level in an inheritance hierarchy.

pijyoi:
Based on a search of the repo, #1991 led me to https://github.com/sem-geologist/HussariX/blob/master/lib/ui/CustomPGWidgets.py#L98 which has a class that subclasses `pg.ScaleBar`.
The current detection technique would fail there.

j9ac9k:
@sem-geologist has been active in the issue tracker here before; hopefully this ping will get their attention

pijyoi:
The detection would think that CustomScaleBar doesn't use the deprecated method parentChanged, not knowing whether any parent class does override. 

It would then call self.parentChanged_() which would correctly invoke ScaleBar::parentChanged_().

It works in this case only because the parent (pyqtgraph) class was updated to use parentChanged_().

j9ac9k:
Hi @pijyoi 

Thanks for the PR!.  The diff looks good to me, but I'm a little hung up on the name, `parentChanged_`.  The name does achieve the primary objective, which is to differentiate the method from the signal name, but the past-tense wording is in-line with the name of a signal.  An alternative would be `changeParent`, but I'm not set on that.  I'm open to other suggestions.

pijyoi:
A contrived example where the current detection technique fails.
```python
import pyqtgraph as pg 

class CustomParent(pg.GraphicsObject):
    def __init__(self):
        self.text = "OLD PARENT"

    def boundingRect(self):
        return QtCore.QRectF(0, 0, 100, 100)

    def paint(self, p, *args):
        p.setPen(pg.mkPen('w'))
        p.drawText(0, 50, self.text)

    def parentChanged(self):
        self.text = "NEW PARENT"

class CustomChild(CustomParent):
    def __init__(self):
        CustomParent.__init__(self)
        self.text = "CHILD"

pg.mkQApp()
pw = pg.PlotWidget()
pw.show()
custom = CustomChild()
pw.addItem(custom)
pg.exec()
```

j9ac9k:
Maybe I have the wrong attitude about this, but I'm not so concerned about catching every usage of this method... this method is buried fairly deep in the library, and I wouldn't imagine that many end-users overwriting it...

Anyway, I'm fine w/ the current code diff.

pijyoi:
I am generally unfamiliar with all the viewbox machinery. The docstring for `parentChanged()` says `"It should generally be extended, not overridden.`. And yet the only use of `parentChanged()` within the library (`ScaleBar`) doesn't call the parent method.

Maybe merge this "breaking" change after 0.13.0?

j9ac9k:
Since we just released 0.13.0, let's merge this one!   Thanks for your work here @pijyoi

PR #2426, opened 2022-09-24
update multiplePlotSpeedTest.py to use PlotCurveItem instead of QGraphicsPathItem

pijyoi:
This example demonstrates how to plot multiple curves (all of the same color) using a single `Item`.
The original version uses `QGraphicsPathItem`, but apparently using `PlotCurveItem` is several times faster.


j9ac9k:
I was just thinking, we should have an example showing how to do this... Thanks @pijyoi this LGTM, merging!

PR #2437, opened 2022-09-28
Refactor examples using `Interactor` to run on changing function parameters

ntjess:
Changes from #2432 meant examples which relied on default changing parameters needed to be updated. The examples still work as-is, but buttons must be pressed which is unintuitive.


j9ac9k:
Thanks @ntjess for the prompt fix! Merging!

PR #2443, opened 2022-09-29
Parameter: added helper functions (param, group, action) for leaner/easier parameter definitions

Zorkator:
These functions can be optionally used for easier definitions of parameter trees.
The example/parametertree and ImageExporter is updated accordingly.


ntjess:
@Zorkator I've thought about this for a while as well -- thanks for the submission. However, there are a few considerations with the approach in this PR:

* If this was merged, we should have a solution that equally treats all parameters, rather than only implementing shorthand calls for some types. I'm thinking along the lines of
    ```python
    from pyqtgraph.parametertree.Parameter import PARAM_TYPES, Parameter
    from functools import partial

    class ParamGetter:
        def __getattr__(self, parameterType):
            if parameterType not in PARAM_TYPES:
                raise AttributeError(parameterType)
            return partial(self.create, type=parameterType)

        @staticmethod
        def create(name, type=None, value=None, **opts):
            """
            copied from your PR
            """
            type = type or value.__class__
            typeId = getattr(type, "__name__", type)
            return Parameter.create(name=name, type=typeId, value=value, **opts)

        # Can overload things that need special treatment, like:
        def group(self, name, value=None, **opts):
            return Parameter.create(name=name, type="group", children=value, **opts)


    param = ParamGetter()


    # Usage:
    # param.action("myAction")
    # param.color("my color", value="#ff0")
    # ...
    ```

    This prevents lots of duplication during implementation

* I've strayed from submitting this PR in the past due to the disconnect between *value* types and *parameter* types. For instance, the `list` parameter doesn't *actually* handle list values. It handles scalar values that can be selected from a dropdown list. This distinction might be confusing by someone attempting `param.list([a,b,c])` and getting string-like values. There are ways around this though; I've considered a PR that makes a `valueType` module where you can request parameters by value type (i.e. `int` could give you a spinbox or slider, since both handle int types) but this would need more work

In general though, because of the second reason, I would be hesitant to merge this PR until value and parameter types were more synchronized. What are your thoughts?

Zorkator:
@ntjess Thanks for your explanations, I see your point. Of course, it would be best to treat all parameter types the same way.
Actually, my approach just focused on simplicity and getting the flood of parenthesis and brackets under control.
But I like your approach better, because it allows both, using the generic create, and handling some types individually, doing sanity checks or inferring option defaults. **So, I'd vote for your PR** (proposing changes in `group` though: `def group( self, name, *children, **opts)` to avoid the need of putting the children in brackets).

Your second reason seems to me a quite different problem and should be fixed by not using a list for what is actually an enum-type parameter. I had exactly the same problem in another application which uses Qt-Delegates to display/edit config values of various types. After a while, trying to use a `list` for specifying the valid options, I kicked it out, since the "current selection" is missing and has to be stored somehow else, like a "current index" attached to the list, "integrated" by some special list item etc... all too error-prone.
In the end, I replaced it by python's `Enum` which is just handled in a ComboBox:
```python
from enum import Enum 
 
@StyledItemDelegate.map(Enum) 
#--------------------------------------------- 
class EnumEditor(QComboBox): 
#--------------------------------------------- 
 
    def value( self ): 
        return self._ET[ self.currentText() ] #< create new Enum-instance for ComboBox' current selection.
 
    def setValue( self, val ):
        "populate newly created ComboBox by Enum-instance `val`"
        self._ET = type(val)
        [ self.addItem( i.name, i.value ) for i in self._ET ] #< fill ComboBox by valid Enum-options 
        self.setCurrentIndex( self.findText( val.name ) )  # < and set the current 
 
    @classmethod 
    def styleItem( _class, option, item, index, editorOf ): 
        option.text = item.name 
```
Guess, you get the idea, even if you're not familiar with delegates (or the decorator here that just registers the class for handling Enums): Since `Enum` gives you a clear type that contains both, the list of valid options and the current selection it matches perfectly to ComboBoxes (dropdown lists) and makes the implementation easy, since with an Enum-instance you always have all necessary data at hand.

ntjess:
Can you clarify what you mean by the "current selection" being missing? To my knowledge the `list` parameter's `value()` function will give you the current selection.

In my earlier comment I meant to just clarify that the _value_ type only equals the _parameter_ type for ints/floats, bools, and strings. In all other cases, setting the default parameter type to `type(value)` will not be beneficial. This behavior would make more sense if a separate registry kept track of which parameters were delegates for which value types. Something like this:

```python
# I want int values to spawn a `slider` and QColor parameters to spawn `color`
typeRegistry.add(int, "slider")
typeRegistry.add(type(pg.mkColor()), "color")
# ...
```

Now, an unspecified `value` parameter knows what the default should be. But the `typeRegistry` mechanism would have to be more carefully thought out before this is merged.

In the case of a `list` _value_, I would want `parameter.value()` to give me a list. `checklist` fulfills this requirement, but `type([1,2,3]) == list` instead of `checklist`. 

Does that make sense? I am finding it difficult to communicate my train of thought properly through text, so I'm happy to clarify as needed.

j9ac9k:
Hi @Zorkator 

After chatting w/ some of the other maintainers, we decided not to merge this PR. To expand the public API more, there would need to be a more significant benefit in my opinion.

Thanks for the PR regardless, sorry it's taken us so long to follow up.

PR #2451, opened 2022-10-02
Fix action parameter button that is briefly made visible before getting a parent

ntjess:
Before this change, spawning an `action` parameter briefly made its button visible before the button was assigned an invisible parent. This caused a brief flickering when showing a parameter tree for the first time:

https://user-images.githubusercontent.com/23620506/193437758-d245732d-8d16-431b-8467-b0ed7fa7fcbe.mp4


j9ac9k:
Thanks @ntjess LGTM!

PR #2453, opened 2022-10-02
Fix disconnect of signal proxy

dgoeries:
Back then with Python 2 support the SignalProxy disconnect was left broken behind.

**self.slot** stores a **weakref**, we disconnect properly now if needed.


j9ac9k:
Thanks for the PR! This LGMT.

PR #2461, opened 2022-10-03
Allow plotting multiple data items at once

ntjess:
Allow easily plotting multiple `PlotDataItem`s at once through a `PlotItem.multiDataPlot` interface. This way, some properties can be easily cycled for each curve without all the required manual specifications. See the newly added example for some usages.

### DIFF
--- a/pyqtgraph/examples/MultiDataPlot.py
+++ b/pyqtgraph/examples/MultiDataPlot.py
@@ -1,0 +1,83 @@
+import traceback
+import numpy as np
+
+import pyqtgraph as pg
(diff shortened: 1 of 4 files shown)


j9ac9k:
Hi @ntjess This LGTM.

On the example, I think we can do a better job showing what the data structure that's being passed in is:

So instead of:

image

first I was thinking `repr(x)` and `repr(y)` (instead of `xType` and `yType` but the output from that is also a bit tough to read....

j9ac9k:
Should have merged this a while back, merging now!  Thanks for the PR @ntjess

PR #2463, opened 2022-10-04
Add keyword argument in PColorMeshItem to enable consistent colormap scaling during animation

SimenZhor:
I've added a new keyword argument to PColorMeshItem that allows absolute scaling of the colormap (in addition to the autoscaling that is already present, and still default). This is especially useful when plotting animations where the absolute value of the z axis (color) matters. I've also modified the PColorMeshItem example to showcase why this feature might be important in certain cases.

I see that [ColorBarItem]( https://pyqtgraph.readthedocs.io/en/latest/api_reference/graphicsItems/colorbaritem.html) uses both `limits` and `levels` about the corresponding parameter, so I wasn't exactly sure what to call it, but I ended up with `limits` as I felt that was most self explanatory.

This is my first PR in this repo, so I might have missed some mandatory updates. In that case, please let me know, and I will try to amend the mistakes.

### DIFF
--- a/pyqtgraph/graphicsItems/PColorMeshItem.py
+++ b/pyqtgraph/graphicsItems/PColorMeshItem.py
@@ -92,7 +92,13 @@ class PColorMeshItem(GraphicsObject):
             Sets the minimum and maximum values to be represented by the colormap (min, max). 
             Values outside this range will be clipped to the colors representing min or max.
             ``None`` disables the limits, meaning that the colormap will autoscale 
-            each time ``setData()`` is called.
+            each time ``setData()`` is called - unless ``enableAutoLevels=False``.
+        enableAutoLevels: bool, optional, default True
+            Causes the colormap levels to autoscale whenever ``setData()`` is called. 
+            When enableAutoLevels is set to True, it is still possible to disable autoscaling
+            on a per-change-basis by using ``autoLevels=False`` when calling ``setData()``.
+            If ``enableAutoLevels==False`` and ``levels==None``, autoscaling will be 
+            performed once when the first z data is supplied. 
         edgecolors : dict, optional
             The color of the edges of the polygons.
             Default None means no edges.
@@ -113,6 +119,7 @@ class PColorMeshItem(GraphicsObject):
         self.edgecolors = kwargs.get('edgecolors', None)
         self.antialiasing = kwargs.get('antialiasing', False)
         self.levels = kwargs.get('levels', None)
+        self.enableAutoLevels = kwargs.get('enableAutoLevels', True)
         
         if 'colorMap' in kwargs:
             cmap = kwargs.get('colorMap')
@@ -184,7 +191,7 @@ class PColorMeshItem(GraphicsObject):
             ValueError('Data must been sent as (z) or (x, y, z)')
 
 
-    def setData(self, *args):
+    def setData(self, *args, **kwargs):
         """
         Set the data to be drawn.
 
@@ -206,7 +213,13 @@ class PColorMeshItem(GraphicsObject):
 
             "ASCII from: <https://matplotlib.org/3.2.1/api/_as_gen/
                          matplotlib.pyplot.pcolormesh.html>".
+        autoLevels: bool, optional, default True
+            When set to True, PColorMeshITem will automatically select levels
+            based on the minimum and maximum values encountered in the data along the z axis.
+            The minimum and maximum levels are mapped to the lowest and highest colors 
+            in the colormap. The autoLevels parameter is ignored if ``enableAutoLevels==False`` 
         """
+        autoLevels = kwargs.get('autoLevels', True)
 
         # Has the view bounds changed
         shapeChanged = False
@@ -239,10 +252,13 @@ class PColorMeshItem(GraphicsObject):
         lut = self.lut_qbrush
         # Second we associate each z value, that we normalize, to the lut
         scale = len(lut) - 1
-        if self.levels is None:
-            # Autoscale colormap each time setData is called
+        # Decide whether to autoscale the colormap or use the same levels as before
+        if (self.levels is None) or (self.enableAutoLevels == True and autoLevels == True):
+            # Autoscale colormap 
             z_min = self.z.min()
             z_max = self.z.max()
+            if not self.enableAutoLevels:
+                self.levels = (z_min, z_max)
         else:
             # Use consistent colormap scaling
             z_min = self.levels[0]
--- a/pyqtgraph/graphicsItems/PColorMeshItem.py
+++ b/pyqtgraph/graphicsItems/PColorMeshItem.py
@@ -88,6 +88,8 @@ class PColorMeshItem(GraphicsObject):
         colorMap : pyqtgraph.ColorMap
             Colormap used to map the z value to colors.
             default ``pyqtgraph.colormap.get('viridis')``
+        limits: tuple, optional, default None
+            Limits the colormap range to (low, high), None disables the limit.
         edgecolors : dict, optional
             The color of the edges of the polygons.
             Default None means no edges.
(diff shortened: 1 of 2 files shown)


NilsNemitz:
Hi @SimenZhor !

I am sure @j9ac9k will jump right in and have a closer look, but I wanted to comment on the `limits` / `levels` ambiguity:
`ColorBarItem` has an interactive level adjustment and IIRC the `limits` set the limits to this adjustment.

What you are looking for are the levels, as described in the `setLevels()` functionality of `ImageItem`. `ImageItem.setImage()` lets you set these through a keyword argument (...which is described in the docs for `setOpts()`...). That might be a good precedent, API-wise.

Do you think `PColorMeshItem.setData()` would benefit from also replicating the `autoLevels` parameter of `ImageItem.setImage()`? I believe that is a "keep what you have now" setting, and might avoid the need to specify the actual levels every time.


Otherwise this feature seems like a great addition, and indeed quite essential for some applications. Nice to see `PColorMeshItem` getting some attention!

SimenZhor:
Hi @NilsNemitz. Thanks for the encouraging response!

You are of course right about the kwarg naming. I'll make an update where I rename to `levels`. 

Rewriting `PColorMeshItem` to operate like `ImageItem` with `setOpts` seems like a bigger job that I'm not sure if I'm suited for (as mentioned, this is my first PR here, and I'm not super familiar with the APIs in general).

I've played around a bit with the `ImageItem` class in order to understand the `autoLevels` parameter, and it seems like `autoLevels` is just a slightly more verbose way of doing what I'm trying to achieve with this PR. Quick examples based on `pyqtgraph/examples/ImageItem.py`:
``` Python
# Ex1.1:
## Default behavior (autoscaling with each frame)
[...]
img = pg.ImageItem() # Setting levels here makes no difference to the end result in this example
[...]
def updateData():
    [...]
    img.setImage(data[i])

# Ex2.1:
## "Keep what you have" behavior
[...]
img = pg.ImageItem(levels=[1000,1500])
[...]
def updateData():
    [...]
    img.setImage(data[i], autoLevels=False) # autoLevels must be explicitly disabled for each frame
```
Corresponding examples for this PR:
``` Python 
# Ex1.2:
## Default behavior (autoscaling with each frame)
[...]
pcmi = pg.PColorMeshItem()
[...]
def updateData():
    [...]
    pcmi.setData(new_x, new_y, new_z)

# Ex2.2:
## "Set and forget" behavior
[...]
pcmi = pg.PColorMeshItem(levels=(-2,2))
[...]
def updateData():
    [...]
    pcmi.setData(new_x, new_y, new_z) # No need to explicitly disable autoscaling
```

If the autoLevels behavior is a well known and expected I'll gladly add it here, although I personally feel like it is slightly counter intuitive to continuously disable a setting.

SimenZhor:
I think your suggestion sounds like a good middle ground. Should we wait for others to weigh in, or should I suggest an implementation right away?

SimenZhor:
@NilsNemitz, I've just pushed a suggestion for what the implementation of this behavior may look like. Let me know if you have any further comments or suggestions 😄

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -200,7 +213,13 @@ def setData(self, *args):
 
             "ASCII from: <https://matplotlib.org/3.2.1/api/_as_gen/
                          matplotlib.pyplot.pcolormesh.html>".
+        autoLevels: bool, optional, default True
+            When set to True, PColorMeshITem will automatically select levels
```
j9ac9k, line 217: nit-pick typo for `PColorMeshITem`

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -200,7 +213,13 @@ def setData(self, *args):
...
+            based on the minimum and maximum values encountered in the data along the z axis.
+            The minimum and maximum levels are mapped to the lowest and highest colors 
+            in the colormap. The autoLevels parameter is ignored if ``enableAutoLevels==False``
```
j9ac9k, line 220: probably should change to

```
``enableAutoLevels is False``
```

last thing we should do is encourage doing equality operators on booleans :)  

(also a nit-pick level comment)

pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -233,8 +252,17 @@ def setData(self, *args):
         lut = self.lut_qbrush
         # Second we associate each z value, that we normalize, to the lut
         scale = len(lut) - 1
-        z_min = self.z.min()
-        z_max = self.z.max()
+        # Decide whether to autoscale the colormap or use the same levels as before
+        if (self.levels is None) or (self.enableAutoLevels == True and autoLevels == True):
```
j9ac9k, line 256: should likely change to 

```python
if (self.levels is None) or (self.enableAutoLevels and autoLevels):
```

j9ac9k (commented):
Sorry I've been quiet on this subject despite @NilsNemitz 's repeated poking.  This kind of API decision-making is something I'm absolutely rubbish at and I almost always defer to a consensus from other maintainers and contributors.

I have no issue with the proposed change.  I just have some nit-picky comments for the docstring.  

Thanks for the PR @SimenZhor !

pyqtgraph/graphicsItems/PColorMeshItem.py
SimenZhor: Good catch :) Fixed!

pyqtgraph/graphicsItems/PColorMeshItem.py
SimenZhor: Fixed

pyqtgraph/graphicsItems/PColorMeshItem.py
SimenZhor: Fixed

SimenZhor:
I made the improvements that were pointed out by @j9ac9k. Are there any next steps for me now, or just wait for more people to have the time to review my code? :)

j9ac9k:
Hi @SimenZhor I'm good w/ the PR as is, I'm going to leave it open in hopes to get the nod of agreement from @outofculture whose done a lot of the work with ImageItem if nobody brings up issues in a day or so, I'll merge 👍🏻 

Thanks for the submission!

outofculture (approved):
This code looks good. I would say that the only improvement I can think of would be to provide methods for setting the levels and auto-levels attributes. I would like to say this is a trivial thing to add, but part of the job would be refactoring `setData` to separate out the steps that belong in an `_updateDisplayWithCurrentState` method from the rest of it, such that you could elsewhere write e.g.:
```python
def setLevels(self, levels):
    self.levels = levels
    self._updateDisplayWithCurrentState()
```
If you feel up for this challenge, that would be a valuable improvement, but this PR can otherwise be accepted as-is.

SimenZhor:
@outofculture I agree. That makes total sense. I've already started work on a [new branch](https://github.com/SimenZhor/pyqtgraph/commit/5a2dfb5ac4f9bf683a6a89891b320f83c2e1860c) for a future PR where I'm planning to add ColorBar support as well to PColorMeshItem. In this usecase, the `_updateDisplayWithCurrentState()` method you describe has to be implemented (I've based my implementation on ImageItem, so the proof-of-concept commit I've made so far contains a lot of naming-bleed and completely unaltered docstrings - but you'll find the functionality you describe in `updateImage()` if you want to have a quick look).

I don't know if you prefer to wait for the work on that branch to be done or if we should just merge this one to keep the PRs shorter and cleaner. I suspect there will be a lot more discussion for getting the ColorBar support right. So, if possible, I'd prefer to do that in a different PR. I could of course include the setters in this PR, without doing full support for ColorBar. I'll leave that up to you.

j9ac9k:
I'm indifferent to that change being part of this PR vs a separate PR.  I'll likely squash the PR on merge unless the change is quite massive, and if which case it is, I may ask (and I may be willing to help with) rebasing such that there are only a few commits that contain well-contained changes.

SimenZhor:
Sounds good, @j9ac9k. I'd still prefer to keep them separate though. I've barely started looking at the new feature, and if the consensus is that this PR is ok as-is (even though it could be improved) I'm happy to continue working on the other stuff at my own pace.

j9ac9k:
Sounds good to me, I'll merge as is.  thanks for the PR @SimenZhor !

PR #2465, opened 2022-10-04
re-enable hmac authentication for win32

pijyoi:
The commit that disabled it was in November 2013, but didn't specify which version of Windows was having the issue.

This PR has been tested to be working on Windows 7, Python 3.8.10.


j9ac9k:
Thanks @pijyoi LGTM

PR #2466, opened 2022-10-05
Remove STRTransform main

Jaime02:
Why this file has a main function? It is not meant to be run independently, right? Shouldn't it be removed?

### DIFF
--- a/pyqtgraph/SRTTransform.py
+++ b/pyqtgraph/SRTTransform.py
@@ -75,7 +75,7 @@ class SRTTransform(QtGui.QTransform):
         self.update()
         
     def setFromMatrix4x4(self, m):
-        m = SRTTransform3D(m)
+        m = SRTTransform3D.SRTTransform3D(m)
         angle, axis = m.getRotation()
         if angle != 0 and (axis[0] != 0 or axis[1] != 0 or axis[2] != 1):
             print("angle: %s  axis: %s" % (str(angle), str(axis)))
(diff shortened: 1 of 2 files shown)


j9ac9k:
Looks like the `__main__` code there was added 12 years ago, we certainly don't use it, so I'm good with removing it.

That said, read-the-docs CI is failing on build and looks like it was complaining about a circular import; so I would suggest unwinding some of your changes.

EDIT: I see you closed the PR now; going to leave the comment just in case.

Jaime02:
@j9ac9k oh so it is completely pointless? I have seen the test failing due to the imports, but I thought that the main function could have something to do with that 🤓 Then shall I remove it and try to fix it?

j9ac9k:
Looking at the diff more closely, you have indeed introduced a circular import.  SRTTransform.py does imports SRTTransform3d, and SRTTransform3d now imports SRTTransform.py, which causes a circular import.

Jaime02:
> Looking at the diff more closely, you have indeed introduced a circular import. SRTTransform.py does imports SRTTransform3d, and SRTTransform3d now imports SRTTransform.py, which causes a circular import.

Yeah but I fixed that in the lastest patch. I used regular `import x; x.y` instead of `from x import y`

Jaime02:
Hey @j9ac9k can you check this again please?

j9ac9k:
I wouldn't say I like importing just the module instead of the class itself, I suppose I prefer to import the class at the bottom of the file, which feels like a rubbish way to go about handling the circular import.  Both these classes should likely reside in the same module, but that's _well_ outside the scope of this PR.

Thanks for the PR @Jaime02 Merging.

PR #2468, opened 2022-10-05
Remove all import *

Jaime02:
I have removed all the import * code, this makes all imports implicit.

### DIFF
--- a/pyqtgraph/SRTTransform.py
+++ b/pyqtgraph/SRTTransform.py
@@ -169,7 +169,7 @@ if __name__ == '__main__':
     import GraphicsView
 
     from . import widgets
-    from .functions import *
+    from .functions import mkPen
     app = pg.mkQApp()  # noqa: qapp stored to avoid gc
     win = QtWidgets.QMainWindow()
     win.show()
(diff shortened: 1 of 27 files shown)


j9ac9k:
Hi @Jaime02 

Thanks for the PR.  The openGL related changes may need to be undone, those imports were very sensitive (hence the #noqa next to them).

Jaime02:
@j9ac9k why? Is there a specific reason for importing all OpenGL stuff?

j9ac9k:
> @j9ac9k why? Is there a specific reason for importing all OpenGL stuff?

Hi @Jaime02 

The import order is important here as some of the opengl functions/objects are intended to get clobbered/overwritten through the import sequence.  I learned this the hard way when I ran `isort` through the codebase and ran into a whole array of issues.

With the change you have proposed, the next time isort is run, all the opengl code will not work. 

While other maintainers may feel differently, I'm of the opinion that any import statement with `#noqa` should be left as they currently are.

j9ac9k:
thanks @Jaime02 this LGTM, merging!

PR #2469, opened 2022-10-05
Remove HeightColorShader class

Jaime02:
This class is not used anywhere, seems to be unfinished and contains undefined variables

Shall I keep doing PRs like this? I mean, small easy to merge PRs rather than multiple changes, right?


j9ac9k:
Ha, that's a good tip on it being broken, I would definitely accept a PR that fixed it 😆

Issue #2123, opened 2021-11-26
mouseHoverEvent method of ButtonItem class not work


hatawong:
### `def mouseHoverEvent(self, ev) ` of ButtonItem class not work
right method should be  `def hoverEvent(self, ev) `

j9ac9k:
Hi @hatawong can you provide a code example showing it not working?

hatawong:
```
import pyqtgraph as pg
from pyqtgraph import icons


class PlotItem(pg.PlotItem):

    def __init__(self, **kargs):
        super().__init__(**kargs)
        self.autoBtn.setParentItem(None)
        self.autoBtn = ButtonItem(imageFile=icons.getGraphPixmap('auto'), width=14, parentItem=self)
        self.autoBtn.mode = 'show'
        self.autoBtn.clicked.connect(self.autoBtnClicked)


class ButtonItem(pg.ButtonItem):

    def __init__(self, **kargs):
        super().__init__(**kargs)

    def mouseHoverEvent(self, ev):  # original implements - not work
        print("mouseHoverEvent")
        super(ButtonItem, self).mouseHoverEvent(ev)

    def hoverEvent(self, ev):       # fix implements - it work
        print("hoverEvent")
        if not self.enabled:
            return
        if ev.enter:
            self.setOpacity(1.0)
        if ev.exit:
            self.setOpacity(0.7)


win = pg.GraphicsLayoutWidget(show=True, title="Button Hover Example")
win.resize(1000, 600)

pg.setConfigOptions(antialias=True)

win.addItem(PlotItem())

if __name__ == '__main__':
    pg.exec()

```

hatawong:
when mouse hover in to the auto button, output:
```
python.exe button_examplt.py
hoverEvent
hoverEvent
hoverEvent
hoverEvent
...
```

hatawong:
call stack
```
hoverEvent, button.py:25
sendHoverEvents, GraphicsScene.py:285
mouseMoveEvent, GraphicsScene.py:196
mouseMoveEvent, GraphicsView.py:364
exec_, __init__.py:407
<module>, button.py:42
```

hatawong:
GraphicsScene.py:285
```
    def sendHoverEvents(self, ev, exitOnly=False):
        ## if exitOnly, then just inform all previously hovered items that the mouse has left.
        
        if exitOnly:
            acceptable=False
            items = []
            event = HoverEvent(None, acceptable)
        else:
[...]

PR #2473, opened 2022-10-07
Fix ButtonItem hover event

bbc131:
* Fix method name `mouseHoverEvent` -> `hoverEvent`. Therefore, it has never been called.
* Fix changing opacity while hovering. Change it back to "non-hover-state" only after hover event with "is-exit" occurs. (Note, that `hoverEvent` is called multiple times while mouse cursor is on the item.)

Simply have a look at the "Auto-Range-Button" in any plot to see the hovering work now.


j9ac9k:
Fixes #2123

j9ac9k:
Thanks for the PR @bbc131 and thanks for the report @hatawong  This LGTM, merging.

PR #2474, opened 2022-10-07
Raise appropriate Exceptions in place of generic exceptions

### LINKED ISSUE #2167: PlotDataItem.setData() raises generic Exception instead of TypeError.
When using setData() method in the context of a try: except statement (plenty possible application since that's the low level way of redrawing fast) one has to resort to a generic Exception catch which looks nooby and can lead to undetected bugs (just what happened to me). setData() should probably be raising more specific exception types, judging by the Exception messages in place these should probably both be TypeErrors.

### Short description
PlotDataItem.setData() raises generic Exception instead of TypeError.

### Code to reproduce 1
```
import pyqtgraph as pg
plotWidget = pg.plot()
plot_data_item = plotWidget.plot()
plot_data_item.setData(None)
```

### Expected behavior
```
TypeError: Invalid data type <class 'NoneType'>
```

### Real behavior
```
Exception: Invalid data type <class 'NoneType'>
```

### Code to reproduce 2
```
import pyqtgraph as pg
plotWidget = pg.plot()
plot_data_item = plotWidget.plot()
plot_data_item.setData(None,None)
```

### Expected behavior
```
TypeError: When passing two unnamed arguments, both must be a list or array of values. (got <class 'NoneType'>, <class 'NoneType'>)
```

### Real behavior
```
Exception: When passing two unnamed arguments, both must be a list or array of values. (got <class 'NoneType'>, <class 'NoneType'>)
```

### Tested environment(s)
  - python=3.7.6=cpython_h60c2a47_6
  - pyqtgraph=0.11.0=pyh9f0ad1d_0 (this is certainly not the latest but the raise Except statements are there in master still)

PS: There is another 'raise Exception' in PlotDataItem.py relating to wrong dimensioning on numpy arrays. Here some specific class ArrayDimensionError(Exception) could be created or perhaps numpy readily implements some applicable exception type.

Nibba2018:
Fixes #2167 
Raise `TypeError` in place of generic `Exception` for `PlotDataItem.setData()`.
Introduced and raised a new exception named `ArrayDimensionError` when numpy arrays have an Invalid shape.

### Other Tasks 

<details>
  <summary>Raise `TypeError` in place of generic `Exception` for `PlotDataItem.setData()`</summary>

### Files that need updates
    
- [X] `README.md`
- [X] `setup.py`
- [X] `tox.ini`
- [X] `.github/workflows/main.yml` and associated `requirements.txt` and conda `environemt.yml` files
- [X] `pyproject.toml`
- [X] `binder/requirements.txt`

</details>

<details>

### Pre Release Checklist

</details>

<details>

### Steps To Complete

- [ ] Announce on mail list
- [ ] Announce on Twitter

</details>

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -1214,7 +1214,7 @@ def dataType(obj):
             elif obj.ndim == 2 and obj.dtype.names is None and obj.shape[1] == 2:
                 return 'Nx2array'
             else:
-                raise TypeError('array shape must be (N,) or (N,2); got %s instead' % str(obj.shape))
+                raise ValueError('array shape must be (N,) or (N,2); got %s instead' % str(obj.shape))
         elif isinstance(first, dict):
             return 'listOfDicts'
         else:
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -7,6 +7,7 @@ from .. import debug as debug
 from .. import functions as fn
 from .. import getConfigOption
 from ..Qt import QtCore
+from ..exceptionHandling import ArrayDimensionError
 from .GraphicsObject import GraphicsObject
 from .PlotCurveItem import PlotCurveItem
 from .ScatterPlotItem import ScatterPlotItem
@@ -1214,7 +1215,7 @@ def dataType(obj):
             elif obj.ndim == 2 and obj.dtype.names is None and obj.shape[1] == 2:
                 return 'Nx2array'
             else:
-                raise Exception('array shape must be (N,) or (N,2); got %s instead' % str(obj.shape))
+                raise ArrayDimensionError('array shape must be (N,) or (N,2); got %s instead' % str(obj.shape))
         elif isinstance(first, dict):
             return 'listOfDicts'
         else:
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -707,7 +707,7 @@ class PlotDataItem(GraphicsObject):
                     x = np.array(data['x'])
                 if 'y' in data:
                     y = np.array(data['y'])
-            elif dt ==  'listOfDicts':
+            elif dt == 'listOfDicts':
                 if 'x' in data[0]:
                     x = np.array([d.get('x',None) for d in data])
                 if 'y' in data[0]:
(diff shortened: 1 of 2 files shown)


pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -1214,7 +1215,7 @@ def dataType(obj):
             elif obj.ndim == 2 and obj.dtype.names is None and obj.shape[1] == 2:
                 return 'Nx2array'
             else:
-                raise Exception('array shape must be (N,) or (N,2); got %s instead' % str(obj.shape))
+                raise ArrayDimensionError('array shape must be (N,) or (N,2); got %s instead' % str(obj.shape))
```
j9ac9k, line 1218: I do think it's a good idea to introduce more specific exceptions for this sort of thing, but I think for this purpose, we should probably stick with `ValueError` which I believe is what numpy raises when you get a dimension mismatch related exceptions.

j9ac9k:
Hi @Nibba2018 

Thanks for the PR! 

I made a comment, while I think it's a good idea to have more specific exceptions, this is something that users of pyqtgraph aren't expecting, so it would create some problems for folks trying to catch this type of exception (they wouldn't know where to import it from).

I think what would be best to stick w/ numpy's convention, where I believe they raise a `ValueError` for this type of issue.

pyqtgraph/graphicsItems/PlotDataItem.py
Nibba2018: Thank you for the review @j9ac9k , will update it to `ValueError`.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -1214,7 +1214,7 @@ def dataType(obj):
             elif obj.ndim == 2 and obj.dtype.names is None and obj.shape[1] == 2:
                 return 'Nx2array'
             else:
-                raise Exception('array shape must be (N,) or (N,2); got %s instead' % str(obj.shape))
+                raise TypeError('array shape must be (N,) or (N,2); got %s instead' % str(obj.shape))
```
j9ac9k, line 1217: Hi, this one should be a `ValueError` not a `TypeError` 👍🏻

j9ac9k:
diff doesn't match the comment/message, looks like it was replaced with a TypeError ;)

pyqtgraph/graphicsItems/PlotDataItem.py
Nibba2018: OMG sorry about that :rofl:

j9ac9k:
LGTM thanks for the contribution @Nibba2018 !

PR #2476, opened 2022-10-09
ROI: Add test with mouseDrag event and check snapping

dgoeries:
By working on the **scaleSnap (x, y)** feature, i realized that the changes will become noisy.

I do a pre step here by flaking the test region and adding a simple test with a rect roi for snapping. No logical changes are involved.


j9ac9k:
Hi @dgoeries this LGTM, but I'll defer to @outofculture to merge.  Thanks for the PR!

PR #2485, opened 2022-10-11
ROI: Provide feature for snapSize (x,y)

dgoeries:
This is the next PR after #2476

We now allow to declare the snapSize of the ROI as sequence of length 2 to have individual snapping for x and y dimension.

I extended the previously implemented tests. By doing so, I realized that flooring will provide a better snapping than rounding.

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -13,6 +13,7 @@ of how to build an ROI at the bottom of the file.
 
 import sys
 from math import atan2, cos, degrees, hypot, sin
+import warnings
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


j9ac9k:
CI says no due to codescan issue from before, but as the diff involves a line it already had an issue with, it's returning an error.  I'll let @outofculture decide on merging.

Thanks for the PR @dgoeries !

ixjlyons:
Maybe this was discussed elsewhere already, but could we avoid the deprecation by accepting either a single scalar or a sequence? It's a minor convenience if you don't care about independent x/y, but no breaking change. I suppose setting `roi.scaleSnap = 2` and getting back `roi.scaleSnap == (2, 2)` might be a little awkward and could break stuff anyway.

dgoeries:
@ixjlyons Yes, we can also provide both ways for an amount of time. Looking forward for @outofculture opinion ...

dgoeries:
@ixjlyons Yes, we can also provide both ways for an amount of time. Looking forward for @outofculture opinion ...

outofculture:
So sorry for the long delay on this. This feature is a great addition.

Yes, absolutely, this will need to have a deprecation cycle if we don't also build in the ability to use the old interface. I think it's reasonable to do both, though. All the old tests (thanks again for those) will need to be copied to ensure that old behavior continues, too. Are you able to put this change together? If not, I can try to take it on.

PR #2487, opened 2022-10-12
PlotSpeedTest: reflect initial use_opengl state

pijyoi:
Since #2318, when enabling OpenGL from the command line, the associated checkboxes were not reflecting the initial state.
Also, toggling OpenGL would have no effect as `GraphicsView::useOpenGL()` needs to be called.

### DIFF
--- a/pyqtgraph/examples/PlotSpeedTest.py
+++ b/pyqtgraph/examples/PlotSpeedTest.py
@@ -35,6 +35,7 @@ args = parser.parse_args()
 if args.use_opengl is not None:
     pg.setConfigOption('useOpenGL', args.use_opengl)
     pg.setConfigOption('enableExperimental', args.use_opengl)
+use_opengl = pg.getConfigOption('useOpenGL')
 
 # don't limit frame rate to vsync
 sfmt = QtGui.QSurfaceFormat()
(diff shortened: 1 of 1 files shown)


j9ac9k:
Gonna ask @ntjess to make sure this is setup as it's supposed to.

ntjess:
> Since https://github.com/pyqtgraph/pyqtgraph/pull/2318, when enabling OpenGL from the command line, the associated checkboxes were not reflecting the initial state.

Would it work to simply address `useOpenGL` from `updateOptions`? This would be the diff:
```diff
@@ -35,6 +35,7 @@ args = parser.parse_args()
 if args.use_opengl is not None:
     pg.setConfigOption('useOpenGL', args.use_opengl)
     pg.setConfigOption('enableExperimental', args.use_opengl)
+use_opengl = pg.getConfigOption('useOpenGL')
 
 # don't limit frame rate to vsync
 sfmt = QtGui.QSurfaceFormat()
@@ -151,11 +152,11 @@ def updateOptions(
     curvePen=pg.mkPen(),
     plotMethod='pyqtgraph',
     fillLevel=False,
-    enableExperimental=False,
-    useOpenGL=False,
+    enableExperimental=use_opengl,
+    useOpenGL=use_opengl,
 ):
     pg.setConfigOption('enableExperimental', enableExperimental)
-    pg.setConfigOption('useOpenGL', useOpenGL)
+    pw.useOpenGL(useOpenGL)
     curve.setPen(curvePen)
     curve.setFillLevel(0.0 if fillLevel else None)
     curve.setMethod(plotMethod)

```

pijyoi:
I tried out your version and it seems like it's not such a big penalty to reset the `QGraphicsView` viewport when changing other unrelated options.
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/widgets/GraphicsView.py#L153

I have changed the PR to use your version instead.

ntjess:
👍 LGTM. I might make a separate PR to determine if the current viewport already matches `useOpenGL`, but that change would be out of scope for this PR

j9ac9k:
Thanks for the PR @pijyoi and thanks for the review @ntjess LGTM, merging!

PR #2488, opened 2022-10-12
Update readme links

ntjess:
- [x] Update "s3a application" link
- [x]  Update binder for jupyter examples to open in tabs rather than pages
  - Thanks to SO user [Wayne](https://stackoverflow.com/users/8508004/wayne) for the recommendation


j9ac9k:
Thanks @ntjess !

Issue #2328, opened 2022-06-15
Auto-range breaks with downsampling enabled and very small data 


agausmann:
### Short description

When the data to be downsampled is very small relative to the size of the current viewport, downsampling produces an empty data array. That downsampled array is also used when computing `PlotDataItem.dataBounds` (since it defers to `self.curve.dataBounds` or `self.scatter.dataBounds`, which uses the processed data given to those items), and that method is used by the ViewBox when computing its auto-range. However, since it finds an empty array, it won't report any bounds, and the plot will be ignored when auto-ranging.

### Code to reproduce

```python
import sys
from pyqtgraph import PlotWidget, PlotDataItem
from PyQt5.QtWidgets import QApplication

app = QApplication(sys.argv)
plot = PlotWidget()

data = PlotDataItem()
plot.addItem(data)
data.setDownsampling(auto=True)

# OK:
# data.setData(
#     x=[0.0, 1.0],
#     y=[0.0, 1.0],
# )

# BAD:
data.setData(
    x=[0.0, 1e-6],
    y=[0.0, 1e-6],
)

plot.show()
app.exec_()
```

### Expected behavior

If you zoom in enough around the origin (until the downsampling produces a non-empty array), and _then_ enable auto-range, then it works fine:

![image](https://user-images.githubusercontent.com/6611767/173872277-6af0364b-e175-4bfe-9922-a2dc85cbfdd9.png)

### Real Behavior

![image](https://user-images.githubusercontent.com/6611767/173872166-dd49e041-b1bb-41f4-8f2e-fccbb96ed545.png)

### Tested environment(s)

 * PyQtGraph version:  0.12.4
 * Qt Python binding: PyQt5 5.15.6 Qt 5.15.2
 * Python version: 3.9.9 and 3.10.1
 * NumPy version: 1.22.4
 * Operating system: Windows 10
 * Installation method: pip

agausmann:
Using downsampled data for bounds and auto-range calculation is dubious in general, because it's a cyclic dependency. In another example case, I've observed a feedback loop with log plots where the scale updates which triggers resampling which changes the data which changes the scale, and this cycle causes the auto-range to oscillate and never stabilize. And even where it does stabilize, the autorange and resampling is probably being computed several times...

Issue #2206, opened 2022-02-23
Histogram Autolevels does not work for constant arrays


kremeyer:
### Short description
When continuously plotting images that is constant (all values are the same) to an ImageView widget with the automatic levels for the histogram enables, the axis range will keep extending infinitely in both directions around that constant.

### Code to reproduce
In this example you can just click the `Norm` button over and over again to see the behavior.

```python
import pyqtgraph as pg
import numpy as np
from pyqtgraph.Qt import QtWidgets

app = pg.mkQApp('Auto Level Bug')
win = QtWidgets.QMainWindow()
win.resize(800,800)
imv = pg.ImageView()
win.setCentralWidget(imv)
win.show()

def new_img(self):
	imv.setImage(np.zeros((100, 100)))
	print('new img')
imv.__dict__['new_img'] = new_img
imv.ui.menuBtn.clicked.connect(new_img)

if __name__ == '__main__':
    pg.exec()
```

### Expected behavior
Axis bounds should not change.

### Real behavior
Auto range keeps extending axis limits of histogram towards `(-inf,inf)` with every consecutive image drawn

### Tested environment(s)

 * PyQtGraph version: 0.12.3
 * Qt Python binding: PyQt5 5.15.4 Qt 5.15.2
 * Python version: 3.8.10
 * NumPy version: 1.21.2
 * Operating system: Ubuntu & Windows 10
 * Installation method: pip

lyragordon:
Having a similar problem. With any data outside (0,1), it will just peak the colormap. Reproduced with a minor change to @kremeyer 's code (Norm button produces the behavior I'd expect from autolevel):
```
import pyqtgraph as pg
import numpy as np
from pyqtgraph.Qt import QtWidgets

app = pg.mkQApp('Auto Level Bug')
win = QtWidgets.QMainWindow()
win.resize(800, 800)
imv = pg.ImageView()
win.setCentralWidget(imv)
win.show()


def new_img(self):
    imv.setImage(np.random.rand(100, 100) * 10, autoLevels=True)
    print('new img')


imv.__dict__['new_img'] = new_img
imv.ui.menuBtn.clicked.connect(new_img)

if __name__ == '__main__':
    pg.exec()

```

j9ac9k:
Fixed by merging #2489

PR #2489, opened 2022-10-13
Fix `ViewBox.autoRange()` for case of data with constant y-value

bbc131:
Calling `ViewBox.autoRange()` multiple times lead to different results if the plotted data points have all identical y-values. For example, when you plot an array like [1.0, 1.0, ... 1.0], as I do in the mwe below.

The reason for that can be found in `ViewBox.setRange(rect, padding, ...)`, where for this special case of 0 extent of the vertical axis, the current visible range shall be preserved, see also comment "If we requested 0 range, try to preserve previous scale." in line 594.
But with `padding` being not equal to 0.0, the current visible range is not preserved, but rather enlarged by the padding.

The only way I see to circumvent this is to add some other special handling here.

Here a mwe which simply plots some data as described above und further calls `ViewBox.autoRange()` when you click into the widget.

```
import pyqtgraph as pg
from PySide2 import QtCore


app = pg.mkQApp()

class ClickablePlotWidget(pg.PlotWidget):
    clicked: QtCore.Signal = QtCore.Signal()

    def mousePressEvent(self, ev):
        self.clicked.emit()
        super().mousePressEvent(ev)


p = ClickablePlotWidget()
p.show()

p.plot([1.0, 1.0, 1.0])  # bad
# p.plot([1.0, 10.0, 1.0])  # good

def _autorange():
    print("ViewBox.autoRange()")
    vb = p.plotItem.getViewBox()
    vb.autoRange()
    # vb.autoRange(padding=0.0)  # workaround

p.clicked.connect(_autorange)


if __name__ == '__main__':
    pg.exec()
```

### DIFF
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -593,7 +593,9 @@ class ViewBox(GraphicsWidget):
 
             # If we requested 0 range, try to preserve previous scale.
             # Otherwise just pick an arbitrary scale.
+            preserve = False
             if mn == mx:
+                preserve = True
                 dy = self.state['viewRange'][ax][1] - self.state['viewRange'][ax][0]
                 if dy == 0:
                     dy = 1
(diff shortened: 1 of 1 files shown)


j9ac9k:
Hi @bbc131 Thanks for the PR, sorry it's been a while before someone replied, I was on vacation.  The diff LGTM, do you know if this fixes any of the existing open issues involving autoRange such as #2328 ?

bbc131:
> do you know if this fixes any of the existing open issues involving autoRange such as https://github.com/pyqtgraph/pyqtgraph/issues/2328 ?

#2328 is a different problem and won't be fixed.
#2206 would be fixed.

j9ac9k:
Thanks for the PR @bbc131 this LGTM, merging!

PR #2492, opened 2022-10-15
fix: ndarray_from_qimage does not hold a reference to qimage

pijyoi:
Note: `ndarray_from_qimage` is a library internal function.

It is the user's responsibility to ensure that the life-cycle of the `QImage` exceeds that of the created `ndarray` view.
A search of the usages of `ndarray_from_qimage` within the library will show safe ways to use it.


ntjess:
Thanks for the clear explanation and fix. LGTM

PR #2493, opened 2022-10-16
Added 'value' argument to ActionParameterItem that is called on click event

### LINKED ISSUE #1421: Enhancement: function in `opts` for ActionParameter
It would be great if, when creating an `action` parameter using a dict, the function associated with this action could be triggered without having to explicitly create a parameter and bind the function to the `sigActivated` signal. Since `value` is currently unused, it seems like a great place to allow a function that can be activated.

This should be quite straightforward -- just a few extra lines in `ActionParameterItem`:

```python
# ...
self.button  = QtGui.QPushButton(title)
activateFn = param.opts.get('value', None) # Or whatever name should be assigned to this option
self.button.clicked.connect(activateFn)
```

Sample usage:
```python
action = dict(name='test action, type='action', value=lambda: QtWidgets.QMessageBox.information('hello world!'))
```

kmfahey:
Altered pyqtgraph.parametertree.parameterTypes.action.ActionParameterItem to accept a 'value' argument of a zero-argument function and assign it to self.button.clicked. The function is called on the ParameterControlledButton's click event.

This is a more streamlined style of constructing a button with a click event than the previous behavior, which called for defining a function separate from the Parameter group constructor and then manually binding it to the sigActivated event.

Fixes #1421 

### Other Tasks 

<details>

### Files that need updates
    
- [ ] `README.md`
- [ ] `setup.py`
- [ ] `tox.ini`
- [ ] `pyproject.toml`

</details>

<details>

### Pre Release Checklist

</details>

<details>

### Steps To Complete

- [ ] Announce on mail list
- [ ] Announce on Twitter

</details>

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes/action.py
+++ b/pyqtgraph/parametertree/parameterTypes/action.py
@@ -45,7 +45,7 @@ class ParameterControlledButton(QtWidgets.QPushButton):
 
 
 class ActionParameterItem(ParameterItem):
-    """ParameterItem displaying a clickable button."""
+    """ParameterItem displaying a clickable button. Accepts a 'value' parameter of a zero-argument function which is called on click event."""
     def __init__(self, param, depth):
         ParameterItem.__init__(self, param, depth)
         self.layoutWidget = QtWidgets.QWidget()
--- a/pyqtgraph/parametertree/parameterTypes/action.py
+++ b/pyqtgraph/parametertree/parameterTypes/action.py
@@ -57,6 +57,9 @@ class ActionParameterItem(ParameterItem):
         self.layout.addWidget(self.button)
         self.layout.addStretch()
         self.titleChanged()
+        activateFn = param.opts.get('value', None)
+        if activateFn is not None:
+            self.button.clicked.connect(activateFn)
 
     def treeWidgetChanged(self):
         ParameterItem.treeWidgetChanged(self)


ntjess:
@kmfahey Congrats on your first pyqtgraph PR! A few extra things that I've considered since creating the initial issue:
* This code should probably live in `ActionParameter.setValue` instead of `ActionParameterItem.__init__`, otherwise the function will not be called during `ActionParameter.activate()`.
* Since `button.clicked` is not visible from `ActionParameter`, it will probably be best to tie `activateFn` to `self.sigActivated`
* Old connected functions should be disconnected when the value changes, i.e. we will need to call `pyqtgraph.function.disconnect(self.sigActivated, self.opts["value"])` before connecting the new signal.

Let me know if any of these items are confusing and I'll be happy to explain further.

kmfahey:
I'm pretty sure I understand, but for one question: can you give me an example of when ActionParameter.activate normally gets called? (A simple source tree grep didn't turn up examples.) I wrote a test script that manually called it just to confirm the set value executes, but it would be good to understand the changes better.

ntjess:
Sure, it is the function called when an action parameter's button is [clicked](https://github.com/pyqtgraph/pyqtgraph/blob/edc7dc710df0149ab9daefd8dd5a345485aaac47/pyqtgraph/parametertree/parameterTypes/action.py#L17). An important difference between calling `button.click()` directly and calling `activate()` is that `activate()`also calls `self.emitStateChanged('activated', None)` which triggers listeners to [`treeStateChanged`](https://github.com/pyqtgraph/pyqtgraph/blob/edc7dc710df0149ab9daefd8dd5a345485aaac47/pyqtgraph/parametertree/Parameter.py#L69).

ntjess:
Closing since it looks like this discussion resulted in #2496

PR #2496, opened 2022-10-19
ActionParameter.setValue links a function value to sigActivated

### LINKED ISSUE #1421: Enhancement: function in `opts` for ActionParameter

kmfahey:
When instantiating an ActionParameter directly or via a dict() of arguments passed to a group Parameter, if the `value` argument is callable, it is connected to self.sigActivate via an overriden setValue() method. If there was a previous value for `value` that was callable, it is disconnected.

Fixes #1421 

### Other Tasks 

<details>

### Files that need updates
    
- [ ] `README.md`
- [ ] `setup.py`
- [ ] `tox.ini`
- [ ] `pyproject.toml`

</details>

<details>

### Pre Release Checklist

</details>

<details>

### Steps To Complete

- [ ] Announce on mail list
- [ ] Announce on Twitter

</details>


ntjess:
The logic looks good to me, thanks! Are you able to add a test as well to `tests/parametertree/test_parametertypes.py`? Something that checks if disconnection logic and connection logic works properly. I can offer more input if you need

pyqtgraph/parametertree/parameterTypes/action.py
```diff
@@ -94,3 +96,13 @@ class ActionParameter(Parameter):
     def activate(self):
         self.sigActivated.emit(self)
         self.emitStateChanged('activated', None)
+
+    def setValue(self, value, blocksignal=None):
```
ntjess, line 100: It would be good to document this new feature under the `Parameters` docstring block above. After `shortcut: str`, you can add a description of how `value` should be used.

pyqtgraph/parametertree/parameterTypes/action.py
```diff
@@ -1,8 +1,10 @@
 from ...Qt import QtCore, QtWidgets, QtGui
+from ...functions import disconnect
 from ..Parameter import Parameter
 from ..ParameterItem import ParameterItem
 
 
+
```
ntjess, line 7: Very minor: It would be best to remove this blank space to maintain pep-8 guidelines

pyqtgraph/parametertree/parameterTypes/action.py
```diff
@@ -1,8 +1,10 @@
 from ...Qt import QtCore, QtWidgets, QtGui
+from ...functions import disconnect
```
ntjess, line 2: @j9ac9k do we have `isort` guidelines for import ordering? I.e. case-sensitivity and all that? I figure it's out of scope for this PR (lots of noncompliant files throughout the lib), but something to consider for future contributions

j9ac9k: We do... but they're not enforced (and I don't plan on enforcing them for now); I ran isort on the entire repo a while back to mostly get it in good shape, identify bugs, and pain points and address them accordingly.  I have isort as part of my pre-commit hooks, but we don't mandate contributors use pre-commit yet.  I figure we'll mandate this once we can run black/pycln/isort when the PR queue is non-existent and in turn enable pre-commit.ci

pyqtgraph/parametertree/parameterTypes/action.py
kmfahey: Finally getting back to this, sorry for the delay.

Pertinent question: is it appropriate to set the value to self.opts["value"]? I assume so, because that's the behavior of Parameter.setValue() that I'm overriding.

But I have an example script I'm using to check my work and if I have setValue store the value to self.opts the callable fires not once but twice-- ie. I get a popup when I click the button, and when I click OK & it closes I get another identical popup. Example script is attached for context.

[actionParameter_example_py.txt](https://github.com/pyqtgraph/pyqtgraph/files/9966770/actionParameter_example_py.txt)

ntjess: No worries, I haven't looked at it in a minute either. I think this logic might work a bit better:
```python
    def setValue(self, value, blockSignal=None):
        oldValue = self.value()
        if callable(oldValue) and oldValue is not value:
            disconnect(self.sigActivated, oldValue)
        if callable(value) and value is not oldValue:
            self.sigActivated.connect(value)
        return super().setValue(value, blockSignal)
```

This way, we always call `super().setValue()` and disconnect callable old values even when the new value isn't callable

Thanks for your example script! I tried it with this new implementation and it seemed to work OK: let me know what you think

magentatobe: Well, I ran into an issue. I wanted to see your setValue() in action so I pasted it in, saved, then ran that example script using the python debugger. (action.py:76 is the first line of the method if it's pasted in at the end of the ActionParameterItem class.)

`kmfahey@Madeline:~/Workspace/pyqtgraph$ python -m pdb actionParameter_example.py 
> /home/kmfahey/Workspace/pyqtgraph/actionParameter_example.py(1)<module>()
-> import PyQt5
(Pdb) b /home/kmfahey/Workspace/pyqtgraph/pyqtgraph/parametertree/parameterTypes/action.py:76
Breakpoint 1 at /home/kmfahey/Workspace/pyqtgraph/pyqtgraph/parametertree/parameterTypes/action.py:76
(Pdb) c
qt.qpa.xcb: QXcbConnection: XCB error: 3 (BadWindow), sequence: 595, resource id: 24609331, major code: 40 (TranslateCoords), minor code: 0
The program finished and will be restarted
> /home/kmfahey/Workspace/pyqtgraph/actionParameter_example.py(1)<module>()
-> import PyQt5
(Pdb)`

The new setValue() was never called (assuming I pasted it in the right place). Did I err or is this issue resolved by some other code?

ntjess: Sorry for the long delay...

I'm a bit confused -- can you clarify what you mean by "the new setValue() was never called"?

Here's a different script that might better illustrate the code is working:
```python
import PyQt5
import pyqtgraph as pg
from pyqtgraph.parametertree import Parameter, ParameterTree, parameterTypes as ptypes


def a(x=5, y=6):
    QtWidgets.QMessageBox.information(None, "Hello World", f"X is {x}, Y is {y}")


params = Parameter.create(
    name='"a" parameters',
    type="group",
    children=[
        dict(name="x", type="int", value=5),
        dict(name="y", type="int", value=6),
        dict(
            name="Hello, world",
            type="action",
            value=lambda *args: pg.Qt.QtWidgets.QMessageBox.information(
                None, "hello world!", "hello world!"
            ),
        ),
    ],
)

def onChange(_param, _value):
   print(f"function is {_value}")

ap = params.child("Hello, world")
ap.sigValueChanged.connect(onChange)

ap.setValue(a)

app = pg.mkQApp()
tree = ParameterTree()
tree.setParameters(params)
tree.show()
pg.exec()


# ap.activate()

```

PR #2502, opened 2022-10-26
Fix QMenu leak in PlotItem and ViewBox

### LINKED ISSUE #2497: `QMenu`s not destroyed (memory leak)

```
import pyqtgraph as pg

app = pg.mkQApp()
widget.show()

N = 10
for i in range(N):
    plot = pg.PlotItem()
    widget.addItem(plot)
    app.processEvents()
    widget.removeItem(plot)

if __name__ == '__main__':
    pg.exec()
```

pijyoi:
Fixes #2497

`QMenu`s are instantiated and then added to another menu as a sub-menu using `QMenu.addMenu(menu : QMenu)`.
However, this method specifically does not take ownership of the sub-menu. (https://doc.qt.io/qt-6/qmenu.html#addMenu)
Hence the sub-menus remain parent-less.

On PySide{2,6} bindings, these parent-less sub-menus get leaked, as demonstrated by the scripts in #2497.
Setting a parent to them fixes the leak.

However, these changes cause a failure in `test_ref_cycles.py::test_PlotWidget()` for PyQt{5,6} bindings.
Specifically, with this change, the objects in `w.plotItem.getMenu()` already have their underlying C++ object destroyed, resulting in the following error.
```
>           assert ref() is None
E           AssertionError: assert <PyQt6.QtWidgets.QCheckBox object at 0x000001BC7D7EC310> is None
E            +  where <PyQt6.QtWidgets.QCheckBox object at 0x000001BC7D7EC310> = <[RuntimeError('wrapped C/C++ object of type QCheckBox has been deleted') raised in repr()] weakref object at 0x1bc7d804c20>()
```
A possible workaround (done in this PR) is to treat such occurrences as a pass.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -201,7 +201,7 @@ class PlotItem(GraphicsWidget):
         self.ctrlMenu.setTitle(translate("PlotItem", 'Plot Options'))
         self.subMenus = []
         for name, grp in menuItems:
-            sm = QtWidgets.QMenu(name)
+            sm = QtWidgets.QMenu(name, self.ctrlMenu)
             act = QtWidgets.QWidgetAction(self)
             act.setDefaultWidget(grp)
             sm.addAction(act)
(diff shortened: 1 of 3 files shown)


j9ac9k:
Thanks for the PR @pijyoi  This LGTM, merging.

PR #2503, opened 2022-10-26
fix wrong logic for assert_alldead

pijyoi:
#2502 introduced a bug when relaxing the test criteria for pass.
It would have passed, for example, numpy arrays that were still alive.


j9ac9k:
I should have caught that as well, sorry about that!  Thanks for the prompt follow up @pijyoi

j9ac9k:
> The fault is entirely mine...

"entirely":  not even remotely true.  You didn't press the green merge button anyway not like I haven't merged my own bugs into the repo before 😬

PR #2505, opened 2022-10-27
Make `ActionGroup` compatible with existing Parameter conventions

ntjess:
- The name didn't end with `Parameter` which is unlike every other class
- `sigActivated` should emit `self` instead of nothing
- `emitStateChanged` should be fired during `activate`

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes/actiongroup.py
+++ b/pyqtgraph/parametertree/parameterTypes/actiongroup.py
@@ -1,3 +1,5 @@
+import warnings
+
 from ...Qt import QtCore
 from .action import ParameterControlledButton
 from .basetypes import GroupParameter, GroupParameterItem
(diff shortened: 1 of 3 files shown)


j9ac9k (approved):
LGTM!

ksunden:
Any public API concerns from the name change?

ntjess:
@ksunden This class was introduced in 0.13, so thankfully the API issue isn't binding at this point. Also, the parameter type is considered "private" (leading underscore), so even since the 0.13 release I don't think it would be widely adopted: https://github.com/pyqtgraph/pyqtgraph/blob/d6f2da977ece7cebca7e09e30b1b3e32a5f9762b/pyqtgraph/parametertree/parameterTypes/__init__.py#L33

j9ac9k:
I think we should probably leave the old class with a deprecation warning, saying it will be removed in the first pyqtgraph release after January, 2023.  The old class, calling the "correct" named class.  If it wasn't already in a release I would say we can just rename it, but it's been like this for long enough it _might_ impact someone.

ntjess:
Very understandable. Will do

pyqtgraph/parametertree/parameterTypes/actiongroup.py
```diff
@@ -58,3 +60,19 @@ def setButtonOpts(self, **opts):
...
+
+class ActionGroup(ActionGroupParameter):
+    sigActivated = QtCore.Signal()
+
+    def __init__(self, **opts):
+        warnings.warn(
+            "`ActionGroup` is deprecated and will be removed in the first release after "
+            "January 2023. Use `ActionGroupParameter` instead. See "
+            "https://github.com/pyqtgraph/pyqtgraph/pull/2505 for details.",
+            DeprecationWarning,
```
j9ac9k, line 73: probably want to add `stacklevel=2` after `DeprecationWarning`; see any of our usage of deprecation warnings in the library 😆 (it's an awful default in the standard lib)

ntjess:
Good catch!

ntjess:
@j9ac9k Ok to merge?

j9ac9k:
whops, yes! Sorry! I'll press the buttan!

PR #2507, opened 2022-10-27
See if updating mpl fixes CI

ksunden:
A somewhat lazy "I don't want to think too hard" PR to see if we can get 3.10/pyside6-essentials/pip CI runs to pass by simply updating mpl

### DIFF
--- a/.github/workflows/requirements.txt
+++ b/.github/workflows/requirements.txt
@@ -13,7 +13,7 @@ numba==0.55.2; python_version == '3.9'
 pyopengl==3.1.6
 
 # supplimental tools
-matplotlib==3.5.2
+matplotlib==3.6.1
 h5py==3.7.0
 
 # testing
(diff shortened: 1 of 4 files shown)


j9ac9k:
🤦🏻  I didn't realize we had mpl pinned there.  Thanks for looking into it 👍🏻

ksunden:
Well it didn't fix it...

j9ac9k:
https://github.com/matplotlib/matplotlib/pull/24158

This is the PR that added support for PySide6 6.4; not sure if matplotlib has done a release since this was merged.

ksunden:
Hmmm... wonder why I was having trouble reproducing locally

But no, no mpl release since that PR

pijyoi:
BTW, qt5agg and qtagg are aliases. Both work on Qt6 provided you have mpl >= 3.5.0

ksunden:
Yes, but also no... The qt5 variant sets a flag to "Force Qt5 usage"

Now, that flag doesn't have any effect if one of the Qt libs is already imported so in our usage it mostly doesn't matter... but still probably better to use the one that is explicitly for all bindings.

j9ac9k:
Going to merge despite the segfault, YOLO!

j9ac9k:
Thanks @ksunden for looking into this, and for highlighting that we had matplotlib pinned in CI, left it pinned for the time being, but I'll likely revisit on the next time I update the CI.

PR #2509, opened 2022-10-28
Use non-deprecated QMouseEvent signatures

j9ac9k:
In Qt 6.4, the QMouseEvent signature we were using has been deprecated. To work around it, we get the global mouse position by using widget.mapToGlobal(pos), which gives us the global_position.  This feels a bit hacky, but if it is stupid and it works, it is not stupid.


j9ac9k:
Would love to know what's triggering the segfault on linux + pyside6 6.4; if anyone can replicate locally please let me know...

PR #2510, opened 2022-10-28
Changed exportDialog

### LINKED ISSUE #2367: ExportDialog Drawn Off Screen
Depending on the size of the scene the export dialog box can be drawn off or partially off screen. This is due to an implementation of the `show` command that allows moving the box to negative pixel indices.

Problem Code:
https://github.com/pyqtgraph/pyqtgraph/blob/a5f48ec5b58a10260195f1424309f7374a85ece7/pyqtgraph/GraphicsScene/exportDialog.py#L57-L62

To fix this, the position calculation can be clipped using `max`, and the `setGeometry` command can be changed to `move` to account for the size of the window's frame.

Potential Fix:
```python
        if not self.shown:
            self.shown = True
            vcenter = self.scene.getViewWidget().geometry().center()
            x = max(0, int(vcenter.x() - self.width() / 2))
            y = max(0, int(vcenter.y() - self.height() / 2))
            self.move(x, y)
```

I can't say I understand the motivation for moving the dialog box in the first place, but atleast with this modification the dialog box is always accessible with the mouse.

aksy2512:
This pull request fixes #2367. The code has been changed as it was suggested in the issue.

image


j9ac9k:
This LGTM, thanks for the PR @aksy2512 !

Merging

PR #2511, opened 2022-10-28
ErrorBarItem: call update() last

pijyoi:
in test_ErrorBarItem.py, processEvents() was called twice to resolve some flakiness in the test. (2fd533721589cebecf26f2761ccf5eb8db4b94dd)
moving update() to the end of the method seems to make the tests stable without resorting to calling processEvents() twice.

In fact, when running the test suite under WSL2, the error appears reliably. (Running the single test alone will pass though) 
The changes in this PR make those errors go away on WSL2, so hopefully this is the correct fix.


pijyoi:
Errors are showing again on WSL2, so this was not the correct fix after all.

j9ac9k:
You got my hopes up with this PR 😂

PR #2512, opened 2022-10-28
Revert "instantiate QMenu with a parent"

### LINKED ISSUE #2497: `QMenu`s not destroyed (memory leak)

pijyoi:
This reverts commit 2ae77165aea17d517aee0abe920d62cdd86ad299 which was meant to fix #2497.
However, this commit seems to be responsible for causing segfaults in the CI for Ubuntu 20.04 / Python 3.10.

It has been observed to cause segfaults for both PySide6 6.3.2 and PySide6 6.4.0.1 when run under Python 3.10.


j9ac9k:
Thanks @pijyoi for your investigation and the PR; hopefully we get an upstream fix!

PR #2516, opened 2022-10-30
Maintain argument propagatoin to super class

### LINKED ISSUE #2514: Bug: 3D widgets didn't use `parentItem`
# Problem

Widgets like [`GLLineItem`](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/opengl/items/GLImageItem.py#L16) need to have `parentItem` option and propagate it to [`GLGraphicsItem`](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/opengl/GLGraphicsItem.py#L34). In other words, proper inheritance needed to avoid calling `setParentItem` each time.

# Supposed solution

Revision of item graphics widgets required, `__init__` have to contain `parentItem` before `**kwargs` with proper propagation to super class.
```python
def __init__(self, *args, parentItem=None, **kwargs):
    super().__init__(self, parentItem=parentItem)
```

koutoftimer:
All classes from `pyqtgraph.opengl.__init__` based on `GLGraphicsItem` didn't respect `parentItem` argument, which leads to `setParentItem()` call each time you want to group up graphics.

Fixes #2514

### DIFF
--- a/pyqtgraph/opengl/GLGraphicsItem.py
+++ b/pyqtgraph/opengl/GLGraphicsItem.py
@@ -31,14 +31,14 @@ GLOptions = {
 class GLGraphicsItem(QtCore.QObject):
     _nextId = 0
     
-    def __init__(self, parentItem=None):
+    def __init__(self, parentItem: 'GLGraphicsItem' = None):
         super().__init__()
         self._id = GLGraphicsItem._nextId
         GLGraphicsItem._nextId += 1
         
-        self.__parent = None
+        self.__parent: GLGraphicsItem | None = None
         self.__view = None
-        self.__children = set()
+        self.__children: set[GLGraphicsItem] = set()
         self.__transform = Transform3D()
         self.__visible = True
         self.__initialized = False
--- a/pyqtgraph/opengl/items/GLAxisItem.py
+++ b/pyqtgraph/opengl/items/GLAxisItem.py
@@ -12,8 +12,8 @@ class GLAxisItem(GLGraphicsItem):
     
     """
     
-    def __init__(self, size=None, antialias=True, glOptions='translucent'):
-        GLGraphicsItem.__init__(self)
+    def __init__(self, size=None, antialias=True, glOptions='translucent', parentItem=None):
+        super().__init__(parentItem=parentItem)
         if size is None:
             size = QtGui.QVector3D(1,1,1)
         self.antialias = antialias
--- a/pyqtgraph/opengl/items/GLBarGraphItem.py
+++ b/pyqtgraph/opengl/items/GLBarGraphItem.py
@@ -7,7 +7,7 @@ from .GLMeshItem import GLMeshItem
 
 
 class GLBarGraphItem(GLMeshItem):
-    def __init__(self, pos, size):
+    def __init__(self, pos, size, parentItem=None):
         """
         pos is (...,3) array of the bar positions (the corner of each bar)
         size is (...,3) array of the sizes of each bar
@@ -27,4 +27,4 @@ class GLBarGraphItem(GLMeshItem):
         faces = cubeFaces + (np.arange(nCubes) * 8).reshape(nCubes,1,1)
         md = MeshData(verts.reshape(nCubes*8,3), faces.reshape(nCubes*12,3))
 
-        GLMeshItem.__init__(self, meshdata=md, shader='shaded', smooth=False)
+        super().__init__(meshdata=md, shader='shaded', smooth=False, parentItem=parentItem)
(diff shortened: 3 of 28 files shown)


koutoftimer:
I can see that for 2D graphics there are tests for parent, but not for 3D, I suppose I have to add tests to this merge request.

pyqtgraph/opengl/GLGraphicsItem.py
```diff
@@ -31,14 +31,14 @@
...
     _nextId = 0
     
-    def __init__(self, parentItem=None):
+    def __init__(self, parentItem: 'GLGraphicsItem' = None):
         super().__init__()
         self._id = GLGraphicsItem._nextId
         GLGraphicsItem._nextId += 1
         
-        self.__parent = None
+        self.__parent: GLGraphicsItem | None = None
```
j9ac9k, line 39: Love to see type annotations, but I'm not sure we can use this syntax, as we support Python 3.8 still.  We will likely add typing-extensions as a dependency at some point but we are not there yet.

pyqtgraph/opengl/GLGraphicsItem.py
```diff
@@ -31,14 +31,14 @@
...
         self.__view = None
-        self.__children = set()
+        self.__children: set[GLGraphicsItem] = set()
```
j9ac9k, line 41: For Python 3.8; I don't think we can use the generic containers, I think we still need to use 'typing.Set'

j9ac9k (commented):
Love the extra tests!

pyqtgraph/opengl/GLGraphicsItem.py
koutoftimer: To clarify, python3.8 as a runtime happily ignores typehints during execution. It is confirmed by the tests. Typehints needed for tooling you are using. Are you telling that maintaining support for python 3.8 means not only runtime but also all tools that have no clue about anything above 3.8?

j9ac9k: oh nice, I got that in my head that it would raise syntax errors or something; happy to restrict type checking to the current/newest version of python.

j9ac9k:
Hi @koutoftimer 

Could you add some pytest.importskip to skip tests if pyopengl is not installed?  You can see an example of how we do that with matplotlib here: https://github.com/pyqtgraph/pyqtgraph/blob/8a3c4fd6f00085be87ab83660fbe9c37877de63f/tests/exporters/test_matplotlib.py#L8

koutoftimer:
@j9ac9k I can, but I will ask you to explain to me why is it important, because I don't quite understand.

j9ac9k:
> @j9ac9k I can, but I will ask you to explain to me why it is important, because I don't quite understand.
> 
> 
> 
> From the user's perspective I can see that there are no need to install opengl without actually using it, but as developer what is the guarantee that doing your stuff you will not break anything behind `importorskip` ? To asnwer this question you have to install opengl in dev environment, so why to bother?

The main reason is that pyopengl is an optional dependency as you were commenting and the test suite should not fail in the absence of an optional dependency.

On my machine I have about 13 PyQtGraph virtual environments I use for development and testing; not all of them need pyopengl.

koutoftimer:
@j9ac9k why do you need 13 virtual environments? I can imagine a lot of pros and cons of using `impororskip` but I'd like to know real world test case.

j9ac9k:
I have so many environments due to testing different versions of Qt bindings and different versions of Python; and one environment for documentation builds.

Basically it's often I want to try and replicate a test failure, and run `pip install numpy pyqt6 pytest` and then run `pytest tests`.

The CI system has pyopengl for all pipelines so it should catch any issues that way.

j9ac9k:
Oh, the other gotcha w/ pyopengl is that on Python 3.8 and recent versions of macOS pyopengl won't work... we have a note about that in the README.

koutoftimer:
@j9ac9k if you still concerned about annotations, I can add `from __future__ import annotations`

j9ac9k:
> @j9ac9k if you still concerned about annotations, I can add `from __future__ import annotations`

Sorry for the long wait here, a lot happens over here through the holidays and I end up taking an extended break from OSS work during that time. 

I would rather avoid the `from __future__ import annotations` so using string based annotations are fine, they're just new to the library is all :)

test suite passed on my system without pyopengl so thanks for adding that skip check!  This looks good to me, will give one last lookover before merging.

j9ac9k:
@koutoftimer thanks for your patience, this LGTM, merging!

Sorry for the delay!

PR #2517, opened 2022-11-02
Fix vertical alignment of spans for highlighted code block

j9ac9k:
When viewing the source code inside dash, the highlighted code block would have vertical alignment of all the spans.  This did not occur in the regular/html version of the docs.  I believe this is a bug/issue w/ the pydata-sphinx-theme, but this was fixed by specifying `display: block` in `dash.css` 

See https://github.com/hynek/doc2dash/issues/135 for more information


j9ac9k:
really never thought submitting `css` work product would be my way of contributing to the library, yet here we are...

PR #2519, opened 2022-11-04
fix setMaximumHeight(1e6) in  SpinBox.py

sem-geologist:
After upgrading my KDE Neon and thus Qt to 5.15.6 I had found out that my software hiccups at pytqgraph SpinBox. I had tracked down problem to SpinBox.py, where method `setMaximumHeight` is fed with `1e6` which is float, and these are pixels (logically it should be integer numbers). Interestingly older Qt versions had not complained before and was consuming float without any complaining. Simple fix is to change this `1e6`  to `1000000`.

This fix is simple and straightforward.

### DIFF
--- a/pyqtgraph/widgets/SpinBox.py
+++ b/pyqtgraph/widgets/SpinBox.py
@@ -572,7 +572,7 @@ class SpinBox(QtWidgets.QAbstractSpinBox):
         # SpinBox has very large margins on some platforms; this is a hack to remove those
         # margins and allow more compact packing of controls.
         if not self.opts['compactHeight']:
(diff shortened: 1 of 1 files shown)


j9ac9k:
Hi @sem-geologist 

Thanks for this fix; Python 3.10 removed a compatibility layer that I believe under the hood did a lot of the float-to-int conversions for us within the bindings.  When Python 3.10 was coming out, we patched up a bunch of these, but clearly, a few instances slipped through the cracks.

It looks like we have a similar issue with AxisItem.py, although I haven't seen any issue like it.

Thanks for the PR @sem-geologist merging.

Issue #2520, opened 2022-11-05
test_busycursor.py segfaults on Linux PySide6 6.3, 6.4


pijyoi:
### Short description

test_busycursor.py segfaults under the following conditions
* on Linux PySide6 6.3, 6.4 *and*
* *when executed alone*

### Code to reproduce

```python
pytest tests/widgets/test_busycursor.py
```

### Real behavior

```
(py310_pyside63) :~/repos/pyqtgraph$ pytest tests/widgets/test_busycursor.py
============================================== test session starts ===============================================
platform linux -- Python 3.10.6, pytest-7.2.0, pluggy-1.0.0
rootdir: /home/user/repos/pyqtgraph, configfile: pytest.ini
collected 1 item

tests/widgets/test_busycursor.py .                                                                         [100%]

=============================================== 1 passed in 0.07s ================================================
Segmentation fault
```

### Tested environment(s)

 * PyQtGraph version: master 
 * Qt Python binding: PySide6 6.4.0.1, 6.3.2
 * Python version: 3.10.6
 * NumPy version: 1.23.4 
 * Operating system: Ubuntu 22.04
 * Installation method: pip 

### Additional context
When tested on the CI running just this test, it does not happen on macOS and Windows, nor on Linux with PySide 5.15, 6.2.4

While debugging the QMenu leak issue in #2497, various segfaults at end of *successful* pytest run could be triggered on Linux PySide6 >= 6.3. This may be related.

pijyoi:
A MWE would be the following
```python
from PySide6 import QtGui
app = QtGui.QGuiApplication([])
t = app.thread()
```
i.e. the program will segfault on program end if you take the gui thread object.

pijyoi:
Looks like a bug report got filed https://bugreports.qt.io/browse/PYSIDE-2254 and a fix is pending https://codereview.qt-project.org/c/pyside/pyside-setup/+/465817

pijyoi:
The fix has been merged to 6.4 https://code.qt.io/cgit/pyside/pyside-setup.git/commit/?h=6.4&id=4ee2bfc29b9367f14a60a6b81a83bd92d7c24f25
So the skipped tests can probably be reinstated when 6.4.3 gets released.

PR #2518, opened 2022-11-02
partial fix for Qmenu leak

pijyoi:
Partially fixes #2497

There are 3 places that leak `QMenu`s. This PR fixes 2 of them. The last one is left unfixed as fixing it causes a CI segfault on PySide6 >= 6.3 on Linux platforms. It is believed that the last leak is masking a latent issue.

The workaround used has changed from #2502. In #2502, "topmenu" was made the `C++ parent` of "submenu".
It was found that PySide{2,6} binding makes "topmenu" a `Python-wrapper parent` of "submenu", increasing the refcnt of "submenu", even though "submenu" has no `C++ parent`.
The workaround here relies on PySide's `setParent(None)` dropping this refcnt. (Note that the `C++ parent` was already `None` to begin with)
The end-result is that the life-state of "submenu" remains the same before and after `addMenu`, as shown in the output of the following script: refcnt and shiboken.dump output remains the same.

If upstream fixes the leak, then this `setParent(None)` call should have no effect.

```python
from PySide6 import QtGui, QtWidgets
import shiboken6 as shiboken
import sys

app = QtWidgets.QApplication([])
topmenu = QtWidgets.QMenu()
topmenu.setObjectName("topmenu")
submenu = QtWidgets.QMenu()
submenu.setObjectName("submenu")

print(shiboken.dump(submenu))

print(sys.getrefcount(submenu))
topmenu.addMenu(submenu)
submenu.setParent(None) # let Python do the cleanup
print(sys.getrefcount(submenu))
del topmenu
print(sys.getrefcount(submenu))
print(shiboken.dump(submenu))
```
Output
```
C++ address....... PySide6.QtWidgets.QMenu/0000021A1C923C40
hasOwnership...... 1
containsCppWrapper 1
validCppObject.... 1
wasCreatedByPython 1

2
2
2
C++ address....... PySide6.QtWidgets.QMenu/0000021A1C923C40
hasOwnership...... 1
containsCppWrapper 1
validCppObject.... 1
wasCreatedByPython 1
```

As mentioned, this PR only fixes 2 of the leaks. When running the test scripts in #2497, the rate of memory growth is reduced and only 1 QMenu is leaked.

PR #2522, opened 2022-11-05
Fix last QMenu leak and its associated segfaults

### LINKED ISSUE #2497: `QMenu`s not destroyed (memory leak)

pijyoi:
Fixes #2497.
Completes #2518.
With the segfaults being attributed to the reason listed in #2520, this PR avoids taking the gui thread object.
This includes:
1) switching `SignalProxy` to use `QTimer` instead of `ThreadsafeTimer`
2) skipping tests on widgets that take the gui thread object in their implementation

Thus there are 2 bugs in PySide binding
1) QMenu leak on `QMenu::addMenu(QMenu*)` (as detailed in #2518)
2) gui thread object segfault since PySide6 >= 6.3 (as detailed in #2520)

### DIFF
--- a/pyqtgraph/SignalProxy.py
+++ b/pyqtgraph/SignalProxy.py
@@ -20,20 +20,23 @@ class SignalProxy(QtCore.QObject):
 
     sigDelayed = QtCore.Signal(object)
 
-    def __init__(self, signal, delay=0.3, rateLimit=0, slot=None):
+    def __init__(self, signal, delay=0.3, rateLimit=0, slot=None, *, threadSafe=True):
         """Initialization arguments:
         signal - a bound Signal or pyqtSignal instance
         delay - Time (in seconds) to wait for signals to stop before emitting (default 0.3s)
         slot - Optional function to connect sigDelayed to.
         rateLimit - (signals/sec) if greater than 0, this allows signals to stream out at a
                     steady rate while they are being received.
+        threadSafe - Specify if thread-safety is required. For backwards compatibility, it
+                     defaults to True.
         """
 
         QtCore.QObject.__init__(self)
         self.delay = delay
         self.rateLimit = rateLimit
         self.args = None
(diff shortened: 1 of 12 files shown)


pijyoi:
With the cause of the segfault out of the way, a cleaner implementation of building the mouse submenu can be done.

pijyoi:
I am trying to move away from the use of `ThreadsafeTimer` in `SignalProxy`, while maintaining backwards compatibility.
Possibilities:
1) create another class (e.g. SignalThrottler) that uses `QTimer` and change all library uses to this new class.
    * adding yet another public class
2) add a parameter `threadSafe=True` to `SignalProxy` and change all library uses to adding `threadSafe=False`.
    * unwieldy
3) create a new class `SignalProxyThreadsafe` and bind it to `pg.SignalProxy`. Change `SignalProxy` to use `QTimer`. Thus all library uses will end up using the `QTimer` version, while external users that use `pg.SignalProxy` still use the `ThreadsafeTimer` version. (this is what is currently implemented in this PR)
    * having behavior change depending on how you import it doesn't sound good
4) ignore backwards compatibility and just switch to `QTimer`.

pijyoi:
With the change of `ubuntu-latest` to `ubuntu-22.04`, the Linux / Python 3.9 / PySide6.2 pipeline segfaults.
Testing on py38, py39, py310 with PySide6.2 showed segfaults on py38, py39 (but not py310).

Fixed by skipping `test_busycursor` and `test_progressdialog` also on PySide6.2

j9ac9k:
Thanks for your work on this @pijyoi I know this one was a doozy to sort out.

PR #2526, opened 2022-11-16
bump dev version to correct value

ntjess:
After the 0.13.1 release, the current github repo should be be 0.13.2.dev0, but is 0.13.1.dev0. This PR fixes the issue.


ntjess:
@j9ac9k Does this mean I can claim the legendary 1-char PR badge? 😆 It's a shame hacktoberfest is over

j9ac9k:
thanks for the fix @ntjess and yes, suppose this would qualify! 😆

PR #2536, opened 2022-11-26
Update `PenParameter`

### LINKED ISSUE #2388: `pen` parameter fires `sigValueChanged` without setting value
This line should be `setValue` rather than emitting `sigValueChanged`:
https://github.com/pyqtgraph/pyqtgraph/blob/8be2d6a88edc1e26b1f6f85c47a72c400db9a28b/pyqtgraph/parametertree/parameterTypes/pen.py#L52

In all other parameters, the signal doesn't fire until after the value is set.

ntjess:
Pretty straightforward as described in the original issue.

- [x] Fixes #2388
- [x] Adds a default button at the pen level + propagates `setDefault` to parameter children

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes/pen.py
+++ b/pyqtgraph/parametertree/parameterTypes/pen.py
@@ -1,26 +1,52 @@
 import re
 from contextlib import ExitStack
 
-from . import GroupParameterItem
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks @ntjess this LGTM, merging!

PR #2537, opened 2022-11-26
Remove parameter tree size hint overloads

ntjess:
I might be missing something, so it would be great for others to verify. On my machine, parameter tree & item sizing appears to work better by removing all custom size hint logic... The biggest difference is spawning a window with just a parameter tree
- Before this change, the window had to be manually resized to fit all parameters
- Now, (for me) the window is correctly sized immediately.


j9ac9k:
This is what it looks like on `master` for me:

image


This is what it looks like on your branch

image

Not seeing too much of a difference there, but on initial launch of the parametertree example:

Master:
image


This branch: 
image

j9ac9k:
I'm going to close out this PR as it doesn't seem to fix anything :)

PR #2539, opened 2022-11-26
fix: instantiate QApplication for test_Parameter.py

pijyoi:
As `test_Parameter.py` instantiates some gui elements, it needs `QApplication` to have been instantiated. When this test is run alone, it will segfault. i.e. it was relying on other tests in the test-suite having instantiated `QApplication`.


j9ac9k:
Nice catch @pijyoi !  LGTM, merging

PR #2540, opened 2022-11-26
Change viewBox.py

### LINKED ISSUE #2415: `ViewBox.setLimits()` doesn't trigger update if view range not within new limits
### Description
If you call for example `ViewBox.setLimits(minXRange=limit)` with `limit` being larger as the current visible range, the `ViewBox` doesn't update its view range. We are then in a state where the limits are not satisfied.
My expectation is, that the view range is updated immediatly with the call of `setLimits()`.
Only after some other trigger, like updating the view range by `ViewRange.setRange()`, the view range suddenly changes and the limits are satisfied. 
This can be seen as a workaround: Just call `setRange` with the current range.
This is behaviour seems to be the case for all eight limit parameters (`xMin`, ..., `yMax`, `minXRange`, ..., `maxYRange`)

### Code to reproduce
Use the following mwe to check the current behaviour:
The width of the `LinearRegionItem` sets the `minXRange` limit of the right plot.
At the beginning it's all good, the region has a width of 10 and the right plot a range of 20.
Increasing the regions width to above 20 doesn't affect the right plot.
Now, try to pan the right plot just a little bit, this will trigger the limits and the plot will "jump" into a state where the limits are satisfied.

The expected behaviour can be seen, if the workaround is enabled (see the `if 0` in the code).

```
import pyqtgraph as pg

app = pg.mkQApp()

win.resize(1000,600)

p1 = win.addPlot(title="Region defines limit for other plot")
lr = pg.LinearRegionItem([0,10])
p1.addItem(lr)
p1.setXRange(0,50)

p2 = win.addPlot()
p2.setXRange(0,20)

def lr_updated():
    r = lr.getRegion()
    region_span = r[1] - r[0]
    print(f"p2.setLimits(minXRange={region_span})")
    p2.setLimits(minXRange=region_span)

    if 0:  # workaround
        p2.setRange(xRange=p2.viewRange()[0], padding=0.0)

lr.sigRegionChanged.connect(lr_updated)
lr_updated()

if __name__ == '__main__':
    pg.exec()
```

### Suggested solution

I think the solution would be to extend `ViewBox.updateViewRange()` at line 1576.
Between `for axis in [0, 1]:` and `if aspect is not False and ...` one should compare `viewRange` against `limits`, `minRng` and `maxRng` and adjust `viewRange` accordingly.

That shouldn't be that much work, so if you agree with this suggestion, I might find time in the near future to come up with a PR.

aksy2512:

Fixes #2415 

I wrote the following code inside the updateViewRange function 
image


bbc131:
Hi @aksy2512,
I just had a look at your PR and also tested it quickly.

It solves the problem partly ;)
I took my example from #2415 and observe the following:
Making the region larger correctly triggers the right plot at some point to increase its x-range. 
But if you make the region than smaller again, this will decrease the right plots x-range also! This definitely shouldn't be the case.
I just looked around in the `updateViewRange` method and saw, that one might have to adjust `self.state['targetRange']`?! 
I hadn't the time to look closer into it, this is just a wild guess.

Another remark on formatting:
I suggest you to use black [1] to get a nice formatting. 
For example I don't like the brackets at the "if"s and black would also normalize all whitespaces.
[1] https://pypi.org/project/black/

aksy2512:
@j9ac9k I have rewritten the code and used "for" loop instead of the if statements. Please have a look.
Thanks

pyqtgraph/graphicsItems/ViewBox/ViewBox.py
```diff
@@ -1573,6 +1572,14 @@ def updateViewRange(self, forceX=False, forceY=False):
                     maxRng[axis] = min(maxRng[axis], limits[axis][1] - limits[axis][0])
                 else:
                     maxRng[axis] = limits[axis][1] - limits[axis][0]
+        for axis in [0,1]:
```
j9ac9k, line 1575: Can this line be removed and incorporate the logic added into the for loop above?

j9ac9k:
Hi @aksy2512 

Sorry for the really long wait for feedback... ViewBox.py is among the most fragile parts of the library, so naturally I gravitate towards avoiding changes to it unless absolutely necessary.

There is an issue I'm a bit hung up on that I posted in the channel, why should the second limit be adjusted and not the first limit as this PR does.

I was going to suggest adding a test, but I can see we already skip the viewbox limits/resize test already as it's not working, so that would not be a reasonable request for a newcomer to the library :)

j9ac9k:
@bbc131 made a good point in this post here https://github.com/pyqtgraph/pyqtgraph/issues/2415#issuecomment-1396564259

They point out that the current logic in this PR makes no attempt to compare the computed upper bound of the range against xMax/yMax, so there could be a situation where the newly computed upper bound is outside of the specified limits.  If that occurs, the lower bounds should likely be adjusted when the upper bound is placed at the xMax or yMax limit.

PR #2544, opened 2022-12-02
Improve `ChecklistParameter.setValue` logic.

ntjess:
Previously, calling `setValue` would properly fill in child checklist widgets, but wouldn't populate `self.opts["value"]` with the correct option. This was because setting `blockSigal=self._onChildChanging` meant `childrenValues()` wasn't notified that child values were updated. Simply creating and passing a `valueToSet` object prevents these issues.

I found this bug by calling `restoreState`, which updates a checklist's value without bubbling up from its widget children.

As a bonus, I (at least tried to) improve the readability of `setValue` logic.


j9ac9k:
Thanks for the PR @ntjess this LGTM, merging!

PR #2548, opened 2022-12-09
use ThreadsafeTimer class instead of module

ntjess:
Very simple fix: after bf3820a,  `SignalProxy` imports the `ThreadsafeTimer` module instead of class. MWE below fails:

```python
import pyqtgraph as pg
pg.mkQApp()
edit = pg.QtWidgets.QTextEdit()
sp = pg.SignalProxy(edit.textChanged)

# TypeError: 'module' object is not callable
```

This PR imports the class instead.

### DIFF
--- a/pyqtgraph/SignalProxy.py
+++ b/pyqtgraph/SignalProxy.py
@@ -1,9 +1,9 @@
 import weakref
 from time import perf_counter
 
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks @ntjess for this patch!

PR #2554, opened 2022-12-14
compute floating point boundingRect

### LINKED ISSUE #2550: BarGraphItem is not showing if bar coordinates are huge
### Short description
See example below, use `data` to create a `BarGraphItem`, there will be nothing, whereas use `data_ok` everything is normal.

### Code to reproduce
```python
"""
Simple example using BarGraphItem
"""

import numpy as np
import pyqtgraph as pg

win = pg.plot()
win.setWindowTitle('pyqtgraph example: BarGraphItem')

# prepare data

# bars are invisible, might show up when resizing
data = np.random.randint(2E9, 4E9, 1000)

# # if data is not very big, everything is fine
# data_ok = np.random.randint(200, 400, 1000)

counts, edges = np.histogram(data, bins=50)
height = edges[1]-edges[0]
bg1 = pg.BarGraphItem(x0=0, y=edges[:-1], height=height, width=counts, brush='r')
# set viewbox range
vb = win.getPlotItem().getViewBox()
vb.setLimits(xMin=0, 
             xMax=counts.max() * 1.2,
             yMin=edges[0] * 0.9, 
             yMax=edges[-1] * 1.1)

win.addItem(bg1)

if __name__ == '__main__':
    pg.exec()
```

### Real behavior
Blank scene.
image

### Tested environment(s)

 * PyQtGraph version: 0.13.1
 * Python version: 3.9
 * NumPy version: 1.20.3
 * Operating system: macOS
 * Installation method: pip

pijyoi:
Fixes #2550
A few `GraphicsItem`s compute their bounding rectangle with `QPicture::boundingRect() -> QRect`, probably following the example in `customGraphicsItem.py`. This causes issues when the data limits exceed the int32 range of a `QRect`.

This PR fixes the cases where the bounding rectangle can be easily built up from the drawing primitives used.

### DIFF
--- a/pyqtgraph/examples/customGraphicsItem.py
+++ b/pyqtgraph/examples/customGraphicsItem.py
@@ -19,6 +19,7 @@ class CandlestickItem(pg.GraphicsObject):
     def generatePicture(self):
         ## pre-computing a QPicture object allows paint() to run much more quickly, 
         ## rather than re-drawing the shapes every time.
+        bounds = QtCore.QRectF()
         self.picture = QtGui.QPicture()
         p = QtGui.QPainter(self.picture)
         p.setPen(pg.mkPen('w'))
@@ -30,7 +31,9 @@ class CandlestickItem(pg.GraphicsObject):
             else:
                 p.setBrush(pg.mkBrush('g'))
             p.drawRect(QtCore.QRectF(t-w, open, w*2, close-open))
+            bounds |= QtCore.QRectF(t-w, min, w*2, max-min)
         p.end()
+        self._bounds = bounds
     
     def paint(self, p, *args):
         p.drawPicture(0, 0, self.picture)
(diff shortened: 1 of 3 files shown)


j9ac9k:
Oh no, are we going to have to generate new test images?

pijyoi:
> Oh no, are we going to have to generate new test images?

No, there was some self-referencing in `NonUniformImage`. The `QPicture` was used to compute the bounding rectangle; yet while the paint method had not completed drawing onto the `QPicture`, `boundingRect()` was called to draw the optional border.

j9ac9k:
This LGTM, thanks @pijyoi !

pijyoi:
Hmm, the bounding rectangle computed this way ignores the pen width. Practically speaking, it doesn't seem to cause a problem. 

It doesn't seem easy to incorporate the pen width into the bounding rectangle, as the pen width is in device pixels.

PR #2561, opened 2022-12-25
BarGraphItem: implement boundingRect with pen width

pijyoi:
Fixes #2550
Related to #2554, #2560 

After playing around with thicker pen widths, it turns out that `QPicture.boundingRect()` assumes that pens are non-cosmetic; when in fact pyqtgraph usually uses cosmetic pens. This means that `QPicture.boundingRect()` adds the pen widths to the bounding rectangle in data sizes rather than in device pixel sizes.

A sample script below demonstrates the issue. When a thick pen is used, it results in a lot of empty space.

Not being familiar with the very code that this PR changes, this PR is created as a draft.

```python
import numpy as np
import pyqtgraph as pg
from pyqtgraph.Qt import QtCore, QtGui, QtWidgets

class ObjectBounds(QtWidgets.QGraphicsItem):
    def paint(self, painter, *args):
        pen = QtGui.QPen(QtCore.Qt.GlobalColor.red, 0, QtCore.Qt.PenStyle.DashLine)
        rect = self.boundingRect()
        painter.setPen(pen)
        painter.drawRect(rect)
        print(rect)

    def boundingRect(self):
        return self.parentItem().boundingRect()

pg.mkQApp()
pw = pg.PlotWidget()
pw.show()

rng = np.random.default_rng()
vals = np.hstack([rng.normal(size=500), rng.normal(size=260, loc=4)])
y,x = np.histogram(vals, bins=np.linspace(-3, 8, 40))

pen = pg.mkPen('w', width=6)
bgi = pg.BarGraphItem(x0=x[:-1], x1=x[1:], height=y, pen=pen, brush=(0,0,255,150))
pw.addItem(bgi)
rect = ObjectBounds(bgi)

pg.exec()
```

master: PyQt6.QtCore.QRectF(-6.0, -3.0, 17.0, 69.0)
Screenshot 2022-12-25 171523

this PR: PyQt6.QtCore.QRectF(-3.082433908183643, -0.6809981495366643, 11.164867816367286, 67.36199629907333)
Screenshot 2022-12-25 171451

Issue #1141, opened 2020-03-13
Artifacts when moving ROI with significant pen width


ixjlyons:
### Short description

This issue comes from the mailing list: https://groups.google.com/d/msg/pyqtgraph/tWtjJOQF5x4/CnQF6IgmDAAJ

Dragging around an ROI with a large pen width results in artifacts. For me it only seems to occur after zooming/scaling the viewbox.

### Code to reproduce

```python
import pyqtgraph as pg
import numpy as np

app = pg.mkQApp()

imgdata = np.random.rand(100, 100)

plt = pg.plot()
plt.setAspectLocked(True)

img = pg.ImageItem(imgdata, levels=(0, 1))
plt.addItem(img)

roi = pg.RectROI((0, 0), (50, 50), pen=pg.mkPen('r', width=10))
plt.addItem(roi)

app.exec_()
```

### Expected behavior

Only red in the image should be the square ROI border.

### Real behavior

![Screenshot_2020-03-12_20-08-30](https://user-images.githubusercontent.com/943602/76586161-49dbb500-649d-11ea-9692-cadd859574a5.png)

### Tested environment(s)

 * PyQtGraph version: 0.11.0.dev0+gdb483f8em
 * Qt Python binding: 5.14.0
 * Python version: 3.8.2
 * NumPy version: 1.17.4
 * Operating system: Linux
 * Installation method: pip

### Additional context

There's a proposed solution in the reply to the mailing list post, though I think it double accounts for half pen width.

JoeZiminski:
Hi ixjlyons,

I wonder if this is related to this issue: https://github.com/pyqtgraph/pyqtgraph/issues/533 which seems to be a known performance issue with thick line widths in Qt.

A workaround that at least papers over the problem is to edit the ROI to accept the plot as an argument, then force the viewBox to update every time the ROI is moved: 

In line 14 of your reproducable example:
```
roi = pg.RectROI((0, 0), (50, 50), pen=pg.mkPen('r', width=10), master_plt=plt)
```


in pyqtgraph/graphicsItems/ROI.py:

1) from PyQT5.QtWidgets import QApplication
2)  add master_plt=None to the init to the ROI class.
3) replace line 174 with:
```
        self.master_plt = master_plt
        self.mouseDragHandler = MouseDragHandler(self, self.master_plt)
```

4) add master_plt as an attribute of MouseDragHandler()
5) add to the end of the mouseDragEvent() method (line 1515).

```
[...]

Issue #60, opened 2014-04-09
Graphics Items prevent autoscaling of visual-data-only


ghost:
It seems that using BarGraphItem or FillBetweenItem prevents the auto-scaling of the visible-data-only in plots. To reproduce, use the FillBetween example in the pyqtgraph.example demo -- if one sets autoscale to visible data only on the full data set, the y-axis maximum rattles around between about 8 and 12. But if one pushes most of the full data set out of the visible window and just looks at one side of the other, and resets the settings to auto-scale visible-data only, the y-axis is still maximum of 8-12, not something more appropriate for the reduced data span at the beginning or end of the data set. The apparent disabling of the visible-data-only autoscaling occurs even when I add a regular plot curve to the bar data.

Curiously, the CandlestickItem used in the Custom Graphics Example (which I adapted slightly to allow for the close == open) _will_ auto-scale very nicely.

Perhaps all graphics objects need a "boundingRect" function like CandlestickItem has?

PR #2565, opened 2022-12-27
BarGraphItem: implement dataBounds and pixelPadding

pijyoi:
Alternate implementation of #2561.
Implements methods `dataBounds` and `pixelPadding` which are used by `ViewBox`.

One difference that I noticed between this implementation and #2561 is that this implementation takes less iterations to stabilize on Windows. (7 vs 3).

### DIFF
--- a/pyqtgraph/graphicsItems/BarGraphItem.py
+++ b/pyqtgraph/graphicsItems/BarGraphItem.py
@@ -121,6 +121,8 @@ class BarGraphItem(GraphicsObject):
         
         p.setPen(fn.mkPen(pen))
         p.setBrush(fn.mkBrush(brush))
+        dataBounds = QtCore.QRectF()
+        pixelPadding = 0
         for i in range(len(x0 if not np.isscalar(x0) else y0)):
             if pens is not None:
                 p.setPen(fn.mkPen(pens[i]))
@@ -148,8 +150,23 @@ class BarGraphItem(GraphicsObject):
             rect = QtCore.QRectF(x, y, w, h)
             p.drawRect(rect)
             self._shape.addRect(rect)
-            
+
+            pen = p.pen()
+            pw = pen.widthF()
+            if pen.style() == QtCore.Qt.PenStyle.NoPen:
+                pw = 0
+            elif pw == 0:
+                pw = 1
+            pw *= 0.5
+            if pen.isCosmetic():
+                dataBounds |= rect
+                pixelPadding = max(pixelPadding, pw)
+            else:
+                dataBounds |= rect.adjusted(-pw, -pw, pw, pw)
+
         p.end()
+        self._dataBounds = dataBounds
+        self._pixelPadding = pixelPadding
         self.prepareGeometryChange()
         
         
@@ -158,11 +175,6 @@ class BarGraphItem(GraphicsObject):
             self.drawPicture()
         self.picture.play(p)
             
-    def boundingRect(self):
-        if self.picture is None:
-            self.drawPicture()
-        return QtCore.QRectF(self.picture.boundingRect())
-    
     def shape(self):
         if self.picture is None:
             self.drawPicture()
@@ -179,3 +191,27 @@ class BarGraphItem(GraphicsObject):
 
     def getData(self):
         return self.opts.get('x'),  self.opts.get('height')
+
+    def dataBounds(self, ax, frac=1.0, orthoRange=None):
+        if self.picture is None:
+            self.drawPicture()
+        l, t, r, b = self._dataBounds.getCoords()
+        return [l, r] if ax == 0 else [t, b]
+
+    def pixelPadding(self):
+        if self.picture is None:
+            self.drawPicture()
+        return self._pixelPadding
+
+    def boundingRect(self):
+        px = py = 0
+        pxPad = self.pixelPadding()
+        if pxPad > 0:
+            # determine length of pixel in local x, y directions
+            px, py = self.pixelVectors()
+            px = 0 if px is None else px.length()
+            py = 0 if py is None else py.length()
+            # return bounds expanded by pixel size
+            px *= pxPad
+            py *= pxPad
+        return self._dataBounds.adjusted(-px, -py, px, py)


pijyoi:
For a easier comparison of the various ways of computing the bounding rectangle, a standalone script that plots them all together.
Apparently there are some differences when running on macOS, so it would be appreciated if someone could paste the screenshot from running on macOS.
```python
import pyqtgraph as pg
from pyqtgraph import QtCore, QtGui, QtWidgets

class CandlestickItem(pg.GraphicsObject):
    def __init__(self, data):
        self.data = data  ## data must have fields: time, open, close, min, max
        self.generatePicture()
    
    def generatePicture(self):
        ## pre-computing a QPicture object allows paint() to run much more quickly, 
        ## rather than re-drawing the shapes every time.
        dataBounds = QtCore.QRectF()
        self.picture = QtGui.QPicture()
        p = QtGui.QPainter(self.picture)
        pen = pg.mkPen('w', width=6)
        p.setPen(pen)
        brush_pos = pg.mkBrush('r')
        brush_neg = pg.mkBrush('g')
        w = (self.data[1][0] - self.data[0][0]) / 3.
        for (t, open, close, min, max) in self.data:
            p.drawLine(QtCore.QPointF(t, min), QtCore.QPointF(t, max))
            if open > close:
                p.setBrush(brush_pos)
            else:
                p.setBrush(brush_neg)
            p.drawRect(QtCore.QRectF(t-w, open, w*2, close-open))
            dataBounds |= QtCore.QRectF(t-w, min, w*2, max-min)
        p.end()
        self._dataBounds = dataBounds
        self._pixelPadding = pen.widthF()*0.7072

    def paint(self, p, *args):
        p.drawPicture(0, 0, self.picture)
    
    def boundingRect(self):
        if self.picture is None:
            self.generatePicture()

class ItemX(CandlestickItem):
    # provide boundingRect that takes into account pen width
    def boundingRect(self):
        if self.picture is None:
            self.generatePicture()

        px = py = 0.0
        pxPad = self._pixelPadding
        if pxPad > 0:
            # determine length of pixel in local x, y directions
            px, py = self.pixelVectors()
            try:
                px = 0 if px is None else px.length()
            except OverflowError:
                px = 0
            try:
                py = 0 if py is None else py.length()
            except OverflowError:
                py = 0

            # return bounds expanded by pixel size
            px *= pxPad
            py *= pxPad
        boundingRect = self._dataBounds.adjusted(-px, -py, px, py)
        return boundingRect

class ItemY(ItemX):
    # also provide dataBounds and pixelPadding
    def dataBounds(self, ax, frac=1.0, orthoRange=None):
        if self.picture is None:
            self.generatePicture()
        br = self._dataBounds
        if ax == 0:
            return [br.left(), br.right()]
        else:
            return [br.top(), br.bottom()]

    def pixelPadding(self):
        if self.picture is None:
            self.generatePicture()
        return self._pixelPadding

class ObjectBounds(QtWidgets.QGraphicsItem):
        pen = QtGui.QPen(QtCore.Qt.GlobalColor.red, 0, QtCore.Qt.PenStyle.DashLine)
        rect = self.boundingRect()
        painter.setPen(pen)
        painter.drawRect(rect)
        # print(rect)

    def boundingRect(self):
        return self.parentItem().boundingRect()

data = [  ## fields are (time, open, close, min, max).
    (1., 10, 13, 5, 15),
    (2., 13, 17, 9, 20),
    (3., 17, 14, 11, 23),
    (4., 14, 15, 5, 19),
    (5., 15, 9, 8, 22),
    (6., 9, 15, 8, 16),
]

pg.mkQApp()
glw.show()
items = [CandlestickItem(data), ItemX(data), ItemY(data)]
rects = []
for idx, item in enumerate(items):
    plt = glw.addPlot(row=idx, col=0)
    plt.addItem(item)
    rects.append(ObjectBounds(item))
pg.exec()
```

Here's how it looks like on Windows and Linux.
Screenshot 2022-12-27 225050

pijyoi:
I have a test case here where (on Windows) the implementation of #2561 and this PR have different behavior.

With #2561 _and_ on Qt6, the viewbox is not tightly sized on the initial plot. Pressing the "A" button will make it tightly sized. Even minimizing and restoring the window will make it tightly sized.
It doesn't occur with PySide2 and PyQt5. 

With this PR, the viewbox is tightly sized on the initial plot. 

I am not sure if this is a bug of pyqtgraph.

This test-case also shows another deficiency of using `QPicture.boundingRect()` for non-integer coords.

```py
import numpy as np
import pyqtgraph as pg

class ObjectBounds(QtWidgets.QGraphicsItem):
        pen = QtGui.QPen(QtCore.Qt.GlobalColor.red, 0, QtCore.Qt.PenStyle.DashLine)
        rect = self.boundingRect()
        painter.setPen(pen)
        painter.drawRect(rect)
        print(rect)

    def boundingRect(self):
        return self.parentItem().boundingRect()

pg.mkQApp()
pw = pg.PlotWidget()
pw.show()

vals = np.hstack([rng.normal(size=500), rng.normal(size=260, loc=4)])
y,x = np.histogram(vals, bins=np.linspace(-3, 8, 40))

pen = pg.mkPen('w', width=6)
bgi = pg.BarGraphItem(x0=x[:-1], x1=x[1:], height=y, pen=pen, brush=(0,0,255,150))
pw.addItem(bgi)
rect = ObjectBounds(bgi)

pg.exec()
```

pijyoi:
If you have a hidpi screen, you will be able to see that the bounding rectangle has a small gap, compared to a device pixel ratio 1 screen. (This effect is more obvious with thicker line widths.)
This means that the pixel padding computation should by right be dependent on device pixel ratio. It's harmless to have a slightly over-sized bounding rectangle though.

SimenZhor:
I randomly encountered a need for this exact functionality in `PColorMeshItem` only to discover that you've put down some good work here already. For any potential future reference, the [`PlotDataItem`](https://github.com/pyqtgraph/pyqtgraph/blob/b2e6c729e6f89ca42b704c8f4f24ea32b6492906/pyqtgraph/graphicsItems/PlotDataItem.py#L1093) contains a good description of the `frac` and `orthoRange` functionality:

``` Python 
"""
Returns the range occupied by the data (along a specific axis) in this item.
This method is called by :class:`ViewBox` when auto-scaling.
=============== ====================================================================
**Arguments:**
ax              (0 or 1) the axis for which to return this item's data range
frac            (float 0.0-1.0) Specifies what fraction of the total data
                range to return. By default, the entire range is returned.
                This allows the :class:`ViewBox` to ignore large spikes in the data
                when auto-scaling.
orthoRange      ([min,max] or None) Specifies that only the data within the
                given range (orthogonal to *ax*) should me measured when
                returning the data range. (For example, a ViewBox might ask
                what is the y-range of all data with x-values between min
                and max)
=============== ====================================================================
"""
```
I had some trouble trying to figure out exactly what they were supposed to do from reading methods that didn't implement their respective functionality.

Another potentially useful resource is the `PlotCurveItem` which actually implements their functionality in its [`dataBounds` method](https://github.com/pyqtgraph/pyqtgraph/blob/b5390d40bd2c1f39606f5074893428a38a7f3427/pyqtgraph/graphicsItems/PlotCurveItem.py#L262).

Would it be possible to write some of this functionality in a generic way that can be reused across different classes? I think it would be useful for any item that uses `QPicture` for drawing, as the basic boundingRect functionality of directly converting the bounding rect of `QPicture` from `QRect` to `QRectF` can cause huge margins to appear for small x/y data.

pijyoi:
>Would it be possible to write some of this functionality in a generic way that can be reused across different classes?

I don't understand how `frac` and `orthoRange` are actually used by the library enough to implement it.

The scope that I hope to achieve here in this PR is to implement a simple enough replacement for current uses of `QPicture::boundingRect` such that it can be copied easily when implementing new custom `GraphicsObject`s. For instance, besides `BarGraphItem` and `PColorMeshItem`, `customGraphicsItem` and `NonUniformImage` also currently use `QPicture::boundingRect` to compute their bounding rectangle.

"Simple" here means that we don't handle things like non-finites or overflow.
In fact, it now seems to me that non-cosmetic pens would also fall into the category of edge cases.

pijyoi:
There is one example https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/examples/PlotAutoRange.py that exercises the `frac` argument.

There are only two `GraphicsObject`s, `PlotCurveItem` and `ScatterPlotItem`, which do handle the `frac` argument.

j9ac9k:
> One difference that I noticed between this implementation and https://github.com/pyqtgraph/pyqtgraph/pull/2561 is that this implementation takes less iterations to stabilize on Windows. (7 vs 3).

👀 

that's ...absolutely fantastic.

The script you added, here is the output on macOS  (this is with PyQt 6.4)

image

This PR looks great @pijyoi I'm 👍🏻 on merging ... curious if there are any other potential issues we would like to have a closer look, but I don't want to hold off on this PR to handle things like non-finite values and such...

pijyoi:
Slightly related:
#60 some notes on `dataBounds` by campagnola 
#1141 visual artifacts when `boundingRect` doesn't take pen width into account

j9ac9k:
hi @pijyoi thanks again for this PR, this looks good to me.  Thanks for diving into this tricky issue, I'm sure it wasn't easy to sort out!

PR #2566, opened 2022-12-28
Use base_prefix to detect virtual env

### LINKED ISSUE #1052: Multiprocess does not work in a venv

eendebakpt:
See the discussion in #1860.

Detect presence of a virtual enviroment using `sys.prefix != sys.base_prefix`. This is more robust than a check on the environment variable `VIRTUAL_ENV`.

Fixes #1052

@j9ac9k @pijyoi 

A minimal example (using Windows):

1. Create a virtual environment using virtualenv
2. Activate the environment
3. Run `python -c "import os; print(os.environ.get('VIRTUAL_ENV'))"`. The output shows the location of the virtualenv
4. Install Spyder (`pip install spyder`)
5. Start spyder (by typing the command `spyder`). In the Spyder console window run `import sys; print(sys.prefix)`. The output is the location of the virtual environment, which shows the console session in spyder is indeed running the virtual environment.
7. In the same console window of spyder run `import os; print(os.environ.get('VIRTUAL_ENV'))`. The output is `None`. So the original check does not work in this case
7. In the same console window run `import sys; sys.prefix != sys.base_prefix`. The output is `True`

### Other Tasks 

<details>

### Files that need updates
    
- [ ] `README.md`
- [ ] `setup.py`
- [ ] `tox.ini`
- [ ] `pyproject.toml`

</details>

<details>

### Pre Release Checklist

</details>

<details>

### Steps To Complete

- [ ] Announce on mail list
- [ ] Announce on Twitter

</details>

### DIFF
--- a/pyqtgraph/multiprocess/processes.py
+++ b/pyqtgraph/multiprocess/processes.py
@@ -136,7 +136,9 @@ class Process(RemoteEventHandler):
         # the multiprocessing connection. Technically, only the launched subprocess needs to
         # send its pid back. Practically, we hijack the ppid parameter to indicate to the
         # subprocess that pid exchange is needed.
-        xchg_pids = sys.platform == 'win32' and os.getenv('VIRTUAL_ENV') is not None
+        #
+        # We detect a virtual environment using sys.base_prefix, see https://docs.python.org/3/library/sys.html#sys.base_prefix
+        xchg_pids = sys.platform == 'win32' and sys.prefix != sys.base_prefix
 
         ## Send everything the remote process needs to start correctly
         data = dict(
--- a/pyqtgraph/multiprocess/processes.py
+++ b/pyqtgraph/multiprocess/processes.py
@@ -138,6 +138,7 @@ class Process(RemoteEventHandler):
         # subprocess that pid exchange is needed.
         #
         # We detect a virtual environment using sys.base_prefix, see https://docs.python.org/3/library/sys.html#sys.base_prefix
+        # See https://github.com/pyqtgraph/pyqtgraph/pull/2566 and https://github.com/spyder-ide/spyder/issues/20273
         xchg_pids = sys.platform == 'win32' and sys.prefix != sys.base_prefix
 
         ## Send everything the remote process needs to start correctly


pijyoi:
Is this not an issue with Spyder messing with the environment?
https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/ipythonconsole/utils/kernelspec.py#L173

pijyoi:
IPython makes use of the VIRTUAL_ENV environment variable and Spyder wants to fool IPython.
https://github.com/ipython/ipython/blob/27ca9e2e3f8c9220a8b92938e11157a77c2c4893/IPython/core/interactiveshell.py#L845

Maybe the fix should be at Spyder?

eendebakpt:
> IPython makes use of the VIRTUAL_ENV environment variable and Spyder wants to fool IPython. https://github.com/ipython/ipython/blob/27ca9e2e3f8c9220a8b92938e11157a77c2c4893/IPython/core/interactiveshell.py#L845
> 
> Maybe the fix should be at Spyder?

@pijyoi Thanks for the pointer, that gives good information. I do not know whether we should fix this at the spyder or pyqtgraph side. I created https://github.com/spyder-ide/spyder/issues/20273 to get input from the spyder devs.

The PR here is simple, so if there are no downsides and there is no easy solution at the spyder side I would be in favor of this option.

j9ac9k:
Hi @eendebakpt 

I'm ok merging this change, but I would request you add a comment with a link to the issue in spyder, so that future people maintaining the library don't have to scratch their heads as to why this change was made, and in turn can track if the other issue was ever fixed (and if it is fixed/addressed, when can this patch be removed from pyqtgraph).

eendebakpt:
> Hi @eendebakpt
> 
> I'm ok merging this change, but I would request you add a comment with a link to the issue in spyder, so that future people maintaining the library don't have to scratch their heads as to why this change was made, and in turn can track if the other issue was ever fixed (and if it is fixed/addressed, when can this patch be removed from pyqtgraph).
> 
> EDIT: btw thanks for creating the issue w/ the Spyder repo; hopefully there will be an explanation there, ccordoba12 (intentionally not tagging to minimize notification spam) is a friend of the project :)

@j9ac9k No response on the spyder github issue yet, but I added links to the issues in the comments.

j9ac9k:
> > Hi @eendebakpt
> > I'm ok merging this change, but I would request you add a comment with a link to the issue in spyder, so that future people maintaining the library don't have to scratch their heads as to why this change was made, and in turn can track if the other issue was ever fixed (and if it is fixed/addressed, when can this patch be removed from pyqtgraph).
> > EDIT: btw thanks for creating the issue w/ the Spyder repo; hopefully there will be an explanation there, ccordoba12 (intentionally not tagging to minimize notification spam) is a friend of the project :)
> 
> @j9ac9k No response on the spyder github issue yet, but I added links to the issues in the comments.

Thanks, that's perfect!

PR #2579, opened 2023-01-08
AxisItem-based grid item

### LINKED ISSUE #1606: showGrid screws up tick labels at axis boundaries
### Short description
When I use showGrid (current master and 0.11.1), then the tick labels at the corner of the axes rectangle are not drawn correctly.

### Code to reproduce
```python
import pyqtgraph as pg
import numpy as np

win.resize(800,350)
win.setWindowTitle('pyqtgraph example: Histogram')
plt1 = win.addPlot()
plt2 = win.addPlot()

## make interesting distribution of values
vals = np.hstack([np.random.normal(size=500), np.random.normal(size=260, loc=4)])

## compute standard histogram
y,x = np.histogram(vals, bins=np.linspace(-3, 8, 40))

## Using stepMode="center" causes the plot to draw two lines for each sample.
## notice that len(x) == len(y)+1
plt1.plot(x, y, stepMode="center", fillLevel=0, fillOutline=True, brush=(0,0,255,150))
plt2.plot(x, y, stepMode="center", fillLevel=0, fillOutline=True, brush=(0,0,255,150))

plt1.showGrid(x=False, y=False)
plt2.showGrid(x=True, y=True)

plt1.setRange(xRange=(5,10), yRange=(0,10), padding=0)
plt2.setRange(xRange=(5,10), yRange=(0,10), padding=0)


```

### Actual and Expected behavior
This is the output of the above script:
image

### Tested environment(s)

 * PyQtGraph version: master and 0.11.1
 * Qt Python binding: PyQt5 5.15.1 Qt 5.15.1
 * Python version: 3.8.5un
 * NumPy version: 1.18.5
 * Installation method: pip

swvanbuuren:
The idea is to display the grid lines in a `GridItem` and not extend the `AxisItem`s ticks.

This PR attempts to fix #1606, which reports that tick labels never seem to be displayed just right, when the grid is enabled.

Although the current PR seems to accomplish this, I am not quite happy with the current implementation:
- I had troubles replicating the way how `AxisItem` calculates tick positions inside a `GridItem`. Therefore I just added the `AxisItem` objects to the `GridItem` and get the positions by calling the appropriate methods. Not nice but it works.
- With this PR, the `GridItem` has been improved so that it gets the right tick positions from the corresponding `AxisItem`s. However, it needs to be clarified if this approach shall be adopted altogether by `AxisItem`, without the need for users to manually create a `GridItem`.

Therefore, I marked it as draft. I am happy to receive/discuss ideas how the current approach can be improved.

Recently, #2385 was introduced that circumvents this problem, by hiding tick labels, when overlap even slightly the axis boundaries. Thus, the changes of this PR are only relevant if the style option 'hideOverlappingLabels' of all `AxisItem`s are set to `False`.

I think it is useful, to have the possibility, to show tick labels at the plotted range start and/or end, when the tick label positions exactly match the start and/or end coordinate. Afaik matplotlib also does this.

Example to compare the behavior before and after
``` python
import pyqtgraph as pg
import numpy as np

win.resize(800,350)
win.setWindowTitle('pyqtgraph example: random points')
plt1 = win.addPlot()
plt2 = win.addPlot()

# create data
x = np.arange(100)
y = np.random.normal(size=100)

for ax in ('left', 'right', 'top', 'bottom'):
    plt1.getAxis(ax).setStyle(hideOverlappingLabels=False)
    plt2.getAxis(ax).setStyle(hideOverlappingLabels=False)

plt1.plot(x, y)
plt2.plot(x, y)

plt1.setXLink(plt2)
plt1.setYLink(plt2)

plt1.showAxis('right', show=True)
plt2.showAxis('right', show=True)
plt1.showAxis('top', show=True)
plt2.showAxis('top', show=True)

plt1.showGrid(x=True, y=True)

grid = pg.GridItem()
grid.setTextPen(None)
grid.setTickSpacing(x=[None, None], y=[None, None])
plt1.addItem(grid)

plt2.showGrid(x=True, y=True)

QtWidgets.QApplication.instance().exec_()
```


pyqtgraph/graphicsItems/PlotItem/PlotItem.py
```diff
@@ -551,6 +552,18 @@ def addItem(self, item, *args, **kargs):
         if name is not None and hasattr(self, 'legend') and self.legend is not None:
             self.legend.addItem(item, name=name)            
 
+        if isinstance(item, GridItem):
+            item.setAxes({ax['item'].orientation: ax['item'] for ax in self.axes.values()})
+            item.setCtrl(self.ctrl)
+
+    def addDataItem(self, item, *args):
```
j9ac9k, line 559: I think we already removed this method; probably should yank it out of this PR.

pyqtgraph/graphicsItems/GridItem.py
```diff
@@ -26,6 +26,30 @@ def __init__(self, pen='default', textPen='default'):
...
+
+
+    def setAxes(self, axes):
+        """ 
+        Set the axes object, in case the GridItem belongs to plot, whose axes are controlled by AxisItem's
+        """
+        self.axes = axes
+
+
+    def setCtrl(self, ctrl):
```
j9ac9k, line 40: nit-pick suggestion, how do you feel about renaming to `setControl`; I've been trying to avoid using abbreviations for methods/variables/functions in the public API.

pyqtgraph/graphicsItems/GridItem.py
```diff
@@ -26,6 +26,30 @@ def __init__(self, pen='default', textPen='default'):
         self.setPen(pen)
         self.setTextPen(textPen)
         self.setTickSpacing(x=[None, None, None], y=[None, None, None])
+        self.axes = None
+        self.ctrl = None
+
+
+    def setAxes(self, axes):
```
j9ac9k, line 33: if we're going to have a setter, to match the Qt API, we should have a getter method, `def axes(self)` which means we'll have to rename the attribute `self.axis` to `self._axis` or something like that.

pyqtgraph/graphicsItems/GridItem.py
j9ac9k: Also will need a getter method to get the content of the `ctrl` attribute

pyqtgraph/graphicsItems/PlotItem/PlotItem.py
swvanbuuren: Removed with commit e88d5ab

pyqtgraph/graphicsItems/GridItem.py
swvanbuuren: Method name `setCtrl` Changed to `setControl` and added the getter method for the `ctrl` attribute with commit e88d5ab

pyqtgraph/graphicsItems/GridItem.py
swvanbuuren: Also added the getter method `axes`, internal attribute changed to `self._axes` with commit e88d5ab

pyqtgraph/graphicsItems/GridItem.py
```diff
@@ -90,20 +114,71 @@
...
+        p.end()
+
+
+    def gridAlpha(self, orientation):
+        """ Determine the grid lines' alpha value """
+        if orientation in ('top', 'bottom'):
+            show_grid = self.ctrl.xGridCheck.isChecked()
+        elif orientation in ('left', 'right'):
+            show_grid = self.ctrl.yGridCheck.isChecked()
+        if show_grid:
```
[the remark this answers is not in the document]
swvanbuuren: Fixed in commit e88d5ab

j9ac9k:
Hi @swvanbuuren 

Sorry I've let this PR sit for so long, I'll try and review/test here in the coming days.

PR #2584, opened 2023-01-11
feat(PlotItem) define context menu action visibility

### LINKED ISSUE #2351: Disable actions in pg.PlotItem custom menu
### Short description
There is a way to disable custom menu of `pg.PlotItem` with `setMenuEnabled` but we need a way to disable only a set of available options.

I didn't find any way to disable only some items in the available options.

### Expected behavior
We would like to be able to define the available option.

For example:

```
# Don't show all  transform actions
plot_item.transformAction.setVisible(False)
```

For this we will need to store all created action in `pg.PlotItem`. I can create the PR for this.

For specific items in menu we can still use created widget:

`plot_item.ctrl.logYCheck.setVisible(False)`

jmkerloch:
Add method `setContextMenuActionVisible` in `PlotItem`.

Based on action `name` ('Transforms', 'Downsample', 'Average', 'Alpha', 'Grid', 'Points'), action is set visible depending on the argument `visible`

To avoid issue with translation (action text is translated) action name is translated to get correction action.

Fixes #2351

### DIFF
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -1083,6 +1083,23 @@ class PlotItem(GraphicsWidget):
     
     def menuEnabled(self):
         return self._menuEnabled
+
+    def setContextMenuActionVisible(self, name : str, visible : bool) -> None:
+        """
+        Changes the context menu action visibility
+
+        =============== ====================================================================
+        **Arguments:**
+        name            (str) Action name ('Transforms', 'Downsample', 'Average', 'Alpha', 'Grid', 'Points'
+
+        visible            (bool) If `True`, action is visible. If `False` action is invisible
+        =============== ====================================================================
+        """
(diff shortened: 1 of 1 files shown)


j9ac9k:
Hi @jmkerloch 

Thanks for the PR; this feature request has come up a few times, so I really appreciate the PR to implement this.

One nit-pick I have is to switch to a numpy doc style docstring; you can see another example in PlotItem here: https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/PlotItem/PlotItem.py#L1179C2-L1203

Also, how do you feel about accepting `str | Iterable[str]` where you can pass multiple action names.

I'm not sold on the method name, but I can't think of anything better, @outofculture @ixjlyons either of you two think of a better name here (or have other comments?)

jmkerloch:
@j9ac9k I updated the docstring. It seems OK to me in the generated documentation.

j9ac9k:
Thanks @jmkerloch This PR looks good to me, we appreciate your contribution!

PR #2586, opened 2023-01-14
PColorMeshItem: implement dataBounds and pixelPadding

pijyoi:
using `QPicture.boundingRect()` to compute `boundingRect()` has issues, which includes rounding the values to the next integral value.

Other issue fixed: `width()` and `height()` methods assume that the minimum coord is 0

Script that demonstrates the issues fixed by this PR.
- Coords are chosen to be small and the bounding box will be over-sized on `master`.
- print out of width() and height() are not correct on `master`.

```python
import pyqtgraph as pg
import numpy as np

class ObjectBounds(QtWidgets.QGraphicsItem):
        pen = QtGui.QPen(QtCore.Qt.GlobalColor.red, 0, QtCore.Qt.PenStyle.DashLine)
        rect = self.boundingRect()
        painter.setPen(pen)
        painter.drawRect(rect)
        # print(rect)

    def boundingRect(self):
        return self.parentItem().boundingRect()

pg.mkQApp()
pw = pg.PlotWidget()
pw.show()

rng = np.random.default_rng()
shape = (100, 100)
az = np.linspace(np.pi/6, np.pi*5/6, shape[0]+1)
rg = np.linspace(0.25, 1.25, shape[1]+1)
X = rg[:, np.newaxis] * np.cos(az)
Y = rg[:, np.newaxis] * np.sin(az)
Z = rng.uniform(size=shape)

pcmi = pg.PColorMeshItem(X, Y, Z)
pw.addItem(pcmi)
rect = ObjectBounds(pcmi)

print(pcmi.width(), pcmi.height())

pg.exec()
```

### DIFF
--- a/pyqtgraph/examples/PColorMeshItem.py
+++ b/pyqtgraph/examples/PColorMeshItem.py
@@ -67,6 +67,12 @@ wave_length     = 10
 color_speed     = 0.3
 color_noise_freq = 0.05
 
+# display info in top-right corner
+miny = np.min(y) - wave_amplitude
+maxy = np.max(y) + wave_amplitude
+view.setYRange(miny, maxy)
+textitem.setPos(np.max(x), maxy)
+
 timer = QtCore.QTimer()
 timer.setSingleShot(True)
 # not using QTimer.singleShot() because of persistence on PyQt. see PR #1605
@@ -91,11 +97,7 @@ def updateData():
 
     i += wave_speed
 
-    # display info in top-right corner
     textitem.setText(f'{(t2 - t1)*1000:.1f} ms')
-    if textpos is None:
-        textpos = pcmi.width(), pcmi.height()
-        textitem.setPos(*textpos)
 
     # cap update rate at fps
     delay = max(1000/fps - (t2 - t0), 0)
--- a/pyqtgraph/graphicsItems/PColorMeshItem.py
+++ b/pyqtgraph/graphicsItems/PColorMeshItem.py
@@ -12,12 +12,6 @@ from .GraphicsObject import GraphicsObject
 __all__ = ['PColorMeshItem']
 
 
-if Qt.QT_LIB.startswith('PyQt'):
-    wrapinstance = Qt.sip.wrapinstance
-else:
-    wrapinstance = Qt.shiboken.wrapInstance
-
-
 class QuadInstances:
     def __init__(self):
         self.polys = []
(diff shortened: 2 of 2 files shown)


pijyoi:
Looking at the very first commit of the PR #1273 for `PColorMeshItem`, the original implementations of `width()`, `height()` and `boundingRect()` were modeled on the implementation for `ImageItem`, which draws the image at the origin. For `ImageItem`, the user applies scale and translation transformations to "move" it to the correct spot.

As `PColorMeshItem` can be drawn with arbitrary coordinates, this model is not appropriate.
A couple of questions:
1) Are `width()` and `height()` methods even necessary?
2) And if yes, why return `None` instead of just zero? (`ImageItem` returns `None`, but is there a need to copy `ImageItem`'s interface?)

```python
    def width(self):
        if self.x is None:
            return None
        return len(self.x)

    def height(self):
        if self.y is None:
            return None
        return len(self.y)

    def boundingRect(self):

        if self.z is None:
            return QtCore.QRectF(0., 0., 0., 0.)
        return QtCore.QRectF(0., 0., float(self.width()), float(self.height()))
```

pijyoi:
Would it make sense to force/require the edgecolor pen to be cosmetic? It would simplify the `dataBounds` method.
It doesn't seem to make sense to draw the polygon edges with a non-cosmetic pen.

j9ac9k:
Hi @pijyoi sorry for the late response.

I don't think the `width` and `height` methods are necessary.  I can see the case where having thick lines for the edge-colors would be desired; but I'm not sure it's worth it to support that functionality at the expense of complexity.  I would be good with supporting only cosmetic pens.  Should add a note about that in the docstring, and maybe put a comment in the code referencing a post here about not supporting non-cosmetic pens.

pijyoi:
>I can see the case where having thick lines for the edge-colors would be desired; but I'm not sure it's worth it to support that functionality at the expense of complexity. I would be good with supporting only cosmetic pens.

It's okay to support thick cosmetic pens. That code is within `pixelPadding`. Whether the pen is thick or not, half the pen width still needs to be accounted for.

The code that supports non-cosmetic pens is within `dataBounds`, and is the code that I would like to remove.

j9ac9k:
> The code that supports non-cosmetic pens is within `dataBounds`, and is the code that I would like to remove.

I'm good with that.

j9ac9k:
Thanks @pijyoi for patching up PColorMeshItem!  Changes LGTM, merging!

PR #2587, opened 2023-01-14
NonUniformImage: implement floating point boundingRect

pijyoi:
using QPicture.boundingRect() to compute boundingRect() has issues, which includes rounding the values to the next integral value.

The data bounds are easily computed. The bulk of the code deals with handling the optional border rectangle.

### DIFF
--- a/pyqtgraph/graphicsItems/NonUniformImage.py
+++ b/pyqtgraph/graphicsItems/NonUniformImage.py
@@ -131,4 +131,5 @@ class NonUniformImage(GraphicsObject):
         p.drawPicture(0, 0, self.picture)
 
     def boundingRect(self):
-        return QtCore.QRectF(self.picture.boundingRect())
+        x, y, _ = self.data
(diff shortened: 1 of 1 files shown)


pijyoi:
Note: `ImageItem`'s `boundingRect()` does not take into account the pen width of the optional border. Probably the border is not often used.

pijyoi:
I am reducing the scope of this PR. Only `boundingRect()` will be implemented in terms of the data coordinates.

We can cross the bridge of the border pen width when we get there.
Given that `ImageItem`:
- also has this issue
- is much more used
- has no issue being raised about it

there's no point complicating this PR at this point of time.

A simple way to test the effects of this PR is by modifying `examples/NonUniformImage.py`. By scaling the variables `tau` and `w` to be less than 1, you will be able to see that `master` does not range the item properly.

j9ac9k:
Hi @pijyoi 

Thanks for the PR, I'm good w/ limiting the scope.  LGTM, merging!

PR #2589, opened 2023-01-15
typo: dataRange -> dataBounds

pijyoi:
This fix should make it easier for someone to find out that `enableAutoRange()` is the way to exercise the `dataBounds()` method.


j9ac9k:
Nice catch @pijyoi !

Issue #2287, opened 2022-05-04
Is there any way to draw great number of rectangles in the viewbox?


earphonebreaker:
I made a graphic viewer by using pyqtgraph. The code logic is like this:

Graphic Viewer
```
pg = GraphicsLayoutWidget()
vb = pg.addViewBox()
vb.addItem(RectItem)
```
The RectItem is a Class that I copied and modified from other's code:
```
class RectItem(pg.GraphicsObject):
    def __init__(self, rect,line,brush, parent=None,line_width=1):
        super().__init__(parent)
        self._rect = rect
        self.picture = QtGui.QPicture()
        self._generate_picture(line,brush,line_width)

    def _generate_picture(self,line,brush,line_width):
        painter = QtGui.QPainter(self.picture)
        painter.setPen(pg.mkPen(line,width=line_width))
        painter.setBrush(pg.mkBrush(brush))
        painter.drawRect(self._rect)
        painter.end()

    def paint(self, painter, option, widget=None):
        painter.drawPicture(0, 0, self.picture)

    def boundingRect(self):
        return QtCore.QRectF(self.picture.boundingRect())
```
My problem is, when there are only several RectItem added into the viewbox, the program works perfectly fine, but when great numbers of RectItem(i.e. more than 100000) are added, the program will be extremely slow and even crash. I also tried to draw them in one item by modifying the RectItem class into this:
```
class GraphLayerItem(pg.GraphicsObject):
    def __init__(self, rect_list, line,brush,parent=None):
        super().__init__(parent)
        self.rect_list = rect_list
        self.line = line
        self.brush = brush
        self.picture = QtGui.QPicture()
        self._generate_picture()


    def _generate_picture(self):
        painter = QtGui.QPainter(self.picture)
        painter.setPen(pg.mkPen(self.line))
        painter.setBrush(pg.mkBrush(self.brush))
        for rect in self.rect_list:
            painter.drawRect(rect)
        painter.end()

    def paint(self, painter, option, widget=None):
        painter.drawPicture(0, 0, self.picture)

    def boundingRect(self):
        return QtCore.QRectF(self.picture.boundingRect())
```
However, this doesn't help either when the rect_list is very large. Is there any way to change the display options, such as changing dpi while zooming in and out, or making a filter when there are too many items displayed in current viewbox range? Or any other ways to fix this problem? Thanks!

j9ac9k:
Hi @earphonebreaker 

[...]

PR #2591, opened 2023-01-24
create internals.PrimitiveArray

pijyoi:
Refactor the array code in `ScatterPlotItem` and `PlotCurveItem` into PrimitiveArray.

It may look as though that there is one behavior change made in this PR. Previously in `ScatterPlotItem`, if the number of items was zero, then the call to `drawPixmapFragments` would be skipped.
I have looked through the logs, and this skipping behavior was added when adding support for PyQt's `sip.array`. Testing shows that PyQt does currently support zero-length sip arrays. I suspect it was added when testing against preliminary versions of `sip.array`.

Script to verify that zero-length `sip.array`(s) work.
```python
from PyQt6 import QtCore, QtGui, sip

app = QtGui.QGuiApplication([])
array = sip.array(QtGui.QPainter.PixmapFragment, 0)
print(array)

pix = QtGui.QPixmap(51, 51)
pix.fill(QtCore.Qt.GlobalColor.cyan)

qimg = QtGui.QImage(200, 200, QtGui.QImage.Format.Format_RGB32)
qimg.fill(QtCore.Qt.GlobalColor.white)
painter = QtGui.QPainter(qimg)
painter.drawPixmapFragments(array, pix)
painter.end()

print('done')
```

Besides cleaning up the code, this PR will hopefully make it simpler to add support for `drawRects` and `drawPoints`.

Sample script to exercise `drawRects` using this PR. This may help #2287. 
```python
import numpy as np
import pyqtgraph as pg

class CheckerBoardItem(pg.GraphicsObject):
    def __init__(self, size):
        super().__init__()
        self.size = size

        x = np.arange(self.size)
        y = np.arange(self.size)[:, np.newaxis]
        bitmap = (y & 1) ^ (x & 1)
        X, Y = np.meshgrid(x, y)
        nzidx = bitmap.nonzero()

        self.array = pg.Qt.internals.PrimitiveArray(QtCore.QRectF, 4)
        self.array.resize(nzidx[0].size)
        memory = self.array.ndarray()
        memory[:, 0] = X[nzidx]
        memory[:, 1] = Y[nzidx]
        memory[:, 2:4] = 1

        rect = QtCore.QRectF(0, 0, self.size, self.size)
        painter.fillRect(rect, QtCore.Qt.GlobalColor.blue)

        inst = self.array.instances()
        painter.setPen(QtCore.Qt.PenStyle.NoPen)
        painter.setBrush(QtCore.Qt.GlobalColor.white)
        painter.drawRects(inst)

    def boundingRect(self):
        return QtCore.QRectF(0, 0, self.size, self.size)

pg.mkQApp()
pw = pg.PlotWidget()
pw.show()
cbi = CheckerBoardItem(32)
pw.addItem(cbi)
pg.exec()
```

### DIFF
--- a/pyqtgraph/Qt/internals.py
+++ b/pyqtgraph/Qt/internals.py
@@ -1,9 +1,13 @@
 import ctypes
+import itertools
 import numpy as np
-from . import QT_LIB
+from . import QT_LIB, QtCore, QtGui
 from . import compat
 
 __all__ = ["get_qpainterpath_element_array"]
+
+if QT_LIB.startswith('PyQt'):
+    from . import sip
 
 class QArrayDataQt5(ctypes.Structure):
     _fields_ = [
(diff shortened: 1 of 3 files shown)


j9ac9k:
Hi @pijyoi I love this PR.   Migrating this complexity from the GraphicsItems themselves to Qt.internals was absolutely the right thing to do.  Abstracting PrimativeArray to work with all bindings, regardless of whether they support the lower-level methods is fantastic.

My one nit-pick is that we should probably stay away from descriptions like "newer" and "older" regarding the versions of Qt bindings and just specify which version sip.array was introduced explicitly (or I suppose, more specifically `PyQt[56]-sip`.  Hopefully that PR in pyside will be merged soon and we'll get equivalent performance improvements from those bindings as well.

pijyoi:
I used more general language in the comments and let the code document itself as to the exact versions required.

j9ac9k:
> I used more general language in the comments and let the code document itself as to the exact versions required.

Saw that, but before you know it, 10 years will have passed :D 

If you'd rather leave it as is, that's fine with me.  This class has fantastic comments as it is.  Merging

PR #2596, opened 2023-01-28
Prepare support for PySide6 drawLines and friends

pijyoi:
As PYSIDE-1924 (https://bugreports.qt.io/browse/PYSIDE-1924) feature has been added, we can prepare pyqtgraph to make use of it when it gets into a released version.
For now, testing can be done with the wheels at https://download.qt.io/snapshots/ci/pyside/dev/latest/

One issue is that PySide6's array API implementation will requirement passing of different arguments than the other codepaths. For this, we add a `drawargs()` method.

A synthetic benchmark to exercise the new API.
```python
import numpy as np
import pyqtgraph as pg
from pyqtgraph.Qt import QtCore, QtGui

from time import perf_counter

def generate_pattern(nsegs, size):
    rng = np.random.default_rng()
    x = rng.random(nsegs) * size
    y = rng.random(nsegs) * size
    arr = np.empty((nsegs, 4), dtype=np.float64)
    arr[:, 0] = x
    arr[:, 1] = y
    arr[:, 2] = x + 2
    arr[:, 3] = y + 2
    return arr

nsegs = 10_000
size = 500
nframes = 100
pattern = generate_pattern(nsegs, size)

def run(use_array):
    # generate lines once
    segments = pg.Qt.internals.PrimitiveArray(QtCore.QLineF, 4, use_array=use_array)
    segments.resize(nsegs)
    memory = segments.ndarray()
    memory[:] = pattern

    # draw multiple frames using the same lines array
    t0 = perf_counter()
    for _ in range(nframes):
        qimg = QtGui.QImage(size, size, QtGui.QImage.Format.Format_RGB32)
        qimg.fill(QtCore.Qt.GlobalColor.transparent)
        painter = QtGui.QPainter(qimg)
        painter.setPen(QtCore.Qt.GlobalColor.cyan)
        painter.drawLines(*segments.drawargs())
        painter.end()
    t1 = perf_counter()
    return t1 - t0

for use_array in [False, None, True]:
    duration = run(use_array)
    fps = int(nframes / duration)
    print(f'{use_array=} {duration=:.3f} {fps=}')
```

PR #2609, opened 2023-02-10
rename "method" to "use_array" and make it keyword only

pijyoi:
Even though PrimitiveArray is not public facing API, its `__init__` argument "`method : bool`" was not named correctly.

This rename was originally meant to be part of #2596, but as #2596 needs more testing, the rename is extracted out to get it in before the next release.


j9ac9k:
@pijyoi I really appreciate your work on getting these lower-level interfaces working!

PR #2625, opened 2023-02-24
don't use pg.plot() in tests

pijyoi:
There seems to be a new recurring CI segfault on macOS / Python 3.8 / PySide2 5.15.

This PR seems to help to prevent it from happening.

`pg.plot()` and `pg.image()` will keep a permanent global reference to their created `PlotWidget` and `ImageView`.
This may be a convenience for interactive use, but not suitable for CI where a great many plots are instantiated.

### DIFF
--- a/tests/exporters/test_csv.py
+++ b/tests/exporters/test_csv.py
@@ -17,7 +17,8 @@ def approxeq(a, b):
 
 
 def test_CSVExporter():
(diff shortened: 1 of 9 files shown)


j9ac9k:
Thanks @pijyoi for this PR, always love it when another cause for segfaults in CI is removed!

Issue #2642, opened 2023-03-07
Adding a `TextItem` to a `PlotItem` with no other data breaks auto-scaling


redruin1:
### Short description
I'm attempting to add text labels to a standard `PlotItem` widget. In my production code however, I neglected to add any data before I added the text (I was adding ROIs as "data"), and it resulted in both the x and y scales to trend towards 0 very quickly when auto-scaling is on.

### MinRe Example
```python
import pyqtgraph as pg

if __name__ == "__main__":
    pg.mkQApp()

    pw = pg.PlotWidget()
    pw.show()

    # Having a plot with no data works fine, until you add the TextItem:
    text_item = pg.TextItem("Some text")
    pw.addItem(text_item)

    # Including either (or both) of the following lines produce expected behavior:
    #pw.plot([1, 2, 3], [4, 5, 6])
    #pw.getPlotItem().getViewBox().disableAutoRange()

    pg.exec()
```

### Expected behavior
Either the standard graph display centered around the text "Some text" located at the origin, or the standard graph centered around the range [0, 1] (which when panned to the origin reveals the aforementioned text).

### Real behavior
For the first few seconds of the display, the scale factors are seen changing rapidly every frame; upon close inspection, they can be seen shrinking until they approach 0:

![image](https://user-images.githubusercontent.com/25079873/223302091-e8eb15c5-491b-4709-b152-982506d119d4.png)

Eventually the graph markers break completely as the x and y scales sit precariously above 0, as revealed by the context menu:

![image](https://user-images.githubusercontent.com/25079873/223303232-11256e8d-9cc6-412f-bd45-13e72ddb29fd.png)

Switching the axis modes from Auto to Manual and setting an appropriate range provides the expected outcome:

![image](https://user-images.githubusercontent.com/25079873/223303515-5b0a40aa-9c21-4d96-bdf2-677a45190808.png)

### Tested environment

 * PyQtGraph version: `0.13.2`
 * Qt Python binding: `'PyQt5 5.15.9 Qt 5.15.2'`
 * Python version: `3.10.0`
 * NumPy version: `1.24.2`
 * Operating system: Windows 10
 * Installation method: `pip install pyqtgraph`

j9ac9k:
Thanks for the report @redruin1 !

[...]

PR #2646, opened 2023-03-09
Add `dataBounds()` method to `TextItem` to make autoRange more predictable

redruin1:
### Summary

Sacrifices some visual polish to make the autorange functionality with TextItems much more predictable. "Fixes" #2642, though theoretically with some more effort it should be possible to still get high visual fidelity without runaway scaling.

### Details

Adds the `dataBounds()` method to `TextItem`, which supersedes the result from `boundingRect()` when updating the ViewBox's autorange. Instead, only the anchor point is returned, behaving similarly to how a set of data points on a graph would be scaled to. 

### Benefits
This PR fixes:

* Runaway scaling when there is only one `TextItem` in a `PlotWidget`
  <details>
  <summary>Example</summary>
  <blockquote>
  import pyqtgraph as pg

  pg.mkQApp()
  pw = pg.PlotWidget()
  
  text_item = pg.TextItem("Any text")
  pw.addItem(text_item)
  
  pw.show()
  pg.exec()
  </blockquote>
  </details>
* Runaway scaling when there is multiple `TextItem`s in a `PlotWidget`, but they all share the same axis
  <details>
  <summary>Example</summary>
  <blockquote>
  import pyqtgraph as pg

  pg.mkQApp()
  pw = pg.PlotWidget()
  
  text_item = pg.TextItem("Any text")
  pw.addItem(text_item)
  
  text_item = pg.TextItem("Some other text")
  text_item.setPos(10, 0)
  pw.addItem(text_item)
  
  pw.show()
  pg.exec()
  </blockquote>
  </details>
* Runaway scaling when the visual length of a `TextItem` cannot entirely fit within a `ViewBox` at the current resolution/font size
  <details>
  <summary>Example</summary>
  <blockquote>
  import pyqtgraph as pg

  pg.mkQApp()
  pw = pg.PlotWidget()
  
  pw.plot([1, 2, 3], [4, 5, 6])

  text = 'A' * 1000
  text_item = pg.TextItem(text)
  pw.addItem(text_item)
  
  pw.show()
  pg.exec()
  </blockquote>
  </details>
* Runaway scaling with `TextItem` as a whole should be largely eliminated 

### Downsides

* TextItems are no longer guaranteed to be entirely (or even partially) visible, which reduces the visual quality under non-edge cases:

image

### Files that need updates

<details>

- [x] `README.md` (None needed)
- [x] `setup.py` (None needed)
- [x] `tox.ini` (None needed)
- [x] `.github/workflows/main.yml` and associated `requirements.txt` and conda `environemt.yml` files (None needed)
- [x] `pyproject.toml` (None needed)
- [x] `binder/requirements.txt` (None needed)

</details>

<details>

### Pre Release Checklist

</details>

<details>

### Steps To Complete

- [ ] Announce on mail list
- [ ] Announce on Twitter

</details>


campagnola:
Looks good to me! I am ok with changing the default behavior here, but can you also add an __init__ argument to disable dataBounds and restore the original behavior?

redruin1:
@campagnola Sure, no problem. Any preferences for what that init argument would be called? (So I can be in line with the rest of the API)

j9ac9k:
Hi @redruin1 

First off, I owe you an apology for letting this PR sit for so long without a response, that wasn't fair to you.  This is a _very_ high quality PR that does a fantastic job highlighting the issue, the fix and potential downside.  Thank you for that.

I'm also good w/ changing the default behavior; I'm not sure we need to keep an argument to preserve legacy behavior tho (but @campagnola being the original author outranks me :D ).

I don't think there is an argument in pyqtgraph that we can use for to maintain consistency, but there might be something in the Qt API.  Scrolling through the Qt API looks like perhaps a flag is what we would want to add, and while this topic has come up before, we've never actually implemented out our enums that can be used as flags. Maybe `ensureInViewport` with default to False, and in the docstring referencing this PR (or associated issue).

redruin1:
Alright, this was the least intrusive way to get the desired behavior that I could come up with. Ideally there would probably be some inheritance shuffling to enable/disable the `dataBounds` method properly, but for now all I've done is conditionally set the attribute to `None` in `TextItem.__init__()`:
```python
    if ensureInBounds:
        self.dataBounds = None # Doesn't like del or delattr here, which is probably for the best
```
The only other change (aside from docs) is changing the criteria in `ViewBox`:
```python
if hasattr(item, 'dataBounds') and item.dataBounds is not None
    ...
```
This could maybe be more strict in checking that `item.dataBounds` is a `Callable` object, but I doubt this will be used frequently elsewhere. Tests pass on my machine.

I didn't make helper methods for setting/modifying `ensureInBounds` as it would be complex to maintain and ensure that it works properly, and it seems more proper as just a init parameter. That being said, would you like it to be an accessible attribute so that users can query whether or not the `TextItem` is guaranteed to be fully visible?

redruin1:
Hmm, seems like this interferes with the OpenGL glGetError wrapper function also not expecting `None` in place of a method.

j9ac9k:
I think you're on the right track, let me take a look later today and I can see if I come up with something.

j9ac9k:
oh, the glError in read the docs CI is entirely unrelated, if you rebase `master`, that should go away.

The errors on python 3.11 with PyQt6 are due to a cpython bug here: https://github.com/python/cpython/issues/105497

j9ac9k:
should the file `to_delete.py` be removed from the PR?

redruin1:
whoops

j9ac9k:
Thanks for the PR @redruin1 and thanks for your patience with a review.  Merging!

PR #2648, opened 2023-03-12
Return float values from QColor in getByIndex

### LINKED ISSUE #2627: ColorMap.getByIndex() returns wrong colors
### Short description

ColorMap.getByIndex() returns wrong colors.

### Code to reproduce

```python
In [2]: cm = pg.ColorMap([0.0, 1.0], [(0,0,0), (255,0,0)])
In [3]: cm.getByIndex(0)
Out[3]: PySide6.QtGui.QColor.fromRgbF(0.000000, 0.000000, 0.000000, 0.003922)
In [4]: cm.getByIndex(1)
Out[4]: PySide6.QtGui.QColor.fromRgbF(0.003922, 0.000000, 0.000000, 0.003922)
```

### Tested environment(s)

 * PyQtGraph version: 0.13.2.dev0

nickdimitroff:
Make `ColorMap_getByIndex` return floats for RGB values to fit expectations.

Fixes #2627

### DIFF
--- a/tests/test_colormap.py
+++ b/tests/test_colormap.py
@@ -1,3 +1,4 @@
+import PySide6
 import pytest
 
 import pyqtgraph as pg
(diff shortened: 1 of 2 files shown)


tests/test_colormap.py
```diff
@@ -1,3 +1,4 @@
+import PySide6
```
j9ac9k, line 1: Since pyqtgraph attempts to be binding-agnostic, we should not import specific bindings outside of rare places like `pyqtgraph/Qt/__init__.py`.

In this case, we already have access to the QtGui module from an import a few lines down.

```python
```

So you can delete the PySide6 import, and in your test code instead of saying `PySide6.QtGui.QColor`, you can say `QtGui.QColor`

j9ac9k:
Hi @nickdimitroff 

Thanks for the PR!  I added one comment regarding the imports in the test code.  Once you make that change, I'll approve and run on the CI system 👍🏻

tests/test_colormap.py
nickdimitroff: Thank you, that's good to know! Updated :+1:

j9ac9k:
Thanks @nickdimitroff this LGTM! Merging!

Issue #2650, opened 2023-03-14
GLViewWidget.itemsAt() returns items outside region since Qt 6.4


pijyoi:
### Short description

Since Qt 6.4, GLViewWidget.itemsAt() no longer returns only the items within the region and seemingly returns all items.

### Code to reproduce

```python
import pyqtgraph as pg
import pyqtgraph.opengl as gl
import numpy as np

class MyGLV(gl.GLViewWidget):
    def mouseReleaseEvent(self, ev):
        lpos = ev.position() if hasattr(ev, 'position') else ev.localPos()
        region = [lpos.x()-5, lpos.y()-5, 10, 10]
        # itemsAt seems to take in device pixels
        dpr = self.devicePixelRatioF()
        region = tuple([x * dpr for x in region])
        for item in self.itemsAt(region):
            print(item.objectName())

pg.mkQApp()

glv = MyGLV()
glv.setCameraParams(elevation=90, azimuth=-90, distance=50)
# X points right, Y points up
glv.show()

side = 8
names = ['red', 'green', 'blue']
for idx, name in enumerate(names):
    box = np.zeros((side, side, 4), dtype=np.uint8)
    box[..., [idx, 3]] = 255
    img = gl.GLImageItem(box)
    img.setObjectName(name)
    img.translate(-side/2, -side/2, 0)  # center the box
    img.translate((idx-1)*side*2, (idx-1)*side*1, 0)
    glv.addItem(img)

pg.exec()
```

### Expected behavior

Clicking on any box should result in the print out of its color.

### Real behavior

Clicking on any box results in print out of all three colors.

### Tested environment(s)

 * PyQtGraph version: 0.13.2 
 * Qt Python binding: Qt 6.4.x PySide6 or PyQt6 
 * Python version: 3.10
 * NumPy version: NA 
 * Operating system: Windows 11 
 * Installation method: pip install -e .

pijyoi:
It seems that using a `QOpenGLWindow` would restore some of the functionality that no longer works in `QOpenGLWidget`.
Included below is a patch that would introduce a `GLViewWindow` (modified and rebased from #2063).
By using a `GLViewWindow`,
1) `itemsAt()` seems to work again.
2) swapBuffers() also works

```python
import pyqtgraph as pg
from pyqtgraph.Qt import QtWidgets
import pyqtgraph.opengl as gl
from pyqtgraph.opengl.GLViewWidget import GLViewWindow
import numpy as np

[...]

PR #2653, opened 2023-03-17
GLViewWidget: don't assume mouse tracking is disabled

pijyoi:
In `mouseMoveEvent`, it is assumed that `self.mousePos` has already been set. This amounts to assuming that there is at least one `mousePressEvent` prior to `mouseMoveEvent`.

This amounts to assuming that mouse tracking is left at its default disabled state.

This PR removes this assumption and simplifies the code by removing the `mousePressEvent` handler altogether.

Relevant documentation: https://doc.qt.io/qt-6/qwidget.html#mouseMoveEvent
"If mouse tracking is switched off, mouse move events only occur if a mouse button is pressed while the mouse is being moved. If mouse tracking is switched on, mouse move events occur even if no mouse button is pressed."

This is partially related to the `GLViewWindow` implementation in #2650. In a `QOpenGLWindow`, there is no mouse tracking enable/disable.

### DIFF
--- a/pyqtgraph/opengl/GLViewWidget.py
+++ b/pyqtgraph/opengl/GLViewWidget.py
@@ -418,12 +418,10 @@ class GLViewWidget(QtWidgets.QOpenGLWidget):
         xDist = dist * 2. * tan(0.5 * radians(self.opts['fov']))
         return xDist / self.width()
         
-    def mousePressEvent(self, ev):
-        lpos = ev.position() if hasattr(ev, 'position') else ev.localPos()
-        self.mousePos = lpos
-        
     def mouseMoveEvent(self, ev):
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks @pijyoi always love to see the code cleaned up and unnecessary methods removed.

l-spiecker:
I now had to set the following to not jump between mouse clicks and keep the behaviour the a same as before.

```
w = gl.GLViewWidget()
w.setMouseTracking(True)
```

(Using pyqt5)

pijyoi:
Hi @l-spiecker , thank you for the bug report and sorry for having caused the bug!
I have tried to fix it in #2698. Would you be able to test it out?

l-spiecker:
Nice, works now again. Thanks!

PR #2654, opened 2023-03-17
unify ndarray_from_qpolygonf implementation

pijyoi:
With a better understanding of sip and shiboken, the implementation for `ndarray_from_qpolygonf` can be unified.


j9ac9k:
love the diff @pijyoi thank you! merging

PR #2662, opened 2023-03-25
inherit GraphicsWidgetAnchor on the left-hand-side

pijyoi:
Following the merge of https://code.qt.io/cgit/pyside/pyside-setup.git/commit/?id=e8095467f7d0332cc0987e7c541de9906e19fece slated for PySide6 6.5, inheriting `GraphicsWidgetAnchor` on the right-hand-side no longer works.

In fact, since PyQt{5,6} do implement cooperative inheritance, then the previous implementation of inheriting `GraphicsWidgetAnchor` on the right-hand-side would be wrong. i.e. `GraphicsWidgetAnchor::__init__` was being called twice on PyQt{5,6}. (Indeed, adding simple print statements shows that it was being called twice on PyQt{5,6}
There is a `connect` statement in `GraphicsWidgetAnchor::__init__`, which would make calling it multiple times *not* harmless.

Note: `GraphicsWidgetAnchor` as a mixin should be inherited on the left-hand-side.

### DIFF
--- a/pyqtgraph/graphicsItems/LabelItem.py
+++ b/pyqtgraph/graphicsItems/LabelItem.py
@@ -6,7 +6,7 @@ from .GraphicsWidgetAnchor import GraphicsWidgetAnchor
 
 __all__ = ['LabelItem']
 
-class LabelItem(GraphicsWidget, GraphicsWidgetAnchor):
+class LabelItem(GraphicsWidgetAnchor, GraphicsWidget):
     """
     GraphicsWidget displaying text.
     Used mainly as axis labels, titles, etc.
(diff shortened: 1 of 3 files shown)


j9ac9k:
glad this didn't have other consequences like with the docks!

Thanks for identifying this issue and submitting a fix @pijyoi merging!

pijyoi:
Was contemplating whether we should disinherit `pg.LabelItem` from `GraphicsWidgetAnchor`, as the library itself doesn't make use of this feature.
However, a search turned up code in the wild making use of this feature:
https://github.com/LilElm/python-flying-balls/blob/master/src/gui.py

Issue #1621, opened 2021-03-05
stepMode arguments not working properly 


Cademan10:
Using stepMode=True in a PlotDataItem creates a histogram that associates a given y-value with the left boundary, and using the other string commands ("center", "left", or "right") raises an error regarding array sizes. 

(Tested on Windows 10 using Python 3.7.9 and pyqtgraph  0.11.0)

Cademan10:
The issue was resolved by simply updating to the latest version of PyQtGraph

j9ac9k:
Thanks for following up @Cademan10 

Hopefully this will become less of an issue as we move to a quicker/smaller release cycle.

PR #2664, opened 2023-03-26
Add style sheet in pyqtgraph

edumur:
Dear all,

I thought it will be a nice feature to have style sheet, in the "same" way as matplotlib.
Here, I have just implemented a **very simple** version of that feature in hope to have feedback before going further.
Currently nothing is changed (and normally nothing is crashing) except all style options are now on a dedicated `default` style sheet.

### How it works

When pyqtgraph is loaded, the default style is loaded from the new `core.style` file.
All the style options are stored in `configStyle` which is a dictionnary.
Throughout the lib, default style options are taken from that dictionnary.

To use other styles, use `core.use` and load another style file.

### Some remarks
- There is currently two options dict, the config and the style one. There are not access the same way. We should decide one one way and use that everywhere.
- Is it possible to have different item with and without aliasing?
- I am not at all a python or qt wizards so if something looks weirds it could be because I messed up somewhere. Please tell me if that's the case!
- current extension is `pstyle`... Do we keep it?

### Tasks

Item to go through and transfer to the new styling paradigms:
- [x] LabelItem
- [x] LegendlItem
- [x] AxisItem
- [x] GridItem
- [x] GraphItem
- [x] PlotCurveItem
- [x] PlotDataItem
- [x] BarItem
- [x] ErrorBarItem
- [x] ScaleBarItem
- [x] ScatterPlotItem

What I plan to do if we decide to move forwards
- [ ] Add more style options in the stylesheet (size, color, font, ...).
- [x] Make an official "white" version of pyqtgraph through a second style sheet.
- [ ] Make an example to demonstrate the use of the feature.
- [ ] Add test

To be checked:
- [ ] Working with pyplotter
- [ ] Working with orange

Example

White official style
image

### Breaking change

Currently None.

### Other Tasks 

<details>

### Files that need updates
I confirm that I didn't touch that yet.
- [ ] `README.md`
- [ ] `setup.py`
- [ ] `tox.ini`
- [ ] `pyproject.toml`

</details>

<details>

### Pre Release Checklist

I confirm that I didn't touch that yet.

</details>

<details>

### Steps To Complete

I confirm that I didn't touch that yet.
- [ ] Announce on mail list
- [ ] Announce on Twitter

</details>

### DIFF
--- a/pyqtgraph/__init__.py
+++ b/pyqtgraph/__init__.py
@@ -10,6 +10,7 @@ __version__ = '0.13.4.dev0'
 import importlib
 import os
 import sys
+import warnings
 
 import numpy  # # pyqtgraph requires numpy
 
@@ -25,6 +26,18 @@ from .Qt import mkQApp
     #app = QtWidgets.QApplication([])
 
               ## (import here to avoid massive error dump later on if numpy is not available)
+
+
+
+### Set style options
+## All options related to style are set here
+## Style options must be initialized before the config options
+## Init default style
+from .style.core import (
+    loadDefaultStyle,
+    setConfigStyle
+)
+configStyle = loadDefaultStyle()
 
 
 
@@ -60,8 +73,20 @@ CONFIG_OPTIONS = {
                                   # 'off' or False: lines are never plotted in segments
 }
 
-
 def setConfigOption(opt, value):
+
+    # We keep old style options available with deprecation warning
+    if opt in ('foreground', 'background', 'antialias'):
+        warnings.warn('Setting style option "{}" through setConfigOptions is deprecated. Use setConfigStyle instead.',
+                       DeprecationWarning,
+                       stacklevel=2)
+        if opt in ('background', 'antialias'):
+            setConfigStyle('GraphItem', opt, value)
+        else:
+            setConfigStyle('GraphItem', 'lineColor', value)
+        return
+
+
     if opt not in CONFIG_OPTIONS:
         raise KeyError('Unknown configuration option "%s"' % opt)
     if opt == 'imageAxisOrder' and value not in ('row-major', 'col-major'):
@@ -82,18 +107,6 @@ def getConfigOption(opt):
     """Return the value of a single global configuration option.
     """
     return CONFIG_OPTIONS[opt]
-
-
-
-
-
-### Set style options
-## All options related to style are set here
-## Init default style
-from .style.core import loadDefaultStyle
-configStyle = loadDefaultStyle()
-
-
 
 
 
--- a/pyqtgraph/__init__.py
+++ b/pyqtgraph/__init__.py
@@ -90,8 +90,8 @@ def getConfigOption(opt):
 ### Set style options
 ## All options related to style are set here
 ## Init default style
-from .style.core import loadDefautStyle
-configStyle = loadDefautStyle()
+from .style.core import loadDefaultStyle
+configStyle = loadDefaultStyle()
 
 
 
--- a/pyqtgraph/__init__.py
+++ b/pyqtgraph/__init__.py
@@ -27,6 +27,9 @@ from .Qt import mkQApp
               ## (import here to avoid massive error dump later on if numpy is not available)
 
 
+
+### Set runing options
+## All options not related to style are set here
 ## in general openGL is poorly supported with Qt+GraphicsView.
 ## we only enable it where the performance benefit is critical.
 ## Note this only applies to 2D graphics; 3D graphics always use OpenGL.
@@ -40,10 +43,6 @@ else:
 CONFIG_OPTIONS = {
     'useOpenGL': useOpenGL, ## by default, this is platform-dependent (see widgets/GraphicsView). Set to True or False to explicitly enable/disable opengl.
     'leftButtonPan': True,  ## if false, left button drags a rubber band for zooming in viewbox
-    # foreground/background take any arguments to the 'mkColor' in /pyqtgraph/functions.py
-    'foreground': 'd',  ## default foreground color for axes, labels, etc.
-    'background': 'k',        ## default background for GraphicsWidget
-    'antialias': False,
     'editorCommand': None,  ## command used to invoke code editor from ConsoleWidgets
     'exitCleanup': True,    ## Attempt to work around some exit crash bugs in PyQt and PySide
     'enableExperimental': False, ## Enable experimental features (the curious can search for this key in the code)
@@ -83,6 +82,19 @@ def getConfigOption(opt):
     """Return the value of a single global configuration option.
     """
     return CONFIG_OPTIONS[opt]
+
+
+
+
+
+### Set style options
+## All options related to style are set here
+## Init default style
+from .style.core import loadDefautStyle
+configStyle = loadDefautStyle()
+
+
+
 
 
 def systemInfo():
--- a/pyqtgraph/examples/ColorBarItem.py
+++ b/pyqtgraph/examples/ColorBarItem.py
@@ -42,7 +42,7 @@ class MainWindow(QtWidgets.QMainWindow):
         p2.addItem( i2, title='' )
         # inserted color bar also works with labels on the right.
         p2.showAxis('right')
-        p2.getAxis('left').setStyle('showValues', False )
+        p2.getAxis('left').setStyle(showValues=False)
         p2.getAxis('bottom').setLabel('bottom axis label')
         p2.getAxis('right').setLabel('right axis label')
 
--- a/pyqtgraph/examples/ColorBarItem.py
+++ b/pyqtgraph/examples/ColorBarItem.py
@@ -42,7 +42,7 @@ class MainWindow(QtWidgets.QMainWindow):
         p2.addItem( i2, title='' )
         # inserted color bar also works with labels on the right.
         p2.showAxis('right')
-        p2.getAxis('left').setStyle( showValues=False )
+        p2.getAxis('left').setStyle('showValues', False )
         p2.getAxis('bottom').setLabel('bottom axis label')
         p2.getAxis('right').setLabel('right axis label')
 
--- a/pyqtgraph/examples/CustomGraphItem.py
+++ b/pyqtgraph/examples/CustomGraphItem.py
@@ -8,7 +8,7 @@ import pyqtgraph as pg
 from pyqtgraph.Qt import QtCore
 
 # Enable antialiasing for prettier plots
-pg.setConfigOptions(antialias=True)
+pg.configStyle['graphItem.antialias'] = True
 
 w = pg.GraphicsLayoutWidget(show=True)
 w.setWindowTitle('pyqtgraph example: CustomGraphItem')
@@ -22,7 +22,7 @@ class Graph(pg.GraphItem):
         self.textItems = []
         pg.GraphItem.__init__(self)
         self.scatter.sigClicked.connect(self.clicked)
-        
+
     def setData(self, **kwds):
         self.text = kwds.pop('text', [])
         self.data = kwds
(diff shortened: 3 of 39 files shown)


j9ac9k:
Feel free to use #1625 for inspiration 

That PR never got merged in large part due to the size/scope and my reluctance to merge something with such a big diff... I do regret not reviewing that more thoroughly in a timely fashion.

pyqtgraph/__init__.py
```diff
@@ -85,6 +84,19 @@ def getConfigOption(opt):
     return CONFIG_OPTIONS[opt]
 
 
+
+
+
+### Set style options
+## All options related to style are set here
+## Init default style
+from .style.core import loadDefautStyle
```
NilsNemitz, line 93: should that be "DefaultStyle" throughout, maybe?

NilsNemitz:
Hi @edumur ,

It is great to see someone working on this, I hope you can get this merged!
My own attempts (there's also #1921 in addition to #1621 ) stalled since I couldn't figure out a scope that wouldn't break the entire library at the same time.

But there's some things in that earlier attempt that might be worth looking at. Two things come to mind:

It would be great to expose some of the style colors through the existing color dictionary. For example, so that user generated graphs (drawn with a "foreground color") can function in both light and dark mode.

I had a working method to send redraw requests to all pyqtgraph items. That made it possible to apply changes to existing graphs. I had been expecting to find something built in on the Qt level, but somehow that all failed to work.

There's also a number of color palettes for styling already :)

Best of luck!

Unfortunately, I looks like I'll be in zero-free-time mode until at least end of May.

edumur:
Is there any docstring style recommanded?

j9ac9k:
> Is there any docstring style recommanded?

numpy docstyle for sure, ... maybe one evening if I'm too tired to do any coding, I can start converting existing docstrings over...

pyqtgraph/__init__.py
NilsNemitz: Spelling: "Defaut" != "Default"

edumur: lol that's a very French mistake to do ^^ .

Thanks.

edumur:
Is there a reason why some objects are defined with an Item in there name such as "LabelItem" and others without such as "ScaleBar"?

j9ac9k:
> Is there a reason why some objects are defined with an Item in there name such as "LabelItem" and others without such as "ScaleBar"?

So first I want to acknowledge that our API consistency isn't as good as it should be.

That said, PyQtGraph classes that end in Item typically inherit from `QGraphicsItem`. I'm on mobile so not sure about ScaleBar specifically but classes that don't end in Item are typically QWidgets.

We recently added this diagram to help with understanding the relationships between PyQtGraph and parent Qt classes.

https://pyqtgraph.readthedocs.io/en/latest/api_reference/uml_overview.html

edumur:
Thanks, I am still confused with "item", "object", "view" widget.
The disgram is nice thought!

j9ac9k:
> Thanks, I am still confused with "item", "object", "view" and, "widget". The disgram is nice thought!

SomethingObject inherits from QGraphicsObject, and SomethingView from QGraphicsView, same with QGraphicsWidget for being SomethingWidget.  What complicates things is that QGraphicsObject inherits from QGraphicsItem, and QGraphicsWidget inherits from QGraphicsObject 😆 

Also what can complicate matters is that [I don't believe you can do multiple inheritances of QObject sublcasses](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/GraphicsItem.py#L42)

All that said, we do have inconsistencies in our naming scheme which definitely does not help.

edumur:
Hi all,

Some items have their style option stored in an attribute `self.opts` and others in `self.styles`.
I would like to make it consistante across the lib.
Any preference?

j9ac9k:
> Hi all,
> 
> Some items have their style option stored in an attribute `self.opts` and others in `self.styles`. I would like to make it consistante across the lib. Any preference?

I low-key dislike `self.opts` ... I tend to fall back towards whatever the upstream API is for this sort of thing, in this case, where Qt objects store style information, ...but for this sort of thing I'm not seeing something that's analogous in the Qt API; so I would lean towards `self.styles` myself; but I'll go with whatever you would prefer (given you're the one actually doing the work :) )

ntjess:
Can you link a few locations where `styles` is used? I found some with `opts` in the existing codebase but not with `styles`

Also -- Is there a reason to define a `style.core` with a new file format instead of relying on the builtin `ConfigParser`? It seems the syntax is largely compatible and we wouldn't need to be responsible for maintaining the parser. This is one of the reasons `pyqtgraph/configfile.py` was discouraged for use elsewhere

edumur:
Sure: [here](https://github.com/pyqtgraph/pyqtgraph/blob/d0b9b7c66cb86a526c1f64000428d442a52d7a5c/pyqtgraph/graphicsItems/AxisItem.py).

style.core is made to host any general style related code.
However I would love to use a standard file format instead of the current *.pstyle !
However configparser can only store string, see [here](https://docs.python.org/3/library/configparser.html#supported-datatypes) and currently the *.pstyle file is storing various type such as bool, float, list of tuple, ...
Maybe a python dict? a yaml file? something else?
I also do not like parsing the fle myself.

j9ac9k:
> Sure: [here](https://github.com/pyqtgraph/pyqtgraph/blob/d0b9b7c66cb86a526c1f64000428d442a52d7a5c/pyqtgraph/graphicsItems/AxisItem.py).
> 
> style.core is made to host any general style related code. However I would love to use a standard file format instead of the current *.pstyle ! However configparser can only store string, see [here](https://docs.python.org/3/library/configparser.html#supported-datatypes) and currently the *.pstyle file is storing various type such as bool, float, list of tuple, ... Maybe a python dict? a yaml file? something else? I also do not like parsing the fle myself.

I'm reluctant towards yaml as that would involve an extra dependency.  Python 3.11 introduced `tomllib`; if we don't want to do json, I would be ok with a toml file and require end users to provide some dependency that can read it in if using python < 3.11.

ntjess:
> Sure: [here](https://github.com/pyqtgraph/pyqtgraph/blob/d0b9b7c66cb86a526c1f64000428d442a52d7a5c/pyqtgraph/graphicsItems/AxisItem.py).

Thanks, I was doing a search for `styles` which is why I missed it
> 
> However configparser can only store string, see [here](https://docs.python.org/3/library/configparser.html#supported-datatypes)

ConfigParser is easily customizable:

```python
from configparser import ConfigParser, Interpolation
import ast

class TypedInterpolation(Interpolation):
    def before_get(self, parser, section, option, value, defaults):
        try:
            return ast.literal_eval(value)
        except (ValueError, SyntaxError):
            return super().before_get(parser, section, option, value, defaults)
typed_config = ConfigParser(interpolation=TypedInterpolation())

ini_file = """
[linePlot]
string = 'hello'
tuple = (1, 2, 3)
list = [1, 2, 3]
dict = {'a': 1, 'b': 2, 'c': 3}
int = 1
float = 1.0
bool = True
none = None
"""

typed_config.read_string(ini_file)
for key, value in typed_config.items('linePlot'):
    print(key, value, type(value))

# string hello <class 'str'>
# tuple (1, 2, 3) <class 'tuple'>
# list [1, 2, 3] <class 'list'>
# dict {'a': 1, 'b': 2, 'c': 3} <class 'dict'>
# int 1 <class 'int'>
# float 1.0 <class 'float'>
# bool True <class 'bool'>
# none None <class 'NoneType'>
```

This implementation essentially allows any builtin python type, defaulting to a string otherwise

edumur:
@ntjess thanks, I will try this!

@j9ac9k I just realized that replacing opts by styles was a breaking change anyway so I will leave both as it is.

ntjess:
It is better to avoid breaking changes immediately; instead, consider a deprecation warning while keeping old behavior:

```python
def setStyle(...):
    warnings.warn("`setStyle` is deprecated in favor of `setStyles`, see XXX", FutureWarning)
    ...

```

Also, for pieces of the library that are as old as this, deprecation may not even be an option (884 results on [github](https://github.com/search?q=pyqtgraph+%2F%5C.setStyle%5Cb%2F+lang%3Apython&type=code) alone). We might just want to leave a comment in the docstring that new behavior is preferred, but keep everything else.

Curious about others' thoughts, though

edumur:
Hi all,

Just push the last update.
Not done yet.

* Most of item have the new styles implementation.
* All test are passed
* I added a default white style and so far, it looks good.

j9ac9k:
Hi there @edumur 

Looks like read the docs doesn't like some of the enum namespacing;

```
List[Tuple[QtCore.QRectF, QtCore.Qt.Alignment, str]]]]:
```

I think this should be `QtCore.Qt.AlignmentFlag`

j9ac9k:
I need to go through this diff more closely; but at first glance, I would steer away from `def getAttribute(self)` as method names, and just stick with `def attribute()`.  That's more in-line with the Qt API (Qt has very few methods that start with the word "get").  I know pyqtgraph has some of those, but in time I suspect we'll phase them out.

j9ac9k:
The read the docs error in the CI has already been fixed on `master`, so you'll have to rebase master onto your branch for it to work (also likely will need to do that anyway due to merge conflicts).

edumur:
To be sure I do not f** up things.
Could someone provide the git commands to pull the master and rebase it into my branch?

Thanks.

j9ac9k:
> To be sure I do not f** up things. Could someone provide the git commands to pull the master and rebase it into my branch?
> 
> Thanks.

Since you made a PR, I actually have push rights to your branch; I'd be happy to do it if you'd like.  If you want to do it, these are the steps I would do...

```bash
# go to your branch
git switch style 

 # create a backup in case things get hosed up
git switch -c style-backup 
git switch master
# sync up with upstream locally on your disk
git pull upstream master  
git switch style
git rebase master
```

When you run that last command, you're going to get merge conflicts that stop the rebase process, on each one of those, you'll have to resolve the conflict in the file, then run `git add <path-to-file-with-conflict-now-resolved>` and then `git rebase --continue`

j9ac9k:
Read the docs reports the following errors (read the docs is configured to treat warnings as errors).

```
reading sources... [ 21%] api_reference/graphicsItems/axisitem
:49: (ERROR/3) Unexpected indentation.
:51: (ERROR/3) Unexpected indentation.
:52: (WARNING/2) Block quote ends without a blank line; unexpected unindent.
...
reading sources... [ 26%] api_reference/graphicsItems/errorbaritem
:7: (ERROR/3) Unexpected indentation.
```

edumur:
Is there a way to run these tests locally?

Currently I am running `test.py` in [yqtgraph folder and got this:
```bash
(plotter_env) PS C:\Users\etien\Desktop\edumur\script\pyqtgraph> python .\test.py
sys.platform: win32
sys.version: 3.10.11 | packaged by Anaconda, Inc. | (main, May 16 2023, 00:55:32) [MSC v.1916 64 bit (AMD64)]
qt bindings: PyQt5 5.15.9 Qt 5.15.2
pyqtgraph: 0.13.4.dev0; None
config:
{'crashWarning': False,
 'editorCommand': None,
 'enableExperimental': False,
 'exitCleanup': True,
 'leftButtonPan': True,
 'mouseRateLimit': 100,
 'segmentedLineMode': 'auto',
 'useCupy': False,
 'useNumba': False,
 'useOpenGL': False}
============================================================================================= test session starts ==============================================================================================
platform win32 -- Python 3.10.11, pytest-7.4.0, pluggy-1.2.0
rootdir: C:\Users\etien\Desktop\edumur\script\pyqtgraph
configfile: pytest.ini
collected 417 items

pyqtgraph\examples\test_examples.py ..
```

edumur:
Nevermind, I think I solved it .

j9ac9k:
> Nevermind, I think I solved it .

Let me know if you need help debugging test failures!

edumur:
I realized during this work that this PR was too much in a non pyqt spirit and dropped it.
I will close it.

PR #2671, opened 2023-04-01
ViewBoxMenu: remove unused empty WidgetGroup

pijyoi:
This PR removes the unused instantiated `WidgetGroup`s from ViewBoxMenu.
In fact, until less than a year ago (74fa36da37d36a3da52019f673898ff0fe165de0), the (unused) instantiated `WidgetGroup`s were not even stored.


pijyoi:
Even though the objects removed in this PR are unused, removing them is reliably making the CI fail for Conda/Windows/PySide2.

j9ac9k:
> Even though the objects removed in this PR are unused, removing them is reliably making the CI fail for Conda/Windows/PySide2.

Ugh ... well that sucks

PR #2683, opened 2023-04-11
Avoid changing background colors of text and rows for group parameter…

ntjess:
Most themes (including defaults on windows and gnome) use alternating colors for each table row. Current styling doesn't play nice with this assumption

This PR simply removes row and font coloring to avoid those difficulties:
image

Note that in the "before" samples (before this PR), the colors don't match on alternate rows, styling doesn't stand out very well, and (importantly) on some popular styling choices (there are more on GH), the text isn't properly changed and appears as very low contrast against the background.

I feel this is sufficient reason to restrain from changing background colors. I tested on a few more themes without screenshots, but I don't think there are any cases where the PR causes a _worse_ appearance. Notably, `qt-material` isn't happy either way, but the problem doesn't get exacerbated with these changes.

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes/basetypes.py
+++ b/pyqtgraph/parametertree/parameterTypes/basetypes.py
@@ -337,22 +337,11 @@ class GroupParameterItem(ParameterItem):
         """
         Change set the item font to bold and increase the font size on outermost groups.
         """
-        app = mkQApp()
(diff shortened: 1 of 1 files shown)


j9ac9k:
image

This is what things look like on macOS with dark mode.  Still getting the overlapping text but that's been an issue for ages.  Otherwise it looks good

PR #2689, opened 2023-04-14
Only apply nan mask workaround for cp version below 10.0.

koenstrien:
In [#1595](https://github.com/pyqtgraph/pyqtgraph/pull/1595) a workaround was introduced for masking NaNs in cupy arrays, because the indexing is different than for numpy arrays. As of cupy version 10.0.0 the indexing can be done the same for numpy and cupy arrays again (see cupy PR [#6196](https://github.com/cupy/cupy/pull/6196)) and the workaround causes issues. This is solved by only applying the workaround if the cupy version is lower then 10.0.0. In the future the workaround can be moved completely.

Fixes [#1595](https://github.com/pyqtgraph/pyqtgraph/pull/1595)

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1533,7 +1533,7 @@ def makeARGB(data, lut=None, levels=None, scale=None, useRGBA=False, maskNans=Tr
     if nanMask is not None:
         alpha = True
         # Workaround for https://github.com/cupy/cupy/issues/4693, fixed in cupy 10.0.0
-        if xp == cp and StrictVersion(xp.__version__) < StrictVersion("10.0.0"):
+        if xp == cp and tuple(map(int, cp.__version__.split("."))) < (10, 0):
             imgData[nanMask, :, dst_order[3]] = 0
         else:
             imgData[nanMask, dst_order[3]] = 0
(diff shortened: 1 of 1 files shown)


koenstrien:
Right now `distutils.version.StrictVersion` is used for the version check. Maybe `packaging.version` is better (vendored by setuptools too), but I think this would mean `packaging` should be added to the requirements.

pijyoi:
Several places in the library do something like:
```python
tuple(map(int, cp.__version__.split("."))) < (10, 0)
```

koenstrien:
That is another possibility if preferred.

j9ac9k:
I would avoid using `distutils` as it is being removed in Python 3.12.  In testing code paths, you can use the `packaging` library as that is a dependency of `pytest` but for runtime code, we should probably stick to what @pijyoi suggested above (although I'm certainly open to other suggestions).

j9ac9k:
Hi @koenstrien 

I hope you don't mind, I pushed a commit to your branch removing the import.  The reason I did it is I'm about to leave for a vacation for a little over a week, and I wanted to do a new release before I left.  I wanted this PR to be in the new release, I hope you don't mind.

Thanks for the PR!

Issue #2672, opened 2023-04-02
PlotItem leak since PySide6 6.5.0


pijyoi:
### Short description

Since PySide6 6.5.0, some objects within `PlotItem` will be leaked when `PlotItem` is destroyed.

This issue is a spinoff from #2665 as this issue is specific to PySide6 6.5.0.

### Code to reproduce

Additional `psutil` library is used to show memory growth.

```python
import pyqtgraph as pg
from pyqtgraph.Qt import QtCore
import psutil

app = pg.mkQApp()
timer = QtCore.QTimer()
proc = psutil.Process()

rss0 = proc.memory_info().rss

cnt = 0
def timer_cb():
    global cnt
    rss = proc.memory_info().rss
    print(f'{cnt}: {(rss - rss0)/1048576:.1f} MB')
    pg.PlotItem()
    cnt += 1
    if cnt > 100:
        timer.stop()
        app.exit()

timer.timeout.connect(timer_cb)
timer.start(0)
pg.exec()
```

### Expected behavior

Memory usage should stabilize.

### Real behavior

Memory usage continues to grow with each iteration.

### Tested environment(s)

 * PyQtGraph version: master 
 * Qt Python binding: PySide6 6.5.0 
 * Python version: Python 3.10.10
 * NumPy version: 1.24.2 
 * Operating system: Windows 11
 * Installation method: pip 

### Additional context
Doesn't occur for PySide2 5.12.2, PyQt5 5.15.9, PySide6 6.4.3, PyQt6 6.4.2

pijyoi:
No leak for PyQt6 6.5.0 either.

pijyoi:
Minimal pyqtgraph-free script that leaks on PySide >= 6.5.0
```python
from PySide6 import QtCore, QtWidgets
import psutil

class Leaker:
    def __init__(self):
        parent = None
        parent = QtWidgets.QWidget()    # this line seemingly makes PySide < 6.5.0 not leak
        wgt= QtWidgets.QSpinBox(parent)
        wgt.valueChanged.connect(self.mkChangeCallback(wgt))

    def mkChangeCallback(self, w):
        return lambda *args: self.widgetChanged(w, *args)
        
    def widgetChanged(self, w, *args):
        pass

app = QtWidgets.QApplication([])
timer = QtCore.QTimer()
proc = psutil.Process()

rss0 = proc.memory_info().rss

cnt = 0
def timer_cb():
    global cnt
[...]

PR #2694, opened 2023-04-15
WidgetGroup: avoid lambda by using self.sender()

pijyoi:
workaround #2672 by avoiding the mechanism triggering the leak in PySide6 6.5.0 

1) `WidgetGroup` was creating lambda closures with references to both `self` and `widget`.
   * This is a bad practice and should be avoided.
   * However, prior to PySide6 6.5.0, this didn't cause a leak. (although it would not have been surprising if it had!)
2) https://github.com/pyqtgraph/pyqtgraph/issues/2672#issuecomment-1507056182 demonstrates that the leak occurs even with a lambda that has no closures.
   * i.e. various techniques using weakrefs would still have leaked
   * further testing shows that using custom `def` functions (with no closures) also leaks
3) The whole reason for the closure was to pass in an argument that would allow identification of the widget emitting the signal. This identification can be done with `QObject::sender` (https://doc.qt.io/qt-6/qobject.html#sender) and avoids the closure.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -169,7 +169,7 @@ jobs:
           miniforge-variant: Mambaforge
           environment-file: ${{ matrix.environment-file }}
           auto-update-conda: false
-          python-version: "3.11"
+          python-version: "3.10"
           use-mamba: true
       - name: "Install Test Framework"
         run: pip install pytest pytest-xdist
--- a/pyqtgraph/WidgetGroup.py
+++ b/pyqtgraph/WidgetGroup.py
@@ -172,7 +172,7 @@ class WidgetGroup(QtCore.QObject):
         if signal is not None:
             if inspect.isfunction(signal) or inspect.ismethod(signal):
                 signal = signal(w)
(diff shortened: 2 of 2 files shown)


pijyoi:
CI is crashing like for #2671, but now also includes Conda/Ubuntu/PySide2.
Removing that particular test does make it pass.

pijyoi:
#2606 bumped conda python from 3.9 to 3.11, skipping 3.10.
By dropping to 3.10, the conda CI passes.

According to a comment in https://bugreports.qt.io/browse/PYSIDE-1864, Python > 3.10 will not be supported by Qt.
Conda builds it for Python 3.11 by applying an unofficial patch https://github.com/conda-forge/pyside2-feedstock/blob/main/recipe/patches/python3.11.patch

pijyoi:
#1598 mentions `pyqtgraph.WidgetGroup.WidgetGroup.mkChangeCallback` as being unsolvable at the time.

j9ac9k:
Thanks @pijyoi ...appreciate the touch-up on the CI, testing python 3.11 with pyside2 does feel silly.

PR #2704, opened 2023-04-25
TypeError when using multiple pens for different lines

adrianlubitz:
```python
lastPen = None
for i in range(pts.shape[0]):
    pen = self.pen[i]
    if np.any(pen != lastPen):
```
comparison with None will always cause 
```
if np.any(pen != lastPen):
TypeError: Cannot compare structured or void to non-void arrays.
```
This does not really follow DRY principle but was the best solution I could come up with using static types in structured array

Fixes #<IssueNumber>

### Other Tasks 

<details>

### Files that need updates
    
- [ ] `README.md`
- [ ] `setup.py`
- [ ] `tox.ini`
- [ ] `pyproject.toml`

</details>

<details>

### Pre Release Checklist

</details>

<details>

### Steps To Complete

- [ ] Announce on mail list
- [ ] Announce on Twitter

</details>

### DIFF
--- a/pyqtgraph/graphicsItems/GraphItem.py
+++ b/pyqtgraph/graphicsItems/GraphItem.py
@@ -108,16 +108,25 @@ class GraphItem(GraphicsObject):
             pts = self.pos[self.adjacency]
             pen = self.pen
             if isinstance(pen, np.ndarray):
-                lastPen = None
+                pen = lastPen = self.pen[0]
+                if pen.dtype.fields is None:
+                    p.setPen(
+                        fn.mkPen(color=(pen[0], pen[1], pen[2], pen[3]), width=1))
+                else:
+                    p.setPen(fn.mkPen(color=(
+                        pen['red'], pen['green'], pen['blue'], pen['alpha']), width=pen['width']))
                 for i in range(pts.shape[0]):
                     pen = self.pen[i]
                     if np.any(pen != lastPen):
                         lastPen = pen
                         if pen.dtype.fields is None:
-                            p.setPen(fn.mkPen(color=(pen[0], pen[1], pen[2], pen[3]), width=1))                            
+                            p.setPen(
+                                fn.mkPen(color=(pen[0], pen[1], pen[2], pen[3]), width=1))
                         else:
-                            p.setPen(fn.mkPen(color=(pen['red'], pen['green'], pen['blue'], pen['alpha']), width=pen['width']))
-                    p.drawLine(QtCore.QPointF(*pts[i][0]), QtCore.QPointF(*pts[i][1]))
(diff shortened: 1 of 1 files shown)


pijyoi:
How about:
```python
if lastPen is None or np.any(pen != lastPen):
```

adrianlubitz:
> How about:
> 
> ```python
> if lastPen is None or np.any(pen != lastPen):
> ```

This is obviously the more elegant solution :sweat_smile: 
Edited it. 
@pijyoi Let me know if there is something else I need to do :wink:

j9ac9k:
This LGTM, thanks for the PR @adrianlubitz and thanks for the review/feedback @pijyoi

adrianlubitz:
@j9ac9k When will this be uploaded to PyPI as a new Version? I want to use it as a dependency in [CoBaIR](https://github.com/dfki-ric/CoBaIR) and PyPI only allows dependencies from PyPI itself. Looking forward to Version 0.13.4 :package: :partying_face:

pijyoi:
Looking at the code, I don't think that passing in red,green,blue,alpha,width as a numpy record array to specify the pens seems a good idea. (It's also possible to pass in a regular numpy array to specify r,g,b,a, but that's not documented in the docstring)
The `lastPen` variable is a sort of optimization to avoid recreating `QPen`s, but it only works if the same-colored pens are adjacent.

I would think a more useful input parameter would be a Python `list` of (either `QPen` or something that `mkPen` accepts). This way, the user can use the same pre-created `QPen` object within the list. (Similar to what `BarGraphItem` code is doing)

Prior to this PR, use of multiple pens wouldn't have worked anyway. Would it be a good opportunity to fix the api now?

j9ac9k:
> Would it be a good opportunity to fix the api now?

Now is always better than later.

pijyoi:
>Prior to this PR, use of multiple pens wouldn't have worked anyway.

Now I am puzzled. https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/examples/GraphItem.py specifies multiple pens and does work on 0.13.3. So what was this PR fixing?

```python
In [40]: lines = np.array([
    ...:     (255,0,0,255,1),
    ...:     (255,0,255,255,2),
    ...:     (255,0,255,255,3),
    ...:     (255,255,0,255,2),
    ...:     (255,0,0,255,1),
    ...:     (255,255,255,255,4),
    ...:     ], dtype=[('red',np.ubyte),('green',np.ubyte),('blue',np.ubyte),('alpha',np.ubyte),('width',float)])

In [41]: lines[4] == None
Out[41]: False

In [42]: lines[4] != None
Out[42]: True

In [43]: np.any(lines[4] != None)
Out[43]: True
```

PR #2707, opened 2023-04-28
Restore axes list in ViewBoxMenu

marti1999:
Hi,

This PR is to restore the `axes` list in `ViewBoxMenu`. 

On [this commit](https://github.com/pyqtgraph/pyqtgraph/commit/8a791819e74440adaa88097b658c1eba34ef6af0) the list `self.axes` is deleted from the code. With this change there seems to be no way to modify such axes. 

An example of what we used to do with it:
```python
menu = self.plotItem.getViewBox().menu.axes[0]
action = menu.addAction("Log scale")
action.setCheckable(True)
action.setChecked(self.plotItem.getAxis("bottom").logMode)
action.setParent(menu)
action.toggled.connect(self._onXLogToggled)
self.menu.axes[0].addAction(action) 
```

Looking at `QtWidgets.QMenu` and `ViewBoxMenu`, there seems to be no alternative to access the axes.

### DIFF
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBoxMenu.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBoxMenu.py
@@ -18,7 +18,8 @@ class ViewBoxMenu(QtWidgets.QMenu):
         self.viewAll = QtGui.QAction(translate("ViewBox", "View All"), self)
         self.viewAll.triggered.connect(self.autoRange)
         self.addAction(self.viewAll)
-        
+
+        self.axes = []
         self.ctrl = []
         self.widgetGroups = []
         self.dv = QtGui.QDoubleValidator(self)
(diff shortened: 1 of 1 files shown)


pijyoi:
You could obtain the axis like this:
```python
vb = plt.getPlotItem().getViewBox()
for submenu in vb.menu.findChildren(QtWidgets.QMenu):
    if submenu.title() == "X axis":
        act = submenu.addAction("Log scale")
        act.setCheckable(True)
        act.setChecked(True)
```

The question I would ask is whether `ViewBoxMenu.axes` was there 
1) for public access
2) merely to hold a Python reference to keep the `QMenu` alive.
   * this is no longer necessary as it is a C++ child of `ViewBoxMenu`.
3) both of the above

Consider also `ViewBoxMenu.dv`. The library doesn't make use of this `QDoubleValidator`. But is it also public API?

marti1999:
Thank you @pijyoi 
Your approach to the problem is indeed better. I've just tried it and it is working fine.

PR #2720, opened 2023-05-13
keep track of respective finiteness of x and y

pijyoi:
A common plotting case is when "x" is provided as integers.
This also happens when only "y" is specified, in which case "x" is automatically populated as `np.arange(len(y))`.

In this case, it is wasteful to check the finiteness of "x".

In implementing this PR, it was found to be more useful to keep track of the finiteness of `x` and `y` arrays separately.


j9ac9k:
Thanks @pijyoi this LGTM.  Love the diff; and I agree checking each x and y independently is the way to go.

PR #2726, opened 2023-05-19
guard against auto-downsample oscillation

### LINKED ISSUE #2713: Program loops indefinitely when I start an application with plot set on autosampling.
### Short description
When I start a pyqtgraph application with a plot with auto Downsampling, it loops over different values of the downsampling parameter and then crashes. I also have remarks about the optimization of the Downsampling but will put them in a discussion.

### Code to reproduce

```
import pyqtgraph as pg
from PyQt6 import QtCore, QtWidgets

import numpy as np
pp = pg.mkQApp()
mw = QtWidgets.QMainWindow()
mw.setWindowTitle('pyqtgraph example: PlotWidget')
mw.resize(800,800)
cw = QtWidgets.QWidget()
mw.setCentralWidget(cw)
l = QtWidgets.QVBoxLayout()
cw.setLayout(l)

pw = pg.PlotWidget()  ## giving the plots names allows us to link their axes together
pw.setClipToView( True)
pw.enableAutoRange(False)
pw.setDownsampling( ds=None, auto=1, mode="subsample")
l.addWidget(pw)
data = np.arange(10000000)+np.random.normal(0,10,10000000)
pw.plot(data, title="Simplest possible plotting example")
mw.show()

if __name__ == '__main__':
    pg.exec()
```

### Real behavior
The program loops. When I print the "ds" parameter, it goes down until the program crashes.

### Tested environment(s)

 * PyQtGraph version: 0.13.3 
 * Qt Python binding: PyQt6
 * Python version: Python 3.10
 * NumPy version: 1.21.5
 * Operating system: Ubuntu
 * Installation method: pip

pijyoi:
This PR attempts to fix #2713 where for particularly large data sets, the auto-downsampled value never stabilizes and instead oscillates between two values.

If a newly computed auto-downsample value differs very little from the previous calculated value, then the previous value is used instead.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -447,6 +447,7 @@ class PlotDataItem(GraphicsObject):
         self.opts['logMode'] = [xState, yState]
         self._datasetMapped  = None  # invalidate mapped data
         self._datasetDisplay = None  # invalidate display data
+        self._adsLastValue   = 1     # reset auto-downsample value
         self.updateItems(styleUpdate=False)
         self.informViewBoundsChanged()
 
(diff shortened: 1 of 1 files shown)


j9ac9k:
I'm going to go ahead and merge this despite not hearing from the author of the issue.  The diff LGTM.  Thanks @pijyoi

PR #2728, opened 2023-05-19
For SVG exports, shift PlotCurveItems to be about 0,0

### LINKED ISSUE #1849: SVG export: linearGradient not exported correctly

### LINKED ISSUE #2018: SVGExporter: Background color
SVGExporter generates an additional background tag of 100% width and height in generateSvg(). This tag uses rgba to define the color which seens not to be fully supported by the standard. I.e. inkscape won't recognize the color, causing it to appear black (BTW: the size is also not correct).

### Short description
Background color rectangle is incorrectly rendered.

### Code to reproduce
Call export from context menu and select SVG as output format. Press "Copy" button and insert SVG image to e.g. inkscape (tested with version 1.1). 

### Expected behavior
The background shall be filled by the selected color.

### Real behavior
The background rectangle will be black no matter what color was selected. The size is also wrong. 

### Tested environment(s)

 * PyQtGraph version: 0.12.3
 * Qt Python binding: PySide2 5.13.2 Qt 5.12.9
 * Python version: 3.8
 * NumPy version: 1.20.1
 * Installation method: conda

### Additional context
Proposed solution: either use Qt SVGGenerator to generate the background rectangle or add an option to disable background export at all (no background rectangle to be generated).

### LINKED ISSUE #2661: SVG exporter output has errors when data values are very large
I found a bug when trying to export a plot as an SVG. It seems that incorrect data values are being saved in the .svg when the data values are large (~1,000,000). 

Here is a minimal example:
```import pyqtgraph as pg 
import numpy as np
import scipy


### create data
x = np.arange(0,500,10)
y = np.random.random(50)*0.3 + scipy.signal.gaussian(50, std=10)

### create plot where export works
p1 = pg.plot()
p1.plot(x=x, y=y, pen='g')

### create plot where export doesn't work because we've increased the x-axis values
p2 = pg.plot()
p2.plot(x=x+10000000, y=y, pen='b')

```

And here is an illustration of the output:
4 plots showing the output of exporting p1 and p2 in the above code as svg and png files. the svg plot of p2 is different than the other 3 - it is more jagged and blocky, as if several different x-values were grouped together.

Here's the pg.systemInfo output for my machine:
```sys.platform: linux
sys.version: 3.9.13 (main, Aug 25 2022, 23:26:10) 
[GCC 11.2.0]
qt bindings: PyQt5 5.15.7 Qt 5.15.2
pyqtgraph: 0.13.1; None
config:
{'antialias': False,
 'background': 'k',
 'crashWarning': False,
 'editorCommand': None,
 'enableExperimental': False,
 'exitCleanup': True,
 'foreground': 'd',
 'leftButtonPan': True,
 'mouseRateLimit': 100,
 'segmentedLineMode': 'auto',
 'useCupy': False,
 'useNumba': False,
 'useOpenGL': False}
```

j9ac9k:
Fixes #2661
Fixes #2018
Fixes #1849

This one is a doozy... 

In the SVG exporter; there is a QPainter draw operation that occurs to a QDataBuffer that is interacting w/ a QSvgGenerator.  The issue is that the positional information in the view (x and y coordinates for line to, move to, etc), are relatively high (> 1E8) and small differences between values (in the case of #2661, differences in increments of 10) are lost.  

The way this PR attempts to get around this is by shifting PlotCurveItem's such that they are centered about (0, 0), and compute the transform for that shift, then apply the transform to the QPainter; so the curve appears to be in the correct place on the SVG output.

There are some more todo's with this PR

- [ ] Don't only work with PlotCurveItems (should handle everything with `setData`, and ~`dataBounds`~ Using `viewRect()`) 
- [x] Don't import pg.PlotCurveItem in the exporter, use relative imports instead
- [x] Don't just handle translation, attempt to handle scaling

This PR also removed an unused lengthy PlotItem method that was left over after its usages were removed in 2012.  For now I am commenting out the method as it looks to do that similar kind of translation and scaling we want,  so I may need to borrow some code from it.  `PlotItem.writeSvgCurves` should be removed before this PR is merged.

While testing, I attempted to get scatter plots to work in a similar fashion but could not get it working; so I decided to give up on that effort since I'm not sure if anyone cares about scatter plots that are placed in positions with larger absolute values.


pyqtgraph/exporters/SVGExporter.py
```diff
@@ -358,22 +388,22 @@ def correctCoordinates(node, defs, item, options):
                 # If coords start with L instead of M, then the entire path will not be rendered.
                 # (This can happen if the first point had nan values in it--Qt will skip it on export)
                 if newCoords[0] != 'M':
-                    newCoords = 'M' + newCoords[1:]
+                    newCoords = f'M{newCoords[1:]}'
                 ch.setAttribute('d', newCoords)
             elif ch.tagName == 'text':
-                removeTransform = False
+                removeTransform = True
```
j9ac9k, line 394: Whoops, need to udno this change I made... I enabled it for testing purposes, should comment out again...

j9ac9k:
This PR is working again, but I'm still puzzled by two things.

1. Why am I _not_ having to call `item.setData(xOriginalData, yOriginalData)` after the export occurs?  How is the curve still in the correct place in the view?
2. Why do I have to call `item.setData(xModified, yModified)` to begin with instead of using a QTransform on the curve to shift and scale it over (0, 0)?

Ideally, I would be able to apply the QTransform that shifts the GraphicsItem to the center, if that method works we can use that same method for scatter plots and other graphics items without a bunch of if/elif statements.

j9ac9k:
Looks like this type of annotation needs python 3.9+ ...while I could use the 3.8 compatible variant, we should have dropped python 3.8 already per NEP-29.... will let this PR sit idle until we actually drop python 3.8.

j9ac9k:
...and somewhere along the way I broke the exporter, it's no longer showing the curve in the SVG 😆

j9ac9k:
Since I'm now looking at the SVG exporter and I hope I don't have to look at it for a while in the future, decided to investigate #2018 

[c996598](https://github.com/pyqtgraph/pyqtgraph/pull/2728/commits/c9965980e68675de67462ea8c98c60b658a54e8b) addresses the background issue reported there, but the size (height/width) parameters are still not being accepted correctly.

PR #2731, opened 2023-05-24
Adapt hide icon (invisible eye) to style of other icons

stephan-senkbeil:
The icon that is displayed when hiding a curve by clicking on the corresponding legend entry is out of style compared to all other icons used.

This PR updates this icon to reflect the style of the other icons in `icons.svg`.
I used the eye of the old icon and the background of one of the icons from `icons.svg` to create the new icon.

Before:
old_hide_icon

After:
new_hide_icon

### DIFF
--- a/pyqtgraph/icons/invisibleEye.svg
+++ b/pyqtgraph/icons/invisibleEye.svg
@@ -1,35 +1,68 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg
+   width="110"
+   height="110"
+   viewBox="0 0 110 110"
+   version="1.1"
+   id="svg8"
+   sodipodi:docname="invisibleEye.svg"
+   inkscape:version="1.2.1 (9c6d41e410, 2022-07-14)"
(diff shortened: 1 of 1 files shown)


sem-geologist:
As this is a bit different scope compared to the other pyqtgraph icons, I am not sure. Yes, I agree the changed style is similar to other already built-in buttons, but should it be like that? Is it supposed to resemble button? other icons clearly are to be seen as button, but this - is it or should it be? For me it looks that it is only to hint the state of the curve "DatasetItem"

j9ac9k:
@stephan-senkbeil Thanks for the PR, sorry it took so long for a maintainer to respond  While I'm not completely sold on this icon being _right_, we don't have a particularly consistent style here; and this PR is certainly an improvement, so I'm going to merge.

That said, if someone has experience in this area, and wants to contribute a set of icons that are easily accessible in pyqtgraph, _please_ reach out; either by creating an issue or contacting me via one of the methods in my github user profile.

PR #2734, opened 2023-05-28
PySide6 6.5.1 compatibility updates

pijyoi:
This PR modifies code patterns in pyqtgraph that are triggering failures on the CI with PySide6 6.5.1

1) using inner functions (with value capture) as slots
  * PySide6 6.5.1 is not treating the generated inner functions with different values as distinct functions
  * This is not limited to inner functions, it also applies to lambdas
2) connecting to `Signal.emit` instead of directly to the signal

To be clear, this PR only makes the CI pass.
Whether more functionality has been broken (but not captured by the CI) is unknown.

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes/pen.py
+++ b/pyqtgraph/parametertree/parameterTypes/pen.py
@@ -174,20 +174,13 @@ class PenParameter(GroupParameter):
             name = name.title().strip()
             p.setOpts(title=name, default=default)
 
-        def penPropertyWrapper(propertySetter):
-            def tiePenPropToParam(_, value):
-                propertySetter(value)
-                self.sigValueChanging.emit(self, self.pen)
-
-            return tiePenPropToParam
-
         if boundPen is not None:
             self.updateFromPen(param, boundPen)
             for p in param:
-                setter, setName = self._setterForParam(p.name(), boundPen, returnName=True)
+                setName = f'set{p.name().capitalize()}'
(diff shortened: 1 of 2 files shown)


j9ac9k:
Oh, never noticed the connection to the `emit` method directly; that wasn't right 😅

pyqtgraph/parametertree/parameterTypes/pen.py
```diff
@@ -174,20 +174,13 @@ def _makeChildren(self, boundPen=None):
...
-                propertySetter(value)
-                self.sigValueChanging.emit(self, self.pen)
-
-            return tiePenPropToParam
-
         if boundPen is not None:
             self.updateFromPen(param, boundPen)
             for p in param:
-                setter, setName = self._setterForParam(p.name(), boundPen, returnName=True)
+                _, setName = self._setterForParam(p.name(), boundPen, returnName=True)
```
ntjess, line 180: I recommend removing `_setterForParam` entirely since it's only used here and simply doing `setName = f'set{p.name().capitalize()}'

This is the only place it's used, and it's an internal method, so there should be no problems with its removal

pijyoi:
On further reflection, there is a behavior change made by this PR.
Previously, the old instance method was saved. Now it's assumed that the old instance method is equal to the class method.
Let's hope that nobody feels the need to monkey-patch their own `QPen` instances!

PR #2746, opened 2023-06-15
Fix NumPy warning in test_functions

swt2c:
Fixes the below warning about integer conversion by using astype() - overflow seems OK/desired in this case.

```
tests/test_functions.py:290
  /home/runner/work/pyqtgraph/pyqtgraph/tests/test_functions.py:290: DeprecationWarning: NumPy will stop allowing conversion of out-of-bound Python integers to integer arrays.  The conversion of -1 to uint32 will fail in the future.
  For the old behavior, usually:
      np.array(value).astype(dtype)`
  will give the desired result (the cast overflows).
    np.arange(6, dtype=dtype), np.arange(0, -6, step=-1, dtype=dtype), 'all',
```

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -287,7 +287,7 @@ def _handle_underflow(dtype, *elements):
     "xs, ys, connect, expected", [
         *(
             (
-                np.arange(6, dtype=dtype), np.arange(0, -6, step=-1, dtype=dtype), 'all',
+                np.arange(6, dtype=dtype), np.arange(0, -6, step=-1).astype(dtype), 'all',
                 _handle_underflow(dtype,
                                   (MoveToElement, 0.0, 0.0),
                                   (LineToElement, 1.0, -1.0),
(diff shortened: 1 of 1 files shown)


swt2c:
The test failures seem to be unrelated.  In fact, they seem possibly related to Python 3.11.4?

swt2c:
I think the failures might actually be due to this CPython bug: https://github.com/python/cpython/issues/105497

j9ac9k:
ahh, yeah saw some chatter about this on the pyqt mail list...  looks like it will be fixed in 3.11.5.  I'm good with merging as is, should probably put a note on the README about an incompatibility with PyQt/Python 3.11.4, but that's out of scope for this PR.

@swt2c do you have a preferred method I can reach out to you privately? If you have some time, would you mind sending me a message on your preferred platform? My contact info is on my [github profile page](https://github.com/j9ac9k)

j9ac9k:
Thanks for the PR @swt2c this LGTM, merging!

PR #2752, opened 2023-06-22
Simplify fps computation

pijyoi:
Various examples compute fps by measuring the inverse time elapsed of _each_ frame. As there is quite a bit of jitter, the fps value is then smoothed using some rolling average.

This PR simplifies the computation by counting the number of frames that have been updated over a period of 1 second. The longish duration serves to average out the jitter and removes the need for accurate time measurement of the time elapsed of each update.

### DIFF
--- a/pyqtgraph/examples/utils.py
+++ b/pyqtgraph/examples/utils.py
@@ -1,5 +1,7 @@
 from argparse import Namespace
 from collections import OrderedDict
+from time import perf_counter
+from pyqtgraph.Qt import QtCore
 
 # Avoid clash with module name
 examples_ = OrderedDict([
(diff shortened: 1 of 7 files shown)


pyqtgraph/examples/utils.py
```diff
@@ -135,3 +137,28 @@
...
+        self.frame_cnt = 0
+        self.last_update = 0
+        self.interval = interval
+
+    def update(self):
+        self.frame_cnt += 1
+
+        if self.last_update == 0:
+            self.last_update = perf_counter()
+            self.startTimer(self.interval)
```
j9ac9k, line 156: Do you think we should specify the use of a precise timer here? 

http://doc.qt.io/qt-6/qt.html#TimerType-enum

pijyoi: No need. A coarse timer is sufficient because the actual time elapsed is calculated from perf_counter.

j9ac9k:
Love this diff, especially love doing away w/ the `.processEvents()` calls...I'll have to rework #2418 a fair amount but this PR makes that _much_ easier.  

Merging, thanks @pijyoi

PR #2755, opened 2023-06-25
Fix tox config, bump min numpy to 1.22

j9ac9k:
PR addresses 3 issues

- Fix `tox.ini` to run only on compatible configurations
- Remove usage of `numpy.product` which is deprecated as of NumPy 1.25 and will be removed in NumPy 2.0
- Bump minimum NumPy from 1.21 to 1.22 per NEP-29 timeline

### DIFF
--- a/.github/workflows/etc/requirements.txt
+++ b/.github/workflows/etc/requirements.txt
@@ -1,7 +1,7 @@
 # numpy based on python version and NEP-29 requirements
 numpy; python_version == '3.10'
 numpy; python_version == '3.11'
-numpy~=1.21.0; python_version == '3.9'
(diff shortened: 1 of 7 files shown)


j9ac9k:
Computer says no...

j9ac9k:
I can replicate the locally, only occurs on numpy 1.22.x, not 1.23.x.

pijyoi:
This workaround for `np.clip` on Windows (or rather MSVC) can also be removed.
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L1215-L1219

pijyoi:
Maybe add this link as a comment https://github.com/numpy/numpy/issues/21612

PR #2760, opened 2023-06-27
workaround Python 3.11.4 flag inversion issue

pijyoi:
CI is failing for Python 3.11.4 and PyQt6 due to https://github.com/python/cpython/issues/105497 and the next Python release is only expected to be in August (https://peps.python.org/pep-0664/#bugfix-releases).
Since there's only one place in pyqtgraph that triggers this bug, we can workaround it by using `xor` to toggle away the unwanted flag (even if it's only to make the CI go all green)


j9ac9k:
Thanks @pijyoi 

You make me feel lazy, here I was perfectly content waiting almost 2 months for the next python release before getting an all green CI.

PR #2599, opened 2023-01-29
BarGraphItem: calculate boundingRect without drawing

pijyoi:
This is a rewrite/restructuring of `BarGraphItem` to address some issues with the existing implementation:
1) `boundingRect`, `dataBounds` and `pixelPadding` all trigger the rendering code.
    * in this PR, bounding rectangle is calculated directly from the data values.
    * in this PR, pen width is calculated outside of the rendering code.
2) pens and brushes are always recreated even if only the data values were modified.
    * in this PR, pens and brushes are only recreated if they are modified
3) Even if the user passes in pre-created `QPen`s and `QBrush`es, a copy will be made.
    * in this PR, `QPen`s and `QBrush`es are not copied. The user is able to use the same `QPen` or `QBrush` instance for multiple bars.
4) Computation of the `QRectF`s are done one by one in a Python loop.
    * in this PR, numpy vector operations are used to calculate all the `QRectF`s
5) `drawRects()` is not used.
    * in this PR, `drawRects()` is used if only a single color is being used. This path also skips creation of `QPicture`.

A animated script to exercise and benchmark the various combinations of using `BarGraphItem`.
It seems to demonstrate a largish speedup vs master. 
```python
import itertools
import time
import numpy as np
import pyqtgraph as pg
from pyqtgraph.Qt import QtCore, QtGui, QtWidgets

pg.mkQApp()
glw = pg.GraphicsLayoutWidget()
glw.show()
plt1 = glw.addPlot()
plt2 = glw.addPlot()
glw.nextRow()
plt3 = glw.addPlot()
plt4 = glw.addPlot()

ttt = np.linspace(0, 1, 100)
s = np.sin(2*np.pi*1.0*ttt)
dt = ttt[1] - ttt[0]

# user pen and default brush
bgi1 = pg.BarGraphItem(x=ttt, width=dt, y1=s, y0=0, pen={'color':'w'})
# default pen and user brush
bgi2 = pg.BarGraphItem(y=ttt, height=dt, x1=s, x0=0, brush=(0,0,255,150))

# pre-made QBrush objects shared amongst all the bars
# pens are not pre-made and will be created by BarGraphItem
# (old implementation will re-create them every cycle)
pens = itertools.cycle([{'color': x} for x in 'rgby'])
pens = [next(pens) for _ in range(len(ttt))]
brushes = itertools.cycle([pg.mkBrush(255,0,0), pg.mkBrush(0,255,0), pg.mkBrush(0,0,255)])
[...]

PR #2614, opened 2023-02-13
Add docs for NonUniformImage

j9ac9k:
Recently realized that NonUniformImage didn't have an entry in our documentation.  This PR aims to correct that, and also makes some minor changes to NonUniformImage as well.

PR #2764, opened 2023-07-05
NonUniformImage improvements

pijyoi:
Behavior changes:
1) `NonUniformImage::setLookupTable` was previously implemented to take in a `HistogramLUTItem`. Now it takes in a numpy array, like `ImageItem`.

Improvements:
1) `NonUniformImage` can be used with both `HistogramLUTItem` and `ColorBarItem`, in the same way that an `ImageItem` is used. 
2) Change the example to use `ColorBarItem`. This also removes the hack of inserting a custom colormap into `GradientEditorItem`.
3) Speed up the drawing with `drawRects`. The speed up can be observed with the example at 
https://github.com/pyqtgraph/pyqtgraph/pull/2599#issuecomment-1432873261

Bugs fixed:
1) default color map was specified wrongly with floats
2) non-normalized QRectF: negative height was being computed
3) `HistogramLUTItem` was not using `ImageItem::getLevels()` accessor

### DIFF
--- a/pyqtgraph/graphicsItems/NonUniformImage.py
+++ b/pyqtgraph/graphicsItems/NonUniformImage.py
@@ -108,7 +108,7 @@ class NonUniformImage(GraphicsObject):
 
         # get colormap, lut has precedence over cmap
         if self.lut is None:
-            lut = self.cmap.getLookupTable()
+            lut = self.cmap.getLookupTable(nPts=256)
         elif callable(self.lut):
             lut = self.lut(z)
         else:
--- a/pyqtgraph/graphicsItems/NonUniformImage.py
+++ b/pyqtgraph/graphicsItems/NonUniformImage.py
@@ -1,12 +1,12 @@
-import math
-
+import warnings
 import numpy as np
 
 from .. import functions as fn
-from .. import mkBrush, mkPen
 from ..colormap import ColorMap
+from .. import Qt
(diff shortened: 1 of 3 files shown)


pyqtgraph/graphicsItems/NonUniformImage.py
```diff
@@ -72,62 +80,87 @@ def getHistogram(self, **kwds):
...
         return hist[1][:-1], hist[0]
 
-    def generatePicture(self):
+    def setLevels(self, levels):
+        self.levels = levels
+        self.picture = None
+        self.update()
 
-        x, y, z = self.data
+    def getLevels(self):
```
j9ac9k, line 88: Could this be renamed to just `levels()` (with `setLevels()` being the setter)?

pijyoi: ColorBarItem is looking for getLevels. To change to "levels()" would require corresponding changes to ColorBarItem, ImageItem, PColorMeshItem, HistogramLUTItem.

j9ac9k: Outside the scope of this PR to change all those other objects, but I'd like to deprecate that at point; but that's a future issue 👍

j9ac9k:
Thanks @pijyoi for revitalizing this plot type! I'll rebase #2614 so I can add docs to methods.

PR #2768, opened 2023-07-07
speed up PColorMeshItem

pijyoi:
By applying some of the techniques implemented in #2764, some speed up to PCMI has been obtained.

To make the speed up measurable, the example was modified to use `FrameCounter`.
This is not quite the same behavior as the original code, which was trying to cap the animation to 25 fps.


j9ac9k:
Thanks @pijyoi

I'm loving the impact the `FrameCounter` is having on simplifying our examples even more.

PR #2770, opened 2023-07-09
ScaleBar: allow setParentItem(None)

pijyoi:
This fixes an issue raised on the Discord channel that removing a `ScaleBar` would raise an exception.
The same bug is present in `LegendItem`. However the fix there might not be as trivial.

Note that `setParentItem()` is a void function, so there's no need to return its value.

MWE
```python
import pyqtgraph as pg

pg.mkQApp()

pw = pg.PlotWidget()
pw.show()
vb = pw.getViewBox()
sb = pg.ScaleBar(5, offset = (-30,-30))
sb.setParentItem(vb)
legend = pw.addLegend()
pdi = pw.plot(range(10), name='curve')

QtCore.QTimer.singleShot(1000, lambda: vb.removeItem(sb))
# QtCore.QTimer.singleShot(2000, lambda: vb.removeItem(legend))
pg.exec()
```

### DIFF
--- a/pyqtgraph/graphicsItems/ScaleBar.py
+++ b/pyqtgraph/graphicsItems/ScaleBar.py
@@ -57,12 +57,11 @@ class ScaleBar(GraphicsWidgetAnchor, GraphicsObject):
     def boundingRect(self):
         return QtCore.QRectF()
 
-    def setParentItem(self, p):
-        ret = GraphicsObject.setParentItem(self, p)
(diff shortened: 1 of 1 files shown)


pijyoi:
Not so trivial after all. `GraphicsWidgetAnchor::anchor()` would probably need to be modified to disconnect a `geometryChanged` signal.

PR #2776, opened 2023-07-13
Update ASV config to use paramers

j9ac9k:
* `renderImageItem` now uses parameterization better to keep track of attributes of a given test
*  The correct `cupy` version is better detected
*  Restructured so that only one environment is created; but different `PYQTGRAPH_QT_LIB`'s are cycled through forcing the use of different Qt bindings.

TODO:

- [x] Remove commented-out code
- [x] Re-add PyQt5, PySide2 and PySide6 environments

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,9 @@ python -m pip install numpy scipy pyqt6 -e .
 
 PyQtGraph supports PySide2, PyQt5, PySide6 and PyQt6 bindings, but if a contributor is
 to test against only one set of bindings, we suggest that it be PyQt6.  PyQt6 is the
-only bindings that enforces the use of fully scoped enums. For example, apart from PyQt6, `QGraphicsItem.ItemIgnoresTransformations` would be recognized, but that would error in PyQt6. The cross-binding compatible reference to that enum would be
+only bindings that enforces the use of fully scoped enums. For example, apart from
+PyQt6, `QGraphicsItem.ItemIgnoresTransformations` would be recognized, but that would
+error in PyQt6. The cross-binding compatible reference to that enum would be
 `QGraphicsItem.GraphicsItemFlags.ItemIgnoresTransformations`.
 
 Before making changes to the code-base, create a different branch with a name that
@@ -40,7 +42,8 @@ git switch -c my-new-feature
 ```
 
 The target of the pull request should be the `master` branch in the pyqtgraph repo.
-Pull requests should include only a focused and related set of changes. Mixed features and unrelated changes may be rejected.
+Pull requests should include only a focused and related set of changes. Mixed features
+and unrelated changes may be rejected.
 
 For major changes, it is recommended to discuss your plans on the mailing list or in a
 github issue/discussion before putting in too much effort.
@@ -121,7 +124,7 @@ To run the test suite, after installing the above dependencies run
 pytest tests
 ```
 
-In addition, the examples can be tested as well.  
+In addition, the examples can be tested as well.
 
 ```bash
 pytest pyqtgraph/examples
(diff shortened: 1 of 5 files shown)


benchmarks/renderImageItem.py
```diff
@@ -22,117 +23,156 @@ def renderQImage(*args, **kwargs):
...
+            for dtype in self.params.dtypes:
+                cache[xp.__name__]["data"][dtype] = {}
+                for size in self.params.sizes:
+                    if xp.dtype(dtype) in (xp.float32, xp.float64):
+                        data = xp.random.default_rng().standard_normal(size=size, dtype=dtype)
+                    else:
+                        iinfo = xp.iinfo(dtype)
+                        data = xp.random.default_rng().integers(
+                            low=iinfo.min,
+                            high=iinfo.max,
```
pijyoi, line 104: You need to add the argument `endpoint=True`.

benchmarks/renderImageItem.py
```diff
@@ -22,117 +23,156 @@ def renderQImage(*args, **kwargs):
...
+            dtype: npt.DTypeLike,
+            lut_dtype: typing.Optional[npt.DTypeLike]
+    ):
+        xp = np
+        self.name = f"{size=}_{acceleration=}_{use_levels=}_{dtype=}_{lut_dtype=}"
+        if acceleration == "numba":
+            if numba is None:
+                raise NotImplementedError
+            if lut_dtype is None and not use_levels:
+                # this is a fast code-path bypassing makeARGB
```
pijyoi, line 127: This comment is not correct. The parameters in this benchmark script should (hopefully) never invoke `makeARGB`. If the format of the rendered `QImage` is `ARGB32`, then it has invoked `makeARGB`.

benchmarks/renderImageItem.py
j9ac9k: thank you!

benchmarks/renderImageItem.py
j9ac9k: Thanks for pointing this out; I decided to skip the tests that are able to be rendered to QImage directly if running on non-numpy acceleration permutations of the parameters; as there is no point in testing that.

benchmarks/renderImageItem.py
```diff
@@ -22,117 +23,183 @@
...
+            pg.setConfigOption("useCupy", True)
+            xp = cp  # use cupy instead of numpy
+
+        # these codepaths bypass makeARRGB
+        if acceleration != "numpy":
+            if xp.dtype(dtype) == xp.ubyte and not use_levels:
+                if lut_length is None:
+                    # Grayscale8, RGB888 or RGB[AX]8888
+                    raise NotImplementedError
+                if channels == 1 and xp.dtype(lut_length) == xp.uint8:
```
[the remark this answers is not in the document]
j9ac9k: While this is technically correct; I had to scroll back a ways and think through it a fair amount before I agreed with the static code checker; I'm thinking this might be a case where I should disregard it and keep the check as-is.

benchmarks/renderImageItem.py
```diff
@@ -22,117 +23,183 @@ def renderQImage(*args, **kwargs):
...
+            # handle data caching
+            for dtype in self.params.dtypes:
+                cache[xp.__name__]["data"][dtype] = {}
+                for channels in self.params.channels:
+                    cache[xp.__name__]["data"][dtype][channels] = {}
+
+                    for size in self.params.sizes:
+                        size_with_channels = (size[0], size[1], channels) if channels != 1 else size
+                        if xp.dtype(dtype) in (xp.float32, xp.float64):
+                            data = xp.random.default_rng().standard_normal(
```
pijyoi, line 105: Calling `default_rng()` instantiates a new Generator object. Was that intended or could a single Generator object be used?

j9ac9k: I just went w/ the example in the docs; you're right, should probably just use a single generator object

CONTRIBUTING.md
```diff
@@ -78,20 +103,36 @@ python -m pytest pyqtgraph/examples tests
...
+To ensure this library is performant, we use 
+[Air Speed Velocity (asv)](https://asv.readthedocs.io/en/stable/) to run 
+benchmarks. For developing on core functions and classes, be aware of any 
+impact your changes have on their speed. To configure and run asv:
 
 ```bash
-pip install asv
+pip install asv virtualenv
 python setup.py asv_config
+asv config
```
ntjess, line 127: Error running this command:
```
(asv-test) PS C:\Users\...\pyqtgraph> asv config
usage: asv [-h] [--verbose] [--config CONFIG] [--version]
           {help,quickstart,machine,setup,run,dev,continuous,find,rm,publish,preview,profile,update,show,compare,check,gh-pages} ...
asv: error: argument {help,quickstart,machine,setup,run,dev,continuous,find,rm,publish,preview,profile,update,show,compare,check,gh-pages}: invalid choice: 'config' (choose from 'help', 'quickstart', 'machine', 'setup', 'run', 'dev', 'continuous', 'find', 'rm', 'publish', 'preview', 'profile', 'update', 'show', 'compare', 'check', 'gh-pages')
```

j9ac9k: That's what I get for going off memory, pretty sure it should be `asv setup`

ntjess: Do tests require python 3.11? If so, might be worth specifying in the docs
```
(asv-test) PS C:\Users\ntjess\Desktop\git\pyqtgraph> asv setup
· No executable found for python 3.11
· No executable found for python 3.11
· No executable found for python 3.11
```

j9ac9k: Huh; I'm fairly certain I tested on Python 3.11... haven't seen that error before tho

j9ac9k:
while I'm not big on extending this command via setup.py, migrating the asv setup command to pyproject.toml or something else will be part of a future PR.

PR #2778, opened 2023-07-21
Added turbo colormap to default Gradients of GradientEditorItem

ptuemmler:
Added turbo from https://ai.googleblog.com/2019/08/turbo-improved-rainbow-colormap-for.html to default Gradients dict in GradientEditorItem.py such that it can be used directly in ImageViews and their HistogramLUTItems.

Rainbow colormaps yield far greater contrast and depth differentiation than colormaps such as magma or viridis. Turbo is far more pleasant than the already available spectrum map and does not have the same banding issues as jet.


j9ac9k:
Would you consider adding the colormap in `csv` form here? https://github.com/pyqtgraph/pyqtgraph/tree/71bcf3f6d078b07a19b90791d96d27d28cb545d1/pyqtgraph/colors/maps

GradientEditorItem doesn't get as much usage as our ColorMap functionality.

ptuemmler:
Sure thing!

PR #2779, opened 2023-07-22
Allow menu selection of ColorMap(s)

pijyoi:
**UPDATE: the scope of this PR has evolved from the original `HistogramLevelsItem` (analog to `HistogramLUTItem`) to `ColorMapDisplayItem` (analog to `GradientEditorItem`). The contents of this comment should be considered outdated.**

`HistogramLevelsItem` is a simplified version of `HistogramLUTItem`.
Differences:
1) supports only mono-channel images
2) changes only levels, does not change LUT
3) uses only `ColorMap` property from `ImageItem`. Expects that `ColorMap` has been set.
4) does not make use of `GradientEditorItem`. Colormaps are non-editable.

The `NonUniformImage` example has been modified to use `HistogramLevelsItem`. It was originally using `HistogramLUTItem` but had been changed to `ColorBarItem` to avoid the monkey patch of color presets. Changing it to use `HistogramLevelsItem` makes the example closer to its original look.


pijyoi:
As there is too much code overlap between `HistogramLUTItem` and its derivative `HistogramLevelsItem`, it would make more sense to add an option to `HistogramLUTItem` instead to disable its `GradientEditorItem`. (`HistogramLevelsItem` will be removed from this PR)

When so disabled, the `GradientEditorItem` is replaced with a `DummyGradientWidget` that merely displays the colormap that has been **_pre-set_** on the `ImageItem`.

pijyoi:
To make the levels-only mode closer in behavior to `HistogramLUTItem`, the widget will no longer obtain the colormap by `ImageItem::getColorMap`. Instead, the user will set the colormap by `HistogramLUTItem.gradient.setColorMap`.

Some warts:
1) `enableGradientEditor` could use a better name
2) no way to use this levels-only mode in `ImageView`.

pijyoi:
An example of usage. 
It also demonstrates that using 256-entry LUTs have a large speed advantage over larger LUTs (according to `PYQTGRAPHPROFILE="ImageItem.paint"`).
Setting `enableGradientEditor=False` will, as an additional side-effect, unconditionally use 256-entry LUTs, whereas `enableGradientEditor=True` will use the original behavior of 512-entry LUTs.

```python
import numpy as np
import pyqtgraph as pg
pg.setConfigOption('useNumba', False)

rng = np.random.default_rng()
data = rng.random(size=(10240, 6144), dtype=np.float32)

pg.mkQApp()
glw.show()

imgitem = pg.ImageItem(axisOrder='row-major')
plot = glw.addPlot()
plot.addItem(imgitem)
hist = pg.HistogramLUTItem(image=imgitem, enableGradientEditor=False)
glw.addItem(hist)

cmap = pg.colormap.get('magma')
hist.gradient.setColorMap(cmap)
imgitem.setImage(data)

pg.exec()

print(imgitem.qimage.format())
```

pijyoi:
`ColorMapDisplayItem` is also usable standalone without `HistogramLUTItem`.
```python
import numpy as np
import pyqtgraph as pg

from pyqtgraph.graphicsItems.ColorMapDisplayItem import ColorMapDisplayItem

rng = np.random.default_rng()
data = rng.random(size=(256,256), dtype=np.float32)

pg.mkQApp()
glw.show()

imgitem = pg.ImageItem(image=data, axisOrder='row-major')
plot = glw.addPlot()
plot.addItem(imgitem)
cmdisp1 = ColorMapDisplayItem(orientation='vertical')
cmdisp2 = ColorMapDisplayItem(orientation='horizontal')
glw.addItem(cmdisp1)
glw.nextRow()
glw.addItem(cmdisp2)

cmdisp1.sigGradientChanged.connect(lambda cmdisp: imgitem.setColorMap(cmdisp.cmap))
cmdisp2.sigGradientChanged.connect(lambda cmdisp: imgitem.setColorMap(cmdisp.cmap))

pg.exec()
```

pijyoi:
This is how the example in https://github.com/pyqtgraph/pyqtgraph/pull/2779#issuecomment-1653758746 looks now
Screenshot 2023-08-05 122945

pijyoi:
It's actually possible for `ImageView` to make use of `ColorMapDisplayItem` by monkey-patching it in.
```python
import functools
import numpy as np
import pyqtgraph
import pyqtgraph as pg

HLWCM = functools.partial(pg.HistogramLUTWidget, useColorMaps=True)
pyqtgraph.imageview.ImageViewTemplate_generic.HistogramLUTWidget = HLWCM

rng = np.random.default_rng()
data = rng.rayleigh(size=(1024, 1024)).astype(np.float32)

pg.mkQApp()

imgitem = pg.ImageItem(data, axisOrder='row-major')
imv = pg.ImageView(imageItem=imgitem)
imv.show()

pg.exec()
```

pijyoi:
An alternative to replacing `GradientEditorItem` with `GradientShimItem` is to simply replace `GradientEditorItem`'s gradient selection context menu with `ColorMapMenu`, as is done in the last commit.

Advantages:
1) no need to supply extra option `useColorMaps=True`. Existing users automatically get access to the non-gradient colormaps.

Disadvantages (due to `GradientEditorItem`):
1) 256-entry colormaps will have 256 ticks rendered.
2) non-uint8 image data will request for 512-entry LUT, which will result in a less efficient `QImage` format of `RGBX8888` instead of `Indexed8`.

Other differences:
1) `GradientEditorItem` renders the colorbar as a `QLinearGradient` whereas `GradientShimItem` renders it as a 256-pixel `QImage`.

pijyoi:
The original `ColorMapParameter` has been renamed to `GradientParameter`
while `ColorMapExParameter` has been renamed to `ColorMapParameter`.

There is actually an `examples/ScatterPlotWidget.py` --> `widgets.ScatterPlotWidget` --> `widgets.ColorMapWidget` --> `parametertree.GradientParameter`. (where the arrow signifies "makes use of") (the new name `GradientParameter` is used to be less confusing)

If this is not a desirable change, it can be reverted.

j9ac9k:
Thanks for this PR @pijyoi sorry I let this sit idle for so long... this diff LGTM, ...having the gradient colormaps be different from the regular color maps always seemed less than...ideal.  Merging!

PR #2784, opened 2023-07-28
ImageItem: cache nan check

pijyoi:
While adjusting an `ImageItem`'s levels interactively with `HistogramLUTItem` or `ColorBarItem`, it is a waste to check for nans each time the _same_ image data gets re-rendered. 
(Arguably, checking it even once is a waste, since the common case is that there are _no_ nans.)

The example in https://github.com/pyqtgraph/pyqtgraph/pull/2779#issuecomment-1653758746 can be used to show that this PR gives a good speedup.


j9ac9k:
Thanks for the PR @pijyoi 

While there is a common case of no NaNs, NaNs have slipped into this library in a number of places I would not expect, the library should be able to behave in a predictable fashion if that's happening.

PR #2795, opened 2023-08-08
fix: wrong key set to _pixelVectorCache

pijyoi:
#1212 changed the lookup key used for the local cache from `dt` to a common `key` used also by the global cache.
The local cache update at the end of the function was not similarly modified to use `key`.
At worst, this causes the very _first_ local cache lookup _after_ a computation to always be a miss.

There appears to be another inconsistency present nearby. Elsewhere in the function (2 places), a copy of `pv` is returned. This copying is not done for the return at the end of the function. This was already the behavior prior to #1212.


j9ac9k:
Ha, a bug that got merged not by me 🎊 

Thanks for the PR @pijyoi nice find!  At one point, I was debugging something related and was suspicious of bogus caching... and noticed one of those caches seemed to never hit; but as that wasn't what I was debugging, I moved on...

j9ac9k:
I can't speak to if the intended behavior is to return a copy of the point or not tho...

PR #2796, opened 2023-08-09
Fix #2786: proper handling of slider parameter suffix

### LINKED ISSUE #2786: Bug in slider parameter
### Short description

Bug that precludes use of suffix in SliderParameter
### Code to reproduce

```python
import pyqtgraph as pg
import numpy as np
from pyqtgraph.parametertree import (Parameter, ParameterTree, registerParameterType)

        ...
        colorParams = [
            dict(name='Color Settings', type='group', children=[
                dict(name='Opacity', type='slider', limits=[0,100], value = 60.0, suffix='%'),
            ]),
        ]
        self.parameters = pg.parametertree.Parameter.create(name='Analysis Settings', type='group', children=colorParams)
        ...
```

### Expected behavior

Slider should show a number 0 - 100, followed by a '%' suffix

### Real behavior

Application crashes, as it refers to a non-existing method
```
An error occurred?

AttributeError
'SliderParameterItem' object has no attribute '_updateLabel'
```

### Solution
 1. Please delete line 32, containing ``` self._updateLabel(self.slider.value())``` as this method does not exist
 2. After ```layout = QHBoxLayout()``` Please add a line containing ```layout.setContentsMargins(0, 0, 0, 0)```
 3. In doing so, the height of the slider item stays equal to the other items (=20), now there is a *bulge*

### Additional context
None

ntjess:
Fixes #2786

Also avoids one unnecessary "valueChanged" signal just to update a label when the suffix (not the value) changes.

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes/slider.py
+++ b/pyqtgraph/parametertree/parameterTypes/slider.py
@@ -20,6 +20,7 @@ class SliderParameterItem(WidgetParameterItem):
     def updateDisplayLabel(self, value=None):
         if value is None:
             value = self.param.value()
+        self.sliderLabel.setText(self.prettyTextValue(self.slider.value()))
         value = str(value)
         if self._suffix is None:
             suffixTxt = ''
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for the PR @ntjess Merging!

PR #2807, opened 2023-08-26
Docs: Update Qt binding selection details

veractor:
- Match the binding selection order in documentation to the actual order. This follows up #2157, where it changed.

- Document the ability to set the environment variable `PYQTGRAPH_QT_LIB` for Qt binding selection.


j9ac9k:
@veractor thank you so much for updating the docs I keep forgetting that these docs needed updating

PR #2835, opened 2023-10-06
First PR!

### LINKED ISSUE #2331: Bug (?) from #2041
This new pattern in ``PlotCurveItem.setPen``, ``PlotCurveItem.setBrush``, and others has caused a number of breaks for me:

https://github.com/pyqtgraph/pyqtgraph/blob/4e350d03ac2007c47cbfcca9d3b77a6a2a638e5c/pyqtgraph/graphicsItems/PlotCurveItem.py#L373-L380

I believe it would be safer as

```python
if args and args[0] is None:
    self.opts['pen'] = None
```

As for now, I've had to change all my calls to give the color as a plain argument rather than keyword, which is fine, but inconsistent with the [docs for mkPen](https://pyqtgraph.readthedocs.io/en/latest/style.html#line-fill-and-color).

JayanthBontha:
Will now check if args is not None before accessing args[0]
Fixes #2331 

Hi! ,

I'm excited to contribute for the first time! 😄 This is my first pull request, and I'm eager to learn and improve. Pleasereview and suggest any changes you think are necessary. Your feedback will be highly appreciated.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -371,7 +371,7 @@ class PlotCurveItem(GraphicsObject):
 
     def setPen(self, *args, **kargs):
         """Set the pen used to draw the curve."""
-        if args[0] is None:
+        if args and args[0] is None:
             self.opts['pen'] = None
         else:
             self.opts['pen'] = fn.mkPen(*args, **kargs)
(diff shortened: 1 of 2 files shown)


j9ac9k:
Hi @JayanthBontha 

I'm sorry this sat for a month!  I was off the grid traveling, and then had a busy work trip right after; anyway the diff LGTM; merging and thank you for making pyqtgraph better!

PR #2836, opened 2023-10-08
AxisItem: Add adjustable label overlap tolerances, vertical overlap by default

NilsNemitz:
This extends / adjusts the functionality added in #2385 to stop axis labels sticking out of the allocated space. That certainly avoids overlap, it leaves less room for labels, and makes it so that ticks at the edges of the axis are never labeled. 

The main problem identified in #2385 was an overlap of the x-axis labels of adjacent plots in a layout. While the width of the x-axis plots grows for larger numbers, the height of the y axis labels remains constant, so overlap is much less problematic there.

This PR 
- changes the default AxisItem behavior to allow tick labels for the *vertical* axis to extend outside the boundary box.
- adds the option of passing an integer tolerance value for `hideOverlappingLabels` to allow a user specified amount of overlap: The default of 15 pixels tolerance for enabled overlap is too small at larger font sizes. This functionality was discussed for #2385, but wasn't ultimately added. The previously allowed values of True and False function the same as before.
- fixes a bug where the label tolerance accidentally remained hardcoded for the top axis.
- adds an example to show the effect of the setting and give demonstration code
- fixes an unrelated documentation error in `setTickSpacing()`

example AxisItem - label overlap

### DIFF
--- a/pyqtgraph/examples/AxisItem_label_overlap.py
+++ b/pyqtgraph/examples/AxisItem_label_overlap.py
@@ -1,0 +1,74 @@
+"""
+This example demonstrates many of the 2D plotting capabilities
+in pyqtgraph. All of the plots may be panned/scaled by dragging with 
+the left/right mouse buttons. Right click on any plot to show a context menu.
+"""
+
+import numpy as np
+
+import pyqtgraph as pg
+from pyqtgraph.Qt import QtGui
+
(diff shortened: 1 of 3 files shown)


j9ac9k:
Thanks @NilsNemitz sorry for the wait, this LGTM, merging!

PR #2837, opened 2023-10-08
fix: test Python 3.10 against Qt 6.2

pijyoi:
There are extraneous entries for the Python 3.10 CI matrix, with the result that both Python 3.10 and 3.11 were being tested against Qt 6.5.x, instead of (Python 3.10 with Qt 6.2.x) and (Python 3.11 with Qt 6.5.x).


j9ac9k:
Whoops, nice catch!

PR #2842, opened 2023-10-12
Give siPrefix behavior to values of zero

outofculture:
This PR adds new behavior to the SpinBox widget when `siPrefix` is True. By default, values of 0 will now take on the SI prefix associated with the initial step, as would immediately be present if the user were to press up or down once. This scaling hint will make it clearer what that initial step will be, as well as make it so that initial values typed in manually will be set to that scale. This behavior can also be overridden with a new `scaleAtZero` option, which can be configured to any value regardless of that initial step. Tests and example coverage are included.

This PR also corrects a small, unrelated incompatibility in the (likely not broadly used) GarbageCollector class, as well as adds a test that would have caught the problem.

### Other Tasks 

<details>

### Files that need updates
    
- [x] `README.md`
- [x] `setup.py`
- [x] `tox.ini`
- [x] `pyproject.toml`

</details>

<details>

### Pre Release Checklist

</details>

<details>

### Steps To Complete

- [ ] Announce on mail list
- [ ] Announce on Twitter

</details>


tests/widgets/test_spinbox.py
```diff
@@ -33,14 +37,27 @@ def test_SpinBox_formatting(value, expected_text, opts):
...
+    sb.editingFinishedEvent()
+    assert sb.value() == 100
+
+
 @pytest.mark.parametrize("suffix", ["", "V"])
 def test_SpinBox_gui_set_value(suffix):
     sb = pg.SpinBox(suffix=suffix)
 
-    sb.lineEdit().setText('0.1' + suffix)
+    sb.lineEdit().setText(f'0.1{suffix}')
```
j9ac9k, line 57: love the f-strings!

pyqtgraph/widgets/SpinBox.py
```diff
@@ -401,34 +405,35 @@ def stepEnabled(self):
...
+        ## note all steps (arrow buttons, wheel, up/down keys..) emit delayed signals only.
+        self.setValue(self._stepByValue(n), delaySignal=True)
 
-        n = decimal.Decimal(int(n))   ## n must be integral number of steps.
-        s = [decimal.Decimal(-1), decimal.Decimal(1)][n >= 0]  ## determine sign of step
+    def _stepByValue(self, steps):
+        if isinf(self.val) or isnan(self.val):
+            return self.val
+        steps = int(steps)
+        sign = [decimal.Decimal(-1), decimal.Decimal(1)][steps >= 0]
```
j9ac9k, line 415: Thanks for replacing single character variable names!

j9ac9k:
Hi @outofculture 

The PR looks good to me, I'm wondering if we need to go through a deprecation dance for removing the `prev` argument to `updateText` and `formatText` methods.

I would be open to having those methods accept `**kwargs` and have a check if `prev` is in the `kwargs` dictionary, and if present, emit a deprecation warning saying the argument is ignored, and it's presence will raise an error in a future version of pyqtgraph.

outofculture:
Here's that deprecation warning. Thanks for the review!

PR #2844, opened 2023-10-13
workaround PYSIDE-2487 for pen parameter

pijyoi:
This PR workarounds a regression in PySide 6.5.3 that affects code in pyqtgraph's pen parameter.
(https://bugreports.qt.io/projects/PYSIDE/issues/PYSIDE-2487)

Note: Qt 6.5 is LTS and PySide 6.5.3 will be the last version to have non-commercial wheels.

Script to demonstrate that the proposed fix does disconnect the slot
```python
from PySide6 import QtCore

class Parent(QtCore.QObject):
    sig = QtCore.Signal(object)

    def pslot(self, val):
        pass

class Child(Parent):
    pass

x = Child()
x.sig.connect(x.pslot)

# disconnecting slot defined in parent fails since PySide >= 6.5.3
# x.sig.disconnect(x.pslot) 

# disconnecting all slots works
x.sig.disconnect()

# show that slots are disconnected
x.dumpObjectInfo()
```

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes/pen.py
+++ b/pyqtgraph/parametertree/parameterTypes/pen.py
@@ -192,6 +192,7 @@ class PenParameter(GroupParameter):
                     # that affects PySide 6.5.3
                     # Since the child param was freshly created by us, there can only have been one slot
                     # connected. So we can just disconnect all the slots without specifying which one.
+                    assert p.receivers(QtCore.SIGNAL("sigValueChanged(PyObject,PyObject)")) == 1
                     p.sigValueChanged.disconnect()
                 # Some widgets (e.g. checkbox, combobox) don't emit 'changing' signals, so tie to 'changed' as well
                 p.sigValueChanged.connect(newSetter)
--- a/pyqtgraph/parametertree/parameterTypes/pen.py
+++ b/pyqtgraph/parametertree/parameterTypes/pen.py
@@ -185,7 +185,14 @@ class PenParameter(GroupParameter):
                 if p.type() != 'color':
                     p.sigValueChanging.connect(newSetter)
                 # Force children to emulate self's value instead of being part of a tree like normal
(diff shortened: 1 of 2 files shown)


pyqtgraph/parametertree/parameterTypes/pen.py
```diff
@@ -185,7 +185,14 @@ def _makeChildren(self, boundPen=None):
                 if p.type() != 'color':
                     p.sigValueChanging.connect(newSetter)
                 # Force children to emulate self's value instead of being part of a tree like normal
-                p.sigValueChanged.disconnect(p._emitValueChanged)
+                try:
+                    p.sigValueChanged.disconnect(p._emitValueChanged)
+                except RuntimeError:
+                    # workaround https://bugreports.qt.io/projects/PYSIDE/issues/PYSIDE-2487
+                    # that affects PySide 6.5.3
+                    # Since the child param was freshly created by us, there can only have been one slot
```
ntjess, line 193: Is it possible to guarantee this with `assert p.receivers("_emitValueChanged(object, object)") == 1`?

I tried various adaptations of arguments to `receivers` according to the [docs](https://doc.qt.io/qtforpython-6/PySide6/QtCore/QObject.html#PySide6.QtCore.PySide6.QtCore.QObject.receivers) example, but it always returns 0 when I try to test with a MWE. Do you know whether it works in Python? There aren't too many examples online...

pijyoi: For PySide, the syntax is: `assert p.receivers(QtCore.SIGNAL("sigValueChanged(PyObject,PyObject)")) == 1`
The PySide syntax specific things are:
1) PyObject
2) QtCore.SIGNAL: PyQt has dropped it since moving to the new signal / slot syntax

j9ac9k:
Thanks @pijyoi !

PR #2845, opened 2023-10-15
[CI] add mypy settings for CI

longqzh:
This is the first PR in the PR's sequence to add type annotation.
The settings of mypy in `pyproject.toml` will be updated later, and mypy ignores all of the errors currently

@j9ac9k ,
I think that using the `ignore_errors`, `ignore_missing_imports` settings in mypy can add type annotation incrementally easier than run some custom script. Let me know if I misunderstand :)

@ksunden ,
This PR is an early stage, everything may be changed. If you have better idea, let me know~

Fixes #2833 

### Other Tasks 

<details>

### Files that need updates
    
- [x] `README.md`
- [x] `setup.py`
- [x] `tox.ini`
- [x] `pyproject.toml`

</details>

<details>

### Pre Release Checklist

</details>

<details>

### Steps To Complete

- [ ] Announce on mail list
- [ ] Announce on Twitter

</details>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,6 +15,19 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  run-mypy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install mypy
+        run: pip install mypy
+      - name: Run mypy
+        run: mypy
   test-pip:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,3 +34,14 @@ known_third_party = ["QtCore", "QtGui", "QtWidgets"]
 [tool.pycln]
 all = true
 exclude = '(Template|__init__.py)'
+
+[tool.mypy]
+packages = ["pyqtgraph"]
+show_error_codes = true
+strict = true
+disable_error_code = ["attr-defined"]
+ignore_errors = true
+
+[[tool.mypy.overrides]]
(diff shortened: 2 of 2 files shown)


j9ac9k:
Thanks for the PR @longqzh !

The only question I have right now is does this setup actually check anything? For example, if we were to add a wrong type-annotation in the code-base, would this check catch that?

longqzh:
@j9ac9k, in the current PR, No, this won't check anything because I set `ignore_errors` option to `true` for all of source file.

And as I said, type annotation will be made up of dozen PRs. The [next PR I'm working on](https://github.com/longqzh/pyqtgraph/commits/add-type-flowchart), will check the `flowchart` folder. I will work on one folder per PR :)

.github/workflows/main.yml
```diff
@@ -15,6 +15,19 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  run-mypy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
```
j9ac9k, line 24: Is there a reason not to use 3.12? It should provide access to newer type annotation syntax, and some faster performance checks due to performance boosts.

j9ac9k:
Hi @longqzh 

PR LGTM, I would suggest using the newest available python version for the mypy check; but that's my only suggestion 👍🏻

.github/workflows/main.yml
longqzh: Because pyqtgraph should support Python3.9+.  And I use Python3.9 for development to avoid backward compatibility issue.

for example, 
```python
a: int | str = 10
```
is OK in python3.10, but raise `TypeError` in 3.9

So using lower version's python to check the type makes sense.
Let me know your option, if you want, I can update it to 3.10 :)

j9ac9k: oh right, yeah probably should check on the oldest supported version...

PR #2848, opened 2023-10-17
Fix Python 3.12 `DeprecationWarning`: `datetime.datetime.utcfromtimestamp()`

### LINKED ISSUE #2847: Python 3.12 `DeprecationWarning`: `datetime.datetime.utcfromtimestamp()`
On Python 3.12, I get
> E   DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).

bersbersbers:
`utcfromtimestamp()` is deprecated in Python 3.12. Fixes #2847.

### DIFF
--- a/pyqtgraph/graphicsItems/DateAxisItem.py
+++ b/pyqtgraph/graphicsItems/DateAxisItem.py
@@ -23,7 +23,8 @@ if sys.platform == 'win32':
     def utcfromtimestamp(timestamp):
         return _epoch + timedelta(seconds=timestamp)
 else:
-    utcfromtimestamp = lambda timestamp: datetime.utcfromtimestamp(timestamp, UTC)
+    def utcfromtimestamp(timestamp):
+        return datetime.fromtimestamp(timestamp, UTC)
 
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for the PR @bersbersbers !

Is there a reason we can't remove the platform specific if-statement and just use `datetime.fromtimestamp(0, timezone.utc)` ?

bersbersbers:
> Is there a reason we can't remove the platform specific if-statement and just use `datetime.fromtimestamp(0, timezone.utc)` ?

I haven't tried, to be honest. The platform-specific code was there before, and in an effort to affect as little code as possible while on my phone, I did not touch that part.

bersbersbers:
(But feel free to edit/take over this PR.)

j9ac9k:
Taking a closer look, there could definitely be some room for cleanup; but that would involve a change that's far outside the scope of removing a deprecated method such as this PR aims to do.

This LGTM, thanks for the PR @bersbersbers !

PR #2850, opened 2023-10-18
Make sure the text is not croped

### LINKED ISSUE #2849: DockLabel are croped
### Short description
I'm using a lot the Docks and it's super nice tools. There is however an issue with their DockLabel. Indeed the Vetical label they are inheriting from is not properly taking into account the size of the label (in particular its font size). The MaximumHeight property of the Vertical label seems always too small. As you can see below 

image

when applying a qdarkstyle to it, it's even worse  (and the issue come from pyqtgraph as you have hardcoded the style in the VerticalLabel class):

image

### Code to reproduce

```python
import sys
from qtpy import QtWidgets, QtCore
from pyqtgraph.dockarea import Dock, DockArea


if __name__ == '__main__':

    area = DockArea()
    dock1 = Dock('ggg lll jjj')
    dock2 = Dock('ppp, ggg, zzz')
    area.addDock(dock1)
    area.addDock(dock2, position='right')
    area.show()
    sys.exit(app.exec())
```

and with qdarstyle:

```python
import sys
from qtpy import QtWidgets, QtCore
from pyqtgraph.dockarea import Dock, DockArea


if __name__ == '__main__':
    import qdarkstyle
    app.setStyleSheet(qdarkstyle.load_stylesheet(qdarkstyle.DarkPalette))
    area = DockArea()
    dock1 = Dock('ggg lll jjj')
    dock2 = Dock('ppp, ggg, zzz')
    area.addDock(dock1)
    area.addDock(dock2, position='right')
    area.show()
    sys.exit(app.exec())
```

### Expected behavior
I would expect the label area to expand to let people read completely the label text

### Real behavior
More or less part of the text is being cropped

### Possible Patch
I had to set the MaximumHeight of the VerticalLabel to some value to solve the issue

### Tested environment(s)

 * PyQtGraph version: 0.13.3
 * Qt Python binding: qtpy with PyQt5
 * Python version: 3.10.12
 * NumPy version: 1.25.2
 * Operating system: Win10
 * Installation method: pip

### Additional context

seb5g:

Fixes #2849 

I just added a fixed number (6) to the MaximumWidth or Height of the VerticalLabel with respect to the label size hint. 

### Other Tasks 

<details>

### Files that need updates
    
- [ X] `README.md`
- [ X] `setup.py`
- [ X] `tox.ini`
- [X ] `.github/workflows/main.yml` and associated `requirements.txt` and conda `environemt.yml` files
- [ X] `pyproject.toml`
- [ X] `binder/requirements.txt`

</details>

<details>

### Pre Release Checklist

</details>

<details>

### Steps To Complete

- [ ] Announce on mail list
- [ ] Announce on Twitter

</details>


j9ac9k:
I agree that this is an issue, but I'm not sure this is the fix we want... I generally try and avoid hard-coding pixels to oversize fields and instead try and set the margins correctly.

As we can see, the height/width is set based on the `self.hint` attribute, which is set here:

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/widgets/VerticalLabel.py#L52

...but looking at the Qt documentation, I'm really confused, `QPainter::drawText` should return None... but it is returning a `QRect`? ... I'll investigate a bit more.

seb5g:
I agree with your comment but somehow something seems to be wrong on the qt side. I think the sizehint is often wrong as many times the labels within a widget are not properly visible.

As for the QPainter::drawTest thing, it is because of pyqt mechanism:

You see that the overloaded method:
```
void QPainter::drawText(const QRectF &rectangle, int flags, const QString &text, QRectF *boundingRect = nullptr)
```
take as third argument a pointer to a QRectF. In the usual pyqt5/6 wrapper this translates into a returned value being this QRectF. This is why the hint has been set as the returned value. 

Did you see another possible fix?

seb5g:
running this minimal example reproduce the result:
```
if __name__ == '__main__':
    import sys

    label = VerticalLabel('mysuperlabel', orientation='hor')
    label.setAlignment(QtCore.Qt.AlignmentFlag.AlignTop | QtCore.Qt.AlignmentFlag.AlignHCenter)
    f = label.font()
    f.setPixelSize(25)
    label.setFont(f)
    label.show()
    sys.exit(app.exec())
```
which shows the croping
image


 while this is working fine:

```
    import sys

    label = QtWidgets.QLabel('mysuperlabel')
    label.setAlignment(QtCore.Qt.AlignmentFlag.AlignTop | QtCore.Qt.AlignmentFlag.AlignHCenter)
    f = label.font()
    f.setPixelSize(25)
    label.setFont(f)
    label.show()
    sys.exit(app.exec())
```
image

seb5g:
Allright I found out the bug, 
in fact the sizeHint method is never called during the label creation, therefore the size cannot be set to the boundingrect as discussed above. Just adding a statement with the resize method having the sizeHint returned value as argument solves our issue.
I'll push some code in that direction

seb5g:
well it seems however that the styling in the DockLabel is messing up things also, there is a border width (bottom) in there that is hard coded to 2pxl and this is also explaining why the bottom of the writting are wrong. However I could not figure how to properly solve this... could you have a look in those directions?

seb5g:
@j9ac9k could you have a look at what I dug out?

PR #2853, opened 2023-10-20
Add updatePristineValues arg to Parameter.setDefault

outofculture:
When a Parameter has its default changed, the value could change to match unless it has previously been modified by the user. The `updatePristineValues` argument will turn this behavior on, currently defaulting to False.

- This also adds the `valueModifiedSinceResetToDefault` method, a necessary part of the implementation, but also a nice feature for possible application logic.
- This also adds the `hasValue` method, to be better able to distinguish when a Parameter has an explicit value set.
- This starts the deprecation countdown on being able to call without error `value()` and `defaultValue()` while they're respectively null.
- This completes the deprecation of being able to use `'values'` to specify the limits of a ListParameter.
- This fixes a bug in PenParameter where the Qt setter methods were not being accessed properly.
- This changes the behavior of the reset-to-default gui button: it will now be enabled if the parameter is in a dirty state relative to the default, not just an equal value comparison.
- Various small cleanups and improvements of related code are included in this.

### Other Tasks 

<details>

### Files that need updates
    
- [x] `README.md`
- [x] `setup.py`
- [x] `tox.ini`
- [x] `pyproject.toml`

</details>

<details>

### Pre Release Checklist

</details>

<details>

### Steps To Complete

- [ ] Announce on mail list
- [ ] Announce on Twitter

</details>


pyqtgraph/parametertree/Parameter.py
```diff
@@ -221,6 +227,7 @@ def __init__(self, **opts):
         if 'default' not in self.opts:
             self.opts['default'] = None
             self.setDefault(self.opts['value'])
+        self._modifiedSinceReset = not self.valueIsDefault()
```
campagnola, line 230: I don't think this is right -- if the initial value happens to be the same as the default, then it would end up pinned. Seems like it should be explicitly reset to default for this to happen?

pyqtgraph/parametertree/Parameter.py
```diff
@@ -20,9 +23,17 @@ def registerParameterItemType(name, itemCls, parameterCls=None, override=False):
     """
     global _PARAM_ITEM_TYPES
     if name in _PARAM_ITEM_TYPES and not override:
-        raise Exception("Parameter item type '%s' already exists (use override=True to replace)" % name)
-
-    parameterCls = parameterCls or Parameter
+        raise Exception(
```
j9ac9k, line 26: Can we raise something other than a bare exception here?

j9ac9k:
@ntjess want to look over this one? 😅

pyqtgraph/parametertree/Parameter.py
```diff
@@ -39,11 +50,17 @@ def registerParameterType(name, cls, override=False):
...
     PARAM_NAMES[cls] = name
 
+
 def __reload__(old):
     PARAM_TYPES.update(old.get('PARAM_TYPES', {}))
     PARAM_NAMES.update(old.get('PARAM_NAMES', {}))
 
-class Parameter(QtCore.QObject):
+
+class _ParameterMeta(ABCMeta, type(QtCore.QObject)):
```
j9ac9k, line 59: Do we really want to use metaclasses here and not something like `typing.Protocol`?

outofculture: I had some trouble combining `QObject` inheritance and abstract classes. It might be possible without an explicit metaclass, though, and I just couldn't figure it out.

j9ac9k: I haven't used metaclasses that much, I know they're somewhat more ....controversial in python-land than in other languages.

@ksunden you have any thoughts on the use here?

pyqtgraph/parametertree/parameterTypes/numeric.py
```diff
@@ -27,7 +27,7 @@ def makeWidget(self):
         for k in defs:
             if k in opts:
                 defs[k] = opts[k]
-        if 'limits' in opts:
+        if opts.get('limits') is not None:
```
j9ac9k, line 30: is this change meant to address the case where a `limits` key is preset in `opts` but the value of that entry could be `None`?

ntjess:
Is there a reason `Parameter` needs to be abstract? It can be valuable to track a value, state, and changed signals even if there isn't a defined type associated

pyqtgraph/parametertree/parameterTypes/numeric.py
outofculture: Yeah, in my playing with the new feature, i was auto-generating a bunch of param specs, and setting `'limits'` to `None` blew up. I guess I should make that an actual test, shouldn't I.

pyqtgraph/parametertree/Parameter.py
outofculture: no need; the path to being abstract has been abandoned :)

pyqtgraph/Qt/__init__.pyi
```diff
@@ -5,32 +5,40 @@ This stub file is to aid in the PyCharm auto-completion of the Qt imports.
...
 
             QtCore = QtCore
             QtGui = QtGui
             QtWidgets = QtWidgets
+            QtTest = QtTest
+            QtSvg = QtSvg
         except ImportError:
             try:
-                from PySide6 import QtCore, QtGui, QtWidgets
+                from PySide6 import QtCore, QtGui, QtSvg, QtTest, QtWidgets
```
j9ac9k, line 35: Curious what the motivation here is for these extra imports? I'm not seeing them referenced elsewhere in the PR.

pyqtgraph/parametertree/Parameter.py
```diff
@@ -1,5 +1,7 @@
 import re
+import warnings
 import weakref
+from abc import ABCMeta
```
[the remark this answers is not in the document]
j9ac9k: go ahead and remove the unused import 👍🏻

pyqtgraph/parametertree/Parameter.py
j9ac9k: wait, this shows as resolved, but still shows as a bare exception, some git mixup?

pyqtgraph/Qt/__init__.pyi
outofculture: I noticed while looking at a test that the `pyi` file was incomplete; we already import both of these in the actual `Qt/__init__.py` file. Though, now that I'm looking at it closer, they're both imported in blocks of
```python
try:
    from PySide6 import QtTest
except ImportError as err:
    QtTest = FailedImport(err)
```
I don't know why `QtTest` wouldn't be available for import, but I'm guessing that maybe there are stripped-down distributions that don't include it? I don't know what that should mean for the `pyi` file, though. Probably any development environment is going to include the full qt library, which is what this file is intended for. However, the type-checking is going to run everywhere, regardless, and might throw an `ImportError`. Maybe I should just skip this and accept the lack of type hints... Thoughts?

pyqtgraph/parametertree/Parameter.py
outofculture: Ha! I fixed an identical problem elsewhere in the file. This one's fixed now, too.

pyqtgraph/Qt/__init__.pyi
j9ac9k: Huh, `QtTest` is shown as being present in `PySide6-essentials`   https://pypi.org/project/PySide6-Essentials/

That's weird it's not included, would it be under a different module or something?

j9ac9k: the import worked fine on my machine just now?

```
In [1]: from PySide6 import QtTest

In [2]: import PySide6

In [3]: PySide6.__version__
Out[3]: '6.6.2'

In [4]:
```

outofculture: okay then, I say we tentatively go forward with importing like this it in the `pyi` and expecting that nothing will blow up. I'll put in an issue to clean up `Qt.py` to keep it isolated.

PR #2860, opened 2023-10-28
fixed recursive PlotWidget.__getattr__ calls

nleclercq:
PlotWidget.\_\_getattr\_\_ implicitly wrap methods from PlotItem. However, the current implementation leads to recursive calls
and 'maximum recursion depth exceeded' errors when PlotWidget is used in a multiple inheritance schema - i.e., class SomeClass(PlotWidget, SomeOtherClass). The proposed PlotWidget.\_\_getattr\_\_  prevents the recursive calls by get rid of the hasattr that is the source of the problem.

### DIFF
--- a/pyqtgraph/widgets/PlotWidget.py
+++ b/pyqtgraph/widgets/PlotWidget.py
@@ -44,6 +44,9 @@ class PlotWidget(GraphicsView):
     other methods, use :func:`getPlotItem <pyqtgraph.PlotWidget.getPlotItem>`.
     """
     def __init__(self, parent=None, background='default', plotItem=None, **kargs):
+        ## start by instantiating the plotItem attribute in order to avoid recursive 
+        ## calls of PlotWidget.__getattr__ - which access self.plotItem!
+        self.plotItem = None
         """When initializing PlotWidget, *parent* and *background* are passed to 
         :func:`GraphicsWidget.__init__() <pyqtgraph.GraphicsWidget.__init__>`
         and all others are passed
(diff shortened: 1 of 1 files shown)


pyqtgraph/widgets/PlotWidget.py
```diff
@@ -74,11 +74,17 @@ def close(self):
...
+    ## implicitly wrap methods from plotItem
+    ## the implementation prevents recursive calls to PlotWidget.__getattr__ that lead to a 
+    ## 'maximum recursion depth exceeded' when PlotWidget is used in a multiple inheritance 
+    ## schema - i.e., class SomeClass(PlotWidget, SomeOtherClass)
+    def __getattr__(self, attr): 
+        try:
             m = getattr(self.plotItem, attr)
-            if hasattr(m, '__call__'):
-                return m
+        except:
```
j9ac9k, line 84: Can you specify which exception to catch here instead of the bare exception?

nleclercq: Well, in fact, no need to modify the implementation of PlotWidget.\_\_getattr\_\_. Moreover, the problem has nothing to do with multiple inheritance. It comes from the fact that the   PlotWidget.\_\_getattr\_\_ tries to access the PlotItem attribute before it actually exists - generating recursive calls  PlotWidget.\_\_getattr\_\_ trying to access self.PlotItem. 
The simplest solution is to instantiate the PlotItem attribute at the very beginning of the  PlotWidget.__init__ and we are done withe the problem.

def __init__(self, parent=None, background='default', plotItem=None, **kargs):
        ## start by instantiating the plotItem attribute in order to avoid recursive 
        ## calls of PlotWidget.__getattr__ - which access self.plotItem!
        self.plotItem = None
        ...

j9ac9k:
Love the modified PR, creating the attribute at initialization is a good way to go; not sure why we just don't instantiate the PlotItem instance at `__init__` but I feel like that might have other consequences outside the scope of this PR; so if this fixes your issue, then leave as is 👍🏻

nleclercq:
BTW, when do you plan to publish a new release?

j9ac9k:
Haha probably soon as we have fixed issues with recent bindings. I
introduced a bug in the SVG exporter I want to fix first.  Been doing a lot
of travel (recreational and for work) recently which has really put a
damper on things. I don’t have any scheduled travel for a while… so
hopefully in a week or two.

Issue #2854, opened 2023-10-25
Blurry scatter plot on high DPI monitor


kaktus018:
### Short description
I'm experiencing an issue with blurry scatter plots on high DPI monitors. The plots look fine on a HD screen, but with a resolution of 2560x1440 and the monitor scaling set to 125% in the Windows settings, the single points have blurry outlines.

### Code to reproduce
The following code makes two windows, one with pxMode=True and the other with pxMode=False and antialias=False.
```
import pyqtgraph as pg

plotWidget = pg.plot()
plotWidget.setBackground('w')
scatter = pg.ScatterPlotItem(size=10, symbol='o', pen=None, brush=pg.mkBrush((0,115,190)))
x = [1,2,3,4,5,6,7,8,9,10]
y = [5,4,3,2,1,6,7,8,9,10]
scatter.addPoints(x, y)
plotWidget.addItem(scatter)

plotWidget = pg.plot()
plotWidget.setBackground('w')
scatter = pg.ScatterPlotItem(size=0.5, pxMode=False, antialias=False, symbol='o', pen=None, brush=pg.mkBrush((0,115,190)))
scatter.addPoints(x, y)
plotWidget.addItem(scatter)
```
<img src="https://github.com/pyqtgraph/pyqtgraph/assets/26413439/ea51980e-dba8-4909-866a-e7084e395c52.png" width=50% height=50%>

The image to the right is with pxMode=False and does not have blurry points. However, I want to have the points properly scaled, which is not possible with pxMode=False (as far as I know). I suspect that the blurriness originates from antialiasing, which cannot be turned off when pxMode=True.

Any ideas?

### Tested environment(s)

 * PyQtGraph version: 0.13.3
 * Qt Python binding: PySide6 6.6.0 Qt 6.6.0
 * Python version: 3.11.4
 * NumPy version: 1.25.2
 * Operating system: Windows 10

pijyoi:
I find that using `pxMode=True` (default True) together with `useCache=False` (default True) removes the blurriness.

kaktus018:
Thanks for the remark! Indeed, the points look much better when using `useCache=False`.

PR #2863, opened 2023-10-30
ScatterPlotItem: implement hidpi pixmap

pijyoi:
This PR came about while investigating #2854.
Although a workaround for #2854 was to disable `useCache`, the true reason for the blurriness on hidpi displays is due to the pixmap atlas only being rendered at a dpi of 1.0.

1) The improved sharpness of this PR vs master (on a hidpi display) can be seen from the `ScatterPlot.py`, `Symbol.py` examples.
2) On a hidpi display, this PR achieves a higher frame-rate than master on `ScatterPlotSpeedTest.py`. i.e. there is a drawing penalty when a lo-res pixmap needs to be scaled to be painted onto a hidpi display.

### DIFF
--- a/pyqtgraph/graphicsItems/ScatterPlotItem.py
+++ b/pyqtgraph/graphicsItems/ScatterPlotItem.py
@@ -85,7 +85,7 @@ def drawSymbol(painter, symbol, size, pen, brush):
     painter.drawPath(symbol)
 
 
-def renderSymbol(symbol, size, pen, brush, device=None):
+def renderSymbol(symbol, size, pen, brush, device=None, dpr=1.0):
     """
     Render a symbol specification to QImage.
     Symbol may be either a QPainterPath or one of the keys in the Symbols dict.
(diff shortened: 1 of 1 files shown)


j9ac9k:
I think you can programatically get that ratio you need by doing the following inside the paint method

```python
dpi = self.getViewWidget().metric(QtGui.QPaintDevice.PaintDeviceMetric.PdmPhysicalDpiX)
pixmap_dpi = self.fragmentAtlas.pixmap.metric(QtGui.QPaintDevice.PaintDeviceMetric.PdmPhysicalDpiX)

scale = dpi / pixmap_dpi
```

Likely don't want to be calling this during every paint call...

pijyoi:
>I think you can programatically get that ratio you need by doing the following inside the paint method
>dpi = self.getViewWidget().metric(QtGui.QPaintDevice.PaintDeviceMetric.PdmPhysicalDpiX)
pixmap_dpi = self.fragmentAtlas.pixmap.metric(QtGui.QPaintDevice.PaintDeviceMetric.PdmPhysicalDpiX)
>scale = dpi / pixmap_dpi

Although we call it "hi-dpi", the quantity that we are interested in is the DPR rather than the DPI.

pijyoi:
As a bonus, this PR also fixes the following artifacts that appear for the bottom right plot of `ScatterPlot.py` example the moment the mouse is moved over it.
(Windows 10 device pixel ratio 1.25, PySide6 6.6.0)
before_br

j9ac9k:
Thanks @pijyoi for the PR, this LGTM merging!

PR #2867, opened 2023-11-01
[type annotation] add type hint for flowchart

longqzh:
This is the second PR in the PR's sequence to add type annotation.
type hints are added in `flowchart`folder.

Discussion : #2833

### Other Tasks 

<details>

### Files that need updates
    
- [x] `README.md`
- [x] `setup.py`
- [x] `tox.ini`
- [x] `pyproject.toml`

</details>

<details>

### Pre Release Checklist

</details>

<details>

### Steps To Complete

- [ ] Announce on mail list
- [ ] Announce on Twitter

</details>


longqzh:
@j9ac9k, after adding the type hint, the github CodeQL start to scan the code, and reject the PR because mismatch override etc
Those potential problems are caused by code itself, and all fo them are marked by `# type: ignore` in mypy.

So how about disable the CodeQL before we're ready?

j9ac9k:
Depending on the reason codeql is failing, we can merge with it giving a red checkmark.  CodeQL is intended to be a check for some basic stuff like unused imports; referencing potentially unassigned variables and so on.

That said the reason for the errors is using `typing.Unpack` which was added in python 3.11.

I would be good with adding typing-extensions as a dependency if using new type syntax results in inline annotations that are far easier to read/simplified.

longqzh:
@j9ac9k , Good catch! `typing.Unpack` cannot use in python 3.9. But I don't think only adding typing-extensions can solve this issue.

Firstly, codeQL can support python 3.5~3.11 [codeql-overview](https://codeql.github.com/docs/codeql-overview/supported-languages-and-frameworks/),

Secondly, codeQL said "Overriding method process has signature mismatch with overridden method",
in  the parent class `Node`, the `process` method's signature is
```python
    def process(self, **kargs: Unpack) -> dict:
```

but in the child class (like `ImageViewNode`), the signature is
```python
def process(self, data: Optional[np.ndarray], display: bool = True) -> None:
```

the signature of method in the parent & child class are **Different**. I guess this is the real reason to fail.
I'll try to find other way to solve it, (ex, remove type hint in `process` method). and will share the progress :)

j9ac9k:
I wouldn't worry about the CodeQL issue, it being red on a PR is not a deal breaker on it merging; there are inherent shortcomings w/ that check.

Right now, there are 418 CodeQL issues in the codebase (there used to be ~1,000 but periodically I make a PR to address some of the low-hanging fruit).

longqzh:
You're right, I didn't check the detail of failed checks.
I'll fix the `Unpack` issue :)

longqzh:
@j9ac9k , Thanks for your explanation :)

I have added `type-extensions` dependency in `setup.py` and `main.yaml`(for conda)
let me know if I should update any more file

longqzh:
@j9ac9k , please merge it if it's no problem :)

j9ac9k:
Hi @longqzh

Sorry for the silence on my end.  I'm moved across the country, and am preparing for an international move in a few months so my time on this library came to a standstill for a while.

Given the [NEP-29 timeline](https://numpy.org/neps/nep-0029-deprecation_policy.html) we will be dropping Python 3.9 support is just a few months.  This will take away the need for using the `typing_extensions` module.  I'm going to hold off on merging this PR until we drop Python 3.9 support.  Feel free to remove the `typing_extensions` module.  I understand if enough time has passed and you've moved on with your own projects, I would be good with taking over this PR too 👍🏻

pyqtgraph/flowchart/library/Display.py
```diff
@@ -11,24 +14,24 @@ class PlotWidgetNode(Node):
...
+
+    def __init__(self, name: str) -> None:
         Node.__init__(self, name, terminals={'In': {'io': 'in', 'multi': True}})
-        self.plot = None  # currently selected plot 
-        self.plots = {}   # list of available plots user may select from
-        self.ui = None 
-        self.items = {}
-        
-    def disconnected(self, localTerm, remoteTerm):
+        self.plot: Optional[PlotWidget] = None  # currently selected plot
```
rwarren, line 20: Instance properties should be typed at the class level, not in the `__init__` function (see [here](https://mypy.readthedocs.io/en/stable/class_basics.html#instance-and-class-attributes))... like this:

```python
class PlotWidgetNode(Node):
    items: dict[int, QGraphicsItem | PlotDataItem
    plot: PlotWidget | PlotItem | None  # personal preference, but I only use `Optional` for function args
    plots: dict[str, PlotWidget | PlotItem]  
    ui: ComboBox | None
    
    def __init__(self, name: str) -> None:
        Node.__init__(self, name, terminals={'In': {'io': 'in', 'multi': True}})
        self.plot = None  # <-- no need for typing here! Type is defined at class level.
```

longqzh:
Hi, @j9ac9k 

I hope everything is going well with you. 
I'm not sure if I'll have time to finish this PR, so feel free to take it over later if you have time :)

PR #2873, opened 2023-11-02
ButtonItem: draw pixmap to logical size

pijyoi:
On hidpi displays, `PlotItem`'s "auto" icon is rendered smaller, making it harder to click.
This is due to `ButtonItem` drawing the pixmap in device pixels.

This PR changes the drawing to use logical sizes so that the icon is the same size across different DPI displays.

Remarks:
The "auto" graphic is actually 32x32 while the logical size used is 14, which means that even on a DPI 2.0 display, we will have a nice non-blurry rendering.


j9ac9k:
Thanks @pijyoi 

This is one of those issues in hindsight makes me think "oh yeah, that button is smaller on my laptop vs. my desktop", but never really realized it even as I was switching between displays.

diff LGTM, merging!

Issue #2830, opened 2023-09-25
Persist histogram levels during live image updating


kmcgrathgenerate:
I have a stream of images coming in at a very high rate. I want to be able to set static levels of the image and have the user be able to adjust the levels using the GUI widget but persist them through image updates.

<img width="332" alt="image" src="https://github.com/pyqtgraph/pyqtgraph/assets/107209886/c100ace3-ff0c-4f4b-8c32-8789264aa16a">

I want them to be able to adjust the levels of the histogram (as seen in image above) and have it persist throughout each image. As far as I can tell, there isn't a way to do that.

 I was hoping there would be something like keeping the "state" as you do with keeping the panning static throughout each image, but could not find one for histogram levels

pijyoi:
I do something like the below. The relevant lines are the ones marked with "AAA" and "BBB".
```python
import pyqtgraph as pg
from pyqtgraph.Qt import QtCore
import numpy as np

rng = np.random.default_rng(0)

app = pg.mkQApp()
glw = pg.GraphicsLayoutWidget()
glw.show()
plot = pg.PlotItem()
img = pg.ImageItem(axisOrder='row-major')
plot.addItem(img)
hlut = pg.HistogramLUTItem()
hlut.setImageItem(img)
img.setImage(rng.rayleigh(size=(100,100)))
hlut.setHistogramRange(0, 10)               # AAA
glw.addItem(plot, row=0, col=0)
glw.addItem(hlut, row=0, col=1)

def update():
    noise = rng.rayleigh(size=(100,100))
    XY = np.sinc(np.linspace(-10, 10, 100))
    s = rng.random(1)
    data = s * 10 * np.abs(XY[:, np.newaxis] * XY)**2 + noise
    img.setImage(data, autoLevels=False)    # BBB

timer = QtCore.QTimer()
timer.timeout.connect(update)
timer.start(100)

app.exec()
```

PR #2876, opened 2023-11-11
pcmi: refactor setData()

### LINKED ISSUE #2875: pg.PColorMeshItem crashes when x, y array shapes are changed (setData(x, y, z...)).
I am creating a plot using the **PColorMeshItem** over time (i have a timer loop in pyside6), if i change the oversampling of my datas, then the x, y and z array shapes are changing and the python code generates an error when setting the data using the setData(x, y, z) function : 

  File "D:\3-Python_Programs\...\Radar_Olivier.py", line 203, in updatePlot
    self.img4.setData(x_vals, y_vals, np.flipud(np.abs(angle_arr.T)))
  File "D:\3-Python_Programs\...\venv\lib\site-packages\pyqtgraph\graphicsItems\PColorMeshItem.py", line 243, in setData
    if np.any(self.x != args[0]) or np.any(self.y != args[1]):
ValueError: operands could not be broadcast together with shapes (129,801) (65,401) 

I found the problem... 
In the "PColorMeshItem.py" file on the line **242** : 
replace:
      if np.any(self.x != args[0]) or np.any(self.y != args[1]): 
with : 
      **if np.any(self.x != args[0].shape[0]) or np.any(self.y != args[1].shape[1]):**

Then the problem is solved.
Thank you !

pijyoi:
`PColorMeshItem::setData()` was a giant function that was also being called by `setLevels()` and `setLookupTable()`.

This PR refactors `setData()` to `_rerender()` and `_drawPicture()` while removing `_updateDisplayWithCurrentState()`

`_rerender()` is for redrawing when only levels or colors have changed.
i.e. data bounds have not changed. 
i.e. after a call to `setLevels()` or `setLookupTable()`.

may fix #2875

It is not clear to me why `prepareGeometryChange()` is being called unconditionally, but care is taken to call `informViewBoundsChanged()` only if bounds have been detected to have changed.
```python
self.prepareGeometryChange()
if boundsChanged:
```

### DIFF
--- a/pyqtgraph/graphicsItems/PColorMeshItem.py
+++ b/pyqtgraph/graphicsItems/PColorMeshItem.py
@@ -91,7 +91,7 @@ class PColorMeshItem(GraphicsObject):
             Sets the minimum and maximum values to be represented by the colormap (min, max). 
             Values outside this range will be clipped to the colors representing min or max.
             ``None`` disables the limits, meaning that the colormap will autoscale 
-            each time ``setData()`` is called - unless ``enableAutoLevels=False``.
+            the next time ``setData()`` is called with new data.
         enableAutoLevels: bool, optional, default True
             Causes the colormap levels to autoscale whenever ``setData()`` is called. 
             It is possible to override this value on a per-change-basis by using the
--- a/pyqtgraph/graphicsItems/PColorMeshItem.py
+++ b/pyqtgraph/graphicsItems/PColorMeshItem.py
@@ -219,32 +219,35 @@ class PColorMeshItem(GraphicsObject):
             The minimum and maximum levels are mapped to the lowest and highest colors 
             in the colormap. The autoLevels parameter is ignored if ``enableAutoLevels is False`` 
         """
+        old_bounds = self._dataBounds
+        self._prepareData(args)
+        boundsChanged = old_bounds != self._dataBounds
+
         autoLevels = kwargs.get('autoLevels', True)
+        self._rerender(autoLevels=autoLevels)
 
-        # Has the view bounds changed
-        shapeChanged = False
-        if self.qpicture is None:
-            shapeChanged = True
-        elif len(args)==1:
-            if args[0].shape[0] != self.x[:,1][-1] or args[0].shape[1] != self.y[0][-1]:
-                shapeChanged = True
-        elif len(args)==3:
-            if np.any(self.x != args[0]) or np.any(self.y != args[1]):
-                shapeChanged = True
+        if boundsChanged:
+            self.prepareGeometryChange()
+            self.informViewBoundsChanged()
 
-        if len(args)==0:
-            # No data was received.
-            if self.z is None:
-                # No data is currently displayed, 
-                # so other settings (like colormap) can not be updated
-                return
-        else:
-            # Got new data. Prepare it for plotting
-            self._prepareData(args)
+        self.update()
 
+    def _rerender(self, *, autoLevels):
+        self.qpicture = None
+        if self.z is not None:
+            if (self.levels is None) or (self.enableautolevels and autoLevels):
+                # Autoscale colormap
+                z_min = self.z.min()
+                z_max = self.z.max()
+                self.setLevels( (z_min, z_max), update=False)
+            self.qpicture = self._drawPicture()
 
-        self.qpicture = QtGui.QPicture()
-        painter = QtGui.QPainter(self.qpicture)
+    def _drawPicture(self) -> QtGui.QPicture:
+        # on entry, the following members are all valid: x, y, z, levels
+        # this function does not alter any state (besides using self.quads)
+
+        picture = QtGui.QPicture()
+        painter = QtGui.QPainter(picture)
         # We set the pen of all polygons once
         if self.edgecolors is None:
             painter.setPen(QtCore.Qt.PenStyle.NoPen)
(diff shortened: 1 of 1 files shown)


j9ac9k:
If memory serves the `prepareGeometryChange` call is left over from a debug effort of when PColorMesh item was first being introduced but there were visible artifacts left over when the mesh changed in some way.  It may indeed not be needed.

pijyoi:
>If memory serves the prepareGeometryChange call is left over from a debug effort of when PColorMesh item was first being introduced but there were visible artifacts left over when the mesh changed in some way. It may indeed not be needed.

I believe both calls are needed, but both should either be called together, or both not called.

pijyoi:
A script to exercise the no-bounds-changed code paths.
```python
import random
import numpy as np

import pyqtgraph as pg

rng = np.random.default_rng(12345)

pg.mkQApp()

glw.show()

xsize = 15
ysize = 30
z = rng.standard_normal(size=(xsize, ysize))

pcmi1 = pg.PColorMeshItem(z)
pcmi2 = pg.PColorMeshItem(z)

view1 = glw.addPlot(title="change values", enableMenu=False)
view2 = glw.addPlot(title="change colormap", enableMenu=False)
view1.addItem(pcmi1)
view2.addItem(pcmi2)

colormaps = pg.colormap.listMaps()

def updateData():
    z = rng.standard_normal(size=(xsize, ysize))
    pcmi1.setData(z)

    cmap = pg.colormap.get(random.choice(colormaps))
    lut_qcolor = cmap.getLookupTable(nPts=256, mode=cmap.QCOLOR)
    pcmi2.setLookupTable(lut_qcolor)

timer = QtCore.QTimer()
timer.timeout.connect(updateData)
timer.start(100)

pg.exec()
```

pijyoi:
It seems rather unsatisfactory to me that there should be an enableAutoLevels state and yet the user is allowed to disable it on a case-by-case basis in a call to setData().

Furthermore, one can only override an enabled state to disabled, but not the other way round.

j9ac9k:
> It seems rather unsatisfactory to me that there should be an enableAutoLevels state and yet the user is allowed to disable it on a case-by-case basis in a call to setData().
> 
> Furthermore, one can only override an enabled state to disabled, but not the other way round.

I agree.  That doesn't sound ideal.

pijyoi:
Hello @SimenZhor, I am doing some refactoring of PCMI in #2879 (merged) and this PR, which would possibly modify the changes introduced in #2463.

What is the use case for having both an overall `enableAutoLevels` state and a per-data-update `autoLevels` setting?
The PCMI example exercises only the overall `enableAutoLevels` state setting.

I would propose to remove the per-data-update `autoLevels` setting.

SimenZhor:
Hi @pijyoi , thanks for taking care of the `PColorMeshItem`!
I don't remember the exact reason behind keeping both `enableAutoLevels` and `autoLevels`. But reading back the comments in my PR it looks like I somewhat reluctantly implemented `autoLevels` to keep the API consistent with `ImageItem` after feedback provided in code review. See this comment for more info: https://github.com/pyqtgraph/pyqtgraph/pull/2463#issuecomment-1268074184

I don't have an issue with changing this behavior.

pijyoi:
If the per-update `autoLevels` was a requested addition, then instead of removing it, I would propose to change its behavior to be tri-state.
If:
1) not set (i.e. equal to None), then defer to the overall `enableAutoLevels` state
2) set (to either True / False), then `autoLevels` over-rides `enableAutoLevels`.

SimenZhor:
It seems to me like we all agree that this part of the API is not intuitive or easy to understand. Since it was not clear to you why this solution was implemented, and since this confusion arrived so quickly after my PR, I'm inclined to take that as a sign that the API needs an overhaul (both for `PColorMeshItem` and `ImageItem`). Maybe @NilsNemitz wants to weigh in here, since he requested that I implement `autoLevels` in the first place?

If a breaking API change is not desired, I think your tri-state alternative sounds reasonable.

NilsNemitz:
I certainly won't oppose a nicely thought out improvement to the API. 

The `autoLevels` parameter exists for compatibility with the `setImage` method of `ImageItem`, which is functionally very similar.

But having to call that at every update is inconvenient, so `enableAutoLevels` serves to set a persistent state. 

If given, `autoLevels` should (for this calll only) override the persitent state. If not given, the persistent state should be used. I would consider it a bug that it doesn't work like that right now - the "tristate" logic looks more correct. I think we simply didn't work that out properly in the previous PR. (And I didn't check later on, my apologies!)

I think we should add the same `enableAutoLevels` logic to ImageItem, too. This sets one additional constraint:
If we make `enableAutoLevels` default to `True`, then we'll have
- a consistent data setting API across `ImageItem` and `PColorMeshItem`
- a convenient persistent setting
- backwards compatibility in `ImageItem`

NilsNemitz:
The naming of `enableAutoLevels` parallels `enableAutoRange` in`Viewbox`, which  I believe has a similar function and functionality for axis scaling. 
https://pyqtgraph.readthedocs.io/en/latest/api_reference/graphicsItems/viewbox.html#pyqtgraph.ViewBox.enableAutoRange

I am not saying it cannot be simplified, but that was the reasoning in the previous discusion.

pijyoi:
>The naming of enableAutoLevels parallels enableAutoRange inViewbox, which I believe has a similar function and functionality for axis scaling

I added the `{enable,disable}AutoLevels` method to `ImageItem`. If you refer to the `setOpts` method, all the setting methods are actually of the form `set<Something>`, e.g. `setAutoDownsample`.

I couldn't come up with the right wording for the documentation of `ImageItem::setImage::autoLevels`. The most unambiguous would be pseudo-code...

To be complete, `enableAutoLevels` would need to be added as a keyword argument to `setOpts`. But then that would require more documentation.

The example from https://github.com/pyqtgraph/pyqtgraph/issues/2830#issuecomment-1746888360 becomes
```python
import pyqtgraph as pg
import numpy as np

app = pg.mkQApp()
glw.show()
plot = pg.PlotItem()
img = pg.ImageItem(axisOrder='row-major')
plot.addItem(img)
hlut = pg.HistogramLUTItem()
hlut.setImageItem(img)
img.setImage(rng.rayleigh(size=(100,100)))
hlut.setHistogramRange(0, 10)               # AAA
glw.addItem(plot, row=0, col=0)
glw.addItem(hlut, row=0, col=1)

img.disableAutoLevels()                     # BBB

def update():
    s = rng.random(1)
    img.setImage(data)

timer = QtCore.QTimer()
timer.timeout.connect(update)
timer.start(100)

app.exec()
```

pijyoi:
changes to `ImageItem` have been moved to #2883

j9ac9k:
I know `enableAutoLevels` and `disableAutoLevels` pre-date this PR, but looking at this PR I'm wondering why we just didn't do it like any other setter/getter attribute with `setAutoLeveling()` but that's outside the scope of this PR 👍🏻 

Thanks @pijyoi this LGTM, merging.

PR #2877, opened 2023-11-11
Use QColor.fromString instead of deprecated QColor.setNamedColor

zariiii9003:
`QColor.setNamedColor` is deprecated since Qt 6.6. Use `QColor.fromString` if available.


pijyoi:
We could use the constructor form instead, which is not deprecated and works across all supported versions, and thus avoid having to differentiate between versions.
```python
>>> QtGui.QColor("steelblue")
```

j9ac9k:
Probably should go w/ the constructor form instead, `QColor::fromString` was only introduced in Qt 6.4, so instead of adding if-checks for `hasattr` using the constructor, using the `QColor(namedColor)` instead would probably e simpler

zariiii9003:
Thanks for your comments, i'll update the PR soon

pijyoi:
Your changes have made the following to no longer work:
```python
In [2]: pg.mkColor('steelblue')
Out[2]: PySide6.QtGui.QColor.fromRgbF(0.274510, 0.509804, 0.705882, 1.000000)
```

Referring to:
https://github.com/pyqtgraph/pyqtgraph/pull/1944#issuecomment-892276991
it seems that this feature was left undocumented and "experimental"

pyqtgraph/functions.py
```diff
@@ -253,45 +253,26 @@ def mkColor(*args):
...
                 except KeyError:
                     raise ValueError('No color named "%s"' % c)
-            have_alpha = len(c) in [5, 9] and c[0] == '#'  # "#RGBA" and "#RRGGBBAA"
-            if not have_alpha:
-                # try parsing SVG named colors, including "#RGB" and "#RRGGBB".
-                # note that QColor.setNamedColor() treats a 9-char hex string as "#AARRGGBB".
-                qcol = QtGui.QColor()
-                qcol.setNamedColor(c)
+            # match hex color codes
+            match = re.match(r"#([0-9a-fA-F]{3,8})", c)
```
NilsNemitz, line 260: My personal impression is that it might be better to try avoiding the overhead of running the regex parser just to confirm the hex nature of the input string. As a relatively low level function, I am sure that there's a use case where `mkColor` gets called a lot.
Maybe we can get away with just checking the lead '#' character and *requiring* that this is followed by a valid hex sequence. Ideally, we'll fail with a readable error message if the conversion fails later?

pyqtgraph/functions.py
```diff
@@ -253,45 +253,26 @@ def mkColor(*args):
...
+                    # convert RRGGBB to RRGGBBAA
+                    c += "ff"
+                r, g, b, a = bytes.fromhex(c)
+            else:
+                # 'c' might be an SVG color keyword
+                qcol = QtGui.QColor(c)
                 if qcol.isValid():
                     return qcol
-                # on failure, fallback to pyqtgraph parsing
-                # this includes the deprecated case of non-#-prefixed hex strings
```
NilsNemitz, line 268: Did we intend to throw a deprecation warning when the code falls through into the path for non #-prefixed hex codes?

Hex codes without leading '#' character have been out of the documentation for a while (two years?) now, but it seems that we've continued to just silently accept them.

pyqtgraph/functions.py
```diff
@@ -253,45 +253,26 @@ def mkColor(*args):
...
             if len(c) == 1:
                 try:
-                    return Colors[c]
+                    return QtGui.QColor(Colors[c])  # return copy
                 except KeyError:
                     raise ValueError('No color named "%s"' % c)
-            have_alpha = len(c) in [5, 9] and c[0] == '#'  # "#RGBA" and "#RRGGBBAA"
-            if not have_alpha:
-                # try parsing SVG named colors, including "#RGB" and "#RRGGBB".
-                # note that QColor.setNamedColor() treats a 9-char hex string as "#AARRGGBB".
```
NilsNemitz, line 262: We could convert pyqtgraph's internal `#RRGGBBAA` hex string format to Qt's `#AARRGGBB` format by simply reshuffling the string, then the code could go through the QColor(string) constructor in *all* cases where the input is a string.

If length is '#' + 4, then the format is RGBA and needs to be expanded manually by e.g. the neat `"".join([x + x for x in c])` construction used below.

After that, or when length is '#' + 8, then reshuffle `c = '#' + c[8:10] + c[1:7]`

Then, `QColor(c)` should handle everything.

j9ac9k:
Hi @zariiii9003 

Thanks for the PR, this LGTM, I'm about to merge.  I wanted to apologize for the delay on the merge.  

Thanks!

PR #2881, opened 2023-11-14
Fix UnderflowError in _try_rescale_float when maxVal == 0.

nup002:
In ImageItem.py, an UnderflowError exception is thrown in _try_rescale_float by numpy.nextafter() if the histogram levels minVal and maxVal are both 0. This PR fixes the problem by instead setting maxVal to the machine precision if maxVal is 0.

Tested and verified to resolve the issue without introducing new ones.

### DIFF
--- a/pyqtgraph/graphicsItems/ImageItem.py
+++ b/pyqtgraph/graphicsItems/ImageItem.py
@@ -640,7 +640,7 @@ class ImageItem(GraphicsObject):
 
         minVal, maxVal = levels
         if minVal == maxVal:
-            maxVal = xp.nextafter(maxVal, 2*maxVal)
+            maxVal = xp.nextafter(maxVal, 2*maxVal) if maxVal != 0 else numpy.finfo(maxVal).eps
         rng = maxVal - minVal
         rng = 1 if rng == 0 else rng
 


pyqtgraph/graphicsItems/ImageItem.py
```diff
@@ -640,7 +640,7 @@ def _try_rescale_float(self, image, levels, lut):
 
         minVal, maxVal = levels
         if minVal == maxVal:
-            maxVal = xp.nextafter(maxVal, 2*maxVal)
+            maxVal = xp.nextafter(maxVal, 2*maxVal) if maxVal != 0 else numpy.finfo(maxVal).eps
```
NilsNemitz, line 643: I have a dumb question:
Is it actually intentional that the existing code moves `maxval` to the next more *negative* number when `maxval` is already < 0? That makes `rng` negative below.

If the intention is to clamp `rng` to a small positive number, then wouldn't
```maxVal = math.nextafter( minVal, math.inf )```
be what we want?

File: pyqtgraph/graphicsItems/ImageItem.py [hunk 2]
```diff
@@ -640,7 +640,7 @@ def _try_rescale_float(self, image, levels, lut):
...
         rng = maxVal - minVal
         rng = 1 if rng == 0 else rng
```
NilsNemitz, line 645: This (pre-existing) line seems redundant if the code above works as intended. If it is indeed still needed, then there's a kind of problem that gets past the check `minVal == maxVal`... Possibly something originating from mixed or non-float datatypes?

Is the real thing we want to avoid ```float(maxval - minval) == 0``` ? Then the most robust path might be to *directly* check that. Something like
```
rng = float(maxVal - minVal)
if rng == 0.:
   maxVal = math.nextafter( minVal, math.inf )
   rng = float(maxVal - minVal)
```

pijyoi:
The code chunk actually comes from `functions::makeARGB()`:
https://github.com/pyqtgraph/pyqtgraph/blob/21391e08a3f93aaf83ab22af642dabfa85e00e25/pyqtgraph/functions.py#L1441-L1462
Back then, the goal was to make the code behave the same as `makeARGB()`.

I would replace it with the following:
```python
minVal, maxVal = levels
# if minVal == maxVal:
#     maxVal = xp.nextafter(maxVal, 2*maxVal)
rng = maxVal - minVal
rng = 1 if rng == 0 else rng
```
The goal is simply to avoid a division by zero shortly after.

NilsNemitz:
Well, that looks like a dramatic reduction in complexity!

And unless I am missing some side effect in the rescale functions, the exact value of the combined scaling factor `scale/rng` should not matter when the output range is zero - or it should at least be very obvious if it does.

j9ac9k:
I belive the issue this PR was meant to address is now addressed in #2944 ?

j9ac9k:
Indeed superseded by #2944

pijyoi:
Note that while #2944 fixes the `ImageItem` "fast path", it is not fixed in the `functions.makeARGB` "slow path"

PR #2883, opened 2023-11-18
ImageItem: implement persistent default autolevels

pijyoi:
Allow user to set whether to perform auto levels by default instead of having to specify it on each data change.
The original behavior is that if unspecified, `autoLevels==True`. i.e. the user needs to specify `autoLevels=False` explicitly on each data change.

The example from https://github.com/pyqtgraph/pyqtgraph/issues/2830#issuecomment-1746888360 becomes

```python
import pyqtgraph as pg
import numpy as np

pg.mkQApp()
glw.show()
plot = pg.PlotItem()
img = pg.ImageItem(axisOrder='row-major', autoLevels=False)
assert img.getLevels() is None
plot.addItem(img)
hlut = pg.HistogramLUTItem()
hlut.setImageItem(img)
# HistogramLUTItem immediately sets levels [0, 1] to the ImageItem 
assert img.getLevels() is not None
hlut.setHistogramRange(0, 10)
glw.addItem(plot, row=0, col=0)
glw.addItem(hlut, row=0, col=1)

# set the levels manually 
img.setLevels([0, 4])

def update():
    s = rng.random(1)
    img.setImage(data)

timer = QtCore.QTimer()
timer.timeout.connect(update)
timer.start(100)

pg.exec()
```

Oh, a plug for https://github.com/pyqtgraph/pyqtgraph/pull/2779: if you are running off master, right-clicking on the colorbar now exposes all the colormaps bundled with pyqtgraph.

### DIFF
--- a/pyqtgraph/graphicsItems/ImageItem.py
+++ b/pyqtgraph/graphicsItems/ImageItem.py
@@ -84,6 +84,7 @@ class ImageItem(GraphicsObject):
         self._defferedLevels = None
         self._imageHasNans = None    # None : not yet known
         self._imageNanLocations = None
+        self._defaultAutoLevels = True
 
         self.axisOrder = getConfigOption('imageAxisOrder')
         self._dataTransform = self._inverseDataTransform = QtGui.QTransform()
@@ -167,6 +168,14 @@ class ImageItem(GraphicsObject):
         if (height := self.height()) is None:
             height = 0.
         return QtCore.QRectF(0., 0., float(width), float(height))
+
+    def setAutoLevels(self, bState: bool):
+        """
+        Controls whether automatic image scaling takes place for this ImageItem,
+        if not otherwise overridden by ``autoLevels`` or ``levels`` keyword
+        arguments in a call to :func:`~pyqtgraph.ImageItem.setImage`.
+        """
+        self._defaultAutoLevels = bState
 
     def setLevels(self, levels: npt.ArrayLike | None, update: bool=True):
         """
@@ -390,6 +399,8 @@ class ImageItem(GraphicsObject):
         --------
         :meth:`setAutoDownsample`
             Accepts the value of ``kwargs['autoDownsample']``.
+        :meth:`setAutoLevels`
+            Accepts the value of ``kwargs['autoLevels']``.
         :meth:`setNanPolicy`
             Accepts the value of ``kwargs['nanPolicy']``.
         :meth:`setBorder`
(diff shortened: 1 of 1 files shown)


pijyoi:
If I were to implement `enableAutoLevels` as a keyword argument for `ImageItem` to follow `PColorMeshItem`'s API to allow setting during object instantiation, the API becomes weird for `ImageItem`.

`ImageItem` has a `setOpts` method that is called in both `__init__` and `setImage`. If `enableAutoLevels` were implemented in `setOpts`, then it would become possible for the user to do the following:
`setImage(..., autoLevels=<something>, enableAutoLevels=<something>)`

pijyoi:
Perhaps `enableAutoLevels` should be named as `autoLevels`. This is consistent with the rest of the keyword arguments.
Then in `setImage`, if `autoLevels` is present as a keyword argument, it will not get passed on to `setOpts`.
This is also intuitive in that `setImage(autoLevels=<something>)` is only supposed to override the persistent value temporarily.

pijyoi:
I have reworked things to be more consistent with the API of `ImageItem` itself.

There is a common edge-case (!) shown in the updated example above:
The user wants the image to auto-scale on the very first data set and thereafter use the same levels, based on the premise that images from the same data source have similar intensity levels.

The `ImageItem` code detects that no levels have yet been set or computed, and forces an autoLevels, despite any default autoLevels setting to the contrary.

This logic fails when used with a `HistogramLUTItem`, which will automatically set the levels to [0, 1] upon initialization. So the user needs to either reset the levels to `None` or set levels manually.

pijyoi:
So on master and this PR, the following throws an exception of `Exception: levels argument is required for float input types`, because no levels have been set.
```python
import pyqtgraph as pg
import numpy as np

pg.mkQApp()
plt = pg.PlotWidget()
plt.show()
img = pg.ImageItem()
data = rng.rayleigh(size=(100,100))
img.setImage(data, autoLevels=False)
plt.addItem(img)
pg.exec()
```

On master and on this PR, the following doesn't throw an Exception, but for different reasons.

On master, `autoLevels` is an unrecognized keyword and is silently ignored. Since `setImage` doesn't specify `autoLevels=False`, it defaults to `True` and auto scaling takes place (for this and subsequent `setImage`).

On this PR, `autoLevels` is recognized and stored. On the first `setImage`, since no levels have been set, auto scaling takes place for this very first `setImage` only.

```python
import pyqtgraph as pg
import numpy as np

pg.mkQApp()
plt = pg.PlotWidget()
plt.show()
img = pg.ImageItem(autoLevels=False)
data = rng.rayleigh(size=(100,100))
img.setImage(data)
plt.addItem(img)
pg.exec()
```

pijyoi:
Unfortunately, forcing `autoLevels` if `self.levels is None` looks like to be the wrong thing to do for `ImageItem`, because in `ImageItem::render()`, `self.levels is None` is absolutely a permitted value and signifies to use the min and max values of the underlying (integer) data type.

So the safe thing to do is to remove the force `autoLevels` if `self.levels is None` logic.

j9ac9k:
Thanks @pijyoi for this PR, I'll try and not take 2 years to merge your PRs in the future 😆

PR #2892, opened 2023-12-02
adding support for pyqtgraph editable install

spchamp:
This change adds build-system metadata to pyproject.toml and ensures the dir for setup.py is in sys.path.

The sys.path modification is to ensure the project ./tools/ modules can be imported in setup.py during editable install.

With this tweak to the project metadata, it should be possible to install pyqtgraph with an editable install, using setuptools

```
mkdir newproject; cd newproject
python3 -m venv --prompt newproject env
source env/bin/activate
pip install -e ../pyqtgraph/ --config-settings editable_mode=compat
```

Previous to this change, while it was possible to run `pip install -e ../pyqtgraph` but that might not have resulted in the modules actually being installed and available under the user environment.

### Files that need updates
    
- [n/a] `README.md` 
- [x] `setup.py`
- [n/a ] `tox.ini`
- [?] `.github/workflows/main.yml` and associated `requirements.txt` and conda `environemt.yml` files
- [x] `pyproject.toml`
- [n/a] `binder/requirements.txt`

I don't think this would affect the workflow builds. Hoping it would test out though lol

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,7 +46,7 @@ jobs:
             qt-version: "PySide2~=5.15.0"
           - python-version: "3.10"
             qt-lib: "pyqt"
-            qt-version: "PyQt6~=6.2.0 PyQt6-Qt6~=6.2.0"
+            qt-version: "PyQt6==6.2.0 PyQt6-Qt6==6.2.0"
           - python-version: "3.10"
             qt-lib: "pyside"
             qt-version: "PySide6~=6.2.0"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,7 +52,7 @@ jobs:
             qt-version: "PySide6~=6.2.0"
           - python-version: "3.11"
             qt-lib: "pyqt"
-            qt-version: "PyQt6"
+            qt-version: "PyQt6==6.2.0 PyQt6-Qt6==6.2.0"
           - python-version: "3.11"
             qt-lib: "pyside"
             qt-version: "PySide6-Essentials"
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,4 +1,5 @@
 PyQt6==6.6.0
+PyQt6-Qt6==6.6.0
 sphinx==7.2.6
 pydata-sphinx-theme==0.14.3
 sphinx-design==0.5.0
--- a/tox.ini
+++ b/tox.ini
@@ -27,8 +27,8 @@ deps=
     {[base]deps}
     pyside2_515: pyside2
     pyqt5_515: pyqt5
-    pyqt6_62: pyqt6~=6.2.0
-    pyqt6_62: PyQt6-Qt6~=6.2.0
+    pyqt6_62: pyqt6==6.2.0
+    pyqt6_62: PyQt6-Qt6==6.2.0
     pyside6_62: pyside6~=6.2.0
     pyqt6: pyqt6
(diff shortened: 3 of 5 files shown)


spchamp:
Did this change break the rtd build? 

Noticed the following in the [build log](https://readthedocs.org/projects/pyqtgraph/builds/22727687/) for Read the Docs:
> `PyQtGraph requires one of PyQt5, PyQt6, PySide2 or PySide6; none of these packages could be imported."`

Of course PyQt6 is listed in doc/requirements.txt and it was  installed under that step:
> python -m pip install --exists-action=w --no-cache-dir -r doc/requirements.txt
> [...]
> Successfully installed PyQt6-6.6.0 [...]

Maybe this change produced some kind of a side effect for how the rtd build resolves the imports for the docs build?

I'll try to see if I can figure this out with a glance at `doc/source/conf.py`

spchamp:
Huh. If this change is breaking the read the docs build, my second guess is that it's because of `sys.path` lol :shrug:

j9ac9k:
There is a new pyqt6 build that having issues, I wouldn’t worry about it

spchamp:
added a patch for PyQt6 version parity in doc/requirements.txt
```
PyQt6==6.6.0
PyQt6-Qt6==6.6.0
```

fwiw the tests all run successfully under Python 3.12 and the docs build

spchamp:
fwiw I've folded a couple more patches into this pull request, to ensure version alignment for the pyqt6 parts in the CI builds

.github/workflows/main.yml
```diff
@@ -46,13 +46,13 @@ jobs:
             qt-version: "PySide2~=5.15.0"
           - python-version: "3.10"
             qt-lib: "pyqt"
-            qt-version: "PyQt6~=6.2.0 PyQt6-Qt6~=6.2.0"
+            qt-version: "PyQt6==6.2.0 PyQt6-Qt6==6.2.0"
```
j9ac9k, line 49: why this change, I'm fairly certain that `~=` is what I want in this case, I want to get the latest patch version of PyQt6 6.2

doc/requirements.txt
```diff
@@ -1,4 +1,5 @@
 PyQt6==6.6.0
+PyQt6-Qt6==6.6.0
```
j9ac9k, line 2: Got fixed in another PR already, go ahead and remove this 👍🏻

tox.ini
```diff
@@ -27,8 +27,8 @@ deps=
     {[base]deps}
     pyside2_515: pyside2
     pyqt5_515: pyqt5
-    pyqt6_62: pyqt6~=6.2.0
-    pyqt6_62: PyQt6-Qt6~=6.2.0
+    pyqt6_62: pyqt6==6.2.0
+    pyqt6_62: PyQt6-Qt6==6.2.0
```
j9ac9k, line 31: Again, why is there the `==` and not `~=` here?

j9ac9k:
hi @spchamp 

Thanks for the PR, I didn't realize that some modules might not be captured via editable installs in the current form.  I have no problem merging the changes to pyproject.toml and setup.py, but I think the other changes for `~=` to `==` are unrelated and undo the intentional behavior of installing the latest patch version of 6.2.

tox.ini
spchamp: This ensures that the same version is drawn in for both

j9ac9k:
Was superseded by #2935 I cherry picked the first two commits here, thanks for the PR @spchamp

PR #2894, opened 2023-12-02
various fixes for NumPy 2.0

pijyoi:
This PR fixes the test suite to pass with NumPy 2.0, which may be releasing soon (https://github.com/numpy/numpy/issues/24300).

### DIFF
--- a/pyqtgraph/examples/NonUniformImage.py
+++ b/pyqtgraph/examples/NonUniformImage.py
@@ -31,7 +31,7 @@ V = np.ones_like(TAU) * v
 P_loss = kfric * W + kfric3 * W ** 3 + (res * (TAU / psi) ** 2) + k_v * (V - v_ref)
 
 P_mech = TAU * W
-P_loss[P_mech > 1.5e5] = np.NaN
+P_loss[P_mech > 1.5e5] = np.nan
 
 app = pg.mkQApp("NonUniform Image Example")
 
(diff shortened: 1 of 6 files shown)


pijyoi:
Just a short script to verify that pyqtgraph's `clip_array` workaround is no longer needed since `numpy` >= 1.25.0
```python
import time
import numpy as np
from pyqtgraph.functions import clip_array

print(np.__version__)

shape = 10000
data = rng.random(size=shape, dtype=np.float32)
niters = 10000

for alpha in [0.1, 0.25]:
    for name, func in [('np', np.clip), ('pg', clip_array)]:
        vmin, vmax = alpha, 1 - alpha
        output = np.zeros_like(data)
        t0 = time.perf_counter()
        for _ in range(niters):
            res = func(data, vmin, vmax, out=output)
        t1 = time.perf_counter()
        print(name, data.dtype, alpha, t1 - t0)
```

```
1.24.4
np float32 0.1 0.3700788999994984
pg float32 0.1 0.204399000009289
np float32 0.25 0.36337140000250656
pg float32 0.25 0.20933669999067206

1.25.0
np float32 0.1 0.21235800000431482
pg float32 0.1 0.22272969999175984
np float32 0.25 0.213919599991641
pg float32 0.25 0.23375790000136476
```

j9ac9k:
I'm surprised more changes weren't needed for NumPy 2.0, guess goes to show you how mature their API is.  Thanks for the PR @pijyoi

PR #2903, opened 2023-12-16
fix: paint method widget argument can be None

pijyoi:
This PR fixes a bug introduced in #2863.
In the `paint` method, the DPR is obtained by calling `widget.devicePixelRatioF()`. However, the `widget` argument can be `None`. https://doc.qt.io/qt-6/qgraphicsitem.html#paint

This bug manifests itself at least in the Jupyter Notebook examples using `ScatterPlotItem`.


j9ac9k:
walrus operator has entered the codebase! 

Thanks @pijyoi sorry it's taken me so long to get to this.  This LGTM, I appreciate your continued work and effort on this library!

PR #2907, opened 2023-12-23
class Handle Fix allowing dict to be passed for handlePen and removing hard reset for width and cosmetic attributes

### LINKED ISSUE #1960: ROI overrides pen parameters explicitly
I would expect ROIs that take a `pen` parameter to preserve aspects of that pen, but `Handle` instances explicitly reset their `width` and `cosmetic` properties during init. Is there a reason these aren't simply inferred from the starting pen? Otherwise, I would request that these lines be removed.

https://github.com/pyqtgraph/pyqtgraph/blob/0d8eba36a6a19ab6d8453f66a7dcf8d7e6f3c6a3/pyqtgraph/graphicsItems/ROI.py#L1361-L1362

### LINKED ISSUE #2765: Scale Handle line width not changing when passed directly as keyword of mkPen or as dict thru handlePen within ROI class
### my Story

 Passing ROI class and adding Scale handles, due to the nature of the background colors and data plotted below the active ROI, the default scale handle is not really visible so tried to increase width of line and change color, by setting `self.handlePen = {'color': (150,75,255), 'width': 5}` in the custom class calling the pyqtgraph.ROI I am able to change the default color used by mkPen when creating the Handle, but the width of the line is not changing...
I noticed the example used in the mkPen docstring is throwing an error (NameError: name 'width' is not defined) so changed only `width: 5` to `'width': 5`

I tested out as well passing directly a Handle as item object into assScaleHandle, but no success], color works, but width nope again, I even tested a dash and it worked...

It seems like mkPen is not reading properly the width keyword argument I am passing, or it is not really passing it properly to QPen??? I'm kind of lost here, unless something somewhere else is replacing the width being used by the Handle.

### Class using pg.ROI, with handlePen set as dictionary Try#1: 
```python
import pyqtgraph as pg
import numpy as np

class EventRoi(pg.ROI):
    def __init__(self, pos, size, custom_format, mainApp=None, *args, **kwargs):
        super().__init__(pos, size, *args, **kwargs)

        self.mainApp = mainApp
        self.handlePen = {'color': (255, 255, 0),'width': 4}
```
### Function using myROI with Handle class passed as item when adding scaleHandle try#2:
```python
def add_roi(self, x0, x1, y0, y1, format = {}, is_restored_event=False, ignoreBounds=True):
        roi = EventRoi(
            [x0, y0],
            [x1-x0, y1-y0],
            custom_format=format,
            mainApp=self)

        roi.addScaleHandle([1, 0.5],
                           [0, 0.5], 
                           item=pg.graphicsItems.ROI.Handle(roi.handleSize,
                                                            typ='s',
                                                            pen=pg.functions.mkPen(color=(255, 255, 0, 255), width=4, dash= [1,1]),
                                                            hoverPen=roi.handleHoverPen,
                                                            parent=roi))
```
### Expected behavior
Handles with wider lines and set color

### Real behavior
Handles have set color, and in the second try even the correct dashing, but line is still default width, this is the same for both versions mentioned above.
No errors occurred during runtime.

### Tested environment(s)

 * PyQtGraph version: 0.13.3
 * Qt Python binding: 'PyQt5 5.15.7 Qt 5.15.2'
 * Python version: 3.10.8
 * NumPy version: 1.24..3
 * Operating system: Windows
 * Installation method: conda install

pdmkdz:
Adding ability to modify `handlePen` for `class ROI` removing hard coded defaults for `mkPen` `width=0` and `cosmetic=True` in ROI.py `class Handle`. Keeping `width=0` and `cosmetic=True` as default for assignation without customization.

Fixes #1960
Fixes #2765

This will allow to pass a dictionary for pen customization when needed due to visibility.

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -1336,7 +1336,7 @@ class Handle(UIGraphicsItem):
     sigClicked = QtCore.Signal(object, object)   # self, event
     sigRemoveRequested = QtCore.Signal(object)   # self
     
-    def __init__(self, radius, typ=None, pen={'color': (200, 200, 220),'width':0, 'cosmetic':True},
+    def __init__(self, radius, typ=None, pen={'color': (200, 200, 220)},
                  hoverPen=(255, 255, 0), parent=None, deletable=False):
         self.rois = []
         self.radius = radius
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -1336,7 +1336,7 @@ class Handle(UIGraphicsItem):
     sigClicked = QtCore.Signal(object, object)   # self, event
     sigRemoveRequested = QtCore.Signal(object)   # self
     
-    def __init__(self, radius, typ=None, pen={'color': (200, 200, 220)},
+    def __init__(self, radius, typ=None, pen=(200, 200, 220),
                  hoverPen=(255, 255, 0), parent=None, deletable=False):
         self.rois = []
         self.radius = radius
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -1336,7 +1336,7 @@ class Handle(UIGraphicsItem):
     sigClicked = QtCore.Signal(object, object)   # self, event
     sigRemoveRequested = QtCore.Signal(object)   # self
     
-    def __init__(self, radius, typ=None, pen=(200, 200, 220),
+    def __init__(self, radius, typ=None, pen={'color': (200, 200, 220),'width':0, 'cosmetic':True},
                  hoverPen=(255, 255, 0), parent=None, deletable=False):
         self.rois = []
         self.radius = radius
@@ -1344,8 +1344,6 @@ class Handle(UIGraphicsItem):
         self.pen = fn.mkPen(pen)
         self.hoverPen = fn.mkPen(hoverPen)
         self.currentPen = self.pen
-        self.pen.setWidth(0)
-        self.pen.setCosmetic(True)
         self.isMoving = False
         self.sides, self.startAng = self.types[typ]
         self.buildPath()


pdmkdz:
New PR removed all the unnecessary changes to other files.

pijyoi:
The cosmetic=True can be dropped because pg.mkPen defaults to making cosmetic pens.

From what I understand, pen widths of 0 are a holdover from Qt4 days. It is supposed to be equivalent to width=1 and cosmetic=True; in which case, even the width=0 can be dropped because pg.mkPen defaults to width=1.

That said, leaving width=0 is the safe option.

Note: pen widths of zero forces code that need to know about pen widths to handle the special case of zero.

pijyoi:
>Note: pen widths of zero forces code that need to know about pen widths to handle the special case of zero.

Places that convert 0 pen width to 1:
1) https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ArrowItem.py#L143
2) https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ScatterPlotItem.py#L97
3) https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/BarGraphItem.py#L281
4) https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/PColorMeshItem.py#L367

Places that failed to convert 0 pen width to 1:
1) https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/PlotCurveItem.py#L289-L292

pdmkdz:
Considered there is a single failure maybe removing the 0 and setting to 1 is best?
While I am not prone to fight the cosmetic = True drop, I like the idea to see the width set to a default together with the color, but maybe 1 is better than 0 after seeing that single point of failure.
Let me know if I need to update this.

pdmkdz:
New commit removing those unnecessary default attributes values, I tested on my own app and worked perfectly.

j9ac9k:
Hi @pdmkdz, thanks for the PR, sorry this has sat idle so long.

The main issue I have with this PR is that we are specifying a mutable object as a default parameter in a function declaration.  This is one of the [main python gotchas](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) which can cause really confusing behavior.

Throughout the library it's common to use an immutable object (boolean, None, etc) and have an if-conditional that assigns the appropriate attribute to the dictionary.  I think passing in a `dict` to that argument is fine, it just can't have a `dict` as a default value.

pdmkdz:
I removed the dict though in the last iteration.
Didn't the commit update?

I just removed the hard set to width=0, that is pretty bad because whatever
I pass in mkPen get overridden.

*Michele Del Zoppo*
M.Sc. Structural Geology & Geodynamics
Cell: +1 (336) 456 8213
Houston, TX
* [image: LinkedIn Account]
<https://www.linkedin.com/in/michele-del-zoppo-a8791761> *

*“An investment in knowledge pays the best interest.” **— Benjamin Franklin*

-----------------------------------------------------------------------------------------
This message may contain confidential and privileged information. If it has
been sent to you in error, please reply to advise the sender of the error
and then immediately delete it. If you are not the intended recipient, do
not read, copy, disclose or otherwise use this message. The sender
disclaims any liability for such unauthorized use.


Il giorno mer 14 feb 2024 alle ore 05:40 Ogi Moore ***@***.***>
ha scritto:

pijyoi:
No, there's still a {} dictionary.
```python
    def __init__(self, radius, typ=None, pen={'color': (200, 200, 220)},
```

j9ac9k:
Hi @pdmkdz 

Thanks for the PR, I hope you don't mind me adding a commit to your branch, I wanted to get your PR into the next release, so I thought I would just make the small change myself.  Thanks for your contribution to the library!

pdmkdz:
Happy to help, my very first around here 😊

PR #2923, opened 2024-02-01
disconnect only slots that got connected

pijyoi:
PySide 6.7 will stop raising an exception when attempting to disconnect a slot that wasn't connected in the first place.
It will instead print out a warning.
See:
https://code.qt.io/cgit/pyside/pyside-setup.git/commit/?id=d7aa15abe25bd71ea19180743ce9b41e0b788520

This PR fixes:
1) the many warnings printed due to `GraphicsItem`.
2) https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ViewBox/ViewBox.py#L1780-L1795
     - this warning gets emitted on program exit

Remaining code that still cause warnings:
1) https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/parametertree/interactive.py#L115-L118
     - this warning gets emitted in the test suite

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -3206,12 +3206,18 @@ def disconnect(signal, slot):
     """
     while True:
         try:
-            signal.disconnect(slot)
-            return True
+            success = signal.disconnect(slot)
+            if success is None:     # PyQt
+                success = True
         except (TypeError, RuntimeError):
-            slot = reload.getPreviousVersion(slot)
-            if slot is None:
-                return False
+            success = False
+
+        if success:
+            return True
+
+        slot = reload.getPreviousVersion(slot)
+        if slot is None:
+            return False
 
 
 class SignalBlock(object):
--- a/pyqtgraph/graphicsItems/GraphicsItem.py
+++ b/pyqtgraph/graphicsItems/GraphicsItem.py
@@ -493,10 +493,9 @@ class GraphicsItem(object):
 
         ## disconnect from previous view
         if oldView is not None:
-            for signal, slot in [('sigRangeChanged', self.viewRangeChanged),
-                                 ('sigDeviceRangeChanged', self.viewRangeChanged), 
-                                 ('sigTransformChanged', self.viewTransformChanged), 
-                                 ('sigDeviceTransformChanged', self.viewTransformChanged)]:
+            Device = 'Device' if hasattr(oldView, 'sigDeviceRangeChanged') else ''
+            for signal, slot in [(f'sig{Device}RangeChanged', self.viewRangeChanged),
+                                 (f'sig{Device}TransformChanged', self.viewTransformChanged)]:
                 try:
                     getattr(oldView, signal).disconnect(slot)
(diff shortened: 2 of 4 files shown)


pijyoi:
I don't understand the condition in the following code:
`self.destroyed.connect(lambda: ViewBox.forgetView(sid, name) if (ViewBox is not None and 'sid' in locals() and 'name' in locals()) else None)`
from https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ViewBox/ViewBox.py#L284-L288
and explicitly added in https://github.com/pyqtgraph/pyqtgraph/commit/e5f383fbb5e3953d44b9767d3a73b32d2a32e480

Based on my testing script below, the expression `'sid' in locals() and 'name' in locals()` is always `True`, since they are referenced in the expression `ViewBox.forgetView(sid, name)`
```python
def func_indirect(x):
    return x

def func1(x):
    # "'x' in locals" is True 
    return lambda : func_indirect(x) if 'x' in locals() else None

def func2(x):
    y = 2 * x
    # "'y' in locals" is False
    return lambda : func_indirect(x) if 'y' in locals() else None

print(func1(1)())   # 1
print(func2(2)())   # None
```

samtygier:
I got interested in what was happening with `locals()` in the lambda. I agree, looks like it must always be true. I wondered if there had been a change in python since that was written, but see the same output for your test in python 2.7 and 3.2.

pijyoi:
While trying to fix `interactive.py` to not trigger warnings on PySide 6.7, I am not sure if I have stumbled onto a bug. The 3rd commit in this PR does a check for connections before attempting to disconnect; and this does manage to prevent some of the warnings.

However, despite this check, there are still 2 failures to disconnect of `sigValueChanged`.
These failures are actually present in all bindings.
i.e. previously, these real failures were just being ignored

j9ac9k:
Thanks @pijyoi Always nice when we can address an issue before the upstream library makes a change.  Appreciate your effort to look ahead and mitigate issues per-emptively.

campagnola:
> I don't understand the condition in the following code: `self.destroyed.connect(lambda: ViewBox.forgetView(sid, name) if (ViewBox is not None and 'sid' in locals() and 'name' in locals()) else None)` from https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/ViewBox/ViewBox.py#L284-L288 and explicitly added in [e5f383f](https://github.com/pyqtgraph/pyqtgraph/commit/e5f383fbb5e3953d44b9767d3a73b32d2a32e480)
> 
> Based on my testing script below, the expression `'sid' in locals() and 'name' in locals()` is always `True`, since they are referenced in the expression `ViewBox.forgetView(sid, name)`

Sorry for the late reply! The reson for this code is that the callback can be invoked during python teardown (in which case some of these locals may already have been deleted). I think this just avoids error messages that would appear on exit, but it's also possible that it prevented segmentation faults at exit, which were common at the time this was written.

PR #2930, opened 2024-02-06
Center exportDialog using screen geometry

MorbidCuriosity84:
The fix for Issue 2510 https://github.com/pyqtgraph/pyqtgraph/pull/2510 can result in the y coordinate being 0.  See attached image where application is not full screen and exportDialog appears at the top of the screen

This fixes that by using the screen geometry to find the center of  the screen and then centers the exportDialog to the center of the active screen.

Fixes #2510

image

### DIFF
--- a/pyqtgraph/GraphicsScene/exportDialog.py
+++ b/pyqtgraph/GraphicsScene/exportDialog.py
@@ -51,10 +51,11 @@ class ExportDialog(QtWidgets.QWidget):
         self.selectBox.setVisible(True)
         if not self.shown:
             self.shown = True
-            vcenter = self.scene.getViewWidget().geometry().center()
-            x = max(0, int(vcenter.x() - self.width() / 2))
-            y = max(0, int(vcenter.y() - self.height() / 2))
-            self.move(x, y)
+            screen = QtWidgets.QApplication.desktop().screenNumber(QtWidgets.QApplication.desktop().cursor().pos())
+            centre = QtWidgets.QDesktopWidget().availableGeometry(screen).center()
+            frame = self.frameGeometry()
+            frame.moveCenter(centre)
+            self.move(frame.topLeft())
         
     def updateItemList(self, select=None):
         self.ui.itemTree.clear()


j9ac9k:
Hi @MorbidCuriosity84 

Thanks for the PR, I'm not convinced we want that dialog in the center of the screen as a default, but this PR does address an existing issue and thus is an improvement, so I'm going to merge.   I need to pay more attention to what other applications do in this regard.

MorbidCuriosity84:
Hi @j9ac9k 

The other options I can think of:

- Center it on the main window 
- Center it on the parent if it exists
- Center on active window

All have their issues:

- There might be more than one QMainWindow 
- The parent/active window might be at the bottom/top of the screen and cause the dialog to overflow the screen

I would lean towards using the active window (can be easily accessed from QApplication) and checking that there is sufficient room for the dialog to not overflow and if not either moving so that it doesn't overflow or defaulting back to the center of the screen.  A little complex perhaps, but should cover most use cases and still keep the dialog close to the graph.

If you've got any other ideas, let me know and I can try to implement them.

j9ac9k:
I may have merged pre-emptively.  Using PyQt6 6.6.1, with the export dialog...

```
  File "/Users/ogi/Developer/pyqtgraph/pyqtgraph/GraphicsScene/GraphicsScene.py", line 543, in showExportDialog
    self.exportDialog.show(self.contextMenuItem)
  File "/Users/ogi/Developer/pyqtgraph/pyqtgraph/GraphicsScene/exportDialog.py", line 54, in show
    screen = QtWidgets.QApplication.desktop().screenNumber(QtWidgets.QApplication.desktop().cursor().pos())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: type object 'QApplication' has no attribute 'desktop'
```

j9ac9k:
Ahh, QDesktopWidget is deprecated, and was removed in Qt6

https://doc.qt.io/qt-6/widgets-changes-qt6.html#qdesktopwidget-and-qapplication-desktop

j9ac9k:
Able to work around it with:

```python
            # screen = QtWidgets.QApplication.desktop().screenNumber(QtWidgets.QApplication.desktop().cursor().pos())
            # centre = QtWidgets.QDesktopWidget().availableGeometry(self.screen()).center()
            center = self.screen().availableGeometry().center()
            frame = self.frameGeometry()
            frame.moveCenter(center)
```

PR #2491, opened 2022-10-14
Add link of Antenna Array Analysis tool

rookiepeng:
Add the link of Antenna Array Analysis tool that I am maintaining. I am using pyqtgraph in this project.

PR #2934, opened 2024-02-11
Fix 'import pyqtgraph.canvas' crash

### LINKED ISSUE #2838: import pyqtgraph.canvas fails.
### Short description
Importing pyqtgraph.canvas causes the program to abort and exit with an error code..

### Code to reproduce

```python
print("Mark 0", flush=True)
import time
print("Mark 1", flush=True)
import pyqtgraph.canvas
print("Mark 2", flush=True)
time.sleep(10)
print("Mark 3", flush=True)

```

### Expected behavior
1. The program should print out all the markers four ,0 to 3, with a 10 seconds delay between marker 2 and 3.
2. The program should exit with status code 0 (ok)

### Real behavior
1. The program prints only markers 0 and 1.
2. The program exist with status code 127 (error)

### An error occurred?
No error messages were noticed. The program just exits when it tries to import pyqtgraph.canvas.

### Tested environment(s)

 * PyQtGraph version:  0.13.3
 * Qt Python binding:   PyQt6 6.5.2 Qt 6.5.2
 * Python version:  3.12.0
 * NumPy version:  1.26.0
 * Operating system:  Windows 10, Ryzen
 * Installation method:  pip

### Additional context

The issue is discussed here in the pyinstaller repository https://github.com/pyinstaller/pyinstaller/issues/7991#issuecomment-1752032919

Swapping the two lines of CanvasManager as described the link above seems to fix the problem.

nicoddemus:
Just importing pyqtgraph.canvas causes the interpreter to crash.

Saving the `SINGLETON` after the call to `QObject.__init__` solves the issue.

Fix #2838

### DIFF
--- a/pyqtgraph/canvas/CanvasManager.py
+++ b/pyqtgraph/canvas/CanvasManager.py
@@ -13,8 +13,9 @@ class CanvasManager(QtCore.QObject):
     def __init__(self):
         if CanvasManager.SINGLETON is not None:
             raise Exception("Can only create one canvas manager.")
-        CanvasManager.SINGLETON = self
+        # It is important to save SINGLETON *after* the call to QObject.__init__, see #2838.
         QtCore.QObject.__init__(self)
+        CanvasManager.SINGLETON = self
         self.canvases = weakref.WeakValueDictionary()
 
     @classmethod
--- a/tests/test_canvas_manager_crash.py
+++ b/tests/test_canvas_manager_crash.py
@@ -1,0 +1,8 @@
+import subprocess
+import sys
+
+
+def test_canvas_manager_singleton_crash() -> None:
+    """Regression test for #2838."""
+    proc = subprocess.run([sys.executable, "-c", "import pyqtgraph.canvas"])
+    assert proc.returncode == 0


j9ac9k:
Thanks @nicoddemus 

I'm curious, so you use the canvas module?

Jamesong7822:
Hi @j9ac9k, will this be in the next upcoming release? if so, when is the expected release date? Thank you very much

j9ac9k:
> Hi @j9ac9k, will this be in the next upcoming release? if so, when is the expected release date? Thank you very much

😅  I'm awful at predicting when I'm going to do a release.  I recently moved across the country and have been dealing with everyone in my household getting sick... there are some PRs I would want to merge before the next release, and ideally I would fix a regression I introduced in the SVG exporter... anyway I have no idea when I'm going to do it.  No doubt that my other responsibilities have pulled me away from this project... hoping to resume my work before much longer.

Jamesong7822:
in that case, for anyone facing this issue, consider installing pyqtgraph from source, instead of via pip. It helped for my project using PyQt6 and pyqtgraph when I was freezing it using PyInstaller.

j9ac9k:
I generally try and cut releases shortly after breakages are addressed, i didn't realize this issue was impacting so many pyinstaller users.

nicoddemus:
@j9ac9k anything I can help to get this into a new release? Seems minor but this prevents users from using pyinstaller.

I would like to do some testing with `pixi`, which uses conda packages and does not support a development install from git (which is how I would workaround this). 😁

j9ac9k:
> @j9ac9k anything I can help to get this into a new release? Seems minor but this prevents users from using pyinstaller.
> 
> I would like to do some testing with `pixi`, which uses conda packages and does not support a development install from git (which is how I would workaround this). 😁

I had been holding off on a fix until I can resolve #2491.  I should point out that the issue is poorly written (I'm the author, I need to spend some more time rewriting and be more specific with the intricacies 🙃 ), but given the series of fixes that have been submitted since the last release (including the pyinstaller issue addressed here), I have come to the conclusion that I should undo the "bug-fix" I made (which included a regression described in #2941) and work on a better way to resolve  #2661 

Sorry @meganbkratz I'm going to have to roll back the bug fix for your issue for a while 

I hope to cut a release in the next 24 hours.  Thanks for following up @nicoddemus  ...if you have some time/interest, I would love your assistance with re-fixing #2661 after I cut this next release.

nicoddemus:
Awesome, thanks a lot @j9ac9k, appreciate it!

> if you have some time/interest, I would love your assistance with re-fixing https://github.com/pyqtgraph/pyqtgraph/issues/2661 after I cut this next release.

To be honest, I know almost nothing about `pyqtgraph`'s codebase, so I'm not sure how much help I could provide. My initial offer to get the release out was along the lines of "release-wise" work, like tidying up changelog, helping reviewing some outstanding PRs, that kind of thing.

meganbkratz:
@j9ac9k No worries about rolling back the svg exporter fix. I have a workaround (subtracting out a constant to make the axis values smaller), so it's not urgent. Thank you for your work on it!!!

j9ac9k:
> To be honest, I know almost nothing about `pyqtgraph`'s codebase, so I'm not sure how much help I could provide. My initial offer to get the release out was along the lines of "release-wise" work, like tidying up changelog, helping reviewing some outstanding PRs, that kind of thing.

Fair enough, issue referenced isn't something requiring lots of tinkering in the pyqgraph codebase or knowledge of its working.  My PR did one attempt to workaround for Qt's implementation of painting QSvgGenerator object

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/exporters/SVGExporter.py#L268-L273

The numbers to the output XML where the precision is lots (so exporting a line that has points with a large numerical value in x (or y), but the values are similar/close together, you'll lost the precision.  My attempt to work around this is to move the line/curve to be between (-1, 1) and then write the SVG out. 

I suspect there is some more fanciness we can do with applying a transform to the graphics items, and then setting the `QPainter::setTransform` , and then applying the inverse after the painter writes to the QSvgGenerator,... anyway just need to spend more time with it.

@meganbkratz Thanks for chiming in and for your understanding 👍🏻

PR #2938, opened 2024-02-15
Use non-deprecated method to access screen info

j9ac9k:
Fixup for #2930 which used a deprecated API that was removed in Qt6.


j9ac9k:
ugh...why isn't CI running?

PR #2944, opened 2024-02-25
Refactor ImageItem QImage creation routines

pijyoi:
The fast-path `QImage` creation methods residing within `ImageItem` can be factored out to be free-standing functions. 
This has the following benefits:
1) Makes the routine stateless, which ensures that it has no side-effects from other methods.
2) Allows the functionality to be used outside of `ImageItem`

Some behavior changes:
1) replaces the use of `nextafter` with the snippet in https://github.com/pyqtgraph/pyqtgraph/pull/2881#issuecomment-1810241405
2) doesn't cache the effective lut. 
    - Caching the effective lut makes the code stateful.
    - The effective lut computation is relatively small. While the overhead of recomputing the effective lut may be significant for *very small* images, displaying such small images should be fast enough already.
      - The largest computation that this caching would have saved is for the case "uint16 image *and* lut with more than 256 entries". This results in an effective lut size of 128KB. However, since luts are usually created with 256 entries, this case would be seldom encountered.
3) the test of ImageFormat no longer needs to instantiate an `ImageItem` object for each test. This could be problematic if the CI was flakey to begin with.

### DIFF
--- a/pyqtgraph/functions_qimage.py
+++ b/pyqtgraph/functions_qimage.py
@@ -1,0 +1,295 @@
+import numpy
+
+from .Qt import QtGui
+from . import functions
+from .util.cupy_helper import getCupy
+from .util.numba_helper import getNumbaFunctions
+
+
+def _apply_lut_for_uint16_mono(xp, image, lut):
+    # Note: compared to makeARGB(), we have already clipped the data to range
+    augmented_alpha = False
+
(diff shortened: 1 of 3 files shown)


j9ac9k:
Thanks @pijyoi for this PR.  ImageItem was getting complicated and there was some good stuff there so I'm a very big 👍🏻 for moving that functionality into free-standing functions.

PR #2947, opened 2024-03-02
ViewBox: Only update auto-range before painting.

agausmann:
Instead of recalculating autorange for each call to addItem, defer the calculation until the next paint event.

I was experiencing a lot of lag when creating >1000 curve items with 100 points each. Profiling showed that 70% of the runtime was spent in updateAutoRange. With this patch, updateAutoRange is less than 1% of the runtime.

### DIFF
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -436,7 +436,7 @@ class ViewBox(GraphicsWidget):
 
         if not ignoreBounds:
             self.addedItems.append(item)
-        self.updateAutoRange()
+        self.queueUpdateAutoRange()
 
     def removeItem(self, item):
         """Remove an item from this view."""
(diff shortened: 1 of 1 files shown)


j9ac9k:
Here be dragons in this part of the code, would need to test on a number of platforms, and with log-mode to ensure plots are behaving as appropriate.  I can't remember which PR touched on this last (would need to look at git blame) but we had some very undesirable consequences if memory serves...

agausmann:
I completely understand!

Give me some time and I can make an example script and open an issue. If this PR doesn't work out, we can still track the performance issue and discuss other solutions.

j9ac9k:
Not trying to suggest this PR is a nogo, just that it would need more scrutiny than one would think

j9ac9k:
Hi @agausmann 

Sorry for the year+ wait on this from my end.  To say it's been a chaotic year for me (outside of pyqtgraph) would be an understatement.

On further reflection, I think the PR is fine.  Unfortunately as sensitive as ViewBox.py is, there is very poor test coverage of it, and I don't think it's fair to punish contributors with holding up merges because of that.  

My skepticism came from some extreme frustrations from debugging: https://github.com/pyqtgraph/pyqtgraph/issues/1373  This PR doesn't actually touch on that logic, which I think makes it safe to merge.  Thanks for your contribution @agausmann

j9ac9k:
Closing/Re-Opening to trigger a new CI run

Issue #2110, opened 2021-11-18
`test_PolyLineROI` fails on aarch64 linux


risicle:
pyqtgraph 0.12.3's `test_PolyLineROI` fails on aarch64 linux systems when comparing `closed_init` with the reference image.

```
        pxdiff = diff.max(axis=2)  # largest value difference per pixel
        mask = np.abs(pxdiff) >= pxThreshold
        if pxCount is not None:
>           assert mask.sum() <= pxCount
E           AssertionError
```

Generated image: [closed_init](https://user-images.githubusercontent.com/807447/142303551-8db8c005-a5ec-405b-8fa1-4b89f7709a88.png)

Full build log visible at https://hydra.nixos.org/log/16xzbqm8fif20l3iq3m45i0s1k4a1c93-python3.8-pyqtgraph-0.12.2.drv - this is actually from 0.12.2 but 0.12.3's is essentially identical.

This is with:
 - numpy-1.21.2
 - PyQt5-5.15.4
 - qt 5.15.3
 - scipy-1.7.1
 - pyopengl-3.1.4
 - both python 3.8 & 3.9

pijyoi:
I had noticed this before on the Raspberry Pi 3B. This failure occurs on aarch64 but not arm32.

On the recently released Raspberry Pi OS 32-bits "Bullseye", the test passes.
On both ArchLinux and Ubuntu aarch64, the test fails.

The test does image comparison against a reference image and can be very sensitive to slight pixel differences, even if the differences don't impact usage of the library.

risicle:
I think it's just a matter of the tolerances being too tight.

This is the build process of the NixOS package. We can always disable that test on aarch64, it just seems better not to leave a known failure.

j9ac9k:
We have expanded on the tolerance for image testing based on platform, I wouldn't mind increasing the tolerance for the image test for this platform a bit.  I'm on mobile so I can't find the exact place in the library we modify toleraces based on platform.

risicle:
Why vary it by platform? If a result is good enough for one platform why is it not good enough for another?

j9ac9k:
> Why vary it by platform? If a result is good enough for one platform why is it not good enough for another?

totally depends on what the test is trying to catch.  In this case, the test is trying to catch the case of someone making a change in the library that makes a difference on the positioning of pixels being rendered.

[...]

PR #2949, opened 2024-03-04
close ROI handle subpath

pijyoi:
this ensures that the path is a closed polygon.
this matters for PenCap and PenJoin purposes.

`test_PolylineROI` has the following series of test images:
1) init
2) hover_roi
3) drag_roi
4) hover_handle
5) drag_handle
6) hover_segment
7) click_segment
8) drag_new_handle
9) clear

Prior to this PR, aarch64 (and ppc64le) platforms fail at `init`. (See #2110)
With this PR, the failure occurs later at `click_segment` (and `drag_new_handle`)

When the path is not closed, we end up with a polyline with its ends rendered with `SquareCap`, which according to Qt docs: "a square line end that covers the end point and **_extends beyond it by half the line width_**."
It does so happen that the failing pixels in `init` correspond to the closing point of the ROI handle. 

Failure screenshot for `click_segment`. For this case, a lot more pixels are involved.
Screenshot 2024-03-04 230107

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -1468,14 +1468,15 @@ class Handle(UIGraphicsItem):
         self.path = QtGui.QPainterPath()
         ang = self.startAng
         dt = 2 * np.pi / self.sides
-        for i in range(0, self.sides+1):
+        for i in range(0, self.sides):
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for the investigation @pijyoi and the partial improvement!

PR #2963, opened 2024-03-15
ImageView example: use levelMode="rgba"

pijyoi:
Since #1611, the image data used in this example has been changed to RGB.
Using `rgba` mode is more appropriate here.

Setting a colormap has no effect on RGB images, so delete those lines too.


j9ac9k:
Thanks @pijyoi for keeping the examples up to date!

PR #2964, opened 2024-03-15
Aaryamantriescode SyntaxWarnings #2962

### LINKED ISSUE #2962: SyntaxWarnings
While running an `apt upgrade` I noticed:

```
/usr/lib/python3/dist-packages/pyqtgraph/examples/SpinBox.py:38: SyntaxWarning: invalid escape sequence '\$'
  regex='\$?(?P<number>(-?\d+(\.\d+)?)|(-?\.\d+))$')),
```

The `\$` should be written `\\$` or `r'\$'` since a few Python releases (same for all backslash escape that have no meanings). I don't have the time to search for other occurrences of this fact, but running the tests with `PYTHONDEVMODE=1` should help spotting them :)

aaryamantriescode:
Fixes #2962
This was the issue [The \$ should be written \\$ or r'\$' since a few Python releases (same for all backslash escape that have no meanings). I don't have the time to search for other occurrences of this fact, but running the tests with PYTHONDEVMODE=1 should help spotting them :)]
I think I  found the issue and made changes

### DIFF
--- a/pyqtgraph/examples/SpinBox.py
+++ b/pyqtgraph/examples/SpinBox.py
@@ -35,7 +35,7 @@ spins = [
      pg.SpinBox(value=1e9, siPrefix=True)),
     ("Float with custom formatting", 
      pg.SpinBox(value=23.07, format='${value:0.02f}',
-                regex=r'\\$?(?P<number>(-?\\d+(\\.\\d+)?)|(-?\\.\\d+))$')),
+                regex = r'\\$?(?P<number>(-?\d+(\.\d+)?)|(-?\.\d+))$')),
     ("Int with suffix",
      pg.SpinBox(value=999, step=1, int=True, suffix="V")),
     ("Int with custom formatting", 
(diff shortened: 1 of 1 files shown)


ixjlyons:
Slight preference for a raw string (`r'...'`) rather than escaping. I think I would go ahead and do the same for the other example here that sets a custom regex, though it's not strictly necessary.

aaryamantriescode:
Ok ill change that

j9ac9k:
Hi @aaryamantriescode 

Thanks for the PR, when using raw strings (`r''`) you should not take care to escape special characters using the `/`.   Go ahead and remove the extra `/` characters in this PR, and that should be good 👍🏻

aaryamantriescode:
Is this fine now?

j9ac9k:
image

Need to remove the first (or second) slash here in the screen shot.

aaryamantriescode:
😅got it

j9ac9k:
Thanks for the PR @aaryamantriescode ❤️

PR #2965, opened 2024-03-16
set SymbolAtlas dpr to primary screen's dpr

### LINKED ISSUE #2960: PyQtGraph 0.1.14 takes approximately twice as long to plot colored points in a ScatterPlotItem that 0.1.13
In PyQtGraph version `0.13.4`, coloring points in a `ScatterPlotItem` with `brush` set to a long index of integers (to rotate over the default number of colors) takes approximately twice as long as in PyQtGraph `0.13.3`.

Example code:

```python
...
self.scatter = pg.ScatterPlotItem(
    size=5,
    pen=pg.mkPen(None),
    brush=pg.mkBrush(255, 255, 255, 128),  # To be overridden in the next call depending on a setting
    hoverable=True,
    hoverSymbol="s",
    hoverSize=5,
    hoverPen=pg.mkPen("w", width=2),
    hoverBrush=None,
)
...
self.scatter.addPoints(x=x, y=y, data=tid, brush=brushes,)  # Here brushes is equal to tid
``` 

For `x`, `y`, and `tid` of about 200,000 elements, and `np.unique(tid)` of approximately 600, on my machine PyQtGrpah `0.1.13` takes approximately **10 seconds** to plot all colored points, whereas `0.1.14` takes approximately **19 seconds**.

_Originally posted by @aarpon in https://github.com/pyqtgraph/pyqtgraph/issues/2719#issuecomment-1997515263_

pijyoi:
Set the startup DPR of `ScatterPlotItem`'s `SymbolAtlas` to the primary screen's DPR, so that users with hidpi screens are not penalized by a guaranteed `SymbolAtlas` rebuild.

https://doc.qt.io/qt-6/qguiapplication.html#primaryScreen-prop says that:
`This will be the screen where QWindows are initially shown, unless otherwise specified.`

Fixes #2960
Test script can be found in the comments of the linked issue.

Remarks:
1) `ScatterPlotItem` can (surprisingly) be instantiated without a `QGuiApplication`, in which case there will be no primary screen.

### DIFF
--- a/pyqtgraph/graphicsItems/ScatterPlotItem.py
+++ b/pyqtgraph/graphicsItems/ScatterPlotItem.py
@@ -361,6 +361,8 @@ class ScatterPlotItem(GraphicsObject):
 
         self.picture = None   # QPicture used for rendering when pxmode==False
         self.fragmentAtlas = SymbolAtlas()
+        if screen := QtGui.QGuiApplication.primaryScreen():
+            self.fragmentAtlas.setDevicePixelRatio(screen.devicePixelRatio())
 
         dtype = [
             ('x', float),


j9ac9k:
walrus operator strikes again!  

Thanks for the PR @pijyoi LGTM 👍🏻

PR #2970, opened 2024-03-23
add fastpath for float images with nans

pijyoi:
This PR augments #1693 by adding support for float images containing nans. 

To test it out.
```
set PYQTGRAPHPROFILE="ImageItem.paint"
```
Then run the following script and change the levels using the `HistogramLUTItem`. The GUI should feel more responsive with this PR versus master.
```python
import pyqtgraph as pg
import numpy as np
# pg.setConfigOptions(useNumba=True)

Y = np.sinc(np.linspace(-6, 10, 4000))[:, np.newaxis]
X = np.sinc(np.linspace(-15, 9, 6000))
data = np.abs(Y * X)
data[data > 0.95] = -1
data = 10*np.log10(data)
data = data.astype(np.float32)

pg.mkQApp()
imv = pg.ImageView()
imv.show()
imv.setImage(data, levels=(-50, 0))
hlut = imv.getHistogramWidget()
hlut.gradient.setColorMap(pg.colormap.get("CET-C1"))
hlut.gradient.showTicks(False)
pg.exec()
```

timings from `PYQTGRAPHPROFILE`
|   |row-major master | row-major this PR | col-major master | col-major this PR |
| -------- | -------- | ------- | ------- | ------- |
| numba | 224 ms | 35 ms | 558 ms | 165 ms |
| numpy | 258 ms | 128 ms | 303 ms | 252 ms |

numba codepath with col-major is no longer worse than numpy. If you were interested in performance, you wouldn't be using col-major anyway...

### DIFF
--- a/pyqtgraph/functions_qimage.py
+++ b/pyqtgraph/functions_qimage.py
@@ -6,38 +6,63 @@ from .util.cupy_helper import getCupy
 from .util.numba_helper import getNumbaFunctions
 
 
-def _apply_lut_for_uint16_mono(xp, image, lut):
+def _apply_lut_for_uint(xp, image, lut):
     # Note: compared to makeARGB(), we have already clipped the data to range
-    augmented_alpha = False
 
     # if lut is 1d, then lut[image] is fastest
     # if lut is 2d, then lut.take(image, axis=0) is faster than lut[image]
+    lut = _convert_2dlut_to_1dlut(xp, lut)
 
-    if not image.flags.c_contiguous:
-        image = lut.take(image, axis=0)
-
-        # if lut had dimensions (N, 1), then our resultant image would
-        # have dimensions (h, w, 1)
-        if image.ndim == 3 and image.shape[-1] == 1:
-            image = image[..., 0]
-
-        return image, augmented_alpha
-
-    # if we are contiguous, we can take a faster codepath where we
-    # ensure that the lut is 1d
-
-    lut, augmented_alpha = _convert_2dlut_to_1dlut(xp, lut)
-
-    fn_numba = getNumbaFunctions()
-    if xp == numpy and fn_numba is not None:
+    if xp == numpy and (fn_numba := getNumbaFunctions()) is not None:
+        # numba "take" supports only the 1st 2 arguments of np.take,
+        # therefore we have to convert the lut to 1d.
+        # "take" will output a c contiguous array regardless of its input.
         image = fn_numba.numba_take(lut, image)
     else:
+        # advanced indexing is memory order aware.
+        # its output can be either C or F contiguous.
         image = lut[image]
 
     if image.dtype == xp.uint32:
+        # "view" requires c contiguous for numpy < 1.23
+        image = xp.ascontiguousarray(image)
         image = image[..., xp.newaxis].view(xp.uint8)
 
-    return image, augmented_alpha
+    return image
+
+
+def _convert_lut_to_rgba(xp, lut):
+    # converts:
+    #   - None to (256, 4)
+    #   - uint8 (N,) to uint8 (N, 4)
+    #   - uint8 (N, 1) to uint8 (N, 4)
+    #   - uint8 (N, 3) to uint8 (N, 4)
+
+    if not (
+        lut is None
+        or lut.ndim == 1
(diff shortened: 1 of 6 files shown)


pyqtgraph/functions_qimage.py
```diff
@@ -40,6 +40,32 @@ def _apply_lut_for_uint16_mono(xp, image, lut):
...
+    N = lut.shape[0] if lut is not None else 256
+
+    if lut is None:
+        lut = xp.arange(N, dtype=xp.uint8)
+
+    # convert (N,) to (N, 1)
+    if lut.ndim == 1:
+        lut = lut[:, xp.newaxis]
+
+    assert lut.ndim == 2 and lut.shape[1] in (1, 3, 4)
```
j9ac9k, line 59: Is an `assert` statement what we actually want here? These are bypassed when Python is run with the `-O` argument.

pyqtgraph/functions_qimage.py
```diff
@@ -61,7 +87,13 @@ def _convert_2dlut_to_1dlut(xp, lut):
     return lut, augmented_alpha
 
 
-def _rescale_float_mono(xp, image, levels, lut):
+def _rescale_and_lookup_float(xp, image, levels, lut, *, forceApplyLut):
+    # It is usually more performant to _not_ apply the lut and
+    # instead use it as an Indexed8 ColorTable. This is only
+    # applicable if the lut has <= 256 entries.
+
+    assert (not forceApplyLut) or (lut is not None)
```
j9ac9k, line 95: Same comment about `assert` as earlier.

j9ac9k (commented):
Thanks for the PR @pijyoi 

As always, really appreciate the comments, it helps a ton as I read through this.

My only comment is regarding the assert statements, which can easily be bypassed if running Python with the `-O` argument.  Outside of test code we should probably just have an if check and raise an exception of some kind (I don't have strong opinions on what exceptions to raise but maybe one of the numpy ones?)

pyqtgraph/functions_qimage.py
pijyoi: This one makes sense to be an exception.

pyqtgraph/functions_qimage.py
pijyoi: I am not so sure that this makes more sense as an exception. Sometimes I use assertions as comments to document the code flow.

I converted it to an exception but I moved it to the top of the function, otherwise it doesn't make sense to raise a complaint that the lut was not 2d.

pijyoi:
While improving the numba performance for the float with nans case, we also found out how to improve the numba performance for the float with no nans case. (24ms --> 12ms)

The gap between numpy vs numba has widened.

pijyoi:
Change uint16 mono images to use `functions_numba.rescale_only` instead of `functions_numba.rescaleData`.
Timings have halved.

pijyoi:
Added support for float RGB images with nans.
The supported data types can thus be summarized as (uint8, uint16, float32, float64) x (1 chan, 3 chans) provided no multi-channel levels are used.

```python
import pyqtgraph as pg
import numpy as np
# pg.setConfigOptions(useNumba=True)

bias = 50

Y = np.sinc(np.linspace(-6, 10, 4000))[:, np.newaxis]
X = np.sinc(np.linspace(-15, 9, 6000))
data = np.abs(Y * X)
mask1 = data > 0.95
mask2 = (0.5 < data) & (data < 0.6)
data = 10*np.log10(data)
data += bias
data = np.dstack((data, data * 0.5, data * 0.25))
data[..., 1][mask1] = np.nan
data[..., 2][mask2] = np.nan
data = data.astype(np.float32)

pg.mkQApp()
imv = pg.ImageView(levelMode='mono')
imv.show()
imv.setImage(data, levels=(-50 + bias, 0 + bias))
pg.exec()
print(imv.imageItem.qimage.format())
```

j9ac9k:
that numba `rescale_and_clip` function seems to be exactly what's needed haha, wow, that's _really_ fast.

j9ac9k:
Every time I think you've gotten as much out of ImageItem performance as we can get, you prove me wrong 😆  Thanks for the PR @pijyoi Merging.

PR #2972, opened 2024-03-27
Add stubs for Qt modules

maflAT:
Enables editor and type checker support for Qt classes.

This change provides type information to editors and type checkers.

VSCode / Pylance will still report the import statements in the new stubs as missing imports if the given package is not installed. Nevertheless, if at least one of the packages is installed, you can import and use those classes with full auto-complete support.

It would be nice if someone could test if this also works in PyCharm or other editors / type checkers. Unfortunately I'm not able to do so.

### Other Tasks 

- [x] Check interaction with VSCode / Pylance
- [ ] Check interaction with PyCharm

### DIFF
--- a/pyqtgraph/Qt/QtCore/__init__.pyi
+++ b/pyqtgraph/Qt/QtCore/__init__.pyi
@@ -1,0 +1,5 @@
+# Import packages in reverse libOrder as defined in pyqtgraph/Qt/__init__.py
+from PySide2.QtCore import *
+from PyQt5.QtCore import *
+from PySide6.QtCore import *
+from PyQt6.QtCore import *
--- a/pyqtgraph/Qt/QtGui/__init__.pyi
+++ b/pyqtgraph/Qt/QtGui/__init__.pyi
@@ -1,0 +1,5 @@
+# Import packages in reverse libOrder as defined in pyqtgraph/Qt/__init__.py
+from PySide2.QtGui import *
+from PyQt5.QtGui import *
+from PySide6.QtGui import *
+from PyQt6.QtGui import *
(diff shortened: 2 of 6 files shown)


j9ac9k:
Hi @maflAT 

My apologies for not chiming in earlier... thanks for the PR! 

This issue has admittedly been a sore point for us for some time... I will test this out right now on both vscode and pycharm 👍🏻

j9ac9k:
Can confirm works on vscode: 
image

Can confirm working on pycharm

image

Testing on _not_ PyQt6 also works, thanks so much for this PR @maflAT !!!

maflAT:
@j9ac9k my pleasure, thanks for the help with checking!

j9ac9k:
Hi @maflAT just a follow up on this, do you know if there is a limitation of reading stubs from relative paths?

pyqtgraph internally uses relative paths, but seems like pyright is not reading the stubs.

example:

Import statement:

```python
from ..Qt import QtCore, QtGui, QtWidgets
```

Usage of one of the modules:

image

maflAT:
@j9ac9k I'm not aware of any specific limitations.
However, when I tried to reproduce this, Pylance was acting up for me too.
In one of my virtualenvs it didn't recognize PySide6 at all, even when directly importing PySide6.
In another environment, it works as expected:
grafik

This is with Pylance v2024.5.102 and PySide6 installed (no other Qt bindings).

j9ac9k:
Weird!!! In my case I'm using PyQt6, but I'll try and experiment with more bindings and combinations of dependencies.

Other potential oddity is that I have PyQtGraph installed as an editable install; maybe that behaves differently vs a regular install.

I'll try and tinker with it more when I'm on my computer next.

PR #2974, opened 2024-03-31
numpy 2.0: workaround regression in rescaleData

pijyoi:
Testing against NumPy 2.0rc1, a performance regression was noticed in `functions.rescaleData`.

The cause was traced to:
https://numpy.org/devdocs/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion

It does seem rather counter-intuitive that the new behavior should affect in-place operations:
e.g.
```python
data -= offset
data *= scalar
```
i.e. the resultant `dtype` can't change, so the performance regression might imply that a temporary higher precision ndarray was created? This would defeat the purpose of doing in-place operations.

~The fix in this PR is only applied to the `numpy` codepath, and not to the `cupy` and `numba` codepaths.
The `numba` codepath (`functions_numba.rescaleData`) already marks the `offset` and `scale` arguments as type `f8`.~

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1274,6 +1274,14 @@ def rescaleData(data, scale, offset, dtype=None, clip=None):
     else:
         work_dtype = np.float64
 
+    # from: https://numpy.org/devdocs/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for keeping the library up to date w/ issues with upcoming releases of our dependencies!

PR #2992, opened 2024-04-20
Set Application Icon on windows

j9ac9k:
This PR makes it so the peegee icon is visible on the taskbar on Windows.  Furthermore, we set the `fusion` theme to applications created using `mkQApp()`, which handled dark mode detection on windows with Qt 6.5+.  

What still isn't working is the peegee icon not showing up on gnome/kde desktops.

### DIFF
--- a/pyqtgraph/examples/ExampleApp.py
+++ b/pyqtgraph/examples/ExampleApp.py
@@ -362,10 +362,11 @@ class ExampleLoader(QtWidgets.QMainWindow):
             try:
                 darkMode = app.styleHints().colorScheme() == QtCore.Qt.ColorScheme.Dark
             except AttributeError:
-                palette = app.pallete()
+                palette = app.palette()
                 windowTextLightness = palette.color(QtGui.QPalette.ColorRole.WindowText).lightness()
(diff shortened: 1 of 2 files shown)


pyqtgraph/examples/ExampleApp.py
```diff
@@ -357,6 +356,20 @@ def onComboChanged(searchType):
         self.codeBtn.clicked.connect(self.runEditedCode)
         self.updateCodeViewTabWidth(self.ui.codeView.font())
 
+    def event(self, event):
+        if event.type() == QtCore.QEvent.Type.ApplicationPaletteChange:
+            app = QtWidgets.QApplication.instance()
+            try:
+                darkMode = app.styleHints().colorScheme() == QtCore.Qt.ColorScheme.Dark
+            except AttributeError:
+                palette = app.pallete()
```
pijyoi, line 365: A typo here?

j9ac9k: > A typo here?

...yes

j9ac9k:
actually while this looked good on Windows, just loaded up macOS with pyqt 5.15, and the example app looks like rubbish there, ...needs more investigation before merging.

PR #2993, opened 2024-04-20
Remove deprecated use of paletteChanged

j9ac9k:
Also try and address some of the downsizes of using `fusion` style on Windows with the odd alternating row colors for ParameterTree.

### DIFF
--- a/pyqtgraph/parametertree/ParameterTree.py
+++ b/pyqtgraph/parametertree/ParameterTree.py
@@ -21,7 +21,6 @@ class ParameterTree(TreeWidget):
         self.setAnimated(False)
         self.setColumnCount(2)
         self.setHeaderLabels(["Parameter", "Value"])
-        self.setAlternatingRowColors(True)
         self.paramSet = None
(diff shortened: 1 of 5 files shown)


pyqtgraph/parametertree/ParameterTree.py
```diff
@@ -30,9 +30,8 @@ def __init__(self, parent=None, showHeader=True):
         self.itemCollapsed.connect(self.itemCollapsedEvent)
         self.lastSel = None
         self.setRootIsDecorated(False)
-
+        self.setAlternatingRowColors(False)
         app = mkQApp()
```
pijyoi, line 34: can be removed now

j9ac9k: oh, just noticed 10 lines up where that property is set to `True` ... whoops, yeah just remove those entries and that's better...

It's too bad on Windows the alternate color that is picked is the "accent" color on the windows theme, not something subtly different

j9ac9k:
alright, I'm done tinkering ...

PR #2999, opened 2024-04-21
test_PolyLineROI now passes on non-AMD64 platforms

### LINKED ISSUE #2998: Test failure on test_PolyLineROI on arm64
CI is failing for `test_PolyLineROI` and in one case results in a segfault.  The test failure can be replicated locally.

When setting `export PYQTGRAPH_AUDIT=1` we get an image demonstrating the failure.

image

Note: the segfault is avoided when running `pytest tests -k "not test_PolyLineROI"`

j9ac9k:
The test suite now passes due to using a forgiving pxCount values when the test suite is running on not on `platform.machine() == 'AMD64'`.

Did formatting changes to `test_PolyLineROI` and migrated the for loop to instead use `pytest.mark.parametrize`

Question now is, is this the "fix" we want to have?

An alternative would be to add new images for arm based platforms to test against.

Fixes #2998


j9ac9k:
```python
pxCount=0 if platform.machine() == "AMD64" else 36
```
😬

j9ac9k:
ok, done tinkering, I think this PR is ready.  I wish I had a better idea why the segfault in make_ARGB, but setting the `im1` arrays to read only seems to resolve whatever the issue is there.

PR #3007, opened 2024-04-28
Add note about NaN to int conversion

j9ac9k:
Fixes #2969

PR #3008, opened 2024-04-28
Fixup 3007

j9ac9k:
Re-add commented out tests in #3007

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -34,7 +34,7 @@ __all__ = [
     'solve3DTransform', 'solveBilinearTransform',
     'clip_scalar', 'clip_array', 'rescaleData', 'applyLookupTable',
     'makeRGBA', 'makeARGB',
-    # 'try_fastpath_argb', 'ndarray_to_qimage',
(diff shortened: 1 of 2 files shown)


pyqtgraph/functions.py
```diff
@@ -1257,7 +1257,7 @@ def rescaleData(data, scale, offset, dtype=None, clip=None):
         # Cupy does not support nditer
         # https://github.com/cupy/cupy/issues/5021
 
-        data_out = data.astype(work_dtype, copy=True)
+        data_out = data.astype(work_dtype)
```
pijyoi, line 1260: I think `copy=True` is necessary as we are going to mutate the array in the next 2 statements

pijyoi: Oh I see, `copy=True` is the default

pijyoi: Oh, I remember why it was there. To make it symmetric to the `return data_out.astype(out_dtype, copy=False)`

j9ac9k: oh no..... I deleted the wrong `copy=True`

PR #3014, opened 2024-04-29
test image formats outside of ImageItem

pijyoi:
This was originally in #2944, but was triggering a segfault in the CI.
Now that the segfault has been resolved by #3008, we can put it back in.

Note that this also fixes a usage error in the test(s) for multichannel images with nan. Providing a lut to `ImageItem` has no effect for multichannel images.


j9ac9k:
look at all those green check-boxes 😍

PR #3031, opened 2024-05-22
RawImageGLWidget: implement using QOpenGLTextureBlitter

pijyoi:
This PR reimplements RawImageGLWidget without use of PyOpenGL, using instead QtOpenGL.
Some consequences:
1) RawImageGLWidget is always available for use in VideoSpeedTest
2) RawImageGLWidget now also works on GLES
3) OpenGL object leaks are more visible since Qt will warn.


j9ac9k:
Thanks for the PR @pijyoi  Love seeing RawImageGLWidget getting to be functional with VideoSpeedTest again!

PR #3045, opened 2024-06-05
enable blending to take into account alpha transparency

pijyoi:
The bundled peegee png files are examples of images that make use of transparency and do not look right on `RawImageGLWidget` without blending enabled.

The script from #3043 can be used to test.


j9ac9k:
Thanks @pijyoi It's awesome to have a basis to look at when doing opengl stuff.

PR #3047, opened 2024-06-05
Fix types of signals of `SpinBox`

StSav012:
In the constructors of the signals, use the actual types of the values passed to the corresponding `emit` calls. Namely, there should be `float`s in two places.

I haven't checked other classes for the same issue. Please take a look if you have time.

### DIFF
--- a/pyqtgraph/widgets/SpinBox.py
+++ b/pyqtgraph/widgets/SpinBox.py
@@ -41,9 +41,9 @@ class SpinBox(QtWidgets.QAbstractSpinBox):
     ## I have no idea how to get around this..
     
     
-    valueChanged = QtCore.Signal(object)     # (value)  for compatibility with QSpinBox
+    valueChanged = QtCore.Signal(float)     # (value)  for compatibility with QSpinBox
     sigValueChanged = QtCore.Signal(object)  # (self)
-    sigValueChanging = QtCore.Signal(object, object)  # (self, value)  sent immediately; no delay.
+    sigValueChanging = QtCore.Signal(object, float)  # (self, value)  sent immediately; no delay.
 
     def __init__(self, parent=None, value=0.0, **kwargs):
         """


StSav012:
> The reasoning I suspect was due to various issues with Qt

On the contrary, consider the following:

```plaintext
>>> from pyqtgraph.widgets.SpinBox import SpinBox
>>> SpinBox.valueChanged
<unbound PYQT_SIGNAL double)>
>>> SpinBox.sigValueChanging
<unbound PYQT_SIGNAL PyQt_PyObject,double)>
```

There are actually `double`s there, so I'd expect to see `float` in Python.

Please see if the following code works for you in two cases: 
1. when `valueChanged = QtCore.Signal(object)` and
1. when `valueChanged = QtCore.Signal(float)`.

In both cases, I used `PyQt6`. I saw the same with `PyQt5`, but not with `PySide6`.

```python
# coding=utf-8
import sys

from pyqtgraph.widgets.SpinBox import SpinBox
from pyqtgraph.Qt import QtCore, QtWidgets


class C(QtWidgets.QWidget):
    @QtCore.pyqtSlot(float)
    def callback(self, value: float) -> None:
        print(value)


if __name__ == '__main__':
    app: QtWidgets.QApplication = QtWidgets.QApplication(sys.argv)
    c: C = C()
    w: SpinBox = SpinBox()
    w.valueChanged.connect(c.callback)
```

I see the following error in the first case:
```none
  File "pyqtgraph-pr-3047.py", line 18, in <module>
    w.valueChanged.connect(c.callback)
TypeError: decorated slot has no signature compatible with SpinBox.valueChanged[object]

Process finished with exit code 1
```
In the second case, everything is fine.

j9ac9k:
> > The reasoning I suspect was due to various issues with Qt
> 
> On the contrary, consider the following:

Sorry, I meant _historically_, like early Qt4 days ... there's a lot of code in this codebase to work around Qt issues that are no longer relevant.  

I appreciate the example, will take a closer look later today, but yeah, there's a whole lot of `Signal(object, ...)`  bits throughout the codebase.

pijyoi:
Based on the outcome of your example, if someone had previously decorated their slot as "pyqtSlot(object)", wouldn't this PR break their code?

Changing the signature of the signals _is_ an api change.

j9ac9k:
What I'm unclear of is what is the benefit of being more specific about types in the Signal and Slot declarations.  I get that it would be nice to be able to declare more specific types in the decorator, such as `float` or `int`, but does that actually fix/improve anything on a functional level, or is it merely a better declarative representation?

Also as @pijyoi said, being more specific w/ the types would break stuff for existing users who have decorated their slots with the generic `object` type.

StSav012:
> wouldn't this PR break their code?

Yeah, unfortunately. I'd expect the `object` slot to catch all the argument types, since `object` is the base class for all other classes. In other words, `issubclass(float, object)` is `True`. Still, the casting never happens.

> or is it merely a better declarative representation?

The [docs](https://doc.qt.io/qtforpython-6/tutorials/basictutorial/signals_and_slots.html#specifying-signals-and-slots-by-method-signature-strings) read
> Slots should be decorated using [@Slot](https://doc.qt.io/qtforpython-6/tutorials/basictutorial/signals_and_slots.html#slot-decorator).

The reason is a mystery to me. I don't see a difference in the execution, although this can raise a `TypeError` if the types given to a `@Slot` mismatch what the signal sends. With the flexibility of Python, this adds a thin layer of protection.

j9ac9k:
I agree that pyqtgraph should have the slots decorated (and they're not right now).  What I'm less convinced of is the utility on decorating the slots with precise types (in this case, `float`) as opposed to a generic `object`.  To introduce a potential API break, I'd need a stronger case to explain why it's needed.

StSav012:
Fair point.

j9ac9k:
Not saying we have to close the issue yet, there may be a functional need that is addressed by more accurate annotations; but I just don't know.

PR #3049, opened 2024-06-06
PlotItem.autoBtnClicked to emit PlotItem.sigRangeChangedManually

tikuma-lsuhsc:
As all the mouse interactions in `ViewBox` emits `ViewBox.sigRangeChangedManually`, it seems logical for pressing the "auto" button on `PlotItem` to also emit this signal.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
+++ b/pyqtgraph/graphicsItems/PlotItem/PlotItem.py
@@ -71,12 +71,13 @@ class PlotItem(GraphicsWidget):
     
     The ViewBox itself can be accessed by calling :func:`getViewBox() <pyqtgraph.PlotItem.getViewBox>` 
     
-    ==================== =======================================================================
+    ======================= =======================================================================
     **Signals:**
-    sigYRangeChanged     wrapped from :class:`ViewBox <pyqtgraph.ViewBox>`
-    sigXRangeChanged     wrapped from :class:`ViewBox <pyqtgraph.ViewBox>`
-    sigRangeChanged      wrapped from :class:`ViewBox <pyqtgraph.ViewBox>`
-    ==================== =======================================================================
+    sigYRangeChanged        wrapped from :class:`ViewBox <pyqtgraph.ViewBox>`
+    sigXRangeChanged        wrapped from :class:`ViewBox <pyqtgraph.ViewBox>`
+    sigRangeChanged         wrapped from :class:`ViewBox <pyqtgraph.ViewBox>`
+    sigRangeChangedManually wrapped from :class:`ViewBox <pyqtgraph.ViewBox>`
+    ======================= =======================================================================
     """
     
     sigRangeChanged = QtCore.Signal(object, object)    ## Emitted when the ViewBox range has changed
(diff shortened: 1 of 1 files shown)


tikuma-lsuhsc:
Here is a script to test the added feature.

```python
import numpy as np
from PyQt6.QtCore import pyqtSlot
import pyqtgraph as pg

app = pg.mkQApp("Plotting Example")
win = pg.GraphicsLayoutWidget(show=True, title="Basic plotting examples")
p1 = win.addPlot(y=np.random.normal(size=100))
p1.vb.setMouseEnabled(True, True)

@pyqtSlot(object)
def rangeChangedManually(mask):
    print(f'sigRangeChangedManually emitted ({mask})')

p1.vb.sigRangeChangedManually.connect(rangeChangedManually)

if __name__ == "__main__":
    pg.exec()
```

j9ac9k:
Hi @tikuma-lsuhsc 

Thanks for the PR, sorry for the long wait to reply, I recently had a large move.

My first question is why emit the signal inside the PlotItem instance, but not have it emitted as part of `ViewBox.enableAutoRange` instead?

hokiedsp:
@j9ac9k - I'm glad you got around to my PRs. Thanks for reviewing them. 

> why emit the signal inside the PlotItem instance, but not have it emitted as part of ViewBox.enableAutoRange instead

Because it was the path of the least resistance although I agree with you on its awkwardness. I didn't think it would be appropriate for `ViewBox.enableAutoRange` to emit the `sigRangeChangedManually` signal as the signal name suggests that it ought to be triggered by user (UI). Its emission by `enableAutoRange` introduces a potential emission by a non-user command (at least not directly). For example, I can see myself using `enableAutoRange` to reset the view when I load a new data. (There was a specific reason I needed this signal emitted in my app, but it has escaped me at the moment.)

Hope this makes sense. I'm 100% open to your suggestions/decisions and will be happy to modify the PR accordingly.

tikuma-lsuhsc:
@j9ac9k - I'm glad you got around to my PRs. Thanks for reviewing them. 

> why emit the signal inside the PlotItem instance, but not have it emitted as part of ViewBox.enableAutoRange instead

Because it was the path of the least resistance although I agree with you on its awkwardness. I didn't think it would be appropriate for `ViewBox.enableAutoRange` to emit the `sigRangeChangedManually` signal as the signal name suggests that it ought to be triggered by user (UI). Its emission by `enableAutoRange` introduces a potential emission by a non-user command (at least not directly). For example, I can see myself using `enableAutoRange` to reset the view when I load a new data. (There was a specific reason I needed this signal emitted in my app, but it has escaped me at the moment.)

Hope this makes sense. I'm 100% open to your suggestions/decisions and will be happy to modify the PR accordingly.

j9ac9k:
Actually you can chain signals together so we can just tie this signal to the signal that is emitted when the button is pressed. I'm on mobile right now so I can't look it up but maybe that signal makes more sense there?

j9ac9k:
Sorry for the long wait on merging this @tikuma-lsuhsc 

This PR LGTM, merging.  Thanks for the contribution to the library!

PR #3050, opened 2024-06-06
Transform3D.map should return the type of obj passed in

### LINKED ISSUE #3044: Segfault when iterating on a transform-mapped Vector in a thread
### Short description
Take a Vector, map it with a transform, put it in a _threading_ Thread, iterate, and BOOM.

### Code to reproduce

```python
import pyqtgraph as pg
import threading


def do_it():
    xform = pg.SRTTransform3D()
    v = xform.map(pg.Vector((0, 0, 0)))
    tuple(v)


pg.mkQApp()
threading.Thread(target=do_it).start()
```

### Real behavior

`SIGSEGV`

### Tested environment(s)

 * PyQtGraph version: 0.13.8dev0
 * Qt Python binding: PyQt5 5.15.10 Qt 5.15.2
 * Python version: 3.11
 * NumPy version: 1.26.3
 * Operating system: linux
 * Installation method: git

### Additional context

Workaround: use a `QThread`, and this isn't an issue.

outofculture:
Closes #3044

### DIFF
--- a/pyqtgraph/Transform3D.py
+++ b/pyqtgraph/Transform3D.py
@@ -1,5 +1,3 @@
-import warnings
-
 import numpy as np
 
 from . import functions as fn
--- a/pyqtgraph/Transform3D.py
+++ b/pyqtgraph/Transform3D.py
@@ -47,8 +47,11 @@ class Transform3D(QtGui.QMatrix4x4):
             v = QtGui.QMatrix4x4.map(self, Vector(obj))
             return type(obj)([v.x(), v.y(), v.z()])[:len(obj)]
         else:
-            return type(obj)(QtGui.QMatrix4x4.map(self, obj))
-            
+            retval = QtGui.QMatrix4x4.map(self, obj)
+            if not isinstance(retval, type(obj)):
+                return type(obj)(retval)
+            return retval
+
     def inverted(self):
         inv, b = QtGui.QMatrix4x4.inverted(self)
         return Transform3D(inv), b
--- a/pyqtgraph/Transform3D.py
+++ b/pyqtgraph/Transform3D.py
@@ -1,3 +1,5 @@
+import warnings
+
 import numpy as np
 
 from . import functions as fn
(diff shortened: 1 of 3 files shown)


pyqtgraph/Transform3D.py
```diff
@@ -1,3 +1,5 @@
+import warnings
```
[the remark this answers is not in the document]
j9ac9k: Can you remove this unused import?

tests/test_srttransform3d.py
```diff
@@ -36,3 +36,18 @@ def testMatrix():
     assert_array_almost_equal(tr3.getRotation()[1], tr.getRotation()[1])
     assert_array_almost_equal(tr3.getScale(), tr.getScale())
     assert_array_almost_equal(tr3.getTranslation(), tr.getTranslation())
+
+
+def testMapTypes():
```
j9ac9k, line 41: this would probably be much better as a parametrized test, I can make that change if you'd like.

pyqtgraph/Transform3D.py
```diff
@@ -45,8 +47,11 @@ def map(self, obj):
             v = QtGui.QMatrix4x4.map(self, Vector(obj))
             return type(obj)([v.x(), v.y(), v.z()])[:len(obj)]
         else:
-            return QtGui.QMatrix4x4.map(self, obj)
-            
+            retval = QtGui.QMatrix4x4.map(self, obj)
+            if not isinstance(retval, type(obj)):
```
j9ac9k, line 51: is `isinstance` what's wanted here? wouldn't that capture subclasses when we don't want them to?

For example, if `type(obj)`was of type `pg.Point`, wouldn't this return a `QPointF`?

(haven't actually tested this, just speculating)

outofculture: It would fail in the other direction. If `.map` returns a _subclass_ of the `obj(type)`, then it would pass this test and not try to downgrade the object into the less specific `obj(type)`, which seems like the correct behavior. So if we passed in a `QPoint` and it spat out a `QPointF`, we would still return the `QPointF`.

While I'm parameterizing this test, I can add those types, too.

j9ac9k:
Thanks for the added tests @outofculture !

PR #3051, opened 2024-06-07
Remove MetaArray, MultiPlotItem and MultiPlotWidget

j9ac9k:
Also bumping up the version to 0.14.0dev0 since the deprecation warnings indicated `MetaArray` would be removed in 0.14.0.

If the `MetaArray` functionality is still needed, it can be gotten via the [MetaArray package](https://github.com/outofculture/metaarray).


j9ac9k:
the CodeQL counter at the time of merging was 305... curious what it will go down to.

PR #3052, opened 2024-06-07
Fix ViewBox wheel event propagation

tikuma-lsuhsc:
This PR makes `ViewBox` to ignore a wheel event if both `mouseEvent` flags are false. This enables graphic items to capture the wheel event when pyqtgraph's zoom/pan mechanism is turned off.

Below is a test script to demonstrate the PR. WIthout it, `MyPlotItem.wheelEvent` never gets called. (As far as I can tell, this event is the only one not caught by a custom PlotItem.)

```python
import numpy as np
from PyQt6.QtCore import Qt
import pyqtgraph as pg


app = pg.mkQApp("Plotting Example")

class MyPlotItem(pg.PlotItem):

    def wheelEvent(self, event):
        print("custom wheelEvent")


p1 = MyPlotItem()


class MyWindow(pg.GraphicsLayoutWidget):

    def keyPressEvent(self, ev):

        if ev.key() == Qt.Key.Key_Space:
            # toggle zoom/pan
            if any(p1.vb.mouseEnabled()):
                print("turn off zoom/pan")
                p1.vb.setMouseEnabled(False, False)
            else:
                print("turn on  zoom/pan")
                p1.vb.setMouseEnabled(True, True)

        return super().keyPressEvent(ev)


win = MyWindow(show=True)
p1.plot(y=np.random.normal(size=100))
win.addItem(p1)

if __name__ == "__main__":
    pg.exec()
```


j9ac9k:
Thanks @tikuma-lsuhsc this LGTM!  Merging.

Issue #2878, opened 2023-11-12
Curve fillLevel behaviour with enableExperimental


dmax366:
### Short description
Are fill levels for PlotDataItems supposed to be supported while useOpenGL and enableExperimental are True?
A comment in another issue (https://github.com/pyqtgraph/pyqtgraph/issues/2257) mentions that fill levels are not supported while enableExperimental is False, but I observe the opposite.
Also, why does fill level support depend on enableExperimental, but FillBetweenItem objects are able to display whether enableExperimental is true or false?

### Code to reproduce
The code below generates one curve with the fillLevel set and a FillBetweenItem using a different curve, so the two behaviours can be compared.
There is a button to toggle enableExperimental.

```python
import pyqtgraph as pg
from pyqtgraph.Qt import QtWidgets

import numpy as np

pg.setConfigOptions(useOpenGL = True, enableExperimental = False)

app = pg.mkQApp('OpenGL / Experimental Test')
win = pg.GraphicsLayoutWidget(show = True, title = 'OpenGL / Experimental Test')

plot = win.addPlot(title = 'Sinusoidal')
x_data = np.linspace(-10, 10, 1000)
curve = plot.plot(x_data, np.sin(x_data), pen = 'green', fillLevel = 0, fillBrush = pg.mkBrush((150, 0, 200, 100)), name = 'Sinusoidal')
y_data = np.sin(x_data*2.0)
curve2 = plot.plot(x_data, y_data, pen = 'red', name = 'Sinusoidal 2')

filler = pg.FillBetweenItem(curve, curve2, brush = pg.mkBrush((200, 200, 0, 100)), pen = 'pink')
plot.addItem(filler)

button = QtWidgets.QPushButton('enableExperimental = False')
en_exp = False
def toggle_experimental():
    global en_exp, button
    en_exp = not en_exp
    pg.setConfigOptions(enableExperimental = en_exp)
    button.setText(f'enableExperimental = {en_exp}')

button.clicked.connect(toggle_experimental)
proxy = QtWidgets.QGraphicsProxyWidget()
proxy.setWidget(button)

win.addItem(plot, 0, 0)
win.addItem(proxy, 1, 0)

win.show()
pg.exec()
```

### Expected behavior
The curve with fillLevel set should be filled while enableExperimental = True.

### Real behavior
The curve with fillLevel set is not filled while enableExperimental = True. Instead, it is filled while enableExperimental = False.

### Tested environment(s)

[...]

Issue #2417, opened 2022-09-15
Precision issues in opengl renderer when zoomed in with high values


gedalia:
### Short description
The gpu's ability to deal with interpolating numbers can be somewhat limited, it's not just the values you are sending down because the GPU needs to be able to interpolate between the points to draw the line which can require a bunch of additional precision.
The paintGL code winds up with 2 issues if your range is like 5000.01 to 5000.02.
*The stencil buffer code winds up subject to precision problems leading to the vertexes of the triangles to hop around, this leads to the graph not drawing inside the expected plot area. 
*The graph winds up quantized 

The first issue can be fixed by simplifying the transforms used to draw the stencil buffer to avoid ever leaving screen coordinates. 
Currently the mapRectToItem call transforms the screen coords to model coordinates. Those initial coords can just be used raw (with an offset for the left axis if the model view matrix is reset:
```
        rect = view.boundingRect()
        gl.glPushMatrix()
        gl.glLoadIdentity()

        gl.glEnable(gl.GL_STENCIL_TEST)
        gl.glColorMask(gl.GL_FALSE, gl.GL_FALSE, gl.GL_FALSE,
                       gl.GL_FALSE)  # disable drawing to frame buffer
        gl.glDepthMask(gl.GL_FALSE)  # disable drawing to depth buffer
        gl.glStencilFunc(gl.GL_NEVER, 1, 0xFF)
        gl.glStencilOp(gl.GL_REPLACE, gl.GL_KEEP, gl.GL_KEEP)

        ## draw stencil pattern
        gl.glStencilMask(0xFF)
        gl.glClear(gl.GL_STENCIL_BUFFER_BIT)

        margin = widget.width() - rect.width()
        gl.glBegin(gl.GL_TRIANGLES)
        gl.glVertex2f(rect.x() + margin, rect.y())
        gl.glVertex2f(rect.x() + rect.width() + margin, rect.y())
        gl.glVertex2f(rect.x() + margin, rect.y() + rect.height())
        gl.glVertex2f(rect.x() + rect.width() + margin, rect.y() + rect.height())
        gl.glVertex2f(rect.x() + rect.width() + margin, rect.y())
        gl.glVertex2f(rect.x() + margin, rect.y() + rect.height())
        gl.glEnd()

        gl.glColorMask(gl.GL_TRUE, gl.GL_TRUE, gl.GL_TRUE, gl.GL_TRUE)
        gl.glDepthMask(gl.GL_TRUE)
[...]

PR #3055, opened 2024-06-08
Port PlotCurveItem experimental opengl codepath to shaders

pijyoi:
Changes
1) no more dependency on PyOpenGL, uses QtOpenGL only
2) works on OpenGLES 2.0 (so far tested on Windows ANGLE)
3) trivially add support for connect="pairs" (assumed to contain finite data only)
4) add support for connect="all" and connect="finite"
5) add support for connect=ndarray (assumed to contain finite data only)
6) implement shadowPen
7) implement origin shift to mitigate quantization effects due to single precision


pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -15,6 +14,126 @@
 __all__ = ['PlotCurveItem']
 
 
+class GLC:
```
j9ac9k, line 17: This might be a good use-case for a [dataclass](https://docs.python.org/3/library/dataclasses.html) with `frozen=True`

pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -866,89 +988,90 @@ def paint(self, p, opt, widget):
...
-                    gl.glEnable(gl.GL_LINE_SMOOTH)
-                    gl.glEnable(gl.GL_BLEND)
-                    gl.glBlendFunc(gl.GL_SRC_ALPHA, gl.GL_ONE_MINUS_SRC_ALPHA)
-                    gl.glHint(gl.GL_LINE_SMOOTH_HINT, gl.GL_NICEST)
-                else:
-                    gl.glDisable(gl.GL_LINE_SMOOTH)
+            program = self.glstate.m_program
+            program.bind()
+            with warnings.catch_warnings():
+                # PySide2 : RuntimeWarning: SbkConverter: Unimplemented C++ array type.
```
j9ac9k, line 1028: Appreciate the comment specifying what warning we're looking to catch!

j9ac9k (approved):
Should there be other curves with the other connect mechanisms like `finite` that test with a `np.nan` value?

Love what you're doing with this!!!

pijyoi:
> Should there be other curves with the other connect mechanisms like `finite` that test with a `np.nan` value?

Implementing "pairs" already allows the user to plot arbitrary line segments and strips.
If we ignore details like join and miter, every additional mode could be reduced to `GL_LINES` aka `pairs`.

It shouldn't be difficult by taking reference from the following:
https://github.com/pyqtgraph/pyqtgraph/blob/3743e57523d9bbb8558638e9e6bc8f2270605cec/pyqtgraph/graphicsItems/PlotCurveItem.py#L18

Alternatively for "finite" mode, line strips could be extracted for multiple calls to `GL_LINE_STRIP`:
https://github.com/pyqtgraph/pyqtgraph/blob/3743e57523d9bbb8558638e9e6bc8f2270605cec/pyqtgraph/functions.py#L1883

_However_, I don't think it's a particularly useful thing to implement if the only purpose is to support more of the API of `PlotCurveItem`. `PlotCurveItem` already has too many extras baked in like histogram and fill that don't play well with non-finites.

j9ac9k:
> _However_, I don't think it's a particularly useful thing to implement if the only purpose is to support more of the API of `PlotCurveItem`. `PlotCurveItem` already has too many extras baked in like histogram and fill that don't play well with non-finites.

I'm in complete agreement here. The histogram and fill should either be removed from `PlotCurveItem` or we create another GraphicsItem (`PlotLineItem`?) that is a bit more primitive in the functionality, but that's a separate discussion.

Should the library emit warning (or raise an Exception) when using openGL with `PlotCurveItem` where if a user tries to do a histogram or a fill?

pijyoi:
A better solution is to only call `paintGL` for supported modes.
The script in #2878 now "works" whether `enableExperimental` is `True` or `False`.

NilsNemitz:
Hi @pijyoi , this looks great!

This *does* look like a good opportunity to rethink the architecture we now have with PlotCurveItem and PlotDataItem. It does not seem necessary to replicate everything that PlotCurveItem now does.

From my perspective it might be reasonable for a revised PlotCurveItem to take data
- that is known-good (e.g. no non-finites)
- does not require any modification (e.g. no step plots, no log mode)  

That would reasonably move the task of preprocessing for NaNs to PlotDataItem. That could include processing all data into 'pairs', which would indeed be a very generic solution that should work for almost everything.

One thing I wanted to add here is that the data I encounter might have very long "good" segments (all finite), then a short section where a failure generates NaN, followed by another "good" segment. If the fastest path is still the equivalent of 'finite', 'skip-tests' mode, then being able to plot multiple long segments of known-good data might be interesting. PlotDataItem would need to figure out what mode to use.

----

As an additional comment/question, OpenGL is probably very happy to render the filled shapes used with fillLevel or by the FillBetweenItem. Is there a particular arrangement of code that avoids transferring the same curve data to the GPU twice?

pijyoi:
>That would reasonably move the task of preprocessing for NaNs to PlotDataItem. That could include processing all data into 'pairs', which would indeed be a very generic solution that should work for almost everything.

>One thing I wanted to add here is that the data I encounter might have very long "good" segments (all finite), then a short section where a failure generates NaN, followed by another "good" segment. If the fastest path is still the equivalent of 'finite', 'skip-tests' mode, then being able to plot multiple long segments of known-good data might be interesting. PlotDataItem would need to figure out what mode to use.

I think a good middle ground would be a hypothetical `PlotLineItem` that accepts a (not too large) list of "good" segments.
This would satisfy the current `connect="finite"` mode as well as users wanting to plot a great many line segments using a single `GraphicsItem`.
The `connect="all"` mode would just be a list of one segment.
This would map to multiple calls to  `QPainter.drawPolyline` / `glDrawArrays(GL_LINE_STRIP)`

Although this would also support `connect="pairs"`, it would incur a large overhead. (Nothing would stop a user from doing it, however) So "pairs" and "ndarray" mode could instead map to `QPainter.drawLines` / `glDrawArrays(GL_LINES)`.

>As an additional comment/question, OpenGL is probably very happy to render the filled shapes used with fillLevel or by the FillBetweenItem. Is there a particular arrangement of code that avoids transferring the same curve data to the GPU twice?

The filling API of `pyqtgraph` does not require the shape to be closed. Take for example the filling of the area under a sine curve. Even for the current pyqtgraph `QGraphicsView` implementation, this requires two separate calls:
1) `QPainter.fillPath` to flood fill the "enclosed" area (pyqtgraph library does the closing)
2) `QPainter.drawPath` to draw the sine curve

For OpenGL (caveat, my knowledge of OpenGL is limited), filling the area would require generating the vertices for the triangle strip that covers the area to be filled. So it would in any case be two different sets of vertices to be sent to the GPU. One for the curve to be drawn using `GL_LINE_STRIP` and the other to be drawn using `GL_TRIANGLE_STRIP`.

j9ac9k:
I think implementing a theoretical PlotLineItem so we can be aggressive about API breakage is a great idea.  Also gives us the opportunity to use `QPainter.drawPolyline` which the benchmark seems to indicate it works better.

The one question I have is, would it make sense for any of the downsampling stuff to occur at that level, or should that remain in the PlotDataItem/higher level objects to implement?

NilsNemitz:
@pijyoi , thank you for your reply!

Your suggestions look very reasonable to me, I have no further comments. I am certain that if you encounter a way to streamline the OpenGL data transfer, you'll be the very first to jump on it. :)


@j9ac9k, 

I think the downsampling would want to be placed at the same level where transforms like log mode are handled.
- for the heavier transforms, it makes sense to have the option to downsample first, THEN transform.
- either of these actions has to inspect the dataset and generates insights into e.g. the existance of non-finites if it runs. This should help choose the best drawing mode, for example. (See PlotDataItem's 'connect=auto' mode.)  Having this information in one place may make things a little more elegant than having cascaded checks through the different levels.

j9ac9k:
> I think the downsampling would want to be placed at the same level where transforms like log mode are handled.
> 
>     * for the heavier transforms, it makes sense to have the option to downsample first, THEN transform.
> 
>     * either of these actions has to inspect the dataset and generates insights into e.g. the existance of non-finites if it runs. This should help choose the best drawing mode, for example. (See PlotDataItem's 'connect=auto' mode.)  Having this information in one place may make things a little more elegant than having cascaded checks through the different levels.

I agree that the transforms and downsamples should stick together.  In what order that happens, I think there is an argument to be made on how non-linear the transform is; but that's a discussion for another PR.

The next concern I have right now is involving "thick" lines.  I don't think this needs to be addressed in this specific PR, but I'd like to at least bring it up.  The issue right now is the use of `glLineWidth(width_values)`, The supported `glLineWidth` range of values to my understanding is limited by the OpenGL driver, and in some drivers it is limited to just 1 pixel, and even if the desired thickness is within the reported range, there are issues with it not working.  There is a small github repo that points this out and proposes some alternatives (of various complexities).

https://github.com/mhalber/Lines?tab=readme-ov-file#why

Looking at some of the alternatives listed, the geometry shader seems like a good place to inject some functionality here, but that requires OpenGL 3.2.  I'd be curious to hear anyone else's thoughts.

pijyoi:
>Looking at some of the alternatives listed, the geometry shader seems like a good place to inject some functionality here, but that requires OpenGL 3.2. I'd be curious to hear anyone else's thoughts.

Actually it uses a number of OpenGL 4.5 API (https://github.com/mhalber/Lines/blob/f20b5e2509d98e5d8e273d9730036ec5aecaa02d/geometry_shader_lines.h#L203-L214)

j9ac9k:
> > Looking at some of the alternatives listed, the geometry shader seems like a good place to inject some functionality here, but that requires OpenGL 3.2. I'd be curious to hear anyone else's thoughts.
> 
> Actually it uses a number of OpenGL 4.5 API (https://github.com/mhalber/Lines/blob/f20b5e2509d98e5d8e273d9730036ec5aecaa02d/geometry_shader_lines.h#L203-L214)

Oh yeah, OpenGL 4.5 API is a non-starter, anything above 4.1 is a non-started <shakes fist at apple>, alternative being QRhi, but that's a whole other thing.

pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -1,20 +1,136 @@
...
+
+    def setup(self, widget):
+        context = widget.context()
+        if self.widget is widget and self.context is context:
+            return self.functions
+
+        # if self.widget is not None and self.widget is not widget:
+        #     print(f'widget changed {hex(id(self.widget))} to {hex(id(widget))}')
+        # if self.context is not None and self.context is not context:
+        #     print(f'context changed {hex(id(self.context))} to {hex(id(context))}')
```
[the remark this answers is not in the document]
j9ac9k: should we remove this commented out code?

pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -801,10 +918,25 @@ def paint(self, p, opt, widget):
...
 
-        if getConfigOption('enableExperimental'):
-            if HAVE_OPENGL and isinstance(widget, QtWidgets.QOpenGLWidget):
-                self.paintGL(p, opt, widget)
-                return
+        do_fill = self.opts['brush'] is not None and self.opts['fillLevel'] is not None
+        do_fill_outline = do_fill and self.opts['fillOutline']
+
+        if (
+            getConfigOption('enableExperimental')
```
j9ac9k, line 925: do you think we should hide this functionality being the `enableExperimental` config option at this point? The downside of doing this is that it won't get as much usage and we may not get any feedback on its functionality.

pijyoi: #2257 shows that there's 1 user using OpenGL while simultaneously leaving `enableExperimental=False`. So at this point, enabling `paintGL` codepath w/o `enableExperimental=True` may constitute an API breakage for them.
@bbc131, would you care to chime in on this?

pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -1,20 +1,136 @@
...
 from .. import Qt, debug
 from .. import functions as fn
 from .. import getConfigOption
+from ..Qt import OpenGLConstants as GLC
 from .GraphicsObject import GraphicsObject
 
 __all__ = ['PlotCurveItem']
 
 
+class OpenGLState:
```
j9ac9k, line 18: This _may_ need to inherit from `QObject` due to the `cleanup` method being a `Slot`.

In some non-pyqtgraph Qt related work, I can into issues when connecting to a slot that wasn't decorated and wasn't decorated with the `@Slot()` decorator.

https://www.riverbankcomputing.com/static/Docs/PyQt6/signals_slots.html#the-pyqtslot-decorator

> Connecting a signal to a decorated Python method also has the advantage of reducing the amount of memory used and is slightly faster.

I know pyqtgraph doesn't decorate its slots but I anticipate this being a change that we'll roll out in the near future.

pijyoi: Was your issue because of being called from another thread?
In this case here, cleanup has to be a direct call and thus doesn't need to be registered as a QObject metamethod.

campagnola:
@pijyoi, thank you so much for the fantastic work on pyqtgraph performance. This is an excellent addition like so many that came before.

> From my perspective it might be reasonable for a revised PlotCurveItem to take data
> 
>     * that is known-good (e.g. no non-finites) 
>     * does not require any modification (e.g. no step plots, no log mode)

I am all for rethinking the API and separating concerns, especially where it leads to better performance. However I don't think this PR is the right place to start that work, mostly because (you can hear my experience with vispy here) OpenGL is really weird in some ways, and it tends to guide one toward strange architectural contortions that otherwise make life more difficult. 

As an example of that point, there's an even more interesting performance target which is to write GPU shaders that perform all of the data processing. Ideally, we should send only the raw data as provided by the user to the GPU and the rest happens there--filtering for finite values, transformations (both linear and nonlinear), construction of new geometry like triangle strips or histogram steps, etc. In that case it may not be necessary to have a simplifed PlotLineItem if, in the end, PlotCurveItem is able to handle all types of data efficiently. 

I think I prefer a somewhat different approach (that we have discussed before), using a PlotData class that encapsulates raw x,y data plus a set of filters / transforms and then allows the user of that data (PlotDataItem, PlotCurveItem, etc) to specify the result that is needed. For example, if PlotCurveItem wants strips of nan-free data, then PlotDataItem can provide that in the most efficient way. If there is a ScatterPlotItem that uses the same PlotData instance, then when it asks for nan-free values, the PlotData instance can return previously cached results. If you wanted to implement all or part of this in GPU, then you could simply request PlotData give you raw or partially-processed data, plus the extra metadata needed to complete the processing in a GPU shader. 

> The issue right now is the use of glLineWidth(width_values), The supported glLineWidth range of values to my 
> understanding is limited by the OpenGL driver, and in some drivers it is limited to just 1 pixel,

This is indeed a major issue, and one that I don't recommend taking on lightly. For now, just having the high performance of 1px-width lines covers the vast majority of high performance use cases. If you ever want to go further than that, then let's have a long talk about the architecture of vispy and how similar concepts could be implemented here.

pijyoi:
I have implemented the origin shifting suggestion in #2417.
This does fix the obvious issue demonstrated in https://github.com/pyqtgraph/pyqtgraph/issues/2020#issuecomment-945135581

This necessitates extra computation in that the data has to be converted to double precision first to perform the translation before finally converting it back to single precision for OpenGL vertex buffers.
But correctness over speed, right?

j9ac9k:
Correctness over speed for sure 👍

pijyoi:
Modified the MouseSelection example to demonstrate the quantization effect if origin shift not applied:
Screenshot 2024-06-18 235904

pyqtgraph/graphicsItems/PlotCurveItem.py
bbc131: > #2257 shows that there's 1 user using OpenGL while simultaneously leaving `enableExperimental=False`. So at this point, enabling `paintGL` codepath w/o `enableExperimental=True` may constitute an API breakage for them. @bbc131, would you care to chime in on this?

We don't use this configuration (OpenGL and `enableExperimental=False`) anymore, so this won't break anything for us.

j9ac9k:
Thanks for this PR @pijyoi 

Getting a more robust OpenGL implementation throughout the library would be such an amazing improvement. Also having some implementations done makes it easier for contributors to attempt to add OpenGL support for other parts too.

I hope we can revisit the `enableExperimental` setting that this is being kept behind, but that's a decision for a future PR.

pijyoi:
non-working: if `QPen` is created with a non-solid brush.
https://github.com/pyqtgraph/pyqtgraph/pull/3068#issuecomment-2184311958

PR #3057, opened 2024-06-08
Documentation text review for PlotDataItem

NilsNemitz:
After the recent formatting update to the PlotDataItem documentation in PR #3039, I have gone through the text itself.

I would like to propose some changes for consistency and readability. Very likely not all changes are good, but I hope some of the better ones can be picked out and moved into the documentation.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -467,7 +467,7 @@ class PlotDataItem(GraphicsObject):
 
         skipFiniteCheck     ``bool``, default ``False``
 
-                            If `True`, the special handling of non-finite values such as
+                            If ``True``, the special handling of non-finite values such as
                             ``NaN`` in :class:`~pyqtgraph.PlotCurveItem` is skipped.
                             This speeds up the plot, but creates error or causes the
                             plotting to fail entirely if any such values are present.
@@ -690,7 +690,6 @@ class PlotDataItem(GraphicsObject):
         ----------
         alpha : float
             Value passed to :meth:`QGraphicsItem.setOpacity`.
-
         auto : bool
             Receives the ``autoAlpha`` value from a parent
             :class:`~pyqtgraph.PlotItem`, but has no function within PlotDataItem
@@ -704,7 +703,7 @@ class PlotDataItem(GraphicsObject):
         if self.opts['alphaHint'] == alpha and self.opts['alphaMode'] == auto:
             return
         self.opts['alphaHint'] = alpha
-        self.opts['alphaMode'] = auto # 'alphaMode' option is never used or transferred, should be deprecated.
+        self.opts['alphaMode'] = auto
         self.setOpacity(alpha)
 
     def setFftMode(self, state: bool):
@@ -1006,7 +1005,7 @@ class PlotDataItem(GraphicsObject):
 
         Parameters
         ----------
-        size : int
+        size : int | list[int]
             Diameter of the symbols, or array-like list of diameters. Diameter is
             either in pixels or data-space coordinates depending on the value of
             `pxMode`.
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -61,7 +61,7 @@ class PlotDataset:
     ``dataset.y``.
     
     After a search has been performed, typically during a call to
-    :func:`dataRect() <pyqtgraph.PlotDataset.dataRect>`, ``dataset.containsNonfinite``
+    :meth:`dataRect <pyqtgraph.PlotDataset.dataRect>`, ``dataset.containsNonfinite``
     is ``True`` if any coordinate values are non-finite (e.g. ``NaN`` or ``Inf``) or 
     ``False`` if all values are finite. If no search has been performed yet,
     `dataset.containsNonfinite` is ``None``.
@@ -251,7 +251,7 @@ class PlotDataItem(GraphicsObject):
     * If you know that all points in your data set will have numerical, finite values,
       :meth:`setSkipFiniteCheck` can disable a check to identify points that require
       special treatment.  
-    * When passing `x` and `y` data to :func:`PlotDataItem.setData`, use
+    * When passing `x` and `y` data to :meth:`PlotDataItem.setData`, use
       :class:`numpy.ndarray` instead of python's built-in lists.
 
     **Bases:** :class:`~pyqtgraph.GraphicsObject`
@@ -290,7 +290,7 @@ class PlotDataItem(GraphicsObject):
         The supported keyword arguments can be grouped into several categories:
 
         *Point Style Keyword Arguments*, see
-        :func:`ScatterPlotItem.setData <pyqtgraph.ScatterPlotItem.setData>` for more
+        :meth:`ScatterPlotItem.setData <pyqtgraph.ScatterPlotItem.setData>` for more
         information.
     
         =========== ====================================================================
@@ -847,7 +847,7 @@ class PlotDataItem(GraphicsObject):
         """
         Set the :class:`QBrush` used to fill the area under the curve.
          
-        Use :func:`~setFillLevel` to enable filling and set the boundary value. 
+        Use :meth:`setFillLevel` to enable filling and set the boundary value. 
 
         Parameters
         ----------
@@ -876,7 +876,7 @@ class PlotDataItem(GraphicsObject):
 
     def setBrush(self, *args, **kwargs):
         """
-        An alias to :func:`~pyqtgraph.PlotDataItem.setFillBrush`.
+        An alias to :meth:`setFillBrush`.
 
         Parameters
         ----------
@@ -922,7 +922,7 @@ class PlotDataItem(GraphicsObject):
         """
         Set the symbol or symbols for drawing the points.
 
-        See :func:`pyqtgraph.ScatterPlotItem.setSymbol` for a full list of accepted
+        See :meth:`pyqtgraph.ScatterPlotItem.setSymbol` for a full list of accepted
         arguments.
 
         Parameters
@@ -934,7 +934,7 @@ class PlotDataItem(GraphicsObject):
         
         See Also
         --------
-        :func:`~pyqtgraph.ScatterPlotItem.setSymbol`
+        :meth:`pyqtgraph.ScatterPlotItem.setSymbol`
             Recognized symbols are detailed in the description of this method.
         """
         if self.opts['symbol'] == symbol:
@@ -1165,7 +1165,7 @@ class PlotDataItem(GraphicsObject):
         See Also
         --------
         :class:`PlotDataItem`
-            The arguments accepted by :func:`setData` are the same used during 
+            The arguments accepted by :meth:`setData` are the same used during 
             initialization, and are listed in the opening section.
         :func:`~pyqtgraph.arrayToQPath`
             Explains the constructions of the draw paths.
@@ -1422,7 +1422,7 @@ class PlotDataItem(GraphicsObject):
 
         See Also
         --------
-        :func:`getData`
+        :meth:`getData`
             This method returns the transformed data displayed on the screen instead.
         """
         dataset = self._dataset
@@ -1643,7 +1643,7 @@ class PlotDataItem(GraphicsObject):
 
         See Also
         --------
-        :func:`getOriginalDataset`
+        :meth:`getOriginalDataset`
             This method returns the original data provided to PlotDataItem instead.
         """
         dataset = self._getDisplayDataset()
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -62,8 +62,8 @@ class PlotDataset:
     
     After a search has been performed, typically during a call to
     :func:`dataRect() <pyqtgraph.PlotDataset.dataRect>`, ``dataset.containsNonfinite``
-    is ``True`` if any coordinate values are non-finite (e.g. ``NaN`` or ``Inf``) or ``False``
-    if all values are finite. If no search has been performed yet,
+    is ``True`` if any coordinate values are non-finite (e.g. ``NaN`` or ``Inf``) or 
+    ``False`` if all values are finite. If no search has been performed yet,
     `dataset.containsNonfinite` is ``None``.
 
     Parameters
@@ -206,9 +206,11 @@ class PlotDataset:
 
 class PlotDataItem(GraphicsObject):
     """
-    PlotDataItem is PyQtGraph's primary way to plot 2D data. It provides a unified 
-    interface for displaying plot curves, scatter plots, or both. The library's
-    convenience functions such as :func:`pyqtgraph.plot` create PlotDataItem objects.
+    PlotDataItem is PyQtGraph's primary way to plot 2D data.
+    
+    It provides a unified interface for displaying plot curves, scatter plots, or both.
+    The library's convenience functions such as :func:`pyqtgraph.plot` create
+    PlotDataItem objects.
 
     .. code-block::
 
@@ -218,9 +220,9 @@ class PlotDataItem(GraphicsObject):
 
     The scatter plot renders a symbol for each point of the data. The curve connects
     adjacent points. This is realized by a combination of a
-    :class:`~pyqtgraph.ScatterPlotItem` and a :class:`~pyqtgraph.PlotCurveItem`. Although
-    these classes can be used individually, :class:`~pyqtgraph.PlotDataItem` is the
-    recommended way to interact with them.
+    :class:`~pyqtgraph.ScatterPlotItem` and a :class:`~pyqtgraph.PlotCurveItem`.
+    Although these classes can be used individually, :class:`~pyqtgraph.PlotDataItem`
+    is the recommended way to interact with them.
 
     PlotDataItem contains methods to transform the original data:
       
@@ -239,18 +241,18 @@ class PlotDataItem(GraphicsObject):
     following.
 
     * Use a :class:`QPen` with ``width=1``. All wider pens cause a loss in performance.
-      This loss can be partially mitigated by using fully opaque colors (``alphaF=1.0``),
-      solid lines, and no anti-aliasing. 
-    * For scatter plots that use multiple pen or brush settings, passing a list of string
-      representation to `symbolPen` or `symbolBrush` creates many internal :class:`QPen`
-      and :class:`QBrush` objects. Instead, create the needed :class:`QPen` and
-      :class:`QBrush` objects in advance and pass these as a list instead. This lets a 
-      smaller number of stored instances be reused.
+      This loss can be partially mitigated by using fully opaque colors 
+      (``alphaF=1.0``), solid lines, and no anti-aliasing. 
+    * For scatter plots that use multiple pen or brush settings, passing a list of
+      string representation to `symbolPen` or `symbolBrush` creates many internal
+      :class:`QPen` and :class:`QBrush` objects. Instead, create the needed
+      :class:`QPen` and :class:`QBrush` objects in advance and pass these as a list
+      instead. This lets a smaller number of stored instances be reused.
     * If you know that all points in your data set will have numerical, finite values,
       :meth:`setSkipFiniteCheck` can disable a check to identify points that require
       special treatment.  
     * When passing `x` and `y` data to :func:`PlotDataItem.setData`, use
-      :class:`numpy.ndarray` s instead of python's built-in lists.
+      :class:`numpy.ndarray` instead of python's built-in lists.
 
     **Bases:** :class:`~pyqtgraph.GraphicsObject`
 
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -62,7 +62,7 @@ class PlotDataset:
     
     After a search has been performed, typically during a call to
     :func:`dataRect() <pyqtgraph.PlotDataset.dataRect>`, ``dataset.containsNonfinite``
-    is ``True`` if any coordinate values are non-finite (e.g. NaN or inf) or ``False``
+    is ``True`` if any coordinate values are non-finite (e.g. ``NaN`` or ``Inf``) or ``False``
```
j9ac9k, line 65: I keep going back and forth on this, but should we replace usages of `NaN` with `np.nan` and `np.inf`  for `Inf`?

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -206,102 +206,127 @@ def applyLogMapping(self, logMode: tuple[bool, bool]):
 
 class PlotDataItem(GraphicsObject):
     """
-    Provides a unified interface for displaying plot curves, scatter plots, or both.
-
-    The symbols of a scatter point are rendered over the "point" of the data, and the 
-    curve connects adjacent points.
+    PlotDataItem is PyQtGraph's primary way to plot 2D data. It provides a unified
```
j9ac9k, line 209: First line in the docstring needs to be confined to 1-line per numpydoc:

https://numpydoc.readthedocs.io/en/latest/format.html#short-summary

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -804,21 +840,21 @@ def setShadowPen(self, *args, **kwargs):
...
+            Arguments directed to :func:`mkBrush <pyqtgraph.mkBrush>`.
         
         See Also
         --------
-        :func:`~pyqtgraph.mkBrush`
-            See for supported arguments.
-        :func:`~pyqtgraph.mkColor`
-            See for supported color arguments.
+        :func:`mkBrush <pyqtgraph.mkBrush>`
+            for supported arguments.
```
j9ac9k, line 855: Needs to be capitalized.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -804,21 +840,21 @@ def setShadowPen(self, *args, **kwargs):
...
+        :func:`mkColor <pyqtgraph.mkColor>`
+            for supported color arguments.
```
j9ac9k, line 857: Needs to be capitalized.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -860,7 +896,7 @@ def setFillLevel(self, level: float | None):
         See Also
         --------
         :class:`pyqtgraph.FillBetweenItem`
-            See for another :class:`~pyqtgraph.GraphicsItem` that fills in regions.
+            for another :class:`~pyqtgraph.GraphicsItem` that fills in regions.
```
j9ac9k, line 899: Needs to be capitalized.

j9ac9k:
I'll review the content in a bit, but numpydoc has some ...opinions.

Link to the style guide, but I know it's a lot of text so it's tough to recognize all the issues.

https://numpydoc.readthedocs.io/en/latest/format.html#sections

Main ones when I ran numpydoc (I should note, that for numpydoc to run, I'm having to run my own fork, as they haven't merged my PR).

Some samples of the errors/issues it returns:

Descriptions should be capitalized:

```
| ../pyqtgraph/graphicsItems/PlotDataItem.py:933  | PlotDataItem.PlotDataItem.setSymbolPen         | SA03    | Description should be capitalized for See Also     |
|                                                 |                                                |         | "mkPen" reference
```

First line is a "short summary" which should fit on one line:

```
| ../pyqtgraph/graphicsItems/PlotDataItem.py:958  | PlotDataItem.PlotDataItem.setSymbolBrush       | SS06    | Summary should fit in a single line                |
```

Here's the whole battery of "errors" it returned:

```
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| file                                            | item                                           | check   | description                                        |
+=================================================+================================================+=========+====================================================+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:207  | PlotDataItem.PlotDataItem                      | GL07    | Sections are in the wrong order. Correct order is: |
|                                                 |                                                |         | Parameters, Attributes, See Also, Notes            |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:207  | PlotDataItem.PlotDataItem                      | SS06    | Summary should fit in a single line                |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:207  | PlotDataItem.PlotDataItem                      | PR09    | Parameter "**kwargs" description should finish     |
|                                                 |                                                |         | with "."                                           |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:841  | PlotDataItem.PlotDataItem.setFillBrush         | SA03    | Description should be capitalized for See Also     |
|                                                 |                                                |         | "mkBrush" reference                                |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:841  | PlotDataItem.PlotDataItem.setFillBrush         | SA03    | Description should be capitalized for See Also     |
|                                                 |                                                |         | "mkColor" reference                                |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:886  | PlotDataItem.PlotDataItem.setFillLevel         | SA03    | Description should be capitalized for See Also     |
|                                                 |                                                |         | "pyqtgraph.FillBetweenItem" reference              |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:933  | PlotDataItem.PlotDataItem.setSymbolPen         | SS06    | Summary should fit in a single line                |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:933  | PlotDataItem.PlotDataItem.setSymbolPen         | SA03    | Description should be capitalized for See Also     |
|                                                 |                                                |         | "mkPen" reference                                  |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:933  | PlotDataItem.PlotDataItem.setSymbolPen         | SA03    | Description should be capitalized for See Also     |
|                                                 |                                                |         | "mkColor" reference                                |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:958  | PlotDataItem.PlotDataItem.setSymbolBrush       | SS06    | Summary should fit in a single line                |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:958  | PlotDataItem.PlotDataItem.setSymbolBrush       | SA03    | Description should be capitalized for See Also     |
|                                                 |                                                |         | "mkBrush" reference                                |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:958  | PlotDataItem.PlotDataItem.setSymbolBrush       | SA03    | Description should be capitalized for See Also     |
|                                                 |                                                |         | "mkColor" reference                                |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:1063 | PlotDataItem.PlotDataItem.setDynamicRangeLimit | GL03    | Double line break found; please use only one blank |
|                                                 |                                                |         | line to separate sections or paragraphs, and do    |
|                                                 |                                                |         | not leave blank lines at the end of docstrings     |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:1127 | PlotDataItem.PlotDataItem.setData              | SA03    | Description should be capitalized for See Also     |
|                                                 |                                                |         | "PlotDataItem" reference                           |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:1127 | PlotDataItem.PlotDataItem.setData              | SA03    | Description should be capitalized for See Also     |
|                                                 |                                                |         | "~pyqtgraph.arrayToQPath" reference                |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:1295 | PlotDataItem.PlotDataItem.updateItems          | SS03    | Summary does not end with a period                 |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:1295 | PlotDataItem.PlotDataItem.updateItems          | SS05    | Summary must start with infinitive verb, not third |
|                                                 |                                                |         | person (e.g. use "Generate" instead of             |
|                                                 |                                                |         | "Generates")                                       |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:1295 | PlotDataItem.PlotDataItem.updateItems          | SS06    | Summary should fit in a single line                |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:1390 | PlotDataItem.PlotDataItem.getOriginalDataset   | RT04    | Return value description should start with a       |
|                                                 |                                                |         | capital letter                                     |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:1390 | PlotDataItem.PlotDataItem.getOriginalDataset   | RT04    | Return value description should start with a       |
|                                                 |                                                |         | capital letter                                     |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:1390 | PlotDataItem.PlotDataItem.getOriginalDataset   | SA03    | Description should be capitalized for See Also     |
|                                                 |                                                |         | "getData" reference                                |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:1611 | PlotDataItem.PlotDataItem.getData              | GL02    | Closing quotes should be placed in the line after  |
|                                                 |                                                |         | the last text in the docstring (do not close the   |
|                                                 |                                                |         | quotes in the same line as the text, or leave a    |
|                                                 |                                                |         | blank line between the last text and the quotes)   |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:1611 | PlotDataItem.PlotDataItem.getData              | GL03    | Double line break found; please use only one blank |
|                                                 |                                                |         | line to separate sections or paragraphs, and do    |
|                                                 |                                                |         | not leave blank lines at the end of docstrings     |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:1611 | PlotDataItem.PlotDataItem.getData              | SA03    | Description should be capitalized for See Also     |
|                                                 |                                                |         | "getOriginalDataset" reference                     |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
| ../pyqtgraph/graphicsItems/PlotDataItem.py:1695 | PlotDataItem.PlotDataItem.pixelPadding         | SS06    | Summary should fit in a single line                |
+-------------------------------------------------+------------------------------------------------+---------+----------------------------------------------------+
```

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -645,15 +686,16 @@ def setAlpha(self, alpha: float, auto: bool):
         """
         Set the opacity of the item to the value passed in.
 
-        This method is likely
-
         Parameters
         ----------
         alpha : float
             Value passed to :meth:`QGraphicsItem.setOpacity`.
+
```
j9ac9k, line 693: I don't think we want a newline here?

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -860,7 +907,9 @@ def setFillLevel(self, level: float | None):
         See Also
         --------
         :class:`pyqtgraph.FillBetweenItem`
-            See for another :class:`~pyqtgraph.GraphicsItem` that fills in regions.
+            This
+            :class:`~pyqtgraph.GraphicsItem` creates a filled in region between two
```
j9ac9k, line 911: probably doesn't need to be a new line here?

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -945,13 +1002,14 @@ def setSymbolBrush(self, *args, **kwargs):
 
     def setSymbolSize(self, size: int):
```
j9ac9k, line 1003: oh, this should be updated to be `size: int | list[int]`  ...not your change, but a mistake I made.

pyqtgraph/graphicsItems/PlotDataItem.py
NilsNemitz: The argument for ``NaN`` is that
- it is more generic
- data input is not necessarily in numpy format.

When *we* generate a NaN, it should probably be `np.nan`, but I think that does not occur here as far as the documentaiton is concerned.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -662,7 +704,7 @@ def setAlpha(self, alpha: float, auto: bool):
         if self.opts['alphaHint'] == alpha and self.opts['alphaMode'] == auto:
             return
         self.opts['alphaHint'] = alpha
-        self.opts['alphaMode'] = auto
+        self.opts['alphaMode'] = auto # 'alphaMode' option is never used or transferred, should be deprecated.
```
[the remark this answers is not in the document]
NilsNemitz: Thank you, I wrote that first and then forgot to delete it.
My impression is that `PlotItem` is using this to leave itself notes.
Maybe the real issue here is if we consider this a public function or if it is ONLY intended to be called internally.

pyqtgraph/graphicsItems/PlotDataItem.py
NilsNemitz: There is some air of uncertainty on whether list includes np.ndarray, or if and when that should be array_like.

I have no suggested answer to that question.

NilsNemitz:
The linter comments are very specific and really helpful in targetting the correct style.

There is just one issue I wanted to bring up for discussion: In the "See Also" section, the linter demands a full sentence of explanation. Here is an example:
```
See Also
--------
    :func:`mkBrush <pyqtgraph.mkBrush>`
        For supported arguments see the convenience function constructing QBrush instances.
    :func:`mkColor <pyqtgraph.mkColor>`
        For supported color arguments see the convenience function constructing QColor instances.
```

These sentences often seem a little awkward and difficult to connect to the linked target. My impression is that it is easier, shorter, and at least equally clear to write in a way that joins the section title, linked target and comment into one sentence:
```
See Also
--------
    :func:`~pyqtgraph.mkBrush`
        for supported arguments.
    :func:`~pyqtgraph.mkColor`
        for supported color arguments.
``` 
that is "See also \ mkBrush \ for supported arguments."

I do like that the linter encourages a consistent style here, which would be harder if we disable the rule that encourages capitalization, i.e. a full sentence after the reference.

Do we have any additional opinions on this?

j9ac9k:
> I do like that the linter encourages a consistent style here, which would be harder if we disable the rule that encourages capitalization, i.e. a full sentence after the reference.
> 
> Do we have any additional opinions on this?

Not all the linter rules make sense for us, I have no problem disabling it (and think in this case it makes sense to do so).

An alternative is we do not use the `See Also` section as liberally, and have our cross linking to `mkBrush` (and friends) be sufficient.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -206,126 +206,161 @@ def applyLogMapping(self, logMode: tuple[bool, bool]):
...
                     ``'auto'``
                     
-                    - ``'auto'`` - if dataset contains non-finite values, switch to
-                      ``'all'``, otherwise change to ``'finite'``.
-                    - ``'all'`` - connects all points.  
-                    - ``'pairs'`` - generates lines between every other point.
-                    - ``'finite'`` - creates a break when a non-finite points is
-                      encountered. 
-                    - :class:`~numpy.ndarray` - it should contain `N` elements that of 
+                    - ``'auto'`` connects all points, but creates a break where a
```
pijyoi, line 352: _"connects all points, but creates a break..."_ sounds contradictory.
Looking at the implementation, "auto" semantically means "finite".

In the absence of non-finites, "all" is equivalent to "finite".
Since "auto" only uses "all" when there are no non-finites, it is equivalent to using "finite".

Under the hood, the combination of "all" + `skipFiniteCheck=True` goes through a different processing path,  but that is an implementation detail.

j9ac9k: I know this isn't that relevant to this PR, but the `skipFiniteCheck` was intended as a band-aid, which may no longer be relevant, in which case I would be fine with removing.

NilsNemitz: Thank you for the comments! I'll reconsider the wording to avoid the whiplash going from "connect all" to "creates breaks".

My understanding is that the mode we want to encourage is the one currently called 'auto'. It is the same as 'finite', but if e.g. log mode calculations have *already* given us the info that there are no non-finites, then the test is skipped for free.

Two alternative proposals to make this nicer and simpler:
(A) If `skipFiniteCheck` is indeed not needed anymore (is it?), then we should simply stop advertising 'auto' mode and eventually deprecate it.
(B) If the 'auto' behavior is our best default solution, then we should rename 'auto' to 'finite'. This would restore parity between the 'connect' options in PlotCurveItem and PlotDataItem. As you correctly point out, there is no functional difference on the user side. ('auto' would remain as a hidden alias for compatibility.)

pijyoi: There are 3 possible modes here:
1) connect="all" + skipFiniteCheck=True
    * purpose: data is known to be a single segment with all data finite
2) connect="all" + skipFiniteCheck=False
    * purpose: data may have non-finites, but user wants to draw it as a single segment anyway (by effectively eliminating the non-finite values)
3) connect="finite" + skipFiniteCheck=False ("finite" implies that checking cannot be skipped)
    * purpose: draw runs of finite values as separate line segments

It can be observed that (1) is a special case of (3). (ignoring side-effects like whether one runs more efficiently than the other)

In the current implementation of `PlotDataItem`, `connect="auto"` will result in either (1) or (3) only. Hence, rendering-wise, `connect="auto"` is equivalent to `"finite"`.

One possible reason for the existence of (2) is that things like fill and histogram do not behave well in the presence of non-finites. i.e. (2) exists to sanitize the data for code that can't (and/or won't be updated to) handle non-finites.

pijyoi: I don't understand your second bullet point about respect-(ing) `skipFiniteCheck`. Did you mean that if no known information, then use the user supplied value of `skipFiniteCheck`?

Currently, as implemented and documented, "auto" means user supplied `skipFiniteCheck` is ignored. This makes sense and is the current value-add of "auto" over "finite".
Perhaps `skipFiniteCheck` in `PlotDataItem` should be removed? (while retaining it for `PlotCurveItem`)
It can retroactively be defined that `PlotDataItem`'s `skipFiniteCheck` argument was only ever a hint that may or may not be respected.

In fact, we could follow `ImageItem`'s model where instead of just indicating whether non-finites are present, we provide (and cache) the locations (indices) of the non-finites to `PlotCurveItem`. An empty locations would indicate that all data is finite.
In such a model, `PlotDataItem` doesn't get to say that it doesn't know if the data has any non-finites. It has to find it. After all, any time saved in `PlotDataItem` just gets kicked down the line to `PlotCurveItem`.

NilsNemitz: You are right, there is clearly an alternative architecture that would make this a lot more logical and accessible.

Having PlotDataItem always check for and locate any non-finite values would fit well with my current understanding of how processing should be split between PlotDataItem and its curve and scatter children.

A specific proposal for that will need its own PR though. For this PR, I have tried to clean up the description for the *current* state of implementation a little, but that is where I would like to leave it for now.

PR #3058, opened 2024-06-09
Do upgrade to CONTRIBUTING.md, it's been a while

j9ac9k:


j9ac9k:
I'm on the fence for `CONTRIBUTING.md` adding instructions to use my fork of numpydoc with the fix for accepting cross-reference syntax.  There is a PR for numpydoc to accept that, but even if merged, it's unclear when they will have their next release.

NilsNemitz:
I would think that the fork is already suitable for the very limited use proposed here.
When a patched version of the main numpydoc becomes available, it is a two-line change here to point to that instead.

j9ac9k:
> I would think that the fork is already suitable for the very limited use proposed here. When a patched version of the main numpydoc becomes available, it is a two-line change here to point to that instead.

Added instructions to install my fork.  Updated other parts of the doc as well... going to merge.  Thanks for the review @NilsNemitz

Issue #2684, opened 2023-04-11
'Flowcharts' example: metaArray exception and plots not positioned correctly


ghylander:
### Short description
In the 'Flowcharts' example there are two bugs. First, the use of `pyqtgraph.metaarray` results in an exception that causes the input signal to not be plotted.

Installing the MetaArray library, and changing lines 16 and 52 to use said library fixes the issue.

Second, the output of the gaussian filter should be plotted in the bottom plot. Instead, it's plotted in the top one.

This seems to be caused by a bug in the code. I'm not sure if this is the desired behaviour, but the following lines (pyqtgraph/examples/Flowcharts.py lines 65-67):

```
pw2Node = fc.createNode('PlotWidget', pos=(150, -150))
pw2Node.setPlot(pw2)
pw2Node.setPlotList(plotList)
```

Result in the 'output' bring plotted on the top plot. Changing the lines to (notice how lines 66 and 67 have been swapped):

```
pw2Node = fc.createNode('PlotWidget', pos=(150, -150))
pw2Node.setPlotList(plotList)
pw2Node.setPlot(pw2)
```
 
Correctly plots the 'output' in the bottom plot.

As stated, I don't know if this behaviour is intentional, or there is a deeper bug in the code that causes this unexpected behaviour.

j9ac9k:
The flowcharts module definitely doesn't get as much love/attention as the rest of the library.  

The metaarray module was removed from pyqtgraph some time ago, ... that's embarrassing that we missed the usage of it in the flowchart module tho!

j9ac9k:
The metaarray issue is no longer relevant as it's been removed from pyqtgraph.  The other issue is still relevant.

j9ac9k:
Fixed wit #3060 being merged.  thanks for the fix @ghylander

PR #3060, opened 2024-06-10
Update Flowchart.py

ghylander:
As stated in issue #2684:

The output of the gaussian filter should be plotted in the bottom plot. Instead, it's plotted in the top one.
The fix is simply to switch the positions of lines 64 and 65.

I left the issue open because I didn't (and still don't) know if it was a bug or intended behaviour.
Should this happen? Or should calling the methods in the original order display the output in the bottom plot?


j9ac9k:
Hi @ghylander 

Thanks for the PR.  Unfortunately the flow-chart module does not get the attention other parts of the library get (PlotCurveItem, PlotDataItem, ScatterPlotItem or ImageItem), so it would not surprise me if this is indeed a bug and not intended behavior.  Thanks you for the fix!

PR #3061, opened 2024-06-10
Support downsampling of image data with NaNs

nup002:
This PR adds tolerance for NaNs in ImageItem autodownsampling. 

In the original code, NaNs in the image data would dominate after autodownsampling. With these changes, NaNs are ignored during downsampling.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1765,7 +1765,7 @@ def gaussianFilter(data, sigma):
     return filtered + baseline
     
     
-def downsample(data, n, axis=0, xvals='subsample', has_nans=False):
+def downsample(data, n, axis=0, xvals='subsample', *, has_nans=False):
     """Downsample by averaging points together across axis.
     If multiple axes are specified, runs once per axis.
     """
@@ -1773,7 +1773,7 @@ def downsample(data, n, axis=0, xvals='subsample', has_nans=False):
         if not hasattr(n, '__len__'):
             n = [n]*len(axis)
         for i in range(len(axis)):
-            data = downsample(data, n[i], axis[i])
+            data = downsample(data, n[i], axis[i], has_nans=has_nans)
         return data
     
     if n <= 1:
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1765,7 +1765,7 @@ def gaussianFilter(data, sigma):
     return filtered + baseline
     
     
-def downsample(data, n, axis=0, xvals='subsample'):
+def downsample(data, n, axis=0, xvals='subsample', has_nans=False):
     """Downsample by averaging points together across axis.
     If multiple axes are specified, runs once per axis.
     """
@@ -1786,7 +1786,10 @@ def downsample(data, n, axis=0, xvals='subsample'):
     sl[axis] = slice(0, nPts*n)
     d1 = data[tuple(sl)]
     d1.shape = tuple(s)
-    d2 = d1.mean(axis+1)
+    if has_nans:
+        d2 = np.nanmean(d1, axis+1)
+    else:
+        d2 = d1.mean(axis+1)
     return d2
 
 def _compute_backfill_indices(isfinite):
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1765,7 +1765,7 @@ def gaussianFilter(data, sigma):
     return filtered + baseline
     
     
-def downsample(data, n, axis=0, xvals='subsample', *, has_nans=False):
+def downsample(data, n, axis=0, xvals='subsample', *, nanPolicy='ignore'):
     """Downsample by averaging points together across axis.
     If multiple axes are specified, runs once per axis.
     """
@@ -1773,7 +1773,7 @@ def downsample(data, n, axis=0, xvals='subsample', *, has_nans=False):
         if not hasattr(n, '__len__'):
             n = [n]*len(axis)
         for i in range(len(axis)):
-            data = downsample(data, n[i], axis[i], has_nans=has_nans)
+            data = downsample(data, n[i], axis[i], nanPolicy=nanPolicy)
         return data
     
     if n <= 1:
@@ -1786,10 +1786,12 @@ def downsample(data, n, axis=0, xvals='subsample', *, has_nans=False):
     sl[axis] = slice(0, nPts*n)
     d1 = data[tuple(sl)]
     d1.shape = tuple(s)
-    if has_nans:
+    if nanPolicy == 'ignore':
         d2 = np.nanmean(d1, axis+1)
-    else:
+    elif nanPolicy == 'propagate':
         d2 = d1.mean(axis+1)
+    else:
+        raise ValueError(f"Keyword argument {nanPolicy=} must be one of ['propagate', 'ignore'].")
     return d2
 
 def _compute_backfill_indices(isfinite):
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1765,7 +1765,7 @@ def gaussianFilter(data, sigma):
     return filtered + baseline
     
     
-def downsample(data, n, axis=0, xvals='subsample', *, nanPolicy='ignore'):
+def downsample(data, n, axis=0, xvals='subsample', *, nanPolicy='propagate'):
     """Downsample by averaging points together across axis.
     If multiple axes are specified, runs once per axis.
     """
@@ -1788,10 +1788,10 @@ def downsample(data, n, axis=0, xvals='subsample', *, nanPolicy='ignore'):
     d1.shape = tuple(s)
     if nanPolicy == 'propagate':
         d2 = d1.mean(axis+1)
-    elif nanPolicy == 'ignore':
+    elif nanPolicy == 'omit':
         d2 = np.nanmean(d1, axis+1)
     else:
-        raise ValueError(f"Keyword argument {nanPolicy=} must be one of ['propagate', 'ignore'].")
+        raise ValueError(f"Keyword argument {nanPolicy=} must be one of {'propagate', 'omit'}.")
     return d2
 
 def _compute_backfill_indices(isfinite):
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -1786,10 +1786,10 @@ def downsample(data, n, axis=0, xvals='subsample', *, nanPolicy='ignore'):
     sl[axis] = slice(0, nPts*n)
     d1 = data[tuple(sl)]
     d1.shape = tuple(s)
-    if nanPolicy == 'ignore':
-        d2 = np.nanmean(d1, axis+1)
-    elif nanPolicy == 'propagate':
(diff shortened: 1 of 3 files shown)


pyqtgraph/functions.py
```diff
@@ -1786,7 +1786,10 @@ def downsample(data, n, axis=0, xvals='subsample'):
     sl[axis] = slice(0, nPts*n)
     d1 = data[tuple(sl)]
     d1.shape = tuple(s)
-    d2 = d1.mean(axis+1)
+    if has_nans:
+        d2 = np.nanmean(d1, axis+1)
+    else:
+        d2 = d1.mean(axis+1)
```
j9ac9k, line 1792: I'm curious if we should just run `np.nanmean` for all codepaths, or is there a performance penalty for running `nanmean` vs. `mean`?

nup002: np.nanmean has a significant performance penalty compared to np.mean:

```
>>> timeit.timeit('np.mean(arr)', number=1000, setup='import numpy as np; arr = np.random.random(int(1E6))')
0.24538480000046548
>>> timeit.timeit('np.nanmean(arr)', number=1000, setup='import numpy as np; arr = np.random.random(int(1E6))')
3.1709375999998883
```

j9ac9k: oh...wow... didn't realize it was _that_ much slower.  Thanks for the post.

pyqtgraph/functions.py
```diff
@@ -1765,7 +1765,7 @@ def gaussianFilter(data, sigma):
     return filtered + baseline
     
     
-def downsample(data, n, axis=0, xvals='subsample'):
+def downsample(data, n, axis=0, xvals='subsample', has_nans=False):
```
j9ac9k, line 1768: Can you make `has_nans` be keyword only?    This can be done via `..., xvals='subsamble', *, has_nans=False):`

j9ac9k (changes requested):
The `has_nans` keyword argument is propagated into the recursive call on line 1776 in `functions.py`

j9ac9k:
Hi @nup002 

I posted a couple of comments, the other thing I'm not sure about is if we want the keyword argument to be `has_nans`.  It's a bit late now, but I'd be curious to know if there is anything similar in the library that we can have the keyword match.  If nothing inside pyqtgraph, then is there anything in Qt that's equivalent and if not there, anything in numpy/scipy.

pyqtgraph/functions.py
nup002: This has been resolved.

nup002:
> Hi @nup002
> 
> I posted a couple of comments, the other thing I'm not sure about is if we want the keyword argument to be `has_nans`. It's a bit late now, but I'd be curious to know if there is anything similar in the library that we can have the keyword match. If nothing inside pyqtgraph, then is there anything in Qt that's equivalent and if not there, anything in numpy/scipy.

The only thing I can think of which is similar is Scipy's `nan_policy`: https://docs.scipy.org/doc/scipy/dev/api-dev/nan_policy.html
We can implement this instead of the simpler `has_nans` for finer control of the desired behaviour of `downsample`. But I am not sure I think this makes much sense since when downsampling the desired behaviour is very likely that NaNs must not propagate.

NilsNemitz:
If there *is* a potential discussion on the desired behavior, then isn't the parameter really `skip_nans` instead of `has_nans`?

nup002:
Yes, but `skip_nans` implies a NaN check within `downsample`, and this is expensive to do every time it is called. Instead, the proposed implementation uses the `self._imageHasNans` attribute of `ImageItem` to avoid repeated NaN checks. I also do not feel that permitting NaN propagation in the downsampling makes much sense, except in certain edge cases.

NilsNemitz:
Sorry, I misunderstood the discussion on the keyword naming, I thought this is exposed to the user. But the change is really that the existing 'has_nans' information is now passed on to the downsampling function which takes on a different behaviour then.

It seems to me that you are introducing a new downsampleMethod, like we already have in PlotDataItem:
https://github.com/pyqtgraph/pyqtgraph/blob/8b60c0e4d9959595dd6ca4a45eb4a42dd903cdc3/pyqtgraph/graphicsItems/PlotDataItem.py#L406 

Would it be reasonable to add a similar switch here, creating a user choice between `downsampleMethod='mean'` and `downsampleMethod='nanmean'`?

nup002:
I propose that we expose an optional keyword argument to the user in ``ImageItem.__init__`` named 'nan_policy' that defines how to handle NaNs during downsampling. The valid values would be 'propagate' which uses ``np.mean`` and 'ignore' which uses ``np.nanmean``. We should at the same time add this functionality to ``PlotDataItem``, which has the same limitation with regards to NaNs propagating into the downsampled data.

nup002:
I have implemented `nanPolicy` as an optional keyword argument for ImageItem which defaults to 'ignore', and which can be set with a the method ``ImageItem.setNanPolicy``:

```
Control how NaN values are handled during downsampling for this ImageItem.

Parameters
----------
nanPolicy : str
    Must be one of ['ignore`, 'propagate']. If 'nanPolicy' is 'ignore', NaNs
    are automatically ignored during downsampling, at the expense of
    performance. If 'nanPolicy' is 'propagate', NaNs are kept during
    downsampling. ImageItem is instantiated with nanPolicy set to 'ignore'.
```

If this feature request is approved, I will work on adding matching functionality to `PlotDataItem` so that it also can ignore NaNs during downsampling.

pyqtgraph/functions.py
```diff
@@ -1765,15 +1765,15 @@ def gaussianFilter(data, sigma):
     return filtered + baseline
     
     
-def downsample(data, n, axis=0, xvals='subsample'):
+def downsample(data, n, axis=0, xvals='subsample', *, nanPolicy='ignore'):
```
NilsNemitz, line 1768: renaming 'nan_policy' to 'nanPolicy' seems like a good idea for matching our Qt-inspired case style.
Shouldn't 'ignore' be 'omit' if we are aiming to resemble SciPy conventions?

pyqtgraph/functions.py
```diff
@@ -1786,7 +1786,12 @@ def downsample(data, n, axis=0, xvals='subsample'):
     sl[axis] = slice(0, nPts*n)
     d1 = data[tuple(sl)]
     d1.shape = tuple(s)
-    d2 = d1.mean(axis+1)
+    if nanPolicy == 'ignore':
+        d2 = np.nanmean(d1, axis+1)
```
NilsNemitz, line 1790: Since the simple mean of 'propagate' is the fast solution, I'd propose to make that the default, and to also test for it first.

pyqtgraph/graphicsItems/ImageItem.py
```diff
@@ -323,6 +324,24 @@ def setAutoDownsample(self, active: bool=True):
...
+        """
+        Control how NaN values are handled during downsampling for this ImageItem.
+
+        Parameters
+        ----------
+        nanPolicy : str
+            Must be one of ['ignore`, 'propagate']. If 'nanPolicy' is 'ignore', NaNs
+            are automatically ignored during downsampling, at the expense of
+            performance. If 'nanPolicy' is 'propagate', NaNs are kept during
+            downsampling. ImageItem is instantiated with nanPolicy set to 'ignore'.
```
NilsNemitz, line 337: You've actually added a switch to make a new ImageItem with any nanPolicy. Maybe:
```python
Unless a different policy was specified, a new ImageItem is created with ``nanPolicy='propagate'``.
```

pyqtgraph/graphicsItems/ImageItem.py
```diff
@@ -323,6 +324,24 @@ def setAutoDownsample(self, active: bool=True):
...
+        """
+        assert nanPolicy in ['propagate', 'ignore'], (f"{nanPolicy=} must be one "
+                                                      f"of ['propagate', 'ignore']")
+        self.nanPolicy = nanPolicy
```
NilsNemitz, line 341: Since this is catching user errors, it should probably raise a ValueError (instead of using assert).

pyqtgraph/functions.py
```diff
@@ -1786,7 +1786,12 @@ def downsample(data, n, axis=0, xvals='subsample'):
...
+    elif nanPolicy == 'propagate':
+        d2 = d1.mean(axis+1)
+    else:
+        raise ValueError(f"Keyword argument {nanPolicy=} must be one of ['propagate', 'ignore'].")
```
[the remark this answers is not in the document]
j9ac9k: If a keyword only argument can be one of several things (in this case string) the syntax is:

`{ 'option_1', 'option_2' }`

You can see an example of it here (but you have to scroll down a bit) https://numpydoc.readthedocs.io/en/latest/format.html#parameters

pyqtgraph/graphicsItems/ImageItem.py
j9ac9k: Agreed if nanPolicy is not in the above, we can raise a `ValueError`, let's not have an `assert` check here.

pyqtgraph/graphicsItems/ImageItem.py
```diff
@@ -399,6 +421,8 @@ def setOpts(self, update: bool=True, **kwargs):
             self.menu = None
         if 'autoDownsample' in kwargs:
             self.setAutoDownsample(kwargs['autoDownsample'])
+        if 'nanPolicy' in kwargs:
+            self.nanPolicy = kwargs['nanPolicy']
```
j9ac9k, line 425: I think we want to call `self.setNanPolicy(kwargs['nanPolicy'])`.

`self.nanPolicy` should not be an attribute, but should be a method that gets the currently stored value.  I would suggest renaming to `self._nanPolicy` and have another method

```python
def nanPolicy(self):
    return self._nanPolicy
```

That will be more in line w/ Qt's getters and setters.

pyqtgraph/graphicsItems/ImageItem.py
j9ac9k: Also rename the attribute to `._nanPolicy`, see comment below about getter/setter.

j9ac9k:
Hi @nup002 

My apologies for sitting on this for so long before commenting.  I recently underwent a major international move, and it's taken me a while to get sorted out in my current home.

I agree with @NilsNemitz 's comments.  I have a couple of nit-pick items

1. `self.nanPolicy` should not be an attribute, but a method.  The attribute should be `self._nanPolicy` and the getter method for that attribute should be `self.nanPolicy()`. This lines up with our API and the API of Qt.
2. There is an assert check to ensure the nanPolicy string is one of the expected kids, as assert checks can easily be turned off, I avoid having them outside of testing code.  Instead I suggest having an if-statement, where if the value of `nanPolicy` is not one of the expected kinds, it raises a `ValueError`.

nup002:
@j9ac9k All of your suggestions, and those of @NilsNemitz, are reasonable. I will get on them once I can find some time.

nup002 (commented):
All suggestions implemented.

j9ac9k:
Thanks @nup002 for this PR, this looks good to me, merging.  I really appreciate your patience on this!

PR #3070, opened 2024-06-23
Fixes: Error when using FFT on data with one point

### LINKED ISSUE #3018: Error when using FFT on data with one point
Activating the FFT transform on data with one point raises the following error:

> File ~\anaconda3\lib\site-packages\pyqtgraph\graphicsItems\PlotDataItem.py:1190 in _fourierTransform
uniform = not np.any(np.abs(dx-dx[0]) > (abs(dx[0]) / 1000.))
IndexError: index 0 is out of bounds for axis 0 with size 0

This occurs because `dx = np.diff(x)` is empty when x is of size 1.

A possible solution is to add the following condition `if len(x) == 1: return np.array([0]), abs(y)` in the beginning of _fourierTransform in PlotDataItem.py.

 * PyQtGraph version: tested on 0.13.3 but will also raise same error in 0.13.7 has the relevant code has not changed.

Silas-K:
see #3018

### Handle single data point case in Fourier Transform
- Adds a condition to handle cases where there is only one data point
- Ensures the function returns zero frequency and the magnitude of the single data point as fallback

### Credits
The solution was proposed by [Python-simulation](https://github.com/Python-simulation) in issue #3018.

EDIT (added for automatic issue closing by @j9ac9k):

Fixes #3018

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -1771,6 +1771,8 @@ class PlotDataItem(GraphicsObject):
     def _fourierTransform(x, y):
         # Perform Fourier transform. If x values are not sampled uniformly,
         # then use np.interp to resample before taking fft.
+        if len(x) == 1: 
+            return np.array([0]), abs(y)
         dx = np.diff(x)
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for the PR @Silas-K ! Appreciate the PR!

Issue #2312, opened 2022-05-26
artifacts in fill under curve for histograms in log mode


servaenergy:
### Short description
Some histograms are not properly filled when log mode is enabled (for X or Y axis).

### Steps to reproduce
1. Run `histogram.py` from examples.
2. Right-click to bring up menu, select "Plot Options" -> "Transforms" -> "log X" or "log Y"'

Note that the dataset is randomly generated and it does not happen every single time. It does happen about 7 out of 10 times for me.

### Expected behavior
Histogram stepped-curve is properly filled all the way down to fill level.

### Real behavior
Fill is incomplete, cut off. See attached screenshot

### Tested environment(s)

 * PyQtGraph version: `0.12.4`
 * Qt Python binding: `PySide6 6.2.1 Qt 6.2.1`
 * Python version: `Python 3.9.7`
 * NumPy version: `1.21.6`
 * Operating system: `Linux fes 5.13.0-44-generic #49-Ubuntu SMP Wed May 18 13:28:06 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

 * Installation method: `pip3 install -r requirements.txt`

```
pyqtgraph>=0.12.3
PySide6==6.2.1 # 6.2.2 throws exception with pyqtgraph == 0.12.3
cython>=0.29.24
psutil~=5.8.0
numpy~=1.21.4
sharedmem>=0.3.8
aiohttp~=3.8.1
scipy>=1.6.0
setuptools>=60.5.0
h5py>=3.6.0
Pillow>=9.0.0
aiortc==1.2.1
wheel>=0.34.2
fast-histogram>=0.10
```

pijyoi:
git bisect points to 0cc35806872c1b658a239046de34f9b18c8189dd

A reproducible script that initializes the random seed.
```python
import pyqtgraph as pg
import numpy as np

rng = np.random.default_rng(seed=0)

pg.mkQApp()
pw = pg.PlotWidget()
pw.show()

vals = np.hstack([rng.normal(size=500), rng.normal(size=260, loc=4)])
y,x = np.histogram(vals, bins=np.linspace(-3, 8, 40))
plt = pw.plot(x, y, stepMode="center", fillLevel=0, fillOutline=True, brush=(0,0,255,150))
plt.setLogMode(False, True)

pg.exec()
```

pijyoi:
[...]

PR #3071, opened 2024-06-24
implement fill for connect="finite"

pijyoi:
Implement the fill behavior for `connect="finite"` as defined in #3068

### DIFF
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -850,7 +850,7 @@ class PlotCurveItem(GraphicsObject):
         connect = self.opts['connect']
         return (
             # not meaningful to fill disjoint lines
-            isinstance(connect, str) and connect == 'all'
+            isinstance(connect, str) and connect in ['all', 'finite']
             # guard against odd-ball argument 'enclosed'
             and isinstance(self.opts['fillLevel'], (int, float))
         )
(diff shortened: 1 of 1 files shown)


NilsNemitz:
I have no further comments.   :)

pijyoi:
this PR seems to fix #2312

pyqtgraph/graphicsItems/PlotCurveItem.py
```diff
@@ -869,17 +869,6 @@ def _getFillPathList(self, widget):
                 baseline=None
             )
 
-        if not self.opts['skipFiniteCheck']:
```
j9ac9k, line 872: I love that this check on `skipFiniteCheck` is going away!

pijyoi:
An animation to demonstrate this PR as well as #3068
```python
import random
import numpy as np
import pyqtgraph as pg

# pg.setConfigOptions(useOpenGL=True, enableExperimental=True)

pg.mkQApp()
plt = pg.PlotWidget()
plt.show()
data = np.zeros(1000)
data[:-1] = np.nan

curve = pg.PlotCurveItem(data, connect="finite", fillLevel=0.0, brush=(50, 50, 200, 100))
plt.addItem(curve)

def update():
    val = random.gauss(0, 1)
    data[:-1] = data[1:]
    data[-1] = data[-2] + val

    # inject a nan every now and then
    if random.randrange(500) == 0:
        data[-2] = np.nan

    level = np.nanmean(data)
    curve.setData(data, fillLevel=level)

timer = QtCore.QTimer()
timer.timeout.connect(update)
timer.start(0)

pg.exec()
```

j9ac9k:
Love the diff @pijyoi thanks for simplifying one of our most used GraphicsItems.

Issue #3073, opened 2024-06-25
Change Plot Background Color via Context Menu


mohass98:
### Description:
I would like to introduce a feature that allows users to change the background color of a plot directly through its context menu. This enhancement aims to provide a more interactive and customizable plotting experience.

### Current Behavior:
Currently, users cannot modify the background color of a plot through the context menu. The plot background is set to a default color and can only be changed within the code.

### Expected Behavior:
Users should be able to right-click on a plot to access the context menu and choose a new background color for the plot from the QColorDialog. The color change should be applied immediately and should remain effective until the plot is reset or the application is closed.

### GraphicsScene.py:
- Added an action to the context menu of the GraphicsScene:

```python
self.setBackgroundAction = QtGui.QAction(
    QtCore.QCoreApplication.translate("GraphicsScene", "Set Background color"), self
)
self.setBackgroundAction.triggered.connect(self.setBackgroundFromContextMenu)
self.contextMenu.append(self.setBackgroundAction)
```

- Added a new signal to pass the new color as an argument:

```python
sigBackgroundChanged = QtCore.Signal(str)   ## emitted when the Background Color was changed via the context menu
```

- Emit the signal when Color was selected

```python
    def setBackgroundFromContextMenu(self):
        # Open the color picker dialog
        color = QColorDialog.getColor()
        
        # Check if a valid color was selected
        if color.isValid():
            # Store the selected color as a hex string
            selected_color_hex = color.name()  # Color in hex format

            # Emit the signal with the selected color
            self.sigBackgroundChanged.emit(selected_color_hex)
```

### GraphicsView.py:
- Listen to the signal from GraphicsScene (which is a member of GraphicsView named sceneObj) and connect it to the setBackground method:
```python
    self.sceneObj.sigBackgroundChanged.connect(self.setBackground)
  ```

### Implementation
![ChangeBG2](https://github.com/pyqtgraph/pyqtgraph/assets/153856978/7bbcc24b-2904-43eb-baab-fc263671a858)
![ChangeBG1](https://github.com/pyqtgraph/pyqtgraph/assets/153856978/e531695b-fb7e-4ebc-aa0c-1e19dd2e9be5)

**I would appreciate feedback on this proposed feature and any suggestions for improvement.**

PR #3080, opened 2024-06-29
Change background color in context menu

mohass98:
### Summary

Introduces a new feature that enables users to change the background color of plots directly through the context menu.

### Issue
Reference to issue #3073 in which the feature was introduced

### Changes Made

Added functionality to the context menu of plots to allow users to select and apply a new background color with the QColorDialog.
Implemented immediate application of the selected color, persisting until plot reset or application closure.

### How to Test

- Right-click on any plot to open the context menu.
- Select "Set Background Color" from the menu.
- Choose a color from the QColorDialog.
- Observe that the plot background color changes immediately and remains as selected until the plot is reset or the application is closed.

This feature enhancement aims to provide users with a more interactive and customizable plotting experience, enhancing visual clarity and usability in presentations and data analysis without changing the code.

### DIFF
--- a/pyqtgraph/GraphicsScene/GraphicsScene.py
+++ b/pyqtgraph/GraphicsScene/GraphicsScene.py
@@ -2,7 +2,6 @@ import warnings
 import weakref
 from time import perf_counter, perf_counter_ns
 
-from PyQt6.QtWidgets import QColorDialog
 
 from .. import debug as debug
 from .. import getConfigOption
@@ -116,7 +115,7 @@ class GraphicsScene(QtWidgets.QGraphicsScene):
 
     def setBackgroundFromContextMenu(self):
         # Open the color picker dialog
-        color = QColorDialog.getColor()
+        color = QtWidgets.QColorDialog.getColor()
 
         # Check if a valid color was selected
         if color.isValid():
--- a/pyqtgraph/GraphicsScene/GraphicsScene.py
+++ b/pyqtgraph/GraphicsScene/GraphicsScene.py
@@ -114,6 +114,23 @@ class GraphicsScene(QtWidgets.QGraphicsScene):
         self._lastMoveEventTime = 0
 
     def setBackgroundFromContextMenu(self):
+        """
+        Opens a color picker dialog to allow the user to select a background color from the context menu.
+    
+        This method performs the following steps:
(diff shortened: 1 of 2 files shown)


pyqtgraph/GraphicsScene/GraphicsScene.py
```diff
@@ -80,6 +81,7 @@ class GraphicsScene(QtWidgets.QGraphicsScene):
     sigItemAdded = QtCore.Signal(object)  ## emits the item object just added
     sigItemRemoved = QtCore.Signal(object)  ## emits the item object just removed
 
+    sigBackgroundChanged = QtCore.Signal(str)   ## emitted when the Background Color was changed via the context menu
```
j9ac9k, line 84: why is it emitting a `str`? as the dtype?

j9ac9k: disregard me!

j9ac9k:
Thanks for the PR @mohass98 !

I'm going to assign this one to @NilsNemitz as he's worked on this before.  Bit I did want to make some comments.

We might be able to just use a `QAction` that starts a QColorPicker, and sets the result to `setBackground`.  `setBackground` accepts as its argument anything that is accepted by `mkBrush` which is quite a bit!   I'm not a fan of the method name as the implication is that it is only called form the contextMenu, and we'll likely want something more generic.

NilsNemitz:
Hi!
I think what @j9ac9k is talking about here is PR #1625 . That one got too far out of hand to parse into a manageable sequence of patches, but it did have some working features. Maybe we can bring some of that functionality into your PR, if it is needed to e.g. trigger a redraw after updating the color setting.

I hope I may ask some questions on your suggested implementation.

Could you let me know why you picked a signal to transfer the background color information?
Does ```self.sigBackgroundChanged.emit(selectedColorHex)``` just act as an indirect way to call `setBackground` on all connected `GraphicsView` objects? 

Wouldn't the signal also be received by *all* possible `GraphicsView` items? Would selecting a background color in the context menu of *one* graph change the background color of *all* graphs? That might not be the expected result.

If it is possible to reach the appropriate `GraphicsView` by crawling up the chain of parent items, does calling `setBackground` on that directly achieve the desired result?

PR #3082, opened 2024-06-30
don't draw non-finite pairs

pijyoi:
Comments
https://github.com/pyqtgraph/pyqtgraph/pull/2011#issuecomment-942358983
through
https://github.com/pyqtgraph/pyqtgraph/pull/2011#issuecomment-942910785
had a discussion (and a demonstrating example) on the behavior to employ when `connect="pairs"` had non-finites in the data.

This PR finally implements the saner behavior of not drawing pairs containing any non-finite coordinate.

image

The image comparison of the following should have failed, but it doesn't because there are too few white pixels to trigger the image-comparison tolerance threshold.
https://github.com/pyqtgraph/pyqtgraph/blob/master/tests/images/plotcurveitem/connectpairs.png
Which raises the question of whether any of the 4 plotcurveitem reference images were useful at all.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -2068,8 +2068,12 @@ def arrayToQPath(x, y, connect='all', finiteCheck=True):
 
     # decide which points are connected by lines
     if connect == 'pairs':
+        mask = 1                # by default connect every 2nd point to every 1st one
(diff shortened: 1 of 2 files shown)


j9ac9k:
Thanks @pijyoi this LGTM, merging.

PR #3089, opened 2024-07-07
bugfix: PlotCurveItem OpenGL didn't invalidate render_cache on cleanup

pijyoi:
`cleanup()` is intended to be called for the following reasons:
1) on program termination (but this may not always happen due to reasons out of our control)
2) if a context change is detected
For both cases, OpenGL objects are destroyed.

For case (2), the state of the class is expected to be restored to a freshly initialized state, in preparation for a re-initialization.

The bug here is that `render_cache` was not invalidated even though the VBO that it was relying on had just been nuked.

Included is a change to `MouseSelection.py` to trigger this bug.


j9ac9k:
Nice catch, thanks for the fix!

PR #3090, opened 2024-07-07
PColorMeshItem: implement opengl rendering

pijyoi:
`PColorMeshItem` is unusable when a non-small mesh is used.
Even panning a static `PColorMeshItem` is very slow. (i.e. just redrawing the rendered `QPicture` is slow)

Following the example of the recent OpenGL code updates to `PlotCurveItem`, we can make `PColorMeshItem` usable by rendering using OpenGL.

An example to demonstrate the speed.
We can use the peegee icons of various sizes to test out the performance.
On my laptop, `peegee_256px.png` without OpenGL gets a dismal 5 fps; whereas with OpenGL, `peegee_512@2x.png` hits 30 fps. i.e. an improvement of (1024/256)**2 * 30/5 == 96x
```python
import sys

import numpy as np

import pyqtgraph as pg

qimage = QtGui.QImage(sys.argv[1])
qimage.convertTo(QtGui.QImage.Format.Format_Grayscale8)
z = pg.functions.ndarray_from_qimage(qimage)


Z = z
nx, ny = z.shape
x = np.linspace(0, nx, nx+1, endpoint=True)
y = np.linspace(0, ny, ny+1, endpoint=True)
X, Y = np.meshgrid(x, y)
X += (nx * 0.025) * np.sin(2*np.pi*2*np.linspace(0, 1, nx+1))[:, np.newaxis]
Y += (ny * 0.025) * np.sin(2*np.pi*2*np.linspace(0, 1, ny+1))

pg.setConfigOptions(useOpenGL=True, enableExperimental=True)

pg.mkQApp()
win = pg.PlotWidget()
win.invertY(True)
cmap = pg.colormap.ColorMap(None, [0.0, 1.0])
pcmi = pg.PColorMeshItem(X, Y, Z, colorMap=cmap)
win.addItem(pcmi)
win.setTitle('')
win.show()

cnt = 0
def update():
    global cnt
    cnt = (cnt + 1) % z.shape[1]
    Z = np.roll(z, cnt, axis=1)
    pcmi.setData(None, None, Z)
    framecnt.update()

timer = QtCore.QTimer()
timer.timeout.connect(update)
timer.start()

framecnt = FrameCounter()
framecnt.sigFpsUpdate.connect(lambda fps: win.setTitle(f'{fps:.1f} fps'))

pg.exec()
```

An addition to the `PColorMeshItem` API. A use-case that I have is that the mesh (XY) is fixed and only the Z values are varying. I have overloaded the `setData(None, None, Z)` to mean that the existing X, Y should be re-used. This allows to not waste time uploading vertices that have not changed.
(The converse also works: `setData(X, Y, None)` means that Z should be re-used; although I don't have such a use-case)

One wart is that on termination, the following error message is shown:
```
QOpenGLTexturePrivate::destroy() called without a current context.
Texture has not been destroyed
```
This message doesn't show up on a pure `QOpenGLWidget` implementation. (as opposed to this PR's doing OpenGL painting within a `QGraphicsView`)

### DIFF
--- a/pyqtgraph/graphicsItems/PColorMeshItem.py
+++ b/pyqtgraph/graphicsItems/PColorMeshItem.py
@@ -1,11 +1,25 @@
+import enum
+import importlib
+import warnings
+
 import numpy as np
 
 from .. import Qt, colormap
 from .. import functions as fn
-from ..Qt import QtCore, QtGui
+from .. import getConfigOption
+from ..Qt import compat
+from ..Qt import OpenGLConstants as GLC
+from ..Qt import QtCore, QtGui, QtWidgets, QT_LIB
 from .GraphicsObject import GraphicsObject
 
 __all__ = ['PColorMeshItem']
+
+
+class DirtyFlag(enum.Flag):
+    XY = enum.auto()
+    Z = enum.auto()
+    LUT = enum.auto()
+    DIM = enum.auto()
 
 
 class QuadInstances:
@@ -116,6 +130,7 @@ class PColorMeshItem(GraphicsObject):
         self.y = None
         self.z = None
         self._dataBounds = None
+        self.glstate = None
 
         self.edgecolors = kwargs.get('edgecolors', None)
         if self.edgecolors is not None:
@@ -144,11 +159,13 @@ class PColorMeshItem(GraphicsObject):
             self.setData(*args)
 
 
-    def _prepareData(self, args):
+    def _prepareData(self, args) -> DirtyFlag:
         """
         Check the shape of the data.
         Return a set of 2d array x, y, z ready to be used to draw the picture.
         """
+
+        dirtyFlags = DirtyFlag.XY | DirtyFlag.Z | DirtyFlag.DIM
 
         # User didn't specified data
         if len(args)==0:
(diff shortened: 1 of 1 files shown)


pyqtgraph/graphicsItems/PColorMeshItem.py
```diff
@@ -1,13 +1,173 @@
...
+        self.m_texture = QtOpenGL.QOpenGLTexture(QtOpenGL.QOpenGLTexture.Target.Target2D)
+
+        return self.functions
+
+    def getFunctions(self, context):
+        if QT_LIB == 'PyQt5':
+            # it would have been cleaner to call context.versionFunctions().
+            # however, when there are multiple PlotCurveItems, the following bug occurs:
+            # all except one of the C++ objects of the returned versionFunctions() get
+            # deleted.
```
j9ac9k, line 103: Should this comment be reworded to make it more applicable to PCMI?

j9ac9k:
Just want to post the screenshot for the fun of it:

image

j9ac9k:
Thanks for this @pijyoi  this looks good to me! Merging!

PR #3094, opened 2024-07-13
pcmi: add support for OpenGL ES 3.0

pijyoi:
On master, the OpenGL implementation of `PColorMeshItem` uses GLSL 1.40 if the OpenGL version is detected to be >= 3.1, otherwise it falls back to shaders that are compatible with both OpenGL 2.0 and OpenGL ES 2.0.

This would fail on a system with OpenGL ES >= 3.1, as the GLSL 1.40 shaders are not compatible. 

This PR makes the flat shaders compatible with OpenGL ES >= 3.0.

However, I don't have a system with OpenGL ES 3.0 *and* a Qt compiled against OpenGL ES to actually test it out.
(A Raspberry Pi 4 running 32-bit Raspberry Pi OS would be such a candidate system)

A less ideal fix would be to unconditionally use the non-flat shaders if any version of OpenGL ES is being used.

### DIFF
--- a/pyqtgraph/graphicsItems/PColorMeshItem.py
+++ b/pyqtgraph/graphicsItems/PColorMeshItem.py
@@ -652,7 +652,6 @@ class OpenGLState:
     """
 
     VERT_SRC = """
-        #version 140
         in vec4 a_position;
         in float a_luminance;
         flat out float v_luminance;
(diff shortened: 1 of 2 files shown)


pijyoi:
It turns out that Raspberry Pi OS 32-bits does run on the RPI5 (It runs with a 64-bit kernel and 32-bit userland), so I was able to test out this PR to be working on an OpenGL ES >= 3.0 system and Qt libraries built against OpenGL ES.

The import also needs to be fixed to import PyQt5's ES2 module, since on such systems, only that module is available to be imported.

pijyoi:
Also found out that on Raspberry Pi OS 32-bits (Bookworm),
while Qt5 is compiled against OpenGL ES,
Qt6 is compiled against OpenGL Desktop.

j9ac9k:
Thanks @pijyoi for expanding capability! LGTM Merging.

PR #3096, opened 2024-07-14
Fix crashing and exception throwing when plotted range is very small (~ < 1e-305). Introduce lower limit to zooming in ViewBox.py

### LINKED ISSUE #2370: Very small range data can crash application due to divide by zero issues
I've hit crashes when trying to plot degenerate y data that's very close to 0 but not actually zero. along the lines of a series of 10^-310
This causes the following 2 functions to blow up because when reasonable numbers are divided by almost zero it's winding up as inf and nan in some of the code 
https://github.com/pyqtgraph/pyqtgraph/blob/243c287044ca5a720b02b37e9a456d7bed1adfa3/pyqtgraph/graphicsItems/AxisItem.py#L972
I believe that this needs an epsilon in generateDrawSpecs:
if dif == 0: 
->
if math.fabs(dif) < 0.00001:

https://github.com/pyqtgraph/pyqtgraph/blob/243c287044ca5a720b02b37e9a456d7bed1adfa3/pyqtgraph/graphicsItems/ViewBox/ViewBox.py#L1656
and updateMatrix:
 if vr.height() == 0 or vr.width() == 0:
 ->
 if math.fabs(vr.height()) < 0.00001 or vr.width() == 0:
 vr.width might need it as well for a vertical plot

visar77:
## Goal

- Fixes #2370 (Technically speaking another person wanted to fix it, but after two years It still was there, so I thought it would be okay, if I tried it)
- Introduces lower zoom limit to ViewBox
- Introduces three new functions in functions.py

The issue, which cause #2370 was the inverting of the QTransform (in functions.py), which calculated infinite values for certain QTransform and the PyQt map() function which resulted to NaN, when infinities or very large numbers where involved. I didn't really fix this issue, instead I wrote three new functions in functions.py, which use numpy to calculate it more accurately. 

The fix was already over, when implementing those functions into the code, the only issue was that the ticks in AxisItem broke, when xScale became -inf or +inf in the generateDrawSpecs() function:
```python
dif = self.range[1] - self.range[0]
        if dif == 0:
            xScale = 1
            offset = 0
        else:
            if axis == 0:
                xScale = -bounds.height() / dif
                offset = self.range[0] * xScale - bounds.height()
            else:
                xScale = bounds.width() / dif
                offset = self.range[0] * xScale
```

I solved this by using the new turnInfToSysMax() function, but this results in the ticks breaking after approx. 4e-307.
Why? Well, `xScale` reaches infinity (after rounding sys.max) after approx. 4e-306, due to bounds.height() or bounds.width(). Whilst xScale reaches it's maximum value, the offset still scales with self.range, which results in very weird (and wrong) tick placements.

That's why I also added a lower zoom limit to prevent the ticks from breaking. It is currently set to sys.float_info.min * 200, which is approx. 4.45e-306. 

## Functions
I had to round a lot of values from +-inf to +- sys.max, that's why I implemented turnInfToSysMax.
The functions should hopefully be self-explanatory, if not I would gladly help to understand. The names are not the best out there, if you have a better proposal, I would gladly rename them.

```python
def turnInfToSysMax(val):
    """Turns infinite values into sys.float_info.max and -infinite values into -sys.float_info.max.
       Used for QPointF and QTransform or normal float values.
    """
    if isinstance(val, QtCore.QPointF):
        return QtCore.QPointF(turnInfToSysMax(val.x()), turnInfToSysMax(val.y()))
    if isinstance(val, QtGui.QTransform):
        return QtGui.QTransform(turnInfToSysMax(val.m11()), turnInfToSysMax(val.m12()), turnInfToSysMax(val.m13()),
                                turnInfToSysMax(val.m21()), turnInfToSysMax(val.m22()), turnInfToSysMax(val.m23()),
                                turnInfToSysMax(val.m31()), turnInfToSysMax(val.m32()), turnInfToSysMax(val.m33()))
    if val == math.inf:
        return sys.float_info.max
    if val == -math.inf:
        return -sys.float_info.max
    return val


def np_invert_qtransform(tr):
    """
    Inverts a QTransform without using the Qt API. This is useful when the Qt API fails to invert a matrix and produces NaNs.
    """
    np_inv_tr = np.linalg.inv(np.array([[tr.m11(), tr.m12(), tr.m13()],
                                       [tr.m21(), tr.m22(), tr.m23()],
                                       [tr.m31(), tr.m32(), tr.m33()]]))
    q_transform = QtGui.QTransform(*np_inv_tr.ravel().tolist())
    q_transform = turnInfToSysMax(q_transform)
    return q_transform

def np_map(tr, pos):
    """
    Maps a QPointF using a QTransform without using the Qt API. This is useful when the Qt API fails to map a point and produces NaNs.
    """
    new_x = turnInfToSysMax(tr.m11() * pos.x()) + turnInfToSysMax(tr.m21() * pos.y()) + turnInfToSysMax(tr.dx())
    new_y = turnInfToSysMax(tr.m22() * pos.y()) + turnInfToSysMax(tr.m12() * pos.x()) + tr.dy()

    new_x = turnInfToSysMax(new_x)
    new_y = turnInfToSysMax(new_y)

    if tr.isAffine():
        w = tr.m13() * pos.x() + tr.m23() * pos.y() + tr.m33()
        w = turnInfToSysMax(w)
        new_x /= w
        new_y /= w

    return QtCore.QPointF(new_x, new_y)
```

## Tests

I have tested everything with this example:
```python
"""
This example is meant to illustrate the issue #2370, where the AxisItem breaks
"""
import sys

import numpy as np

import pyqtgraph as pg

app = pg.mkQApp("AxisItem")

win = pg.GraphicsLayoutWidget(show=True, title="AxisItem")
win.resize(800, 600)
win.setWindowTitle("AxisItem - Lower bound of y axis is not respected")

# Enable antialiasing for prettier plots

x_data = [0.1 * i for i in range(100)]
x_data = [1e-308 * i for i in range(100)]
#y_data = np.random.uniform(-5e-306, 5e-306, 100)
y_data = np.random.uniform(-5e-307, 5e-307, 100)
#y_data = np.random.uniform(-2e-308, 2e-308, 100)
#y_data = np.random.uniform(-5e-309, 5e-309, 100)
#y_data = [1 for _ in range(100)]
print(y_data)

font = QtGui.QFont()
font.setPointSize(14)   # A larger font makes the effects more visible

p1 = win.addPlot(x=x_data, y=y_data)
for axis_key in ('top', 'bottom', 'left', 'right'):
    ax = p1.getAxis(axis_key)
    ax.setTickFont(font)

if __name__ == '__main__':
    pg.exec()
```

I have also run `tox`, which worked with every Python 3.11 and 3.12 (I didn't have Python 3.10) and `pytest test` also work with `1001 passed, 708 skipped, 8 xfailed`. The test result was exactly the same even without the changes, which should mean, that everything is fine and I haven't broken anything else.

## Extras
I have left some print statements in the zooming bits to indicate, that you can't zoom further. If I have to remove them, I will.
I also tried to not change too much code, that's also the reason why I introduced three new functions instead of rewriting invertedQTransform.

This is my very first PR, so I might have done some amateur mistakes, but I would love to fix them, if there are any.


visar77:
@j9ac9k I know that you are extremely busy, but can you please allow the checking of the PR, so I at least know, if the PR breaks something?

I'm sorry for pinging / notifying, but I thought that it shouldn't be too much of a hassle for you to allow the checks. Still, I don't want you to feel pressured by me, just wanted to remind you. Take your time and thank you for maintining pyqtgraph.

pyqtgraph/graphicsItems/ViewBox/ViewBox.py
```diff
@@ -1293,8 +1341,15 @@ def wheelEvent(self, ev, axis=None):
...
 
         s = 1.02 ** (ev.delta() * self.state['wheelScaleFactor']) # actual scaling factor
+
+        if s < 1.02 and any(self._reachedMaxZoom):
+            print("Can't zoom further")
+            return
+
         s = [(None if m is False else s) for m in mask]
-        center = Point(fn.invertQTransform(self.childGroup.transform()).map(ev.pos()))
+        inv = fn.np_invert_qtransform(self.childGroup.transform())
```
j9ac9k, line 1350: not using `invertQTransform` here is going to have a _significant_ performance penalty, why is the numpy version needed here?

j9ac9k:
I don't mind being periodically tagged, I do forget about PRs so an occasional reminder/prodding is perfectly fine.

I'll try and review more closely today, but at first glance, does `math.inf` also work w/ `np.inf`?  

Also `invertQTransform` is _significantly_ faster than the numpy equivalent, and it things work correctly, if it fails (determinant being 0) it should fall back to the numpy equivalent method anyway; to merge a change in ViewBox.py that goes to the numpy method I would need a convincing case why `invertQTransform` is not sufficient.

These are just initial thoughts after doing a look over, I'll review more closely later today hopefully.

visar77:
Thank you for your understanding and your response.

1. [`math.inf` is equal to `np.inf`](https://stackoverflow.com/questions/42315541/difference-between-np-inf-and-floatinf)
2. `invertQTransform` and `tr.map()` cause the issue #2370, by creating NaNs and Infinities. The only way I saw for solving this, was to replace them with numpy_versions that 1. don't cause Infinities and NaNs and 2. round infinties to sys.float_max. This fix allows one even to zoom further than sys.float_min ( I tried values of 1e-314, it should be also possible to go till 4e-324, which is the real sys.float_min, but I haven't tested it). But because the ticks of the AxisItem break after a 1e-307,  I tried adding a zoom limit.

pyqtgraph/graphicsItems/ViewBox/ViewBox.py
visar77: Well, if I have tested it correctly, it breaks, if you don't use it there.  Every mapping with `invertQTransform` will break with Infinities and NaNs popping up after a certain threshhold.  Also the mapping with Qt using `tr.map()` will break if the values are near sys.float_max, so I have to both use np_invert and np_map, so it doesn't break. 

I didn't notice any differences in speed, but I also didn't check for them. It would makea lot of sense for the Qt versions to be faster, as they use the C++ implementation, but I would also argue, that the numpy versions are not that far off, as e.g. the inversion algorithm is heavily optimized as well.  The mapping function np_map on the other hand should be (much) slower than Qt's map function, but that would also need to be tested. One way to fix possible performance issue is to use the numpy variant of inverting and mapping only after the zoom has reached certain threshold (e.g. over 1e304).

visar77:
During the weekend I will use timeit to measure the performance difference and maybe even implement a solution for it.

visar77:
You were absolutely right. I tested the functions with the code below and here are the results:

```
Time taken by normal qt_inverse:  5.650940000006813
Time taken by numpy inverse:  23.995663300011074
Time taken by normal qt mapping:  9.853801899997052
Time taken by numpy mapping:  13.534976599999936
```
As you can see my functions are a slower than the normal qt-versions. I will try to optimize them a bit and implement a check, which uses the other functions after a specific zoom threshold. What do you recommend doing, because implementing a check for zooming is maybe not the best solution?

Code:
```
import numpy as np
import timeit

from pyqtgraph.Qt.QtCore import QPointF
from pyqtgraph.Qt.QtGui import QTransform

def qt_inverse():
    x = np.random.uniform(-1e300, 1e300, 9)
    x_tr = QTransform(*x)
    fn.invertQTransform(x_tr)


def qt_numpy_inverse():
    x = np.random.uniform(-1e300, 1e300, 9)
    x_tr = QTransform(*x)
    fn.np_invert_qtransform(x_tr)


def qt_map():
    x = np.random.uniform(-1e300, 1e300, 9)
    x_tr = QTransform(*x)
    pos = x_tr.map(QPointF(*np.random.uniform(-1e300, 1e300, 2)))


def qt_numpy_map():
    x = np.random.uniform(-1e300, 1e300, 9)
    x_tr = QTransform(*x)
    pos = fn.np_map(x_tr, QPointF(*np.random.uniform(-1e300, 1e300, 2)))


if __name__ == '__main__':
    print("Time taken by normal qt_inverse: ", timeit.timeit(qt_inverse, number=1000000))
    print("Time taken by numpy inverse: ", timeit.timeit(qt_numpy_inverse, number=1000000))
    print("Time taken by normal qt mapping: ", timeit.timeit(qt_map, number=1000000))
    print("Time taken by numpy mapping: ", timeit.timeit(qt_numpy_map, number=1000000))

```

j9ac9k:
A little context here regarding how we came to the optimized invertQTransform: https://github.com/pyqtgraph/pyqtgraph/pull/1493

(More context here too: https://github.com/pyqtgraph/pyqtgraph/pull/1418 )

We actually looked at the source code for Qt to see when the `qFuzzyIsNull` took effect, and utilized methods of `QTransform` that wouldn't use that within.

https://github.com/qt/qtbase/blob/e13173c112b729da8f53dd2e81e8116a1ed857cf/src/gui/painting/qtransform.cpp#L347

The best bet would be to try and catch the error condition inside the `invertQTransform` method here: https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L2925-L2940 and either raise an error, or fall back to the numpy/slower method in the error case (which we presently do for ZeroDivisionError, but maybe that isn't sufficient).

visar77:
Hey, I got busy with work, but I'm back and I (hopefully) solved the issue by modifying the ```invertQTransform()``` function slightly. The extra if won't cause performance problems and the usage of the numpy variant will only happen after the determinant goes to infinity,  which shouldn't happen in most cases.

New invertQTransform():
```python
def invertQTransform(tr):
    """Return a QTransform that is the inverse of *tr*.
    A pseudo-inverse is returned if tr is not invertible.
    
    bugs in that method. (specifically, Qt has floating-point precision issues
    when determining whether a matrix is invertible)
    """
    try:
        det = tr.determinant()
        if np.isinf(det):
            return np_invert_qtransform(tr)
        detr = 1.0 / det    # let singular matrices raise ZeroDivisionError
        inv = tr.adjoint()
        inv *= detr
        return inv
    except ZeroDivisionError:
        return _pinv_fallback(tr)
```
The zooming is still limited and won't zoom further than 1e-306. The new invertQTransform() rendered np_map useless, that's why I removed it completely. Sorry for the long wait and for the weird last commits, I messed up whilst trying to push the newest commit.

For testing I still used the python example (in the comments) above.

j9ac9k:
Hi @visar77 

I'm finally getting to this, I'm so sorry for the long delay, I do have some questions that I'll put in line.

pyqtgraph/functions.py
```diff
@@ -2951,6 +2953,34 @@ def invertQTransform(tr):
...
+                                turnInfToSysMax(val.m21()), turnInfToSysMax(val.m22()), turnInfToSysMax(val.m23()),
+                                turnInfToSysMax(val.m31()), turnInfToSysMax(val.m32()), turnInfToSysMax(val.m33()))
+    if val == math.inf:
+        return sys.float_info.max
+    if val == -math.inf:
+        return -sys.float_info.max
+    return val
+
+
+def np_invert_qtransform(tr):
```
j9ac9k, line 2973: Why use this instead of https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L2927-L2933

the pseudo-inversion would handle cases where even if the determinant is 0?

pyqtgraph/graphicsItems/AxisItem.py
```diff
@@ -1492,9 +1492,12 @@
             xScale = -bounds.height() / dif
             offset = self.range[0] * xScale - bounds.height()
         else:
-            xScale = bounds.width() / dif
-            offset = self.range[0] * xScale
-
+            if axis == 0:
+                xScale = fn.turnInfToSysMax(-bounds.height() / dif)
```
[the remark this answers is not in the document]
j9ac9k: Whoops, looks like I messed this up in the merge, I'll fix...

j9ac9k:
It got late last night, and I decided to call it a night, but testing this morning there is an issue with using the scroll wheel on the vertical axis of your example

```python
"""
This example is meant to illustrate the issue #2370, where the AxisItem breaks
"""
import sys

import numpy as np

import pyqtgraph as pg

app = pg.mkQApp("AxisItem")

win = pg.GraphicsLayoutWidget(show=True, title="AxisItem")
win.resize(800, 600)
win.setWindowTitle("AxisItem - Lower bound of y axis is not respected")

# Enable antialiasing for prettier plots

# x_data = [0.1 * i for i in range(100)]
x_data = [1e-308 * i for i in range(100)]
y_data = np.random.uniform(-5e-306, 5e-306, 100)
# y_data = np.random.uniform(-5e-307, 5e-307, 100)
# y_data = np.random.uniform(-2e-308, 2e-308, 100)
# y_data = np.random.uniform(-5e-309, 5e-309, 100)
#y_data = [1 for _ in range(100)]
print(y_data)

font = QtGui.QFont()
font.setPointSize(14)   # A larger font makes the effects more visible

p1 = win.addPlot(x=x_data, y=y_data)
for axis_key in ('top', 'bottom', 'left', 'right'):
    ax = p1.getAxis(axis_key)
    ax.setTickFont(font)

if __name__ == '__main__':
    pg.exec()
```

With the mouse hovering over the vertical axis, when you zoom in, you get the prompt "Can't zoom further".  When you zoom out, it zooms out normally as expected.  But when you try and zoom in again (after having zoomed out), you'll get the "Can't zoom further" message and be unable to zoom in towards the original size.

j9ac9k:
Also I cleaned up some of the diff in `functions.py`

```python
def _pinv_fallback(tr):
    elements = [tr.m11(), tr.m12(), tr.m13(), tr.m21(), tr.m22(), tr.m23()]
    for i, element in enumerate(elements):
        match element:
            case float('inf'):
                elements[i] = sys.float_info.max
            case float('-inf'):
                elements[i] = sys.float_info.min
            case _:
                continue
    arr = np.array(elements + [0., 0., 1.])
    arr.shape = (3, 3)
    pinv = np.linalg.pinv(arr)
    return QtGui.QTransform(*pinv.ravel().tolist())


def invertQTransform(tr):
    """Return a QTransform that is the inverse of *tr*.
    A pseudo-inverse is returned if tr is not invertible.
    
    bugs in that method. (specifically, Qt has floating-point precision issues
    when determining whether a matrix is invertible)
    """
    det = tr.determinant()
    if np.isinf(det):
        return _pinv_fallback(tr)
    try:
        detr = 1.0 / det    # let singular matrices raise ZeroDivisionError
    except ZeroDivisionError:
        return _pinv_fallback(tr)
    else:
        inv = tr.adjoint()
        inv *= detr
        return inv
    

def turnInfToSysMax(val):
    """Turns infinite values into sys.float_info.max and -infinite values into -sys.float_info.max.
       Used for QPointF and QTransform or normal float values.
    """
    if isinstance(val, QtCore.QPointF):
        return QtCore.QPointF(turnInfToSysMax(val.x()), turnInfToSysMax(val.y()))
    
    if isinstance(val, QtGui.QTransform):
        return QtGui.QTransform(turnInfToSysMax(val.m11()), turnInfToSysMax(val.m12()), turnInfToSysMax(val.m13()),
                                turnInfToSysMax(val.m21()), turnInfToSysMax(val.m22()), turnInfToSysMax(val.m23()),
                                turnInfToSysMax(val.m31()), turnInfToSysMax(val.m32()), turnInfToSysMax(val.m33()))
    match val:
        case float('inf'):
            return sys.float_info.max
        case float('-inf'):
            return sys.float_info.min
        case _:
            return val
```

I'm not convinced we need `turnInfoToSysMax` in `functions.py` at all, at places where we think this could be impactful, we can replace calls with something like...

```python
...
elif axis == 0:
    xScale = max(-bounds.height() / diff, sys.float_info.min)
    offset = self.range[0] * xScale - bounds.ehight()
else:
    xScale = min(bounds.width() / diff, sys.float_info.max)
    offset = self.range[0] * xScale
```

in short instead of calling  `turnInfToSysMax` we just use `min` or `max` calls where appropriate.

That said there is still a bug in ViewBox.py.

PR #3111, opened 2024-07-29
Add legend double-click and legend sample click signals.

kevinanewman:
An application may need/want to know which plot data items are visible, the sigSampleClicked signal can be used for this purpose.

An application may want to resize the legend text after a legend has been created. The sigDoubleClicked signal can be used for this.

The example Legend.py has been updated to illustrate example use cases.

Tested against PyQt6 and PySide6.

Thank you for you consideration, 
Kevin

### DIFF
--- a/pyqtgraph/graphicsItems/LegendItem.py
+++ b/pyqtgraph/graphicsItems/LegendItem.py
@@ -29,6 +29,9 @@ class LegendItem(GraphicsWidgetAnchor, GraphicsWidget):
 
     """
 
+    sigDoubleClicked = QtCore.Signal(object, object)
+    sigSampleClicked = QtCore.Signal(object)
+
     def __init__(self, size=None, offset=None, horSpacing=25, verSpacing=0,
                  pen=None, brush=None, labelTextColor=None, frame=True,
                  labelTextSize='9pt', colCount=1, sampleType=None, **kwargs):
@@ -218,6 +221,10 @@ class LegendItem(GraphicsWidgetAnchor, GraphicsWidget):
             sample = item
         else:
             sample = self.sampleType(item)
+
+        if hasattr(sample, 'sigClicked'):
+            sample.sigClicked.connect(self.sampleClickEvent)
+
         self.items.append((sample, label))
         self._addItemToLayout(sample, label)
         self.updateSize()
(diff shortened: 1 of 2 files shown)


pyqtgraph/graphicsItems/LegendItem.py
```diff
@@ -218,6 +221,10 @@ def addItem(self, item, name):
             sample = item
         else:
             sample = self.sampleType(item)
+
+        if hasattr(sample, 'sigClicked'):
+            sample.sigClicked.connect(self.sampleClickEvent)
```
j9ac9k, line 226: There is no need for the sampleClickEvent method, you can connect one signal directly to another

```python
sample.sigClicked.connect(self.sigSampleClicked)
```

j9ac9k:
Thanks for the PR @kevinanewman  !

I mentioned the issue with the method that I don't think we need; the other issue I have is with what the signals are carrying.

`sigSampleClicked` contains the sample that was clicked, but `sigDoubleClicked` contains the double-click event itself, which is maybe what is wanted, but I'm not sure we need to emit that event as a signal?

pyqtgraph/graphicsItems/LegendItem.py
```diff
@@ -339,11 +346,20 @@ def mouseDragEvent(self, ev):
             dpos = ev.pos() - ev.lastPos()
             self.autoAnchor(self.pos() + dpos)
 
+    def mouseDoubleClickEvent(self, ev):
+        self.sigDoubleClicked.emit(self, ev)
```
j9ac9k, line 350: We should probably pass the event to the superclass.

```python
super().mouseDoubleClickEvent(ev)
self.sigDoubleClicked.emit(self, ev)
```

kevinanewman: In my case I wanted the app get a notification, I'm not making a new class or subclass.  If you try out the Legend.py example I updated you'll see what I mean.  I use the double click signal to pop up a dialog box to change the legend font size.  Of course other style changes could be made available as well.

j9ac9k:
Sorry for the delay, this LGTM.  Thanks for the PR @kevinanewman

PR #3113, opened 2024-07-30
fix ComboBox.items()

MrHenning:
The `ComboBox.items()` method returns itself (i.e. a function).
This PR makes return a copy of the `_items` dictionary, as I believe was intended.


j9ac9k:
oh wow, ...how did we miss this 😬 Thanks for the PR @MrHenning

PR #3121, opened 2024-08-06
Raise ParseError on duplicate keys in configfile

outofculture:
Maybe we shouldn't really be maintaining our own config file format, but since we're doing so anyway, here's a behavior we should have.

This includes some minor style updates, as well, which my editor demanded I make before it would let me edit the code (I kid).

### DIFF
--- a/pyqtgraph/configfile.py
+++ b/pyqtgraph/configfile.py
@@ -21,7 +21,7 @@ from .colormap import ColorMap
 from .Point import Point
 from .Qt import QtCore
 
-GLOBAL_PATH = None # so not thread safe.
+GLOBAL_PATH = None  # so not thread safe.
 
 
 class ParseError(Exception):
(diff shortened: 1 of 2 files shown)


pyqtgraph/configfile.py
```diff
@@ -172,21 +174,20 @@ def parseString(lines, start=0, **scope):
                     raise ParseError("Error evaluating expression '%s': [%s: %s]" % (v, ex.__class__.__name__, str(ex)), (ln+1), l)
             else:
                 if ln+1 >= len(lines) or measureIndent(lines[ln+1]) <= indent:
-                    #print "blank dict"
                     val = {}
                 else:
-                    #print "Going deeper..", ln+1
                     (ln, val) = parseString(lines, start=ln+1, **scope)
+            if k in data:
+                raise ParseError(f'Duplicate key: {k}', ln+1, l)
```
outofculture, line 181: Here's the only substantive change.

PR #3123, opened 2024-08-11
Annotate slots with `QtCore.Slot`

bersbersbers:
Compare similar work in https://github.com/matplotlib/matplotlib/pull/27215.

Issues remaining:
- [x] `QLoggingCategory` missing on PySide on Python 3.10
- [x] segmentation faults with PyQt on Python 3.11 and 3.12


j9ac9k:
The matplotlib errors are my doing 😅 feel free to disregard those

bersbersbers:
> The matplotlib errors are my doing 😅 feel free to disregard those

Not quite sure what you mean with those. Anyway, I fixed all issues that I could. Remaining are only 6 tests that crash with segmentation faults. I was able to minimize the code for these segfaults to two handfuls of lines, exposing a crash in the most current PyQt6 v6.7.1. I could not yet find a workaround, and not sure where to report those. Unfortunately, those are blockers for this PR...

```python
# Minimum version of
# https://github.com/pyqtgraph/pyqtgraph/blob/master/tests/test_reload.py
#
# pip install pytest==8.3.2 PyQt6==6.7.1
# pytest test_reload.py

import sys
from importlib import reload
from pathlib import Path


def test_reload() -> None:
    tmp_path = Path("mod_folder").absolute()
    tmp_path.mkdir(exist_ok=True)
    sys.path.insert(0, str(tmp_path))

    with open(tmp_path / "mod_file.py", "w") as f:
        f.write("from PyQt6 import QtCore\n")
        f.write("class Class(QtCore.QObject):\n")
        f.write("    signal = QtCore.pyqtSignal()\n")
        f.write("    @QtCore.pyqtSlot()\n")
        f.write("    def slot(self): pass")

    from mod_file import Class

    c = Class()
    c.signal.connect(c.slot)

    module = sys.modules["mod_file"]
    reload(module)
    c.__class__ = module.__dict__["Class"]

    c.signal.disconnect(c.slot)

```

j9ac9k:
> Not quite sure what you mean with those

Whoops I got my PRs confused. That's what I get for making drive by comments.

Thanks for taking on the effort, I'll try and take a closer look but that test reload function I remember being problematic before.

bersbersbers:
> I'll try and take a closer look

Problem is, there is little you can actually do in terms of the test. The code snippet does not depend on pyqtgraph at all, and segmentation faults should never be able to be triggered from Python code. To me, this is clearly a bug in PyQt.

PySide people usually fix those things very quickly, not sure about PyQt. Haven't even found a way to report bugs for PyQt exception some ... mailing list :)

bersbersbers:
fyi, https://www.riverbankcomputing.com/pipermail/pyqt/2024-August/045989.html

j9ac9k:
> fyi, https://www.riverbankcomputing.com/pipermail/pyqt/2024-August/045989.html

Thanks for making the bug report with Phil. Given his response, I think it would be ok to skip the problematic tests on PyQtX, and add a comment where the test is skipped linking to that post on the mail list.

tests/test_reload.py
```diff
@@ -19,6 +19,7 @@
 
 class C(pg.QtCore.QObject):
     sig = pg.QtCore.Signal()
+    @pg.QtCore.Slot()
```
j9ac9k, line 22: If you remove this, would you still get the segfault?

bersbersbers: No - that is exactly the part that triggers it.

bersbersbers:
> Given his response, I think it would be ok to skip the problematic tests on PyQtX, and add a comment where the test is skipped linking to that post on the mail list.

Agreed - see https://github.com/bersbersbers/pyqtgraph/commit/2dd608f931280e00d3283bc47a280c41f6e32516. I reverted that test to its previous version, adding two comments with a link to the bug report. The test is still run in its old form, I am just not failing it by ignoring a warning that is emitted.

Keep in mind that if the test with slot annotation crashes, actual code with slot annotations that reloads a module can crash in the same way - skipping the test will not save us from that. The question (that I cannot answer) is how representative that test is for things that `pyqtgraph` does at runtime.

j9ac9k:
That test reload is a pretty old test function, I suspect it represents a
test for one of the functions in debug.py; which honestly doesn’t get much
use these days, so I suspect is safe to skip.

Given the response on the mail list, it sounds like this may not have
worked for some time on the PyQt bindings independent of the segfault so i
am not too worried about it.

.github/workflows/etc/requirements.txt
```diff
@@ -18,5 +18,6 @@ h5py==3.11.0
 
 # testing
 pytest==8.3.2
+pytest-qt==4.4.0
```
j9ac9k, line 21: Suppose it was only a matter of time before this amazing pytest plugin made it's way into the library; with it being added, we can certainly use more of its capabilities (in another PR)

bersbersbers: 👍 

Note that I ensured it is an _optional_ dependency at this point: all tests can be run without it.

j9ac9k: Appreciate that, but it won't be long before it's mandatory. Would love to remove some of the stuff in our ui testing files and offload to pytest-qt!

bersbersbers:
Unless I have overlooked a review comment somewhere, this is ready to merge from my point of view.

j9ac9k:
Sorry I've been a bit off the grid; I'll take another glance, but I'm
fairly certain I put all my issues in the last set of comments... My
apologies for the delay.

j9ac9k:
Hi @bersbersbers 

Trying to resolve the git conflicts, but the fact that you made a PR from the `master` branch is complicating things, I'm going to try and resolve on my end, but FYI for future PRs, make a branch name that doesn't match the branch name of the upstream repo 👍🏻

Issue #404, opened 2016-11-22
Interacting with Console Widget and Console help() broken


bennames:
I'm working on a code in which I want the user to primarily use a command line to input commands and then be able to visualize them in an GLViewWidget. Are there any examples of using the console widget to update data appearing in a GLViewWidget (or really any other widget)?

Additionally, in the I believe the help() command in the console widget is broken as it freezes the window when I try to use it.

2xB:
I edited minimally the Console Widget example to answer your question. Just try to run

```python3
setPlotData([1,2,3],[1,4,3])
```

in the console opened by:

```python3
# -*- coding: utf-8 -*-
"""
ConsoleWidget is used to allow execution of user-supplied python commands
in an application. It also includes a command history and functionality for trapping
and inspecting stack traces.

"""
import initExample ## Add path to library (just for examples; you do not need this)

import pyqtgraph as pg
from pyqtgraph.Qt import QtCore, QtGui
import numpy as np
import pyqtgraph.console

app = pg.mkQApp()

pw = pg.plot([1,2],[3,4])
di = pw.getPlotItem().listDataItems()[0]

## build an initial namespace for console commands to be executed in (this is optional;
## the user can always import these modules manually)
namespace = {'pg': pg, 'np': np, 'setPlotData': di.setData}

## initial text to display in the console
text = """
This is an interactive python console. The numpy and pyqtgraph modules have already been imported 
as 'np' and 'pg'. 

Go, play.
"""
c = pyqtgraph.console.ConsoleWidget(namespace=namespace, text=text)
c.show()
c.setWindowTitle('pyqtgraph example: ConsoleWidget')

## Start Qt event loop unless running in interactive mode or using pyside.
if __name__ == '__main__':
    import sys
    if (sys.flags.interactive != 1) or not hasattr(QtCore, 'PYQT_VERSION'):
        QtGui.QApplication.instance().exec_()
```

Your issue regarding the help() command is not reproduced in my environment using the latest pyqtgraph version. As this issue is open since a while, it may be fixed. Please test this and comment further information if it is not.

PR #3130, opened 2024-08-20
Option to execute repl commands in a non-GUI thread

outofculture:
This allows application designers to offer their users command execution in either the GUI thread (the current default) or a non-GUI thread. Users may want this if their command is expected to run a long time, does not interact directly with any GUI elements, and they don't want the entire application to freeze for the duration. GUI execution is still the default, and the checkbox to control this isn't even shown without being explicitly configured, so all existing applications will be unchanged. I regret that this wasn't built very testably, The example henceforth uses it, so it will at least get instatiated during a test run.

### DIFF
--- a/pyqtgraph/console/Console.py
+++ b/pyqtgraph/console/Console.py
@@ -1,12 +1,12 @@
 import os
-import sys
 import pickle
 import subprocess
+import sys
 
 from .. import getConfigOption
 from ..Qt import QtCore, QtWidgets
-from .repl_widget import ReplWidget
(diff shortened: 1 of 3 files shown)


jack-mil:
This seems like a nice addition, and will prevent friction that users encounter when typing the `help()` command into the console widget, and immediatly the GUI freezes up (I assume because `help()` is an interactive method that waits for user input).

I encounter this issue often when testing (the first thing I think of is to try `help()`). Was also mentioned way back in #404.

PR #3131, opened 2024-08-20
ParseError line numbers should account for empty lines

outofculture:
Comments and blank lines were already being skipped during line processing, so we don't need to pre-exclude them. Doing so impacted the accuracy of the ParseErrors. Test included. General style improvements and python-3-isms also added.

### DIFF
--- a/pyqtgraph/configfile.py
+++ b/pyqtgraph/configfile.py
@@ -166,10 +166,10 @@ def parseString(lines, start=0, **scope):
             if re.search(r'\S', v) and v[0] != '#':  ## eval the value
                 try:
                     val = eval(v, scope)
-                except Exception:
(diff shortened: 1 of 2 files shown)


outofculture:
Shoot; comments mess up the line indentation calculation. I have to solve that before this is viable.

outofculture:
Ugh, same bug, different direction. I should test it in production before reopening next time.

j9ac9k:
This module needs to die... python already offers a [config parser](
https://docs.python.org/3/library/configparser.html ) and [tomllib](https://docs.python.org/3/library/tomllib.html#module-tomllib) (in python 3.11), which I would be happy to add support for.

I'll merge this but I'm >< close to slapping a deprecation warning for some point a year+ out.

Issue #2887, opened 2023-11-23
Dock restoreState fails silently if a dock object is the only item in a V- or HContainer


UsYer:
### Short description
Calling restoreState() on a dockArea where a dock is the only object inside a container leads to the state being not correctly restored.
I don't know how it happened that a dock could appear as the only item in a container but one user managed to do that.
I couldn't find out how to trigger that.
Anyway, in that case way too many container are apoptose'd and the state is nor correctly restored

### Code to reproduce
```python
import json
import sys

import pyqtgraph as pg
from PySide6.QtWidgets import QMainWindow
from pyqtgraph.dockarea import Dock, DockArea


class DockApp(QMainWindow):
    def __init__(self, title):
        super().__init__()
        self.dock_area = DockArea()
        self.setCentralWidget(self.dock_area)
        self.dock_area.addDock(Dock(name="Plot 1", closable=False), 'left')
        self.dock_area.addDock(Dock(name="Plot 2", closable=False), 'left')
        self.dock_area.addDock(Dock(name="Plot 4", closable=False), 'left')
        self.dock_area.addDock(Dock(name="Table 1", closable=False), 'left')
        self.dock_area.addDock(Dock(name="Table 2", closable=False), 'left')
        self.dock_area.addDock(Dock(name="Table 3", closable=False), 'left')

        state = json.loads("""{
          "main": [
            "vertical",
            [
              [
                "horizontal",
                [
                  [
                    "vertical",
                    [
                      [
                        "vertical",
                        [
                          [
                            "dock",
                            "Plot 1",
                            {}
                          ]
                        ],
                        {
                          "sizes": [
                            314
                          ]
                        }
                      ],
                      [
                        "dock",
                        "Plot 2",
                        {}
                      ]
                    ],
                    {
                      "sizes": [
                        314,
                        313
                      ]
                    }
                  ],
                  [
                    "vertical",
                    [
                      [
                        "dock",
                        "Table 3",
                        {}
                      ],
                      [
                        "dock",
                        "Table 2",
                        {}
                      ],
                      [
                        "dock",
                        "Table 1",
                        {}
                      ]
                    ],
                    {
                      "sizes": [
                        208,
                        207,
                        208
                      ]
                    }
                  ]
                ],
                {
                  "sizes": [
                    784,
                    783
                  ]
                }
              ],
              [
                "dock",
                "Plot 4",
                {}
              ]
            ],
            {
              "sizes": [
                631,
                210
              ]
[...]

PR #3134, opened 2024-08-25
Apoptose dock area when temporary window is closed

### LINKED ISSUE #3125: DockArea's restoreState is broken when a Dock was undocked and re-docked before calling saveState
### Short description
When one undocks a dock and either closes the temporary window or drags the dock back into the main window, clicks "Save dock state" and then "Restore dock state" the following exception occurs:

>   File "pyqtgraph\examples\dockarea.py", line 77, in load
>     area.restoreState(state)
>   File "pyqtgraph\dockarea\DockArea.py", line 244, in restoreState
>     a.buildFromState(s[0]['main'], docks, a, missing=missing)
>   File "pyqtgraph\dockarea\DockArea.py", line 264, in buildFromState
>     typ, contents, state = state
>     ^^^^^^^^^^^^^^^^^^^^
> TypeError: cannot unpack non-iterable NoneType object

### Code to reproduce
Just use the example in examples/dockarea.py undock and then re-dock a dock and click "Save dock state" and then "Restore dock state".

### Expected behavior
The saved state should have been restored.

### Real behavior
The above exception occured and the state is not restored

```
  File "pyqtgraph\examples\dockarea.py", line 77, in load
    area.restoreState(state)
  File "pyqtgraph\dockarea\DockArea.py", line 244, in restoreState
    a.buildFromState(s[0]['main'], docks, a, missing=missing)
  File "pyqtgraph\dockarea\DockArea.py", line 264, in buildFromState
    typ, contents, state = state
    ^^^^^^^^^^^^^^^^^^^^
TypeError: cannot unpack non-iterable NoneType object
```

### Tested environment(s)

 * PyQtGraph version: Tested in 0.13.7 and 0.13.4
 * Qt Python binding: PySide6 6.6.2 Qt 6.6.2
 * Python version:  3.11.6
 * NumPy version: 2.0.1
 * Operating system:  Windows 11 10.0.22621
 * Installation method: pip

### Additional context
I also tested 0.13.3, but there the error does not occur.
That leads me to believe that I introduced this error by fixing #2887.
If I find the time I'll try to find a solution that hopefully fixes both problems in the next days.
But any help is appreciated.

UsYer:
After the fix for #2887 was introduced floating temp areas were not removed anymore when it was closed.
 This lead to a TypeError when calling saveState and then restoreState, because it tried to restore half-present floating windows.
Now the apoptose method of the temporary area of the floating window is called explicitly in its closeEvent method.
This leads to a duplicate call to dock_area.apoptose only in the case when a floating window is closed by calling restoreState on the original dock area. In that case an exception would be raised be removeArea, because the temporary dock_area was alread removed,
This is handled by setting the reference to the original dock area to None in the dock_areas apoptose method, which breaks the cycle.
Introduced regression tests for #2887 and #3125 as well as another potential failure case discovered while trying to fix the issue.
 Fixes #3125

### DIFF
--- a/pyqtgraph/dockarea/Dock.py
+++ b/pyqtgraph/dockarea/Dock.py
@@ -210,7 +210,7 @@ class Dock(QtWidgets.QWidget):
     def containerChanged(self, c):
         if self._container is not None:
             # ask old container to close itself if it is no longer needed
-            self._container.apoptose(propagate=False)
+            self._container.apoptose()
         self._container = c
         if c is None:
             self.area = None
--- a/pyqtgraph/dockarea/DockArea.py
+++ b/pyqtgraph/dockarea/DockArea.py
@@ -286,7 +286,11 @@ class DockArea(Container, QtWidgets.QWidget):
                 self.buildFromState(o, docks, obj, depth+1, missing=missing)
             # remove this container if possible. (there are valid situations when a restore will
             # generate empty containers, such as when using missing='ignore')
-            obj.apoptose(propagate=False)
+            # There are 2 cases where this container should be removed:
+            #   1. There are no child items which might happen in missing is set to 'ignore'
+            #   2. To remove a superfluous container (Another container as the sole children of this container)
+            if obj.count() == 0 or (obj.count() == 1 and isinstance(obj.widget(0), Container)):
+                obj.apoptose(propagate=False)
             obj.restoreState(state)  ## this has to be done later?     
 
     def findAll(self, obj=None, c=None, d=None):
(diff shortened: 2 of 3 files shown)


UsYer:
I found another case where preventing the propagation of apoptose calls leads to temporary areas not being closed and therefore breaking restoreState functionality.
If one drags the dock from the temporary area back into the main area the temporary window needs to close. 
But I now found a way to reintroduce the propagation by only calling the containers apoptose method if it is needed during restoreState.
I added a unittest for this case.

j9ac9k:
Thank you @UsYer for this PR, and patching up the dock functionality. This part of the library doesn't get as much attention as others, so it's always refreshing to know contributors are using it and even better, making contributions to fix it up.

I apologize for the long duration from when you made this PR to me following up, life just gets in the way... Anyway, merging!

MariusMeyerDraeger:
Thanks for reviewing and merging the PR.
I can imagine that the effort to maintain this package is a significant burden besides your regular life, so thanks a bunch for that!

PR #3136, opened 2024-08-26
allow conditional siPrefixing of axis labels

### LINKED ISSUE #3126: Request: limit upper/bottom units when auto SI-scaling
Hi folks,

First of all thanks a lot for all the work put into `pyqtgraph`, it is definitely one of the best graph libraries for Python/Qt out there.

### Short description

Currently it is possible to auto-scale axes using SI notation, which is very neat and useful.

It would be nice to be able to define bottom and upper units when auto-scaling SI units. 

Currently it goes all the way down to yocto and up to yotta:

https://github.com/pyqtgraph/pyqtgraph/blob/442f08a476644e7cac1babbb3076a767a265800f/pyqtgraph/functions.py#L62-L63

After that, it switches to scientific notation.

Depending on the application domain, it is interesting to specify a maximum bottom/upper unit that could be scaled to (for example micro/Mega), and after that switch to scientific notation.

### Additional context

I don't have a suggestion on how to accomplish this, but I would say ideally it would be nice to be able to configure this per-plot (or per-axis), as opposed to a global setting.

outofculture:
- siPrefixEnableRanges argument to AxisItem.setLabel and all pass-through callers thereof
- reduce duplication
- f-strings where % were encountered during development
- use ValueError when appropriate
- simplify logic
- closes #3126

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -16,7 +16,7 @@ __all__ = ['AxisItem']
 class AxisItem(GraphicsWidget):
     """
     GraphicsItem showing a single plot axis with ticks, values, and label.
-    Can be configured to fit on any side of a plot, 
+    Can be configured to fit on any side of a plot,
     Can automatically synchronize its displayed scale with ViewBox items.
     Ticks can be extended to draw a grid.
     If maxTickLength is negative, ticks point into the plot.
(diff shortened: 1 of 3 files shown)


j9ac9k:
readthedocs failure is due to the following:

```
Text in column margin in table line 10.

==============  =============================================================
**Arguments:**
text                   The text (excluding units) to display on the label for this
                       axis.
units                  The units for this axis. Units should generally be given
                       without any scaling prefix (eg, 'V' instead of 'mV'). The
                       scaling prefix will be automatically prepended based on the
                       range of data displayed.
unitPrefix             An extra prefix to prepend to the units.
siPrefixEnableRanges   The ranges in which automatic SI prefix scaling is enabled.
                       Defaults to everywhere, unless units is empty, in which case
                       it defaults to ((0, 1), (1e9, inf)).
args                   All extra keyword arguments become CSS style options for
                       the <span> tag which will surround the axis label and units.
==============  =============================================================
```

Would recommend switching to numpydoc style, some examples here: https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/PlotDataItem.py

j9ac9k:
Thanks @outofculture for the PR and the updating of the docstring format <3

PR #3137, opened 2024-08-27
Use Monospace font in Console Widget CmdInput

jack-mil:
The repl uses a monospace font (Courier New by default) to display the command output, but the text input widget was using the system default. This commit applies the same font styling to the LineEdit widget used in the Console widget.

before:
before

after:
after

This looks a lot more consistent in my opinion.

### DIFF
--- a/pyqtgraph/console/CmdInput.py
+++ b/pyqtgraph/console/CmdInput.py
@@ -11,8 +11,8 @@ class CmdInput(QtWidgets.QLineEdit):
         self.ps2 = "... "
         self.history = [""]
         self.ptr = 0
-        font = QtGui.QFont("Courier New", )
-        font.setStyleStrategy(QtGui.QFont.StyleStrategy.PreferAntialias)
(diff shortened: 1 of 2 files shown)


j9ac9k:
Hi @jack-mil 

Thanks for the PR! When specifying fonts by name for this, I'm always concerned about one of our systems not matching the font and erroring. Usually it's better to specify a style hint.

Do you get functional results with

```python
font = QFont()
font.setStyleHint(QtGui.QFont.StyleHint.Monospace, QtGui.QFont.StyleStrategy.PreferAntialias)
self.setFont(font)
```

Usually that should handle the intention across platforms more reliably.  Let me know if that works for you.

jack-mil:
will try is out, and also go ahead and change the hard coded font name for the repl widget if this works

j9ac9k:
Thanks for the PR and incorporating the feedback @jack-mil this LGTM!

PR #3140, opened 2024-08-29
Fix ConsoleWidget ReplWidget.write() always scrolling to bottom (`scrollToBottom` not working)

jack-mil:
Despite having `scrollToBottom='auto'` in `ReplWidget` set to auto, the TextEdit scroll area would always scroll to the bottom, to most recently inserted text. 
Desired behavior: have the scroll view stay in place if not at the bottom.
I messed around with lots of things trying to get it to work as expected. I found that moving the text cursor to the end also scrolled the view. There is no mention of this behavior in the QT docs. It does not seem to matter whether `QTextEdit.moveCursor()` is used, or first getting a reference, moving the cursor, and setting again using `QTextEdit.setCursor()` (as was being used here before. This issue was made especially hard to diagnose because the cursor is not visible. 

The `QTextEdit.insertPlainText()` method does work as expected, and moves the cursor to the end of the inserted text. This does *not* scroll the view (unlike `append()`).

I eventually settled on just removing the line that moves the cursor to the end near the beginning of ReplWidget.write(). This fixed the issue, and I only had to change the one other place the QTextEdit has text added to use insert instead of set in order to update the cursor position to the right place upon initialization (in `ConsoleWidget.__init__()`).

The other changes are just pre-commit auto formatting.

### DIFF
--- a/pyqtgraph/console/Console.py
+++ b/pyqtgraph/console/Console.py
@@ -1,12 +1,12 @@
 import os
-import sys
 import pickle
 import subprocess
+import sys
 
 from .. import getConfigOption
 from ..Qt import QtCore, QtWidgets
-from .repl_widget import ReplWidget
 from .exception_widget import ExceptionHandlerWidget
+from .repl_widget import ReplWidget
 
 
 class ConsoleWidget(QtWidgets.QWidget):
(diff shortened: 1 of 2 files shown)


j9ac9k:
Thank you for the PR @jack-mil this LGTM, merging!

PR #3145, opened 2024-09-08
connect darkMode slots only once

### LINKED ISSUE #3143: Dark/Light mode change will no longer work in PySide6 > 6.7.2
In `mkQApp()`, a `UniqueConnection` is made to a free-standing function to handle light and dark modes.
https://github.com/pyqtgraph/pyqtgraph/blob/20bc47ecabe659ffe83e3cc88c560091882473aa/pyqtgraph/Qt/__init__.py#L348

This will no longer work in PySide6 > 6.7.2, as `UniqueConnection` will only be allowed on methods of `QObject` instances.

An example to demonstrate the change in behavior.
On PySide6 6.7.2, "HELLO" will be printed.
On development versions of PySide6, "HELLO" _does not_ get printed.
```python
from PySide6 import QtCore, QtWidgets

def name_changed(name):
    print('name_changed to', name)

# conntype = QtCore.Qt.ConnectionType.AutoConnection
conntype = QtCore.Qt.ConnectionType.UniqueConnection
app.objectNameChanged.connect(name_changed, type=conntype)
app.objectNameChanged.connect(name_changed, type=conntype)
app.dumpObjectInfo()
app.setObjectName('HELLO')
```

pijyoi:
fixes #3143

In PySide6 > 6.7.2, it is no longer possible to use `UniqueConnection` with free-standing functions. The end result of supplying that mode is that no connection will be made at all.

This can be worked-around by manually keeping track of whether a connection has already been made.

In fact, the original code has a recursion issue. In the slot, `mkQApp()` is called, which then attempts to connect the slot again, relying on `UniqueConnection` to mask out this recursion.

### DIFF
--- a/pyqtgraph/Qt/__init__.py
+++ b/pyqtgraph/Qt/__init__.py
@@ -345,6 +345,7 @@ App = QtWidgets.QApplication
 # subclassing QApplication causes segfaults on PySide{2, 6} / Python 3.8.7+
 
 QAPP = None
+_pgAppInitialized = False
 def mkQApp(name=None):
     """
     Creates new QApplication or returns current instance if existing.
@@ -355,6 +356,7 @@ def mkQApp(name=None):
     ============== ========================================================
     """
     global QAPP
+    global _pgAppInitialized
 
     QAPP = QtWidgets.QApplication.instance()
     if QAPP is None:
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for fixing this @pijyoi a global is the right solution here for sure.

PR #3152, opened 2024-09-17
prevent grid lines from getting mouse events

### LINKED ISSUE #2062: ScatterPlotItem right click and ScatterPlotItem(listOfPoints) no longer works between pyqtgraph 0.11 and 0.12
Hello guys,

I have upgraded to pyqtgraph 0.12 from 0.11 and i've noticed behavioral differences between versions.

```python
def plotObjects(self):
        objects= list()
        for objectId in range(OBJ_NUMBER):
                x = getX
                y = getY
		measured = getMeasured()
                if abs(x) > float(0.0) and abs(y) > float(0.0):
		obj_info = {'pos': (fus_dy, fus_dx),
                	    'size': FUS_OBJECT_SYMBOL_SIZE,
                            'symbol': FUS_OBJECT_SYMBOL,
                            'brush': pg.mkBrush(QColor(color_set['OBJ_COLOR'])),
                            'pen': pg.mkPen(QColor(color_set['OBJ_COLOR'])),
                            'data': f"FUS_{objectId}"}
                    if measured == 1:
			obj_info['brush'] = pg.mkBrush(None)

                    objects.append(obj_info)
                    # add text item that indicates object id
                    label = pg.TextItem(text=str(objectId), color=color_set['OBJ_COLOR'], anchor=([1, 1]))
                    label.setPos(x, y)
                    self.plotInstance.addItem(label)

        scatterPlotItem = pg.RightClickableScatterPlotItem(objects)
        self.plotInstance.addItem(scatterPlotItem)
        scatterPlotItem.sigClicked.connect(self.point_clicked)
```

RightClickableScatterPlotItem is below.

```python
class RightClickableScatterPlotItem(pg.ScatterPlotItem):
    sigRightClicked = pg.QtCore.Signal(object, object)

    def mouseClickEvent(self, ev):
        if ev.button() == pg.QtCore.Qt.RightButton:
            pts = self.pointsAt(ev.pos())
            if len(pts) > 0:
                self.ptsClicked = pts
                ev.accept()
                self.sigRightClicked.emit(self, self.ptsClicked)
            else:
                ev.ignore()
        else:
            super().mouseClickEvent(ev)
```

This used to plot different points with their separate brushes.
But now it plots all points with the last brush used.
I've noticed this does not happen if pxMode=True but then the points look all wrong.

How do i fix this?

The second problem i've noticed is that the right-click overwrite no longer works.
That last line (scatterPlotItem.sigClicked.connect(self.point_clicked)) used to be trigger the emit to this:

```python
    def point_right_clicked(self, item, ptsClicked):
        # Assume that we click only an object once
        # e.g. there is exactly one object under our exact cursor position
        assert(len(ptsClicked) == 1)

        x_intersect = np.argwhere(item.data['x'] == ptsClicked[0].pos().x())
        y_intersect = np.argwhere(item.data['y'] == ptsClicked[0].pos().y())
        point_index = np.intersect1d(x_intersect, y_intersect)[0]

        fus_index = item.data[point_index]['data']

        selected_points = {}
        selected_points['id'] = index

        self.menu.popupWithContext(selected_points)
```

which takes the list of points that is emitted and tries to figure out which one is clicked.

Anyway now I cannot get to trigger the Right-Click on the object.
It shows the background menu.

"View All, 
Y Axis,
X Axis,
Mouse Mode
etc..."

It does work on Left-Click on the ScatterPlotItem if i replace

if ev.button() == pg.QtCore.Qt.RightButton:
        
with 

if ev.button() == pg.QtCore.Qt.LeftButton:
                
How to fix this issue?

Thanks for the help and the great work on this amazing library btw :)

pijyoi:
fixes #2062

Override `shape()` so that grid lines are not considered for the purposes of collision detection / mouse event delivery.

Grid lines are drawn by `AxisItem` as part of an extended tick line. As `AxisItem`s are given a zValue of +0.5 vs `ViewBox`'s  _negative_ 100, children of `ViewBox` get stacked underneath `AxisItem`s.
When grid lines are enabled, this has the side effect of mouse events getting delivered to the `AxisItem` before any other items.

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -660,6 +660,13 @@ class AxisItem(GraphicsWidget):
         else:
             return self.mapRectFromParent(self.geometry()) | linkedView.mapRectToItem(self, linkedView.boundingRect())
 
+    def shape(self):
+        # override shape() to exclude grid lines from getting mouse events
+        rect = self.mapRectFromParent(self.geometry())
+        path = QtGui.QPainterPath()
+        path.addRect(rect)
+        return path
+
     def paint(self, p, opt, widget):
         profiler = debug.Profiler()
         if self.picture is None:


j9ac9k:
Thanks for this fix @pijyoi nice work thinking up a minimally invasive fix to a problem that would otherwise be really difficult to address!

PR #3158, opened 2024-10-01
Minor fixes for a memory debugging tool

campagnola:
.. that probably doesn't belong in this library, but it's super handy when you need it (which is almost always in projects that use pyqtgraph)


j9ac9k:
you're right, it shouldn't be in this library :P but as it already is, fixes definitely appreciated!

PR #3160, opened 2024-10-08
Parameter Tree: ListParameter: Add deprecation warning when values is present.

Tillsten:
The silent change to ignore `values` instead of `limis` has bitten me. This adds a deprecation warning and will allow `values` to still work for the next release.


j9ac9k:
Thanks for the PR @Tillsten I'm sorry the change was made silently, we try and avoid this but that snuck in.  Merging.

PR #3161, opened 2024-10-08
Adding support for connect PlotItem connect vector during view clippng and downsampling

tblum:
When using the connect vector argument to plot (PlotItem) one can not use ClitToView or Downsampling. Using them will cause an Exception (ValueError): 

E.g running this code and activating "clip to view" or downsampling via right click menu:
```
import numpy as np
import pyqtgraph as pg

x = np.linspace(0.0, 7.0, num=10000)
x = np.concatenate((x[:3000], x[7000:]))
y = np.sin(x)
c = np.ones(len(x), dtype=bool)
c[2999] = False
pg.plot(x,y, connect=c)

if __name__ == '__main__':
    pg.exec()
```
This PR "fixes" that.
Clipping works as expected.
Down-sampling will no linger cast exception. Don't know if it is the best strategy for down-sampling the connect vector. But I didn't want to make the calculations too expensive.  

Best regards
Troels Blum

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -93,13 +93,15 @@ class PlotDataset:
         x: np.ndarray,
         y: np.ndarray,
         xAllFinite: bool | None = None,
-        yAllFinite: bool | None = None
+        yAllFinite: bool | None = None,
+        connect: np.ndarray | None = None
     ):
         super().__init__()
         self.x = x
         self.y = y
         self.xAllFinite = xAllFinite
         self.yAllFinite = yAllFinite
+        self.connect = connect
         self._dataRect = None
 
         if isinstance(x, np.ndarray) and x.dtype.kind in 'iu':
@@ -1345,6 +1347,8 @@ class PlotDataItem(GraphicsObject):
 
         x = dataset.x
         y = dataset.y
+        if dataset.connect is not None:
+            curveArgs['connect'] = dataset.connect
         #scatterArgs['mask'] = self.dataMask
         if (
             self.opts['pen'] is not None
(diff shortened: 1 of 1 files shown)


tblum:
I also create a patch for version 13.7 since that is the version installed by pip (and that I am using)
Don't know if that is useful?
[pyqtgraph_connect.patch](https://github.com/user-attachments/files/17293225/pyqtgraph_connect.patch)

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -93,13 +93,15 @@ def __init__(
         x: np.ndarray,
         y: np.ndarray,
         xAllFinite: bool | None = None,
-        yAllFinite: bool | None = None
+        yAllFinite: bool | None = None,
+        connect: np.ndarray | None = None
```
j9ac9k, line 97: Add a blurb in the doc-string above (line 81) about this parameter please.

j9ac9k:
Hi @tblum 

Thanks for this PR, this change LGTM. 

Can you add a test in [`test_PlotDataItem.py`](https://github.com/pyqtgraph/pyqtgraph/blob/master/tests/graphicsItems/test_PlotDataItem.py) that will demonstrate this functionality?  There is a `test_clipping` function there already, maybe would benefit from parameterizing `x`, `y` and `connect` arguments. Let mek now if you would need some guidance.

Thanks for the PR!

tblum:
Updated the doc-string and added a test based on the example I posted.

tblum:
@j9ac9k Am I missing something, or is this good to go?

j9ac9k:
Hi @tblum 

Thanks for the PR, sorry this has taken me so long to get to it.  This LGTM, merging!

Issue #3154, opened 2024-09-19
Features request: append and update last data


Khanhlinhdang:
Hello, I am using pyqtgraph to develop my live trading project.
It has a problem.
when update to the last data or append new data, I have to re-use function setData. it will be re-caculate all again. it made the code to be slowly.
can you add new function like update and append data.
I see this function in PlotdataItem.py but it is not working.
![image](https://github.com/user-attachments/assets/0999debb-a73d-484a-8067-6fbc26e140dd)
Thank you very much for your awesome library

psyuktha:
A possible approach to improve performance involves direct data manipulation. Thus I can create a class with custom update and append methods. 
Can I work on this issue , please let me know if this approach is fine.

Khanhlinhdang:
Add new data and update the last data without recaculate all data. And I am very happy if you can do that.

NilsNemitz:
Hi!
We have gotten the suggestion to add a function to append data multiple times, and the devs agree that this would be a nice addition for convenience. That's why there is a stub for adding it already.

But it seems hard to find a solution that works with good performance across different applications. I tried to write a version with some pre-allocated buffer space in #1453, but it wasn't particularly elegant. My understanding is that appending to the internal numpy array or concatenating arrays needs to allocate new memory for all data every time, and that becomes slow with large datasets.

The usual advice is to try to move the optimization to user space:
When `setData()` is called with a (float format) numpy array, very little processing of data is actually required and plotting is fast. You might want to look at the "Line Plot update" benchmark in the examples application. If your application moves significantly slower than that, then your original data is probably in some different format that takes a while to convert. 

If that's the case, then you should be able to speed up plotting by maintaining your own "buffer" numpy array for the plot data and converting your new points to that format only once when you "append" them. That also gives you the option of updating just the last points. Then you pass the entire array to `setData()`, but since it is already in the right format, that's fast.

That's almost exactly the code that @psyuktha has kindly submitted in the PR. So you should be able to look there for inspiration!

Khanhlinhdang:
Thank you so much @psyuktha and thank for your advice @NilsNemitz 
I will test the new update and compare with origin function setData. And I will reply the result here!

PR #3164, opened 2024-10-13
Add PlotDataItem.appendData

psyuktha:
Added methods appendData and updateData as referred to the issue #3154

- appendData: Appends data without clearing and recalculating the plot. This is useful for real-time data plotting where you are continuously adding new points.

- updateData: Replaces the entire dataset if needed, which triggers a full plot update (similar to the original setData behavior).

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -1741,13 +1741,17 @@ class PlotDataItem(GraphicsObject):
         xData = np.concatenate((self._dataset.x, new_x))
         yData = np.concatenate((self._dataset.y, new_y))
 
-        # Update the dataset
-        self._dataset = PlotDataset(xData, yData)
+            # Update the internal dataset with the concatenated arrays
+        self._dataset.x = xData
+        self._dataset.y = yData
         
-        # Update display data
+        # Invalidate any display data cache
         self._datasetDisplay = None  # Invalidate display data
-        self.updateItems(styleUpdate=True)  # Update items
 
+        # Trigger a visual update of the plot without recalculating everything
+        self.updateItems(styleUpdate=False)  # Only update items, not recalculate all data
+
+        
     def updateLastData(self, new_x, new_y):
         """
         Update the last data point in the dataset.
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -1744,14 +1744,16 @@ class PlotDataItem(GraphicsObject):
             # Update the internal dataset with the concatenated arrays
         self._dataset.x = xData
         self._dataset.y = yData
-        
         # Invalidate any display data cache
-        self._datasetDisplay = None  # Invalidate display data
 
+        self._datasetMapped  = None
+        self._datasetDisplay = None  # Invalidate display data
+        self._adsLastValue   = 1
+        
         # Trigger a visual update of the plot without recalculating everything
         self.updateItems(styleUpdate=False)  # Only update items, not recalculate all data
 
-        
+
     def updateLastData(self, new_x, new_y):
         """
         Update the last data point in the dataset.
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -1719,18 +1719,7 @@ class PlotDataItem(GraphicsObject):
 
     def appendData(self, new_x, new_y):
         """
-        Append new data to the existing dataset.
-
-        .. warning::
-        This method may be significantly slower than calling `setData`, as it 
-        appends to the current data rather than replacing it. For high-performance
-        applications, consider using `setData` when possible.
-
-        .. warning::
-            This method will be undergoing further changes in future versions. 
-            Currently, the accepted arguments are only a subset of those for `setData`.
-            Ensure compatibility by using only the supported arguments.
-    
+        Append new data to the existing dataset.    
         
         Parameters
         ----------
@@ -1738,6 +1727,16 @@ class PlotDataItem(GraphicsObject):
             The new x-values to append.
         new_y : array-like
             The new y-values to append.
+
+        Warning
+        -------
+        This method may be significantly slower than calling `setData`, as it 
+        appends to the current data rather than replacing it. For high-performance
+        applications, consider using `setData` when possible.
+
+        This method will be undergoing further changes in future versions. 
+        Currently, the accepted arguments are only a subset of those for `setData`.
+        Ensure compatibility by using only the supported arguments.
         
         """
         if self._dataset is None:
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -1752,17 +1752,16 @@ class PlotDataItem(GraphicsObject):
         xData = np.concatenate((self._dataset.x, new_x))
         yData = np.concatenate((self._dataset.y, new_y))
 
-            # Update the internal dataset with the concatenated arrays
+        # Update the internal dataset with the concatenated arrays
         self._dataset.x = xData
         self._dataset.y = yData
-        # Invalidate any display data cache
 
+        # Invalidate any display data cache
         self._datasetMapped  = None
-        self._datasetDisplay = None  # Invalidate display data
+        self._datasetDisplay = None  
         self._adsLastValue   = 1
         
-        # Trigger a visual update of the plot without recalculating everything
-        self.updateItems(styleUpdate=False)  # Only update items, not recalculate all data
+        self.updateItems(styleUpdate=False)  
 
     @QtCore.Slot(object, object)
     def curveClicked(self, _: PlotCurveItem, ev):
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -1716,8 +1716,6 @@ class PlotDataItem(GraphicsObject):
         self.curve.clear()
         self.scatter.clear()
 
-    # def appendData(self, *args, **kwargs):
-    #     pass
 
     def appendData(self, new_x, new_y):
         """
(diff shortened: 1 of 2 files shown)


Khanhlinhdang:
thank you @psyuktha , I will test its peformance. and talk you latter.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -1716,8 +1716,62 @@
         self.curve.clear()
         self.scatter.clear()
 
-    def appendData(self, *args, **kwargs):
-        pass
+    # def appendData(self, *args, **kwargs):
+    #     pass
```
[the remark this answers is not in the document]
psyuktha: Should I remove the commented out code?

Khanhlinhdang:
@psyuktha can you add example for appendData function, it is seem not working. updateData function is work well

psyuktha:
@Khanhlinhdang You can append new data to the plot like this:

Assuming plot_item is an instance of PlotDataItem

new_x_value = [10, 11, 12]  # Example new X values
new_y_value = [5, 6, 7]     # Example new Y values

plot_item.appendData(new_x_value, new_y_value)

Khanhlinhdang:
Thanks @psyuktha 
What are type of the inputs can be? a list of float or numpy.array?or both of them?

psyuktha:
Both,
After conversion, both lists and NumPy arrays are handled the same way internally, as np.array ensures uniform data types for further operations.

Khanhlinhdang:
@psyuktha 
I have test some options and when I add some line into function AppendData, it work well. 
`def appendData(self, new_x, new_y):
        """
        Append new data to the existing dataset.
        
        Parameters
        ----------
        new_x : array-like
            The new x-values to append.
        new_y : array-like
            The new y-values to append.
        """
        if self._dataset is None:
            # If there is no existing dataset, set the new data as the dataset.
            self.setData(new_x, new_y)
            return
        new_x = np.array(new_x)
        new_y = np.array(new_y)
        # Concatenate new data to existing dataset
        self._dataset.x = np.concatenate((self._dataset.x, new_x))
        self._dataset.y = np.concatenate((self._dataset.y, new_y))
        self._datasetMapped  = None
        self._datasetDisplay = None
        self._adsLastValue   = 1`

psyuktha:
Thank you , required changes are done

psyuktha:
@campagnola can you please review the changes and merge the PR

psyuktha:
@j9ac9k could you please review the changes done?

j9ac9k:
Hi @psyuktha 

There are a couple of issues on my mind regarding this PR.

1. `updateLastData` I'm convinced is not needed. If you can update a value, you can wrap that value in an array and call `appendData` with the same arguments. 
2. the `appendData` is not accepting of all the inputs that `setData` accepts.  To be fair `setData` accepts arguments in all sorts of formats/containers, which if we want to be able to call `appendData`, and ideally those should match up.  This might be a bigger undertaking than you might think as `setData` accepts data in so many different structures.
3. In the docstring, there should be a warning about the performance, `appendData` in the current form would be _significantly_ slower than calling `setData`.  I'm ok with not trying to resolve this performance issue in this PR, but there should be a note/warning about that.

j9ac9k:
On further reflection, I think I would be ok merging this with

1. Remove `updateLastData`, again, not needed
2. Add a warning in the docstring for `appendData` that this method may be significantly slower than calling `setData`
3. Add a warning in the docstring that this method will be undergoing further changes, and the accepted arguments are only a subset of those for `setData`

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -1716,8 +1716,54 @@ def clear(self):
...
+        # Convert new data to arrays
+        new_x = np.array(new_x)
+        new_y = np.array(new_y)
+
+        # Concatenate new data to existing dataset
+        xData = np.concatenate((self._dataset.x, new_x))
+        yData = np.concatenate((self._dataset.y, new_y))
+
+            # Update the internal dataset with the concatenated arrays
+        self._dataset.x = xData
```
NilsNemitz, line 1757: Wouldn't it be more robust to just call `setData()` with the new arrays?
Currently it isn't entirely clear to me if the remaining code in `setData()`, such as emitting "sigPlotChanged()" are safe to omit in this case.

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -1716,8 +1716,54 @@ def clear(self):
...
+        self._dataset.y = yData
+        # Invalidate any display data cache
+
+        self._datasetMapped  = None
+        self._datasetDisplay = None  # Invalidate display data
+        self._adsLastValue   = 1
+        
+        # Trigger a visual update of the plot without recalculating everything
+        self.updateItems(styleUpdate=False)  # Only update items, not recalculate all data
```
NilsNemitz, line 1766: If we do replicate the code here, I am not so sure about the comments on "without recalculating". There's a lot of code that does need to run on the new data, e.g. log scale conversion and range checking.

To me it seems that the main selling point of `appendData()` is that it is possible to feed in updates from some complex data structure (maybe from pandas?) piece by piece, rather than converting the entire thing all the time.

psyuktha: That makes sense! Emphasizing the method's ability to handle incremental updates from complex data structures like pandas DataFrames while avoiding any implication of performance optimizations would create clearer expectations for users.
I have revised the comments accordingly

pyqtgraph/graphicsItems/PlotDataItem.py
psyuktha: Calling setData() would indeed ensure consistency by running all the additional code, like sigPlotChanged() and other updates that setData handles. However, in this case, as requested by @Khanhlinhdang  in issue #3154 the purpose of appendData() is to optimize for performance, especially for live data scenarios where frequent updates are required.

That said, appendData might not fully replicate all internal updates from setData for certain plot settings, so users should still rely on setData() when working with static datasets or when requiring a complete refresh.

Thus I have revised the docstring accordingly by adding the warnings as suggested by @j9ac9k

NilsNemitz: How much performance improvement are you seeing by NOT going through `setData()`?

pyqtgraph/graphicsItems/PlotDataItem.py
```diff
@@ -1716,8 +1716,52 @@ def clear(self):
...
+        # Update the internal dataset with the concatenated arrays
+        self._dataset.x = xData
+        self._dataset.y = yData
+
+        # Invalidate any display data cache
+        self._datasetMapped  = None
+        self._datasetDisplay = None  
+        self._adsLastValue   = 1
+        
+        self.updateItems(styleUpdate=False)
```
NilsNemitz, line 1764: `self._datasetDisplay` is the cache for the processed data.

If you set that to `None`, doesn't the call to `updateItems()` go to `_getDisplayDataset()` and refreshes the cache by recalculating everything?

Could you tell me which part of the processing you are avoiding?

psyuktha: We still need to clear the display cache (self._datasetDisplay) to accommodate new data points right?

NilsNemitz: Yes.

Your first post said "without clearing and recalculating the plot", I am still trying to figure out where this manages to skip something that might be slow in a real-time application.

The automatic refresh of the cleared cache seems like it is still processing the entire dataset, after all. And I think the current design of the `PlotDataItem` internals do need that.

psyuktha: In live data scenarios, recalculating everything with each update can indeed slow things down, especially if data transformations (like log scaling or other range calculations) are re-applied each time.

pyqtgraph/graphicsItems/PlotDataItem.py
NilsNemitz: Oh, maybe we are just misunderstanding each other?

I meant to say that you could call 'setData()' *here*, in line 1757 of your PR. That would avoid duplicating the cache clears etc. 

That then runs all the processing inside of pyqtgraph as usual. But the PR is not trying to avoid that. The PR is trying to help reduce processing *on the user side*. Do we agree on this? Or am I still getting it wrong?

psyuktha: Yes, its clear

psyuktha:
@j9ac9k I have added the warnings and removed UpdateLastData()

pyqtgraph/graphicsItems/PlotDataItem.py
NilsNemitz: Can I then once again suggest to replace L 1756 and the following lines with
`self.setData(xData, yData)` 
?
Would that not achieve the same thing with less duplication of code?

pijyoi:
I have a script here that makes use of this `appendData` method with live data from the sound card.
The longer the program runs, the longer the data set, and eventually the program hangs and has to be forcibly terminated.

Normally for such live acquisition scenarios, the number of samples displayed would need to be capped, showing only the newest N number of samples.

Notice also that in the script below, the user still needs to keep track of the total number of samples.
```python
from PySide6 import QtMultimedia
import pyqtgraph as pg
import numpy as np

pg.mkQApp()
plt = pg.PlotWidget()
plt.show()
curve = plt.plot()

num_samples = 0

def audio_readable():
    global num_samples
    data = qiodev.readAll()
    y = np.frombuffer(data, dtype=np.float32)
    x = np.arange(num_samples, num_samples + len(y))
    curve.appendData(x, y)
    num_samples += len(y)
    framecnt.update()

fmt = QtMultimedia.QAudioFormat()
fmt.setChannelCount(1)
fmt.setSampleRate(44100)
fmt.setSampleFormat(QtMultimedia.QAudioFormat.SampleFormat.Float)

ainput = QtMultimedia.QMediaDevices.defaultAudioInput()
ainput = QtMultimedia.QAudioSource(fmt)
qiodev = ainput.start()
qiodev.readyRead.connect(audio_readable)

framecnt = FrameCounter()
framecnt.sigFpsUpdate.connect(lambda fps: plt.setTitle(f'{fps:.1f} fps'))

pg.exec()
```

PR #3168, opened 2024-10-16
Update LinearRegionItem.py

dingo9:
fix bug in Examples/InfiniteLine.py
```
lr = pg.LinearRegionItem(values=[70, 80])
p1.addItem(lr)
label = pg.InfLineLabel(lr.lines[1], "region 1", position=0.95, rotateAxis=(1,0), anchor=(1, 1))
```
if change ```lr.lines[1]``` to ```lr.lines[0]```,  label moved in drag event. Don't block InfiLine Signal directly while draging event?

### DIFF
--- a/pyqtgraph/graphicsItems/LinearRegionItem.py
+++ b/pyqtgraph/graphicsItems/LinearRegionItem.py
@@ -320,11 +320,11 @@ class LinearRegionItem(GraphicsObject):
         if not self.moving:
             return
             
-        self.lines[0].blockSignals(True)  # only want to update once
+        self.blockLineSignal = True  # only want to update once
         for i, l in enumerate(self.lines):
             l.setPos(self.cursorOffsets[i] + ev.pos())
-        self.lines[0].blockSignals(False)
         self.prepareGeometryChange()
+        self.blockLineSignal = False
         
         if ev.isFinish():
             self.moving = False


ixjlyons:
Note this looks like it aims to fix #3188.

I'll say I'm not too familiar with the events handling in `LinearRegionItem`, so I'd like to take a bit of a closer look, but it does seem to fix the problem.

Minimal example:

```python
import pyqtgraph as pg

pl = pg.plot([0, 1, 2, 3], [4, 5, 6, 7])

lr = pg.LinearRegionItem(values=[0, 1])
pl.addItem(lr)
pg.InfLineLabel(lr.lines[0], "l0: {value:.2f}")
pg.InfLineLabel(lr.lines[1], "l1: {value:.2f}")

pl.show()
pg.exec()
```

Note also without this change, the label on line 0 can move vertically, until you trigger an event by e.g. panning/zooming, after which it snaps back to the middle.

ixjlyons (approved):
I reviewed things a little more carefully and I still think this looks fine. I can't quite explain why it might've been just blocking the first `InfiniteLine` altogether, because that definitely breaks the connection between line and label while the region is being dragged. Furthermore, I would think other (e.g. user) connections to the `InfiniteLine` position change signals should still work while the region is dragged. This only blocks those events from reaching back up to the region level to avoid triplet region change events.

j9ac9k:
Thanks for the PR @dingo9 and thanks for the review @ixjlyons  This LGTM, merging!

PR #3172, opened 2024-10-21
Fix LabelItem Y-axis inversion issue

Khagendra01:


j9ac9k:
Is this PR just to add a file to the top level directory? It should go somewhere else.  Also probably should make use of `pg.mkQApp()` and `pg.exec()` instead of the standard Qt calls to prevent starting multiple instances of QApplication.

PR #3173, opened 2024-10-21
ENH: Ignore PySide6 warning

larsoner:
Swallow this unnecessary warning that happens when closing a window on latest PySide6:
```
/home/larsoner/python/pyqtgraph/pyqtgraph/graphicsItems/ViewBox/ViewBox.py:1812: RuntimeWarning: Failed to disconnect (NoneType) from signal "destroyed()".
  k.destroyed.disconnect()
```
There are a bunch of unrelated whitespace removals that I can undo if necessary. Also happy to open a separate PR that enforces something like `black`. (Most scientific Python projects I work on use something like this nowadays!)

### DIFF
--- a/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
+++ b/pyqtgraph/graphicsItems/ViewBox/ViewBox.py
@@ -2,6 +2,7 @@ import math
 import sys
 import weakref
 from copy import deepcopy
+import warnings
 
 import numpy as np
 
@@ -164,12 +165,12 @@ class ViewBox(GraphicsWidget):
             'wheelScaleFactor': -1.0 / 8.0,
 
             'background': None,
-            
+
             'logMode': [False, False],
 
             # Limits
             # maximum value of double float is 1.7E+308, but internal caluclations exceed this limit before the range reaches it.
-            'limits': { 
+            'limits': {
                 'xLimits': [-1E307, +1E307],   # Maximum and minimum visible X values
                 'yLimits': [-1E307, +1E307],   # Maximum and minimum visible Y values
                 'xRange': [None, None],   # Maximum and minimum X range
@@ -472,7 +473,7 @@ class ViewBox(GraphicsWidget):
 
             self.background.setRect(self.rect())
             self.borderRect.setRect(self.rect())
-        
+
             self.sigStateChanged.emit(self)
             self.sigResized.emit(self)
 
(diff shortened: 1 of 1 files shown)


pijyoi:
Do you have a reproducing example?
I don't get the error by running examples/SimplePlot.py with PySide 6.8.0.1 (on Windows)

#2923 (present since 0.13.4) was supposed to have addressed this.

If the following block didn't catch the condition, that would mean that the object is not alive? (Your pasted error message says "NoneType")
https://github.com/pyqtgraph/pyqtgraph/blob/5c8ecb5a7e4679322d8d257fc7be921fa7ddb731/pyqtgraph/graphicsItems/ViewBox/ViewBox.py#L1799-L1806

By the way, #2923 didn't go the route of suppressing the warning because it was not possible at the time (PySide 6.7.x). It was a message that would get printed out unconditionally to stderr. Perhaps that has changed in later versions of PySide6.

larsoner:
> Do you have a reproducing example?

Not exactly minimal but if you `pip install mne mne-qt-browser` you should be able to do this (just replicated on PySide 6.7.3 from conda and pyqtgraph 0.13.4 on macOS arm64, previous tests were on Linux):
```Python
>>> import mne                                                                                                                                                  
>>> import numpy as np                                                                                                                               
>>> raw = mne.io.RawArray(np.zeros((1, 1000)), mne.create_info(1, 1000., "eeg"))
>>> raw.plot()                                                                                                                                                  
<mne_qt_browser._pg_figure.MNEQtBrowser(0x600003704de0) at 0x144d72fc0>                                                                                         
>>> /Users/larsoner/python/mne-qt-browser/mne_qt_browser/_pg_figure.py:3316: RuntimeWarning: Failed to disconnect (None) from signal "triggered()".             
  sig.disconnect()                                                                                                                                              
>>> exit()                                                                                                                                                          
/Applications/MNE-Python/1.7.0_0/.mne-python/lib/python3.12/site-packages/pyqtgraph/graphicsItems/ViewBox/ViewBox.py:1798: RuntimeWarning: Failed to disconnect (NoneType) from signal "destroyed()".                                                                                                                           
  k.destroyed.disconnect()                                                                                                                                      
QObject::startTimer: Timers can only be used with threads started with QThread  
```
I manually closed the window before the `exit()` call. So it happens when Python is exiting.

larsoner:
No idea what the underlying issue is, but it would be nice not to get these messages somehow!

j9ac9k:
Thanks for the PR @larsoner 

I'm reluctant to merge this, as this warning message is usually an indication that there is a problematic reference somewhere, which we take effort to minimize when we see this with library issue.   That said, emitting a more helpful warning message would be worthwhile.

j9ac9k:
In the interest of getting the outstanding PR count down, I'm going to close this w/o merging.  The issue isn't in pyqtgraph, and suppressing this warning would make debugging issues for pyqtgraph users more difficult.  A PR to give more context and helpful information in the warning to help debug would certainly be appreciated.

Issue #2921, opened 2024-01-28
AttributeError: 'DockLabel' object has no attribute 'pressPos'


andrewchenshx:
### Short description
mouse over a docklabel, cause this error:
`AttributeError: 'DockLabel' object has no attribute 'pressPos'`

### Expected behavior
no error
### Real behavior

```
Traceback (most recent call last):
  File "C:\Users\Administrator\py_envs\py310\lib\site-packages\pyqtgraph\dockarea\Dock.py", line 345, in mouseMoveEvent
    self.mouseMoved = (lpos - self.pressPos).manhattanLength() > QtWidgets.QApplication.startDragDistance()
AttributeError: 'DockLabel' object has no attribute 'pressPos'
```

### Tested environment(s)

 * PyQtGraph version: 0.13.3
 * Qt Python binding: PyQt6 6.6.1 Qt 6.6.1
 * Python version: Python 3.10.11
 * NumPy version: 1.26.3
 * Operating system: Windows10 60bit
 * Installation method: pip

### Additional context
before i converted the project to pyqt6, there is no error. the env is as:
 * PyQtGraph version: 0.13.3
 * Qt Python binding: PyQt5 5.15.10 Qt 5.15.2
 * Python version: Python 3.9.13
 * NumPy version: 1.24.4
 * Operating system: Windows10 60bit
 * Installation method: pip

andrewchenshx:
self.pressPos is read within mouseMoveEvent(), and is written within mousePressEvent().
if i first press the docklabel which i think will trigger the mousePressEvent(), then subsequent mouseover will NOT cause this error.

Studemann:
I have this issue under Linux Debian but not under Windows (?).
I am not an expert in Python but to solve it, I have initialized the "pressPos" attribut in the def __init__(...) method with a initial point:

self.pressPos = QtCore.QPointF(0.0,0.0)

Now the error does not appear anymore.
Maybe it could help someone ...

seb5g:
indeed I solved it like this, I'll do a PR for this

PR #3177, opened 2024-10-29
initialize the pressPos attribute with default value to prevent error

seb5g:
This will solve #2921

The pressPos attribute of the DockLabel object is not initialized in the __init__. Therefore when the mouse is hovering on a DockLabel it triggers an error. Adding a default value for this attribute in the __init__ method solves it.


j9ac9k:
Thanks for the PR @seb5g !

PR #3180, opened 2024-10-31
add setSymbol method

kevinanewman:
Add setSymbol method to TargetItem

Allows symbol to be updated after TargetItem is created, which can be used to create user-editable or dynamic targets.

Updated error message for unknown string type symbols.


pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -106,19 +106,35 @@ def __init__(
...
+            :func:`~pyqtgraph.ScatterPlotItem.setSymbol`
+
+        """
         if isinstance(symbol, str):
             try:
-                self._path = Symbols[symbol]
+                path = Symbols[symbol]
             except KeyError:
-                raise KeyError("symbol name found in available Symbols")
+                raise KeyError("Symbol name '%s' not found in available Symbols" % symbol)
```
j9ac9k, line 131: nit-pick, can we use a f-string here instead of the `%s` syntax?

pyqtgraph/graphicsItems/TargetItem.py
```diff
@@ -106,19 +106,35 @@ def __init__(
...
         self.setPos(pos)
 
+        self._path = None
+        self.setSymbol(symbol)
+
+        self.scale = size
+        self.setLabel(label, labelOpts)
+
+    def setSymbol(self, symbol):
+        r"""Method to set the TargetItem symbol, during or after creation
```
j9ac9k, line 116: Fairly sure this does not need to be a raw string.

j9ac9k:
Thanks for the PR, this LGTM; I just had a couple of nit-picks which I commented.

pyqtgraph/graphicsItems/TargetItem.py
j9ac9k: Also this docstring should probably have the `Raises` section to specify the KeyError that is raised if the symbol isn't found.

pyqtgraph/graphicsItems/TargetItem.py
kevinanewman: I've revised this in my latest commit

pyqtgraph/graphicsItems/TargetItem.py
kevinanewman: I made it a raw string since that seemed to be how some of the other docstrings were formatted and was for consistency, I've made these changes in my latest commit.

kevinanewman:
Latest commit available with updates per your feedback, thank you!

j9ac9k:
Thanks @kevinanewman for updating the PR, sorry it's taken me so long to get to this!

PR #3182, opened 2024-11-04
Fix bug in PlotItem.setAxisItem when called befor scene is set

tblum:
Calling PlotItem.setAxisIten before the Plot is added to a scene will cause an AttributeError exception to be raises.
E.g. in this example. This pullrequest avoids that.

```python
import numpy as np
import pyqtgraph as pg

w = pg.GraphicsLayoutWidget()
x = np.linspace(0, 6, 1000)
y = np.sin(x)
p = pg.PlotItem(x=x, y=y)
p.setAxisItems({'bottom': pg.AxisItem(orientation='bottom', pen='r')})
w.addItem(p)
w.show()

if __name__ == '__main__':
    print(pg.__version__)
    pg.exec()

```


ixjlyons (approved):

j9ac9k:
Thanks for the PR @tblum  This diff LGTM, merging!

PR #3183, opened 2024-11-05
REPL fonts should be consistently monospaced

kevinanewman:
Current REPL fonts are not consistent cross-platform and/or between input and output.

These changes make the fonts monospaced cross-platform and consistent between the REPL input and output, at least for windows and macOS.

Thanks, Kevin

 ---
code change comments:

set output char format to self.output.font for consistency 
set input font to be the same as the output font for consistency 
drop "TypeWriter" style hint, it seems to map to "American Typewriter" on macos, which is NOT fixed-width...

### DIFF
--- a/pyqtgraph/console/repl_widget.py
+++ b/pyqtgraph/console/repl_widget.py
@@ -33,10 +33,12 @@ class ReplWidget(QtWidgets.QWidget):
         outputFirstLineBlockFormat = QtGui.QTextBlockFormat(outputBlockFormat)
         outputFirstLineBlockFormat.setTopMargin(5)
         outputCharFormat = QtGui.QTextCharFormat()
+        outputCharFormat.setFont(self.output.font())
         outputCharFormat.setFontWeight(QtGui.QFont.Weight.Normal)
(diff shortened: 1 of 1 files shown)


pyqtgraph/console/repl_widget.py
```diff
@@ -55,7 +57,7 @@ def _setupUi(self):
 
         self.output = QtWidgets.QTextEdit(self)
         font = QtGui.QFont("monospace")
-        font.setStyleHint(QtGui.QFont.StyleHint.TypeWriter, QtGui.QFont.StyleStrategy.PreferAntialias)
+        # font.setStyleHint(QtGui.QFont.StyleHint.TypeWriter, QtGui.QFont.StyleStrategy.PreferAntialias)
```
j9ac9k, line 60: Can you remove the comment, and just delete the line? Thanks!

kevinanewman: Certainly

j9ac9k:
thanks for the PR @kevinanewman Merging.

Issue #3185, opened 2024-11-07
DateAxisItem will often not display that tick marks and other AxisItem strangeness


tblum:
### Short description
My situation is that I need the user to be able to switch between different axes. E.g. relative times (seconds since start), and absolute time (DateTime)

This should be pretty straight forward in my mind, but has been the cause of a lot of frustration. So I have created a small example which demonstrates some of this 

Most often the issue is that the tick labels will simply not be displayed when using the DateAxisItem

**Method 1) Works**
1. create new AxisItem
2. setAxisItem
3. update plotdateitem with setData

**Method 2) Does not work for DateAxisItem**
1. update plotdateitem with setData
2. create new AxisItem
3. setAxisItem

**Method 3) Does not work for DateAxisItem**
1. clear() plotitem
2. create new AxisItem
3. setAxisItem
4. create PlotDateItem
5. add plotdateitem

In my mind all three methods should work (I am sure you can come up with more methods).
the reason the tick labels are not being displayed is that they are all being excluded due to line 1277 - 1279 in AxisItem.py:
```python
    br = self.boundingRect()
    if not br.contains(rect):
        continue
```
For some reason boundingRect() does not yield the same result when I call it from my script.

####Second Issue
Setting the same AxisItem (and data) twice results in the ticks being between 0 and 1, no matter the type of axis. This is also not the desired result (I am guessing)

### Code to reproduce

```python
from datetime import datetime
import numpy as np
from PySide6.QtCore import Qt
import pyqtgraph as pg

start = datetime(2024,7,7)
stop = datetime(2024,7,9)
x = np.linspace(start.timestamp(), stop.timestamp(), 24*60*60)
y = np.cumsum(np.random.rand(len(x))-0.5)
p: pg.PlotItem
pdi: pg.PlotDataItem

class GLW(pg.GraphicsLayoutWidget):
    global p, pdi, x, y
    def keyPressEvent(self, event):
        match event.key():
            case Qt.Key.Key_S:
                p.setAxisItems({'bottom': pg.AxisItem(orientation='bottom')})
            case Qt.Key.Key_D:
                p.setAxisItems({'bottom': pg.DateAxisItem(orientation='bottom')})
            # Method 1
            case Qt.Key.Key_W:
                p.setAxisItems({'bottom': pg.AxisItem(orientation='bottom')})
                p.items[0].setData(x=x-x[0] , y=y)
            case Qt.Key.Key_E:
[...]

PR #3187, opened 2024-11-08
Fix bug causing DateTimeAxis to not display tick labels #3185

tblum:
This pull-request fixes the tick-labels not being displayed part of Issue #3185

I found that the bug in due to a floating point rounding error, which is not being taken into account in the QRectF.contains method. But the equality operator does take floating point rounding into account.

### Demonstration of floating point rounding errer and work-around using ```intersected``` and ```==```
 ```python 
In [41]: QtCore.QRectF(0, -5, 600, 19.77).contains(QtCore.QRectF(300, 2, 30, 12.77))
Out[41]: True
In [42]: QtCore.QRectF(0, -5, 600, 19.775).contains(QtCore.QRectF(300, 2, 30, 12.775))
Out[42]: False
In [46]: QtCore.QRectF(0, -5, 600, 19.775) & QtCore.QRectF(300, 2, 30, 12.775) == QtCore.QRectF(300, 2, 30, 12 .775)
Out[46]: True
```

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -1275,7 +1275,8 @@ class AxisItem(GraphicsWidget):
                 #p.drawText(rect, textFlags, vstr)
 
                 br = self.boundingRect()
-                if not br.contains(rect):
+                # br.contains(rect) suffers from floating point rounding errors
+                if not br & rect == rect:
                     continue
 
                 textSpecs.append((rect, textFlags, vstr))


oyvindlr:
Just want to mention that I am experiencing the same issue and would be happy if it this PR is merged.

j9ac9k:
Thanks for the PR @tblum I didn't realize the floating point errors with the use of `QRectF.contains()` having this issue.  We make use of `.contains()` in other parts of the library, but looks like most of our usages are for checking if a point is inside a `QRectF`

PR #3190, opened 2024-11-12
PlotDataItem setData() listOfDicts fix?

kevinanewman:
It's entirely possible I'm confused here, but the way it is now, 'data' (etc) would have to be in the list ``data``, but then a string isn't a dict and the list interpretation fails...

So I think it should be ``if k in data[0]``, not ``if k in data``, in other words, the same handling as ``'x'`` and ``'y'``.

I should have merged first, but see commit  https://github.com/pyqtgraph/pyqtgraph/pull/3190/commits/369a751fc1b3241121ec75d3bfccf49e0770550e for the code change...

Otherwise, any help appreciated!

Thanks,
Kevin


j9ac9k:
You're absolutely right. Huh, this code is ancient (git blame says from 13 years ago!) ...how did this get avoided for so long!  Nice find!

PR #3205, opened 2024-12-20
assume matplotlib supports Qt6

pijyoi:
Since https://github.com/matplotlib/matplotlib/pull/27948, pyqtgraph's test suite has been skipping matplotlib tests on Qt6.

As pyqtgraph had added support for Qt6 earlier than matplotlib, pyqtgraph's CI would skip matplotlib tests if it was determined that the installed version of matplotlib did not support Qt6. This was done by testing for the presence of the backend "QtAgg" in matplotlib.

Since https://github.com/matplotlib/matplotlib/pull/27948, matplotlib backend names have been changed to lower-case and this results in "QtAgg" not being a match.

In addition, pyqtgraph's method of detection was also deprecated, resulting in the following warning:
```
  C:\work\repos\pyqtgraph\tests\exporters\test_matplotlib.py:17: MatplotlibDeprecationWarning: The interactive_bk attribute was deprecated in Matplotlib 3.9 and will be removed in 3.11. Use ``matplotlib.backends.backend_registry.list_builtin(matplotlib.backends.BackendFilter.INTERACTIVE)`` instead.
    pg.Qt.QT_LIB in ["PySide6", "PyQt6"] and "QtAgg" not in matplotlib.rcsetup.interactive_bk,
```

This PR fixes the above-mentioned issues by removing the conditional testing and simply assuming that matplotlib supports Qt6.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -9,6 +9,8 @@ faulthandler_timeout = 60
 qt_log_level_fail = WARNING
 qt_log_ignore =
     QXcbConnection: XCB error
+    # ignore PySide6 warnings in Matplotlib < 3.10.1
+    Registering dynamic slot.*FigureCanvasQTAgg
 markers =
     qt_log_ignore: fallback if pytest-qt is not installed
     qt_no_exception_capture: fallback if pytest-qt is not installed
--- a/tests/exporters/test_matplotlib.py
+++ b/tests/exporters/test_matplotlib.py
@@ -6,20 +6,10 @@ import pyqtgraph as pg
 from pyqtgraph.exporters import MatplotlibExporter
 
 pytest.importorskip("matplotlib")
-import matplotlib
 from packaging.version import Version, parse
 
 
 app = pg.mkQApp()
-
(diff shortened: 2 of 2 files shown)


pijyoi:
PySide6 test failures are due to `pytest-qt` added since #3123.
The following Qt warnings in matplotlib code result in the CI failures.
```
QtWarningMsg: Warning: Registering dynamic slot "_update_screen(QScreen*)" on "FigureCanvasQTAgg". Consider annotating with @Slot(QScreen)
QtWarningMsg: Warning: Registering dynamic slot "_update_pixel_ratio()" on "FigureCanvasQTAgg". Consider annotating with @Slot()
```

Note that these failures had previously not surfaced as the tests concerned had been skipped (wrongly).

pijyoi:
I have submitted https://github.com/matplotlib/matplotlib/pull/29355, which if accepted, will allow us to ***not*** ignore the warning.

j9ac9k:
Thanks for following up on this @pijyoi

PR #3206, opened 2024-12-20
To align mark and the label vertically in a Legend

### LINKED ISSUE #3027: Increasing Legend text generate a vertical misalignment with the sample.
Hello,
I increased the text size of the labels inside a legend, but now there is a vertical misalignment with the samples, as in the following image.

image

I can't find a way to vertical align the two elements.

Thanks in advance, pyqtgraph is awesome!

### LINKED ISSUE #3199: Vertical mis-alignment in LegendItem between ItemSample and Label
### Short description

*ItemSample* and corresponding *Label* are vertically mis-aligned.
It is clearly observable if the size of the text is large.
Following code is an example from pyqtgraph examples. (PyQtGraph Examples > GraphicsItems > LegendItem)
`labelTextSize='20pt'` was given to `LegendItem` to show the issue.

### Code to reproduce

```python
"""
Demonstrates basic use of LegendItem
"""

import numpy as np

import pyqtgraph as pg

win = pg.plot()
win.setWindowTitle('pyqtgraph example: BarGraphItem')

# # option1: only for .plot(), following c1,c2 for example-----------------------
# win.addLegend(frame=False, colCount=2)

# bar graph
x = np.arange(10)
y = np.sin(x+2) * 3
bg1 = pg.BarGraphItem(x=x, height=y, width=0.3, brush='b', pen='w', name='bar')
win.addItem(bg1)

# curve
c1 = win.plot([np.random.randint(0,8) for i in range(10)], pen='r', symbol='t', symbolPen='r', symbolBrush='g', name='curve1')
c2 = win.plot([2,1,4,3,1,3,2,4,3,2], pen='g', fillLevel=0, fillBrush=(255,255,255,30), name='curve2')

# scatter plot
s1 = pg.ScatterPlotItem(size=10, pen=pg.mkPen(None), brush=pg.mkBrush(255, 255, 255, 120), name='scatter')
spots = [{'pos': [i, np.random.randint(-3, 3)], 'data': 1} for i in range(10)]
s1.addPoints(spots)
win.addItem(s1)

# # option2: generic method------------------------------------------------
legend = pg.LegendItem((80,60), offset=(70,20), labelTextSize='20pt')
legend.setParentItem(win.graphicsItem())
legend.addItem(bg1, 'bar')
legend.addItem(c1, 'curve1')
legend.addItem(c2, 'curve2')
legend.addItem(s1, 'scatter')

if __name__ == '__main__':
    pg.exec()
```

### Expected behavior

*Mark* and *Text* are on the same level in Legend.

image

### Real behavior

*Marks* are higher than corresponding *Texts* in Legend.

image

### Tested environment(s)

 * PyQtGraph version: '0.13.7'
 * Qt Python binding: 'PySide6 6.4.2 Qt 6.4.2'
 * Python version: 3.9.13
 * NumPy version: '1.23.4'
 * Operating system: Windows
 * Installation method: pip

### Additional context
I found that `ItemSample` had *zero* width and height.
Following change in `LegendItem.py` could produce expected behavior.
However, as I'm not familiar with PyQtGraph source code, I hope you can check and apply it if it is proper.

image

jiban:
This will solve https://github.com/pyqtgraph/pyqtgraph/issues/3199

`ItemSample` had zero width and height.
This change sets width and height according to the `boundingRect` and adjusts horizontal spacing accordingly.
Besides, vertical alignment option has been given.

Fixes #3199
Fixes #3027

### DIFF
--- a/pyqtgraph/graphicsItems/LegendItem.py
+++ b/pyqtgraph/graphicsItems/LegendItem.py
@@ -32,7 +32,7 @@ class LegendItem(GraphicsWidgetAnchor, GraphicsWidget):
     sigDoubleClicked = QtCore.Signal(object, object)
     sigSampleClicked = QtCore.Signal(object)
 
-    def __init__(self, size=None, offset=None, horSpacing=25, verSpacing=0,
+    def __init__(self, size=None, offset=None, horSpacing=5, verSpacing=0,
                  pen=None, brush=None, labelTextColor=None, frame=True,
                  labelTextSize='9pt', colCount=1, sampleType=None, **kwargs):
         """
@@ -245,7 +245,7 @@ class LegendItem(GraphicsWidgetAnchor, GraphicsWidget):
                     # MAKE NEW ROW
                     col = 0
                     row += 1
-        self.layout.addItem(sample, row, col)
+        self.layout.addItem(sample, row, col, alignment=QtCore.Qt.AlignmentFlag.AlignVCenter)
         self.layout.addItem(label, row, col + 1)
         # Keep rowCount in sync with the number of rows if items are added
         self.rowCount = max(self.rowCount, row + 1)
@@ -359,6 +359,8 @@ class ItemSample(GraphicsWidget):
     def __init__(self, item):
         GraphicsWidget.__init__(self)
         self.item = item
+        self.setFixedWidth(20)
+        self.setFixedHeight(20)
 
     def boundingRect(self):
         return QtCore.QRectF(0, 0, 20, 20)


ixjlyons (approved):
This seems reasonable enough to me.

I'll point out that making the label text really small gives the opposite issue where the sample dictates the row size and the text appears aligned to the top of the row, but setting `AlignVCenter` on the label doesn't fix it. I think this is fairly unlikely unless/until we make the sample size modifiable in some way rather than hard-coding 20 (i.e. it's not really noticeable until you get down to `labelTextSize="6pt"` or so).

Note this should also close #3027

j9ac9k:
Thanks for the PR @jiban and thanks for the review @ixjlyons 

I updated the original post so the relevant issues are automatically closed.  Diff LGTM, merging.

PR #3216, opened 2025-01-08
Fix Comment Error in examples/dockarea.py

ChenQiang-AISS:
A minor fix in the code comments in examples/dockarea.py. 

### Change Made:
Updated the comment from:
`area.addDock(d6, 'top', d4)   ## place d5 at top edge of d4`
to:
`area.addDock(d6, 'top', d4)   ## place d6 at top edge of d4`


ixjlyons:
Good catch, thanks!

PR #3223, opened 2025-01-16
GLLinePlotItem: restore changes to state

pijyoi:
Issue: when running the `GLMeshItem` example, the line edges and wireframe can be seen to be thicker than before (e.g. 0.13.7)

Reason:
Drawing the grid enables anti-aliasing which is not disabled after. Thus when drawing the edges of the meshes, anti-aliasing is also enabled, where previously it was not.

For compatibility with OpenGL ES and also Core profile, ecab72bc257686fdb2b1a3d8d4083b9e5f525869 removed calls to `glPushAttrib` and `glPopAttrib`. Those calls wrapped around the rendering of each `GLGraphicsItem`, effectively allowing each item to make changes to the global OpenGL state without restoring them.

Fix:
Restore the global state to their initial defaults in `GLLinePlotItem`.

### DIFF
--- a/pyqtgraph/opengl/MeshData.py
+++ b/pyqtgraph/opengl/MeshData.py
@@ -99,10 +99,10 @@ class MeshData(object):
         return self._edges
         
     def setFaces(self, faces):
-        """Set the (Nf, 3) array of faces. Each rown in the array contains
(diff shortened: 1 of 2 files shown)


j9ac9k:
Thanks for the PR @pijyoi love to see the OpenGL support keep improving ❤️

PR #3226, opened 2025-01-23
GraphicsLayout: Fix removeItem() regression

### LINKED ISSUE #3195: GraphicsLayoutWidget does not expand its item to use all space
### Short description

Since #3167, GraphicsLayoutWidget no longer expands its item to use all space.

Note: A real program would of course not repeatedly clear and add a new PlotItem, the example is just to demonstrate the issue.

### Code to reproduce

```python
import numpy as np
import pyqtgraph as pg

rng = np.random.default_rng()

glw.show()

def update():
    glw.clear()
    pi = glw.addPlot()
    pi.plot(rng.random(100))

timer = QtCore.QTimer()
timer.timeout.connect(update)
timer.start(100)

pg.exec()

```

### Expected behavior

After clearing and adding a new plot, the new plot should occupy the whole window.

### Real behavior

After clearing and adding a new plot, the new plot occupies only a small top-left corner of the window.

### Tested environment(s)

 * PyQtGraph version: 0.14.0dev0
 * Qt Python binding: PySide6 6.8.1 Qt 6.8.1
 * Python version: 3.10.11
 * NumPy version: 2.1.3
 * Installation method: pip

dnadlinger:
This fixes a regression introduced in
54efcbd2398c0f39a11ec1d14fe8f7f3f3a6b901, where the stretch factors for the row/column previously containing a removed item were zeroed. In general, this is of course incorrect, as there will be other items sharing the same row/column that should not be disturbed. The intention behind that part of the change is not clear to me; this commit simply reverts it.

Fixes #3195.

### DIFF
--- a/pyqtgraph/graphicsItems/GraphicsLayout.py
+++ b/pyqtgraph/graphicsItems/GraphicsLayout.py
@@ -182,10 +182,6 @@ class GraphicsLayout(GraphicsWidget):
         # Clear the row and column where the item was
         for r, c in self.items[item]:
             del self.rows[r][c]
-            
-            # Adjust the layout by updating row/column stretch factors
-            self.layout.setRowStretchFactor(r, 0)
-            self.layout.setColumnStretchFactor(c, 0)
         
         # Clean up the references to the removed item
         del self.items[item]


dnadlinger:
@j9ac9k: Apologies to ping you, but since you merged the PR introducing this regression, I was hoping you could have a quick look. The regression affects all pyqtgraph applications using `clear()`/`removeItem()`, so would be important to have fixed before the next release.

j9ac9k:
Hi @dnadlinger 

Sorry it's taken me so long to get to this.  When it comes to undoing PRs (to address regressions), I try and test more carefully to ensure the original issue is addressed as well as the bug from the initial fix.

Can you submit a reproducible example that I can use to work from?  Thanks.

dnadlinger:
No worries, @j9ac9k, I know the trouble that maintaining open-source libraries brings all to well…

A reproduction case can be found in https://github.com/pyqtgraph/pyqtgraph/issues/3195 (@pijyoi seems to have independently run into the same issue as I did in my project).

j9ac9k:
Closing/Opening to trigger a new CI run.

j9ac9k:
Ugh, who merged that change 🙃  The other PR doesn't actually recover the space either... Thanks for the fix @dnadlinger

PR #3231, opened 2025-01-28
GLMeshItem: allow ubyte color buffers

pijyoi:
When specifying colors for a large number of vertices, keeping colors as `float32` (16-bytes) instead of `uint8` (4-bytes) can be wasteful in memory and bandwidth.

Furthermore, `trimesh` colors are retrieved as `uint8`, which then requires conversion before being able to be used in `pyqtgraph`. (See the example in #3230)

`MeshData` doesn't do any computation on the colors besides indexing and reshaping. i.e. it is actually agnostic to the dtype.
`GLMeshItem` only needs to tell OpenGL whether the VBO is float32 or uint8.


j9ac9k:
Thanks @pijyoi for this PR!  Wonderful to see more expanding compatibility within the opengl codepaths!

PR #3235, opened 2025-02-02
fix GLIsosurface overflow warning on numpy 2

pijyoi:
`pg.isosurface()` relies on NumPy < 2.0 promotion behavior.
An overflow warning occurs on NumPy >= 2.0.
Fix by explicitly promoting to an `int`.

```python
In [1]: import numpy as np
In [2]: val = np.arange(10, dtype=np.uint16)[0] - 1
In [3]: val, val.dtype
Out[3]: (-1, dtype('int32'))
In [4]: np.__version__
Out[4]: '1.26.4'
```

```python
In [1]: import numpy as np
In [2]: val = np.arange(10, dtype=np.uint16)[0] - 1
<ipython-input-2-04d47eed1b3c>:1: RuntimeWarning: overflow encountered in scalar subtract
  val = np.arange(10, dtype=np.uint16)[0] - 1
In [3]: val, val.dtype
Out[3]: (np.uint16(65535), dtype('uint16'))
In [4]: np.__version__
Out[4]: '2.2.1'
```

### DIFF
--- a/pyqtgraph/examples/GLIsosurface.py
+++ b/pyqtgraph/examples/GLIsosurface.py
@@ -38,12 +38,11 @@ data = np.abs(np.fromfunction(psi, (50,50,100)))
 print("Generating isosurface..")
 verts, faces = pg.isosurface(data, data.max()/4.)
 
-md = gl.MeshData(vertexes=verts, faces=faces)
-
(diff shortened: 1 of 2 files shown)


j9ac9k:
Thanks @pijyoi !

PR #3236, opened 2025-02-04
fix: GL_POINT_SPRITE was not enabled for NoProfile

pijyoi:
The previous implementation had notably neglected to enable GL_POINT_SPRITE for OpenGL <= 2.1.
If it had previously worked, it would probably be due to the permissiveness of the various OpenGL implementations.

Remarks:
1) For Core Profile, it is not needed and trying to enable it will result in an error.
2) macOS is the platform that will default to OpenGL 2.1 (NoProfile). However, point sprites was working despite not having enabled GL_POINT_SPRITE.

### DIFF
--- a/pyqtgraph/opengl/items/GLScatterPlotItem.py
+++ b/pyqtgraph/opengl/items/GLScatterPlotItem.py
@@ -1,9 +1,9 @@
+import math
 import importlib
 
 from OpenGL.GL import *  # noqa
 import numpy as np
 
-from ... import functions as fn
 from ...Qt import QtGui, QT_LIB
 from .. import shaders
 from ..GLGraphicsItem import GLGraphicsItem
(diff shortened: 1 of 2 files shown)


pijyoi:
In addition to fixing the bug, I also took the opportunity to move the computation of `pxMode=False` into the vertex shader.
Previously, moving the camera around without any data change would necessitate a recomputation of the scaled point size on the CPU. Now, that code executes in the vertex shader.

A script to demonstrate the improvement. On my laptop, `master` runs at 15 fps while this PR runs at 120 fps. Alternatively, task manager shows that the CPU utilization has decreased while the GPU utilization has increased. Your mileage may vary, depending on your system.
```python
import importlib

import numpy as np

import pyqtgraph as pg
import pyqtgraph.opengl as gl

pngfile = importlib.resources.files("pyqtgraph.icons.peegee") / "peegee_512px@2x.png"
qimg = QtGui.QImage.fromData(pngfile.read_bytes()).mirrored(True, True)
qimg.convertTo(QtGui.QImage.Format.Format_RGBA8888)
data = pg.functions.ndarray_from_qimage(qimg).astype(np.float32) / 255
colors = data.reshape((-1, 4))

pg.mkQApp()
win = gl.GLViewWidget()
win.show()
win.setCameraPosition(distance=2, elevation=30)

rows, cols = data.shape[:2]
pos3 = np.zeros((rows,cols,3))
pos3[...,:2] = np.mgrid[:rows, :cols].transpose(1,2,0) / rows
pos3 = pos3.reshape(-1, 3)
item = gl.GLScatterPlotItem(pos=pos3, color=colors, size=0.001, pxMode=False, glOptions='translucent')

win.addItem(item)

def update():
    win.orbit(-1, 0)
    framecnt.update()

timer = QtCore.QTimer()
timer.timeout.connect(update)
timer.start(0)

framecnt = FrameCounter()
framecnt.sigFpsUpdate.connect(lambda fps: win.setWindowTitle(f'{fps:.1f} fps'))

pg.exec()
```

j9ac9k:
I had identical speedups on macOS (using M2 Max chip).  Nice work!

pijyoi:
If you look closely, the left and right dots are smaller than the one in the center, even though they are all at the same depth.
The formulation used by `GLScatterPlotItem` is that size is scaled to the distance from the eye, so dots at the edges appear smaller (at the same depth).

image

An alternative implementation would be to have dots at the same depth to have the same size. This is a simpler calculation and doesn't require the camera position and view transform to be sent to the shader.

image

PR #3237, opened 2025-02-04
Use filtfilt for bidirectional filtering

MxFxM:
using filtfilt instead of two interleaved lfilter calls to scipy.signal is more elegant and easier to understand. we want to make use of this functionality


j9ac9k:
Thanks for the PR @MxFxM   I agree, the resulting code is much easier to read!

PR #3240, opened 2025-02-07
Plot option to subtract mean value from all samples

wetzeld:
Hello Team,

this PR introduces a new plot transformation that subtracts the mean value of a dataset from all samples.

This makes plotting / analyzing two signals with different offsets much easier if you are only interested in the "AC" part of the signal.

For this I added a new menu item to the Plot Options -> Transformations menu. This was done using the Qt Designer and pyuic6 in v6.1.0 to keep the diff rather small.

image


wetzeld:
@j9ac9k, I think the failed check has nothing to do with this PR but seems to be a temporary failure. Could you maybe re-run the checks?

Besides that, any thoughts on the PR or the feature itself?

wetzeld:
Rebased on current master

j9ac9k:
Thanks for the PR @wetzeld 

We discussed in another issue/PR the addition of having a "plugin" system to add other items, but we don't have that yet so I'll merge this.  Thanks for the PR @wetzeld

PR #3246, opened 2025-02-15
create a default VAO if using Core profile

pijyoi:
This PR makes it more likely for pyqtgraph's OpenGL code to run if the user had made `GLViewWidget` use a Core profile.

In Core profiles, the default VAO is removed and it is mandatory to make use of user-created VAOs.
`glDrawArrays` and `glDrawElements` fail automatically if a user-created VAO is not bound.
One workaround is to create a global VAO and leave it bound throughout the program.

One issue encountered is that `GLTextItem` and `GLGradientLegendItem` make use of `QPainter`, which internally unbinds any existing VAO at the end of its execution. To workaround this, we simply rebind our global VAO before calling each `GLGraphicsItem`'s `paint()`.

Note that this PR is not a sufficient condition for `pyqtgraph` to run on any system using the Core profile. An implementation of the Core profile is not required to support the GLSL 120 syntax that `pyqtgraph` uses.

### DIFF
--- a/pyqtgraph/examples/GLGraphItem.py
+++ b/pyqtgraph/examples/GLGraphItem.py
@@ -2,10 +2,20 @@
 Demonstrates use of GLGraphItem
 """
 
+import sys
+
 import numpy as np
 
 import pyqtgraph as pg
+from pyqtgraph.Qt import QtGui
 import pyqtgraph.opengl as gl
+
+if 'darwin' in sys.platform:
+    fmt = QtGui.QSurfaceFormat()
+    fmt.setRenderableType(fmt.RenderableType.OpenGL)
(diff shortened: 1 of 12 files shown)


pijyoi:
Most desktop OpenGL (except macOS) implementations should support the `GL_ARB_ES3_compatibility` extension, which allows them to parse OpenGL ES 3.0 shaders. On such systems, use the es3 shader for GLVolumeItem.

macOS when running in OpenGL 4.1 Core supports only `GL_ARB_ES2_compatibility`.

pijyoi:
`renderToArray()` has been updated to work in Core profile and OpenGL ES.

GLGraphicsItem(s) making use of QPainter won't get exported, but at least they won't stop the rest from working.

j9ac9k:
Love to see the embracing of OpenGL Core Profile!

Issue #3245, opened 2025-02-12
Add LTTB downsampling support


hinxx:
Consider adding LTTB downsampling support for timeseries based plots. See demo at https://www.base.is/flot/.

Python based modules to consider:

https://github.com/dgoeries/lttbc/
https://pypi.org/project/lttb/

Tmarinus:
I liked the algorithm for my own application so decided to add it to pyqtgraph. No idea if it will be accepted but otherwise you can use the code to patch your pyqtgraph locally.

Tmarinus:
Glad I could help. I will keep track of the pull request and am willing to put in the time to implement the final solution that is requested to get this added. 
I was happy I found your post because I am a researcher that has come across the problem of needing to down-sample data where we have sharp changes in the data. Mean and peak are not usable in those situations so I stayed with sub-sampling which in the end slows down the program quite a lot because I cannot put high down-sampling since that makes us miss important information. 

I attempted to find good algorithms for this (searching for weighted averaging where the window size changes based on the xy, step size) but never found something that worked well. The lttb algorithm works fantastic and is even quite fast. 

Besides all of this, later in the day I realized that the implementation above is already based on pyqtgraph 0.14.0 which makes huge changes to pyqtgraph plotdataitem, for me breaking my tools. So I have also made a branch for 0.13.7 which only adds lttb, I would recommend using that. Please let me know if you find any bugs or anything.

https://github.com/Tmarinus/pyqtgraph/tree/lttb-branch

hinxx:
Yeah, I like LTTB for similar reasons.

On a related note, there is this https://github.com/predict-idlab/MinMaxLTTB. You might find it useful, I have not used it myself yet.

hinxx:
Works as expected! Great job! I've posted a little video at https://github.com/slaclab/pydm/issues/1166#issuecomment-2676225449.

PR #3255, opened 2025-02-21
Added lttb down-sampling

Tmarinus:
As mentioned in #3245 added lttb down-sampling. This does require the lttbc package, so I chose to ignore lttb down sampling if lttbc is not installed. 

First time I do a pull request for pyqtgraph so let me know if anything needs to change.


j9ac9k:
Thanks for the PR.  Can you tell me more about the lttb downsampling algorithm, the circumstances it would be useful to be used and the benefits it provides over mean

I'm not feeling great about making this package an optional dependency _but_ we don't provide a good way for users to inject their own downsampling mechanisms so I'd be fine to exercise it.

I would like to benchmark this a bit and see if it would be better to move the try / except ImportError elsewhere.


Also, the readthedocs build error:

```
/home/docs/checkouts/readthedocs.org/user_builds/pyqtgraph/checkouts/3255/doc/source/../../pyqtgraph/graphicsItems/PlotItem/PlotItem.py:docstring of pyqtgraph.graphicsItems.PlotItem.PlotItem.PlotItem.setDownsampling:19: ERROR: Malformed table.
Text in column margin in table line 17.

=============== ====================================================================
**Arguments:**
ds              (int) Reduce visible plot samples by this factor, or

                (bool) To enable/disable downsampling without changing the value.

auto            (bool) If `True`, automatically pick ``ds`` based on visible range

mode            'subsample': Downsample by taking the first of N samples. This
                method is fastest but least accurate.

                'mean': Downsample by taking the mean of N samples.

                'peak': Downsample by drawing a saw wave that follows the min and
                max of the original data. This method produces the best visual
                representation of the data but is slower.
lttb_threshold (int) Threshold for lttb down-sampling method
=============== ====================================================================
```

Tmarinus:
I can provide a pyqtgraph example (visual and performance) in a later stage. Think examples.SpeedPlotTest can be adjusted to test downsampling performance (in fps). I noticed that lttb sits between mean and subsampling. But will do this more systematically when I have time. 

The import, I think a solution to not force dependencies, cause I agree that would be sub-optimal. Would be a import test on the top, if lttbc is not installed do not add the lttbc. 
plotConfigTemplate_generic.py
```
try:
    import lttbc
    self.lttbRadio.setToolTip(_translate("Form", "Downsample using Largest-Triangle-Three-Buckets requires lttb package to be installed."))
    self.lttbRadio.setText(_translate("Form", "Lttb"))
    self.lttbSpin.setToolTip(_translate("Form", "Threshold value for lttb down-sampling."))
    self.lttbSpin.setSuffix(_translate("Form", "x"))
except ImportError as e:
    pass
```

Then in both PlotItem and PlotDataItem
```
try:
    import lttbc
    lttbc_installed = True
except: 
    lttbc_installed = False


Before lttb functions a check if lttbc_installed is true or false
```

Tho I feel that it is not the prettiest solution, for performance it only adds a single check once running. 

Lastly, a example why lttb can be interesting (from: https://www.datylon.com/blog/line-charts-sampling-time-series-data-sets) I can try to make an example similar to this for peak, subsampling, mean and lbbc in pyqtgraph. 
image

Tmarinus:
> I'm not feeling great about making this package an optional dependency _but_ we don't provide a good way for users to inject their own downsampling mechanisms so I'd be fine to exercise it.

Side note on this. I would absolutely love the ability to get access to the right click menu. I know it is possible to remove it and build your own. I think being able to inject and adjust the context menu would also bring us one step closer to injecting specific down-sampling algorithms. One could now inherit from PlotDataItem, which I have attempted before. And while this works it does not play nicely with autodownsampling and the context menu. 
(in the end downsampling got improved in pyqtgraph at some point allowing me to remove my own horrible code) But overwriting and changing the context menu would have been super nice.

Tmarinus:
Not the most fair comparison because lttb is not dependent on the zoom right now (it ignores auto, since it doesnt use downsampling value but a threshold). 
However depending on the zoom peak looks really ugly in my data (lots of nose high amount of datapoints). 
Mean and Sub look alright here, but once the data gets bigger (multiple lines for example, it starts removing peaks). 

Lttb keeps looking nicely even when decreasing (lower threshold == higher downsampling) the threshold value. 

mean
image
sub
image
peak
image
lttb
image

hinxx:
>Thanks for the PR. Can you tell me more about the lttb downsampling algorithm, the circumstances it would be useful to be used and the benefits it provides over mean

If I may, I want to add some context, as I was the one requesting the addition of LTTB (Largest-Triangle-Three-Buckets) algorithm to the pyqtgraph.

According to the original author's thesis https://skemman.is/bitstream/1946/15343/3/SS_MSthesis.pdf the abstract says:
```
As human beings, we often wish to visualize certain information in order to make better sense of it. This can be a somewhat challenging enterprise for large amounts of data and might require downsampling the data, retaining only the important visual characteristics.
```

The most important piece of text being "retaining only the important visual characteristics". In some fields of science it is not desirable to decimate or calculate the mean of several samples of data as that would result in losing the information contained in the original data set.

> I'm not feeling great about making this package an optional dependency but we don't provide a good way for users to inject their own downsampling mechanisms so I'd be fine to exercise it.

The question of dependencies on the external package could well be voided by adding the https://github.com/dgoeries/lttbc/blob/master/lttbc.c to the pyqtgraph code base and have it always available to the user as a part of package. I do not see any maintenance overhead as it seem like a final implementation of the algorithm; original authors code in JS has not changed in the last 12 years (https://github.com/sveinn-steinarsson/flot-downsample/) and https://github.com/dgoeries/lttbc/blob/master/lttbc.c has not been touched in the last 5 years.

ixjlyons:
> The question of dependencies on the external package could well be voided by adding the https://github.com/dgoeries/lttbc/blob/master/lttbc.c to the pyqtgraph code base and have it always available to the user as a part of package. I do not see any maintenance overhead as it seem like a final implementation of the algorithm; original authors code in JS has not changed in the last 12 years (https://github.com/sveinn-steinarsson/flot-downsample/) and https://github.com/dgoeries/lttbc/blob/master/lttbc.c has not been touched in the last 5 years.

I'm -1 on bringing over the C code. We'd need to support building that C code into an extension module when pyqtgraph is installed and we'd want to be building platform-specific pyqtgraph wheels with the extension module pre-built. That's precisely what the lttbc package already does.

That said, lttbc is currently only shipping wheels for Python 3.10 through 3.12 and for Linux and Windows [[ref](https://pypi.org/project/lttbc/#files)]. This is something I could potentially help out with (I think [cibuildwheel](https://cibuildwheel.pypa.io/en/stable/) could probably just handle it), and I'd personally much rather improve the standalone project than complicate pyqtgraph's infrastructure.

---

Aside from all that, this algorithm produces some impressive results.

I'd love to see hooks in the "processing for visualization" pipeline as well as a less opaque interface with the built-in context menu. Something like this could then be more like a "plugin" of sorts, but it's not a trivial problem. I suppose if we add this as an optional feature now, it could eventually be swapped for a package implementing such a plugin without breaking anything.

Tmarinus:
Alright so many messages and ideas. 

> I'd love to see hooks in the "processing for visualization" pipeline as well as a less opaque interface with the built-in context menu. Something like this could then be more like a "plugin" of sorts, but it's not a trivial problem. I suppose if we add this as an optional feature now, it could eventually be swapped for a package implementing such a plugin without breaking anything.

Someway to implement this directly for now and then move into a more plugin type of idea sounds ideal for me. This would allow me to use this relatively soon already (my/our software that I am using pyqtgraph for will be published soon $^{tm}$ .

Things I know I should still fix before ready:
- [ ] Enable auto settings. I think a calculation for number of points can be done similarly to how auto is done for the other methods
- [ ] Understand `plotConfigTemplate.ui` and use that instead of `plotConfigTemplate_generic.py` to make the ui modifications (or is this not used? 

Then we mainly need to decide how to handle the import for now? Add dependency or use the current except check? And should the UI be not showing lttb option if it is not available?

j9ac9k:
> > The question of dependencies on the external package could well be voided by adding the https://github.com/dgoeries/lttbc/blob/master/lttbc.c to the pyqtgraph code base and have it always available to the user as a part of package. I do not see any maintenance overhead as it seem like a final implementation of the algorithm; original authors code in JS has not changed in the last 12 years (https://github.com/sveinn-steinarsson/flot-downsample/) and https://github.com/dgoeries/lttbc/blob/master/lttbc.c has not been touched in the last 5 years.
> 
> I'm -1 on bringing over the C code. We'd need to support building that C code into an extension module when pyqtgraph is installed and we'd want to be building platform-specific pyqtgraph wheels with the extension module pre-built. That's precisely what the lttbc package already does.
> 
> That said, lttbc is currently only shipping wheels for Python 3.10 through 3.12 and for Linux and Windows [[ref](https://pypi.org/project/lttbc/#files)]. This is something I could potentially help out with (I think [cibuildwheel](https://cibuildwheel.pypa.io/en/stable/) could probably just handle it), and I'd personally much rather improve the standalone project than complicate pyqtgraph's infrastructure.
> 
> Aside from all that, this algorithm produces some impressive results.
> 
> I'd love to see hooks in the "processing for visualization" pipeline as well as a less opaque interface with the built-in context menu. Something like this could then be more like a "plugin" of sorts, but it's not a trivial problem. I suppose if we add this as an optional feature now, it could eventually be swapped for a package implementing such a plugin without breaking anything.

To be clear, I'm not suggesting we _embed_ this library in ours, or make it a dependency that is fetched when pip installing pyqtgraph.  We have precedence for supporting optional dependencies for specific functionality (scipy, h5py, matploblib, numba, cupy...).

That said, if lttbc doesn't have wheels for supported python versions for NEP-29, and covers the common platforms (windows, linux, macOS) that _is_ a problem, and I'd probably hold off on adding even optional support until lttbc can support what PyQtGraph supports.

I do think providing a plugin type architecture (and documentation on how to use it) would be the way to go.  That would likely be a larger effort well outside the scope of this PR.

If memory services, `plogConfigTemplate_generic.py` is _generated from_ `plotConfigTemplate.ui` (when using PyQt6), but we redo the imports to import from `pyqtgraph.Qt` instead of `PyQt6`. 

To add extra options to the menu, you may not need to make any changes to the template.  You can probably fetch the QMenu instance programatically at runtime, and run `QMenu.addAction()` to add the appropriate options in the right click menu that way.  I would have to play with it to come up with more specific versions.

pijyoi:
Is there some reason that numpy-only implementations like https://github.com/dexter2206/pylttb are not being considered over https://github.com/dgoeries/lttbc?
The former is MIT-licensed and the single function that it implements could just be vendored?

j9ac9k:
oh, if there is a pure numpy implementation, vendoring that would definitely be the way to go.

Tmarinus:
> Is there some reason that numpy-only implementations like https://github.com/dexter2206/pylttb are not being considered over https://github.com/dgoeries/lttbc? The former is MIT-licensed and the single function that it implements could just be vendored?
Purely that I did not spot its existence. 

Once I have some downtime I can have a look into it, removing the requirement for lttbc from this pr.

PR #3256, opened 2025-02-21
Make shaders in shaders.py run on macOS OpenGL 4.1 Core

pijyoi:
This is a continuation of #3246 for `GLMeshItem` and its dependencies.

As there are quite a number of remaining shaders in `shaders.py` (all used by GLMeshItem only), an alternate method than #3246 was used to make the shaders run on macOS OpenGL 4.1 Core. Since macOS OpenGL 4.1 Core has the ARB_ES2_compatibility extension, it is able to run ES2 shaders, with the proviso that they are marked as such with `#version 100`.


j9ac9k:
Thanks @pijyoi ! 

Side note, love that we're getting rid of some of the `#noqa` on the imports

PR #3257, opened 2025-02-22
Make PlotCurveItem and PColorMeshItem work in Core Profile

pijyoi:
There are 3 shaders involved:
1) Stencil
2) PlotCurveItem
3) PColorMeshItem

(3) by itself already has a non-legacy code path, but since it depends on (1), it also doesn't work.

The `ARB_ES2_compatibility` trick doesn't quite work due to the use of `QOpenGLShader` which does some mangling of its own.

### DIFF
--- a/pyqtgraph/Qt/OpenGLHelpers.py
+++ b/pyqtgraph/Qt/OpenGLHelpers.py
@@ -114,10 +114,13 @@ class GraphicsViewGLWidget(QtOpenGLWidgets.QOpenGLWidget):
             frag_src = "void main() { gl_FragColor = vec4(1.0); }"
 
         program = QtOpenGL.QOpenGLShaderProgram()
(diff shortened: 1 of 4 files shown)


pijyoi:
This PR would allow macOS to use the flat shading code path for PColorMeshItem, which has better performance. See example script in #3090.

pijyoi:
The initial implementation was for CoreProfile only to use the 140 shaders. But actually we can relax the condition to "OpenGL >= 3.1". This means that Windows and Linux platforms will now use the 140 shaders by default.

pijyoi:
The CI didn't use to detect errors in shader compilation. (See #3248 for a bug that slipped through)
Check and raise an exception if the shader failed to compile or link.

j9ac9k:
Thanks @pijyoi !

PR #3262, opened 2025-02-25
GLVolumeItem: calculate camera pos using ModelView matrix

pijyoi:
This PR uses the information encoded in the ModelView matrix to compute the camera's position in the model's local coordinates rather than using the `cameraPosition` and `mapFromParent`.

There is in fact a bug in the old implementation. `mapFromView` should have been used instead of `mapFromParent`.
`mapFromParent` uses only the model's own transformations whereas `mapFromView` includes any further parent transformations.

In other words, the computation would have been wrong if the `GLVolumeItem` was a child of another `GLGraphicsItem`.

### DIFF
--- a/pyqtgraph/opengl/items/GLVolumeItem.py
+++ b/pyqtgraph/opengl/items/GLVolumeItem.py
@@ -1,4 +1,3 @@
-import ctypes
 import importlib
 
 from OpenGL import GL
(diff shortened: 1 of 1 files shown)


pyqtgraph/opengl/items/GLVolumeItem.py
```diff
@@ -140,10 +139,14 @@ def paint(self):
         mat_mvp = self.mvpMatrix()
         mat_mvp = np.array(mat_mvp.data(), dtype=np.float32)
 
-        view = self.view()
+        # calculate camera coordinates in this model's local space.
+        # (in eye space, the camera is at the origin)
+        modelview = self.modelViewMatrix()
+        cam_local = modelview.inverted()[0].map(QtGui.QVector3D())
```
j9ac9k, line 145: Do you know if we get the same rounding errors that we get w/ `QTransform.inverted()` ?

j9ac9k: Looking a little more closely, maybe it's ok?

https://github.com/qt/qtbase/blob/0770171dcda24a567d6f57b87fdd284011e56e16/src/gui/math3d/qmatrix4x4.cpp#L316-L390

j9ac9k:
thanks @pijyoi this LGTM!

PR #3264, opened 2025-02-26
Fix parameter item compare unpack order

### LINKED ISSUE #3229: ChecklistParameter bug
### Code to reproduce

```python
import pyqtgraph as pg
import numpy as np
```

### Real behavior

```
An error occurred?
```

### Tested environment(s)

 * PyQtGraph version: 0.13.7
 * Qt Python binding: 2.4.2
 * Python version: 3.9.0
 * NumPy version: 1.26.4
 * Installation method: 

### Additional context

```python
    @staticmethod
    def mapping(limits):
        # Return forward and reverse mapping objects given a limit specification
        forward = OrderedDict()  ## {name: value, ...}
        reverse = ([], [])       ## ([value, ...], [name, ...])
        if not isinstance(limits, dict):
            limits = {str(l): l for l in limits}
        for k, v in limits.items():
            forward[k] = v
            reverse[0].append(v)
            reverse[1].append(k)
        return forward, reverse

    def _intersectionWithLimits(self, values: list):
        """
        Returns the (names, values) from limits that intersect with ``values``.
        """
        allowedNames = []
        allowedValues = []
        # Could be replaced by "value in self.reverse[0]" and "reverse[0].index",
        # but this allows for using pg.eq to cover more diverse value options
        for val in values:
            for limitName, limitValue in zip(*self.reverse):
                if fn.eq(limitValue, val):
                    allowedNames.append(limitName)
                    allowedValues.append(val)
                    break
        return allowedNames, allowedValues
```

In the `mapping(limits),reverse = ([], [])       ## ([value, ...], [name, ...])` assigns the first array of a set as' value 'and the second array as' name'. In the `_intersectionWithLimits(self, values: list)` function, assign the value of the first score group to `limitName` and the value of the second score group to `limitValue`. Resulting in normal use only when the array element is str in the 'limits' of `ChecklistParameterItem`

j9ac9k:
Fixes #3229 

Thank you @Qi-henry for reporting the issue and diagnosing the cause.  Added a test case that caught this error.

Funny enough, this error did not occur when using `str` types for values instead of `int` types.

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes/checklist.py
+++ b/pyqtgraph/parametertree/parameterTypes/checklist.py
@@ -274,7 +274,7 @@ class ChecklistParameter(GroupParameter):
         # Could be replaced by "value in self.reverse[0]" and "reverse[0].index",
         # but this allows for using pg.eq to cover more diverse value options
         for val in values:
-            for limitName, limitValue in zip(*self.reverse):
+            for limitValue, limitName in zip(*self.reverse):
                 if fn.eq(limitValue, val):
                     allowedNames.append(limitName)
                     allowedValues.append(val)
--- a/tests/parametertree/test_parametertypes.py
+++ b/tests/parametertree/test_parametertypes.py
@@ -189,6 +189,34 @@ def test_checklist_show_hide():
     pi.setHidden.assert_called_with(False)
     assert p.opts["visible"]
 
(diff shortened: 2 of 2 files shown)


j9ac9k:
Decided since we have a difference in behavior from `int` to `str` values, I should check both, so Ill parametrize this test for both conditions before merging.

Issue #3265, opened 2025-02-27
Memleak in arrayToQPath() function


coells:
### Short description
Using ErrorBarItem with PlotWidget causes small and gradual memleaks that become noticable when app is running a longer time.

### Code to reproduce
This code is extracted from **ErrorBarItem.drawPath()** -> **arrayToQPath(..., connect="pairs")** which builds QPainterPath using QByteArray and numpy buffer interface.

Note that the same behaviour can be easily reproduced by calling **ErrorBarItem.drawPath()** repeatedly. Only the run is slower.

```python
import numpy as np
from pyqtgraph.Qt import QtCore

if __name__ == '__main__':
    n = 1024

    for i in range(100000):
        if i % 1000 == 0:
            print('step', i)

        backstore = QtCore.QByteArray()
        backstore.resize(n * 8)
        np.frombuffer(backstore, dtype=[('x', '>f8')], count=n)

    input('done')
```

### Tested environment(s)

 * PyQtGraph version: 0.13.7
 * Qt Python binding: PySide6 6.8.2 Qt 6.8.2
 * Python version: 3.12.9
 * NumPy version: 2.1.3
 * Operating system: Linux HOST 5.4.191-el7.lime.2.x86_64 # 1 SMP Thu Jul 14 19:09:52 EDT 2022 x86_64 x86_64 x86_64 GNU/Linux
 * Installation method: conda

### Additional context
I'm not sure if this bug is part of Qt or numpy, but it triggers inside pyqtgraph. As a workaround it would help to clear QByteArray before end of `arrayToQPath()`. In which case the memory footprint of the leak is very small.

coells:
Interesting, your example is leaking heavily for me.

```
...
400000: 3223.4 MB
410000: 3303.8 MB
420000: 3384.3 MB
430000: 3464.5 MB
440000: 3545.0 MB
450000: 3625.3 MB
460000: 3705.8 MB
```

coells:
I've tested Python 3.10 and still could see the memleak. 

As far as I could explore the problem, Numpy should rely on memoryview which increases refcount and never decreases. This snippet reproduces problem as well.
```
    backstore = QtCore.QByteArray()
    memoryview(backstore)
    print(sys.getrefcount(backstore))
```

Not sure if the problem is in Qt, Python or even old version of gcc (?).

[...]

PR #3268, opened 2025-02-28
workaround PySide QByteArray leak on conda builds

pijyoi:
Addresses #3265 at the expense of penalizing the official Qt build of PySide6.

The leak only gets triggered for connect types "pairs", "array". 
"finite" is also affected if there are a great many non-finite elements.

Note that 2 extra memory writes are made with this workaround.
1) `bytearray` always initializes its memory to 0
2) when the `bytearray` is converted into a `QByteArray`.
    (this second memory access could have been avoided if `QByteArray.fromRawData` was working. But alas, the binding for it is broken)

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -2044,15 +2044,29 @@ def arrayToQPath(x, y, connect='all', finiteCheck=True):
     if hasattr(path, 'reserve'):    # Qt 5.13
         path.reserve(n)
 
-    if hasattr(path, 'reserve') and getConfigOption('enableExperimental'):
+    if getConfigOption('enableExperimental'):
         backstore = None
         arr = Qt.internals.get_qpainterpath_element_array(path, n)
     else:
-        backstore = QtCore.QByteArray()
-        backstore.resize(4 + n*20 + 8)      # contents uninitialized
-        backstore.replace(0, 4, struct.pack('>i', n))
-        # cStart, fillRule (Qt.FillRule.OddEvenFill)
-        backstore.replace(4+n*20, 8, struct.pack('>ii', 0, 0))
+        if QT_LIB == 'PySide2':
+            # When PySide{2,6} is built without Py_LIMITED_API,
+            # it leaks memory when a memory view to a QByteArray
+            # object is taken.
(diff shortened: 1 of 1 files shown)


j9ac9k:
Since the "fix" has been merged into the conda-forge feedstock, do we want to have the workaround enabled only for specific versions of pyside bindings; or is there minimal downside to the workaround that that's not needed?

pijyoi:
I did a check, the issue is also present in conda-forge's build of PySide2 5.15.15.
The issue is not present in Qt's official build PySide2 5.15.2.1.

refcnt leak introduced:
https://github.com/pyside/pyside2-setup/commit/821480b1f0d0cfb59df427315c29c364096818c4

the following comment says that Qt defaults to limited-api builds (and therefore their builds aren't affected by the leak introduced):
https://github.com/pyside/pyside2-setup/commit/032cb10444f786a7ed0a9321395f398c992eb15d

I couldn't find where pyside2-feedstock was configuring for non-limited-api build though.

Given that conda-forge has patched PySide2 to build for Python 3.11, and seem willing (https://github.com/conda-forge/pyside2-feedstock/issues/246) to push it to Python 3.12, what is pyqtgraph's policy regarding support of PySide2 going to be?

>do we want to have the workaround enabled only for specific versions of pyside bindings;

For PySide6, we can don't apply the workaround. conda-forge PySide6 6.7.3 through PySide6 6.8.2 build 0 will have the leak.

For PySide2, maybe we can apply the workaround? If PySide2 is a non-preferred binding for pyqtgraph, I guess penalizing it is not really an issue? i.e. better to be less efficient than having to worry about whether a leak is present for non-official builds.

pijyoi:
Other places where the leak occurs:
* Ubuntu 22.04 / Python 3.10.12 / PySide2 5.15.3
* Ubuntu 24.04 / Python 3.12.3 / PySide2 5.15.13

pijyoi:
For reference, pyqtgraph first started taking a memory view to QByteArray in this commit. (2021 August)
https://github.com/pyqtgraph/pyqtgraph/commit/287ec9fbeacce5b346d29f32797b2df8516fe36b

PySide2 5.15.2 had already been released for some time.

j9ac9k:
Thanks @pijyoi !

PR #3270, opened 2025-03-02
pyside6: don't take QByteArray memoryview too

pijyoi:
Continuing the investigation of #3268, although conda-forge pyside2-feedstock has patched the PySide6 bug such that PySide6 6.8.2-build1 no longer triggers the leak, this is not the case for the distro builds of PySide6.

In general, we do not know whether a distro has built PySide6 with or without the limited api.
Some limited sampling:
- OpenSUSE Tumbleweed : leak absent
- Arch Linux : leak present

Given this situation, we will have to use the less efficient codepath for PySide6 until the leak is fixed upstream in PySide6.
See https://bugreports.qt.io/browse/PYSIDE-3031.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -2048,7 +2048,7 @@ def arrayToQPath(x, y, connect='all', finiteCheck=True):
         backstore = None
         arr = Qt.internals.get_qpainterpath_element_array(path, n)
     else:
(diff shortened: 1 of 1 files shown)


pijyoi:
`-DFORCE_LIMITED_API=no`
https://src.fedoraproject.org/rpms/python-pyside6/blob/rawhide/f/python-pyside6.spec

PR #3274, opened 2025-03-06
Probe QByteArray leak at runtime

pijyoi:
Rather than penalize non-leaking builds of PySide{2,6}, we can check for the leak and only use the workaround for leaking builds.

For builds that do leak, the check will leak one instance of `QByteArray` just once for the whole application.


j9ac9k:
I like this solution much more, thanks @pijyoi

PR #3277, opened 2025-03-07
Allow OpenGL without enableExperimental

pijyoi:
There are currently two `GraphicsItem`s that have an OpenGL implementation:
- `PlotCurveItem`
- `PColorMeshItem`

Their use was conditional on `useOpenGL` and `enableExperimental`.

Originally, the OpenGL implementation for `PlotCurveItem` was limited to a small subset of the `PlotCurveItem` api. Now that it supports most of the common use cases, it should be fine to remove it from "experimental" status.
(The example `MouseSelection` exercises much of the implemented functionality)

`PColorMeshItem`'s implementation was complete from the beginning, but it also required `enableExperimental`, but only to follow `PlotCurveItem`.

Note that `PyOpenGL` is not required at all to use this OpenGL mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Through 3rd party libraries, additional functionality may be added to PyQtGraph,
 | Library        | Added functionality |
 |----------------|-|
 | [`scipy`]      | <ul><li> Image processing through [`ndimage`]</li><li> Data array filtering through [`signal`] </li><ul> |
-| [`pyopengl`]   | <ul><li> 3D graphics </li><li> Faster image processing </li></ul> |
+| [`pyopengl`]   | <ul><li> 3D graphics </li></ul> |
 | [`h5py`]       | <ul><li> Export in hdf5 format </li></ul> |
 | [`colorcet`]   | <ul><li> Add a collection of perceptually uniform colormaps </li></ul> |
 | [`matplotlib`] | <ul><li> Export of PlotItem in matplotlib figure </li><li> Add matplotlib collection of colormaps </li></ul> |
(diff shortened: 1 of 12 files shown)


pijyoi:
The commit that changes `glBlendFunc` to `glBlendFuncSeparate` fixes an issue that is most evident when running under WSL. e.g. `GLVolumeItem` example will be see-through such that the contents behind the window will bleed through.
e.g. OpenGL lines with anti-aliasing enabled will also be see-through.

The blend equation `SRC_ALPHA * SRC_CHAN + (1 - SRC_ALPHA) * DST_CHAN` is not the correct action to perform on the alpha channel. An originally opaque destination pixel blended with a source pixel with transparency will result in a pixel with transparency, which is not the intention.

pijyoi:
The following script demonstrates the issue that the last commit is fixing.
This issue is actually present in 0.13.7 and before.

The issue is that the rendering performed by `GLScatterPlotItem` assumes that blending is enabled. The natively square point is given a circular alpha, which then relies on blending to achieve a circular point.
This is fixed by simply using the `discard` command at the areas outside the circle.

```python
import numpy as np
import pyqtgraph as pg
import pyqtgraph.opengl as gl

pg.setConfigOptions(background=(128, 128, 128))

pg.mkQApp()
win = gl.GLViewWidget()

pos = np.array([[0, 0, 0]])
item = gl.GLScatterPlotItem(pos=pos, size=10, color=(1.0, 0.0, 0.0, 0.5), pxMode=False)
item.setGLOptions('opaque')

win.addItem(item)
win.show()
pg.exec()
```
before the fix:
red_square

after the fix:
red_circle

j9ac9k:
Thanks for your work in this area @pijyoi this looks great.

PR #3278, opened 2025-03-08
Fix recarray support in PlotDataItem

### LINKED ISSUE #3275: TypeError when using a record array to plot data
### Short description

Using a NumPy record array as data to plot results in the following:
`TypeError: Cannot compare structured or void to non-void arrays.`

### Code to reproduce

```python
import pyqtgraph as pg
import numpy as np

data_array = np.recarray((10,), dtype=[('x', float), ('y', float)])
pg.plot(data_array)

pg.exec()
```

### Expected behavior

According to the API reference, a PlotDataItem can be initialized with a NumPy recarray with `dtype=[('x', float), ('y', float), ...]`, so the above code should plot the (uninitialized) recarray data.

### Real behavior

```
  File "C:\Users\_\code\recarray_test.py", line 5, in <module>
    pg.plot(data_array)
  File "C:\Users\_\code\.venv\lib\site-packages\pyqtgraph\__init__.py", line 392, in plot
    w.plot(*args, **dataArgs)
  File "C:\Users\_\code\.venv\lib\site-packages\pyqtgraph\graphicsItems\PlotItem\PlotItem.py", line 630, in plot
    item = PlotDataItem(*args, **kargs)
  File "C:\Users\_\code\.venv\lib\site-packages\pyqtgraph\graphicsItems\PlotDataItem.py", line 375, in __init__
    self.setData(*args, **kargs)
  File "C:\Users\_\code\.venv\lib\site-packages\pyqtgraph\graphicsItems\PlotDataItem.py", line 719, in setData
    if 'x' in data:
TypeError: Cannot compare structured or void to non-void arrays.
```

### Tested environment(s)

 * PyQtGraph version: 0.13.7
 * Qt Python binding: PyQt6 6.8.1 Qt 6.8.2
 * Python version: 3.9.13
 * NumPy version: 2.0.2
 * Installation method: pip

ixjlyons:
Fixes #3275 

I'm not sure if this ever worked. As far as I know, `recarray` hasn't supported this kind of membership check, and it doesn't make sense semantically for normal `ndarray`.

During my testing, I figured normal arrays with named fields might also be worth supporting. The `dataType` helper function identifies them as "recarray", so the modified code path handles both. Only the `recarray` subclass implements attribute access to the fields, otherwise `getattr(data, "x", None)` would've been slightly more concise.

### DIFF
--- a/tests/graphicsItems/test_PlotDataItem.py
+++ b/tests/graphicsItems/test_PlotDataItem.py
@@ -81,7 +81,19 @@ def test_setData():
     pdi.setData([],[])
     assert pdi.xData is None
     assert pdi.yData is None
-    
+
+    # recarray (issue #3275)
+    data = np.recarray((10,), dtype=[('x', float), ('y', float)])
+    pdi.setData(data)
+    assert all(pdi.xData == data["x"])
+    assert all(pdi.yData == data["y"])
+
+    # array with named fields
+    data = np.array([(1, 2), (3, 4), (5, 6)], dtype=[("x", float), ("y", float)])
+    pdi.setData(data)
+    assert all(pdi.xData == data["x"])
+    assert all(pdi.yData == data["y"])
+
 def test_nonfinite():
     def _assert_equal_arrays(a1, a2):
         assert a1.shape == a2.shape
--- a/tests/graphicsItems/test_PlotDataItem.py
+++ b/tests/graphicsItems/test_PlotDataItem.py
@@ -84,6 +84,8 @@ def test_setData():
 
     # recarray (issue #3275)
     data = np.recarray((10,), dtype=[('x', float), ('y', float)])
+    data["x"] = np.linspace(0, 1, len(data))
+    data["y"] = np.linspace(10, 20, len(data))
     pdi.setData(data)
     assert all(pdi.xData == data["x"])
     assert all(pdi.yData == data["y"])
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -1160,7 +1160,12 @@ class PlotDataItem(GraphicsObject):
             elif dt == 'Nx2array':
                 x = data[:, 0]
                 y = data[:, 1]
-            elif dt == 'recarray' or dt == 'dictOfLists':
+            elif dt == 'recarray':
+                if "x" in data.dtype.names:
+                    x = data["x"]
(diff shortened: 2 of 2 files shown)


tests/graphicsItems/test_PlotDataItem.py
```diff
@@ -81,7 +81,19 @@ def test_setData():
     pdi.setData([],[])
     assert pdi.xData is None
     assert pdi.yData is None
-    
+
+    # recarray (issue #3275)
+    data = np.recarray((10,), dtype=[('x', float), ('y', float)])
+    pdi.setData(data)
+    assert all(pdi.xData == data["x"])
+    assert all(pdi.yData == data["y"])
```
pijyoi, line 89: I don't think we should be performing this comparison as recarray is created empty. NANs may have been created. Then the CI may randomly fail? 

Better to create an initialised recarray.

But why even add support for such an input type if it has never worked before? The user could just pass in data.x and data.y as inputs instead?

ixjlyons: Yep, good point on the uninitialized data - was mostly just using the issue as a test case but there's not much reason to do so.

It's a documented feature, but that is an interesting point that if it hasn't actually worked in the past, we could just remove it. I could go either way on that to be honest.

j9ac9k:
I think I came across this not working when doing the documentation for PlotDataItem a while back...  I don't have a particularly strong opinion about whether or not `PlotDataItem` should accept `recarray` or not, but I am of the opinion that `setData` accepts data in _way_ too many formats, and we should start deprecating some of the more weird ones (or at the very least, having tests to test all the supported input cases).

I'm trying to think of a good argument for why `recarray` _should_ be supported that doesn't involve "well, it isn't that much code".   Is there a functionality that is unlocked by supporting it.

ixjlyons:
> I'm trying to think of a good argument for why `recarray` _should_ be supported that doesn't involve "well, it isn't that much code". Is there a functionality that is unlocked by supporting it.

@hjholmberg care to weigh in on this?

hjholmberg:
I was using `recarray` as a convenient way to store extra information for each data point. I have no particular reason for supporting it because (as pijyoi said) it's very easy to just pass in `data.x` and `data.y` instead.

j9ac9k:
Merging as this fixes an issue that we have in our documentation as working (but doesn't).  The discussion about whether we *should* support this is for another time.  Thanks for the report @hjholmberg and thanks for the fix @ixjlyons

PR #3283, opened 2025-03-20
DO NOT MERGE - Fix ParameterTree warnings

### LINKED ISSUE #2994: ParameterTree Example Generated Deprecation Warning
Run the parameter tree example, at the top of the console output, you will get the following `DeprecationWarning`

```
/Users/ogi/Developer/pyqtgraph/pyqtgraph/examples/parametertree.py:28: DeprecationWarning: Parameter has no default value. Pass a default, or use setDefault(). This will no longer set an implicit default after January 2025.
  pTypes.GroupParameter.__init__(self, **opts)
```

Listing dependencies for completion, but this happens on all supported platforms and dependency combinations.

pyqtgraph version: 0.13.4
numpy version: 1.26.4
qt-bindings: pyqt6 6.2.3

j9ac9k:
Fixes #2994 

A while ago we rolled out a change to Parameter Items that required a `default` value be specified if a `value` was specified.  We emitted warnings when this condition wasn't met; but our own test suite (and library usage in general) did not conform to this new requirement.  I pushed back the deprecation period until August 2025 (previously January 2025).

### DIFF
--- a/pyqtgraph/examples/_buildParamTypes.py
+++ b/pyqtgraph/examples/_buildParamTypes.py
@@ -43,7 +43,7 @@ def makeMetaChild(name, cfgDict):
     children = []
     for chName, chOpts in cfgDict.items():
         if not isinstance(chOpts, dict):
-            ch = Parameter.create(name=chName, type=chName, value=chOpts)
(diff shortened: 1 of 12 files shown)


j9ac9k:
Superseded by #3286

PR #3284, opened 2025-03-24
Fix 'GridItem.generatePicture' for cases of invalid size

bbc131:
This commit fixes errors in `GridItem` that can occur when the widget is not visible or has an invalid size.

I encountered a
`RuntimeWarning: divide by zero encountered in log10`
in `GridItem.generatePixmap` at this line:
```
d = 10. ** np.floor(np.log10(np.abs(dist/nlTarget)) + 0.5)
```
However, I cannot reproduce this issue. But if the bounding rect is null (`boundingRect()` returns `QtCore.QRectF()`), `dist` becomes zero, triggering the warning.

While trying to reproduce this issue, I encountered another error:
`AttributeError: 'NoneType' object has no attribute 'determinant'`.
This can be reproduced by adding:
```
grid = pg.GridItem()
p1.addItem(grid)
```
to `examples/InfiniteLabel.py` at line 20. Running the example and resizing the window to its minimal possible size will trigger the error.

Both issues occur when the widget is not in a state where it can be safely painted.
This fix prevents drawing in such cases.

### DIFF
--- a/pyqtgraph/graphicsItems/GridItem.py
+++ b/pyqtgraph/graphicsItems/GridItem.py
@@ -97,14 +97,20 @@ class GridItem(UIGraphicsItem):
         if self.picture is None:
             #print "no pic, draw.."
             self.generatePicture()
-        p.drawPicture(QtCore.QPointF(0, 0), self.picture)
+        if self.picture is not None:
+            p.drawPicture(QtCore.QPointF(0, 0), self.picture)
         #p.setPen(QtGui.QPen(QtGui.QColor(255,0,0)))
         #p.drawLine(0, -100, 0, 100)
         #p.drawLine(-100, 0, 100, 0)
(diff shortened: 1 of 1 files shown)


j9ac9k:
Hey @bbc131 

Thanks for the PR; I have no issue with the diff; but I'm side-eyeing returning `None` from `.deviceTransform()`.  I'm going back and forth on whether it would be better to return the device transform even if it is singular instead of making `.deviceTransform` return an optional.

bbc131:
I'm not sure which approach would be better...

Always returning a `QTransform` seems like the right choice and is more consistent with Qt classes. However, it could also be more error-prone, if you're not familiar with these details. It's obvious that `None` shouldn't be used for anything. But it might be less clear that you need to check the `QTransform` object before inverting it.

PR #3288, opened 2025-03-25
Fix issue with slider span being set to default values when optChanged is called

mbqnano:
Fixed: When optChanged is called without span related options being included (eg. only enable option), the span is reset to default values

### DIFF
--- a/pyqtgraph/parametertree/parameterTypes/slider.py
+++ b/pyqtgraph/parametertree/parameterTypes/slider.py
@@ -95,20 +95,21 @@ class SliderParameterItem(WidgetParameterItem):
             # This is fine, since errors are from the parent scope which will stabilize after the widget is
             # constructed anyway
             pass
-        span = opts.get('span', None)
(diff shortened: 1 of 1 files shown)


pyqtgraph/parametertree/parameterTypes/slider.py
```diff
@@ -95,20 +95,21 @@ def optsChanged(self, param, opts):
...
-            span = np.arange(start, stop + step, step)
-        precision = opts.get('precision', 2)
-        if precision is not None:
-            span = span.round(precision)
-        self.span = span
-        self.charSpan = np.char.array(span)
-        w = self.slider
-        w.setMinimum(0)
-        w.setMaximum(len(span) - 1)
+        if not hasattr(self,'span') or 'limits' in opts or 'step' in opts or 'span' in opts:
```
ntjess, line 98: I feel it will be more readable/maintainable to move this check to after the `suffix` check, and early return on false

PR #3292, opened 2025-03-27
Lazy evaluation of mkColor error string

jdranczewski:
At the start of `mkColor`, an error message is constructed just in case an Exception needs to be raised. This is fair enough, as the message is then used in a couple of different places within the function.

Unfortunately this process involves an array to string conversion, and this does add up, especially when setBrush is used on a scatter plot with different colours for every point, where mkColor is called for _every_ point. In my testing with around 200 points, perceptible lag is introduced, and around 50% of the `setBrush` call time is actually used up generating these error messages that never get used.

I propose to switch the error message to lazy evaluation - it's still defined at the top, but only evaluated when the error message needs to be raised.

Only problem I can see with this is if `mkColor` changes the value of `args` at any point, the value shown in the error message will not match the arguments provided by the user. I don't see any code modifying the values in `args` within `mkColor` though.

Example benchmarking code, run within Jupyter for convenience, with the `snakeviz` extension for profiling:
```python
import pyqtgraph as pg
import numpy as np
%load_ext snakeviz
%gui qt

w = pg.PlotWidget()
w.show()

w.addItem(scatter := pg.ScatterPlotItem(
    np.random.random(300),
    np.random.random(300)
))
colours = np.random.random((300, 4))*255
%snakeviz scatter.setBrush(colours)
```
Results: `setBrush` takes 0.170s, of which `mkColor` is 0.117s, and `_array_repr_implementation` is 0.0974s (54% of `setBrush`).

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -267,7 +267,7 @@ def mkColor(*args) -> QtGui.QColor:
      QColor          QColor instance; makes a copy.
     ================ ================================================
     """
-    err = 'Not sure how to make a color from "%s"' % str(args)
+    err = lambda: 'Not sure how to make a color from "%s"' % str(args)
     if len(args) == 1:
         if isinstance(args[0], str):
             c = args[0]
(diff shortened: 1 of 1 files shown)


j9ac9k:
...my goodness.  That error string constructor has been with us since the beginning of the library

image

The lazy constructor is a good way to go; I would say since this error message is just used in 3 places, we just have the string constructor happen in each of the 3 places ... but ... wow... wonder how much of our debugging of ScatterPlotItem performance issues has been because of this...

j9ac9k:
Thank you so much for the PR @jdranczewski  Merging!

PR #3295, opened 2025-04-03
don't relink shader program where it is not needed

pijyoi:
Explicitly binding vertex attrib 0 and relinking the shader program was a workaround needed for `GLGraphicsItem`s that use constant attribute values. i.e. those that make calls to `glVertexAttrib*`

`GLImageItem` and `GLVolumeItem` do not fall into that category but the workaround was copy-and-pasted into their implementation. This PR removes the unneeded code.


j9ac9k:
Thanks for the cleanup @pijyoi Merging

PR #3297, opened 2025-04-04
fix: GLLinePlotItem uploading vbo perpetually

pijyoi:
`GLLinePlotItem` was uploading vbo perpetually as it did not reset its dirty state.


j9ac9k:
Thanks for the fix/correction!

PR #3309, opened 2025-05-15
Fix bug preventing console showing more than one uncaught exception

campagnola:
Console has the ability to display tracebacks for caught exceptions, but will ignore all but the first one. 

To test:
- run examples/console_exception_inspection.py 
- uncheck "only uncaught exceptions"
- click "raise and catch exception"; traceback is shown
- click "clear stack"
- click "show next exception"
- click "raise and catch exception" again; traceback _should be_ shown again


j9ac9k:
Thanks @campagnola LGTM, merging

PR #3314, opened 2025-05-28
Alignment option for GLTextItem

swvanbuuren:
This commit adds the option for aligning `GLTextItem` instances. Alignment is set using a new input argument in the form of a `QtCore.Qt.AlignmentFlag`. The corresponding `GLTextItem` example has been updated to also demonstrate the alignment feature.

I plan to also contribute 3D grid axes ([as demonstrated in the showcase section](https://github.com/pyqtgraph/pyqtgraph/discussions/2588#discussioncomment-11258934)), which requires this feature.


j9ac9k:
Thanks for the PR @swvanbuuren  it's good to see the 3D plots getting some attention! Diff LGTM, merging.

Issue #3250, opened 2025-02-17
Add option to hide grid on minor ticks


mfeingesicht:
### Short description
I have recently updated from PyQtGraph version 0.13.3 to version 0.13.7. Users of the software I write have noticed that the plots using version 0.13.7 now have vertical/lines for all minor ticks instead of only major ticks as in version 0.13.3. Would it be possible to have the option to hide the grid on minor ticks ?

### Code to reproduce
```python
import pyqtgraph as pg
from PyQt5.QtGui import QGuiApplication

# Create window
win = pg.GraphicsLayoutWidget(show=True)
win.setWindowTitle('pyqtgraph example: Scrolling Plots')

# Add plot
axis = win.addPlot(1, 1, pen=pg.mkPen(color='#000000'))

# Show grid
axis.showGrid(x=True, y=True)

QGuiApplication.exec_()
```

### Tested environment(s)

 * PyQtGraph version: 0.13.3 (grid only on major ticks), 0.13.7 (grid on major and minor ticks)
 * Qt Python binding: PyQt5 5.15.6 Qt 5.15.2
 * Python version: 3.10.1
 * NumPy version: 1.26.4
 * Operating system: Windows 10
 * Installation method: pip

PR #3318, opened 2025-06-02
AxisItem: Allow hiding minor ticks

2xB:
Fixes pyqtgraph#3250, which originated from #2827 (therefore maybe @NilsNemitz has an opinion on this?).

This can be used as follows, adapting the example from the above issue:

```python3
import pyqtgraph as pg

app = pg.mkQApp()

# Create window
win.setWindowTitle('pyqtgraph example: Scrolling Plots')

# Add plot
plot = win.addPlot(1, 1, pen=pg.mkPen(color='#000000'))

# Show grid
plot.showGrid(x=True, y=True)
#plot.getAxis('bottom').style["maxTickLevel"] = 0 # replaced by proper setter thanks to comment by NilsNemitz
#plot.getAxis('left').style["maxTickLevel"] = 0
plot.getAxis('bottom').setStyle(maxTickLevel=0)
plot.getAxis('left').setStyle(maxTickLevel=0)

app.exec()
```

@mfeingesicht Could you test if this solves your use case?


NilsNemitz:
Hi @2xB, thank you for adding internal logic for the sub ticks. The code should absolutely have that!

There's probably a follow-up discussion to be had what a reasonable API might look like, since currently `AxisItem` doesn't really seem to expect the user to interact with its settings very much. For the moment communicating through the `style` property looks surprisingly reasonable.

#2827 tried to touch the overall logic as little as possible. It tried to adjust the math to produce a helpful number of ticks (e.g. more than one, I think) in a wider range of cases. The same adjustments also allowed the existing minor tick logic to actually produce visible ticks more regularly, I believe that produced the change that #3250 points out.

There's no deeper meaning to the appearance of extra ticks.

NilsNemitz:
@2xB this looks good to me and is a nice, efficiently minimal change.

Could you maybe add a short description of the `maxTickLevel` parameter to the documentation of `setStyle()`?

2xB:
@NilsNemitz Sure, I added a short documentation to `setStyle`!

NilsNemitz:
Thank you!

Given the name of the `maxTickLevel` parameter, I think it might be better if the documentation discusses major ticks and minor ticks here, rather than the grid. Although considering that the grid motivated the PR, maybe
```
 - 0: Show only major ticks
 - 1: Show major ticks and one level of minor ticks
 - 2: Show two levels of minor ticks (higher CPU usage) 

This setting also affects the displayed grid.
```
might be a reasonable inclusion?

2xB:
Sure! I minorly rephrased it so the grid is mentioned in a concise description at the beginning of the text.

NilsNemitz:
Hi @2xB , we've been having some trouble with the tests, which might be better now.

If you make another change, the tests should rerun automatically, and we'll see. On that note, maybe you could go to line 234 and change 
```
``float``,``int`` or ``None``               to
``float``, ``int`` or ``None``    
```
?
It's not related to your PR, but it is right next to your maxTickLevel documentation. I hope that will fix the rendering of [int] in the documentation.

PR #3330, opened 2025-06-19
Fixed bug that significantly degraded zooming performance for line plots

oyvindlr:
I noticed when going from the latest stable release (13.7) to the development version, that the performance of zooming in a large line plot (using peak auto-downsampling, but I'm unsure if that was relevant) was significantly worse. 
By binary searching commits for the issue, I found that it occured after commit no  d13e2ce, which should be purely about documentation and linting. 
The offending line caused "all_finite" to never be updated, and stay as None forever.


pijyoi:
Nice catch! https://github.com/pyqtgraph/pyqtgraph/commit/d13e2ce27cba1ff45fb5128e7f5c553f1b97cdaa effectively undid the optimization made in #2720

PR #3331, opened 2025-06-19
Improve DateAxisItem skipFactor calculation

Servall4:
Replace local array copy with int multiplier.

### DIFF
--- a/pyqtgraph/graphicsItems/DateAxisItem.py
+++ b/pyqtgraph/graphicsItems/DateAxisItem.py
@@ -129,13 +129,13 @@ class TickSpec:
     def skipFactor(self, minSpc):
         if self.autoSkip is None or minSpc < self.spacing:
             return 1
-        factors = np.array(self.autoSkip, dtype=np.float64)
+        multiplier = 1
         while True:
-            for f in factors:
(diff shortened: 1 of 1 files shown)


oyvindlr:
If you're optimizing this, might as well try to skip the outer loop altogether. Something like

```python
def skipFactor(self, minSpc):
    if self.autoSkip is None or minSpc < self.spacing:
            return 1
    minSkipFactor = minSpc/self.spacing
    multiplier = 1
    #autoskip is sorted, so autoskip[-1] is the largest value
    if self.autoskip[-1]< minSkipFactor:
        multiplier = minSkipFactor/self.autoskip[-1] #The exact multiplier needed for multiplier*autoskip[-1] to be equal to minSkipFactor
        multiplier = 10**np.ceil(np.log10(multiplier)) #The closest power of 10 that causes multiplier*autoskip[-1] to be larger than minSkipFactor
    index = numpy.searchsorted(self.autoskip, minSkipFactor/multiplier) #First index in autoskip*multiplier which gives a larger value than minSkipFactor
    return self.autoskip[index] * multiplier
```
Note: I have not tested this

PR #3333, opened 2025-06-20
ensure comparison returns bool instead of np.bool

pijyoi:
Numpy 2.3.0 includes an expired deprecation: https://github.com/numpy/numpy/pull/28254
`'np.bool' scalars can no longer be interpreted as an index (deprecated since 1.19) `

This triggers an error in `pg.TableWidget` when a structured array is used, as demonstrated by the `TableWidget` example itself. (https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/examples/TableWidget.py)

On PyQt6:
`TypeError: invalid result from TableWidgetItem.__lt__(), a 'bool' is expected not 'numpy.bool'`

On PySide6:
`TypeError: 'numpy.bool' object cannot be interpreted as an integer`

### DIFF
--- a/.github/workflows/etc/environment-pyside.yml
+++ b/.github/workflows/etc/environment-pyside.yml
@@ -2,7 +2,7 @@ name: test
 channels:
   - conda-forge
 dependencies:
-  - pyside6
+  - pyside6[version='!=6.9.1']
   - numpy
   - nomkl
   - scipy
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,13 +76,13 @@ jobs:
             qt-version: "PyQt6"
           - python-version: "3.12"
             qt-lib: "pyside"
-            qt-version: "PySide6-Essentials"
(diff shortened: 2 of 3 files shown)


pijyoi:
Following is an MWE that demonstrates the error when upgrading to NumPy 2.3.0

```python
import numpy as np
from PySide6 import QtCore, QtWidgets

class Item(QtWidgets.QTableWidgetItem):
    def __init__(self, value):
        super().__init__(str(value))
        self.value = value

    def __lt__(self, other):
        return self.value < other.value

data = np.arange(10).reshape((5, 2))
nrows, ncols = data.shape

wgt = QtWidgets.QTableWidget(nrows, ncols)

for row in range(nrows):
    for col in range(ncols):
        wgt.setItem(row, col, Item(data[row, col]))

wgt.sortByColumn(0, QtCore.Qt.SortOrder.DescendingOrder)

wgt.show()
app.exec()
```

PR #3339, opened 2025-06-26
Update syntax.py to correctly colorize commented code

kevinanewman:
Strings with '#' in them, or keywords, were not highlighted correctly, two-pass approach handles this issue.


j9ac9k:
Thanks @kevinanewman for the PR!

PR #3349, opened 2025-07-31
Add overlay grid option to GLSurfacePlotItem

swvanbuuren:
This PR adds the possibility to overlay rectangular grids on 3D surfaces as provided by `GLSurfacePlotItem`. It also extends the current example of 3D surface plots to showcase the rectangular grids.

Here's a screenshot of the updated example:
image

The current implementation already features adding a wireframe to a 3D surface plot, but unfortunately the grid lines don't look nice and it's only possible to show the actual triangle mesh wireframes and not rectangular grids, which IMO is useful to visualize and analyze data.

To nicely overlay the grid, the only feasible I could find, is to create additional `GLLinePlotItem`s that are added to the scene. To have this working, I added the `_setView` method, to make sure the view is available. As soon as this is the case, the grid lines are added to the scene. For proper display, certain OpenGL options need be set for the `GLMeshItem` and `GLLinePlotItem`.

Not sure if there would be a better way to handle this, but at least this seems to do the trick. I am open for suggestions for improvements or alternatives to this approach.

~Also note, that adding rectangular grids to animated surfaces imposes a serious performance drop. Because of this, I didn't add the grid to animated surface in the example. Maybe we should a note about this somewhere, although I am note sure where.~ (With the latest changes, this doesn't seem to be an issue anymore)

### DIFF
--- a/pyqtgraph/opengl/items/GLSurfacePlotItem.py
+++ b/pyqtgraph/opengl/items/GLSurfacePlotItem.py
@@ -47,7 +47,7 @@ class GLSurfacePlotItem(GLMeshItem):
         # in GLViewWidget.drawItemTree(), at the same depth value, child items
         # come before the parent. make it such that our grid lines get drawn
         # after the surface mesh.
-        self.lineplot.setDepthValue(1)
+        self.lineplot.setDepthValue(self.depthValue() + 1)
         self.setParentItem(parentItem)
 
         self.setData(**surface_kwds)


pyqtgraph/opengl/items/GLSurfacePlotItem.py
```diff
@@ -133,3 +178,53 @@ def generateFaces(self):
...
+        tr = self.viewTransform()
+        opts = {
+            'glOptions': self.gl_line_options,
+            'antialias': self._lineAntialias,
+            'color': self._lineColor,
+            'width': self._lineWidth,
+        }
+
+        def map_pts(arr):
+            return np.vstack([
```
pijyoi, line 202: It seems like you should be able to vectorize this. See https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/Transform3D.py#L38

pyqtgraph/opengl/items/GLSurfacePlotItem.py
```diff
@@ -2,34 +2,73 @@
 
 from ..MeshData import MeshData
 from .GLMeshItem import GLMeshItem
+import pyqtgraph as pg
```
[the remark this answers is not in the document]
swvanbuuren: Remove with [8c9966d](https://github.com/pyqtgraph/pyqtgraph/pull/3349/commits/8c9966dae40e39297299f8d4138e63b4c31aff4a)

pyqtgraph/opengl/items/GLSurfacePlotItem.py
swvanbuuren: @pijyoi where you thinking about something like this: [ed84150](https://github.com/pyqtgraph/pyqtgraph/pull/3349/commits/ed84150338858cac1e180a6e747bdf5688ed4ad4)?
It seems to more performant at least. Enabling the grid for the animated surface (in the example) runs much smoother.

pijyoi:
It should be possible to use a single GLLinePlotItem in "lines" mode (default is "line-strip") to draw the entire grid.

GLGridItem uses this technique.

This is equivalent to the "pairs" mode of PlotCurveItem.

pyqtgraph/opengl/items/GLSurfacePlotItem.py
pijyoi: Yes.
But looking closer at the code now, you shouldn't even need to compute the transform manually.
If GLLinePlotItem is set as a child of GLSurfacePlotItem, the effective transformation matrix will be sent to OpenGL during rendering of the lines.

The relevant lines performing this is at:
https://github.com/pyqtgraph/pyqtgraph/blob/6eb0bf1a928c3e8fef332bbebe8a9da3be0ab19a/pyqtgraph/opengl/GLViewWidget.py#L263-L267

swvanbuuren: Yes, you're right. By setting `GLLinePlotItem` as a child of `GLSurfacePlotItem`, the transformation isn't required anymore and code simplifies significantly.

Actually, I already tried this (before submitting this PR), but in this case, it's not possible to enable anti-aliasing for the `GLLinePlotItem`s (this possibly also holds for other GL options). I couldn't find a solution for this. Maybe you could point me in the right direction?

swvanbuuren:
> It should be possible to use a single GLLinePlotItem in "lines" mode (default is "line-strip") to draw the entire grid.
> 
> GLGridItem, GLGraphItem use this technique.
> 
> This is equivalent to the "pairs" mode of PlotCurveItem.

I will look into this and will try to come up with a version which only calls `GLLinePlotItem` once with `mode='lines'`.

pyqtgraph/opengl/items/GLSurfacePlotItem.py
pijyoi: I tried playing with your PR. The trick seems to be to make the grid lines draw after the surface mesh.
https://github.com/pijyoi/pyqtgraph/commit/4931b5137b3b4530a3413d06a9ba43b9a13a40a8

On the master branch (which your PR is based-off of anyway), changing options via glOptions is sticky. That is to say, any options set will persist for other GLGraphicsItems. It is the responsibility of the developer to unset any options that they have set.

In my experience, enabling line-antialiasing makes the rendered image look bad on some hardware. (like the Raspberry Pi)

pijyoi: The 3 standard options (opaque, additive and translucent) don't need to be reset because all 3 of them define the same set of options (blend, cull, depth) to set.

pyqtgraph/opengl/items/GLSurfacePlotItem.py
```diff
@@ -2,34 +2,58 @@
...
+        self._lineWidth = 1.0
+        self._lineAntialias = False
         self._vertexes = None
         self._meshdata = MeshData()
-        super().__init__(parentItem=parentItem, meshdata=self._meshdata, **kwds)
-        
-        self.setData(x, y, z, colors)
-        
-        
+        self._grid_lines = []
```
pijyoi, line 36: unused member variable

pyqtgraph/opengl/items/GLSurfacePlotItem.py
```diff
@@ -2,34 +2,58 @@
...
+            if arg in kwds:
+                surface_kwds[arg] = kwds.pop(arg)
+
+        super().__init__(meshdata=self._meshdata, **kwds)
+        
+        self.lineplot = GLLinePlotItem(parentItem=self, mode='lines', glOptions='translucent')
+        # in GLViewWidget.drawItemTree(), at the same depth value, child items
+        # come before the parent. make it such that our grid lines get drawn
+        # after the surface mesh.
+        self.lineplot.setDepthValue(1)
```
pijyoi, line 51: I am not sure, is render order still important with the use of POLYGON_OFFSET_FILL?

In any case, to be robust, the depth value of the line plot should be set relative to the surface plot's depth value, not hardcoded as 1.

I suspect that the implemented behavior of children being drawn before parent at the same depth value was unintentional. For `QGraphicsItem`, the docs say that `parents are always drawn before their children`

pyqtgraph/opengl/items/GLSurfacePlotItem.py
swvanbuuren: Removed

pyqtgraph/opengl/items/GLSurfacePlotItem.py
swvanbuuren: I changed the code to set the depth value relative to that of the surface.

This is needed when anti-aliasing is enabled, i.e. when I don't set the depth value, anti-aliasing fails.

Without setting the depth value, anti-aliasing is not activated.

Issue #3300, opened 2025-04-10
cannot mouse zoom-in when aspect ratio locked


pijyoi:
### Short description

When aspect ratio is set to locked, using mouse move with right-button pressed no longer zooms properly.
The PlotItem can only be zoomed-out but can no longer be zoomed-in.

### Code to reproduce

```python
import pyqtgraph as pg
import numpy as np

Z = np.linspace(0, 1, 12).reshape((3, 4))

pg.mkQApp()
win = pg.PlotWidget()
win.setAspectLocked(True)
pcmi = pg.PColorMeshItem(Z)
win.addItem(pcmi)
win.show()
pg.exec()
```

### Additional context

Bisecting points to 55d90aa04bf1d77a818fc7146802bd8c7b693440 (2024/07/16) as the culprit (PR #3100).

pijyoi:
PR #3100 seems to have changed the logic from:
```python
if self.state['aspectLocked'] is not False:
    scale[0] = scale[1]
```
to:
```python
if self.state['aspectLocked'] is not False:
    if x is None:
        scale[0] = scale[1]
    if y is None:
        scale[1] = scale[0]
```

It is only when scrolling the mouse wheel while over an axis that "x" or "y" can be `None`.
In the case of moving the mouse while the right-mouse button is pressed, both "x" and "y" are not `None`.
So perhaps a possible fix is to change the code to:

```python
if self.state['aspectLocked'] is not False:
    if x is None:
        scale[0] = scale[1]
    if y is None:
        scale[1] = scale[0]
    if x is not None and y is not None:
        scale[0] = scale[1]
```

PR #3354, opened 2025-08-11
fix aspect-locked zoom with mouse-drag while right-button pressed

pijyoi:
This fixes issue #3300, a regression bisected to #3100.

This issue arises in the following conditions:
1) aspect-ratio locked
2) mouse zoom by dragging the mouse with right-button pressed


j9ac9k:
Thanks for the fix @pijyoi, LGTM, merging!

PR #3370, opened 2025-09-13
adapt to changes in QPainterPathPrivate in Qt 6.10

pijyoi:
In Qt 6.10 (https://github.com/qt/qtbase/commit/adcbac38cec5a849af6f2ee43e59b4f905ff5142), `QPainterPathPrivate` goes from
```c++ 
    QExplicitlySharedDataPointer<QPainterPathPrivate> d_ptr;
```
to
```c++
    QPainterPathPrivate *d_ptr;
```

`pyqtgraph`'s CI pipeline will start failing once Qt 6.10 is available but only affects users who set `enableExperimental=True`.
This PR adds support for this new data structure.


j9ac9k:
Thanks @pijyoi for looking out and testing ahead of the 6.10 release so we don't get unexpected breakages!

PR #3380, opened 2025-10-02
Annotate ViewBox slots

moritz-h:
Same as #3123, seems these two slots were missing.

Fixes PySide6 warning `qt.pyside.libpyside: Warning: Registering dynamic slot "linkedXChanged()" on "ViewBox". Consider annotating with @Slot()`


j9ac9k:
Thanks for the PR @moritz-h !

Issue #218, opened 2015-08-19
improve dock tabs look-and-feel


sbourdeauducq:
The look and feel of tabs in applications like Google Chrome or Sublime Text is better IMO. Drag-and-drop of tabs to change their order should also be supported.

Also, they should follow the user's color scheme - the hardcoded blue colors do not always look the best.

j9ac9k:
I'm going to close out this issue as it's not quite descriptive enough.  Our docks are customizable, and I will admit the defaults aren't great.  We would certainly welcome a PR to make them look more modern.

sem-geologist:
I  think what @sbourdeauducq meant with "tabs" is actually `DockLabels`. Colors shapes and border color is hardcoded which makes it look acceptable in some Qt Themes, and in other it looks completely out of place. Currently I am preparing a Poster to promote my software which uses `pyqtgraph` and found that Docking tab color  bites with the rest of whole theme (It had not bothered me so much to use everyday, but making presentation for other people made me realize that), and so I dived into the code to find out if this can't be somehow changed. As I want to be ready also to show that my software also can be dark themed or light themed, I looked if colors could not be looked up and picked up dynamically from `.palette()`. It indeed could to do that easily as DockLabel class has that getter inherited!

The Modern look - To be honest I am not sure what that is. It has round corners - some would say that is modern other would say that it is already _passe_ - it is kind of taste thing. As I am used to DarkTheme Breeze KDE Neon or Light Breeze themes, I would move the border line to top and make it the highlighting element, where background and foreground would pick that from `.pallete().color()` with `QPalette.Window` and `QPalette.WindowText` enumerators.

I could try opening PR with such improvements. But then why it was not done like that from the beginning? maybe @campagnola wanted it to stand-out like this? I think blue colors used in pytgraph.org webpage is with similar hue, and InfiniteLine range or other default colors are also somehow connected.

Now draggable tabs are not so easy to achieve. We should remember that this pyqtgraph implemented Docking Framework is minimalistic - and that I think is the point.
If all bells and whistles are needed then there are something like this (a full blown library dedicated for docking and tabs) https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System , which actually seems to have at last python bindings installable with pip.

sem-geologist:
Oh I see Isomehow missed the #2274

PR #3391, opened 2025-10-15
Add tabColor property to pyqtgraph docks

runrunm:
Hello pyqtgraph,

I was looking at a way to custom the dock colors, everything but the font size is hard-coded (tab background color, font color, border-radius...). Related to [https://github.com/pyqtgraph/pyqtgraph/issues/218](https://github.com/pyqtgraph/pyqtgraph/issues/218).

- This PR makes it possible to choose the tab color by setting the `tabColor` property to either a 6-digit HEX color string `#rrggbb` or a RGB tuple `(r, g, b)` with 0-255 values.
- The luminance of the tab background color is checked and the tab font color is set to black or white accordingly (see Dock 1 has black font, Docks 2 & 3 have white font)
- When 2 docks are stacked, the selected tab has the `tabColor` color (see Dock 3) and the unselected tab is set to a default light grey color (see Dock 4).

For example:
```python
dock1 = Dock("Dock 1", tabColor=(105, 105, 105))
dock2 = Dock("Dock 2", autoOrientation=False, tabColor='#cb5d40')
dock3 = Dock("Dock 3", autoOrientation=False)
dock4 = Dock("Dock 4", autoOrientation=False)
```

image

A possible improvement would be to have a `borderRadius` property which could change the look of docks quite a lot (nice for customization).

### DIFF
--- a/pyqtgraph/dockarea/Dock.py
+++ b/pyqtgraph/dockarea/Dock.py
@@ -262,7 +262,27 @@ class DockLabel(VerticalLabel):
     sigClicked = QtCore.Signal(object, object)
     sigCloseClicked = QtCore.Signal()
 
-    def __init__(self, text, closable=False, fontSize="12px"):
+    def __init__(self, text, closable=False, fontSize=12, tabColor='#4a5c96'):
+        self.r = '5px'  # Tab radius
+        if isinstance(tabColor, str) and tabColor.startswith('#') and len(tabColor) ==7:
+            h = tabColor.lstrip('#')      # Remove #
+            r, g, b = tuple(int(h[i:i+2], 16) for i in (0, 2, 4))  # Get 1 color channel per variable (0-255)
+            self.bg = tabColor            # Set tab background color value
+        elif isinstance(tabColor, tuple) and len(tabColor) == 3:
+            r, g, b = tabColor            # Get 1 color channel per variable (0-255)
+            h = f'{r:02x}{g:02x}{b:02x}'  # Get HEX color value from r, g, b
+            self.bg = '#' + h             # Set tab background color value
+        else:
+            raise ValueError(
+                "tabColor must be either a 6-digit HEX color string '#rrggbb'"
+                "or a RGB tuple (r, g, b) with values in 0-255."
+                )
+
+        if (r*0.299 + g*0.587 + b*0.114) > 186:  # Check tab color luminance and set font color accordingly (black or white)
+            self.fg = '#000000'
+        else:
+            self.fg = '#ffffff'
+        
         self.dim = False
         self.fixedWidth = False
         self.fontSize = fontSize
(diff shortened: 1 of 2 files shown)


pyqtgraph/dockarea/Dock.py
```diff
@@ -262,7 +262,27 @@ class DockLabel(VerticalLabel):
     sigClicked = QtCore.Signal(object, object)
     sigCloseClicked = QtCore.Signal()
 
-    def __init__(self, text, closable=False, fontSize="12px"):
+    def __init__(self, text, closable=False, fontSize=12, tabColor='#4a5c96'):
+        self.r = '5px'  # Tab radius
+        if isinstance(tabColor, str) and tabColor.startswith('#') and len(tabColor) ==7:
```
campagnola: I suggest using `pg.mkColor(tabColor)` here rather than implementing your own color parsing.

pyqtgraph/dockarea/Dock.py
```diff
@@ -262,7 +262,27 @@ class DockLabel(VerticalLabel):
     sigClicked = QtCore.Signal(object, object)
     sigCloseClicked = QtCore.Signal()
 
-    def __init__(self, text, closable=False, fontSize="12px"):
+    def __init__(self, text, closable=False, fontSize=12, tabColor='#4a5c96'):
```
campagnola: If you want the default background color / radius / font size changed, that requires a bit more discussion.. I suggest keeping these as they were.

pyqtgraph/dockarea/Dock.py
```diff
@@ -281,15 +301,12 @@ def __init__(self, text, closable=False, fontSize="12px"):
...
-        r = '3px'
         if self.dim:
+            bg = '#ddd'
             fg = '#aaa'
-            bg = '#44a'
-            border = '#339'
         else:
-            fg = '#fff'
-            bg = '#66c'
-            border = '#55B'
```
campagnola: Again here, I woldn't change the defult border color.

pyqtgraph/dockarea/Dock.py
j9ac9k: yeah, definitely use `mkColor` as it handles all sorts of formats.

pyqtgraph/dockarea/Dock.py
```diff
@@ -262,7 +262,27 @@ class DockLabel(VerticalLabel):
...
+            r, g, b = tabColor            # Get 1 color channel per variable (0-255)
+            h = f'{r:02x}{g:02x}{b:02x}'  # Get HEX color value from r, g, b
+            self.bg = '#' + h             # Set tab background color value
+        else:
+            raise ValueError(
+                "tabColor must be either a 6-digit HEX color string '#rrggbb'"
+                "or a RGB tuple (r, g, b) with values in 0-255."
+                )
+
+        if (r*0.299 + g*0.587 + b*0.114) > 186:  # Check tab color luminance and set font color accordingly (black or white)
```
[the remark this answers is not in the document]
runrunm: Would it be okay if the default behavior was to choose tab color and font color manually ?

We could then add an "auto" parameter or be able to set "auto" as an argument of the tabColor parameter.

runrunm:
Taking your comments into consideration, I went back to previous hard-coded style parameters (border color, border radius). 

I am now using mkColor making it much easier to handle color. I am unsure though how to import the function (I used `from ..functions import mkColor`).

I added the `fontColor` parameter for the user to be able to choose the font color. If it is not specified, the most appropriate value of `fontColor` (black or white) is guessed based on tabColor luminance value. If `tabColor` is not specified, it is set to be the default pyqtgraph hex #55B dock color.

PR #3394, opened 2025-10-20
Correctly handle SI-prefixes for nonlinear units

### LINKED ISSUE #3389: Support for inverse length units
Hi,
I need to handle a lot of diffraction data that use inverse length units, e.g., 1/m, 1/nm, etc. The built-in unit conversion functions only handle the real space length but cannot handle the reciprocal space units. It would be nice to add the support for the inverse length units. Thank you!

oyvindlr:
In cases where your unit has a power different from 1, the SI-prefix scaling becomes incorrect. For instance, 1000 m² will be shown as 1 km², which is off by a factor of 1000. And 1000 s^(-1) certainly is not 1 ks^(-1).
I've implemented a "unitPower" parameter that can be used when creating an axis label. The parameter will work for any power, positive, negative and non-integral. 
Fixes #3389 and #3392


pyqtgraph/functions.py
```diff
@@ -152,15 +206,14 @@ def siFormat(x, precision=3, suffix='', space=True, error=None, minVal=1e-25, al
...
+        return fmt % (x*p, pref, suffix, plusminus, siFormat(error, precision=precision, suffix=suffix, space=space, minVal=minVal, power=power))
 
 
 def siParse(s, regex=FLOAT_REGEX, suffix=None):
     """Convert a value written in SI notation to a tuple (number, si_prefix, suffix).
 
-    Example::
-
-        siParse('100 µV")  # returns ('100', 'µ', 'V')
+    Example:
```
campagnola: The original formatting with `::` and the extra blank line are needed for sphinx to render the documentation correctly (see: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks)

j9ac9k:
Whoops, I mishandled the merge ... let me fix that.

j9ac9k:
Thanks for the PR @oyvindlr !

PR #3399, opened 2025-10-23
Two minor corrections

campagnola:
1. Fixed thread name printing in debug.ThreadTrace
2. Converted some flowchart-related values from float to int, due to changes in Qt API


j9ac9k:
👋🏻 @campagnola Thanks for the PR!

PR #3413, opened 2025-11-07
Make SpinBox locale-aware

oyvindlr:
Spinboxes in Qt are locale-aware, in the sense that they use a period or a comma as a decimal separator depending on locale. This pull request fixes pyqtgraph's Si-scaling spinbox so it acts the same way.


pijyoi:
```python
from PyQt6 import QtCore

print('Qt Version', QtCore.QT_VERSION_STR)

loc_def = QtCore.QLocale('en')
loc_eur = QtCore.QLocale('de')

numbers = [123.456789, 123456.789, 123456789.]

for number in numbers:
    s1 = loc_def.toString(number, 'g')
    s2 = loc_eur.toString(number, 'g')
    print(f'{s1}     {s2}')
```

Qt6 versions up to 6.8.x seem to have a bug in capitalizing the exponent character even though 'g' (and not 'G') was specified.
```
Qt Version 6.2.3
123.457     123,457
123,457     123.457
1.23457E+08     1,23457E+08

Qt Version 6.8.2
123.457     123,457
123,457     123.457
1.23457E+08     1,23457E+08

Qt Version 6.9.0
123.457     123,457
123,457     123.457
1.23457e+08     1,23457e+08
```

Another issue exposed here is that there's also the https://doc.qt.io/qt-6/qlocale.html#groupSeparator that is locale dependent.

oyvindlr:
@pijyoi , that's some thorough investigation for the CI-failure! I've updated the tests to expect failure for old versions of Qt. 

Regarding group separator, I did consider it. But it's not supported in the standard Qt DoubleSpinBox,  it wasn't implemented in the previous version of this SpinBox, and I felt it would be too complex to implement for a niche feature of a niche GUI-component.

PR #3418, opened 2025-11-19
Bump up dependencies to conform to SPEC-0

j9ac9k:
Also bumps documentation dependencies and fixes a handful of warnings.

This PR Does not remove pyside2 specific code paths,  but pyside2 is no longer supported (or tested).


j9ac9k:
Ha no it was not! Nice catch, I'll make an update.

PR #3431, opened 2025-12-02
Fix type saving of Parameter sublcalsses with saveState() / restoreState() (issue #3430)

### LINKED ISSUE #3430: Type of Parameter subclasses incorrectly saved with saveState() / restoreState()
### Context
@seb5g  and I are working on the development of PyMoDAQ, which makes use of `Parameter` objects, and we are currently implementing their serialization into a binary format.
For this, we use `saveState()` and `restoreState()`, which should normally allow to save and recreate a parameter.

During this work, we encountered issues with the `ComplexParameter` and `ScalableGroup` parameter subclasses from examples/parametertree.py. 

### Short Description
In examples/parametertree.py, the subclasses `ComplexParameter` and `ScalableGroup` **cannot be correctly restored** after calling `saveState()` and `restoreState()`.
The restored parameter **does not keep its subclass type** because the subclass constructor overwrites the `type` option with an already existing registered type (e.g. 'group').
As a result, `restoreState()` instantiates a `Parameter` instead of an instance of the original subclass (e.g. `ScalableGroup`).

### Code to reproduce
```python
from pyqtgraph.parametertree.utils import compare_parameters
import pyqtgraph.parametertree.parameterTypes as pTypes
from pyqtgraph.parametertree import Parameter

class ScalableGroup(pTypes.GroupParameter):
    def __init__(self, **opts):
        opts['type'] = 'group' #issue: hide the real subclass
        opts['addText'] = "Add"
        opts['addList'] = ['str', 'float', 'int']
        pTypes.GroupParameter.__init__(self, **opts)

    def addNew(self, typ):
        val = {
            'str': '',
            'float': 0.0,
            'int': 0
        }[typ]
        self.addChild(
            dict(name="ScalableParam %d" % (len(self.childs) + 1), type=typ, value=val, removable=True, renamable=True))


p = ScalableGroup(name='Scalable parameter group')
p_state = p.saveState()
restored_p = Parameter.create(**p_state)

print(p) # <ScalableGroup 'Scalable parameter group' at 0x7eaee1b0e3f0>
print(restored_p) # <GroupParameter 'Scalable parameter group' at 0x75061b256030> <= wrong class

print(compare_parameters(p, restored_p)) # False <= should be True
```

### Expected Behavior
Calling `saveState()` and `restoreState()` should **preserve the actual class** of each parameter.
A restored `ScalableGroup` should be an instance of `ScalableGroup`, not a `Parameter`.

### Actual Behavior
The restored object is a `GroupParameter` because:
- The subclass set `type='group'`, which is already mapped to `GroupParameter` in `PARAM_TYPES`.
- `Parameter.create()` looks up the type in `PARAM_TYPES`, finds 'group', and instantiates the wrong class.

Thus, the restored parameter is not equal to the original one, because it doesn't have the same class.

### Cause
Subclass constructors (`ComplexParameter`, `ScalableGroup`) assign an existing registered type for the `type` option. This prevents the registration and restoration of the correct subclass.

### Proposed Solution (See PR #3431)

1. Assign unique parameter types to each custom subclass of `Parameter` (e.g. 'scalablegroup'). Register them in  `PARAM_TYPES` with `registerParameterType('scalablegroup', ScalableGroup)`
2. Add a type check in the `Parameter` constructor to ensure that the class referred by the 'type' option in `PARAM_TYPES` corresponds to the actual class.

### Proposal for architecture improvement
Moreover, @seb5g  and I would like to propose an improvement of the current architecture for parameter types management.
We suggest refactoring the code with a **factory pattern**, in order to provide a better encapsulation and control of PARAM_TYPES, PARAM_NAMES.
This would provide a more robust architecture and a cleaner way to define and register new parameter types.
If this idea is of interest, we would be happy to work on this new design.

MariaF203:
### Description
Resolves #3430

This PR fixes the issue where custom `Parameter` subclasses (e.g. `ComplexParameter` and `ScalableGroup`, defined in examples/parametertree.py) cannot be correctly restored using `saveState()` and `restoreState()`.

The cause of this problem is that these classes overwrite `opts['type']` with existing types (e.g. 'group') already associated to an other `Parameter` subclass. 

Thus, for example, the restore mechanism instantiates a `GroupParameter` instead of the expected subclass `ScalableGroup`. Similarly, it instantiates a `SimpleParameter` instead of a `ComplexParameter`. 

### Summary of changes

**1. Use and register unique parameters types**

We use unique type names ('complexparameter', 'scalablegroup') for the example subclasses `ComplexParameter` and `ScalableGroup` and register them using `registerParameterType()`.

**2. Add a type check in `Parameter`**

Created a new method `check_type()`. It validates that:
- the specified type is registered in `PARAM_TYPES`
- the corresponding registered class matches the actual class.

This method is then called in the `__init__` function of `Parameter`.

**3. Add a `compare_parameters()` function**

Created a new function `compare_parameters()` that allows to compare two parameters. In particular, it checks that the parameters classes are the same.
The existing `eq()` function only compare saved states and cannot compare classes.

**4. Add tests**

- Tests for `compareParameters()`
- Update of test in test_parametertypes.py to check that the restored parameters keep the correct class.

**5. Fix `BoolOrRadioParameter` inconsistency**

A `BoolOrRadioParameter` could represent a 'bool' or a 'radio', but only one class per type can be registered in `PARAM_TYPES`. Thus, it caused inconsistencies. To resolve this:
- Removed the `BoolOrRadioParameter` class.
- Created a new registered class `RadioParameter`, associated to the 'radio' type;
- Updated `ChecklistParameter` so the created list items are either a `BoolParameter` or a `RadioParameter`.

### Conclusion
-  `Parameter` subclasses are now properly saved and restored.
- Type registration is consistent.
- Parameters comparison is now fully coherent.

### DIFF
--- a/pyqtgraph/examples/_paramtreecfg.py
+++ b/pyqtgraph/examples/_paramtreecfg.py
@@ -99,6 +99,9 @@ cfg = {
         }
     },
 
+    'radio': {
+    },
+
     'pen': {
         'Pen Information': {
             'type': 'str',
--- a/pyqtgraph/examples/parametertree.py
+++ b/pyqtgraph/examples/parametertree.py
@@ -16,14 +16,14 @@ from pyqtgraph.Qt import QtWidgets
 
 app = pg.mkQApp("Parameter Tree Example")
 import pyqtgraph.parametertree.parameterTypes as pTypes
-from pyqtgraph.parametertree import Parameter, ParameterTree
+from pyqtgraph.parametertree import Parameter, ParameterTree, registerParameterType
 
 
 ## test subclassing parameters
 ## This parameter automatically generates two child parameters which are always reciprocals of each other
 class ComplexParameter(pTypes.GroupParameter):
     def __init__(self, **opts):
-        opts['type'] = 'bool'
+        opts['type'] = 'complexparam'
         opts['value'] = True
         pTypes.GroupParameter.__init__(self, **opts)
         
@@ -45,7 +45,7 @@ class ComplexParameter(pTypes.GroupParameter):
 ## this group includes a menu allowing the user to add new parameters into its child list
 class ScalableGroup(pTypes.GroupParameter):
     def __init__(self, **opts):
-        opts['type'] = 'group'
+        opts['type'] = 'scalablegroup'
         opts['addText'] = "Add"
         opts['addList'] = ['str', 'float', 'int']
         pTypes.GroupParameter.__init__(self, **opts)
@@ -59,10 +59,13 @@ class ScalableGroup(pTypes.GroupParameter):
         self.addChild(dict(name="ScalableParam %d" % (len(self.childs)+1), type=typ, value=val, removable=True, renamable=True))
 
 
+all_params_types = makeAllParamTypes()
 
+registerParameterType('complexparam', ComplexParameter)
+registerParameterType('scalablegroup', ScalableGroup)
 
 params = [
-    makeAllParamTypes(),
+    all_params_types,
     {'name': 'Save/Restore functionality', 'type': 'group', 'children': [
         {'name': 'Save State', 'type': 'action'},
         {'name': 'Restore State', 'type': 'action', 'children': [
--- a/pyqtgraph/examples/parametertree.py
+++ b/pyqtgraph/examples/parametertree.py
@@ -61,7 +61,7 @@ class ComplexParameter(pTypes.GroupParameter):
 ## this group includes a menu allowing the user to add new parameters into its child list
 class ScalableGroup(pTypes.GroupParameter):
     def __init__(self, **opts):
-        opts["type"] = "group"
+        opts["type"] = "scalablegroup"
         opts["addText"] = "Add"
         # opts['addList'] = ['str', 'float', 'int']
         addMenu = [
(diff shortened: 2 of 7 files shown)


seb5g:
could someone review this PR please?

j9ac9k:
Docs error:

```
/home/docs/checkouts/readthedocs.org/user_builds/pyqtgraph/checkouts/3431/doc/source/../../pyqtgraph/parametertree/Parameter.py:docstring of pyqtgraph.parametertree.Parameter.Parameter.check_type:3: ERROR: Unexpected indentation. [docutils]
```

j9ac9k:
Thanks for the PR @MariaF203 sorry it's taken me so long to look at this.  The diff looks good, and treating it as a radio button and introducing a comparison method along w/ appropriate testing is greatly appreciated!

Let me know if you would like me to fix the docs build error.

seb5g:
@j9ac9k as @MariaF203 was my student and moved on to her studies, I will fix the doc error and go on with this PR!

j9ac9k:
@MariaF203 sorry I didn't get to this PR during the relevant academic term. Hope that my slowness to respond isn't discouraging you from other open source contributions!

MariaF203:
@j9ac9k thanks for the review, I'm glad this PR was able to be merged!

PR #3434, opened 2025-12-02
HistogramLUTItem: add option colorMapMenu

pijyoi:
This is a continuation of #2779 and #2955.

#2779 added expanded colormaps to the menu selection in `HistogramLUTItem`.
#2955 added colormaps menu to `ColorBarItem`; and in addition allowed the colormaps menu to be customized.

Usually, the user does not need the functionality of `GradientEditorItem` that allows editing of the gradient. The user just wants to visualize the colormap.
Cons of using `GradientEditorItem`:
1) The ticks rendered by `GradientEditorItem` take up some screen estate.
2) Each tick is a `QGraphicsWidget`. This means that our 256-color colormaps instantiate the same number of ticks, even though they are hidden.

This PR adds the `ColorBarItem` functionality developed in #2955 to `HistogramLUTItem`:
-  None: use the existing `GradientEditorItem` (default)
- `False`: disable change of colormap
- `True`: uses all available colormaps in menu (with the exception of the legacy preset gradients)
- ColorMapMenu: user-provided custom menu.

### DIFF
--- a/pyqtgraph/examples/imageAnalysis.py
+++ b/pyqtgraph/examples/imageAnalysis.py
@@ -38,7 +38,13 @@ iso.setParentItem(img)
 iso.setZValue(5)
 
 # Contrast/color control
-hist = pg.HistogramLUTItem()
+
+# use custom colormap menu
+userList = ['viridis', 'cividis', 'magma', 'turbo']
+cmapMenu = pg.ColorMapMenu(userList=userList)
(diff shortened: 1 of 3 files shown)


j9ac9k:
Thanks for this PR @pijyoi

Sorry I didn't get to this earlier, this LGTM, merging!

PR #3439, opened 2025-12-07
PColorMeshItem: make default display grid non-transposed

### LINKED ISSUE #3437: PColorMeshItem with only Z specified renders as column-major
The simplest usage of `PColorMeshItem` in `pyqtgraph` to visualize a 2D array will render it as column-major. This differs from the `matplotlib` behavior. 

In order to render it as row-major, the user will need to provide full X, Y arrays.

This behavior stems from the `pyqtgraph` implementation using dimensions 0 and 1 for X and Y respectively. This additionally requires specifying `indexing="ij"` to `np.meshgrid` to override the default of `indexing="xy"`.

I am not sure if we should fix this behavior to be consistent with `matplotlib?

Image

Image

```python
import numpy as np

nrows = 3
ncols = 5
Z = np.arange(nrows * ncols).reshape((nrows, ncols))
X, Y = np.meshgrid(np.arange(ncols+1), np.arange(nrows+1))

fig, axes = plt.subplots(2)
axes[0].pcolormesh(Z, shading='flat')
axes[1].pcolormesh(X, Y, Z, shading='flat')

plt.show()
```

```python
import numpy as np
import pyqtgraph as pg

nrows = 3
ncols = 5
Z = np.arange(nrows * ncols).reshape((nrows, ncols))
X, Y = np.meshgrid(np.arange(ncols+1), np.arange(nrows+1))

pg.mkQApp()
axes = [win.addPlot(row=r, col=0) for r in range(2)]

cmap = pg.colormap.get('viridis')
axes[0].addItem(pg.PColorMeshItem(Z, colorMap=cmap))
axes[1].addItem(pg.PColorMeshItem(X, Y, Z, colorMap=cmap))

win.show()
pg.exec()
```

pijyoi:
when only a single argument (Z) is passed to `PColorMeshItem`, the more intuitive behavior is that the 2d-array is displayed non-transposed.

fixes #3437

### DIFF
--- a/pyqtgraph/graphicsItems/PColorMeshItem.py
+++ b/pyqtgraph/graphicsItems/PColorMeshItem.py
@@ -5,7 +5,6 @@ import numpy as np
 
 from .. import Qt, colormap
 from .. import functions as fn
-from .. import getConfigOption
 from ..Qt import compat
 from ..Qt import OpenGLConstants as GLC
 from ..Qt import OpenGLHelpers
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks @pijyoi

PR #3442, opened 2025-12-11
Fix missing tickvalues on inverted DateTimeAxis

nup002:
This PR fixes a bug where an inverted DateTimeAxis would not display any ticks because minVal and maxVal passed to `tickValues()` were not being sorted.


j9ac9k:
Nice catch!  Thanks for the PR @nup002  sorry it took me so long to get to this.

PR #3444, opened 2025-12-12
JSON serialization for Parameter

MariaF203:
### Summary
This PR introduces JSON serialization and deserialization for `Parameter` objects.
### Motivation
The result of `Parameter.saveState()` may contain objects such as numpy arrays and ColorMap that cannot be serialized in JSON. This makes the export of Parameters as JSON structures impossible.
This PR proposes a clean and reusable solution to encode and decode `Parameter` objects in JSON.
### Changes
1. Two new methods in `Parameter`:
- `Parameter.to_json(parameter)`
- `Parameter.from_json(json_str)`
2. A dedicated `JsonEncoderDecoder` class supporting serialization of :
- numpy arrays
- ColorMap
- tuples

Several `Parameter` attributes accept multiple data types. Handling JSON encoding for special cases in a dedicated class keeps the code clean and maintainable.

3. Test for JSON encoder and decoder.

### DIFF
--- a/pyqtgraph/parametertree/utils.py
+++ b/pyqtgraph/parametertree/utils.py
@@ -1,0 +1,137 @@
+from typing import Any
+from json import JSONEncoder, JSONDecoder
+import json
+import numpy as np
+
+from ..colormap import ColorMap
+from ..Qt.QtGui import QColor
+
+"""
+This class provides custom JSON encoder and decoder for objects that are not supported by the standard json library.
+
+Parameter objects may contains several data types of objects that cannot be directly serialized into JSON.
+To ensure full serialization and deserialization of Parameter objects, this class introduced custom encoding for:
+	- Numpy ndarray: Not serializable in JSON by default.
+	- ColorMap: Not serializable in JSON by default.
+	- tuple: the json module encodes tuples as lists, so a custom encoding is required to preserve and restore the tuple
+	 type.
+
+We choose to handle JSON encoding in a dedicated class rather than directly inside the Parameter classes. Indeed, many
+Parameter attributes can accept multiple data types, making direct handling tricky. A dedicated encoder ensures a clean
+and maintainable encoding for all Parameter classes.
+"""
+
+
+class JsonEncoderDecoder(JSONEncoder):
+    def default(self, o: Any) -> Any:
+        """
+        Serialize additional objects for JSON encoding.
+        Override the default() method from JSONEncoder to support:
+            - NumPy ndarray,
+            - ColorMap.
+        Returns a serializable object for o, or call the base implementation.
+
+        Parameters
+        ----------
+        o: Any
+            The object to serialize.
+
+        Returns
+        -------
+        Any
+            A serializable representation of the object.
+        """
+        if isinstance(o, np.ndarray):
+            return dict({'__ndarray__': o.tolist()})
+
+        elif isinstance(o, ColorMap):
+            attrs = dict()
+            attrs['pos'] = json.dumps(o.pos, cls=JsonEncoderDecoder)
+            attrs['color'] = o.color
+            attrs['mapping_mode'] = o.mapping_mode
+            attrs['name'] = o.name
+            attrs['stopsCache'] = o.stopsCache
+
+            return dict({'__colormap__': attrs})
+
+        return super().default(o)
+
+    def encode(self, o: Any) -> str:
+        """
+        Encode an object into a JSON string.
+        Extends encode() method from JSONEncoder to support tuples.
+
+        Parameters
+        ----------
+        o: Any
+            The object to encode.
+
+        Returns
+        -------
+        str
+            A JSON string representation of the object.
+        """
+
+        def hint_special(o: Any):
+            if isinstance(o, tuple):
+                return dict({'__tuple__': [hint_special(e) for e in o]})
+            elif isinstance(o, list):
+                return [hint_special(e) for e in o]
+            elif isinstance(o, dict):
+                return {k: hint_special(v) for k, v in o.items()}
+            else:
+                return o
+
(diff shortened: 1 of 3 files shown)


pyqtgraph/parametertree/utils.py
```diff
@@ -0,0 +1,137 @@
...
+        -------
+        Any
+            A serializable representation of the object.
+        """
+        if isinstance(o, np.ndarray):
+            return dict({'__ndarray__': o.tolist()})
+
+        elif isinstance(o, ColorMap):
+            attrs = dict()
+            attrs['pos'] = json.dumps(o.pos, cls=JsonEncoderDecoder)
```
pijyoi: Aren't `pos` and `color` both `ndarray`s? Why does `pos` need an explicit call to `json.dumps`?
Also `pos` and `color` could be obtained by calling the public API `ColorMap.getStops`.

pyqtgraph/parametertree/utils.py
```diff
@@ -0,0 +1,137 @@
...
+            attrs['color'] = o.color
+            attrs['mapping_mode'] = o.mapping_mode
+            attrs['name'] = o.name
+            attrs['stopsCache'] = o.stopsCache
```
pijyoi: I don't think cache values need to be serialized? In any case, this PR doesn't make use of it to reconstruct the ColorMap.

pyqtgraph/parametertree/utils.py
```diff
@@ -0,0 +1,137 @@
...
+        o: Any
+            The object to serialize.
+
+        Returns
+        -------
+        Any
+            A serializable representation of the object.
+        """
+        if isinstance(o, np.ndarray):
+            return dict({'__ndarray__': o.tolist()})
```
pijyoi: No need the `dict` wrapper?

pyqtgraph/parametertree/utils.py
```diff
@@ -0,0 +1,137 @@
...
+
+        elif isinstance(o, ColorMap):
+            attrs = dict()
+            attrs['pos'] = json.dumps(o.pos, cls=JsonEncoderDecoder)
+            attrs['color'] = o.color
+            attrs['mapping_mode'] = o.mapping_mode
+            attrs['name'] = o.name
+            attrs['stopsCache'] = o.stopsCache
+
+            return dict({'__colormap__': attrs})
```
pijyoi: No need the `dict` wrapper?

pyqtgraph/parametertree/utils.py
```diff
@@ -0,0 +1,137 @@
...
+
+        Returns
+        -------
+        str
+            A JSON string representation of the object.
+        """
+
+        def hint_special(o: Any):
+            if isinstance(o, tuple):
+                return dict({'__tuple__': [hint_special(e) for e in o]})
```
pijyoi: No need the `dict` wrapper?

pyqtgraph/parametertree/utils.py
```diff
@@ -0,0 +1,137 @@
...
+        np.ndarray or ColorMap or tuple or dict
+            The deserialized object.
+        """
+        if '__ndarray__' in dct:
+            return np.array(dct['__ndarray__'])
+
+        if '__colormap__' in dct:
+            elt = dct['__colormap__']
+            rgba_color = []
+            for c in elt['color']:
```
pijyoi: I think `elt['color']` can be passed directly to `ColorMap` without converting the values to `QColor`? The colors are stored internally as floats.

pyqtgraph/parametertree/utils.py
MariaF203: I removed the explicit call to `json.dumps`, it was a mistake.
I now use `ColorMap.getStops` to obtain `pos` and `color`.

pyqtgraph/parametertree/utils.py
MariaF203: Yes, I had forgotten to remove it.

pyqtgraph/parametertree/utils.py
MariaF203: I removed it.

pyqtgraph/parametertree/utils.py
MariaF203: I removed it.

pyqtgraph/parametertree/utils.py
MariaF203: I removed it.

pyqtgraph/parametertree/utils.py
MariaF203: Now that I use `getStops()` to obtain `color` I don't need to convert the values to `QColor` anymore.

MariaF203:
Thank you for your review. I fixed all the problems you pointed out.

seb5g:
Hello, just checking if this will be accepted soon?

Ashwolaa:
I suggest using the implementation of 
https://github.com/PyMoDAQ/pyqtgraph/tree/feature/json_serialization

- not relying on a class but using utility functions. 
- extends the serialization for QIcon (taking the path or the index) + frozenset

seb5g:
I'm actually not that happy with the current implementation as it relies on a class or methods listing the various special cases not handled by the base JSON stuff. I would rather propose something where the parameter itself expose a method/recipe on how to serialize itself. I'm not sure it can be done easily with the json library and will have to look into it.

seb5g:
@Ashwolaa do you have a PR made based on the content of the pymodaq/pyqtgraph fork? I guess we could in first place use it like this then maybe add some per/Parameter serialization handling.

Ashwolaa:
This was the PR 
https://github.com/PyMoDAQ/pyqtgraph/pull/1

I just added support for QIcon + frozenset in this branch
https://github.com/PyMoDAQ/pyqtgraph/tree/feature/json_serialization

Issue #3436, opened 2025-12-06
Clarify `setClipToView` data requirements on documentation


azelcer:
The implementation of `setClipToView` expects `x` data to be ordered. The behaviour is ill-defined otherwise. It would be good to make that clear on the documentation.

If it's OK, I'd like to make a PR clarifying this issue.

azelcer:
Hi @j9ac9k , I just made the PR.

I tried to add (in a local branch) some related functionality, and I got puzzled about the internal data representation. I see there are at least three properties where `PlotDataItem` keeps the data: `_dataset`, `_datasetMapped` and `_datasetDisplay`.

Is there a place where I can find information about how the data is managed? The "Internals" part of the documentation is quite short.

PR #3449, opened 2026-01-02
docs: Clarify `setClipToView` documentation

azelcer:
Clipping to view only works for sorted x-data, but it is not stated in the docs. This PR adds this information to the docs and adjusts wording in related places.

Related issue: #3436


j9ac9k:
Thanks for the PR @azelcer  sorry this took me so long to get to.  Appreciate the clearing up of the restrictions!

PR #3452, opened 2026-01-06
Add 3D grid planes with axes, ticks and labels

swvanbuuren:
This PR adds the class `GLGridAxis` representing 3D grid planes with axes, ticks and labels for use in a `GLViewWidget` scene. Depending on the azimuth and elevation view angles, only certain planes and axes are shown.

`GLGridAxis` is only meant to be added once to the scene in a centered position. In order to get the optimal camera position, `GLGridAxis` also offers a convenience function `best_camera` to provide a dict with the corresponding parameters to set this.

This PR also comes with an example to showcase the new 3D grid axis.
Please have a look and see if it works properly.

Here's a screenshot of the example:
image

I am looking forward to your feedback and reviews.

### DIFF
--- a/pyqtgraph/opengl/items/GLGridAxis.py
+++ b/pyqtgraph/opengl/items/GLGridAxis.py
@@ -204,9 +204,9 @@ class GLAxis(GLGraphicsItem):
         ====================  ==================================================
         """
         args = ('coords', 'coords_labels', 'limits', 'other_limits', 'axis',
-                'faces', 'tick_axis', 'label_side', 'font', 'label_color',
-                'line_color', 'line_antialias', 'line_width', 'azimuth_range',
-                'elevates')
+                'faces', 'tick_axis', 'label_side', 'tick_offset_factor',
+                'font', 'label_color', 'line_color', 'line_antialias',
+                'line_width', 'azimuth_range', 'elevates')
         for arg in args:
             if arg in kwargs:
                 setattr(self, arg, kwargs[arg])
--- a/pyqtgraph/opengl/items/GLGridAxis.py
+++ b/pyqtgraph/opengl/items/GLGridAxis.py
@@ -1,0 +1,461 @@
+import numpy as np
+
+from ... import getConfigOption
+from ...Qt import QtGui, QtCore
+from ...Vector import Vector
+from ..GLGraphicsItem import GLGraphicsItem
+from ...opengl import GLLinePlotItem, GLMeshItem, GLTextItem
+
+
+def check_visibility(azimuth_range, azimuth, elevation_range=None, elevation=None):
+    """Check if item should be visible based on camera angles.
+
+    Args:
+        azimuth_range: tuple (min, max) or None
+        elevation_range: tuple (min, max) or None
+        azimuth: current azimuth angle
+        elevation: current elevation angle
+
+    Returns:
+        bool: True if visible, False otherwise
+
+    """
+    if azimuth_range:
+        min_az, max_az = azimuth_range
+        az = azimuth + 360 if max_az > 360 and azimuth < min_az % 360 else azimuth
+        if not (min_az <= az < max_az):
+            return False
+
+    if elevation_range:
+        min_el, max_el = elevation_range
+        if not (min_el <= elevation <= max_el):
+            return False
+
+    return True
+
+
+def other_axes(axis):
+    """Return the two axes other than the given one."""
+    return [i for i in range(3) if i != axis]
+
+
+class GLGridPlane(GLGraphicsItem):
+    """Grid plane in 3D space."""
+
+    def __init__(self, parentItem=None, **kwargs):
+        super().__init__(parentItem=parentItem)
+        self.axis = 0
+        self.offset = 0.0
+        self.coords = (0, 1), (0, 1)
+        self.limits = (-0.05, 1.05), (-0.05, 1.05)
+        white_bg = getConfigOption('background') == 'w'
+        self.face_color = (0.95, 0.95, 0.95, 1) if white_bg else (0.05, 0.05, 0.05, 1)
+        self.line_color = (0.7, 0.7, 0.7, 1) if white_bg else (0.3, 0.3, 0.3, 1)
+        self.line_antialias = False
+        self.line_width = 1.0
+        self.azimuth_range: tuple | None = None
+        self.elevation_range: tuple | None = None
+
+        self._lineplot = GLLinePlotItem(parentItem=self, mode='lines')
+        self._lineplot.setDepthValue(self.depthValue() + 1)
+        self.setParentItem(parentItem)
+
+        self._mesh = GLMeshItem(parentItem=self)
+
+        self.setData(**kwargs)
+
+    def setData(self, **kwargs):
+        """Update the grid plane
+
+        ====================  ==================================================
+        **Arguments:**
+        ------------------------------------------------------------------------
+        axis                  int indicating the grid plan axis 0: x, 1: y, 2: z
+        offset                the offset along the axis orthogonal to the plane
+        coords                tuples with the coordinates for the first and
+                              second axis and the grid
+        limits                tuples with the limits for the first and
+                              second axis and the grid
+        face_color            RGBA tuple for the grid plane surface color
+        line_color            RGBA tuple for the grid lines color
+        line_antialias        boolean indicating if grid lines are antialiased
+        line_width            float indicating the grid lines width
+        azimuth_range         tuple (min, max) or list of tuples for visibility
+        elevation_range       tuple (min, max) or list of tuples for visibility
+        ====================  ==================================================
+        """
+        args = ('axis', 'offset', 'coords', 'limits', 'face_color',
+                'line_color', 'line_antialias', 'line_width', 'azimuth_range', 
+                'elevation_range')
+        for arg in args:
+            if arg in kwargs:
+                setattr(self, arg, kwargs[arg])
+        self._mesh.setMeshData(
+            **dict(self._backplane_face()), color=self.face_color
+        )
+        self._lineplot.setData(
+            pos=self._line_segments_positions(),
+            color=self.line_color,
+            antialias=self.line_antialias,
+            width=self.line_width,
+        )
+        self.update()
+
+    def is_visible(self, azimuth, elevation):
+        """Check if plane should be visible based on camera angles."""
+        return check_visibility(
+            self.azimuth_range, azimuth, self.elevation_range, elevation
+        )
+
+    def _grid_positions(self):
+        """Create grid positions in the specified plane."""
+        axes = other_axes(self.axis)
+
+        for idx, coord in enumerate(self.coords):
+            for c in coord:
+                p1, p2 = [0.0, 0.0, 0.0], [0.0, 0.0, 0.0]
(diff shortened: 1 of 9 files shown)


pijyoi:
Using one `GLTextItem` per label should be very inefficient. The OpenGL framebuffer has to be retrieved, then `QPainter` is used to render to it and finally the framebuffer has to be sent back.

I have always thought that `GLTextItem` would be much more useful if its API allowed the user to specify multiple labels.

https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/examples/GLPainterItem.py is an example where all the stuff that needs to be rendered by `QPainter` is done all together.

By the way, the `GLMixIn::renderToArray` function doesn't currently work for any `GLGraphicsItem` that makes use of `QPainter`. That includes `GLTextItem` and `GLGradientLegendItem`. One issue is that `QPainter` needs to be instantiated on the frame buffer object during the export, and not on the `GLViewWidget`. I think the other issue was that the `QPainter` drawing code assumes that the whole view is to be rendered to, whereas the `renderToArray` code by default splits it up into multiple smaller textures.

swvanbuuren:
During implementation I also noticed the issue that `GLTextItem` can only print one label per instance. I agree that that we should improve this, but decided not to do this in this PR, to keep it small and manageable.

The question now is how we proceed. I see two options:

1. We continue with this PR review as is and try to get it merged. In the meantime I will start working on `GLTextItem` support for multiple labels. When that's finished, we can modify `GLGridAxis` to use `GLTextItem`s with multiple labels.
2. We first implement multi label support for `GLTextItem` and merge it. Then we rebase this PR, to use `GLTextItem` with multiple labels and try to get it merged.

If you see any other possibility, please let me know.

I personally prefer the first option. Although the current implementation could be more efficient, I did not experience any performance issues (although that might of course differ per use case and platform/machine). Also, first waiting for another PR to get merged before we can merge this one can take a long time.

@pijyoi What do you think?

The other two issues with `QPainter` you mentioned are bit beyond me and this PR, but when I find the time I might look into them. No promises here though.

pijyoi:
Option 1 sounds great. But I also want to add that there's no obligation on your part to modify `GLTextItem`.

pyqtgraph/opengl/items/GLGridAxisItem.py
```diff
@@ -0,0 +1,461 @@
...
+        coords                tuple with the coordinates for the axis ticks
+        coords_labels         tuple with the axis tick labels
+        limits                tuple with the axis limits
+        other_limits          tuples with the limits for the other two axes
+        axis                  int indicating the axis 0: x, 1: y, 2
+        faces                 tuple with the faces for the other two axes
+        tick_axis             int indicating which axis the ticks point to
+        label_side            int indicating the side of the labels 0: low, 1:
+                              high
+        tick_offset_factor    float for tick label offset scaling factor
```
pijyoi, line 200: `tick_offset_factor` is not set into `self.tick_offset_factor`

pyqtgraph/opengl/items/GLGridAxis.py
```diff
@@ -0,0 +1,461 @@
...
+        limits                tuple with the axis limits
+        other_limits          tuples with the limits for the other two axes
+        axis                  int indicating the axis 0: x, 1: y, 2
+        faces                 tuple with the faces for the other two axes
+        tick_axis             int indicating which axis the ticks point to
+        label_side            int indicating the side of the labels 0: low, 1:
+                              high
+        tick_offset_factor    float for tick label offset scaling factor
+        font                  QFont for the labels
+        label_color           RGBA tuple for the labels and axis line color
```
pijyoi: I think `label_color` only affects the text color and not the line color?
It's inconsistent that `label_color` takes integer RGBA but `line_color` takes float RGBA.

pyqtgraph/opengl/items/GLGridAxisItem.py
swvanbuuren: Fixed with [b1882e8](https://github.com/pyqtgraph/pyqtgraph/pull/3452/commits/b1882e850ac515779b9ed03a1f350109d3435442)

pyqtgraph/opengl/items/GLGridAxis.py
swvanbuuren: Fixed with [1895e09](https://github.com/pyqtgraph/pyqtgraph/pull/3452/commits/1895e0949f42f560b99c4b574c3bbcbf3d5fa2f5)

swvanbuuren:
@pijyoi Thank you those remarks, I fixed them.

~~Unfortunately, I noticed another issue during testing: with black background/white foreground everything works as expected. When switching to white background/black foreground the grid/axis lines aren't plotted correctly anymore: they appear white, where they should be black or grey.~~

~~I believe we had similar issues in #3349, but the measure we took there don't work here.~~

~~Adding `glOptions='translucent'` helps a bit, but with that it still doesn't look good.~~

~~When anti-aliasing is enabled, the result is good / as expected.~~

~~Do you have clue what's the issue here? How can it be resolved?~~

Sorry, I wasn't paying attention and realized the approach from `GLSurfacPlotItem` also works here to solve the above issue. I will add the corresponding fix momentarily.

pyqtgraph/opengl/items/GLGridAxis.py
```diff
@@ -0,0 +1,478 @@
...
+        self.limits = (-0.05, 1.05), (-0.05, 1.05)
+        white_bg = getConfigOption('background') == 'w'
+        self.face_color = (0.95, 0.95, 0.95, 1) if white_bg else (0.05, 0.05, 0.05, 1)
+        self.line_color = (0.7, 0.7, 0.7, 1) if white_bg else (0.3, 0.3, 0.3, 1)
+        self.line_antialias = False
+        self.line_width = 1.0
+        self.azimuth_range: tuple | None = None
+        self.elevation_range: tuple | None = None
+
+        self._mesh = GLPolygonOffsetMeshItem(parentItem=self)
```
pijyoi: Since you are using `GLMeshItem` to draw just a quad with uniform color, you could add `computeNormals=False` to the constructor call.

swvanbuuren: Yes, that's a good idea. I have changed this now.

pijyoi:
For consistency with the other `GLGraphicsItem`s, this class should probably be named `GLGridAxisItem`?

swvanbuuren:
> For consistency with the other `GLGraphicsItem`s, this class should probably be named `GLGridAxisItem`?

Yes, that would be consistent. I have changed this.

pyqtgraph/opengl/items/GLGridAxisItem.py
```diff
@@ -0,0 +1,478 @@
...
+
+    return True
+
+
+def other_axes(axis):
+    """Return the two axes other than the given one."""
+    return [i for i in range(3) if i != axis]
+
+
+class GLPolygonOffsetMeshItem(GLMeshItem):
```
pijyoi: Now that there are 2 users of this polygon offset fill (`GLSurfacePlotItem` and `GLGridAxisItem`), do you think it should be made as an option of `GLMeshItem`?

It would also fix #3404.

swvanbuuren: This also occurred to me, but I wasn't sure if this would affect the behavior of application using `GLMeshItem` elsewhere. Or do think this can be enabled by default without causing any issues? I find it hard to give any definitive answer here.

Anyway, we still could implement it optionally with a flag to enable it:

```python
class GLMeshItem(GLGraphicsItem):

    ...

    def __init__(self, parentItem=None, **kwds):
        self.opts = {
            'meshdata': None,
            'color': (1., 1., 1., 1.),
            'drawEdges': False,
            'drawFaces': True,
            'edgeColor': (0.5, 0.5, 0.5, 1.0),
            'shader': None,
            'smooth': True,
            'computeNormals': True,
            'polygonOffset': False,
        }
    
    ...

    def paint(self):
        if self.opts['polygonOffset']:
            GL.glEnable(GL.GL_POLYGON_OFFSET_FILL)
            GL.glPolygonOffset(1.0, 1.0)

        # existing implementation of paint()

        if self.opts['polygonOffset']:
            GL.glDisable(GL.GL_POLYGON_OFFSET_FILL)
            GL.glPolygonOffset(0.0, 0.0)
```

BTW, just setting this polygon offset alone doesn't suffice. We also need to set the `glOptions` variable of the lines plotted with `GLLinePlotItem` to `'translucent'` (by default it's set to `'additive'`).

Given we decide to add this to `GLMeshItem`, should it be part of this PR or do we create a new one?

pijyoi: It should be fine to add it to this PR since it will be disabled by default.

pijyoi: It should be fine to add it to this PR since it will be disabled by default.

swvanbuuren: Ok, I just added this option to `GLMeshItem`. It's now used by `GLSurfacePlotItem` and `GLGridAxisItem`. Also, I updated the `GLMeshItem` example to use this option when edges are plotted on top of faces.

pyqtgraph/opengl/items/GLSurfacePlotItem.py
```diff
@@ -109,6 +109,9 @@ def setData(self, **kwds):
         if self._z is None:
             return
         
+        if self._showGrid:
```
pijyoi: Should this be
```python
self.setPolygonOffset(self._showGrid)
```
Otherwise once enabled, it can't be disabled.

swvanbuuren: You are totally right!

pijyoi:
I have added #3457, which allows multiple labels to be drawn by a single `GLTextItem`.

swvanbuuren:
> I have added #3457, which allows multiple labels to be drawn by a single `GLTextItem`.

@pijyoi I can rebase this PR's branch with the one from #3457 and use it to render all axis tick labels at once. Then of course, #3457 should be merged before this PR.

pijyoi:
#3457 has been merged, so you could do the rebase now

swvanbuuren:
> #3457 has been merged, so you could do the rebase now

Done!

j9ac9k:
Hi @swvanbuuren 

Sorry for taking so long to get to this! Due to the armed conflict in the middle east, my kids and I got evacuated.  The disruption to our lives was rather significant which is why I haven't been active here in these parts.  Thanks for working with @pijyoi (and thanks @pijyoi for reviewing and providing feedback on this PR) on this, it's wonderful to see the opengl capabilities improve!

LGTM, merging.

swvanbuuren:
> Hi @swvanbuuren
> 
> Sorry for taking so long to get to this! Due to the armed conflict in the middle east, my kids and I got evacuated. The disruption to our lives was rather significant which is why I haven't been active here in these parts. Thanks for working with @pijyoi (and thanks @pijyoi for reviewing and providing feedback on this PR) on this, it's wonderful to see the opengl capabilities improve!
> 
> LGTM, merging.

No worries! Private life goes first, always. Thanks for merging. Looking forward to the next release.

PR #3457, opened 2026-01-25
allow GLTextItem to render multiple text items

pijyoi:
This PR allows a single `GLTextItem` to render multiple text items. The interface follows the style of `ScatterPlotItem`, where the textitems are specified as a `list` of `dict`s.

This should allow rendering multiple text items to be more efficient.


j9ac9k:
Thanks for the PR @pijyoi sorry it took me so long to get to this, it's wonderful seeing the OpenGL implementation get better!

Issue #2302, opened 2022-05-13
Bug in ROI rotation in conjunction with maxBounds


chrisbpoint:
### Short description

When using the maxBounds parameter of ROI to limit the ROI to the image size, the ROI can be moved over the boundaries
in case it has been rotated. Moreover, it cannot read some regions of the image anymore. See the modified example of pyqtgraph.examples where the ROI is in its maximum upper right position:

![example](https://user-images.githubusercontent.com/48798172/168255603-4c1dec0c-7476-4236-a5c7-62e1e24fd1a6.png)

### Code to reproduce
Simply use the Image Analysis of pyqtgraph.examples and modify one line and add one line. See the two lines below "# Custom ROI for selecting an image region" in the screenshot:

![code](https://user-images.githubusercontent.com/48798172/168255904-3e32bc53-857d-490b-bdf1-19718d8a684e.png)

### Expected behavior
Rotated and bounded ROI should be bounded to the image.

### Real behavior
Rotated and bounded ROI can be moved outside image and some image regions cannot be reached at all. 

### Additional context
To fix this behaviour, the following change in the method "def stateRect(self, state)" of ROI.py can be made:
tr.rotate(-state['angle']) ->tr.rotate(state['angle'])

oceanc-mean:
Thank you for solving the problem that has been bothering me for a long time. I strongly recommend fixing this bug.

yyren7:
I faced the same problem why i trying to bound the range of a rotated roi these days. Hope someone can fix this bug...

z894345865:
Thank you for solving the problem
My question is, Why hasn't this bug been officially fixed for so long

z894345865:
After following this method of modification, the bug has been fixed on the rectangular ROI. However, there are still bugs on the elliptical ROI, as it cannot be close to the image boundary

z894345865:
for elliptical ROI, override the stateRect function can solve this bug
``` python
    def stateRect(self, state):  # fix bug
        """
        正椭圆方程：
        x^2/a^2+y^2/b^2=1
        旋转椭圆后，Ax^2+Bxy+Cy^2+D=0
        A = a^2 sin^2(theta) + b^2 cos^2(theta)
        B = 2(a^2-b^2)sin(theta)cos(theta)
        C = a^2 cos^2(theta) + b^2 sin^2(theta)
        D = -a^2b^2
        边缘点坐标偏移值
[...]

PR #3466, opened 2026-02-21
Fix ROI bounding rectangle calculation for rotated regions

z894345865:
- Correct angle usage in ROI.stateRect from -state['angle'] to state['angle']
- Add proper stateRect implementations for EllipseROI, PolyLineROI, LineSegmentROI, and TriangleROI
- Add helper function _calculate_bounding_rect for consistent bounding rectangle calculations
- Adjust CrosshairROI handle positions for better behavior
- Fix variable name typo: verticies -> vertices in TriangleROI

#2302


j9ac9k:
@campagnola @outofculture this PR may affect you all; mind taking a look?

PR #3468, opened 2026-02-23
Cross platform monospace for command input and REPL

kevinanewman:
The 'Typewriter' style hint doesn't always work, and "monospace" font may not fall back as expected, update hint to 'Monospace'.

Tested on win32 and darwin.


j9ac9k:
Thanks for the PR @kevinanewman sorry it took me so long to get to this!

PR #3469, opened 2026-02-23
syntax highlighter updates

kevinanewman:
fix issue with full line comments after previous highlightBlock() update that handled #'s inside strings

align ExampleApp syntax highlighting with syntax.py and vice versa (shouldn't ExampleApp import the highlighter from syntax.py?)

### DIFF
--- a/pyqtgraph/examples/ExampleApp.py
+++ b/pyqtgraph/examples/ExampleApp.py
@@ -194,19 +194,44 @@ class PythonHighlighter(QSyntaxHighlighter):
         return DARK_STYLES if app.property('darkMode') else LIGHT_STYLES
 
     def highlightBlock(self, text):
-        """Apply syntax highlighting to the given block of text.
-        """
-        # Do other syntax formatting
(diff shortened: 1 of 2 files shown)


j9ac9k:
Hi @kevinanewman 

Thanks for the PR, sorry this took me so long to test out, this PR came in right as my kids and I got evacuated from the middle east, so our lives were turned upside down for quite a while.  Just now I'm starting to get enough spare time to come back here... I really appreciate your patience.

I tested the PR out locally and it works great!

kevinanewman:
Oh, wow.  I'm glad you and your family are OK, that's all that matters!  Thank you!

j9ac9k:
> Oh, wow. I'm glad you and your family are OK, that's all that matters! Thank you!

We're good, new felt like we were in any danger (which I know is a lot more than many others tangled up in this conflict can say), we were just displaced and I needed to direct all my attention outside of work towards my kids for a while.  Hopefully I can clear out the PR backlog now haha

PR #3471, opened 2026-02-27
draw AxisItem without QPicture intermediary

pijyoi:
This is a pre-emptive fix for the pre-release Qt 6.11.
The issue occurs on Wayland, but not on X11 nor Windows.

reported at https://qt-project.atlassian.net/browse/QTBUG-144549

The following screenshot shows that `AxisItem` is not rendered correctly. The most visually obvious errors can be seen at Y=0.0 and Y=-1.0.
image

A pyqtgraph-free MRE:
```python
# misalignment occurs under the following conditions:
# - PySide6 6.11
# - Wayland
# - rendering through a QPicture

from PySide6 import QtCore, QtGui, QtWidgets

class SimpleWidget(QtWidgets.QGraphicsWidget):
    def paint(self, painter, option, widget):
        self.render(painter, QtCore.Qt.GlobalColor.darkRed)

        pic = QtGui.QPicture()
        with QtGui.QPainter(pic) as p:
            self.render(p, QtCore.Qt.GlobalColor.darkGreen)
        pic.play(painter)

    def render(self, painter, color):
        pen = QtGui.QPen(QtGui.QColor(color))
        painter.setPen(pen)
        painter.drawRect(self.rect())

scene = QtWidgets.QGraphicsScene()
view = QtWidgets.QGraphicsView(scene)
layout = QtWidgets.QGraphicsGridLayout()
container = QtWidgets.QGraphicsWidget()
container.setLayout(layout)

for row in range(3):
    for col in range(3):
        item = SimpleWidget()
        layout.addItem(item, row, col)

# issue occurs when we set the stretch factor 
layout.setRowStretchFactor(1, 10)
layout.setColumnStretchFactor(1, 10)

container.setGeometry(0, 0, 600, 400)
scene.addItem(container)
view.show()
app.exec()
```
image

The fix is to not use `QPicture` to draw the axes. Using `PYQTGRAPHPROFILE` shows that `AxisItem` doesn't take so much time to render to begin with. (1 to 2 millisecs). In any case, generation of the "drawSpecs" is still cached.

### DIFF
--- a/pyqtgraph/Qt/QtCore/__init__.py
+++ b/pyqtgraph/Qt/QtCore/__init__.py
@@ -1,0 +1,8 @@
+import importlib
+from .. import QT_LIB
+module = importlib.import_module(f'{QT_LIB}.QtCore')
+
+def __getattr__(name):
+    x = getattr(module, name)
+    globals()[name] = x
+    return x
--- a/pyqtgraph/Qt/QtGui/__init__.py
+++ b/pyqtgraph/Qt/QtGui/__init__.py
@@ -1,0 +1,8 @@
+import importlib
+from .. import QT_LIB
+module = importlib.import_module(f'{QT_LIB}.QtGui')
+
+def __getattr__(name):
+    x = getattr(module, name)
+    globals()[name] = x
+    return x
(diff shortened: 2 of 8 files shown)


pijyoi:
Looks like this bug is also present in 6.10 development branch.
So it is not present in 6.10.2, but will be present in 6.10.3.

pijyoi:
This has been fixed upstream in Qt 6.11.0 and 6.10.3

PR #3477, opened 2026-03-20
BugFix: Add check for None data in Matplotlib exporter

ullmannJan:
Skip processing for curves with None data. Otherwise no other lines will be displayed.


ixjlyons (approved):

j9ac9k:
Thanks for the PR @ullmannJan 

I'm just starting to pick up on the PR backlog, my life got turned upside down due to the conflict in the Middle East, so I can't say enough how much I appreciate your patience.

PR #3484, opened 2026-04-24
Fix reset button state after setToDefault

### LINKED ISSUE #3483: Parameter.setToDefault() function does not disable the reset button
The following code illustrates the issue. If one clicks on the `setToDefault` button at the bottom, the value of the parameter changes, but the reset button stays enabled.

```
import sys
from PySide6.QtWidgets import QApplication, QMainWindow, QPushButton, QVBoxLayout, QWidget


    def __init__(self):
        super().__init__()

        self.param = Parameter.create(name='param', type='int', value=1, default=2)
        self.paramTree = ParameterTree(self)
        self.paramTree.addParameters(self.param)

        self.button = QPushButton('setToDefault')

        self.layout = QVBoxLayout()
        self.layout.addWidget(self.paramTree)
        self.layout.addWidget(self.button)

        self.widget = QWidget(self)
        self.widget.setLayout(self.layout)
        self.setCentralWidget(self.widget)

        self.button.clicked.connect(self.buttonClick)

    def buttonClick(self):
        self.param.setToDefault()


app = QApplication(sys.argv)
window = MainWindow()
window.show()
app.exec()
```

Image
<br/>

The `Parameter.setToDefault()` updates the value and then resets the `self._modifiedSinceReset` flag.

```
def setToDefault(self):
    """Set this parameter's value to the default. Raises ValueError if no default is set."""
    with self.treeChangeBlocker():
        self.setValue(self.defaultValue())
        self._modifiedSinceReset = False
```

Firstly, the reset button state is updated by `WidgetParameterItem.updateDefaultBtn()`, called from `WidgetParameterItem.valueChanged()`. So it is processed _before_ the self._modifiedSinceReset flag is set.

Secondly, even if the flag were set upfront, the `WidgetParameterItem.valueChanged()` function contains these lines:

```
self.widget.setValue(val)
self.param.setValue(self.widget.value())
```

Basically, it writes the parameter value back, causing the `self._modifiedSinceReset` flag to be set to `True`.

The second point also complicates the picture. Imagine that, due to coercion, the widget changes the value, and the new value is assigned back to the parameter. Would it mean that the parameter is modified after reset or not? (I would rather say _it is_ modified. And if so, an unconditional `self._modifiedSinceReset = False` in `setToDefault()` would be incorrect.)

cookesan:
## Summary
- Recompute the reset-to-default modified flag from the final stored value after `setToDefault()`.
- Refresh attached parameter tree items so programmatic resets update the reset button state.
- Add coverage for the reset button state and a coerced-default edge case.

Fixes #3483.

## Testing
- `QT_QPA_PLATFORM=offscreen .venv/bin/python -m pytest --no-qt-log tests/parametertree`

### DIFF
--- a/pyqtgraph/parametertree/Parameter.py
+++ b/pyqtgraph/parametertree/Parameter.py
@@ -474,7 +474,11 @@ class Parameter(QtCore.QObject):
         """Set this parameter's value to the default. Raises ValueError if no default is set."""
         with self.treeChangeBlocker():
             self.setValue(self.defaultValue())
-            self._modifiedSinceReset = False
+            self._modifiedSinceReset = not self.valueIsDefault()
+            for item in list(self.items):
+                updateDefaultBtn = getattr(item, 'updateDefaultBtn', None)
+                if updateDefaultBtn is not None:
+                    updateDefaultBtn()
 
     def hasDefault(self):
         """Returns True if this parameter has a default value."""
--- a/pyqtgraph/parametertree/parameterTypes/basetypes.py
+++ b/pyqtgraph/parametertree/parameterTypes/basetypes.py
@@ -218,7 +218,6 @@ class WidgetParameterItem(ParameterItem):
 
     def defaultClicked(self):
         self.param.setToDefault()
-        self.updateDefaultBtn()
 
     def optsChanged(self, param, opts):
         """Called when any options are changed that are not
--- a/tests/parametertree/test_Parameter.py
+++ b/tests/parametertree/test_Parameter.py
@@ -89,6 +89,12 @@ def test_parameter_defaults_and_pristineness():
     p.setDefault(7, updatePristineValues=True)
     assert p.value() == 7
 
+    # if the value is coerced away from the raw default, the parameter remains modified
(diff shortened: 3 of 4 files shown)


q64:
I believe the fix also renders `updateDefaultBtn()` in `WidgetParameterItem.defaultClicked()` redundant, because `setToDefault()` already updates the default button.

For reference:
```
def defaultClicked(self):
    self.param.setToDefault()
    self.updateDefaultBtn()
```

cookesan:
Good catch, thanks. Removed the redundant `updateDefaultBtn()` call from `defaultClicked()`. The focused parameter-tree tests still pass.

cookesan:
Following up here after the requested cleanup: the redundant `updateDefaultBtn()` call was removed from `defaultClicked()`, and the focused parameter-tree tests were still passing after that change.

Is there anything else you would like adjusted before this moves forward?

Ashwolaa:
Thanks for the work @cookesan 
This looks good for me. I will have to slightly tweak #3479 once this is merged.

seb5g:
I'm now merging it!

PR #3489, opened 2026-05-13
Fixes #3487 :  Restore backward compatibility for blockSignal=callable in Parameter.setValue()

### LINKED ISSUE #1234: Plots with custom x-axis and default y-axis have the y-axis missing
### Short description
This is such a hilariously random but game-breaking bug; here goes: When a plot has a custom x-axis, but **not** a custom y-axis, the y-axis doesn't appear.
This issue has been introduces somewhere in-between 5cd19ba7a0eccf5158b16257470279d1acf4e6a1 and 6bd029915207e96ec4bd37cf6f52be31c3f10ee0 (inclusive).

### Code to reproduce

```python
import pyqtgraph as pg

app = QtGui.QApplication([])

w = pg.PlotWidget(axisItems={"bottom": pg.DateAxisItem()})

w.show()
app.exec_()
```

### Expected behavior
The y axis should be there

### Real behavior
The y axis is not there

### Tested environment(s)

 * PyQtGraph version: 6bd029915207e96ec4bd37cf6f52be31c3f10ee0
 * Qt Python binding: PyQt5 5.14.2 Qt 5.14.2
 * Python version: 3.8.3
 * NumPy version: 1.17.4
 * Operating system: Arch Linux
 * Installation method: pip 

### Additional context
Screenshot_20200607_214925

### LINKED ISSUE #3487: New semantics of blockSignal in Parameter.setValue() suppresses ParameterItem updates
The PR #3305 changes `Parameter.setValue()` such that setting `blockSignal` prevents `sigValueChanged` from being emitted.

That silently breaks the old `blockSignal=callable` pattern, used both in user code and in pyqtgraph's own `examples/parametertree.py`.

Previously, `blockSignal` was used to specify a single slot to ignore during the value update. When a parameter was supposed to reflect a model value, `blockSignal` could help prevent recursion while still allowing corresponding `ParameterItem`s to display updated values.

The following example demonstrates such use (see `MainWindow.lightSwitched()`):

```
import sys
from PySide6.QtCore import QObject, Signal
from PySide6.QtWidgets import QApplication,QPushButton, QVBoxLayout, QWidget


class Light(QObject):
    sigSwitched = Signal(bool)

    def __init__(self):
        super().__init__()
        self.on = False

    def switch(self, on):
        self.on = on
        self.sigSwitched.emit(self.on)


class MainWindow(QWidget):
    def __init__(self):
        super().__init__()

        self.param = Parameter.create(name='param', type='bool', value=False)
        paramTree = ParameterTree(self)
        paramTree.addParameters(self.param)
        toggleButton = QPushButton('toggle light')

        layout = QVBoxLayout()
        layout.addWidget(paramTree)
        layout.addWidget(toggleButton)
        self.setLayout(layout)

        toggleButton.clicked.connect(self.toggleButtonClick)
        self.param.sigValueChanged.connect(self.paramValueChanged)

        self.light = Light()
        self.light.sigSwitched.connect(self.lightSwitched)

    def paramValueChanged(self, param, value):
        self.light.switch(value)

    def lightSwitched(self, on):
		#
        # here we want to update the parameter value, but avoid re-switching the light
        #
        self.param.setValue(on, blockSignal=self.paramValueChanged)

    def toggleButtonClick(self):
        self.light.switch(not self.light.on)


app = QApplication(sys.argv)
window = MainWindow()
window.show()
sys.exit(app.exec())

```

Image

<br>

Now, when `toggleButton` is clicked, `param.setValue()` does not emit `sigValueChanged` at all. So while the parameter changes its value, the `ParameterItem` is not updated.

The same technique can be seen in `examples/parametertree.py`:

```
    def aChanged(self):
        self.b.setValue(1.0 / self.a.value(), blockSignal=self.bChanged)

    def bChanged(self):
        self.a.setValue(1.0 / self.b.value(), blockSignal=self.aChanged)
```

A backward-compatible solution could be to dispatch on the type of `blockSignal`, supporting both behaviors:

```
    def setValue(self, value, blockSignal=None):
        """
        Set the value of this Parameter; return the actual value that was set.
        (this may be different from the value that was requested).
        blockSignal specifies what to block during emission: a slot (callable) or the entire signal (True).
        """
        value = self._interpretValue(value)
        if fn.eq(self.opts.get('value', None), value):
            return value
        self._modifiedSinceReset = True
        self.opts['value'] = value
        if callable(blockSignal):
            self.sigValueChanged.disconnect(blockSignal)
            try:
                self.sigValueChanged.emit(self, value)
            finally:
                self.sigValueChanged.connect(blockSignal)
        elif not blockSignal:  # falsy: emit normally
            self.sigValueChanged.emit(self, value)

        return self.opts['value']
```

<Screenshot here>

That extends the existing `setValue()` by adding the `if callable(....)` branch. Or, as already suggested by @cogsy23, different `kwargs` can be used for different behaviors. Assuming the new behavior is already in use, this would be a bigger breaking change, perhaps, but makes the API and implementation code clearer.

Would it make sense to create a PR?

P.S. Regarding naming.
`blockSignal` is confusing under the old semantics, when _a single slot_ was blocked, not the signal.
`blockSignals` (proposed in #3305) is confusing under the new semantics, suggesting that all signals are blocked, whereas only `sigValueChanged` is.

With a single kwarg, `blockSignal` can be kept and reinterpreted as the blocking mode - callable for a specific slot, boolean for the whole signal. The name is not perfect, but the API stays backward-compatible, and the docstring can clarify its dual nature.

Rathod-shubhamm:
Hi! This is my first contribution. 
Could a maintainer please approve the CI workflow so the tests can run? 
This PR fixes the regression introduced in #3487.
## Description
This PR addresses an unintended breaking change introduced in PR #3305. The previous change modified `Parameter.setValue()` such that setting `blockSignal` to a truthy value prevented `sigValueChanged` from being emitted entirely. However, this silently broke the old `blockSignal=callable` pattern used in both user code and `pyqtgraph`'s own examples (e.g., `examples/parametertree.py`).

Previously, `blockSignal` could be a callable, specifying a single slot to ignore during the value update. This is crucial for preventing recursion when a parameter reflects a model value, while still allowing the corresponding `ParameterItem`s to display the updated values in the UI.

## Changes Made
- Updated `Parameter.setValue()` to dispatch on the type of `blockSignal` to support both behaviors.
- **If `blockSignal` is a callable:** It temporarily disconnects only that specific slot, emits `sigValueChanged`, and then reconnects it. (Restores the pre-#3305 behavior)
- **If `blockSignal` is not callable (e.g., `True`):** It continues to block the entire signal from being emitted. (Retains the new behavior introduced in #3305)
- Updated the docstring to clarify the dual nature of the `blockSignal` argument.

## Motivation and Context
Resolves the issue where using `param.setValue(value, blockSignal=self.someSlot)` failed to update the parameter item in the UI because the signal was completely suppressed instead of just blocking the single recursive slot. This provides a clean, backward-compatible solution without needing to introduce new kwargs or cause further breaking changes.

## Related Issues
*(Note: If there is an open issue number for this, add it here like `Fixes #1234`)*

### DIFF
--- a/pyqtgraph/parametertree/Parameter.py
+++ b/pyqtgraph/parametertree/Parameter.py
@@ -309,15 +309,22 @@ class Parameter(QtCore.QObject):
     def setValue(self, value, blockSignal=None):
         """
         Set the value of this Parameter; return the actual value that was set.
-        (this may be different from the value that was requested)
+        (this may be different from the value that was requested).
+        blockSignal specifies what to block during emission: a slot (callable) or the entire signal (True).
         """
         value = self._interpretValue(value)
         if fn.eq(self.opts.get('value', None), value):
             return value
         self._modifiedSinceReset = True
         self.opts['value'] = value
-        if not blockSignal:
-            self.sigValueChanged.emit(self, value)  # value might change after signal is received by tree item
+        if callable(blockSignal):
+            self.sigValueChanged.disconnect(blockSignal)
+            try:
+                self.sigValueChanged.emit(self, value)
+            finally:
+                self.sigValueChanged.connect(blockSignal)
+        elif not blockSignal:  # falsy: emit normally
+            self.sigValueChanged.emit(self, value)
 
         return self.opts['value']
 
--- a/tests/parametertree/test_Parameter.py
+++ b/tests/parametertree/test_Parameter.py
@@ -564,3 +564,90 @@ def test_interact_existing_parent():
     assert outParam in parent.names.values()
     outParam.activate()
     assert lastValue == 5
+
+
+# ---------------------------------------------------------------------------
+# Tests for Parameter.setValue() blockSignal behaviour (regression for #3305)
+# ---------------------------------------------------------------------------
+
+def test_setValue_blockSignal_callable_blocks_only_that_slot():
+    """
+    blockSignal=<callable> must temporarily disconnect *only* that slot while
+    still emitting sigValueChanged to every other connected listener.
+    This exercises the backward-compatible `if callable(blockSignal)` branch.
+    """
+    p = Parameter.create(name="param", type="float", value=0.0)
+
+    blocked_received = []
+    other_received = []
+
+    def blocked_slot(param, value):
+        blocked_received.append(value)
+
+    def other_slot(param, value):
+        other_received.append(value)
+
+    p.sigValueChanged.connect(blocked_slot)
+    p.sigValueChanged.connect(other_slot)
+
+    # blockSignal=blocked_slot → blocked_slot should NOT fire, other_slot SHOULD
+    p.setValue(1.0, blockSignal=blocked_slot)
+
+    assert blocked_received == [], (
+        "Blocked slot must not receive the signal when passed as blockSignal"
+    )
+    assert other_received == [1.0], (
+        "Other connected slots must still receive sigValueChanged"
+    )
+
+    # After the call the slot must be reconnected; next plain setValue fires both
+    p.setValue(2.0)
+    assert blocked_received == [2.0], (
+        "Blocked slot must be reconnected after the setValue call"
+    )
+    assert other_received == [1.0, 2.0]
+
+
+def test_setValue_blockSignal_true_suppresses_signal_entirely():
+    """
+    blockSignal=True must prevent sigValueChanged from being emitted at all.
+    This preserves the behaviour introduced in #3305.
+    """
+    p = Parameter.create(name="param", type="float", value=0.0)
+
+    received = []
+
+    def slot(param, value):
+        received.append(value)
+
+    p.sigValueChanged.connect(slot)
+
+    p.setValue(99.0, blockSignal=True)
+
+    assert received == [], (
+        "sigValueChanged must not be emitted when blockSignal=True"
+    )
+    # Value must still be stored despite the signal being blocked
+    assert p.value() == 99.0
+
+
+def test_setValue_blockSignal_none_emits_normally():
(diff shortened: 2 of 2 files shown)


Ashwolaa:
Hey @Rathod-shubhamm,
Thanks for the suggestion. I believe it could be interesting to keep blockSignal as a strict boolean (block the whole signal) and add a new blockSlot kwarg that accepts a callable or a list/tuple of callables to temporarily disconnect just those slots — that also allows an extension of the initial PR scope to be able to block multiple slots.
 The main tradeoff is backward compatibility: the old blockSignal=callable pattern (used in examples/parametertree.py) would need a deprecation shim — detect callable(blockSignal), route it to blockSlot internally, and emit a DeprecationWarning.

seb5g:
> The main tradeoff is backward compatibility: the old blockSignal=callable pattern (used in examples/parametertree.py) would need a deprecation shim — detect callable(blockSignal), route it to blockSlot internally, and emit a DeprecationWarning.

I like this solution to create a dedicated blockSlot, propably blockSlots (one or several slots) and route to it the current implementation using the callable if case while raising a deprecation warning. 

The examples would need modification to show this behaviour.

Ashwolaa:
Hey @Rathod-shubhamm,
What is the status on your side for some of the requested/proposed changes?

If you do not have enough spare time for it, we will follow up on your PR on our side.

seb5g:
closing this PR as the original author does not seem to answer anymore and @Ashwolaa took over in #3617

Issue #1767, opened 2021-05-07
Plot legend font looses size by PNG export at high width parameter


ikost:
### Short description
When I export an scene or a plot item from the API or from the context menu to the PNG file, the font of the legend is not as set and shown on the screen. If the width of the export image is big for higher DPI (width > 2000) the axis fonts and plot title font are as shown/as set but the legend font becomes very small as if the font scaling setting is not applied. With lower width (say 600) this behavior is not pronounced.

### Expected behavior
Scale legend font in the exported PNG as on the screen

### Tested environment(s)

 * PyQtGraph version: 12
 * Qt Python binding: PySide2
 * Python version: 5.9.2
 * NumPy version: 
 * Operating system: Windows 10
 * Installation method: pip

### exported image with width = 3000 - look on the legend
![image](https://user-images.githubusercontent.com/20628668/117428087-04dad380-af26-11eb-9572-bc605e916856.png)

### how plot looks in the app on the 4k screen
![image](https://user-images.githubusercontent.com/20628668/117428273-36539f00-af26-11eb-8a4a-1b91f3a5b15b.png)

pijyoi:
I find that if I disable the ```ItemIgnoresTransformations``` flag, then the legend text will be correctly sized.
This flag is enabled in https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/graphicsItems/LegendItem.py#L68.
Maybe someone knows why it was needed to be enabled?

```python
import numpy as np
import pyqtgraph as pg
import pyqtgraph.exporters

plt = pg.plot()
leg = plt.addLegend()
leg.setFlag(leg.ItemIgnoresTransformations, False)   # the relevant line
plt.plot(3*np.sinc(np.linspace(-3, 1, 1001)), pen='r', name='Hello')
plt.plot(np.sinc(np.linspace(-4, 1, 1001)), pen='g', name='Cruel')
plt.plot(np.sinc(np.linspace(-5, 1, 1001)), pen='b', name='World')

exporter = pg.exporters.ImageExporter(plt.plotItem)
exporter.parameters()['width'] = 2560
exporter.export('export.png')
```

j9ac9k:
Duplicate of #1552 (but this one shows a potential fix so I'll leave both open for now)

Doing a git history of that line takes us back to effectively the initial commit of LegendItem.  

Reading the docs on `ItemIgnoresTransformations` I can see why one would think to enable it:

[...]

PR #3492, opened 2026-05-24
Fix legend not scaling on large exports

### LINKED ISSUE #1552: Legend doesn't scale on large exports
```python
import pyqtgraph as pg
import numpy as np
app = pg.mkQApp()
pw = pg.plot()

ii = pg.ImageItem()
ii.setBorder(pg.mkPen('r', width=5))
ii.setImage(np.random.random((6000,6000,3)))
pw.addItem(ii)

spi = pg.ScatterPlotItem()
spi.setData([0, 2000, 5000], [0, 2000, 5000], symbol='s', width=5, pxMode=True, size=10)
pw.addItem(spi)

leg = pg.LegendItem()
leg.setParentItem(pw.plotItem)
leg.addItem(spi, name='test')
leg.setPen('k')
leg.setBrush('k')

# Open "Export" menu, Image Export, set resolution to 6000x4500
# Legend is miniscule!
```
### Expected behavior
Legend should be visible
### Real behavior

I consider myself young, but even I have a tough time reading the text...
image

### Tested environment(s)

 * PyQtGraph version: today afternoon master branch
 * Qt Python binding: 5.14.2
 * Python version: 3.8.2

### Additional context

I work a lot with image items, so a hack I use when I am guaranteed the existence of something that can scale for me:
```python
def _paintLegend(self, painter: QtGui.QPainter, floatLegend=False):
    oldPos = self.legend.pos()
    try:
      exportScene = GraphicsScene(parent=None)
      # Let the legend grow if needed, but it can't shrink without obliterating the text
      oldScale = np.clip(0.5/np.array(self.imgItem.pixelSize()), 1, np.inf)
      # Put into a different scene so alpha doesn't render what's behind the legend in the original scene
      # Just calling legend.paint(painter) doesn't paint child items, that's why I use a different scene
      exportScene.addItem(self.legend)
      # Legend does not scale itself, handle this
      self.legend.scale(*oldScale)
      painter.save()
      if floatLegend:
        # Saves the position of the legend as it appears in the original scene,
        # compared to docking in the top-left corner
        viewPos = self.viewbox.viewRect()
        viewX = max(viewPos.x(), 0)
        viewY = max(viewPos.y(), 0)
        imPos = self.legend.mapRectToItem(self.imgItem, self.legend.rect())
        painter.translate(int(imPos.x() - viewX), int(imPos.y() - viewY))
      exportScene.render(painter)
      painter.restore()
      self.legend.scale(*(1/oldScale))
    finally:
      # Put the legend back in the original scene
      self.legend.setParentItem(self.plotItem)
      self.legend.autoAnchor(oldPos, relative=False)
```

Is there a better way of achieving the same result?

mayank0030:
Removes the ItemIgnoresTransformations flag from LegendItem, which prevented the legend from scaling during high-resolution exports. The legend font and items now scale correctly with the rest of the plot.

Fixes #1552

### DIFF
--- a/pyqtgraph/graphicsItems/LegendItem.py
+++ b/pyqtgraph/graphicsItems/LegendItem.py
@@ -67,7 +67,6 @@ class LegendItem(GraphicsWidgetAnchor, GraphicsWidget):
         """
         GraphicsWidget.__init__(self)
         GraphicsWidgetAnchor.__init__(self)
-        self.setFlag(self.GraphicsItemFlag.ItemIgnoresTransformations)
         self.layout = QtWidgets.QGraphicsGridLayout()
         self.layout.setVerticalSpacing(verSpacing)
         self.layout.setHorizontalSpacing(horSpacing)


j9ac9k:
I seem to remember LegendItem needing that for some reason. I would need to research why this was added to begin with.

Do you have a reproducible example of a large export that demonstrates the issue?

mayank0030:
Thanks for looking into it! I found that multiple users on issues #1552 and #1767 confirmed that removing the ItemIgnoresTransformations flag fixes the legend scaling issue during high-resolution exports. The flag prevents the legend from scaling with the rest of the plot, causing the font and items to remain small when exporting at higher resolutions.

I checked the git history and that line was present from the initial commit of LegendItem. Looking at the Qt docs, ItemIgnoresTransformations is typically used for items like text labels that should remain a fixed size on screen — but in this case, the legend is a child of the plot and should scale along with it during exports.

If there are any edge cases where removing this flag causes issues, I'm happy to help test and investigate further.

ixjlyons:
The only thing I can think of is maybe the intention was to keep the legend font/item size the same at 100% zoom, but in that case I think you'd want the axis labels and tick text to work the same way, and the actual plotted symbols, for that matter.

Also I found exporting to SVG works so a large/small export look the same if you scale them to the same size on screen.

So that said, this seems ok to me.

Need to merge/rebase onto master to get #3458 and fix CI failures.

j9ac9k:
Welp, since we can't think of a reason of why it is the way it is, I'm good with merging.  Thanks for the PR @mayank0030 and thanks for the review @ixjlyons

PR #3493, opened 2026-05-24
Avoid unnecessary import of ExampleApp

pijyoi:
This PR arises from an observation that the following 2 invocations are not strictly equivalent:
- `python -m pyqtgraph.examples.VideoSpeedTest`
- `python /path/to/examples/VideoSpeedTest.py`

This difference was traced to `__init__.py` importing `ExampleApp`. When `ExampleApp` gets imported, it:
1) import the various Qt modules
2) instantiates an `QApplication`
   * This instantiation of an `QApplication` turns out to be unnecessary and has been removed in this PR
3) adds the examples directory to the python path
   * This side effect turns out to be relied upon by the `parametertree` example and `test_parametertypes`; it has thus been added to `__init__.py` in this PR

This PR is based on top of #3458 because otherwise the CI would fail.

### DIFF
--- a/pyqtgraph/examples/__init__.py
+++ b/pyqtgraph/examples/__init__.py
@@ -1,1 +1,8 @@
-from .ExampleApp import main as run
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath(os.path.dirname(__file__)))
+
+def run():
+    from . import ExampleApp
+    ExampleApp.main()
--- a/pyqtgraph/Qt/QtCore/__init__.py
+++ b/pyqtgraph/Qt/QtCore/__init__.py
@@ -1,0 +1,8 @@
+import importlib
+from .. import QT_LIB
+module = importlib.import_module(f'{QT_LIB}.QtCore')
+
+def __getattr__(name):
+    x = getattr(module, name)
+    globals()[name] = x
+    return x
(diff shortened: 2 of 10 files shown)


pyqtgraph/examples/__init__.py
```diff
@@ -1 +1,8 @@
-from .ExampleApp import main as run
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath(os.path.dirname(__file__)))
```
ixjlyons, line 4: Since the examples are now part of the package, it seems like it'd be preferable to change the imports of e.g. `pyqtgraph.examples.utils` to avoid this.

pijyoi: It would indeed be much cleaner not to modify the module path. I did try it while making this PR and some other parts of the codebase just failed.

The things that I had trouble with were the examples that wanted to import a file within the same directory

ixjlyons: Sorry, didn't read the PR description carefully enough. It seems the `PYTHONPATH` manipulation has essentially moved -- the `sys.path.insert` in `ExampleApp.py` can be removed now right?

I will say still adding the parent of a `__init__.py` file to `PYTHONPATH` _in_ the `__init__.py` file itself is strange, but in a way it's actually more explicit than a side effect of importing `ExampleApp`.

I guess I'd still like to see absolute imports everywhere in the examples eventually, but since this wasn't actually introduced in this PR, I'm good with punting to an issue.

Also, I noticed CI runs the examples through pytest, which inserts the test module path in `PYTHONPATH` (by default), so we don't really have automated coverage of running any given example directly from outside the examples directory.

j9ac9k:
Thanks for the PR @pijyoi and thanks for the review @ixjlyons

PR #2595, opened 2023-01-27
PlotCurveItem error with stepMode="center", autoRange and autoVisible

djdt:
An IndexError is raised for PlotCurveItems with stepMode="center" when dataBounds  is called and orthoRange is not None. The mask created in dataBounds is 1 element too long (len(x) == len(y) + 1) and must be trimmed.

Fixes #2594

PR #3496, opened 2026-06-05
PlotCurveItem error with x axis and stepMode="center", autoRange and autoVisible

djdt:
Similar to #2595, an IndexError is raised for PlotCurveItems with stepMode="center" when dataBounds is called and orthoRange is not None. The previous pull request only fixed this for the y axis, and not x. In this pull request a False is appended to the mask array.


j9ac9k:
Thanks for the PR @djdt 

My apologies this took so long to get merged.  LGTM.

PR #3504, opened 2026-06-20
use native pyside6 loadui

pijyoi:
Since PySide6 has its own native `loadUiType` function, we can remove the old implementation.


j9ac9k:
Love getting rid of our custom implementations when there is now builtin equivalents!  Thanks for the PR @pijyoi

PR #3505, opened 2026-06-21
ExampleApp: fix undefined app

pijyoi:
#3493 introduced a regression to `ExampleApp` when it removed the globally instantiated `app` because a couple of places assumed that `app` had been globally defined.

To trigger the regression:
- launch ExampleApp
- In the combo-box, change "Title Search" to "Content Search"
- Type something to search
-


j9ac9k:
Nice catch! Thanks for the PR!

PR #3507, opened 2026-06-21
Fix FFT autorange with auto downsampling

### LINKED ISSUE #438: fftMode fails to plot when autoDownsample==True and no x coords provided
code below exhibits the problem

```python
import numpy as np
import pyqtgraph as pg

Fs = 1000.0
ttt = np.arange(1000.0)/Fs
y = np.zeros_like(ttt)
for f in (100,200,300,400):
    y += np.sin(2*np.pi*f*ttt)

pw = pg.plot()
plotitem = pw.getPlotItem()
if 1:
    # fails to plot
    plotitem.setDownsampling(auto=True)
    curve = plotitem.plot(y)
elif 1:
    # plots fine
    plotitem.setDownsampling(auto=False)
    curve = plotitem.plot(y)
elif 1:
    # plots fine
    plotitem.setDownsampling(auto=True)
    curve = plotitem.plot(ttt, y)
curve.setFftMode(True)

app = pg.mkQApp()
app.exec_()

```

Kkkakania:
## Summary
- Fix auto range for `PlotDataItem` when FFT mode and automatic downsampling are both enabled.
- Reuse the mapped dataset before display-only processing so `dataBounds()` is not calculated from a stale, downsampled FFT view.
- Add a regression test for FFT mode with implicit x-values and `PlotItem.setDownsampling(auto=True)`.

Fixes #438.

## Root cause
When `setFftMode(True)` was called after a y-only curve had already been auto-ranged, `updateItems()` recomputed display data using the previous wide x-range (`0..999`). With automatic downsampling enabled, that reduced the FFT display data to only a couple of points. The subsequent auto-range pass then asked `PlotDataItem.dataBounds()` for bounds, but it delegated to the already-downsampled `PlotCurveItem`, so the view could remain near the old pre-FFT range instead of shrinking to the FFT frequency range (`0..0.5` for implicit x-values).

## Validation
- `python -m pytest tests/graphicsItems/test_PlotDataItem.py::test_fft_auto_downsample_autorange_with_implicit_x -q`
- `python -m pytest tests/graphicsItems/test_PlotDataItem.py -q`
- `python -m pytest tests/graphicsItems/test_PlotCurveItem.py tests/graphicsItems/ViewBox/test_ViewBox.py tests/graphicsItems/ViewBox/test_ViewBoxZoom.py -q`
- `python -m pytest tests/graphicsItems -q`
- Manual reproduction of #438: after enabling FFT, the x view range now moves from roughly `[-542, 542]` to `[-0.021, 0.521]`.

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -1441,6 +1441,48 @@ class PlotDataItem(GraphicsObject):
         dataset = self._dataset
         return (None, None) if dataset is None else (dataset.x, dataset.y)
 
+    def _getMappedDataset(self) -> PlotDataset | None:
+        if self._dataset is None:
+            return None
+
+        if self._datasetMapped is not None:
+            return self._datasetMapped
+
+        x = self._dataset.x
+        y = self._dataset.y
+        if y.dtype == bool:
+            y = y.astype(np.uint8)
+        if x.dtype == bool:
+            x = x.astype(np.uint8)
+        if self.opts['subtractMeanMode']:
+            y = y - np.mean(y)
+        if self.opts['fftMode']:
+            x, y = self._fourierTransform(x, y)
+            # Ignore the first bin for fft data if we have a logx scale
+            if self.opts['logMode'][0]:
+                x = x[1:]
+                y = y[1:]
+        if self.opts['derivativeMode']:  # plot dV/dt
+            y = np.diff(self._dataset.y) / np.diff(self._dataset.x)
+            x = x[:-1]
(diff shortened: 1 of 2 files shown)


Kkkakania:
Rebased this branch onto the current `master` and reran the focused PlotDataItem tests locally.

Local check:

- `.venv/bin/python -m pytest tests/graphicsItems/test_PlotDataItem.py -q`

Result: `11 passed`.

Kkkakania:
Friendly follow-up: this branch is current and the full GitHub matrix, mypy, wheel build, coverage, docs, and CodeQL checks are all green. The change remains scoped to the FFT plus automatic-downsampling autorange regression from #438 and includes focused coverage. It is ready for maintainer review.

PR #3508, opened 2026-06-21
Fix ExampleApp unable to search Scatter Plot example by content

pijyoi:
How to trigger the bug:
- launch the ExampleApp
- switch to searching by content
- type "invariant", which is a word found in the Scatter Plot example `ScatterPlot.py`. It will not be found.

Cause:
In the code for searching by content, the nested dictionary of examples is collapsed into a single level. As the 2D and 3D Scatter Plot examples share the same title of 'Scatter Plot', the latter entry of 3D Scatter Plot overwrites the earlier 2D Scatter Plot.

Solution:
Mark found by filename instead of by title. The examples have unique filenames since they are all placed in the same directory.
(A quick and dirty fix would have been to make the titles unique)


pijyoi:
Based on the comment regarding `ExampleApp.py` not using `syntax.py`
https://github.com/pyqtgraph/pyqtgraph/pull/3469#issue-3980338477

As far back as #1302, the contents of `syntax.py` seemed to have been copied into `ExampleApp.py`.
Since #3469 has synchronized the contents, we can import from syntax.py and delete all the redundant lines of code.

My guess is that it had something to do with wanting to make the examples runnable whether as a module or directly as a file.

j9ac9k:
Your guess is almost certainly correct. The examples historically were expected to be able to run from all sorts of various ways.

Thanks for the PR!

PR #3515, opened 2026-06-27
Preserve SVG curve precision for large coordinates

### LINKED ISSUE #2661: SVG exporter output has errors when data values are very large

cookesan:
## Summary
- Preserve SVG curve precision for large plot coordinates by drawing curves around a local export offset before Qt SVG serializes path data.
- Add regression coverage for exporting a curve near x=10_000_000 without collapsing adjacent x positions.
- Make the SVG exporter test module import pyqtgraph.exporters explicitly so the file runs on its own.

Fixes #2661.

## Testing
- QT_QPA_PLATFORM=minimal pytest tests/exporters/test_svg.py tests/graphicsItems/test_PlotCurveItem.py -q
- python -m compileall pyqtgraph/exporters/SVGExporter.py pyqtgraph/graphicsItems/PlotCurveItem.py tests/exporters/test_svg.py
- git diff --check

### DIFF
--- a/pyqtgraph/exporters/SVGExporter.py
+++ b/pyqtgraph/exporters/SVGExporter.py
@@ -258,7 +258,10 @@ def _generateItemSvg(item, nodes=None, root=None, options=None):
         p = QtGui.QPainter()
         p.begin(svg)
         if hasattr(item, 'setExportMode'):
-            item.setExportMode(True, {'painter': p})
+            item.setExportMode(
+                True,
+                {'painter': p, 'svgCoordinatesOffset': True}
+            )
         try:
             p.setTransform(tr)
             opt = QtWidgets.QStyleOptionGraphicsItem()
--- a/pyqtgraph/graphicsItems/PlotCurveItem.py
+++ b/pyqtgraph/graphicsItems/PlotCurveItem.py
@@ -988,32 +988,62 @@ class PlotCurveItem(GraphicsObject):
         )
         do_fill_outline = do_fill and self.opts['fillOutline']
 
-        if do_fill:
-            if self._shouldUseFillPathList(brush):
-                paths = self._getFillPathList(widget)
-            else:
-                paths = [self._getFillPath()]
+        path_transform = None
+        if (
+            self._exportOpts is not False
+            and self._exportOpts.get('svgCoordinatesOffset', False)
+        ):
+            center = self.boundingRect().center()
+            if np.isfinite(center.x()) and np.isfinite(center.y()):
+                path_transform = QtGui.QTransform.fromTranslate(
+                    -center.x(), -center.y()
+                )
+                p.save()
+                p.translate(center)
 
-            profiler('generate fill path')
-            for path in paths:
-                p.fillPath(path, brush)
-            profiler('draw fill path')
-
-        for pen_kind in ['shadowPen', 'pen']:
-            pen = self.opts[pen_kind]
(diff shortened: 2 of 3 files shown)


j9ac9k:
Nice work! This one has been bugging me for ages!

cookesan:
Thanks, glad it helps.

pijyoi:
I noticed that there is an asymmetry in the logic for `_shouldUseFillPathList` and `_showUseDrawLineSegments`.
`_showUseDrawLineSegments` is not allowed when path transform is active, whereas `_shouldUseFillPathList` is.

Now, I understand that must be because `_getLineSegments` doesn't return a path that can be transformed.

However, as both `_shouldUseFillPathList` and `_showUseDrawLineSegments` are workarounds to speed up rendering, and since for exporting being slow isn't such an issue, do you think it would make sense to also *not* use `_shouldUseFillPathList` during SVG export?

cookesan:
Good point. I pushed an update so SVG coordinate offsetting now takes the single fill path route too, matching the stroke path handling. Added a focused exporter test that records _getFillPathList calls during SVG export and keeps that path unused.

pijyoi:
Wouldn't adding a comment as to the rationale be better than enforcing in the tests that a certain private method isn't called? 

In my opinion, the tests shouldn't know too much about implementation internals.

cookesan:
Thanks, that makes sense. I replaced the private-method assertion with a short rationale in `PlotCurveItem.paint` and kept the SVG tests focused on exported output.

Local validation: `PYQTGRAPH_QT_LIB=PySide6 /tmp/pyqtgraph-pr3530-venv/bin/python -m pytest tests/exporters -q` (`9 passed, 2 skipped`).

j9ac9k:
Finally got around to testing this, I can verify it works, I also verified the test on `master` fails, but on this branch works, so I'll merge.  That said, I hate that the logic for this has to reside in `PlotCurveItem` and not exclusively in `SVGExporter` but I don't see how we can have this logic in `SVGExporter`.

PR #3516, opened 2026-06-27
fix: recognize Greek mu as SI prefix

### LINKED ISSUE #3498: Request: recognise Greek mu character as SI prefix
The siEval() and siParse() functions can recognise the micro sign (U+00B5) and the Latin u character (U+0075) as the SI prefix for 10**-6. However, the Greek mu character (U+03BC) is ignored, despite looking almost identical to the micro sign, which causes much frustration. Recognition of the Latin u character is achieved by [line 65 of functions.py](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L65) and this issue could be simply fixed a similar line for the Greek mu character:
```
SI_PREFIX_EXPONENTS['μ'] = -6
```

This change would allow Greek mu characters as inputs, but retain micro signs in all outputs and displays. Further, the note in the documentation of siParse() would no longer be necessary:

> Note that in the above example, the µ symbol is the “micro sign” (UTF-8 0xC2B5), as opposed to the Greek letter mu (UTF-8 0xCEBC).

cookesan:
### Summary
Accept Greek small letter mu as a micro prefix in SI parsing, alongside the existing SI micro sign and ASCII `u`.
Keep SI formatting examples and tests aligned with the formatter's micro-sign output.

### Testing
- `.venv/bin/python -m pytest tests/test_functions.py -q`

Closes #3498

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -61,15 +61,17 @@ Colors = {
 
 SI_PREFIXES = 'yzafpnµm kMGTPEZY'
 SI_PREFIXES_ASCII = 'yzafpnum kMGTPEZY'
+SI_PREFIXES_INPUT = SI_PREFIXES + 'uμ'
 SI_PREFIX_EXPONENTS = dict([(SI_PREFIXES[i], (i-8)*3) for i in range(len(SI_PREFIXES))])
 SI_PREFIX_EXPONENTS['u'] = -6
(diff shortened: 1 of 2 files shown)


j9ac9k:
Thanks for the PR @cookesan and thanks for adding this use-case to the test.  LGTM, merging

PR #3521, opened 2026-06-28
ROI boundingRect: take into account pen width

### LINKED ISSUE #1141: Artifacts when moving ROI with significant pen width

pijyoi:
fixes #1141

~Only handles ROI, not for other ROI types as mentioned in https://github.com/pyqtgraph/pyqtgraph/issues/1141#issuecomment-686897055~

### DIFF
--- a/pyqtgraph/graphicsItems/ROI.py
+++ b/pyqtgraph/graphicsItems/ROI.py
@@ -1066,12 +1066,36 @@ class ROI(GraphicsObject):
             round(pos[1] / snap[1]) * snap[1]
         )
     
-    def boundingRect(self):
+    def dataBoundsRect(self):
         return QtCore.QRectF(0, 0, self.state['size'][0], self.state['size'][1]).normalized()
 
+    def dataBounds(self, ax, **kwds):
+        rect = self.dataBoundsRect()
+        if ax == 0:
+            return (rect.left(), rect.right())
+        else:
+            return (rect.top(), rect.bottom())
+
+    def _pixelPadding(self):
+        # note that this is deliberately not named "pixelPadding" to prevent ViewBox from finding it.
+        # test_ROI.py fails otherwise.
+
+        # pen is assumed to be cosmetic
+        pen = self.currentPen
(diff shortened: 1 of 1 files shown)


pijyoi:
This PR has been fixed to pass the CI. However, as I am not familiar with ROIs, this PR needs a close review from people familiar with the ROI machinery.

The issue:
- `boundingRect()` was used by
    * Qt to determine the paint update area
    * pyqtgraph ROI to determine the data bounds to grab affine slices

The fix:
- define `dataBoundsRect()` and make pyqtgraph ROI use it to determine data bounds
- modify `boundingRect()` to include pen width
- define `dataBounds()` for `ViewBox`'s benefit

j9ac9k:
Gonna tag @outofculture and @campagnola as they are big ROI users.  @ixjlyons not sure if you use ROIs much.

j9ac9k:
I'm inclined to merge this as we haven't heard back from anyone, and the diff LGTM. 

Thanks @pijyoi for this PR 👍🏻

PR #3523, opened 2026-06-28
Fix filled log plot auto-connect

### LINKED ISSUE #2312: artifacts in fill under curve for histograms in log mode

cookesan:
## Summary
- keep filled `PlotDataItem` curves continuous when `connect="auto"` and log mode creates non-finite points
- add a seeded histogram regression for filled center-step plots in log Y mode

Fixes #2312.

## Tests
- `QT_QPA_PLATFORM=minimal .venv/bin/pytest tests/graphicsItems/test_PlotDataItem.py::test_filled_log_plot_auto_connects_across_nonfinite_values -q`
- `QT_QPA_PLATFORM=minimal .venv/bin/pytest tests/graphicsItems/test_PlotDataItem.py -q`
- `QT_QPA_PLATFORM=minimal .venv/bin/pytest tests/graphicsItems/test_PlotCurveItem.py -q`
- `QT_QPA_PLATFORM=minimal .venv/bin/pytest tests/graphicsItems -q`
- `.venv/bin/python -m compileall pyqtgraph/graphicsItems/PlotDataItem.py tests/graphicsItems/test_PlotDataItem.py`
- `git diff --check`

### DIFF
--- a/pyqtgraph/graphicsItems/PlotDataItem.py
+++ b/pyqtgraph/graphicsItems/PlotDataItem.py
@@ -1402,6 +1402,10 @@ class PlotDataItem(GraphicsObject):
                     # all points can be connected, and no further check is needed.
                     curveArgs['connect'] = 'all'
                     curveArgs['skipFiniteCheck'] = True
+                elif self.opts['fillLevel'] is not None:
+                    # Filled curves need one continuous outline across log(0) gaps.
+                    curveArgs['connect'] = 'all'
+                    curveArgs['skipFiniteCheck'] = False
                 else:   # True or None
                     # True: (we checked and found non-finites)
                     #   don't connect non-finites
--- a/tests/graphicsItems/test_PlotDataItem.py
+++ b/tests/graphicsItems/test_PlotDataItem.py
@@ -121,6 +121,40 @@ def test_nonfinite():
     _assert_equal_arrays( dataset.x, x_log )
     _assert_equal_arrays( dataset.y, y_log )
 
+def test_filled_log_plot_auto_connects_across_nonfinite_values(monkeypatch):
+    rng = np.random.default_rng(seed=0)
+    values = np.hstack([rng.normal(size=500), rng.normal(size=260, loc=4)])
+    y, x = np.histogram(values, bins=np.linspace(-3, 8, 40))
+
+    pdi = pg.PlotDataItem(
+        x,
+        y,
+        stepMode="center",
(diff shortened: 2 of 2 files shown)


pijyoi:
#3071 implements filling under the curve when connect="finite". You can see how it works in the MouseSelection example.

It seems to me that in this PR, this drawing mode would be bypassed for all plots, Log mode or not?

pijyoi:
Disregard the above comment. The logic in this PR only applies if `connect="auto"`

So the below snippet will work properly. i.e. break the curve into continuous segments and fill them.
```python
import numpy as np

import pyqtgraph as pg

pg.mkQApp()
plt = pg.PlotWidget()

x = np.arange(0, 1000, dtype=float)
x[(250, 500, 750),] = np.nan
y=np.sin(np.linspace(1, 21, 1000))
plt.plot(x=x, y=y, pen='g', connect="finite", fillLevel=0, brush=(50,50,200,100))

plt.show()
pg.exec()
```

PR #3529, opened 2026-06-28
Clip oblique InfiniteLine painting to the ViewBox

### LINKED ISSUE #1089: Infinite line hangs when drawn on plot with large difference between `x` and `y` ranges
### Short description
When adding an infinite line to a graph with a large difference between `x` and `y` ranges, the program hangs. It can't even be killed with Ctrl+C.

### Code to reproduce
```python
import sys
from PyQt5 import QtWidgets
from PyQt5.QtCore import Qt, QPointF
import pyqtgraph as pg


    def __init__(self):

        self.resize(800, 600)

        self.setCentralWidget(glw)

        x = [0, 1]  # replacing this line with x = [0, 1e8] would not reproduce the problem
        y = [0, 1e8]

        plot = pg.PlotItem()
        line = plot.plot(x=x, y=y, pen="r")
        inf_line = pg.InfiniteLine(
            pos=QPointF(0, 0), angle=45,
            pen=pg.mkPen('r', width=2, style=Qt.DashLine),
        )
        plot.addItem(inf_line)
        plot.addItem(line)
        glw.addItem(plot, 0, 0)


if __name__ == '__main__':
    ex = Example()
    ex.show()
    sys.exit(app.exec_())
```

### Tested environment(s)
This was reproduced on two machines with the following configurations:
 * PyQtGraph version: `'0.10.0'` and `'0.10.0'`
 * Qt Python binding: `'PyQt5 5.12.3 Qt 5.12.5'` and `‘PyQt5 5.13.1 Qt 5.13.0’`
 * Python version: `'3.8.0'` and `'3.6.5'`
 * NumPy version: `'1.17.2'` and `'1.16.2'`
 * Operating system: Linux and OSX
 * Installation method: `conda` and `pip`

cookesan:
Fixes #1089.

## Summary
- Clip oblique `InfiniteLine` drawing bounds to the visible `ViewBox` intersection before applying `span`.
- This keeps dashed lines with extreme axis-range ratios from sending huge off-screen endpoints to Qt's painter.
- Add regression coverage for the large-range dashed-line case.

## Tests
- `PYTHONPATH=. python -m pytest tests/graphicsItems/test_InfiniteLine.py -q`
- `PYTHONPATH=. python -m pytest tests/graphicsItems/test_InfiniteLine.py tests/graphicsItems/PlotItem/test_PlotItem.py tests/graphicsItems/ViewBox/test_ViewBox.py -q`
- `PYTHONPATH=. python -m pytest tests/graphicsItems -q`
- `PYTHONPATH=. python -m pytest tests -q`
- `git diff --check`

### DIFF
--- a/pyqtgraph/graphicsItems/InfiniteLine.py
+++ b/pyqtgraph/graphicsItems/InfiniteLine.py
@@ -295,6 +295,46 @@ class InfiniteLine(GraphicsObject):
             self.span = (mn, mx)
             self.update()
 
+    def _visibleLineBounds(self):
+        if self.angle % 90 == 0:
+            return None
+
+        view = self.getViewBox()
+        if view is None:
+            return None
+
+        viewRect = view.boundingRect()
+        if viewRect.isEmpty():
+            return None
+
+        corners = (
+            viewRect.topLeft(),
+            viewRect.topRight(),
+            viewRect.bottomRight(),
+            viewRect.bottomLeft(),
+        )
+        points = [self.mapFromItem(view, point) for point in corners]
+        intersections = []
+
+        for p1, p2 in zip(points, points[1:] + points[:1]):
+            y1 = p1.y()
+            y2 = p2.y()
+
+            if y1 == 0 and y2 == 0:
+                intersections.extend([p1.x(), p2.x()])
+                continue
+
+            if (y1 <= 0 <= y2) or (y2 <= 0 <= y1):
+                if y1 == y2:
+                    continue
+                ratio = -y1 / (y2 - y1)
+                intersections.append(p1.x() + ratio * (p2.x() - p1.x()))
+
+        if len(intersections) < 2:
+            return None
+
+        return min(intersections), max(intersections)
+
     def _computeBoundingRect(self):
         #br = UIGraphicsItem.boundingRect(self)
         vr = self.viewRect()  # bounds of containing ViewBox mapped to local coords.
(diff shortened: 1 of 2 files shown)


cookesan:
One CI leg failed in `tests-pip (macos-latest, pyqt, 3.14)` on `tests/graphicsItems/test_ErrorBarItem.py::test_ErrorBarItem_defer_data`:

`assert 1.0 > 1.0` from `r_ebi.height() > r_empty_ebi.height()`.

This is outside the `InfiniteLine` path touched here. The local full suite passed (`1088 passed, 708 skipped, 8 xfailed`), and every other live check is green. I tried to rerun the failed job, but reruns need repo admin rights.

j9ac9k:
> One CI leg failed in `tests-pip (macos-latest, pyqt, 3.14)` on `tests/graphicsItems/test_ErrorBarItem.py::test_ErrorBarItem_defer_data`:
> 
> `assert 1.0 > 1.0` from `r_ebi.height() > r_empty_ebi.height()`.
> 
> This is outside the `InfiniteLine` path touched here. The local full suite passed (`1088 passed, 708 skipped, 8 xfailed`), and every other live check is green. I tried to rerun the failed job, but reruns need repo admin rights.

This test fails intermittently on CI systems.  We haven't figured out a good way to reliably address it, regardless, don't worry about it.

pyqtgraph/graphicsItems/InfiniteLine.py
```diff
@@ -295,6 +295,46 @@ def setSpan(self, mn, mx):
             self.span = (mn, mx)
             self.update()
 
+    def _visibleLineBounds(self):
+        if self.angle % 90 == 0:
+            return None
+
+        view = self.getViewBox()
+        if not isinstance(view, ViewBox):
```
j9ac9k: consider switching to 

```python
if view is None:
    return None
```

No need for an isinstance check, much less one that has a negation in front of it.

j9ac9k:
Hi @cookesan,  This PR LGTM minus the one nit-pick I added the inline comment for.  Thanks for submitting!

cookesan:
Thanks, addressed the nit by switching `_visibleLineBounds` to the direct `view is None` guard. The latest check run is green across the matrix.

Issue #3378, opened 2025-10-01
Recent Parameter changes broke getValues for nested Groups


Tillsten:
### Short description
Some recent changes in Parameters make getValues unuseable with group parameters since they do not have a value anymore.

### Code to reproduce

```python
from pyqtgraph.parametertree import Parameter

children = [Parameter.create(name=f"Child {c}", type="float", value=0.0) for c in "abc"]
group = Parameter.create(name="New Group", type="group", children=children)
upper_group = Parameter.create(name="Upper Group", type="group", children=[group])
upper_group.getValues()
```

### Expected behavior
Before recent changes it returned.
```
OrderedDict([('New Group',
              (None,
               OrderedDict([('Child a', (0.0, OrderedDict())),
                            ('Child b', (0.0, OrderedDict())),
                            ('Child c', (0.0, OrderedDict()))])))])

```
### Real behavior

Now it raises an ValueError since the GroupParameter does not have a value anymopre 

```
An error occurred?
Post the full traceback inside these 'code fences'!
```

### Tested environment(s)

 * PyQtGraph version: 0.14.0dev0'

Commit **6eb0bf1a928c3e8fef332bbebe8a9da3be0ab19a**

pijyoi:
On pyqtgraph-0.13.7, running your script with `-Wd' yields
```
C:\work\repos\pyqtgraph\pyqtgraph\parametertree\Parameter.py:140: DeprecationWarning: Parameter has no default value. Pass a default, or use setDefault(). This will no longer set an implicit default after January 2025.
  return cls(**opts)
C:\work\repos\pyqtgraph\pyqtgraph\parametertree\Parameter.py:360: DeprecationWarning: Parameter has no value set. Pass an initial value or default, or use setValue() or setDefault(). This will be an error after January 2025.
  vals[ch.name()] = (ch.value(), ch.getValues())
```

The warning became an error in ece87378d2a98b0d0ebde35ee55e9d35e9cd2130 (2025-03-24)

Tillsten:
The two solutions I see are just giving the GroupParameter a value of None or change the getValues method to handle items without values.

j9ac9k:
@ntjess is most familiar w/ parameter trees so maybe he'll chime in, but I'm partial to think that just passing a `None` value in for the initial value seems to be the way to go at least in the short term.

seb5g:
[...]

PR #3533, opened 2026-07-02
returning None if no value has been set on a parameter

seb5g:
removing the ValueError recently introduced for no clear reason...

This PR solves #3378 by reverting the Value Error to just returning None if no value has bee set!

### DIFF
--- a/pyqtgraph/parametertree/Parameter.py
+++ b/pyqtgraph/parametertree/Parameter.py
@@ -400,10 +400,7 @@ class Parameter(QtCore.QObject):
         """
         Return the value of this Parameter. Raises ValueError if no value has been set.
         """
-        try:
-            return self.opts['value']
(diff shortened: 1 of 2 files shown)


seb5g:
@outofculture and @j9ac9k you recently worked on PR leading to this ValueError stuf. We discussed it partly on discord but here is the place where we can wrap things up on the subject!

j9ac9k:
I'm good with merging this, but given that it's a change of API behavior and no tests failed, we should likely add a test that checks for this use-case.

seb5g:
added two tests making sure :

* the value method is returning None on no value parameter
* getValues doesn't raise errors when called on complex parameters with groups (no value)

j9ac9k:
blerg, completely forgot about this, I apologize... looking into this now.

Issue #3488, opened 2026-05-08
Document types correctly


StSav012:
Numerous functions possess docs like these:
https://github.com/pyqtgraph/pyqtgraph/blob/d588dd3ec30915e61c8496a0fe1db21fc25e4da5/pyqtgraph/graphicsItems/PlotDataItem.py#L1161-L1164

Disregard what the arguments mean. The _type_ is actually wrong. Type checkers expect the type to be what each of `args` or `kwargs` is. For instance, if a function accepts bare `numpy` arrays of `np.double`, the docs should read `*args: np.ndarray` or `*args: NDArray[np.double]`.

Like it currently is, it causes a barrage of false-positive warnings from linters. They actually distract more than help.

So, please change the docs to `*args: object` and `**kwargs: object` unless a better type is known.

---

Another example is the following:

https://github.com/pyqtgraph/pyqtgraph/blob/d588dd3ec30915e61c8496a0fe1db21fc25e4da5/pyqtgraph/graphicsItems/AxisItem.py#L323-L350

Here, I never get to pass a `tuple` or a `dict` to the function. Instead, I can call it like this:
```python
axes.setLogMode(True)
axes.setLogMode(True, False)
axes.setLogMode(x=True)
axes.setLogMode(y=True)
axes.setLogMode(x=True, y=False)
```
However, both the type annotation and the docs suggest that the function demands a `tuple` or a `dict`.

Instead, the function should start like this:
```python
    def setLogMode(self, *args: bool, **kwargs: bool):
        """
        Set log scaling for x and / or y axes.

        If two positional arguments are provided, the first will set log scaling
        for the x axis and the second for the y axis. If a single positional
        argument is provided, it will set the log scaling along the direction of
        the AxisItem. Alternatively, x and y can be passed as keyword arguments.

        If an axis is set to log scale, ticks are displayed on a logarithmic scale and
        values are adjusted accordingly. The linked ViewBox will be informed of the
        change.

        Parameters
        ----------
        *args : bool
            If a single value is provided, sets log mode regardless of orientation. 
            If two values are given, the first element toggles log mode for x-axis,
            and the second element toggles log mode for the y-axis.
        **kwargs : dict
            Values for `x` and `y` axes, ``bool``, to set the log mode
            for the respective `x` or `y` axis.  Trying to set the `y` axis log mode
            while this axis item is horizontal (or vice versa) will be ignored.

        See Also
        --------
[...]

PR #3534, opened 2026-07-09
docs: clarify AxisItem varargs documentation

dmtr-karan:
Partially addresses #3488.

This updates AxisItem docstrings to describe caller-facing `*args` and `**kwargs` usage as positional and keyword arguments rather than as tuple/dict containers.

Scope:
- Clarifies `setStyle`
- Clarifies `setLogMode`
- Clarifies `setPen`, `setTextPen`, and `setTickPen`
- Leaves behavior unchanged

Validation:
- `pytest tests/graphicsItems -k AxisItem -vv`
- `git diff --check`

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -153,8 +153,8 @@ class AxisItem(GraphicsWidget):
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


j9ac9k:
This PR breaks all sorts of type hints within the docs.

PR #3536, opened 2026-07-09
docs: clarify graphicsItems varargs documentation

dmtr-karan:
Partially addresses #3488.

This updates graphicsItems docstrings to describe caller-facing `*args` and `**kwargs` usage as positional and keyword arguments rather than as tuple/dict containers.

Scope:
- Clarifies `PlotItem`
- Clarifies `PlotDataItem`
- Clarifies `ImageItem`
- Leaves behavior unchanged

Validation:
- `pytest tests/graphicsItems -k PlotItem -vv`
- `pytest tests/graphicsItems -k PlotDataItem -vv`
- `pytest tests/graphicsItems -k ImageItem -vv`
- `git diff --check`

### DIFF
--- a/pyqtgraph/graphicsItems/ImageItem.py
+++ b/pyqtgraph/graphicsItems/ImageItem.py
@@ -41,7 +41,7 @@ class ImageItem(GraphicsObject):
     ----------
     image : np.ndarray or None, default None
         Image data.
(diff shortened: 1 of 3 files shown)


j9ac9k:
Thanks for the PR, but the types are incorrect/

Superseded by #3584

PR #3539, opened 2026-07-11
Keep subnormal plot ranges finite

### LINKED ISSUE #2370: Very small range data can crash application due to divide by zero issues

Kkkakania:
## Summary

- clamp only overflowed ViewBox scale factors to the signed maximum finite float
- apply the same finite-scale guard when AxisItem maps extremely small ranges to tick coordinates
- cover positive and zero-crossing subnormal ranges on both horizontal and vertical axes

## Root cause

For a nonzero range near `1e-310`, dividing a normal pixel width or height by that range overflows to infinity. The resulting Qt transform contains `inf`/`nan`, and AxisItem generates non-finite tick endpoints. This can cause platform-dependent rendering failures or crashes.

A fixed epsilon would change the visible scale of legitimate small-valued data. This change instead preserves the requested range and clamps only an already-overflowed ratio, retaining its sign. Ordinary scales and zero-range handling are unchanged.

## Tests

- focused ViewBox and AxisItem regression matrix: 8 passed
- `QT_QPA_PLATFORM=offscreen python -m pytest tests -q`: 1094 passed, 723 skipped, 8 xfailed
- repository pre-commit hooks for all four changed files: all applicable hooks passed; the removed upstream `fix-encoding-pragma` hook was skipped

Fixes #2370

### DIFF
--- a/pyqtgraph/graphicsItems/AxisItem.py
+++ b/pyqtgraph/graphicsItems/AxisItem.py
@@ -1,5 +1,6 @@
+import sys
 import weakref
-from math import ceil, floor, frexp, isfinite, log10, sqrt
+from math import ceil, copysign, floor, frexp, isfinite, log10, sqrt
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


j9ac9k:
Tweaking AxisItem and ViewBox can be dangerous, but the diff here looks good.  Thanks for the PR!

Kkkakania:
Glad to see this merged. The focused regression coverage exercises both `AxisItem` tick coordinates and `ViewBox` transforms for subnormal ranges, so the finite-clamping behavior is protected on both paths.

PR #3540, opened 2026-07-13
Make makeARGB reference test platform independent

Xeonacid:
NumPy's float-to-integer conversion of NaN varies by platform. On RISC-V, converting NaN to uint8 produces 255, while the generated reference arrays record the zero produced on x86_64. makeARGB already documents those color values as undefined and guarantees transparency by setting the alpha channel to zero, so comparing the RGB bytes rejects valid output.

Build a per-pixel NaN mask from the test input and exclude only the three color bytes at those pixels from the generated-reference comparison. The test continues to check alpha for NaN-containing pixels and every channel for all other pixels, including finite RGBA pixels whose alpha is zero.

Validated on Arch Linux x86_64 and RISC-V with:

    QT_QPA_PLATFORM=offscreen pytest -q tests/test_makeARGB.py::test_makeARGB_against_generated_references

Both runs completed with 594 passed and 702 skipped.


Xeonacid:
The CI failure seems unrelated to this PR.

j9ac9k:
Thanks for the PR @Xeonacid  This LGTM, Merging.

PR #3541, opened 2026-07-19
refactor: create QtOpenGL shim

pijyoi:
All code previously using Qt's built-in OpenGL functions had to import QtOpenGL for Qt6 and QtGui for Qt5. This localizes all that logic into a lazily-loaded QtOpenGL shim.

Some changes to `PlotCurveItem`:
1) Don't support user enabling OpenGL other than by `setConfigOption('useOpenGL', True)`
    * Previously, there was a user doing that (#2257) but apparently that is no longer the case (https://github.com/pyqtgraph/pyqtgraph/pull/3055#discussion_r1646051960)
2) bump the fill chunk size from 50 to 150
    * 150 performs slightly better, and should be more resistant to bad degradations in case we enter edge-case modes that prefer big chunks.

### DIFF
--- a/pyqtgraph/Qt/QtOpenGL/__init__.py
+++ b/pyqtgraph/Qt/QtOpenGL/__init__.py
@@ -1,0 +1,11 @@
+import importlib
+from .. import QtVersionInfo, QT_LIB
+if QtVersionInfo[0] >= 6:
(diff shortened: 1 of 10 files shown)


j9ac9k:
Love the diff @pijyoi thanks for the PR!

PR #3545, opened 2026-07-26
Fix ScatterPlotItem compositionMode handling

### LINKED ISSUE #3117: CompositionMode is not set if sepcified in ScatterPlotItem constructor and setData
setData doc says it will set compositionMode if specified but it does not, looking at the code it seems it never tries to.

JosephDemarest:
## Summary

- honor `compositionMode` passed to `ScatterPlotItem` construction and `setData()`
- retain the selected mode when later data updates omit the option
- add regression coverage for construction and repeated `setData()` calls

Closes #3117

## Testing

- `python -m pytest tests/graphicsItems/test_ScatterPlotItem.py -q` — 3 passed
- `python -m pytest tests -q` — 1116 passed, 705 skipped, 8 xfailed
- `python -m pytest pyqtgraph/examples -q -n 2` — 98 passed, 3 skipped


j9ac9k:
Thanks for the PR, indeed this should address it.

PR #3546, opened 2026-07-26
Fix parsing of symbol-prefixed suffixes

### LINKED ISSUE #3115: functions: FLOAT_REGEX cannot properly match suffix that starts with "°" (degree) character
### Short description

The `FLOAT_REGEX` in `functions.py` cannot match suffixes that start with the `°` character.

### Code to reproduce

```python
print(fn.siEval('451 °F'))  # fn.FLOAT_REGEX used by default
```

### Expected behavior

The code prints `451.0`.

### Real behavior

The code prints `45.0`.

### Additional context

Seems to work fine if I remove the `\w` in the suffix group of the `FLOAT_REGEX`:

```python
>>> import pyqtgraph.functions as fn
>>> fn.siEval('451 °F', regex=re.compile(fn.FLOAT_REGEX.pattern.replace('\w', '')))
451.0
```

JosephDemarest:
Fixes #3115.

`FLOAT_REGEX_PERIOD` and `FLOAT_REGEX_COMMA` previously required suffixes to begin with a word character. For values such as `451 °F`, that made the regex backtrack and parse the number as `45` with a suffix of `1 °F`.

Allowing suffixes to begin with any non-whitespace character parses degree and percent symbols correctly while retaining the existing distinction between a bare unit (`100 m`) and an SI-prefixed unit (`100 mV`). Regression coverage includes period- and comma-decimal parsing and `siEval`.

Testing:
- `pytest tests/test_functions.py -q` — 178 passed
- `pytest tests -q` — 1118 passed, 705 skipped, 8 xfailed
- `pytest pyqtgraph/examples -q` — 98 passed, 3 skipped
- all applicable pre-commit hooks passed (`fix-encoding-pragma` skipped because that configured upstream hook has been removed)

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -11,7 +11,6 @@ import struct
 import sys
 import warnings
 from collections import OrderedDict
-
 from typing import TypeAlias, TypedDict
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


j9ac9k:
Thanks for the PR, I appreciate the extra test.  LGTM, merging.

PR #3547, opened 2026-07-26
Fix integer SpinBox text after coercion

### LINKED ISSUE #3133: Parameter of type int does not update correctly when a float is submitted
### Short description
When submitting a float number to a int type parameter, the value of that parameter does not correct itself if the integer part of the submitted float is the same. For example, when submiting 2.5 to an integer parameter the value will self correct to 2. But if you then submit 2.1 the shown value does not self correct anymore. I suspect this is because the underlying integer value (2) did not change.

### Code to reproduce

```python
import sys
from PyQt5.QtWidgets import QApplication, QMainWindow, QVBoxLayout, QWidget
import pyqtgraph as pg

    def __init__(self):
        super().__init__()
        
        # Create the main widget and layout
        main_widget = QWidget()
        layout = QVBoxLayout(main_widget)
        
        # Create the parameter tree with one int parameter
        params = [
            {'name': 'Integer Parameter', 'type': 'int', 'value': 10}
        ]
        parameter = Parameter.create(name='params', type='group', children=params)
        
        # Create the parameter tree widget
        param_tree = ParameterTree()
        param_tree.setParameters(parameter, showTop=True)
        
        # Add the parameter tree to the layout
        layout.addWidget(param_tree)
        
        # Set the central widget
        self.setCentralWidget(main_widget)

if __name__ == '__main__':
    app = QApplication(sys.argv)
    window = MainWindow()
    window.show()
    sys.exit(app.exec_())
```

### Expected behavior
When submitting a float on an integer parameter the value should always self correct independently of the previously held value.

### Real behavior
The float value is only corrected if it causes a change to the integer part.

### Tested environment(s)

 * PyQtGraph version: 0.13.7
 * Qt Python binding: PyQt5 5.15.11 Qt 5.15.2
 * Python version: 3.10.11
 * NumPy version: 1.26.2
 * Installation method: pip

JosephDemarest:
Fixes #3133.

When an integer `SpinBox` interprets edited text such as `2.1`, it coerces the value to `2`. If the stored value is already `2`, `editingFinishedEvent` returned without updating the line edit, leaving the displayed text inconsistent with the actual integer value.

The equal-value path now reformats the text without emitting a value-change signal. The regression test follows the reported sequence (`2.5`, then `2.1`) and verifies that both edits display the canonical integer text. The configured `pycln` hook also removed the module's unused `warnings` import.

Testing:
- `pytest tests/widgets/test_spinbox.py -q` — 45 passed
- `pytest tests -q` — 1116 passed, 705 skipped, 8 xfailed
- `pytest pyqtgraph/examples -q` — 98 passed, 3 skipped
- all applicable pre-commit hooks passed (`fix-encoding-pragma` skipped because that configured upstream hook has been removed)

### DIFF
--- a/pyqtgraph/widgets/SpinBox.py
+++ b/pyqtgraph/widgets/SpinBox.py
@@ -1,6 +1,5 @@
 import decimal
 import re
-import warnings
 from math import isinf, isnan
 
 from .. import functions as fn
@@ -610,6 +609,7 @@ class SpinBox(QtWidgets.QAbstractSpinBox):
         if val is False:
             return
         if val == self.val:
+            self.updateText()
             return
         self.setValue(val, delaySignal=False)  ## allow text update so that values are reformatted pretty-like
 
--- a/tests/widgets/test_spinbox.py
+++ b/tests/widgets/test_spinbox.py
@@ -90,6 +90,21 @@ def test_evalFunc():
     assert sb.value() == 100
 
 
+def test_SpinBox_reformats_unchanged_integer_value():
+    sb = pg.SpinBox(int=True)
+    sb.setValue(10)
+
+    sb.lineEdit().setText('2.5')
+    sb.editingFinishedEvent()
+    assert sb.value() == 2
+    assert sb.text() == '2'
+
+    sb.lineEdit().setText('2.1')
+    sb.editingFinishedEvent()
+    assert sb.value() == 2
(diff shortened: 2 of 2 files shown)


j9ac9k:
Thanks for the PR, I can confirm the issue is indeed fixed.

PR #3548, opened 2026-07-28
split out low-usage but large functions from functions.py

pijyoi:


j9ac9k:
Anything to get functions.py down!

PR #3550, opened 2026-07-28
pass clickEvent to other items if acceptedItem does not accept the event 

### LINKED ISSUE #3549: lastHoverEvent acceptedItem silently swallows all click events
### Short description
In the sendClickEvent function, the branch where lastHoverEvent is not None and acceptedItem is not none, after the code calls acceptedItem.mouseClickEvent(ev), the sendClickEvent returns - regardless of if acceptedItem accepted the event or not. This makes it impossible to, for example, click on an item within the bounds of an ROI, because the ROI swallows the click.

### Expected behavior
If the acceptedItem does not accept the mouseClickEvent, then sendClickEvent should pass it on to any other items "nearby" - the same as it does in the `acceptedItem is None` branch.

### Real behavior
Only the acceptedItem ever gets the mouseClickEvent, even if it explicitly calls `ev.ignore()`

### Tested environment(s)

 * PyQtGraph version: 0.14.0 (though I checked that the code is the same in GitHub as well)
 * as this is the expected behavior of the GraphicsScene sendClickEvent code, the version of python/OS/etc are irrelevant.

### Additional context
I am using a PolyLineROI subclass, in which I have implemented `mouseClickEvent` to ignore the event (as I only need mouse drag). Ideally, in this scenario, since the ROI ignored the event, the event should pass down to the next item in the view hierarchy (if any). However, due to the way sendClickEvent is structured, this does not happen, and items within the ROI bounds never get the click event.

ibrewster:
Currently, the `GraphicsScene` function `sendClickEvent` just silently swallows the click event if `lastHoverEvent` is not `None` and `acceptedItem` calls `ignore()` on the event. This patch changes the behavior such that if `acceptedItem` doesn't accept the event, the event will be passed on to any other `itemsNearEvent(ev)` (if any) to be handled.

Fixes #3549

### DIFF
--- a/pyqtgraph/GraphicsScene/GraphicsScene.py
+++ b/pyqtgraph/GraphicsScene/GraphicsScene.py
@@ -368,7 +368,7 @@ class GraphicsScene(QtWidgets.QGraphicsScene):
                     acceptedItem.mouseClickEvent(ev)
                 except:
                     debug.printExc("Error sending click event:")
-            if acceptedItem is None or not ev.isAccepted():
+            if not ev.isAccepted() or acceptedItem is None:
                 for item in self.itemsNearEvent(ev):
                     if not item.isVisible() or not item.isEnabled() or item is acceptedItem:
                         continue
--- a/pyqtgraph/GraphicsScene/GraphicsScene.py
+++ b/pyqtgraph/GraphicsScene/GraphicsScene.py
@@ -370,7 +370,13 @@ class GraphicsScene(QtWidgets.QGraphicsScene):
                     debug.printExc("Error sending click event:")
             if not ev.isAccepted() or acceptedItem is None:
                 for item in self.itemsNearEvent(ev):
-                    if not item.isVisible() or not item.isEnabled() or item is acceptedItem:
+                    if any(
+                        (
+                            not item.isVisible(),
+                            not item.isEnabled(),
+                            item is acceptedItem
+                        )
+                    ):
                         continue
                     if hasattr(item, 'mouseClickEvent'):
                         ev.currentItem = item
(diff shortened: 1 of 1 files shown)


pyqtgraph/GraphicsScene/GraphicsScene.py
```diff
@@ -368,9 +368,9 @@ def sendClickEvent(self, ev):
                     acceptedItem.mouseClickEvent(ev)
                 except:
                     debug.printExc("Error sending click event:")
-            else:
+            if acceptedItem is None or not ev.isAccepted():
```
j9ac9k:
```suggestion
            if not ev.isAccepted() or acceptedItem is None:
```

The `or not ...` bit was a bit tougher to read

pyqtgraph/GraphicsScene/GraphicsScene.py
```diff
@@ -368,9 +368,9 @@ def sendClickEvent(self, ev):
...
                 for item in self.itemsNearEvent(ev):
-                    if not item.isVisible() or not item.isEnabled():
+                    if not item.isVisible() or not item.isEnabled() or item is acceptedItem:
```
j9ac9k:
```suggestion
                    if any(
                        (
                            not item.isVisible(),
                            not item.isEnabled(),
                            item is acceptedItem
                        )
                    ):
```

j9ac9k:
Thanks @ibrewster for making the changes, I hate to get that kind of nit-picky but I do thing this is easier to read.  Change LGTM, merging.

j9ac9k:
ugh...maybe merging later, looks like github isn't doing so good... 

image

ibrewster:
> Thanks @ibrewster for making the changes, I hate to get that kind of nit-picky but I do thing this is easier to read. Change LGTM, merging.

Yeah, everyone has a different style. I tend more toward the pragmatic "as long as it is functionally good and not horrible to read" - no nested comprehensions with function calls or the like, but not overly concerned with the most readable possible. It's all good :)

PR #3571, opened 2026-08-09
Fix locale-aware SI value parsing

### LINKED ISSUE #3568: siEval() does not work with the comma decimal separator
Latest `master`.

```
from PySide6.QtCore import QLocale
from pyqtgraph import functions, siEval

regex = functions.float_regex_for_locale(QLocale('de_DE'))  # functions.FLOAT_REGEX_COMMA
siEval('3,14', regex=regex)
```

raises `ValueError: could not convert string to float: '3,14'` instead of returning `3.14`.

Internally `siEval()` calls `siParse()` -- which returns `('3,14', '', '')` -- and then tries to convert with `typ(val)` (`typ` defaults to `float`), which raises.

To fix a similar issue (there with `Decimal`, which would raise `InvalidOperation` instead of `ValueError`), `SpinBox.interpret()` replaces `,` with `.`. If that approach is acceptable, the same replacement can be done within `siEval()` or `siParse()`.

### LINKED ISSUE #3569: siParse()/siEval() + SpinBox: a mismatched decimal separator silently truncates the number and discards the SI prefix
```
from pyqtgraph import functions, siEval, siParse

# default regex is FLOAT_REGEX_PERIOD, so this is locale-independent
siParse('3.14 kB')  # ('3.14', 'k', 'B')   OK
siParse('3,14 kB')  # ('3', '', ',14 kB')  NOK, expected: ValueError
siEval('3.14 kB')   # 3140.0               OK
siEval('3,14 kB')   # 3.0                  NOK, expected: ValueError

# same in the other direction
siParse('3.14 kB', regex=functions.FLOAT_REGEX_COMMA)  # ('3', '', '.14 kB')
siEval('3.14 kB', regex=functions.FLOAT_REGEX_COMMA)   # 3.0
```

The integer part before the unexpected separator is returned, and everything after it - including the SI prefix - is absorbed into the suffix. For example `siEval('0.999 kB') == 999.0`, but `siEval('0,999 kB') == 0.0`.

The cause seems to be regexes in `functions.py`, specifically the suffix part: `(?P<suffix>\S.*)`. The suffix can start with any non-whitespace character, so `,14 kB` is an acceptable choice.

A possible fix would be to tighten the constraint and disallow suffixes starting with `.`, `,` or a digit: `(?P<suffix>[^\s\d.,].*)`.

(Without digits, `siParse('31,4')` would happily return `('3', '', '1,4')`.)

Or, all together:

```
#For comma as decimal separator
FLOAT_REGEX_COMMA = re.compile(r'(?P<number>[+-]?((((\d+(,\d*)?)|(\d*,\d+))([eE][+-]?\d+)?)|((?i:nan)|(inf))))\s*((?P<siPrefix>[' + SI_PREFIXES_INPUT + r']?)(?P<suffix>[^\s\d.,].*))?$')
#For period as decimal separator
FLOAT_REGEX_PERIOD = re.compile(r'(?P<number>[+-]?((((\d+(\.\d*)?)|(\d*\.\d+))([eE][+-]?\d+)?)|((?i:nan)|(inf))))\s*((?P<siPrefix>[' + SI_PREFIXES_INPUT + r']?)(?P<suffix>[^\s\d.,].*))?$')
```

**Impact on `SpinBox`**

`SpinBox` selects its regex from the locale (via `float_regex_for_locale()`), so in a comma locale a user may still type `1.23`. The `.23` is discarded and the widget accepts the value as `1`.

With the proposed change the input is correctly marked invalid because within `SpinBox.interpret()`, `siParse()` raises instead:

```
# tokenize into numerical value, si prefix, and suffix
try:
    val, siprefix, suffix = fn.siParse(strn, self.opts['regex'], suffix=self.opts['suffix'])
except Exception:
    return False
```

Kkkakania:
## Summary

- normalize comma-decimal values before converting parsed SI values to numeric types
- prevent numeric, period, or comma characters from being accepted as the start of a unit suffix
- add coverage for comma-decimal SI evaluation
- add symmetric rejection coverage for mismatched decimal separators

## Root cause

The comma-aware regex returned the localized numeric token unchanged, so `float()` could not convert it. Separately, both locale regexes allowed any non-whitespace suffix start, which let a mismatched decimal separator and the remaining digits be treated as a suffix. That silently truncated the numeric value and could discard the intended SI prefix.

## Validation

- `pytest tests/test_functions.py tests/widgets/test_spinbox.py -q` (228 passed, 2 skipped)
- `git diff --check`

Fixes #3568.
Fixes #3569.

### DIFF
--- a/pyqtgraph/functions.py
+++ b/pyqtgraph/functions.py
@@ -66,9 +66,9 @@ SI_PREFIX_EXPONENTS['u'] = -6
 SI_PREFIX_EXPONENTS['μ'] = -6
 
 #For comma as decimal separator
-FLOAT_REGEX_COMMA = re.compile(r'(?P<number>[+-]?((((\d+(,\d*)?)|(\d*,\d+))([eE][+-]?\d+)?)|((?i:nan)|(inf))))\s*((?P<siPrefix>[' + SI_PREFIXES_INPUT + r']?)(?P<suffix>\S.*))?$')
+FLOAT_REGEX_COMMA = re.compile(r'(?P<number>[+-]?((((\d+(,\d*)?)|(\d*,\d+))([eE][+-]?\d+)?)|((?i:nan)|(inf))))\s*((?P<siPrefix>[' + SI_PREFIXES_INPUT + r']?)(?P<suffix>[^\s\d.,].*))?$')
 #For period as decimal separator
-FLOAT_REGEX_PERIOD = re.compile(r'(?P<number>[+-]?((((\d+(\.\d*)?)|(\d*\.\d+))([eE][+-]?\d+)?)|((?i:nan)|(inf))))\s*((?P<siPrefix>[' + SI_PREFIXES_INPUT + r']?)(?P<suffix>\S.*))?$')
+FLOAT_REGEX_PERIOD = re.compile(r'(?P<number>[+-]?((((\d+(\.\d*)?)|(\d*\.\d+))([eE][+-]?\d+)?)|((?i:nan)|(inf))))\s*((?P<siPrefix>[' + SI_PREFIXES_INPUT + r']?)(?P<suffix>[^\s\d.,].*))?$')
 
 INT_REGEX = re.compile(r'(?P<number>[+-]?\d+)\s*(?P<siPrefix>[u' + SI_PREFIXES + r']?)(?P<suffix>.*)$')
 
@@ -277,6 +277,8 @@ def siEval(s, typ=float, regex=FLOAT_REGEX_PERIOD, suffix=None, unitPower=1):
         siEval("100 μV")  # returns 0.0001
     """
     val, siprefix, suffix = siParse(s, regex, suffix=suffix)
+    if regex is FLOAT_REGEX_COMMA:
+        val = val.replace(',', '.')
     v = typ(val)
     return siApply(v, siprefix, unitPower=unitPower)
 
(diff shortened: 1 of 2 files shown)


oyvindlr:
Looks good. I originally introduced the locale-aware SI-value parsing, so I checked your PR.  The siParse bug seems to have been an oversight, while the other bug was introduced in #3546 when "\w" was changed to "\S" to allow parsing of "°".  This change allows that, and also fixes the bug introduced by that change.

j9ac9k:
This is one of those cases where we will want a test that has many many parametrized inputs, with regular expressions this complicated (especially without the use of `re.VERBOSE`) it's super easy to overlook something.

Kkkakania:
Thanks, I've expanded the coverage into a 26-case parameterized matrix. It now exercises period and comma locales across signed values, leading and trailing decimals, exponents, SI prefixes, Unicode units, and rejection of mixed or mismatched separators. The focused siParse/siEval run passes all 65 selected tests.

j9ac9k:
Thanks for expanding the combinations @Kkkakania 

Can you append those combinations to the existing `test_siParse` function?  

I know  I merged the function with `test_siParse_with_comma_as_decimal_separator` but probably should just have that be one big function that goes through all the parametrizations.

Kkkakania:
Thanks for consolidating the test cases and merging this.

PR #3573, opened 2026-08-10
Image Export: use QImage.invertPixels

pijyoi:
#3499 changed to pre-multiplied format but the image inversion code wasn't changed to take that into account. Using `QImage`'s `invertPixels` method takes care of that. It also removes the need for the endian-dependent handling code.


j9ac9k:
Any time we can get rid of custom numpy code like that in leu of functionality baked into QImage, that's a :100: from me! 

Thanks for the PR, merging.

EDIT: Github is having issues, so not merging right now, but will merge later!

PR #3574, opened 2026-08-10
Refactor HDF5Exporter for better versatility and dataset naming

### LINKED ISSUE #3572: Export into HDF5: ValueError or re-writing a file
Before this line, you should check whether the group is in the file.
https://github.com/pyqtgraph/pyqtgraph/blob/2bb369eb3c560255b6665180571df2903da1a64d/pyqtgraph/exporters/HDF5Exporter.py#L56

Or… more simply, write the file anew. I sincerely don't get why you've done it this way:
https://github.com/pyqtgraph/pyqtgraph/blob/2bb369eb3c560255b6665180571df2903da1a64d/pyqtgraph/exporters/HDF5Exporter.py#L47

Can't you unlink the file manually? Can you provide an MWE for the issue? Has the issue long been resolved on the side of `h5py`?

At least, use a context manager:
```python
        with h5py.File(fileName, 'a') as fd:  # forces append to file... 'w' doesn't seem to "delete/overwrite"
            data = []

            appendAllX = self.params['columnMode'] == '(x,y) per plot'
            # Check if the arrays are ragged
            len_first = len(self.item.curves[0].getData()[0]) if self.item.curves[0] else None
            ragged = any(len(i.getData()[0]) != len_first for i in self.item.curves)

            if ragged:
                dgroup = fd.create_group(dsname)
                for i, c in enumerate(self.item.curves):
                    d = c.getData()
                    fdata = numpy.array([d[0], d[1]]).astype('double')
                    cname = c.name() if c.name() is not None else str(i)
                    dgroup.create_dataset(cname, data=fdata)
            else:
                for i, c in enumerate(self.item.curves):
                    d = c.getData()
                    if appendAllX or i == 0:
                        data.append(d[0])
                    data.append(d[1])

                fdata = numpy.array(data).astype('double')
                fd.create_dataset(dsname, data=fdata)
```

If it's that essential to _append_ to an existing file for someone, add a parameter for those who want to write a new file. Appending is inconsistent with other formats, especially when done silently.

Currently,
```text
  File "pyqtgraph/exporters/Exporter.py", line 78, in fileSaveFinished
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pyqtgraph/exporters/HDF5Exporter.py", line 56, in export
    dgroup = fd.create_group(dsname)
  File "h5py/_hl/group.py", line 118, in create_group
    gid = h5g.create(self.id, name, lcpl=lcpl, gcpl=gcpl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5g.pyx", line 171, in h5py.h5g.create
ValueError: Unable to synchronously create group (name already exists)
```

StSav012:
* Added saving original or displayed data.
* Added an option for saving x and y data as 1D arrays, naming them from the axes.
* Added naming datasets from the names of the curves if provided.
* Made more strings translatable.
* Fixed crash on rewriting an existing file.
* Fixed crash when not a `PlotItem` is selected.
* Removed global name selection as ambiguous.

This might resolve https://github.com/pyqtgraph/pyqtgraph/issues/3572.

But please add the translations. Other exporters might need to translate `'params'`, too.

Thank you, my dear OCD, for the PR.

### DIFF
--- a/pyqtgraph/exporters/HDF5Exporter.py
+++ b/pyqtgraph/exporters/HDF5Exporter.py
@@ -69,8 +69,11 @@ class HDF5Exporter(Exporter):
         items = []
 
         if isinstance(item, GraphicsScene):
+            # why this order???
             items = [
-                _item for _item in item.items() if isinstance(_item, PlotItem)
+                _item
+                for _item in reversed(item.items())
+                if isinstance(_item, PlotItem)
             ]
         if isinstance(item, ViewBox):
             items = [
--- a/pyqtgraph/exporters/HDF5Exporter.py
+++ b/pyqtgraph/exporters/HDF5Exporter.py
@@ -96,14 +96,11 @@ class HDF5Exporter(Exporter):
 
         with h5py.File(fileName, "w") as fd:
             for index, item in enumerate(items, start=1):
-                if len(items) == 1:
-                    pd = fd
-                else:
-                    pd = fd.require_group(
-                        item.titleLabel.text
-                        if item.titleLabel.isVisible()
-                        else f"Plot {index}"
-                    )
+                pd = fd if len(items) == 1 else fd.require_group(
+                    item.titleLabel.text
+                    if item.titleLabel.isVisible()
+                    else f"Plot {index}"
+                )
                 if appendAllX:
                     for i, c in enumerate(item.curves):
                         d = (
@@ -115,7 +112,7 @@ class HDF5Exporter(Exporter):
                             continue
                         fdata = numpy.column_stack(d)
                         cname = c.name() or str(i)
-                        fd.require_dataset(
+                        pd.require_dataset(
                             name=cname,
                             shape=fdata.shape,
                             dtype=fdata.dtype,
@@ -142,7 +139,7 @@ class HDF5Exporter(Exporter):
                             else c.getData()
                         )
                         cname = c.name() or str(i)
-                        cg = fd.require_group(cname)
+                        cg = pd.require_group(cname)
                         if d[0] is not None:
                             cg.require_dataset(
                                 name=x_label,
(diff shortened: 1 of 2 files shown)


pyqtgraph/exporters/HDF5Exporter.py
```diff
@@ -2,74 +2,162 @@
...
     def parameters(self):
         return self.params
-    
+
     def export(self, fileName=None):
         if not HAVE_HDF5:
-            raise RuntimeError("This exporter requires the h5py package, "
-                               "but it was not importable.")
-        
+            QtWidgets.QMessageBox.critical(
```
pijyoi: This shouldn't work because the first argument requires a QWidget instance.
In any case, if HAVE_HDF5 is false, this exporter doesn't get registered, and therefore this method never gets a chance to be executed anyway.

StSav012: Fair point. The `parent` might be `None`, though. I'll remove the lines in several hours. Or you may.

I'm just frustrated that whenever an error occurs when exporting, only the terminal shows it. Otherwise, it's silent. I have to either check the `stderr` output or the file size to ensure the export has gone fine. It triples the time for the operation. Alright, it's a topic for another place.

j9ac9k: In other projects, I've made wrappers to capture Qt's posts to `stderr` to redirect to a logging module; which I've thought would be a welcome addition to pyqtgraph, but as pyqtgraph doesn't use the logging module, it would be odd to incorporate.  That said, adding logging would be great addition... <adds to wish list>.

j9ac9k: For reference, here is how I have done that on past projects.

https://github.com/j9ac9k/barney/blob/68375aeac8299dd0a16c2f36cb52c98f7d683469/src/barney/BarneyApp.py#L58-L95

j9ac9k:
I definitely need to get better about using `translate`.  Thanks for the PR, this LGTM, merging!

PR #3575, opened 2026-08-11
Fix ColorButton dialog palette inheritance

### LINKED ISSUE #3312: Issue: QColorDialog in ColorButton.py does not inherit parent widget's palette
Description:
In ColorButton.py, the QColorDialog is created without passing the parent widget as a parameter. As a result, the dialog does not inherit the parent's palette, leading to inconsistent appearance, especially in custom-themed applications.

Proposed Fix:
Pass the parent parameter to QColorDialog upon initialization to ensure it adopts the same palette as its parent widget.

JosephDemarest:
## Summary

- parent the color dialog to `ColorButton`
- enable Qt window palette propagation so custom palettes remain inherited
- add regression coverage for dialog ownership and palette updates

## Testing

- focused regression on PyQt6, PySide6, and PyQt5
- `pytest tests/widgets -q` (52 passed, 1 skipped)
- `mypy` (350 source files clean)

Fixes #3312

### DIFF
--- a/pyqtgraph/widgets/ColorButton.py
+++ b/pyqtgraph/widgets/ColorButton.py
@@ -22,7 +22,8 @@ class ColorButton(QtWidgets.QPushButton):
         QtWidgets.QPushButton.__init__(self, parent)
         self.padding = (padding, padding, -padding, -padding) if isinstance(padding, (int, float)) else padding
         self.setColor(color)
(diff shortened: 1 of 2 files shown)


j9ac9k:
Thanks for the PR, and I appreciate the extra test!  Merging

PR #3576, opened 2026-08-11
Simplify file dialog life in `Exporter`

StSav012:
While digging for the error handling, I stumbled upon the `Exporter` class. The IDE greeted me with a firework of warnings. So, I'm unsure whether the PR is worth it, but it's essentially up to you to evaluate.

Most of the IDE warnings were about `self.fileDialog`. So, the PR is mostly for it:

* Make it once, use it many times.
* Don't burden it with exporter-specific options.
* Use a global variable for the last directory instead of copying the value to a local var.

As a cosmetic change, I suggest making the `Exporter` class abstract not only in the description no one reads but also for Python.

### DIFF
--- a/pyqtgraph/exporters/Exporter.py
+++ b/pyqtgraph/exporters/Exporter.py
@@ -29,7 +29,12 @@ class Exporter(object):
         """
         object.__init__(self)
         self.item = item
-        
+
+        self.fileDialog = FileDialog()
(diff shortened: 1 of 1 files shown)


j9ac9k:
So much low hanging fruit in this library. This PR came in while I was waiting for something outside, and the size/scope was small enough I could review, so LGTM, merging!

PR #3577, opened 2026-08-12
Fix legend updates when plot data changes

aKa368:
Summary

Fix LegendItem labels not updating when a PlotDataItem name changes through setData().

Changes

•
Connect LegendItem entries to sigPlotChanged.

•
Update the legend label from PlotDataItem.name() when the plot changes.

•
Request a sample redraw and recalculate the legend size.

•
Add a regression test for dynamic name updates.

Testing

QT_QPA_PLATFORM=offscreen pytest -q tests/graphicsItems/test_LegendItem.py tests/graphicsItems/test_PlotDataItem.py

12 passed


pijyoi:
This PR in fact does a legend update upon every signal data update. 

Is it a common use-case to be updating the plot name during runtime?
It's common for me to be only updating the signal data.

pijyoi:
Wait, this seems to be a duplicate of #3524

PR #3583, opened 2026-08-21
Fix tons of spelling mistakes, more pyproject.toml

j9ac9k:
Started with wanting to update pyproject.toml, then decided I didn't want setup.py any more.... and then I decided to check for spelling mistakes.

Some minor type annotations checks added as I started configuring pyright.


doc/source/dictionaries/custom.dic
```diff
@@ -1,39 +1,50 @@
...
+Anderton
+apiref
+ARGB
+asnumpy
+autoLevels
+Campagnola
+cividis
+colorcet
+conda
+deffered
```
pijyoi: this is a spelling error and is probably preventing CSpell from flagging it in `ImageItem`

PR #3587, opened 2026-08-23
compile shaders using Qt instead of PyOpenGL

pijyoi:
This removes usage of OpenGL.GL.shaders from `PyOpenGL`

### DIFF
--- a/pyqtgraph/opengl/shaders.py
+++ b/pyqtgraph/opengl/shaders.py
@@ -1,11 +1,7 @@
 from OpenGL import GL
-from OpenGL.GL import shaders  # noqa
-try:
-    from OpenGL import NullFunctionError
-except ImportError:
-    from OpenGL.error import NullFunctionError
 import numpy as np
-import re
+
+from ..Qt import QtOpenGL
 
 ## For centralizing and managing vertex/fragment shader programs.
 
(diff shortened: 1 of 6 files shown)


pyqtgraph/opengl/shaders.py
```diff
@@ -234,187 +230,123 @@ def initShaders():
...
+                # for reasons that may vary across drivers, having vertex attribute
+                # array generic location 0 enabled (glEnableVertexAttribArray(0)) is
+                # required for rendering to take place.
+                # this only becomes an issue if we are using glVertexAttrib{1,4}f
+                # because that's when we *don't* call glEnableVertexAttribArray.
+                # since we always need vertex coordinates to come from arrays, it is
+                # sufficient for us to bind "a_position" explicitly to 0.
+                prog.bindAttributeLocation("a_position", 0)
+                if not prog.link():
+                    raise Exception("Program link failure:\n%s" % prog.log())
```
j9ac9k: Probably should raise a more specific Exception? Maybe `RuntimeError` ?

j9ac9k:
just rebase from master to fix the errors, sorry about that.

j9ac9k:
Thanks for the PR @pijyoi The diff looks good, appreciate nuking of the commented out code.  Would it be beneficial to have another issue with the areas that PyOpenGL is still being used (so you can check stuff off as you go)? 

Merging.

PR #3537, opened 2026-07-09
fix:  FlowchartCtrlWidget chartWidget method

dmtr-karan:
## Summary
- Rename the internal `FlowchartCtrlWidget` chart widget attribute to avoid shadowing the public `chartWidget()` method.
- Keep `chartWidget()` callable and returning the existing `FlowchartWidget`.
- Add a focused regression test for the issue.

Fixes #3294

## Tests
- `QT_QPA_PLATFORM=offscreen ./.venv/bin/python3 -m pytest tests/flowchart/test_flowchart.py::test_flowchart_ctrl_widget_chart_widget_method`
- `QT_QPA_PLATFORM=offscreen ./.venv/bin/python3 -m pytest tests/flowchart -q`
- `git diff --check`
- `QT_QPA_PLATFORM=offscreen ./.venv/bin/python3 -m pytest tests -q` *(6 failures from Qt offscreen platform warnings in unrelated GUI tests; 1096 passed, 708 skipped, 8 xfailed)*

PR #3589, opened 2026-08-23
Fixup to to #3537 

j9ac9k:
Missed on call to `.chartWidget.` that needed updating.


j9ac9k:
Bought more than I could chew with this PR, started getting Qt warnings about needing slot decorators so I added a bunch until tests passed...

PR #3596, opened 2026-08-25
Fix InfiniteLine marker position when angle isn't 0 or 90

### LINKED ISSUE #1938: InfiniteLine markers aren't positioned correctly when InfiniteLine angle isn't 0 or 90
### Short description

As I understand it, markers added to an `InfiniteLine` are intended to be positioned with respect to the extent of the `InfiniteLine` as represented on the screen. That is, a marker with `pos=0.5` should always appear to be half way along the `InifinteLine` regardless of rotation, axis limits, etc.

For `InfiniteLine`s with an angle that isn't 0 (horizontal) or 90 (vertical), the marker positioning doesn't appear to work correctly. Below is a demonstration of the issue. The marker can even move outside the plot, which should never happen if the position is in the range [0, 1].

### Code to reproduce

Not really a minimal example, but the animation helps show what's going on:

```python
import pyqtgraph as pg
import math

plt = pg.plot()

marker = ("<|>", 0.5, 10)

il_angled = pg.InfiniteLine(pos=(0, 10), angle=45, markers=[marker])
il_horiz = pg.InfiniteLine(pos=10, angle=0, pen="r", markers=[marker])
il_vert = pg.InfiniteLine(pos=5, angle=90, pen="b", markers=[marker])

plt.addItem(il_angled)
plt.addItem(il_horiz)
plt.addItem(il_vert)

plt.setXRange(0, 10)

t = 0
dt = 50
def updateRange():
    global t
    maxy = 20 * math.sin(2 * math.pi * 0.2 * t) + 35
    plt.setYRange(0, maxy)
    t += dt / 1000.

timer = QtCore.QTimer()
timer.timeout.connect(updateRange)
timer.start(dt)

pg.exec()
```

Peek 2021-07-31 11-18

### Tested environment(s)

 * PyQtGraph version: 6a59b7e5b580679f5eba4c685b6c83dd5e3b9874
 * Qt Python binding: 'PyQt5 5.15.4 Qt 5.15.2'
 * Python version: 3.9.6
 * NumPy version: 1.20.2
 * Operating system: Linux
 * Installation method: pip

### Additional context

Not really directly related, but #1878 is another issue with `InfiniteLine` triggered by an angle that's not 0 or 90 and differing x/y ranges.

aymenhafeez:
Fix marker placement for InfiniteLine items when the angle isn't 0 or 90.

Markers were previously being positioned using the line's bounding endpoints which are based on the ViewBox bounding rectangle. For non vertical or horizontal lines this can differ from the actual visible line segment and caused the marker to move as the view range changed.

The PR mirrors the pattern used in `InfLineLabel` and clips angled line endpoints to the ViewBox before positioning the markers.

Fixes #1938

### DIFF
--- a/pyqtgraph/graphicsItems/InfiniteLine.py
+++ b/pyqtgraph/graphicsItems/InfiniteLine.py
@@ -326,6 +326,32 @@ class InfiniteLine(GraphicsObject):
             self._boundingRect = self._computeBoundingRect()
         return self._boundingRect
 
+    def _getVisibleEndpoints(self):
+        left, right = self._endPoints
+        pt1 = Point(left, 0)
+        pt2 = Point(right, 0)
+
+        if self.angle % 90 != 0:
+            view = self.getViewBox()
+            if not isinstance(view, ViewBox):
+                return pt1, pt2
+
+            line = QtGui.QPainterPath()
+            line.moveTo(pt1)
+            line.lineTo(pt2)
+            line = self.itemTransform(view)[0].map(line)
+
+            bounds = QtGui.QPainterPath()
+            bounds.addRect(view.boundingRect())
+
+            paths = bounds.intersected(line).toSubpathPolygons(QtGui.QTransform())
+            if len(paths) > 0:
+                pts = list(paths[0])
+                pt1 = self.mapFromItem(view, pts[0])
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thank you for the fix @aymenhafeez ! LGTM, merging!

PR #3599, opened 2026-08-27
Fixup OpenGL

pijyoi:
1) In the original code, `GL_RGBA` (what Qt calls RGBAFormat) was used as the format. This worked for both Desktop OpenGL and OpenGL ES 2.0.

GL_RGBA was considered obsolete and was supposed to be replaced with `GL_RGBA8` (what Qt calls RGBA8_UNorm).

OpenGL ES 2.0 does not support GL_RGBA8.

2) Cache retrieval of OpenGLFunctions


j9ac9k:
Thanks @pijyoi for the cleanup!

PR #3601, opened 2026-08-29
Fix clobbering of layout method of QWidget

j9ac9k:
While experimenting with pyright, I noticed it was giving a funky error about `self.layout = ...`. On further investigation I realized that this variable assignment was clobbering the `QWidget.layout()` method, which we really should not be doing.

There is an argument to be made that we should not store the layout as an attribute of each class, and just reference it via `self.layout()` as needed, but I'm going to go with the smaller change for now.


j9ac9k:
There is a question about us breaking the public API, but we're breaking the Qt API that our users depend on so....

PR #3602, opened 2026-08-29
can instanciate an ImageView() with ImageItem() with image that is None

genevieve-le-houx:
When doing this

```python
import pyqtgraph

image_view = pyqtgraph.ImageView(imageItem=pyqtgraph.ImageItem())
```

it would fail, because in ``ImageView.__init__``, there is a condition that if the parameter ``imageItem`` is not None the code would try to call ``self.setImage(imageItem.image,, ...)``. 

But if ``imageItem.image`` is None, it would fail. ``ImageItem()`` can be initialized with``image = None`` so this feels like a bug. 

I fixed it by adding a condition that ``self.setImage(...)`` is called only if ``imageItem.image`` is not None

### DIFF
--- a/pyqtgraph/imageview/ImageView.py
+++ b/pyqtgraph/imageview/ImageView.py
@@ -181,7 +181,8 @@ class ImageView(QtWidgets.QWidget):
             self.imageItem = ImageItem()
         else:
             self.imageItem = imageItem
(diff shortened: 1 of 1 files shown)


j9ac9k:
Thanks for the PR @genevieve-le-houx  Never considered this use case, but we should absolutely support it.

PR #3606, opened 2026-09-01
Fix console history double click handling

### LINKED ISSUE #2155: ConsoleWidget history error
There are two issues here but I think they are probably related. 

to reproduce, It may be helpful to set `c.ui.historyList.setAlternatingRowColors(True)` in consoleWidget.py 

**First Issue:**
  - run consoleWidget example 
  - enter 1 in the input line and press enter
  - open the history
  - double click the first row in the history list 2 times
  - double click the third row

**Second** Issue:
  - run consoleWidget example 
  - enter 1 in the input line and press enter
  - open the history
  - double click the first row in the history list once
  - double lick the second row twice ( notice the blank line in the output)
  - an empty 3rd row was added to the historyList, double click it
  
at this point, double clicking the 4th row causes an index error.
  
Traceback for both issues:
  ```
  ===== 2021.12.15 11:12:56 =====
  File "/Users/chris/projects/pyqtgraph/pyqtgraph/console/Console.py", line 275, in cmdDblClicked
    self.input.setHistory(index)
  File "/Users/chris/projects/pyqtgraph/pyqtgraph/console/CmdInput.py", line 40, in setHistory
    self.setText(self.history[self.ptr])
```
I assume the intended behavior for double clicking any row in the historyList would be to execute that code again and to never add an empty row unless the row was intentionally empty(from pressing enter in an empty input)

aymenhafeez:
Fixes #2155 

`ConsoleWidget` was mapping clicked `historyList` rows back into `CmdInput.history` using negative indexing which would cause an IndexError if the last item in the history was double clicked since the two list indexing isn't aligned. The fix uses the clicked list item text directly when selecting or double clicking history rows.

Tested the steps outlined in the original issue to reproduce the error, the error no longer occurs and double clicking items in the history behaves as expected.

### DIFF
--- a/pyqtgraph/console/Console.py
+++ b/pyqtgraph/console/Console.py
@@ -159,13 +159,11 @@ class ConsoleWidget(QtWidgets.QWidget):
             return self.localNamespace
 
     def cmdSelected(self, item):
-        index = -(self.historyList.row(item)+1)
-        self.input.setHistory(index)
+        self.input.setText(item.text())
         self.input.setFocus()
         
     def cmdDblClicked(self, item):
-        index = -(self.historyList.row(item)+1)
-        self.input.setHistory(index)
+        self.input.setText(item.text())
         self.input.execCmd()
         
     def _stackItemDblClicked(self, handler, item):


j9ac9k:
Thanks for the PR @aymenhafeez  

LGTM!

PR #3607, opened 2026-09-02
refactor upload_vbo into OpenGLHelpers

pijyoi:
The code was originally duplicated because there wasn't a defined place to park common code.


j9ac9k:
Love the continued improvements to this part of the codebase.  Thanks for your continuous efforts here @pijyoi

PR #3608, opened 2026-09-04
fix: stop ConsoleWidget help() entering interactive mode

### LINKED ISSUE #3247: pyqtgraph console becoming unresponsive when entering help()
### Short description
If you enter help in the console it prompts you to "Type help() for interactive help, or help(object) for help about object."
If you do enter `help()` it becomes unresponsive.

### Code to reproduce

Start the examples using
`python -m pyqtgraph.examples`
Select the Console example.
Enter `help()`

### Expected behavior

Some sort of help dialog, equivalent to what you get when typing `help()` in an interactive python console.

### Real behavior

The example is becoming unresponsive.
If the console is used in a more complex program, this bug causes the entire program to become unresponsive.

### Tested environment(s)

 * PyQtGraph version: 0.13.7
 * Qt Python binding: PyQt6 6.6.0 Qt 6.6.0
 * Python version: 3.11.0
 * NumPy version: 2.2.2
 * Installation method: pip

aymenhafeez:
Fixes #3247 

The ConsoleWidget would become unresponsive if `help()` was entered as it would try to enter interactive help mode but the console disables input while execution is active. The PR is a bit of a hack and just prints pydoc's intro text on `help()` instead of trying to enter an interactive session. Entering `help(object)` still works unchanged.

A fix to allow properly entering the interactive help would need to add command execution handling in a way that doesn't block the Qt event loop while it's being executed.

In the original issue discussion it was mentioned that the QtConsole is favoured over the ConsoleWidget so i feel this simple fix is probably fine? But happy to reconsider

### DIFF
--- a/pyqtgraph/console/repl_widget.py
+++ b/pyqtgraph/console/repl_widget.py
@@ -1,4 +1,5 @@
 import code
+import pydoc
 import queue
 import sys
 import traceback
@@ -232,6 +233,12 @@ class ReplThread(QtCore.QThread):
                 return
 
             self._commandBuffer = []
+
+            if fullcmd.strip() == "help()":
+                with self._stdoutInterceptor:
+                    pydoc.Helper(output=sys.stdout).intro()
+                    self.sigCommandExecuted.emit()
+                return
 
             # run command
             try:


j9ac9k:
Thanks for the PR @aymenhafeez 

This LGTM; merging!

PR #3611, opened 2026-09-10
Undo removal of GraphicsLayout.layout

j9ac9k:
...but add a deprecation warning.

The change in GraphicsLayout did not leave end users with a way to write code compatible with the next release of pyqtgraph and previous versions.  

To anyone that sees this, know that this deprecation period will likely be (relatively) short.


janezd:
What will happen after the deprecation period? How will we write downstream code compatible with both, old and new versions of pyqtgraph?

The current attribute `layout` shadows the inherited method and stores a Qt property as an attribute; I agree that this is bad. But fixing this will -- in my opinion -- make the situation worse because it will make `layout` useless, and will require adding helper functions like

```
    def getLayout(self):
        if isinstance(self.layout, Callable):
            return self.layout()
        else:
            return self.layout
```

This attempt to make the code cleaner on the pyqtgraph side will make it messier in the downstream code.

The other solution is, of course, to raise the required version of pyqtgraph on downstream, but this is something we try to avoid whenever possible because any pinning increases the risk of unresolvable conflicts.

I would suggest de-deprecating the use of `layout` as property. Sometimes you have to live with the sins of your past.

j9ac9k:
We will be keeping the property `layout_` to give access to the layout, which will work moving forward indefinitely (there prescience for doing this with attribute name conflicts, see `.exec_`).  But as it is planned to stay around indefinitely, I should likely add it to the docs as part of the public API.

I'm not sure what kind of deprecation period we are talking about here, but as this should really only impact GraphicsLayout, I'm okay drawing it out for a while.

PR #3618, opened 2026-09-12
Always send hoverLeaveEvents before hoverMoveEvents and hoverEnterEvents.

### LINKED ISSUE #3465: Hover leave events are not consistently called before adjacent enter events.
### Short description
Hover leave events are not consistently called before adjacent objects hover enter events.

I have overridden the setMouseHover method of a LinearRegionItem and both its' lines to set/restore different override cursors. When i move the mouse from outside the region to the inside the methods are called in the following order:

1. InfiniteLine (enter)
2. LinearRegionItem (enter)
3. InfiniteLine (leave)

This caueses the LinearRegionItems' override cursor to never be shown because it is immediately reset by the InfiniteLines's leave event.

### Expected behavior
Hover leave events are always processed before the next items' enter event.

### Real behavior
A second enter event may be processed before the previous items' leave event.

### Tested environment(s)

 * PyQtGraph version: 0.14.0
 * Qt Python binding: PySide 6.10.2
 * Python version: 3.14.2
 * NumPy version: 2.4.2
 * Installation method: pip

maflAT:
`GraphicsScene.sendHoverEvents` now sends all exit events in reverse enter-order, before sending move and enter events in z-order.

Fixes https://github.com/pyqtgraph/pyqtgraph/issues/3465


aymenhafeez:
Hi thanks for working on this

I tested the PR with the MRE from #3465 and it still prints 'Line entered' before 'Region left'. Also the print statements are coming from `setMouseHover()` which represents what first becomes draggable rather than `HoverEvent.isEnter()`/`isExit()` events. Printing on enter/exit events shows the region isn't actually being left/entered when crossing the overlapping line.

Am i missing a different path the PR is trying to fix, and could you double check the before and after behaviour to confirm that the PR is achieving its intended purpose?

PR #3625, opened 2026-09-13
Gradual Typing - Annotations for GraphicsItem and some frequently used Functions

maflAT:
I started adding type annotations. This is obviously a work in progress.
I began with the color-related functions in the functions module and with GraphicsItem as one of the more base classes.

The public API of GraphicsItem should now be completely annotated, unless I missed something.


maflAT:
Whoops. Looks like I deleted something, I shouldn't have...

PR #3632, opened 2026-09-15
Fix / document color_like annotations.

### LINKED ISSUE #3631: mkColor() does not behave as expected for Tuples of Float.
### Short description
`mkColor((float, float, float))` does not behave as expected.

It does not interpret the values as a range between 0 and 1.0. Instead, it just silently converts them to int.

The same behavior applies to `mkColor(float, float, float)`, `mkColor(float, float, float, float)` and `mkColor((float, float, float, float))`.

### Expected behavior
`mkColor((1.0, 1.0, 1.0))` == 0xFFFFFF

### Real behavior
`mkColor((1.0, 1.0, 1.0))` == 0x010101

### Suggestions
I would suggest not changing any code but either:

- removing the misleading type hints from `color_like`, or
- clearly document the actual behavior,
- both.

In a next step we should search for wrong documentation or worse, implementation, based on this parameter.
For example, I stumbled over this one:
```python
class GLGraphItem(GLGraphicsItem):

    ...

    def setData(self, **kwargs):
        """
        Change the data displayed by the graph. 

        Parameters
        ----------
        edges : np.ndarray
            2D array of shape (M, 2) of connection data, each row contains
            indexes of two nodes that are connected.  Dtype must be integer
            or unsigned.
        edgeColor: color_like, optional
            The color to draw edges. Accepts the same arguments as 
            :func:`~pyqtgraph.mkColor()`.  If None, no edges will be drawn.
            Default is (1.0, 1.0, 1.0, 0.5).
        edgeWidth: float, optional
            Value specifying edge width.  Default is 1.0
        nodePositions : np.ndarray
            2D array of shape (N, 3), where each row represents the x, y, z
            coordinates for each node
        nodeColor : np.ndarray or float or color_like, optional
            2D array of shape (N, 4) of dtype float32, where each row represents
            the R, G, B, A values in range of 0-1, or for the same color for all
            nodes, provide either QColor type or input for 
            :func:`~pyqtgraph.mkColor()`
        """
        ...
```

maflAT:
Fixes #3631

- Remove misleading annotations for tuple of floats.
- Add comments regarding expected values. 
- Add a docstring.
- Convert TypeAlias to Python 3.12 type statement.


j9ac9k:
Thanks for the PR @maflAT 

With TypeAlias being deprecated (and I'm fairly sure I didn't have things right to begin with), this change is very much welcome.

PR #3635, opened 2026-09-17
New Feature: handling multiple changes in Parameter sigTreeStateChanged emission

seb5g:
## Change Transaction in Parameter sigTreeStateChanged signal and its callback

Add change_transaction context manager for atomic and silent parameter tree updates
## Description
This pull request introduces a new change_transaction context manager on the Parameter class in pyqtgraph.parametertree.Parameter.
When performing multiple updates on a parameter or its children (such as updating allowed choices via setLimits followed by selecting a valid value via setValue), Qt widgets like QComboBox frequently trigger intermediate, cascading signals and transient state changes. This results in redundant multiple notifications or unintended side effects in user callbacks connected to sigTreeStateChanged.
The change_transaction context manager addresses this by:

   1. Safely blocking tree-wide change signals at the root level and the targeted emitter level during batch modifications.
   2. Clearing any transient event queues accumulated by Qt widgets during the execution of the block.
   3. Automatically detecting the final change type or accepting explicit payload attributes (change_type, value, emitter).
   4. Supporting a fully silent operation mode when emit_signal=False is requested.
   5. Maintaining full backward compatibility with older Python versions via a native fallback for StrEnum.

## Changes Made

* Added ParameterChangeType enumeration in a dedicate enums module with a Python < 3.11 fallback implementation of StrEnum. Allows to simply/safely import it
* Implemented the @contextmanager def change_transaction(...) method inside the Parameter class.
* Added a comprehensive test class TestChangeTransaction within the parameter test suite covering baseline cascading behaviors, single-emission resolution, quiet transactional updates, custom emitter routing up the node tree, and structural edits like adding children, changing names, and modifying options.

## Example Usage

```py
with param.change_transaction(ParameterChangeType.VALUE):
    param.setLimits(['Option A', 'Option B'])
    param.setValue('Option A')
```

will only trigger a single "value" change type in the callback connected to sigTreeStateChanged!

PR edited with the help of Google Gemini


Ashwolaa:
I'll look for it but have you checked treeChangeBlocker? The application looks fairly close to what you are describing.

pyqtgraph/parametertree/enums.py
```diff
@@ -0,0 +1,19 @@
+try:
+    from enum import StrEnum
+except ImportError:
+    # Fallback for Python < 3.11 to maintain strict compatibility without dependencies
+    from enum import Enum
+    class StrEnum(str, Enum):
+        pass
```
Ashwolaa: pyqtgraph mentions that it only considers 3.12+ so I would not bother keeping it

seb5g:
yes I did but there is a fundamental difference. TreeChangeBlocker is not really blocking the changes, its putting them together to be fired all at once later one. This means you'll have in your callback many many intermediate calls not meant. For instance, changing limits and setting a value may call you callback many many time, creating a bunch of unwanted side effect and where you cannot be sure of the change order. 

This is fundamentally different of blocsignal or blockslots you've worked on. 

Here you control, your change flow, either letting one change go through at the end or no at all!

Ashwolaa:
But here you still get all the signals, you are collecting everything, deleting everything / making an empty list, then manually creating the signal of interest.
I would be more inclined to just rely on treeChangeBlocker and extend it to filter the desired signals in the blocked signals.
In that case, the signals are already blocked from the param itself (self) and you choose which ones get emitted.

In that case no need to introduce root, you just work with all the accumulated signals on the parameter being blocked. 
The final emitter can then be given as an additional arg (if you want to make it start from top root for example)
```
def treeChangeBlocker(self, keep=None, dedupe=False, emitter=None):

        def exitFn():
            self.blockTreeChangeEmit -= 1
            if self.blockTreeChangeEmit == 0:
                changes = self.treeStateChanges # Get all changes that happened on self directly 
                self.treeStateChanges = [] # Reseting treeStateChanges
                if keep is not None or dedupe:
                    changes = coalesceTreeChanges(changes, keep=keep, dedupe=dedupe) # Filter changes and remove duplicates
                if len(changes) > 0:
                    target = self if emitter is None else emitter  # Pick emitter, self as default
                    target.sigTreeStateChanged.emit(target, changes) # Emiter fires selected changes

        return SignalBlocker(self.blockTreeChangeSignal, exitFn)
```

This coalesceTreeChanges (not sure about the naming) is composed of a filter step (keep) and a way to filter similar changes (dedupe). In that context, it only collects the last one.
```
def coalesceTreeChanges(changes, keep=None, dedupe=False):
    """
    Reduce a list of (param, changeType, data) tuples as delivered by
    sigTreeStateChanged / accumulated in Parameter.treeStateChanges.

    keep    : optional set of changeType strings to retain (e.g. {'value', 'limits'}).
              None keeps every change type.
    dedupe  : if True, only the last entry for each (param, changeType) pair is kept,
              in the position it last occurred. Dict-shaped payloads (e.g. 'options'
              fragments like {'readonly': True}) are merged key-by-key instead of being
              replaced outright, since each such fragment only carries the keys that
              changed in that one call, not the full options state.
    """
    #Filter changes    
    if keep is not None:
        changes = [c for c in changes if c[1] in keep]
   #Remove duplicates
    if not dedupe:
        return changes
    last = {}
    order = []    
    for param, changeType, data in changes:
        key = (param, changeType)
        if key not in last: 
            order.append(key)
            last[key] = data 
        elif isinstance(data, dict) and isinstance(last[key], dict): # Managing dict and options for this case
            last[key] = {**last[key], **data}
        else:
            last[key] = data 
    return [(p, t, last[(p, t)]) for (p, t) in order]
```

Overall, you get a similar behavior with in my opnion.

  ```python
with p.treeChangeBlocker(keep={"value"}):
        p.setLimits(["a", "b", "c", "d"])
        p.setValue("b")
# p emits: [(p, 'value', 'b')]

with p.treeChangeBlocker(dedupe=True, emitter=root):
        p.setValue("b")
        p.setValue("c")
# root emits: [(p, 'value', 'c')]

with p.treeChangeBlocker(dedupe=True):
      p.setValue('b')
      p.setValue('a')                 
      p.setLimits(['a', 'b', 'c'])    

  # p emits: [(p, 'value', 'a'), (p, 'limits', ['a', 'b', 'c'])]

with p.treeChangeBlocker(dedupe=True):
    p.setOpts(readonly=True)
    p.show(False)
   # p emits : [[(p, 'options', {'readonly': True, 'visible': False})]] 

with p.treeChangeBlocker(dedupe=True):
    p.setOpts(readonly=True)
    p.setOpts(readonly=False)
#  p emits: [[(p, 'options', {'readonly': False})]]
```

Lastly, the filtering step can be improved if needed. Here it only relies on standard keyword but you could extend it to filter specific child parameter. So from this
```python
with group.treeChangeBlocker(keep={'value'}):
    child_a.setValue(1)
    child_b.setValue(2)
    child_a.setLimits([...])
# group emits  [(child_a, 'value', 1), (child_b, 'value', 2)]
```
One could extend a bit the keep (not shown above here):
```
if keep is not None:
    pred = keep if callable(keep) else (lambda p, t: t in keep)
    changes = [c for c in changes if pred(c[0], c[1])]
```
Then you could even have. 
```python
with group.treeChangeBlocker(keep=lambda p, t: p is child_a and t == 'value'):
    child_a.setValue(1)
    child_b.setValue(2)
    child_a.setLimits([...])
# group emits [(child_a, 'value', 1), ]
```

You can check it directly on https://github.com/Ashwolaa/pyqtgraph/tree/feature/upgrade_treeChangeBlocker

seb5g:
I actually prefer your solution that builds directly on top of treechangeblocker! I modified a bit the tests and proposed a new PR #3636 and I'm closing this one!

PR #3636, opened 2026-09-18
Feature/upgrade tree change blocker

seb5g:
Extending TreeChangeBlocker to filter the finally emitted signals, eventually None or merging some of them into one global change.

Replace #3635 with an extension of TreeChangeBlocker, see #3635 discussion for the details!


pyqtgraph/parametertree/enums.py
```diff
@@ -0,0 +1,19 @@
...
+    """Enumeration of native change types emitted by PyQtGraph Parameters."""
+    VALUE = "value"
+    LIMITS = "limits"
+    OPTIONS = "options"
+    CHILD_ADDED = "childAdded"
+    PARENT = "parent"
+    NAME = "name"
+    READ_ONLY = "readOnly"
+    VISIBLE = "visible"
+    CONTEXT_MENU = "contextMenu"
```
Ashwolaa: I am not so convinced on the added value of it. Do you think it is important to have this ENUM list? 
If yes, then should this be propagated throughout parameter codebase?

Ashwolaa: I think I would just drop this part in fact

tests/parametertree/test_Parameter.py
```diff
@@ -594,6 +595,158 @@ def a():
...
+    def test_treeChangeBlocker_keep_filters_change_types(self):
+
+        with self.root.treeChangeBlocker(keep={ParameterChangeType.VALUE}):
+            self.root.child('group', 'p').setLimits(["a", "b", "c", "d"])
+            self.root.child('group', 'p').setValue("b")
+
+        assert len(self.events) == 1
+        assert self.events[0] == [(self.param, "value", "b")]
+
+    def test_treeChangeBlocker_keep_filters_change_types(self):
```
Ashwolaa: you used the same function name as above. Either merge the two or change it?

pyqtgraph/parametertree/enums.py
```diff
@@ -0,0 +1,19 @@
+try:
+    from enum import StrEnum
+except ImportError:
+    # Fallback for Python < 3.11 to maintain strict compatibility without dependencies
```
j9ac9k: no need to support python < 3.11, pyqtgraph only supports 3.12+ as of right now.

seb5g: good to know! thx

seb5g:
I removed the enum entirely from this PR to not mix things. I 'll do another one dedicated to it

Ashwolaa:
> I removed the enum entirely from this PR to not mix things. I 'll do another one dedicated to it

You still need to change one function in your tests as both share the same name. See comment above and the github-advanced-security

tests/parametertree/test_Parameter.py
```diff
@@ -594,6 +594,158 @@
...
+        # one signal from self carrying every real change, uncoalesced.
+
+        with self.root.treeChangeBlocker():
+            self.param.setValue("b")
+            self.param.setValue("c")
+
+        assert len(self.events) == 1
+        assert self.events[0] == [(self.param, "value", "b"), (self.param, "value", "c")]
+
+    def test_treeChangeBlocker_keep_filters_change_types(self):
```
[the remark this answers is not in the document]
seb5g: changed the names:

Ashwolaa:
All good on my side, thanks!

